PR #4264, opened 2021-03-01
Return to confirm screen in batch payments

### LINKED ISSUE #4260: batch new payments return to confirm payments on submit
### Describe the bug
Not a bug as I intended to patch that workflow out simply because the confirm screen is ambiguous.
Dropping through from the new payment submit provides the same info but still allows editing.
But, I don't care as in the end one user will like and another will not. 
As it's legacy, i'll return.

Also will add some niceties like fix table styles to fit screen and auto select payment amount so user doesn't have to delete content before entering payment amount... annoying for sure and whatever else...

sjpadgett:
- restyle insurance and patient select popups
- fix new payment table to fit screens better

Fixes #4260 #4265

### DIFF
--- a/interface/billing/edit_payment.php
+++ b/interface/billing/edit_payment.php
@@ -535,6 +535,9 @@ $ResultSearchSub = sqlStatement(
         document.onclick = HideTheAjaxDivs;
     </script>
     <style>
+        .amt_input {
+            max-width: 75px;
+        }
         .bottom {
             border-bottom: 1px solid var(--black);
         }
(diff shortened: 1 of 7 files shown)


sjpadgett:
image
image

Also added auto select for Payment Amount so user doesn't have to delete 0.00 to enter payment. how annoying that must be.
image

Would have gotten rid of the "Do you want to Allocate" confirm dialog after clicking the Allocate button(why yes I do. I did click the allocate button after all!:)) but i'll be darn if I could find the darn thing....

bradymiller (approved):

bradymiller:
tenor99

sjpadgett:
@stephenwaite Do you think we should remove that annoying "Do you want to Allocate" confirm after Allocate button click?
I keep going back and forth on it!

stephenwaite:
don't know @sjpadgett , never use it and the community doesn't seem to complain too much about it, thanks for fixing this up

PR #4270, opened 2021-03-06
medex icon sql fix

bradymiller:
@ophthal , noted this discrepancy in tags and wanted to make sure this is what you meant on this


bradymiller:
hi @ophthal , Since rushing to get a patch out, I'm going to bring this minor PR in. please let me know if this change is not ok (then i would revert it). thanks -brady

ophthal:
This one is good.
The save.php has a needed fix for the next patch.
The flow board and recall board bs4 overhaul looks good, just some tweaking to make it flow visually.  The manila style got hammered, but it is coming back too.

PR #4287, opened 2021-03-17
add password max length setting

bradymiller:
add password max length setting


bradymiller:
testing well and bringing in

stephenwaite:
excited_elephant

PR #4288, opened 2021-03-17
fix related to google sign in

bradymiller:
fix related to google sign in

This fixes error when trying to add a new user with a blank google sign in email address. If you already have a user with a blank google sign in email address then it will break. This fixes it by converting blank to null, which does not need to be unique in the database field.


bradymiller:
working well and bringing in

growlingflea:
Good catch, Brady!

PR #4295, opened 2021-03-20
Fix sql error with MySQL 8 when viewing event log.

sunsetsystems:

Fixes SQL error "Incorrect DATETIME value: ''" when viewing the event log using MySQL 8 (and perhaps other versions).


bradymiller (approved):

bradymiller:
looks good. will make sure this also gets into the next 6.0.0 patch.

PR #4319, opened 2021-04-01
Fix to custom filter (via module/event)

kchapple:
Fix to custom filter (via module/event) that allowed searches to bypass filter

Fixes #


bradymiller:
hi @kchapple , code looks good and bringing in. thanks for the fix! -brady

PR #4329, opened 2021-04-07
Fixing import bug in backup.php.

sunsetsystems:
Configuration Import was trying to use mysqldump instead of mysql, causing it to fail. This fixes that.

Should be applied to 6.0.0 as well.


bradymiller (approved):

bradymiller:
@sunsetsystems , thanks for the fix! (I just brought your commit into rel-600 also :) )

PR #4333, opened 2021-04-09
Add events on before-patient-create and before-patient-update

kchapple:
Adds two events: before a patient is created and before a patient is updated. This is helpful to do any extra processing to DB fields before they are inserted into the database. 

We are using this to dynamically generate a PUBPID that is based on a combination of sources. We can listen to the event, and in our module, calculate the PUBPID before it is inserted into the database.


bradymiller:
hi @kchapple , Looks good and bringing in. thanks for the improvements! -brady

PR #4337, opened 2021-04-11
Portal

sjpadgett:
- convert base64 data url image to svg to support AWS
- fix signature meta data
- turn off delete in dashboard.


bradymiller:
2DV (1)

PR #4340, opened 2021-04-15
Dispatch system for the inventory.

juggernautsei:
Fixes #

This is to decouple the inventory system and take advantage of the Symfony event dispatcher. 

Modernizing the inventory system.

### DIFF
--- a/Billing_process_refactor_and_enhancements_(#4185).patch
+++ b/Billing_process_refactor_and_enhancements_(#4185).patch
@@ -1,0 +1,6157 @@
+Index: controllers/C_X12Partner.class.php
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
+diff --git a/controllers/C_X12Partner.class.php b/controllers/C_X12Partner.class.php
+--- a/controllers/C_X12Partner.class.php	(revision 2375db0f8f2e345f1dd36e5c556db30b312fdcc3)
++++ b/controllers/C_X12Partner.class.php	(revision e3fa29dc6ec78e3ed2bdbb5539a5aa281b37a2a7)
+@@ -1,5 +1,19 @@
+ <?php
+ 
++/**
++ * controller class for x-12 partner screen
++ *
++ * @package   OpenEMR
++ * @link      http://www.open-emr.org
++ * @author    Ken Chapple <ken@mi-squared.com>
++ * @author    Daniel Pflieger <daniel@mi-squared.com>, <daniel@growlingflea.com>
++ * @copyright Copyright (c) 2021 Ken Chapple <ken@mi-squared.com>
++ * @copyright Copyright (c) 2021 Daniel Pflieger <daniel@mi-squared.com>, <daniel@growlingflea.com>
++ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
++ */
++
++use OpenEMR\Common\Crypto\CryptoGen;
++
+ class C_X12Partner extends Controller
+ {
+ 
+@@ -10,7 +24,7 @@
+     function __construct($template_mod = "general")
+     {
+         parent::__construct();
+-        $this->x12_partner = array();
++        $this->x12_partners = array();
+         $this->template_mod = $template_mod;
+         $this->assign("FORM_ACTION", $GLOBALS['webroot'] . "/controller.php?" . attr($_SERVER['QUERY_STRING']));
+         $this->assign("CURRENT_ACTION", $GLOBALS['webroot'] . "/controller.php?" . "practice_settings&x12_partner&");
+@@ -32,6 +46,12 @@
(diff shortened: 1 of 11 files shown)


juggernautsei:
Here is the updated code for the inventory dispatch. The module was suggested by @adunsulag. The purpose is to register the inventory subscriber with the dispatcher as to which specific events the subscriber is listening for in the system. @sunsetsystems 
Now, can you assist with your thoughts on what can be the next steps?

sunsetsystems:
As I mentioned before, taking from inventory is an interactive operation (for example it can fail). Also work will need to be done to associate vaccines with inventory items and to check and select from inventory in the course of dispensing. So I'm not sure a model of dispatching an event without acting on an immediate response will work.

More generally, is there consensus within the community that moving towards Symfony/Laravel is desirable? I've not been so active lately so am a bit out of touch on that, and won't be strident as to my own views.

adunsulag:
@sunsetsystems So I don't believe the community is attempting to move towards the entire Symfony framework.  I know several of us have been using the Symfony EventDispatcher library to try and decouple pieces of the core system from addons, vendor specific code, and non-core functionality.  The goal is to allow vendors and developers to respond to core OpenEMR functionality via events in the system.  @sjpadgett and I designed the OpenEMR fax module as an example of this.  @juggernautsei approached me with the desire to do something along those lines with the inventory system.  I believe the expressed goal is to create a system that is decoupled that registered modules (3rd party or otherwise) can then add / remove inventory as needed.

In terms of architecture design, if the process is interactive as @sunsetsystems  is saying I would suggest if this is the route you want to go with your architecture that you send back data in your events.  One option would be to allow the inventory system to send back a dialog URL it needs opened to collect / process additional user information.

If the user interaction will be consistent for all 3rd party inventory modules then you could have the system collect whatever information is needed for the interaction and then send all of that as part of your event to whatever registered module is listening, process the response and have the event pass back whether the inventory change was successful or whether it failed and you can act accordingly.  

I don't know this inventory domain enough to really give you good feedback on how to do the detailed design of what you are wanting, but those are just some thoughts based on what @sunsetsystems has stated.  After all is said and done perhaps the event model may not be appropriate here, that'll be up to you @juggernautsei on the design requirements and the goals you are trying to accomplish here.

sunsetsystems:
Thanks @adunsulag, that's good insight. To @juggernautsei I suggest looking at existing code that calls sellDrug() to see how the process works and why it is interactive. Also see my notes in the previous PR about the desirability of relating CVX codes to drugs.

juggernautsei:
@sunsetsystems I am studying the drug code to learn how it works and where it is connected. My original intent was just to use a better way to have the inventory to include vaccines. I didn't want to do it the old way by just doing an include, 
@adunsulag I'll try to figure out what you are referring to. @bradymiller what is the long range prospect for Symfony being a permanent part of the codebase like Zend framework. I did have the thought that the entire inventory could be over time migrated to the module. When I rewrite the Weno code, I will write it as a module. That is down the road a ways. For right now, I was trying to increase my skill level. Symfony and Laminas have been around as long as OEMR has. I am glad that I learned the free style coding that OpenEMR offers and I am not locked into a framework per say.   For now, I just want to get the inventory to include the immunizations. I will stay focused on that and see where it goes from there. @adunsulag. I don't know much about the adding user input but it might be worth persuing. I thought this would be a behind the scenes thing that takes place without user interfacing with it. I don't yet see why the user would need to add anything other than like you said getting a notice maybe. It really impacts the inventory manager most, not the person recording the vaccine. The inventory manager needs an accurate count of what has been dispensed. @mdsupport

juggernautsei:
A wise man has many counselors.

sunsetsystems:
Yes I think inventory should be able to include vaccines, however inventory tracking has always been optional so we should continue to allow vaccinations without requiring inventory. Perhaps a global setting is in order.

juggernautsei:
That would be wise to add the global since I can only see the request that is being made. It is the COVID vaccine tracking that brought this inventory need to light. The user noticed that as they were recording shots that the QOH was not decreasing and of course the EU thinks it should. But I see your point that some may not want vaccines to change the QOH.

mdsupport:
@juggernautsei, If you look under the covers, vendor/autoload mechanism enables several hundred includes. So includes by themselves are not a bad coding practice. What is being included decides the maintainability of the code.

Inventory has limited integration points. So unless you are considering this PR as a coding exercise, as long as your code ensures SSOT one can live with "imperfection".

adunsulag:
For anyone else following this thread who may not be familiar with SSOT it means ['Single Source of Truth'](https://en.wikipedia.org/wiki/Single_source_of_truth).

mdsupport:
> COVID vaccine tracking that brought this inventory need to light

Not so. Since 4.x it is identified by every clinic after entering their first transaction as a show stopper.

juggernautsei:
Lost funding for this project. I will have to pick it up another time in the future on using this as a springboard on another project.

PR #4341, opened 2021-04-15
Rel 600 open layouts and lists in own tabs

growlingflea:
Fixes #

This week I was creating a bunch of really big LBF's at a customer's request and I found that having separate tabs for layouts and lists really made a difference.  This functionality isn't really used often but the small change really sped things up.


bradymiller:
hi @growlingflea , This PR doesn't look right. Looks like your code may be based on another branch rather than master.

growlingflea:
I thought I cherry-picked it.  I will make changes then resubmit.

PR #4348, opened 2021-04-22
Enhances the billing manager look.

juggernautsei:
Fixes #

Adds contrast to the billing manager

add contrast to the billing manager for readability

### DIFF
--- a/interface/billing/billing_report.php
+++ b/interface/billing/billing_report.php
@@ -532,6 +532,10 @@ $partners = $x->_utility_array($x->x12_partner_factory());
             border-top: none !important;
         }
 
+        .table tr:nth-child(even) {
+            background-color: rgba(82, 208, 204, 0.82) !important;
+        }
+
         a,
         a:visited,
         a:hover {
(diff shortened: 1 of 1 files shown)


juggernautsei:
Here is what this does.

image

stephenwaite:
hi @juggernautsei , the light theme is currently easy on the eyes what theme are you using here?

juggernautsei:
This mod overrides any theme. So, it does not matter the theme. The billing manager should always remain the same. I had someone ask me to fix the visual issue. They were watching the billing video of 4.x and the screen looked  like this
image

I noticed there was still CSS in the billing report so I took advantage of that and let it overwrite the theme. @benmate suggested what to do.

stephenwaite:
here's how it looks in dev demo with light theme
Screenshot from 2021-04-22 09-21-59

juggernautsei:
Right, and the colors are so faint. The lack of contrast makes it difficult to read. The complaint is being able to know where one claim stops and the next begins. The current color scheme makes scrolling through difficult for some people.

stephenwaite:
think it could be done [here](https://github.com/openemr/openemr/blob/9d3ff0974020488b64433ff2361d53b7f5d89b5f/interface/billing/billing_report.php#L1015) as well
wonder why we're not bootstrapping this value

juggernautsei:
I tried there and it did not produce the desired results. The reason that I was given is the spaghetti code. There is no clear separation of logic from the display code. There are too many trs being created for unclear reasons.

bradymiller:
wow, this is definitely a messy css script. Would be nice to clean this up a bit so things are more straightforward. Something i noted that also makes things look nicer is removing the `table-hover` here:
https://github.com/openemr/openemr/blob/9d3ff0974020488b64433ff2361d53b7f5d89b5f/interface/billing/billing_report.php#L886

juggernautsei:
@bradymiller taking out the hover does improve the UIX.

bradymiller:
@juggernautsei , check out this proposed fix: https://github.com/openemr/openemr/pull/4353

PR #4349, opened 2021-04-22
Openemr fhir search

### LINKED ISSUE #4347: FHIR US Core Search Implementation
The current FHIR US Core Search Implementation is not working correctly.  Search currently does just the union of values rather than explicit intersection which FHIR requires.  Only string values with limited token value support is working.  Search with conditionals and modifiers is also mostly missing with encounters being the only thing that does date searching and that is incorrect.

I'm currently building a new search algorithm for US Core Search with FHIR.  The attempt is to continue to follow our 3 tier layer approach.  With the FHIR layer sitting on top of the OpenEMR business service layer which sits on top of the SQL data layer. 

I've gone ahead and implemented string, date, and a limited version of token.  All of the FHIR modifiers for string are now supported as well as all of the date/datetime comparators (gt,lt, etc) except for aproximate which I don't think is a good thing we want in the system right now.  

The algorithm is built so that hopefully only minimal work will need to be done in each of the FHIR and business service classes to do searching is required.  I've gone ahead and heavily modified the FhirPatientServiceQueryTest to show the robustness of the types of data values that can be searched on.

adunsulag:
Fixes #4347 

This is still a work in progress but I'm creating the pull request to give others insight into where I'm going with the FHIR search. I've described my approach in more detail in the issue and you can read about it more.

To get a good feel for how this all works I recommend walking down the code path of FhirPatientRestController->getAll()  That calls the FhirServiceBase->getAll() method which I've completely revamped.  The method in the FhirServiceBase instantiates a search parameter factory class that is responsible for building Search Field objects representing each search term will be searched on for the Service class.  It uses Search Definition objects that are defined in each service class to know how to construct those fields.  The work of that is done in FhirServiceBase->createOpenEMRSearchParameters() and inside of FHIRSearchFieldFactory.

This class also will force each FhirService class that implements the IPatientCompartmentResourceService to bind the patient uuid context (if we are working with the patient/*.read|write scopes) to the search parameters.

After constructing each of the SearchField objects the base class calls the individual service's searchForOpenEMRRecords which for our example would be in FhirPatientService.  The service class can provide any specific field handling that's needed that the parent class can't deal with.  In this case we have to convert our FHIR gender codes for a patient to match the internal OpenEMR value rather than the FHIR search values.  

Once that is all complete we have finished pre-processing everything at the FHIR service layer and the system will now call the business layer logic into the OpenEMR service classes.  In this case it will call PatientService->search().
 
PatientService->search() takes care of any UUID conversion (I'm thinking of a way to handle all uuid field conversions automatically but its not done yet) and then calls the BaseService->search method.

BaseService->search uses a very primitive QueryBuilder system to generate the search array.  It will grab all of the select fields in the service class which are usually the same as the table fields unless there are external tables being joined in (You can look at PractitionerService as an example of this).  

Once the initial select clause is generated, the system attempts to convert each of the search fields into a SearchStatement that represents a SQL where sub clause that has bound data values connected to each.  This is handled by the SearchFieldStatementResolver class (I don't like this name too much, I've been trying to think of a different one).  This class converts ISearchFields into SQL query statements.  If there is a compound search clause it will handle this recursively down to 10 nested hierarchical levels.  This resolver handles all of the search modifiers, and comparators (less than, equal to, greater than, etc).  You can see some great examples of how searching with these modifiers/comparators is done in the FhirPatientServiceQueryTest->testGetAll()+FhirPatientServiceQueryTest->searchParameterDataProvider().  Date searching allows for fuzzy matching from left to right specificity, string matching provides for prefix, contains, and exact matching.  Other systems will come on line eventually.

Once the SearchFieldStatementResolver class has finished everything we join it all together in a SQL AND (set intersection) or SQL OR (set union) clause. We send all of this into the BaseService->selectHelper class which is the final chance for a sub class method to override, change, or add additional query elements to the SQL query before it is executed.  I'm also debating on adding some event dispatcher logic here so people can extend things but until things are farther along and stable I'm going to avoid doing that right now.  You can see an example of how selectHelper is extended in the PractitionerService class right now.  Since I don't have the **:missing** modifier implemented I use the selectHelper class to add some additional processing to the search to get back the practitioner data.

Anyways I hope that provides a good overview of the code and helps to follow more easily what it is that I've built so y'all can review and tell me what I should improve on / change.


adunsulag:
@bradymiller So when did we add this codeclimate thing? I find a couple of its suggestions annoying, but I can fix them if that's what the project now requires. 

I can also see that I've got a couple issues to fix with php7.3 preg_match.

bradymiller:
Ignore codeclimate completely. we never requested that and gotta look into
how to get that eyesore removed.

stephenwaite:
think it's under the organization's settings, like somehow the codeclimate github app was enabled there

bradymiller:
the codeclimate webhook has a "cognitive complexity" of 56, so the OpenEMR project considered "refactoring it", and then decided to officially decimated it :)
(ie. the webhook has been removed)

tests/Tests/Services/FHIR/FhirPatientServiceQueryTest.php
```diff
@@ -57,15 +62,81 @@ private function assertGetAllSearchResults($processingResult, $isExpectedToHaveA
...
+
+            // now we will do less than or equal to, only 1 patient in data set has DOB of 1933-03-22
+            ['birthdate', 'le1933'], // search by year
+            ['birthdate', 'le1933-03'], // search by year, month
+            ['birthdate', 'le1933-03-22'], // search by year, month, day
+
+            // now we will do greater than or equal to, only 1 patient in data set has DOB of 1977-05-02
+            ['birthdate', 'ge1977'], // search by year
+            ['birthdate', 'ge1977-05'], // search by year, month
+            ['birthdate', 'ge1977-05-02'], // search by year, month, day
```
bradymiller, line 134: tenor99

bradymiller:
just read through this and code looks great!

2DV (1)

adunsulag:
@bradymiller which docker image should I use to test against the php7.3 environment?  Do we have a docker release that just does the 7.3 but points against the flex environment?  Just eyeballing the code I can't see what the problem would be with why its failing php7.3.  I've been trying to get the insane docker environment up but I'm having issues with it right now so I'm wondering if there is an easier way to test against php7.3

bradymiller:
hi @adunsulag , although looks like I am late (looks like you got the testing sorted out), check this out for getting 7.3 going on the easy dev docker: https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md#other_php_versions

adunsulag:
@bradymiller I'll have to keep investigating this.  Its just bizarre how one of the unit tests keep failing for random php versions.  I'm thinking there's a database timing issue going on thats creating the problem.  Since I'm not ready for this PR to go in just yet I'll just move forward with the AllergyIntollerance entity and see where we get.

bradymiller:
hi @adunsulag , Looks like getting close :) Just let me know when you want me to review the code.

tenor99

src/Services/AllergyIntoleranceService.php
```diff
@@ -95,64 +187,21 @@ public function getAll($search = array(), $isAndCondition = true, $puuidBind = n
...
-                $row['practitioner'];
-            $row['organization'] = $row['organization'] ?
-                UuidRegistry::uuidToString($row['organization']) :
-            $row['organization'];
-            if ($row['diagnosis'] != "") {
-                $row['diagnosis'] = $this->addCoding($row['diagnosis']);
-            }
-            $processingResult->addData($row);
+        // override our type and our puuid
+        if (isset($puuidBind) && !isset($search['puuid'])) {
```
bradymiller, line 200: this if is throwing me off a bit. doesn't this mean that the bootstrapped patient can be overrided to another patient's data if they put another patient's uuid in the lists.pid search field?

bradymiller:
hi @adunsulag , Only had one question on the review (see above). Github has bugged out on the github actions (when i tried to re-run them manually, looks like they went to an indefinite queue :) ; i am likely going to make another PR just to see what is going on with the automated testing (betting it has to do with a couple "bad' fixtures, since i recall something like this happening in the past).

bradymiller:
@adunsulag ,
Regarding the automated testing, can speed things up a bit by setting this setting to 5:
https://www.open-emr.org/wiki/index.php/Administration_Globals#Authentication_Bcrypt_Hash_Cost

sjpadgett:
@bradymiller Quick question. For rxnorm should we use the SAB(Source Vocabularies) GS (Gold Standard Drug Database) or RXNORM (RxNorm terminology normalized names and codes)

bradymiller:
@adunsulag ,

Figured out how to unmask the automated testing error, which then always happens. Recommend bringing in this fix that removes the issues fixtures after they are used (otherwise they were not being removed):
https://github.com/openemr/openemr/pull/4397/commits/b63f694af4c3937f812c546b317175de751be7ea

With this fix, automated testing then always fails on these tests:
https://github.com/openemr/openemr/blob/f2722b5a8de760c9770639c5e9e4b9f9e7062776/tests/Tests/Services/FHIR/FhirAllergyIntoleranceServiceQueryTest.php#L98-L100

Guessing that this testing is pointing out that this function is not working for when request more than 1 unique patient uuid (guessing that when the fixtures were not being removed that sometimes both of those uuid were the same uuid which is why it was sometimes passing testing):
https://github.com/openemr/openemr/blob/f2722b5a8de760c9770639c5e9e4b9f9e7062776/tests/Tests/Services/FHIR/FhirAllergyIntoleranceServiceQueryTest.php#L129

bradymiller:
hi @sjpadgett ,
Free options would be the rxnorm api (https://mor.nlm.nih.gov/download/rxnav/RxNormAPIs.html#) or rxnorm files (https://www.nlm.nih.gov/research/umls/rxnorm/docs/rxnormfiles.html) . 
Looks like GS cost money and wasn't able to find SAB.

sjpadgett:
The RxNorm free set comes with all the SAB within the table.
Example from table rxnconso for the GS SAB for gabapentin:
image

 And the RXNORM SAB:

image

Appears the Gold standard leaves on Manufacture or named brands. GS seem cleaner for our purpose

bradymiller:
@adunsulag ,
Played around with the automated testing a bit. Made two changes to get the testing to work (note they more point out problem rather than provide solid fixes):
https://github.com/openemr/openemr/pull/4397/commits/ae9842b8427a7e999aa12dd7611ac523a918e0be
https://github.com/openemr/openemr/pull/4397/commits/1740d2ffca5dfa8ed75c952feb4dca64924b06f5

adunsulag:
@bradymiller Thanks for the help on the unit tests. It definitely revealed that we weren't handling multipatient filtering.  Its really odd that I ran the tests at least a dozen times and it worked every time on my machine.  Strange as it means the pid ids were the same as the database identifiers as I could see that the data was cleaned up each time.  That's why I wasn't catching it.  Grateful for your help in diagnosing this issue!

stephenwaite:
southpark_cheering

src/Services/AllergyIntoleranceService.php
adunsulag: Good catch.  I've gone in and fixed that.  When I wrote the code I figured the puuidBind was already being set higher up in the FHIRAllergyIntolerance controller and I didn't want to overwrite it.  That's not a good thing here in this case.  That line should be fixed in the latest push I just sent up.

bradymiller:
@adunsulag , looks good. bringing this in so we can keep the good fhir times rolling :)

2DV (1)

PR #4353, opened 2021-04-24
billing contrast css

bradymiller:
@juggernautsei ,
Here's another proposed solution for https://github.com/openemr/openemr/pull/4348
Test this out and let me know how this works for you. This should deal with contrast issues on the billing listing while also working in all the other themes (ie. dark, colors, etc).
(there's still likely lots of css issues looming in this script, but this should at least deal with contrast issue and the weird blocky hovering stuff that was happening)

### DIFF
--- a/interface/billing/billing_report.php
+++ b/interface/billing/billing_report.php
@@ -879,7 +879,7 @@ $partners = $x->_utility_array($x->x12_partner_factory());
             }
             ?>
             <div class="table-responsive">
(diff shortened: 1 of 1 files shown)


bradymiller:
image

juggernautsei:
This looks great! Best job ever.
Looks the same on our system as well.

stephenwaite:
excited_elephant

PR #4355, opened 2021-04-24
restoresession cleanup in patient tracker

bradymiller:
(this will go in next 6.0.0 patch):
restoresession cleanup in patient tracker
@juggernautsei , Wasn't able to pinpoint why that top.sessionrestore was needed in the patient tracker yet. But I did look through that script(s) and found several place where it was missing that could potentially cause issues.

~~Commit 2:~~
~~trial of removal of redundant restoreSession.php include from popup scripts~~

### DIFF
--- a/interface/main/messages/js/reminder_appts.js
+++ b/interface/main/messages/js/reminder_appts.js
@@ -19,6 +19,7 @@ var show_just;
  * which then populates the form with the select patient data
  */
 function recall_name_click(field) {
+    top.restoreSession();
     dlgopen('../../main/calendar/find_patient_popup.php?pflag=0', '_blank', 500, 400);
 }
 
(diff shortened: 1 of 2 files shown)


bradymiller:
@sunsetsystems ,
Just added a commit in this PR to remove all the restoreSession.php includes from the popups since didn't see why we were doing that (seems like would be better to fail, ie. no existing function, if the main top.restoreSession function can not be found). This is more of a trial and was going to do lots of testing before bringing this in. But wanted to make sure there is something i am not missing on why we are doing this.

sunsetsystems:
@bradymiller I think that's fine, but if there are any dialogs left in separate windows (as opposed to the main frameset) then they will need to call restoreSession() like this:

    if (top.restoreSession) {
        top.restoreSession();
    } else {
        opener.top.restoreSession();
    }

Also code like this in dialog.js bugs me:

    if (typeof top.restoreSession === 'function') {
        top.restoreSession();
    }

What if top.restoreSession() is not defined? Surely that's a serious error?

bradymiller:
@sunsetsystems ,
thanks for pointing out that dialog.js issue, which am pretty sure is there to work with patient portal. This makes me want to scratch what i am doing completely and instead start to think about how to conquer this (then the restoreSession() stuff would all go away):
https://github.com/openemr/openemr/issues/2556

sjpadgett:
> if (top.restoreSession) {
>     top.restoreSession();
> } else {
>     opener.top.restoreSession();
> }

Most likely javascript will throw an error with` if (top.restoreSession) ` if not defined. 
dialog is rarely not instantiated in top. Only portal on occasions has an issue.

sunsetsystems:
I'm pretty sure `if (top.restoreSession)` will not throw a reference error. It won't tell you that it's a function though.

sjpadgett:
okay

bradymiller:
testing well and bringing in

PR #4357, opened 2021-04-25
Fixes for LBF rendering.

sunsetsystems:
Using container-fluid and form-inline classes to take advantage of much-needed screen width.
Optional removal of bootstrapping of the provider list so it does not force a new line.
Fixed failure to use group-specific layout column counts.

### DIFF
--- a/interface/forms/LBF/new.php
+++ b/interface/forms/LBF/new.php
@@ -385,7 +385,7 @@ if (
         }
 
         .linkcolor {
-            color:blue;
+            color: blue;
         }
 
     </style>
@@ -614,6 +614,7 @@ if (
         // Validation logic for form submission.
         // Added prevent restoreSession for remotes
         var submitButtonName = '';
+
         function validate(f, restore = true) {
             var errMsgs = new Array();
             <?php generate_layout_validation($formname); ?>
@@ -828,11 +829,9 @@ if (
         function code_attributes_handler(codetype, code, desc, price, warehouses) {
             if (codetype == 'PROD') {
                 fs_append_product(codetype, code, desc, price, warehouses);
-            }
-            else if (codetype == 'ICD9' || codetype == 'ICD10') {
+            } else if (codetype == 'ICD9' || codetype == 'ICD10') {
                 fs_append_diag(codetype, code, desc);
-            }
-            else {
+            } else {
                 fs_append_service(codetype, code, desc, price);
             }
         }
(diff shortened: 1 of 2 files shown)


interface/forms/LBF/new.php
```diff
@@ -865,16 +859,18 @@ function warehouse_changed(sel) {
 <body class="body_top"<?php if ($from_issue_form) {
     echo " style='background-color:var(--white)'";
                       } ?>>
-<div class="<?php echo $form_container ?>">
+<div>
```
sjpadgett, line 862: Effectively breaks bootstrap. A container is required even if we use a form as container.
This may work on some LBF designs but will break others.
Also form-inline will give unpredictable results across all LBF's.
Moreso, most folks are abandoning desktops in favor of tablets. Easier to manage for provider moving between exam rooms.
I reality, all LBF's should be container-fluid and i'm unsure why container was used as default.
Help me understand why this change is necessary and perhaps I can help find a solution w/o breaking bootstrap.

First thing to break is the group expand buttons due to form-inline.  Flex-row helps with responsiveness but along with col-12, may not be required. We could restyle to only use flexbox with LBF's and lose the BS grid system where only grids used are the LBF row/col grid.

sunsetsystems: A container inside another container wasn't working for me, and the prior code was doing that. What I did was move the container (now container-fluid) to below where the LBF layout starts. What's not bootstrapped here are just the title and provider selector, but those could be as well.

The main reason for form-inline is so multiple input fields in the same column will be side by side and not stacked, except on very small devices. This is how they worked before bootstrapping. For my clients it's important to take advantage of screen width and not have to do unnecessary vertical scrolling. Tablets are OK if they are not too small.

The group expand checkboxes are working fine here. Did you test and see otherwise?

That said, I'm not very experienced with BS and your input is welcomed. I would like to keep it specific though, and if you have layouts that you think will break then let's take a look.

sjpadgett: Yep, I pulled down entire script and see what you did.(I'm horrible at reviewing. Better at big picture!).
It's seems okay except i'm concerned the form is not inside a container(may cause unexpected scroll bars). Perhaps moving the container-fluid block from under form to under body makes more sense. I just haven't seen using a container under a form class as necessary.
However too me, BS4 is just weird and perhaps the reason BS5 followed so quickly.  I now completely stay away from styling with BS4 grid system because it is so cantankerous making me unsure if the form class is considered a container by BS4.
In the end, I planned to refactor out any BS grid in LBF and only use flex and use the appropriate form element classes.

Up to you as I have no objections save for those concerns I mentioned. I'll do some testing on some of my more complicated LBF's but portal seems okay so far. I'll also test with the col grids removed. I think those are what causes most of our styling grief.
I'm just happy to see container replaced with container-fluid for encounter forms.:)

sunsetsystems: Thanks Jerry, I'll play with it a bit more.

sunsetsystems: OK I made a few changes:
1. Moved container to encompass the entire form.
2. Changed flex-row to row, though it doesn't seem to matter.
3. Added row/column divs as needed to avoid inline arrangement of divs.
4. Reverted changes to genProviderSelect. I don't understand why it has a span with form-inline, it was actually forcing the control to be on a separate line... @sjpadgett you may want to look at that... not sure if form-inline is supposed to work outside of the <form> tag.
5. Changed provider selector.
6. Removed use of table-responsive class which was causing a scrollbar regardless of widths.

interface/forms/LBF/new.php
```diff
@@ -1704,7 +1709,11 @@ function warehouse_changed(sel) {
             ?>
             <br />
 
-            <div class="col-12">
+        </div><!-- end column -->
+    </div><!-- end row -->
+    <div class="row"><!-- was flex-row -->
+        <div class='col-12'>
```
sjpadgett, line 1715: div block containing form tag is closed before form contents are complete through L-1805.
Pretty sure this will effect what's included on submit.

I'd replace where the form tag is echoed to a html tag with arg replacements being echoed. Will allow IDE to help with tag closures.
Here's a screen showing html error.
image

interface/forms/LBF/new.php
sjpadgett: > 4\. not sure if form-inline is supposed to work outside of the form tag.

Though it is just a css class rule and may or may not work, it is out of place. I'd remove because it doesn't make sense anyway in a span.

interface/forms/LBF/new.php
sunsetsystems: Not sure I understand about the HTML tag, but I moved the form open and close tags to be at the outer level, containing all the divs. Also removed form-inline from genProviderSelect() as you suggested. Testing OK here. Thanks for the insight and testing, I really appreciate it!

sjpadgett: @sunsetsystems 
Do we really want the dark background on all LBF's?
image

It's always bothered me. I can change the theme builds to take the background out so LBF follows theme color.
image

I have pending PR I can add to....

stephenwaite: i like the contrast, :2nd_place_medal:

sjpadgett: It looks much better in the other colored themes so, maybe i'll pick on something else.:)

sunsetsystems:
@sjpadgett thanks for the contribution! It's testing well here and I'll merge this all in a bit if there are no further objections.

sjpadgett:
I'm good Rod. Thanks.

PR #4361, opened 2021-04-27
Integrate traditional and layout-based forms into their chosen categories in the visit summary dropdowns.

sunsetsystems:
The visit summary page has dropdowns to select new forms. Currently there is a dropdown for each of the traditional form categories, and a separate one for all LBF forms. However users don't care which technology is behind each form.

This PR consolidates LBF entries into the form categories, and sorts the categories and forms. Note that categories and sorting sequence can be assigned in the LBF properties dialog.


bradymiller:
2DV (1)

sunsetsystems:
Alright I'm gonna take that as a review. :) Thanks guys.

PR #4362, opened 2021-04-27
Implemented CSV loader for language translations.

sunsetsystems:
This is based on old work from Kevin Yeh for IPPF. I adapted it as an enhancement to the existing Multi Language Tool.

The idea is that you can easily upload and process a CSV spreadsheet that has language constants in one column, and the desired translations in some other column.

### DIFF
--- a/interface/language/csv/commit_csv.php
+++ b/interface/language/csv/commit_csv.php
@@ -12,6 +12,7 @@
 
 // This script is ajax-loaded into a div by validate_csv.php, so much of
 // the normal framework does not apply.
+header('Content-Type: application/json');
 
 require_once("../../globals.php");
 require_once("translation_utilities.php");
@@ -54,14 +55,14 @@ if (!$errmsg) {
             false,
             $preview
         );
-        if (strpos($result, xl('Definition Exists') . ':') !== 0) {
-            if (strpos($result, xl('Empty Definition')) !== 0) {
+        if (strpos($result, '[2]') !== 0) { // Definition Exists
+            if (strpos($result, '[1]') !== 0) { // Empty Definition
                 if ($result) {
                     array_push($changed, $result);
-                    if (strpos($result, xl('Update From') . ':') === 0) {
-                        array_push($updated, $result);
-                    } else if (strpos($result, xl('Create') . ':') === 0) {
-                        array_push($created, $result);
+                    if (strpos($result, '[3]') === 0) {
+                        array_push($updated, substr($result, 3));
+                    } else if (strpos($result, '[5]') === 0) {
+                        array_push($created, substr($result, 3));
                     }
                 }
             } else {
--- a/interface/language/csv/commit_csv.php
+++ b/interface/language/csv/commit_csv.php
@@ -1,0 +1,91 @@
+<?php
+
+/*
+ * @package OpenEMR
+ * @link    http://www.open-emr.org
+ * @author  Kevin Yeh <kevin.y@integralemr.com>
+ * @author  Rod Roark <rod@sunsetsystems.com>
+ * @copyright Copyright (c) 2014 Kevin Yeh <kevin.y@integralemr.com>
+ * @copyright Copyright (c) 2021 Rod Roark <rod@sunsetsystems.com>
+ * @license https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+// This script is ajax-loaded into a div by validate_csv.php, so much of
+// the normal framework does not apply.
+
+require_once("../../globals.php");
+require_once("translation_utilities.php");
+
+$errmsg = '';
+
+if (!$errmsg && !isset($_REQUEST['translations'])) {
+    $errmsg = xlt("No translations!");
+}
+
+if (!$errmsg && !isset($_REQUEST['lang_id'])) {
+    $errmsg = xlt("No Language ID specified!");
+}
+
+if (!isset($_REQUEST['preview'])) {
+    $preview = true;
+} else {
+    $preview = $_REQUEST['preview'];
+    if ($preview === "false") {
+        $preview = false;
+    }
+}
+
+$unchanged = 0;
+$empty = 0;
+$changed = array();
+$created = array();
+$updated = array();
+
+if (!$errmsg) {
+    $lang_id = $_REQUEST['lang_id'];
+    $translations = json_decode($_REQUEST['translations']);
+    foreach ($translations as $translation) {
+        $result = verify_translation(
+            str_replace("\r\n", "\n", $translation[0]),
+            str_replace("\r\n", "\n", $translation[1]),
+            $lang_id,
+            true,
+            "",
+            false,
+            $preview
+        );
+        if (strpos($result, xl('Definition Exists') . ':') !== 0) {
+            if (strpos($result, xl('Empty Definition')) !== 0) {
+                if ($result) {
+                    array_push($changed, $result);
+                    if (strpos($result, xl('Update From') . ':') === 0) {
+                        array_push($updated, $result);
+                    } else if (strpos($result, xl('Create') . ':') === 0) {
+                        array_push($created, $result);
+                    }
+                }
+            } else {
+                $empty++;
+            }
+        } else {
+            $unchanged++;
+        }
+    }
+}
+
+if ($errmsg) {
+    $created[] = xl('ERROR') . ': ' . $errmsg;
+}
+
+$retval = array();
+$retval['changed'] = $changed;
+$retval['unchanged'] = $unchanged;
+$retval['empty'] = $empty;
+$retval['updated'] = $updated;
+$retval['created'] = $created;
+$changes_html = "";
+foreach ($changed as $change) {
+    $changes_html .= $change . "<br>";
+}
+$retval['html_changes'] = $changes_html;
+echo json_encode($retval);
--- a/interface/language/csv/commit_csv.php
+++ b/interface/language/csv/commit_csv.php
@@ -52,7 +52,6 @@ if (!$errmsg) {
             $lang_id,
             true,
             "",
-            false,
             $preview
         );
         if (strpos($result, '[2]') !== 0) { // Definition Exists
--- a/interface/language/csv/translation_utilities.php
+++ b/interface/language/csv/translation_utilities.php
@@ -1,0 +1,165 @@
+<?php
+
+/* Copyright (C) 2014 Kevin Yeh <kevin.y@integralemr.com>
+ *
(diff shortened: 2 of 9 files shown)


interface/language/csv/commit_csv.php
```diff
@@ -0,0 +1,91 @@
...
+ * @link    http://www.open-emr.org
+ * @author  Kevin Yeh <kevin.y@integralemr.com>
+ * @author  Rod Roark <rod@sunsetsystems.com>
+ * @copyright Copyright (c) 2014 Kevin Yeh <kevin.y@integralemr.com>
+ * @copyright Copyright (c) 2021 Rod Roark <rod@sunsetsystems.com>
+ * @license https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+// This script is ajax-loaded into a div by validate_csv.php, so much of
+// the normal framework does not apply.
```
bradymiller, line 14: does it make sense for scripts that only return json to do this:
```header('Content-Type: application/json');```
(figured this may decrease risk of cross scripting attacks)

**Comment by bradymiller** [comment]
hi @sunsetsystems , do you have any quick csv file that we can use for testing?

File: interface/language/csv/commit_csv.php [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+            false,
+            $preview
+        );
+        if (strpos($result, xl('Definition Exists') . ':') !== 0) {
+            if (strpos($result, xl('Empty Definition')) !== 0) {
+                if ($result) {
+                    array_push($changed, $result);
+                    if (strpos($result, xl('Update From') . ':') === 0) {
+                        array_push($updated, $result);
+                    } else if (strpos($result, xl('Create') . ':') === 0) {
```
bradymiller, line 63: it's odd that there are translated things going on here in comparisons (generally best to avoid things that will break with the 'dummy' language (this is a OpenEMR testing language where all constants are changed to dummy)).

interface/language/csv/commit_csv.php
sunsetsystems: Hmm, I'll try and see if it works.

sunsetsystems:
> hi @sunsetsystems , do you have any quick csv file that we can use for testing?

Try this.

[multilanguage_test.csv](https://github.com/openemr/openemr/files/6389278/multilanguage_test.csv)

interface/language/csv/commit_csv.php
sunsetsystems: Yeah they are translated because they are also displayed. Kind of an odd design and I was trying not to spend a lot of time on this. I'll look it over again.

sunsetsystems:
OK I removed the dependency on translations and added the content-type header. Still testing well here.

interface/language/csv/translation_utilities.php
```diff
@@ -0,0 +1,165 @@
...
+                $sqlDelete = " DELETE FROM lang_constants where cons_id = ? ";
+                sqlStatement($sqlDelete, array($row['cons_id']));
+                $sqlDelete = " DELETE FROM lang_definitions where cons_id = ? ";
+                sqlStatement($sqlDelete, array($row['cons_id']));
+                error_log("DELETED Definitions for duplicate constant:" . $constant . "|" . $row['cons_id']);
+            }
+            return $retval;
+        }
+        if ($row_count == 0) {
+            $sqlInsert = " INSERT INTO lang_constants (constant_name) VALUES (?)";
```
bradymiller, line 45: when do this should also add entry to lang_custom to ensure the csv loader and Manage Translations features are compatible. Can do this by running it through the insert_language_log() function.

interface/language/csv/translation_utilities.php
```diff
@@ -0,0 +1,165 @@
...
+            $new_index = sqlInsert($sqlInsert, array($constant));
+            return $new_index;
+        }
+    }
+}
+
+function update_metainfo($constant, $definition, $source, $set_source = false)
+{
+    if ($source != "") {
+        $sqlUpdate = " UPDATE ippf_lang_definitions set " . $source . "= ? where BINARY constant_name = ?";
```
bradymiller, line 55: looks like a new table here

interface/language/csv/translation_utilities.php
```diff
@@ -0,0 +1,165 @@
...
+        if ($set_source) {
+            $sqlUpdate = " UPDATE ippf_lang_definitions set source = ? where BINARY constant_name = ?";
+            $parameters = array($source, $constant);
+            sqlStatement($sqlUpdate, $parameters);
+        }
+    }
+}
+
+function verify_translation($constant, $definition, $language, $replace = true, $source = "", $set_metainfo = false, $preview = false)
+{
```
bradymiller, line 67: when add/update stuff also need to get this into the lang_custom table via the insert_language_log() function. I wouldn't worry about the $row count > 1 scenario since this should never happen and would get sorted out if the user ever ran through the Manage Translations process

interface/language/csv/validate_csv.php
```diff
@@ -0,0 +1,284 @@
...
+
+<div id="status"></div>
+<div id="information"></div>
+<div id="file-display" data-bind="template:{name: 'file-info', data: filedata}" ></div>
+
+<script type="text/html" id="file-info">
+
+    <div id="verify" data-bind="visible: mode()=='verify' ">
+        <h1>
+            <?php echo xlt('Verify Contents to apply to') . ' ' . $lang_description; ?>
```
bradymiller, line 67: text() around $lang_description

interface/language/csv/validate_csv.php
```diff
@@ -0,0 +1,284 @@
...
+                        <th data-bind="text:$data.text, attr: { index: $index}"></th>                
+                </tr>
+            </thead>
+            <tbody>
+                <?php
+                for ($idx = 1; $idx < count($file_contents); $idx++) {
+                    $row = $file_contents[$idx];
+                    echo "<tr>";
+                    foreach ($row as $cell) {
+                        echo "<td>" . $cell . "</td>";
```
bradymiller, line 87: text() around $cell

interface/language/csv/validate_csv.php
```diff
@@ -0,0 +1,284 @@
...
+            <h3><?php echo xlt('Updated Definitions List'); ?></h3>
+            <div data-bind="foreach: updated">
+                <div data-bind="text:$data"></div>
+            </div>
+        </div>
+    </div>
+    <!-- /ko -->
+
+    <!-- ko if:loading() -->
+        <span data-bind="text:processingStatus"></span><img src='<?php echo $webroot . "/interface/pic/ajax-loader.gif"; ?>'/>
```
bradymiller, line 146: would be nice if did this with a fontawesome icon rather than the gif

interface/language/csv/translation_utilities.php
sunsetsystems: Hmm, looks like that code is never executed. I'll remove it.

interface/language/csv/translation_utilities.php
sunsetsystems: What if the spreadsheet is distributed as "official" translations and are not considered customizations? That's the normal use case here.

bradymiller:
code review complete (didn't test ). very neat feature to have. Just noted couple issues.

interface/language/csv/commit_csv.php
bradymiller: dang, that was quick

interface/language/csv/translation_utilities.php
bradymiller: is that whole preview thing never used (looks like is uses this when not in preview mode)?

interface/language/csv/translation_utilities.php
bradymiller: I think that will be ok and will then just mean that a user could use IPPF official translations via this csv method, which will then take precedence and carry over even if a user downloads the "official" openemr translations followed by the Manage Translations process (Manage Translations is basically adding back all the translation work that the instance had done locally, which would include the csv method in also).

bradymiller: actually, let me think about this more before doing work on this part

bradymiller: quick question:
Are these IPPF instances doing their own custom local translations on top of the csv set (ie. will they have local things that need to override/survive the csv set)?

sunsetsystems: Yes it's likely that some sites will. Probably need to preserve their custom translations.

interface/language/csv/translation_utilities.php
sunsetsystems: Preview is used. But looks to me like the 6th arg of verify_translation() is always false, and as a result it never calls update_metainfo(). Not to mention the table does not exist and yet nothing crashes.

interface/language/csv/validate_csv.php
sunsetsystems: Do you know of something else doing that? Maybe I can clone some code....

interface/language/csv/translation_utilities.php
sunsetsystems: Maybe a checkbox to specify which way to go?

interface/language/csv/validate_csv.php
bradymiller: try to replace the img tag with this:
```
<i class="fa fa-spinner fa-spin"></i>
```

interface/language/csv/translation_utilities.php
bradymiller: Interesting issue.
1. Current Manage Translations allows local instances keep their local changes after importing an official set. So in IPPF's case, won't want to write to lang_custom when do this (since that instance can do the official IPPF import and then run Manage Translations to bring back their local changes).
2. I'm just not sure what will happen in the wild in the future. I think actually best to not write the csv import to lang_custom and not even offer an option (since it will just be confusing). If really necessary and wanted to have it all they could do (ie. by not adding to lang_custom just adds one more step and keeps it less confusing):
- make their local custom changes
- bring in openemr's official translations
- bring in csv translations
- run Manage Translations (this will bring back their local custom changes)

So, would just keep it how you have it :)
And if address the other minor issues and testing well, then feel free to bring it in.

interface/language/csv/translation_utilities.php
bradymiller: disregard above comment since best for the csv import to not mess with lang_custom

interface/language/csv/translation_utilities.php
bradymiller: probably make sense to just yank all that related unused code

sunsetsystems:
Thanks Brady!

PR #4378, opened 2021-05-05
Fixes for restoreSession logic.

sunsetsystems:
restoreSession() was setting a session cookie with different attributes (most notably path) from the one set originally by PHP. This resulted in duplicated cookies and dismal failure of session swapping. This PR fixes that.

### DIFF
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -160,19 +160,23 @@ if ($GLOBALS['login_page_layout'] == 'left') {
             element.disabled = true;
             // nothing fancy. mainly for mobile.
             element.innerHTML = '<i class="fa fa-sync fa-spin"></i> ' + jsText(<?php echo xlj("Authenticating"); ?>);
-            <?php if (!empty($GLOBALS['restore_sessions'])) { ?>
+            <?php if (session_name()) { ?>
+                <?php $scparams = session_get_cookie_params(); ?>
                 // Delete the session cookie by setting its expiration date in the past.
                 // This forces the server to create a new session ID.
                 var olddate = new Date();
                 olddate.setFullYear(olddate.getFullYear() - 1);
-                <?php if (version_compare(phpversion(), '7.3.0', '>=')) { ?>
-                    // Using the SameSite setting when using php version 7.3.0 or above, which avoids browser warnings when cookie is not 'secure' and SameSite is not set to anything
-                    document.cookie = <?php echo json_encode(urlencode(session_name())); ?> + '=' + <?php echo json_encode(urlencode(session_id())); ?> + '; path=<?php echo($web_root ? $web_root : '/');?>; expires=' + olddate.toGMTString() + '; SameSite=Strict';
-                <?php } else { ?>
-                    document.cookie = <?php echo json_encode(urlencode(session_name())); ?> + '=' + <?php echo json_encode(urlencode(session_id())); ?> + '; path=<?php echo($web_root ? $web_root : '/');?>; expires=' + olddate.toGMTString();
-                <?php } ?>
+                var mycookie = <?php echo json_encode(session_name()); ?> + '=' + <?php echo json_encode(session_id()); ?> +
+                    '; path=' + <?php echo json_encode($scparams['path']); ?> +
+                    '; domain=' + <?php echo json_encode($scparams['domain']); ?> +
+                    '; expires=' + olddate.toGMTString();
+                var samesite = <?php echo json_encode(empty($scparams['samesite']) ? '' : $scparams['samesite']); ?>;
+                if (samesite) {
+                    mycookie += '; SameSite=' + samesite;
+                }
+                document.cookie = mycookie;
             <?php } ?>
-            document.forms[0].submit();
+            return true;
         }
     </script>
 </head>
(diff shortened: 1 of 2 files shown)


bradymiller:
@sunsetsystems 
nice stuff
should we also be doing this in login.php where manually building the cookie (see document.cookie in that script) to ensure they are also not different?

library/restoreSession.php
```diff
@@ -28,7 +29,15 @@
 var oemr_session_name = <?php echo json_encode(urlencode(session_name())); ?>;
 var oemr_session_id   = <?php echo json_encode(urlencode(session_id())); ?>;
 var oemr_dialog_close_msg = <?php echo (function_exists('xlj')) ? xlj("OK to close this other popup window?") : json_encode("OK to close this other popup window?"); ?>;
-//
+
+var oemr_scp_lifetime = <?php echo $scparams['lifetime']; ?>; // must be numeric
```
bradymiller, line 33: note json_encode/js_escape should keep it numeric if needed (pretty sure it does not place quotes around numbers)

sunsetsystems:
> should we also be doing this in login.php where manually building the cookie (see document.cookie in that script) to ensure they are also not different?

Dang, didn't know about that. Yes I think so and will take a look.

sunsetsystems:
OK I think we're good now.

sjpadgett:
nice catch @sunsetsystems. I looked at late v5.0.2 and was only creating the one cookie. If I remember right we didn't see a lot of problems with session on that version!

bradymiller:
2DV (1)

bradymiller:
(btw, will get this into next 6.0.0 patch)

sunsetsystems:
Thanks Brady, yeah definitely want it in 6.0.0!

interface/login/login.php
```diff
@@ -160,19 +160,23 @@ function transmit_form(element) {
...
+                <?php $scparams = session_get_cookie_params(); ?>
                 // Delete the session cookie by setting its expiration date in the past.
                 // This forces the server to create a new session ID.
                 var olddate = new Date();
                 olddate.setFullYear(olddate.getFullYear() - 1);
-                <?php if (version_compare(phpversion(), '7.3.0', '>=')) { ?>
-                    // Using the SameSite setting when using php version 7.3.0 or above, which avoids browser warnings when cookie is not 'secure' and SameSite is not set to anything
-                    document.cookie = <?php echo json_encode(urlencode(session_name())); ?> + '=' + <?php echo json_encode(urlencode(session_id())); ?> + '; path=<?php echo($web_root ? $web_root : '/');?>; expires=' + olddate.toGMTString() + '; SameSite=Strict';
-                <?php } else { ?>
-                    document.cookie = <?php echo json_encode(urlencode(session_name())); ?> + '=' + <?php echo json_encode(urlencode(session_id())); ?> + '; path=<?php echo($web_root ? $web_root : '/');?>; expires=' + olddate.toGMTString();
```
bradymiller, line 172: hi @sunsetsystems , just realized why had the urlencode here, which documented here (as I recall, when a comma is in there, it doesn't work as it should):
https://github.com/openemr/openemr/blob/master/src/Common/Session/SessionUtil.php#L25-L27

just checking to make sure you are ok with me adding that back?

sunsetsystems: I'm not seeing why a comma would mess it up... special characters with meaning in cookies are semicolon, equals sign and space. Also not sure that urlencode is the right method of encoding?

bradymiller: looked back into this. Looks like php does this without telling us:
image
(note comma is set as %2C by session_start())
So, when we manually add a session id with a comma, then need to encode it or it won't match the one set by php

bradymiller: also to add, urlencode just happens to convert that comma to %2C :)

sunsetsystems: Bottom line, I'm confused about this. Can't find an authoritative source that says what encoding, if any, PHP expects with session IDs in cookies.

bradymiller: yep, this was a huge painful bug to sort out (I am beginning to remember the pain) when we went to sid_bits_per_character of 6 to increase the randomness of the session id. The bug is more obvious if your remove the urlencode's from the restoresession script. Then the session separations essentially collapse, but only when there is a comma in the session (happens about 1 in 3 session id's). The documentation is super poor on this and I recall finally finding something that discussed this on the internet and noted that PHP was encoding the comma in the cookie it produced. fun stuff (to convince yourself rec. removing the urlencode's on the restoresession script) :)

sunsetsystems: Alrighty then, please put them back! Thanks!

bradymiller: oh goodness, just tested after removed urlencode's from the restoresession script. Log in and out until you get a cookie with a %2C in it, then watch the magic :)
First the restoresession will convert the cookie id %2C to a actual comma, then it will change it to a new session id on next click, and will then get the infamous "Site ID is missing from session data!" message :)

sunsetsystems: urlencode it is then. :)

bradymiller: done :)

PR #4382, opened 2021-05-07
CCDA and Other

sjpadgett:
- procedure form add delimit for new active flag lab providers
- expand procedure results/pending to fluid container.
- CCM codes support for CCDA

More template updates
Some core minor fixes here and there


sjpadgett:
Updating as I go in case there is anyone brave enough to jump in. It's the darn coding systems that are aggravating.
ONC insist on using SNOMEDs for everything. Don't follow spec for what is required or not. If we wish to do ICD10 or CPT4 we have to arrange in advance with ATL.
Making slow progress but, i'm trying to catch up with 5 years of spec changes in a couple months. Wouldn't be so bad if ONC caught up with real world usage!!!!

bradymiller:
2DV (1)

sjpadgett:
UDI done. Down to 2 minor errors on b1 CCD Cures and some html table fixups:
image

sjpadgett:
@bradymiller @stephenwaite Now you can bring on the full cheering section!:) I'm kind of relieved this is done which is the b1 ONC 2015 C-CDA ToC/CCD
I'll update PR to this point then bring in.
Next I'll fix the warnings which turn to errors on CURES and work on Note document. 

image

bradymiller:
tenor99

2DV (1)

stephenwaite:
squirrel

sjpadgett:
Passing CURES except having to use scenarios document dates of like 2015-06-22 instead of current dates at creation. Really stupid! May have to set computer date(old hacking trick):)

Added Birth Name to demo layout for name at birth requirement.

image

I have to figure out setting up the database and upgrade scripts for the feature yet!

PR #4391, opened 2021-05-12
added a tabRefreshByName() api call in the tabs view model

kchapple:
added a tabRefreshByName() api call in the tabs view model so tabs can be refreshed by api.

I use this in a module where a form is opened in a new tab by clicking on an element in a list. There is a delete button at the bottom of the form. When the entity is deleted by clicking the button on the form, we navigate the user back to the list, but we have to refresh the "list tab" first so the deleted entity is removed from the list.

Ken

### DIFF
--- a/interface/main/tabs/js/tabs_view_model.js
+++ b/interface/main/tabs/js/tabs_view_model.js
@@ -100,6 +100,12 @@ function tabRefresh(data,evt)
     }
 }
 
+/**
+ *  Given a name, refresh that tab. This code is used to support custom code where it is required
+ *  to programmatically refresh a tab via javascript. This func may not be used by core code, please do not remove.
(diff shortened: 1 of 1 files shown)


interface/main/tabs/js/tabs_view_model.js
```diff
@@ -100,6 +100,15 @@ function tabRefresh(data,evt)
     }
 }
 
+function tabRefreshByName(name) {
```
bradymiller, line 103: recommend a comment here that this function is used to support custom/extended code so we don't remove it during code cleanups (where we remove code that appears unused)

bradymiller:
hi @kchapple , Looks good and bringing in. thanks! -brady

PR #4399, opened 2021-05-17
New use RxNORM table for drug lookups.

### LINKED ISSUE #4398: Add ability to search RxNORM in prescriptions
Fairly straight forward. If rxnconso (RxNORM free) external table is installed then user has option to search and select drug from there.
Lookup uses the SAB of RXNORM so licensing is valid for free version.

image

sjpadgett:
- if installed option to use rxnorms
- fudge dialog size and replace Done button with Quit.

Fixes #4398

### DIFF
--- a/controllers/C_Prescription.class.php
+++ b/controllers/C_Prescription.class.php
@@ -44,8 +44,9 @@ class C_Prescription extends Controller
         $this->RxList = new RxList();
         // test if rxnorm available for lookups.
         $rxn = sqlQuery("SELECT table_name FROM information_schema.tables WHERE table_name = 'RXNCONSO' OR table_name = 'rxconso'");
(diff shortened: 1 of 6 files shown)


sjpadgett:
I think I should go a head and include RxCUI native load lookup if it is thought to be useful.
I'm not sure what users prefer to load.
@bradymiller @sunsetsystems @stephenwaite

stephenwaite:
only thing easier than the nlm lookup is the monthly prescribe imho

sjpadgett:
what's the best way to check if native rxcui is installed?

stephenwaite:
check if anything in `RXCONSO`?

sjpadgett:
native loaded rxcui is loaded in codes table.

sjpadgett:
Okay so we can't say I took easy way here I added native loaded RxCUI for lookup. 
- if a code set is not enabled then radio will disable for that code set.

Will merge shortly.

image

bradymiller:
2DV (1)

PR #4400, opened 2021-05-18
Quick fix to add lookups.

sjpadgett:
- prevent hidden rxcode overwrite on already selected drug.


sjpadgett:
I've tested this feature on rel v6.0.0 so I may bring back for patch 2.

PR #4411, opened 2021-05-22
Can't save new list in list_edit

sjpadgett:
-fix order of operations on save conditional null coalescing


sunsetsystems:
Good catch, thanks!

sjpadgett:
no problem. thanks for fixing the edit options on core items  in layouts. that'll save some head scratching for users.:)

bringing in as works okay for me...

bradymiller:
2DV (1)

PR #4413, opened 2021-05-24
Eye Form Patch 2

ophthal:
Some tweaks needed for the Eye Form.

The addition in save.php is critical: default value for “billed” was null and it needs to be =“0” explicitly. The Billing Manager requires this.

Other changes are cosmetic.

Fixes #


bradymiller:
hi @ophthal , bringing this in and will also get in next 6.0.0 patch. thanks for the fixes! -brady

PR #4420, opened 2021-05-26
migrate twig to separate class

bradymiller:
migrate twig to separate class


stephenwaite (approved):
looks great, and moving out of core to common will load autoload this?

bradymiller:
Just seemed a better fit in common. Can see it in action here (basically just add the `use` and then the one liner to use it):
https://github.com/bradymiller/openemr/commit/a92217438c9049bb447e92c791efec8bb66cd01b

Issue #4205, opened 2021-01-30
Missing FHIR Resources


sjpadgett:
### Describe the problem
We still have missing resources to complete our core FHIR API. Any help is appreciated and are needed very soon.
- CarePlan https://www.hl7.org/fhir/careplan.html
- Device https://www.hl7.org/fhir/device.html
- DiagnosticReport https://www.hl7.org/fhir/DiagnosticReport.html
- DocumentReference https://www.hl7.org/fhir/DocumentReference.html
- Goal https://www.hl7.org/fhir/Goal.html

@vishnuyar Are any of these in your plans?

vishnuyar:
@sjpadgett Almost done with adding Surgery to Procedures FHIR Resource. 
Started working on Medication.
Can take up these resources next. If there is any importance of order, I can do them accordingly.

sjpadgett:
@vishnuyar thanks
I'd say DiagnosticReport then DocumentReference. Device will need to be last I suppose.

rhausam:
Wondering if there is a status update on where we are with these resources, @vishnuyar?  I'm familiar with them and responsible as an HL7 Orders and Observations Work Group Co-chair for DiagnosticReport and Device.  I'm interested in re-engaging and may be able to help with some of them if that's still needed.

vishnuyar:
@rhausam , I got pulled somewhere else. Haven't made much progress. Currenly working on DiagnosticReport. You can take another FHIR Resource.

PR #4422, opened 2021-05-26
Openemr fhir device, careplan, goal

adunsulag:
This PR implements and passes ONC Inferno requirements for Device, Careplan and Goal (See #4205). I also improved Test Fixtures by allowing us to grab foreign key data and generate uuids and database ids as part of the json definition.

Created a CarePlan service that handles the care plan form where goals and care plans are stored.  Since care plans do not have a unique id we had to implement a surrogate key using the encounter and the form_id of the care plan form.

Implemented the CodeTypesService that wraps around some of the global lookup code functions and provides additional helper method functions for grabbing the OID or system URI for a specific code.

Fixed some style issues in various classes.

Updated PatientService's getOne function to remove duplicate code.

Added some protection on npi number in the PractitionerService class to deal with the 'missing' modifier.

Heavily modified the test fixtures to be able to pull in referenced column data points for the fixture.  Wrote a small DSL to allow whitelisted function calls inside the JSON to do things like generate ids, and uuids.  Wrote a fixture manager for Encounters and care plan.

### DIFF
--- a/library/formdata.inc.php
+++ b/library/formdata.inc.php
@@ -92,9 +92,10 @@ function process_cols_escape($s)
  * @param   string|array        $s       sql column name(s) variable to be escaped/sanitized.
  * @param   array         $tables  The table(s) that the sql columns is from (in an array).
  * @param   boolean       $long    Use long form (ie. table.colname) vs short form (ie. colname).
+ * @param   boolean       $throwException Whether to throw a SQL exception instead of dieing
  * @return  string                 Escaped table name variable.
  */
-function escape_sql_column_name($s, $tables, $long = false)
+function escape_sql_column_name($s, $tables, $long = false, $throwException = false)
 {
     // If $s is asterisk return asterisk to select all columns
     if ($s === "*") {
@@ -140,7 +141,8 @@ function escape_sql_column_name($s, $tables, $long = false)
     }
 
     // Now can escape(via whitelisting) the sql column name
-    return escape_identifier($s, $columns_options, true);
+    $dieIfNoMatch = !$throwException;
+    return escape_identifier($s, $columns_options, $dieIfNoMatch, true, $throwException);
 }
 
 /**
@@ -214,9 +216,10 @@ function mitigateSqlTableUpperCase($s)
  *                                          characters (for example 'a-zA-Z0-9_').
  * @param   boolean      $die_if_no_match  If there is no match in the whitelist, then die and echo an error to screen and log.
  * @param   boolean      $case_sens_match  Use case sensitive match (this is default).
+ * @param   boolean      $throw_exception_if_no_match If there is no match in the whitelist then throw an exception
  * @return  string                         Escaped/sanitized sql identifier variable.
  */
-function escape_identifier($s, $whitelist_items, $die_if_no_match = false, $case_sens_match = true)
+function escape_identifier($s, $whitelist_items, $die_if_no_match = false, $case_sens_match = true, $throw_exception_if_no_match = false)
 {
     if (is_array($whitelist_items)) {
         // Only return an item within the whitelist_items
@@ -237,6 +240,8 @@ function escape_identifier($s, $whitelist_items, $die_if_no_match = false, $case
                     // No match and $die_if_no_match is set, so die() and send error messages to screen and log
                     error_Log("ERROR: OpenEMR SQL Escaping ERROR of the following string: " . errorLogEscape($s), 0);
                     die("<br /><span style='color:red;font-weight:bold;'>" . xlt("There was an OpenEMR SQL Escaping ERROR of the following string") . " " . text($s) . "</span><br />");
+                } else if ($throw_exception_if_no_match) {
+                    throw new \OpenEMR\Common\Database\SqlQueryException("", "ERROR: OpenEMR SQL Escaping ERROR of the following string: " . errorLogEscape($s));
                 } else {
                     // Return first token since no match
                     $key = 0;
@@ -251,6 +256,8 @@ function escape_identifier($s, $whitelist_items, $die_if_no_match = false, $case
                 // Contains illegal character and $die_if_no_match is set, so die() and send error messages to screen and log
                 error_Log("ERROR: OpenEMR SQL Escaping ERROR of the following string: " . errorLogEscape($s), 0);
                 die("<br /><span style='color:red;font-weight:bold;'>" . xlt("There was an OpenEMR SQL Escaping ERROR of the following string") . " " . text($s) . "</span><br />");
+            } else if ($throw_exception_if_no_match) {
+                throw new \OpenEMR\Common\Database\SqlQueryException("", "ERROR: OpenEMR SQL Escaping ERROR of the following string: " . errorLogEscape($s));
             } else {
                 // Contains all legal characters, so return the legal string
                 return $s;
--- a/src/Services/CarePlanService.php
+++ b/src/Services/CarePlanService.php
@@ -24,6 +24,7 @@ use OpenEMR\Services\Search\SearchModifier;
 use OpenEMR\Services\Search\StringSearchField;
 use OpenEMR\Services\Search\TokenSearchField;
 use OpenEMR\Services\Search\TokenSearchValue;
+use OpenEMR\Validators\BaseValidator;
 use OpenEMR\Validators\ProcessingResult;
 use Twig\Token;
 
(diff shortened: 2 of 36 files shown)


src/Services/DeviceService.php
```diff
@@ -0,0 +1,100 @@
...
+            $record['code'] = $this->addCoding($record['diagnosis']);
+            $record['code_full'] = $record['diagnosis'];
+        }
+       
+        try {
+            $dataSet = json_decode($json, JSON_THROW_ON_ERROR);
+            $standardElements = $dataSet['standard_elements'] ?? [];
+            unset($record['udi_data']); // don't send back the JSON array
+            $record['udi_di'] = $standardElements['di'] ?? null;
+            // TODO: @adunsulag check with @brady.miller should this be companyName, or issuingAgency?
```
adunsulag, line 86: @bradymiller In your review can you tell me if this should be companyName for the device manufacturer in the JSON dump that we get back for the UDI, or whether using issuingAgency is the correct approach here.  I don't know enough about the UDI API / devices in general to determine this.

tests/Tests/Fixtures/BaseFixtureManager.php
```diff
@@ -63,4 +85,132 @@ public function getUnregisteredUuid()
...
+
+    private function isForeignReference(array $reference)
+    {
+        return isset($reference['table']) && isset($reference['columnSearch']) && isset($reference['columnSearchValue']) && isset($reference['columnReference']);
+    }
+
+    private function getQueryForForeignReference(array $reference): SearchQueryFragment
+    {
+        // these values come from flat files in the unit tests, I'm torn on whether we should escape these or not
+        // TODO: @adunsulag check with @brady.miller on whether we should escape these or not
```
adunsulag, line 196: @bradymiller These are only used in the unit tests, I've been wondering if it really matters to escape these tables since the json files are part of the codebase.

adunsulag:
Looks like I ran the tests with Jerry's test database.  I'll test with a fresh install and see what my results are.

adunsulag:
@bradymiller When you get a chance can you look into this? I can't for some reason have it resubmit the tests and I have no idea why the UUID generation all failed, nothing to do with any of that changed between 0e69ee3 and de13373

src/Services/CarePlanService.php
```diff
@@ -0,0 +1,276 @@
...
+                        ,codetext
+                        ,description
+                        ,`date`
+                        ,`encounter`
+                        ,`pid`
+                        ,`care_plan_type`
+                    FROM 
+                        form_care_plan
+                    WHERE
+                        `care_plan_type` = '$carePlanType'
```
bradymiller, line 140: prob still better to bind this even though already whitelisted in case anybody ever decides to expand this or decide to allow an item with a ? character in it

bradymiller: actually, after thinking about this, can't see why need to bind this, so can disregard above post :)

src/Services/CarePlanService.php
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /**
+     * Returns a list of all care plan resources.  Search array can be a simple key => value array which does an exact
+     * match on passed in value.  For more complicated searching @see CarePlanService::search().
+     * @param $search a key => value array
+     * @param bool $isAndCondition Whether the search should be a UNION of search values or INTERSECTION of search values
+     * @return ProcessingResult
+     */
+    public function getAll($search, $isAndCondition = true)
```
bradymiller, line 90: does there need to be support for a `$puuid = null` parameter?

src/Services/CarePlanService.php
```diff
@@ -0,0 +1,276 @@
...
+            if (!$value instanceof ISearchField) {
+                $newSearch[] = new StringSearchField($key, [$value], SearchModifier::EXACT);
+            } else {
+                $newSearch[$key] = $value;
+            }
+        }
+        return $this->search($search, $isAndCondition);
+    }
+
+    public function search($search, $isAndCondition = true)
```
bradymiller, line 103: does there need to be support for a `$puuid = null` parameter?

bradymiller: actually, looks like it is usually on the getAll request.

src/Services/DeviceService.php
```diff
@@ -0,0 +1,100 @@
...
+class DeviceService extends BaseService
+{
+    private const DEVICE_TABLE = "lists";
+    private $uuidRegistry;
+
+    public function __construct()
+    {
+        parent::__construct('lists');
+        $this->uuidRegistry = new UuidRegistry(['table_name' => self::DEVICE_TABLE]);
+        $this->uuidRegistry->createMissingUuids();
```
bradymiller, line 29: prob should also do for the patient_data table

src/Services/DeviceService.php
bradymiller: @adunsulag , would use companyName
(Issuing agency is one of the 3 agencies that create/release/track the DI code)

tests/Tests/Fixtures/BaseFixtureManager.php
bradymiller: rec escaping them, since never know where this stuff will lead. Plus will be nice to fail if for some reason the the requested table/column does not exist anymore for some reason.

bradymiller: again, though, no need to go way overboard on escaping here if it overcomplicates things :)

tests/Tests/Services/EncounterServiceTest.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+/**
+ * CarePlanServiceTest.php
+ * @package openemr
+ * @link      http://www.open-emr.org
+ * @author    Stephen Nielson <stephen@nielson.org>
+ * @copyright Copyright (c) 2021 Stephen Nielson <stephen@nielson.org>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
```
bradymiller, line 19: duplicate header

bradymiller:
very nice stuff!
just a couple minor questions
I'll check out the automated testing fails in more detail tomorrow.

src/Services/CarePlanService.php
adunsulag: This should be fixed now.

src/Services/DeviceService.php
adunsulag: This should be fixed now.

tests/Tests/Services/EncounterServiceTest.php
adunsulag: This should be fixed now.

tests/Tests/Fixtures/BaseFixtureManager.php
adunsulag: This should be fixed now.

src/Services/DeviceService.php
adunsulag: Thanks, I got this resolved now.

adunsulag:
@bradymiller I believe I figured out the unit tests issue.  Should be good to go now.

library/formdata.inc.php
```diff
@@ -214,9 +216,10 @@ function mitigateSqlTableUpperCase($s)
  *                                          characters (for example 'a-zA-Z0-9_').
  * @param   boolean      $die_if_no_match  If there is no match in the whitelist, then die and echo an error to screen and log.
  * @param   boolean      $case_sens_match  Use case sensitive match (this is default).
+ * @param   boolean      $throw_exception_if_no_match If there is no match in the whitelist then throw an exception
  * @return  string                         Escaped/sanitized sql identifier variable.
  */
-function escape_identifier($s, $whitelist_items, $die_if_no_match = false, $case_sens_match = true)
+function escape_identifier($s, $whitelist_items, $die_if_no_match = false, $case_sens_match = true, $throw_exception_if_no_match = false)
```
bradymiller, line 222: tenor99

src/Services/DeviceService.php
bradymiller: gotta run createMissingUuid on each one :)
Would prob just do following in case need the main $this->uuidRegistry in the class.
```
$this->uuidRegistry = new UuidRegistry(['table_name' => self::DEVICE_TABLE]);
$this->uuidRegistry->createMissingUuids();
(new UuidRegistry(['table_name' => self::PATIENT_TABLE]))->createMissingUuids();
```

src/Services/FHIR/FhirPractitionerService.php
```diff
@@ -301,7 +271,7 @@ public function updateOpenEMRRecord($fhirResourceId, $updatedOpenEMRRecord)
      */
     public function searchForOpenEMRRecords($openEMRSearchParameters, $puuidBind = null)
```
bradymiller, line 272: this seems odd that there is a parameter (`$puuidBind`) here that is not being used

bradymiller: disregard above post. i see why it is there :)

bradymiller:
@adunsulag , Just noted one super minor thing to fix. So gonna bring this PR and follow it with a commit for the fix (ie. keep things moving forward :) ). Thanks for keeping fhir alive!!

2DV (1)

PR #4423, opened 2021-05-27
CCDA Goals

sjpadgett:
- added new clinical note for assessment(evaluation)
- add new care plan type (Goal)


sjpadgett:
@bradymiller Looking ahead some I'm concerned about the new Health Concerns section of CDA.(see what I did there!:))
I'm thinking we may need another list like problems. Or modify problems to be a concern or problem.
Check out the spec description and let me know your thoughts. @adunsulag not sure if affects FHIR! Thanks.
https://www.hl7.org/ccdasearch/templates/2.16.840.1.113883.10.20.22.2.58.html

adunsulag:
@sjpadgett I went and looked that over, I don't see anything in that Health Concerns document that relates to FHIR at this point.  Doesn't mean it won't get added by ONC later though.

The database I got from you did not have any LOINC codes installed as far as I could tell.

sjpadgett:
Look in codes table for code_type 110. They may not be turned on where you needed them i.e in Code Type list.

adunsulag:
```SELECT * FROM `codes` WHERE code_type = 110``` Returns 0 results

**Comment by sjpadgett** [comment]
@adunsulag Emailed codes table to ya.

**Review by bradymiller — APPROVED** [review]

**Comment by bradymiller** [comment]
2DV (1)

**Comment by bradymiller** [comment]
@sjpadgett ,

Regarding Health Concern, I am not concerned. Ok, maybe a little concerned. It looks like literally anything can be set as a health concern (ie. requiring a status flag and dates of the concern). From a clinical practical standpoint, seems like something like this would just be noise for the provider and guessing there would be a inconsistent checking of everything being a concern in 1 patient and nothing being a concern in another patient dependent on the user/provider and the mood/time available of the mood/provider at the time (EMRs are already typically a mess when listing the myriad of patient medical problems etc., but hey lets add whole nother layer called concerns to clean things up :) ):
https://www.hl7.org/ccdasearch/templates/2.16.840.1.113883.10.20.22.4.132.html

The real question is if this is required :)
(if so, then would prob just make a flexible table that stored/mapped concern pointers that could point to anything)

**Comment by sjpadgett** [comment]
Does this make sense to anybody? 
```
Health Concerns

The tester verifies that the Health concerns are specified in accordance with the standard specified in § 170.205(a)(4) and § 170.205(a)(5) as narrative text.
Note that the system does not have to represent the Health concerns using the C-CDA’s syntax. The system must consistently and independently represent the Health concerns as discrete data that are clearly distinguishable.
```

stephenwaite:
basically what Stephen and Brady are saying right? emr just has to display it like the medical problems?

sjpadgett:
Last commit for this PR. Todo Concerns, Mental Status, Functional Status and Referral Reason sections. Will do all three next PR.
When's the last time CDA has been this complete? However, it doesn't seem to ever end....

image

stephenwaite:
amazed

PR #4427, opened 2021-05-29
New CCDA Templates

sjpadgett:
- Reason for Referral
- Mental Status
- Functional Status
- change encounter form for Mental status select. We don't need an active toggle.


sjpadgett:
Up next is Health Concerns which  I believe is last one for HL7 R2.1 V3 IG.

bradymiller (approved):

bradymiller:
2DV (1)

sjpadgett:
By spec the new health concerns meet requirement however, ATL states following which is unclear because the activity(required) subsection of section could have related entries pointing to specific problem for health concern. Simply does not make sense because a health concern may or may not become a problem and if it is a problem then why document a concern. Circle logic at best!!!
ATL from sample scenario:
```
N) HealthConcerns (Visual Inspection – ATL’s need to visually inspect the System Under Test
(SUT) generated C-CDA for the below narrative content)
a. Chronic Sickness exhibited by patient
b. HealthCare Concerns refer to underlying clinical facts
i. Documented HyperTension problem
ii. Documented HypoThyroidism problem
iii. Watch Weight of patient
```

My take is that the included health concern narrative simply contains the concerns in verbiage with the template activity pointing to the narrative text and not any related entries that point to a problem.

I have set up code in template to be able to add observation for concern and dealing with array of concern activities pointing to problems however, openemr currently doesn't have capability to set up problem to concern relationships. So, during testing if this becomes an issue with ATL then, I can fairly quickly fix though it doesn't make sense to do so.

sjpadgett:
@bradymiller want to take a look and ensure I have new DB items correct then i'll bring in. Unit test fail hasn't anything to do with my commits so will ignore.

bradymiller:
@sjpadgett , looks good to me (agree on ignoring the unit test fail)

sjpadgett:
Leaving open for a bit. Need to add some support for Plan of Care or now Plan of Treatment(how stupid, why change the name of a template!!!). 

Seriously, how is it the government can turn over writing and enforcing specifications to a bunch of middlemen(oh, MIDDLETHEM, gender neutral!) who all have a vested interest in writing specs that never end!

sjpadgett:
these are not conflicts!! they're changes. @bradymiller what's wrong here?

sjpadgett:
I think conflicts were on me where I may have simply forgot I previously committed a file or who knows but the git gods.
Bringing in and yes version for DB needs a bump but for now, i'm green with this PR and will not tempt those git gods!

stephenwaite:
the github actions gods were on your side! - all green checkmarks :)

bradymiller:
tenor99

Issue #4433, opened 2021-05-31
[Help Needed] Updating Bootstrap to 5.0.1


tywrenn:
Our exhaustive upgrade to Bootstrap 4 definitely improved much of the code base. Now that Bootstrap 5 is officially out, we will have to do it all over again. This is going to take awhile and I will probably not be able to help as much as I want to, but this has to be done.

The new version of Bootstrap doesn't really have too many code changes that need to happen. I'll be listing these all today and we will go from there.

sunsetsystems:
Thanks, I'll be interested to see developments. Also it would be nice to see some sort of guidance document for recommended usage of BS in OpenEMR. :)

robertdown:
I would like to recommend we pause this effort and focus on creating an abstraction layer between the code base and bootstrap. We would effectively use Bootstrap 5 class names but leverage a middle file to do some heavy customizations. I have admittedly not looked much into BS5, but I think at this point, for the sake of improving our rather unruly theming system, we stop using SCSS as much as possible, opting for CSS custom properties. This simplifies theming as we'd no longer have to recompile Bootstrap for every theme, instead a theme would simply be a set of variables. This allows us to leverage Boostrap (or whatever CSS framework comes next) without tying thousands of files to the external file. We would be creating a buffer between Bootstrap and the code base.

While on the topic of themes, we need to take a hard look at the way we are doing RTL. There's an RTL attribute for `<html>` tags that seems to flip everything on the screen. However, seeing as I am not fluent (nor can I even pretend to be) in any language that goes right to left I can't adequately understand if that actually solves the problems. The fact we are recompiling bootstrap for every theme twice is really not a sustainable build process. Especially if we are talking about adding a single attribute to the `<html>` tag. Anyway, just some thoughts about our UI.

PR #4434, opened 2021-06-01
Start Upgrade to BS5

tywrenn:
Updates #4433

This PR updates the repo to Bootstrap 5. More PR's will be required in order to get this fully functional. @bradymiller i'm requesting to get a new dedicated branch to fixing this so others can collaboratively fix this issue.


mdsupport:
Did some exploratory checks. This will be quite some effort complicated by ongoing CEHRT development.

Since the code relies on iframes, we need to explore possibility of at least 2 major versions of BS coexisting in the application (obviously not in same script) by explicitly calling BS5 in header for updated scripts. That will also force some cleanup of shared libraries generating complex html.

PR #4435, opened 2021-06-01
Fix for "Site ID is missing from session data!" when session expires.

### LINKED ISSUE #4426: "Site ID is missing from session data!" gives no clue to users as to what the situation is
If a session has expired, and a user goes to the site they will tend to see simply the text "Site ID is missing from session data!" and nothing else whatsoever on the screen. There's no option for them to click on anything to be taken to the login screen. I think that at the very least there should be a link for this. At best, users would be taken to the login page. As it is, users are confused by this very technical message and by the situation, and have to know to trim the URL to just the TLD in the location bar, or select the site from a bookmark, etc.

OpenEMR 6

sunsetsystems:
Fixes #4426

When an OpenEMR session expires, background tasks were closing the PHP session and thus causing the next user action to fail with  "Site ID is missing from session data!".  This PR keeps them from doing that.

Also, the iframe that runs interface/main/tabs/timeout_iframe.php is no longer useful and was removed.


bradymiller:
hi @sunsetsystems , code looks good. thanks for taking this on and simplifying this.

2DV (1)

bradymiller (approved):

sunsetsystems:
Thanks Brady. This should get into 6.0.0 as well.

sjpadgett:
Thanks for proving this Rod. I always thought those were suspect but could never prove it. This is a biggy!

stephenwaite:
bringing back to rel-600 for the upcoming patch 2 @sunsetsystems, thank you!

library/auth.inc
```diff
@@ -90,12 +90,7 @@ if (
     authLoginScreen(true);
 } else {
     // Check if session is valid (already logged in user)
-    if (AuthUtils::authCheckSession()) {
-        // Session is valid
-        if (isset($_SESSION['pid']) && empty($GLOBALS['DAEMON_FLAG'])) {
-            require_once("{$GLOBALS['srcdir']}/patient.inc");
```
stephenwaite, line 96: hi @sunsetsystems, looks like we're going to need to add this somewhere in here or the calling scripts, [like here](https://github.com/openemr/openemr/blob/master/interface/forms/newpatient/new.php#L15)

stephenwaite: just added it to `forms/new.php` in pending [PR](https://github.com/openemr/openemr/pull/4436/commits/014f9772021c392df21ed6e4d8ffcf498070793a)

sunsetsystems: Thanks @stephenwaite for the fix. That require_once was left over from when auth.inc logged patient info here, but it doesn't any more so this was clearly the wrong place for it.

stephenwaite: ok, nice, can't wait for my session to timeout :)

PR #4437, opened 2021-06-01
Weno cleanup

juggernautsei:
Fixes #

Removing old code that is no longer needed.


bradymiller:
hi @juggernautsei , Looks good and bringing in (will also bring into next patch). thanks for the improvements! -brady

PR #4439, opened 2021-06-02
updated timeout mechanism

### LINKED ISSUE #4438: Session timeout should force immediate logout
### Describe the problem

Wasn't sure whether to call this a feature request, bug or security vulnerability. When someone walks away from the computer their session will eventually time out, however whatever was on the screen remains visible until some action is taken. If that was sensitive information there's a problem.

My suggestion is that library/ajax/execute_background_services.php should notice when the user's session expires, and pass back a status code indicating such. The caller than can force a logout.

Even better, would be good to see a prominent notice (bright red?) somewhere at the top of the page when the session is close to expiring.

bradymiller:
fixes #4438

updated timeout mechanism

### DIFF
--- a/interface/main/tabs/js/user_data_view_model.js
+++ b/interface/main/tabs/js/user_data_view_model.js
@@ -133,3 +133,9 @@ function logout()
     top.restoreSession();
     document.getElementById("logoutinnerframe").src=webroot_url+"/interface/logout.php";
 }
+
+function timeoutLogout()
+{
+    top.restoreSession();
+    document.getElementById("logoutinnerframe").src=webroot_url+"/interface/logout.php?timeout=1";
+}
--- a/library/auth.inc
+++ b/library/auth.inc
@@ -85,7 +85,7 @@ if (
     // If session has timed out / been destroyed, logout record for null user/provider will be invalid.
     if (!empty($_SESSION['authUser']) && !empty($_SESSION['authProvider'])) {
         if ((isset($_GET['timeout'])) && ($_GET['timeout'] == "1")) {
-            EventAuditLogger::instance()->newEvent("timeout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
+            EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 0, "timeout, so force logout");
         } else {
             EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
         }
--- a/library/auth.inc
+++ b/library/auth.inc
@@ -85,7 +85,7 @@ if (
     // If session has timed out / been destroyed, logout record for null user/provider will be invalid.
     if (!empty($_SESSION['authUser']) && !empty($_SESSION['authProvider'])) {
         if ((isset($_GET['timeout'])) && ($_GET['timeout'] == "1")) {
-            EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success (timeout)");
+            EventAuditLogger::instance()->newEvent("timeout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
         } else {
             EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
         }
--- a/library/auth.inc
+++ b/library/auth.inc
@@ -84,7 +84,11 @@ if (
     // Logout
     // If session has timed out / been destroyed, logout record for null user/provider will be invalid.
     if (!empty($_SESSION['authUser']) && !empty($_SESSION['authProvider'])) {
-        EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
+        if ((isset($_GET['timeout'])) && ($_GET['timeout'] == "1")) {
+            EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success (timeout)");
+        } else {
+            EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
+        }
     }
     authCloseSession();
     authLoginScreen(true);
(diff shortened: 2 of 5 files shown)


sunsetsystems:
Looks good if it works!

library/auth.inc
```diff
@@ -84,7 +84,11 @@ if (
...
     if (!empty($_SESSION['authUser']) && !empty($_SESSION['authProvider'])) {
-        EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
+        if ((isset($_GET['timeout'])) && ($_GET['timeout'] == "1")) {
+            EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success (timeout)");
+        } else {
+            EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
+        }
     }
     authCloseSession();
     authLoginScreen(true);
```
sunsetsystems, line 94: Probably not for this commit, but at some point it may be useful to move functions authCloseSession and authLoginScreen into a different module so they can be accessed directly instead of going through a dedicated iframe.

library/auth.inc
```diff
@@ -84,7 +84,11 @@ if (
     // Logout
     // If session has timed out / been destroyed, logout record for null user/provider will be invalid.
```
stephenwaite, line 85: hi @bradymiller, so the log will now indicate that it was a successful logout (timeout)?

bradymiller:
I just realize that this code is mixing times from both php and mysql/mariadb. Is this a good idea (my fault since i wrote this class :) ) for such a critical feature (ie. if the time for php is set differently from mysql, then there is a possible chance of always being timed out)? I am thinking of always collecting time in php to prevent this issue.

Also points out another issue in OpenEMR where it looks like we can set the php time in globals, but it does not appear that the mysql/mariadb time is set.

library/auth.inc
bradymiller: yes, for now, but this may not be ideal. let me know if think of a better way to log this (ie. should i make it a 'timeout' instead of a 'logout' for the action)?

stephenwaite:
which class is mixing the time?

bradymiller:
Check out the SessionTracker class. When storing and updating the time, mysql/mariadb NOW() function is used. But when calculating the time to see if expired, php time is used (line 47).

stephenwaite:
forgot about no insertion into `uuid_registry` for these

is it possible to use UTC for these 2 dates?

library/auth.inc
```diff
@@ -84,7 +84,11 @@ if (
...
     if (!empty($_SESSION['authUser']) && !empty($_SESSION['authProvider'])) {
-        EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
+        if ((isset($_GET['timeout'])) && ($_GET['timeout'] == "1")) {
+            EventAuditLogger::instance()->newEvent("timeout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
```
bradymiller, line 88: @stephenwaite , changed to timeout. still a bit awkward, but should do job for now.

src/Common/Session/SessionTracker.php
```diff
@@ -39,12 +41,16 @@ public static function isSessionExpired(): bool
             error_log("OpenEMR Error: session_database_uuid session variable is missing");
             return true;
         }
-        $last_updated = sqlQueryNoLog("SELECT `last_updated` FROM `session_tracker` WHERE `uuid` = ?", $_SESSION['session_database_uuid'])['last_updated'];
-        if (empty($last_updated)) {
+        $sessionTracker = sqlQueryNoLog("SELECT `last_updated`, NOW() as `current_time` FROM `session_tracker` WHERE `uuid` = ?", $_SESSION['session_database_uuid']);
```
bradymiller, line 44: @stephenwaite , ended up being easier to just base all the sessionExpire stuff on the mysql/mariadb server. So, now should always be consistent since time is coming from one place.

library/auth.inc
```diff
@@ -85,7 +85,7 @@ if (
     // If session has timed out / been destroyed, logout record for null user/provider will be invalid.
     if (!empty($_SESSION['authUser']) && !empty($_SESSION['authProvider'])) {
         if ((isset($_GET['timeout'])) && ($_GET['timeout'] == "1")) {
-            EventAuditLogger::instance()->newEvent("timeout", $_SESSION['authUser'], $_SESSION['authProvider'], 1, "success");
+            EventAuditLogger::instance()->newEvent("logout", $_SESSION['authUser'], $_SESSION['authProvider'], 0, "timeout, so force logout");
```
stephenwaite, line 88: smaller

bradymiller: duh, just realized we already had a 'timeout' log entry message below in this same script :)
So, just made it the same :)

bradymiller:
testing well and bringing in

interface/main/tabs/js/user_data_view_model.js
```diff
@@ -133,3 +133,9 @@ function logout()
     top.restoreSession();
     document.getElementById("logoutinnerframe").src=webroot_url+"/interface/logout.php";
 }
+
+function timeoutLogout()
+{
+    top.restoreSession();
+    document.getElementById("logoutinnerframe").src=webroot_url+"/interface/logout.php?timeout=1";
+}
```
sunsetsystems, line 141: Because this file changed, $v_js_includes in version.php should be updated.

bradymiller: good point! will do.

stephenwaite:
should we recommend a `php.ini` setting that ensures [session.gc-maxlifetime](https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime) is at least `>=` `GLOBALS['timeout']` just in case the php session handler decides the session is garbage?

[forum discussion](https://community.open-emr.org/t/timeout-in-globals-not-being-honored/16700/11?u=stephenwaite)

stephenwaite:
thanks @bradymiller, should we add it to the faq in the wiki?

bradymiller:
OpenEMR should be overriding whatever the user sets in their php.ini, though. So, that setting in php.ini should not matter. Are there cases where a non-openemr session is being used?

stephenwaite:
ok, thanks for the explanation, not to my knowledge.

sunsetsystems:
Instead of 14400, might be better to set gc_maxlifetime using the timeout value in global settings plus a few minutes. Unless I'm missing something.

bradymiller:
Could do this. If do, though, need to ensure have a fallback set (to 14400) in case there is no globals context.

PR #4442, opened 2021-06-03
UUID bulk fast gen

adunsulag:
So this is a proof of concept pull request that I wanted to push out and see if this makes sense to the community in how to improve the UUID generation.  I improved the bulk insert speed to be 10X faster for anything larger than 10 records and about 2.5X larger for 2-10 records.  0-1 records missing gets a ~2X slower runtime. 

To generate uuids for the form_vitals a LOT of UUIDs are going to be generated for medium to large businesses.  I benchmarked the UUID stuff to try and see what that would look like, and generation of bulk uuids is very, very, slow.  I went through and wrote a new method called createMissingUuidsImproved() that dramatically improves the UUID generation with the tradeoff of it being slightly slower when there are no uuids missing and when there is only a single uuid missing.

I've committed a unit test that runs the benchmark and you can see the results here.  The times listed are in ms.  
image

You can see that with no records needed you get 0.84ms average with 2.25ms with the new method.  Single record performance is about double the execution with 5.7ms compared to 12.45ms for the new.  The big difference comes with comparing larger records.  At 200 records being generated you get 988ms compared to 118ms for the new method.

Larger data sets show the difference even more significantly:
20,000 records take 1m40s compared with 10s for the new method.

There's a bunch more that could be done to optimize, but this was a fairly straightforward improvement and I'm hoping that the tradeoffs are acceptable to people.  @bradymiller I know you've worked in this space a bunch so any thoughts on this would be appreciated.  If it looks good I'll replace the createMissingUuids() with the improved function.  I left both in so you could compare.


stephenwaite:
southpark_cheering

sunsetsystems:
Sorry I was not paying much attention when UUIDs were put into OpenEMR... is generation slow because of checking for duplicates?

src/Common/Uuid/UuidRegistry.php
```diff
@@ -241,7 +273,11 @@ public function createMissingUuids()
                         $prependAnd = true;
                     }
                     // First, see if it has already been set
-                    $setUuid = sqlQueryNoLog("SELECT `uuid` FROM `" . $this->table_name . "` WHERE `uuid` IS NOT NULL AND `uuid` != '' AND `uuid` != '\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0' AND " . $stringQuery, $arrayQuery)['uuid'];
+                    $sqlStatement = "SELECT `uuid` FROM `" . $this->table_name . "` WHERE `uuid` IS NOT NULL AND `uuid` != '' AND `uuid` != '\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0' AND " . $stringQuery;
+                    $setUuid = sqlQueryNoLog($sqlStatement, $arrayQuery);
+                    // original code here was not checking for whether a result was returned, not sure how it ever worked
+                    // as it should have thrown errors all over the place.
+                    $setUuid = $setUuid === false ? false : $setUuid['uuid'];
```
bradymiller, line 280: Are these several lines to improve performance or to fix something?
Note that the line that was removed has `['uuid']` to the right of it, so then $setUuid will only by not empty if something is returned.
if it returned a uuid then -> $setUuid would equal the uuid
if it did not return a uuid then -> $setUuid would be empty

src/Common/Uuid/UuidRegistry.php
```diff
@@ -305,4 +341,177 @@ public static function isEmptyBinaryUUID($uuidString)
...
+        {
+            return [];
+        }
+        $uuids = $this->getUUIDBatch($limit);
+        $dbUUIDs = [];
+
+        if (!$this->disable_tracker) {
+            $sqlColumns = array_map(function($u) { return '`uuid` = ?';}, $uuids);
+            $sqlWhere = implode(" OR ", $sqlColumns);
+            $dbUUIDs = QueryUtils::fetchRecordsNoLog("SELECT `uuid` FROM `uuid_registry` WHERE " . $sqlWhere, $uuids);
```
bradymiller, line 363: 2DV (1)

src/Common/Uuid/UuidRegistry.php
```diff
@@ -305,4 +341,177 @@ public static function isEmptyBinaryUUID($uuidString)
...
+            // loop through in batches of 1000
+            // generate min(1000, $count)
+            // generate bulk insert statement
+            $gen_count = min($count, self::UUID_MAX_BATCH_COUNT);
+            $batchUUids = $this->getUnusedUuidBatch($gen_count);
+            $ids = QueryUtils::fetchRecords("SELECT " . $this->table_id . " FROM `" . $this->table_name . "` WHERE `uuid` IS NULL OR `uuid` = '' OR `uuid` = '\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0' LIMIT " . $gen_count);
+            $this->insertUuidsIntoRegistry($batchUUids);
+            for ($i = 0; $i < $gen_count; $i++) {
+                // do single updates
+                sqlQueryNoLog("UPDATE `" . $this->table_name . "` SET `uuid` = ? WHERE `" . $this->table_id . "` = ?", [$batchUUids[$i], $ids[$i][$this->table_id]]);
```
bradymiller, line 408: seems odd to be using sqlQueryNoLog function here since that is meant to return a single row as array (so does an Fetchrow process on it). Maybe sqlStatementNoLog instead (possibly save a nano of a nano second :) )?

bradymiller:
nice stuff! Will test it a bit tomorrow.

bradymiller:
btw, rec changing over to using the new function in this PR.

src/Common/Uuid/UuidRegistry.php
adunsulag: It was to fix something.  If the SQL had an empty result set it returns false.  The line then becomes $setUuid = false['uuid'];  On my php setup this was throwing warnings which slows down php.  You can't take an array offset of an unknown key on a boolean.

src/Common/Uuid/UuidRegistry.php
adunsulag: This line was copied from the original createMissingUuids. I can change it to sqlStatementNoLog

adunsulag:
@sunsetsystems My opinion at the time that the UUID generation was put in was to trust the UUID algorithms and ignore the chance of collision since its so infinitesimally small the odds of it happening.  However, I didn't follow the discussion after that and the implementors decided it was necessary to check and make sure there are no duplicates in the database. That process as written was pretty slow as the database was being hit in an 3 times (select 1, insert 1 uuid, update 1) manner.  This batches the process, rolls it into a transaction, and checks against 1000 uuids at a time, batch inserts 1000 uuids, and then does an individual update for each table record.

src/Common/Uuid/UuidRegistry.php
bradymiller:
```
You can't take an array offset of an unknown key on a boolean.
```
yep, that doesn't sound very good :) (agree not good practice and surprised my php8 testing didn't catch that)

src/Common/Uuid/UuidRegistry.php
bradymiller: Lets change it and win back a nanosecond or 2 :)

src/Common/Uuid/UuidRegistry.php
```diff
@@ -199,6 +201,31 @@ private static function appendPopulateLog($table, $count, &$logEntry)
         }
     }
 
+    public function createMissingUuidsImproved()
```
bradymiller, line 204: To convert to this new mechanism, shouldn't this become createMissingUuids() and then the old createMissingUuids() be removed?

adunsulag: Wierd... I did this already, I wonder if I did it in a different branch or my rebase messed something up.  I'll change it again.

adunsulag:
In addition to the observation labs, I've included the following:

Fixed consultation_note LOINC code being incorrect
Implemented DocumentReference implementation for Clinical Notes and Patient Documents.  (#4205)
Patient Document access still has a number of issues such as LOINC categorization and UUID implementation.
Added search capability to base DocumentService.  
Put in a stubbed implementation of DiagnosticReport (#4307)
Added the base fhir url to the FhirServiceBase so we can make FQDN urls for our Document Reference class.



In terms of what's outstanding for the FHIR, we have the following:
**Single Patient API**

Missing Resources (woohoo down to just a couple)
	Diagnostic Report resource
	Procedures
	MedicationRequest
	Observation - Laboratory Results
	Provenance resource -- standalone resource endpoint


Fix Implementation
	Encounter -- missing elements
	Organization -- access right checks
	Practitioners -- access right checks
	CareTeam - team status flag
	DocumentReference - uuid for documents instead of device_uuid
	Observation - blood pressure figure out FHIR bug

API cleanup
	-- fix some general todos and other cleanup

**Multi-Patient API**
	For everything but patient implement the system/group/patient export for each of the resources
	These should go very fast (see FhirPatientService->export() as an example).

Visual Check
	Need to implement Patient previous name in FHIR
	Implement Token Invalidation

Additional Improvements
	Remember EHR-Launch practitioner login so login is not required again

**API Documentation**

Look at all the green checkmarks haha:
image
image
image

bradymiller:
@adunsulag ,
The vertical table mechanism was creating a new uuid for each row rather than each grouping. The more I looked into this, the more complicated it became :)
Was finally able to solve the issue in pure sql with some rather crazy queries, which is testing well so far. Can see the description of the problem in the commit and code comments. Hopefully didn't stray too far on this (happy to revert the commit if solution is no good).
I just brought in 3 commits into this PR:
1. counter return fix
2. fixed the vertical table mechanism
3. removed some redundant code in the older functions

bradymiller:
testing well and bringing in.

@adunsulag , just beautiful code you created btw and was super fun working with it (gonna do some more uuid work to incorporate the mapping stuff into upgrade and populateAllMissingUuids() function).

adunsulag:
@bradymiller glad you liked the code.  I didn't see any problems when I looked through it.

PR #4457, opened 2021-06-09
Fixed rule page rendering problem

gauravprocare:
Fixed rule page rendering problem

Fixes #


bradymiller:
hi @gauravprocare,
Thanks for posting this fix. As i recall, this is where an image link was populating a text field when it was empty. This code looks good. If this fix is testing well for you, then just let me know, and then I'll bring this in.
thanks! -brady

PR #4459, opened 2021-06-09
Temporary fix for deprecated warnings turning into exceptions.

sjpadgett:
I'll just reference issue #4458

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -653,10 +653,9 @@ function UrlIfImageExists($filename, $append = true)
 // Override temporary_files_dir
 $GLOBALS['temporary_files_dir'] = rtrim(sys_get_temp_dir(), '/');
 
-// turn off PHP compatibility warnings
-ini_set("session.bug_compat_warn", "off");
+error_reporting(error_reporting() & ~E_USER_DEPRECATED & ~E_USER_WARNING);
 // user debug mode
 if ((int) $GLOBALS['user_debug'] > 1) {
-    error_reporting(error_reporting() & ~E_WARNING & ~E_NOTICE & ~E_USER_WARNING);
+    error_reporting(error_reporting() & ~E_WARNING & ~E_NOTICE & ~E_USER_WARNING & ~E_USER_DEPRECATED);
     ini_set('display_errors', 1);
 }


stephenwaite:
`ini_set()` is deprecated too? lol

sjpadgett:
Surprised this hasn't come up till now. It will and is a timing/os issue and could be machine specific imo.
Merging when I can for patch 2.

> the `session.bug_compat_warn` i meant

yep is why I pulled it.

sjpadgett:
If ya want. Ya can go ahead and merge this and cherry pick if want something to do. I've seemed to exerted all my energy for the day....

stephenwaite:
be glad to, nice job debugging and patching!

stephenwaite:
will wait for Brady to :+1: :)

interface/globals.php
```diff
@@ -653,10 +653,9 @@ function UrlIfImageExists($filename, $append = true)
 // Override temporary_files_dir
 $GLOBALS['temporary_files_dir'] = rtrim(sys_get_temp_dir(), '/');
 
-// turn off PHP compatibility warnings
-ini_set("session.bug_compat_warn", "off");
+error_reporting(error_reporting() & ~E_USER_DEPRECATED & ~E_USER_WARNING);
```
bradymiller, line 656: should this line go into rel-600, but not master?
(so we fix the underlying warnings in master rather than remove them in order to avoid critical error from creeping up on us)

bradymiller: actually, we could maybe do following so then production versions would always do this while -dev would not:
```
if ( $v_tag != "-dev") {
}
```

sjpadgett: What critical errors? These are user defined which no one pays attention to anyway. Besides, i'd rather see the hard error so it gets taken care of. Besides, it's the wrong way for libraries to deprecate anyway. Should be these same as php does it.
In the end, I don't care.

bradymiller: If the trigger_error() mechanism will break things then agree prob best to always ignore these USER things.

sjpadgett: Agreed. Think about the time that went into tracking this down.
Merging.

bradymiller:
2DV

stephenwaite:
excited_elephant

Issue #4462, opened 2021-06-10
Add relationships to Clinical Notes & Care Plan


sjpadgett:
### Describe the problem
Just makes sense!
Difficult to make these relationships(problems, medications, procedure narratives and more) from API resources.
Also need document attachments in various places with relationships more specific to notes/procedures i.e. Document additions.

adunsulag:
Something to make sure we handle this properly when mapping USCDI v1 Summary of Data Classes and Data Elements.  This is what we need to support for FHIR.

For Clinical Notes:
**Consultation Note** maps to **DocumentReference** using 
**Discharge** maps to **DocumentReference** 
**Summary Note** maps to **DocumentReference**
**History &amp; Physical** maps to **DocumentReference**
**Imaging Narrative** maps to **DocumentReference,DiagnosticReport**
**Laboratory Report** maps **DocumentReference,DiagnosticReport**
**Pathology Report Narrative** maps to **DocumentReference,DiagnosticReport**
**Procedure Note** maps to **DocumentReference,DiagnosticReport**
**Progress Note** maps to **DocumentReference**
**Consultation Note** maps to **DocumentReference**

For the DiagnosticReport note resources we have to support minimum LOINC category codes of 
LP29684-5 | Radiology
LP29708-2 | Cardiology
LP7839-6 | Pathology

So our pathology notes obviously will use the Pathology category code, I'm guessing imaging will use either Radiology or Cardiology codes?  Diagnostic Reports can be any LOINC code but are intended to be anything that we want to present structured code information rather than just a narrative text.  The same information will be represented / contained in a DocumentReference resource.

For the clinical notes DocumentReference each document will need a category flag identifying the note as a clinical-note.  We've talked about using the current labels/tags in our document handling system which has problems since clinics can edit these tags which will break the system (as Jerry can attest to clinics deleting CCDA categories).

In addition our clinical note documents that are generated need to have a single type code that represents what kind of document this is.  The types for clinical notes must be a valid LOINC code but at a minimum must support the 5 core requirements: 18842-5 (Discharge summary), 11488-4 (Consult note), 34117-2 (History and physical note), 11506-3 (Progress note), 28570-0 (Procedure note).

For our Document Reference resource we also need to support who authored the document, the organization, and the encounter the document is attached to.  I think we already have all of this information though but we don't require it on a document upload.  We may want to consider requiring an encounter to be populated when creating these kinds of documents are generated.

sjpadgett:
[...]

PR #4465, opened 2021-06-12
Tag documents to procedures

sjpadgett:
- restyle view template properties
- cleanup classes
- add test type to picker and configuration

see issue #4462


sjpadgett:
Will bring this stage in separate so API can start engineering for use. @adunsulag 
I suspect there is more we can add to procedure order encounter form to better support procedures other than labs. Anyone know of anything missing?
Restyled to be better responsive and not so cluttered:
image

bradymiller:
2DV

PR #4466, opened 2021-06-13
Add issue relations to clinical notes

sjpadgett:
- from Text Templating feature

- Ensure Administration->Globals->Features `Enable Text Templates in Encounter Forms` is enabled.
- Double click in note narrative.
- Add from templating dialog.
image

Now `form_clinical_notes` has the ref mapping to lists table id in `note_related_to` column as json array.
image

See issue #4462

### DIFF
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -681,6 +681,6 @@ ALTER TABLE `form_vitals` ADD `oxygen_flow_rate` FLOAT(5,2) NULL DEFAULT '0.00';
 #EndIf
 
 #IfMissingColumn form_clinical_notes note_related_to
-ALTER TABLE `form_clinical_notes` ADD `note_related_to` TEXT NULL COMMENT 'Reference to lists id for note relationships(json)';
-ALTER TABLE `form_care_plan` ADD `note_related_to` TEXT NULL COMMENT 'Reference to lists id for note relationships(json)';
+ALTER TABLE `form_clinical_notes` ADD `note_related_to` TEXT COMMENT 'Reference to lists id for note relationships(json)';
+ALTER TABLE `form_care_plan` ADD `note_related_to` TEXT COMMENT 'Reference to lists id for note relationships(json)';
 #EndIf
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -679,3 +679,8 @@ INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`
 #IfMissingColumn form_vitals oxygen_flow_rate
 ALTER TABLE `form_vitals` ADD `oxygen_flow_rate` FLOAT(5,2) NULL DEFAULT '0.00';
 #EndIf
+
+#IfMissingColumn form_clinical_notes note_related_to
+ALTER TABLE `form_clinical_notes` ADD `note_related_to` TEXT NULL COMMENT 'Reference to lists id for note relationships(json)';
+ALTER TABLE `form_care_plan` ADD `note_related_to` TEXT NULL COMMENT 'Reference to lists id for note relationships(json)';
+#EndIf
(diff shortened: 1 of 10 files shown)


stephenwaite:
smaller

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -679,3 +679,8 @@ INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`
 #IfMissingColumn form_vitals oxygen_flow_rate
 ALTER TABLE `form_vitals` ADD `oxygen_flow_rate` FLOAT(5,2) NULL DEFAULT '0.00';
 #EndIf
+
+#IfMissingColumn form_clinical_notes note_related_to
+ALTER TABLE `form_clinical_notes` ADD `note_related_to` TEXT NULL COMMENT 'Reference to lists id for note relationships(json)';
+ALTER TABLE `form_care_plan` ADD `note_related_to` TEXT NULL COMMENT 'Reference to lists id for note relationships(json)';
```
bradymiller, line 685: drop the NULL in above 2 lines
(for text fields, we simplify things by not messing with default things)

bradymiller: I see that you replenished your semicolons though :)

2DV

bradymiller:
just noted super simple issue. otherwise looks great. OpenEMR cheer!

signal-2020-11-19-220815

sql/6_0_0-to-6_1_0_upgrade.sql
sjpadgett: Yeah, they're getting hard to get. Plenty colons, periods, commas and smiley faces but, semicolons, big demand!

sjpadgett:
@bradymiller I can only hope this will be useful in a normal production site. I suppose if documented well enough it could be useful. What do you think? Can I make better for clinicians?
Also i'm thinking perhaps best to turn on Text Template by default.

@adunsulag will leave open for a day in case you have comments or needs.

sjpadgett:
oops, left out some important changes like let dialog popup for new rows. Gave care plan some more room.
image

adunsulag:
@sjpadgett I've got your code in a local branch now, I'm just playing around with it to make sure there's no gotchas.  I'll let you know if I see problems.

adunsulag:
@sjpadgett I'm not sure if I'm following.  So in this case if there is a procedure's narrative note.  The connecting issues codes will be in the note_related_to column?  Its a JSON though right?  So I can't join on it or search on it correct?  That will have to be handled at the application layer if I'm following correctly.

sjpadgett:
@adunsulag True however it doesn't have to be json. I could use a delimited string or other. If you can think of better way let me know.

adunsulag:
@sjpadgett I know you're out the rest of this week so I'm still playing with things but so far it appears to be working for me so no problems from me if you want to bring this in.

sjpadgett:
I'm out Wed. on.. Merging.

adunsulag:
@sjpadgett BTW do we have old code in [contrib clinical notes](https://github.com/openemr/openemr/tree/master/contrib/forms/clinical_notes)?  It seems like a duplicate form but the table structure is very different from what is in core?  Should we remove that from the codebase?  It will break the FHIR Clinical Notes stuff if someone were to use that form in place of what is in core (which is the only reason I can think of why its in the codebase).

sjpadgett:
@bradymiller crap:)

bradymiller:
interesting dilemma :) 
Think there is a way to attack this in upgrade, but will take some time to work out ; will look into it after get patch 2 out.

adunsulag:
@sjpadgett @bradymiller so I'm going to guess this isn't old code that can just be thrown away...  I have to add a uuid anyway somewhere so I can fix things if I know what needs to be fixed or how the old form is used.

stephenwaite:
right @adunsulag , you can just never tell who copied that form into their codebase and fired it up so have to allow for it.

bradymiller:
@adunsulag , For your purposes, just pretend like the old form does not exist. My plans to deal with this will be something along the lines of if the old clinical_notes table exists in database (migrate prior clinical_notes to clinical_notes_2 in form folder, fix contrib folder, fix name, migrate sql table, and fix mapping in forms table).

adunsulag:
@bradymiller What I've done is I check in my FHIR search if the **code** column exists, if it does I proceed normally.  If it doesn't exist I just return an empty data set for FHIR.  That way nothing will break.  Should give us time for you to make those changes, or for us to architect a good solution that allows both tables to function.

bradymiller:
Sounds like a great plan!

PR #4467, opened 2021-06-15
Turn on templating by default

bradymiller:
Turn on templating by default

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -761,7 +761,7 @@ $GLOBALS_METADATA = array(
         'text_templates_enabled' => array(
             xl('Enable Text Templates in Encounter Forms'),
             'bool',                           // data type
-            '0',                              // default = false
+            '1',                              // default = true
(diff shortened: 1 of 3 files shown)


library/globals.inc.php
```diff
@@ -761,7 +761,7 @@ function gblTimeZones()
         'text_templates_enabled' => array(
             xl('Enable Text Templates in Encounter Forms'),
             'bool',                           // data type
-            '0',                              // default = false
+            '1',                              // default = true
```
bradymiller, line 764: @sjpadgett ,
Templating stuff is really nice (just tested it out) and definitely think it should be on by default :)
(btw, I finally caught a escaping issue; it had been too long and I was getting worried if I'd ever catch one in your code again :) )
(and the sql upgrade adjustment below if just for the very rare case where somebody may of uninstalled the form_clinical_notes form but kept the form_care_plan form; and vice versa)

tenor (17)

PR #4468, opened 2021-06-15
proper escaping of single quote in synthea filenames

bradymiller:
proper escaping of single quote in synthea filenames


stephenwaite:
PICKLE

PR #4472, opened 2021-06-18
fix dialog content height on IOS mobile devices.

sjpadgett:
See forum issue https://community.open-emr.org/t/ipad-screen-visibility-problem/16726/2

It appears to only show on browsers running on IOS devices (iPad/iPhone)


sjpadgett:
will cherry pick

bradymiller:
2DV

PR #4480, opened 2021-06-22
Fixing the log sync function

juggernautsei:
Fixes #

The changes are to separate the log from the transmit properties. Also, this eliminates the requirement of having an encounter open. This way the log sync can happen automatically in the background.


bradymiller:
hi @juggernautsei , If this is testing well for you and ready to come in, then just let us know. And will plan to bring this into rel-600 when it gets in.

PR #4487, opened 2021-06-28
Mostly PHP8 that can be picked to v6

sjpadgett:
- edit_list string to int

Fixes #

### DIFF
--- a/interface/main/tabs/main.php
+++ b/interface/main/tabs/main.php
@@ -81,6 +81,8 @@ $esignApi = new Api();
         var webroot_url = <?php echo js_escape($web_root); ?>;
         var jsLanguageDirection = <?php echo js_escape($_SESSION['language_direction']); ?>;
         var jsGlobals = {};
+        // used in tabs_view_model.js.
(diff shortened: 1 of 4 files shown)


bradymiller:
signal-2020-11-19-220815

sjpadgett:
Fixed reminders and background services to trigger 10 seconds apart.
I firmly believe they can cause contention via globals.

app now working well enough I can get back to ONC.:) Will leave PR open for a bit watching for more PHP8 errors.

stephenwaite:
checked back on demo site and all the tabs turned to `Unknown`

sjpadgett:
Is the session still good i.e. can you navigate?

stephenwaite:
no, when opening to view frame source it shows `Site id missing...`

sjpadgett:
Okay, I'm going to go ahead and bring this in as I believe services is the issue. Will cherry pick then we can test tomorrow after demos reset or @bradymiller is able to reset before then. I'll also test tonight local.

Thanks for the test Stephen. Very helpful.

sjpadgett:
not going to edit that last commit message as it is wrong so will note the allow enter in the globals search box in merge comment and hope it gets remembered. That had been driving me crazy for awhile.

PR #4500, opened 2021-07-02
Fixes #4499, php8 error, add SNOMED,LOINC codes.

### LINKED ISSUE #4499: PHP8 Edit Existing Procedure Order Form throws error
Procedure Order form throws an error on PHP8 when editing an existing order.

image

Looks like the procedure order array results is not initialized which causes count() to throw an error.

Impacted file is [interface/forms/procedure_order/common.php](https://github.com/openemr/openemr/blob/497c0b1436500328137f5f71924f1b66072c80f6/interface/forms/procedure_order/common.php#L1128)

adunsulag:
Fixed the PHP8 error where count was being used on an un-initialized
variable in procedure_order.  Editing an existing procedure order would trigger the error. Fixes #4499

Added the CPT4 and LOINC codes to the procedure_order file so we can
select them for diagnoses.

### DIFF
--- a/interface/forms/procedure_order/common.php
+++ b/interface/forms/procedure_order/common.php
@@ -662,7 +662,7 @@ function sel_related(varname) {
 // Also note the controlling script here runs from interface/patient_file/encounter/.
     let title = '<?php echo xla("Select Diagnosis Codes"); ?>';
     <?php /*echo attr(collect_codetypes("diagnosis", "csv")); */?>
-    dlgopen(top.webroot_url + '/interface/patient_file/encounter/find_code_dynamic.php?codetype=ICD10', '_blank', 985, 750, '', title);
+    dlgopen(top.webroot_url + '/interface/patient_file/encounter/find_code_dynamic.php?codetype=ICD10,LOINC,CPT4', '_blank', 985, 750, '', title);
 }
 
 // This is for callback by the find-code popup.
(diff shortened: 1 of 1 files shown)


stephenwaite:
southpark_cheering

interface/forms/procedure_order/common.php
```diff
@@ -662,7 +662,7 @@ function sel_related(varname) {
 // Also note the controlling script here runs from interface/patient_file/encounter/.
     let title = '<?php echo xla("Select Diagnosis Codes"); ?>';
     <?php /*echo attr(collect_codetypes("diagnosis", "csv")); */?>
-    dlgopen(top.webroot_url + '/interface/patient_file/encounter/find_code_dynamic.php?codetype=ICD10', '_blank', 985, 750, '', title);
+    dlgopen(top.webroot_url + '/interface/patient_file/encounter/find_code_dynamic.php?codetype=ICD10,LOINC,CPT4', '_blank', 985, 750, '', title);
```
sjpadgett, line 665: LOINC codes are iffy for diagnosis but for sure don't need cpt4. Better to use SNOMED instead

sjpadgett: I think these are unnecessary. perhaps I mislead. can we have a call?

bradymiller: Just checking if there was an outcome on what to include here?

adunsulag: I have to have LOINC codes on the procedure to be compliant with FHIR.  I added SNOMED per @sjpadgett's recommendations.  I'm fine with removing SNOMED if we don't need them.  I haven't talked to @sjpadgett about it yet.

bradymiller:
k, will fix the conflict and bring this in :)

PR #4503, opened 2021-07-05
Refactoring some of the Weno code

juggernautsei:
Fixes #

There is one file that I found that did not make it into any patch and that is the templates/prescription./general_fragment

The log sync is the focus of this refactoring. I had to refactor the date since the prescriptions are being pulled current.  Also, I was supposed to make a manual way to run the log sync. But I don't think that is necessary.

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -126,8 +126,8 @@ function checkBackgroundServices()
      * Setup background services for Weno when it is enabled
      * this is to sync the prescription logs
      */
-    $wenoservices = $GLOBALS['weno_rx_enable'] == 1 ? '1' : '0';
-    updateBackgroundService('WenoExchange', $wenoservices, 1);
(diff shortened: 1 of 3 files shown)


interface/super/edit_globals.php
```diff
@@ -126,8 +126,8 @@ function checkBackgroundServices()
      * Setup background services for Weno when it is enabled
      * this is to sync the prescription logs
      */
-    $wenoservices = $GLOBALS['weno_rx_enable'] == 1 ? '1' : '0';
-    updateBackgroundService('WenoExchange', $wenoservices, 1);
+    $wenoservices = $GLOBALS['weno_rx_enable'] == 1 ? '0' : '1';
```
bradymiller, line 129: this seems odd. Won't this turn on the weno background service when the weno_rx_enable global is set to off?

bradymiller:
hi @juggernautsei , noted an odd change in edit_globals.php above. Also, is the following related PR ready: https://github.com/openemr/openemr/pull/4480

PR #4526, opened 2021-07-13
utilizing cool new missing uuid creation mechanism

bradymiller:
utilizing cool new missing uuid creation mechanism created by @adunsulag

### DIFF
--- a/src/Common/Uuid/UuidMapping.php
+++ b/src/Common/Uuid/UuidMapping.php
@@ -16,10 +16,11 @@
 
 namespace OpenEMR\Common\Uuid;
 
+use Exception;
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Uuid\UuidRegistry;
-use OpenEMR\Services\FHIR\Observation\FhirObservationVitalsService;
 use OpenEMR\Services\FHIR\Observation\FhirObservationSocialHistoryService;
+use OpenEMR\Services\FHIR\Observation\FhirObservationVitalsService;
 
 class UuidMapping
 {
--- a/src/Common/Uuid/UuidMapping.php
+++ b/src/Common/Uuid/UuidMapping.php
@@ -61,7 +61,7 @@ class UuidMapping
                 $mappedCounter += $count;
             } else {
                 foreach ($mapping['codes'] as $code) {
-                    $count = self::createMissingResourceUuids($mapping['resource'], $mapping['table'], "category=" . $mapping['category']. "&code=" . $code['code']);
+                    $count = self::createMissingResourceUuids($mapping['resource'], $mapping['table'], "category=" . $mapping['category'] . "&code=" . $code['code']);
                     $mappedCounter += $count;
                 }
             }
--- a/src/Common/Uuid/UuidMapping.php
+++ b/src/Common/Uuid/UuidMapping.php
@@ -18,9 +18,19 @@ namespace OpenEMR\Common\Uuid;
 
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Uuid\UuidRegistry;
+use OpenEMR\Services\FHIR\Observation\FhirObservationVitalsService;
+use OpenEMR\Services\FHIR\Observation\FhirObservationSocialHistoryService;
 
 class UuidMapping
 {
+    private const UUID_MAPPING_DEFINITIONS = [
+        ['resource' => 'CareTeam', 'table' => 'patient_data'],
+        ['resource' => 'Location', 'table' => 'patient_data'],
+        ['resource' => 'Location', 'table' => 'users'],
+        ['resource' => 'Location', 'table' => 'facility'],
+        ['resource' => 'Observation', 'table' => 'form_vitals', 'codes' => FhirObservationVitalsService::COLUMN_MAPPINGS, 'category' => FhirObservationVitalsService::CATEGORY],
+        ['resource' => 'Observation', 'table' => 'history_data', 'codes' => FhirObservationSocialHistoryService::COLUMN_MAPPINGS, 'category' => FhirObservationSocialHistoryService::CATEGORY]
+    ];
 
     public static function getMappedRecordsForTableUUID($table_uuid)
     {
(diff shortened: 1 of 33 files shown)


stephenwaite:
signal-2021-05-13-141944

src/Services/SocialHistoryService.php
```diff
@@ -25,7 +25,7 @@ class SocialHistoryService extends BaseService
     public function __construct()
     {
         parent::__construct(self::TABLE_NAME);
-        $this->getUuidRegistry()->createMissingUuids();
+        UuidRegistry::createMissingUuidsForTables([self::TABLE_NAME]);
```
bradymiller, line 28: oops, will revert this change

bradymiller: @adunsulag , or should I keep this change and drop the getUuidRegistry() function in this class?

bradymiller: ended up keeping this change and dropping the getUuidRegistry() function :)

adunsulag: @bradymiller I think that's fine. sorry for the late response.  Having 'fun' with vitals.

sjpadgett: Need these? Also, i'm thinking we could automate interpretations from vital item ranges set in lists eg. heart rate 60-80 Normal etc.

    "2.16.840.1.113883.5.83": {
        name: "HL7 Result Interpretation",
        uri: "http://hl7.org/codes/ResultInterpretation#",
        table: {
            "B": "better",
            "D": "decreased",
            "U": "increased",
            "W": "worse",
            "<": "low off scale",
            ">": "high off scale",
            "A": "Abnormal",
            "AA": "abnormal alert",
            "H": "High",
            "HH": "high alert",
            "L": "Low",
            "LL": "low alert",
            "N": "Normal",
            "I": "intermediate",
            "MS": "moderately susceptible",
            "R": "resistant",
            "S": "susceptible",
            "VS": "very susceptible",
            "EX": "outside threshold",
            "HX": "above high threshold",
            "LX": "below low threshold",

        }
    }

adunsulag: These are the codes that FHIR requires which are similar to what you have listed: http://hl7.org/fhir/R4/valueset-observation-interpretation.html

bradymiller:
Just brought in a commit with following updates to UuidMapping class:
1. These are now populated via background service
2. These are populated on upgrade
3. Incorporated @adunsulag speed improvements from UuidRegistration class, so now super fast

src/Common/Uuid/UuidMapping.php
```diff
@@ -18,9 +18,19 @@
 
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Uuid\UuidRegistry;
+use OpenEMR\Services\FHIR\Observation\FhirObservationVitalsService;
+use OpenEMR\Services\FHIR\Observation\FhirObservationSocialHistoryService;
```
bradymiller, line 22: flip this to order alphabetically

src/Common/Uuid/UuidMapping.php
```diff
@@ -39,10 +49,49 @@ public static function getMappingForUUID($uuid, $is_binary = false)
...
+        try {
+            sqlBeginTrans();
+            $counter = 0;
+            do {
+                $count = self::createMissingResourceUuidsStep($resource, $table, $resourcePath);
+                $counter += $count;
+            } while ($count > 0);
+            sqlCommitTrans();
+            return $counter;
+        } catch (Exception $exception) {
```
adunsulag, line 85: @bradymiller Make sure to add a ```use \Exception;``` clause to the file.  Otherwise if an exception is thrown here it will fatal error here as it can't find Exception in the OpenEMR\Common\Uuid namespace.  I ran into this problem somewhere else.  Not sure why PHPStorm doesn't auto-import / detect this problem.

src/Common/Uuid/UuidMapping.php
```diff
@@ -18,9 +18,19 @@
...
+use OpenEMR\Services\FHIR\Observation\FhirObservationSocialHistoryService;
 
 class UuidMapping
 {
+    private const UUID_MAPPING_DEFINITIONS = [
+        ['resource' => 'CareTeam', 'table' => 'patient_data'],
+        ['resource' => 'Location', 'table' => 'patient_data'],
+        ['resource' => 'Location', 'table' => 'users'],
+        ['resource' => 'Location', 'table' => 'facility'],
+        ['resource' => 'Observation', 'table' => 'form_vitals', 'codes' => FhirObservationVitalsService::COLUMN_MAPPINGS, 'category' => FhirObservationVitalsService::CATEGORY],
```
adunsulag, line 31: The work I'm doing may remove the need for the form_vitals table mapping, FYI, I'll know more today if that's going to be possible.

src/Services/FHIR/Observation/FhirObservationVitalsService.php
```diff
@@ -49,7 +49,7 @@ class FhirObservationVitalsService extends FhirServiceBase implements IPatientCo
 
     const CATEGORY = "vital-signs";
 
-    private const COLUMN_MAPPINGS = [
+    const COLUMN_MAPPINGS = [
```
adunsulag, line 52: The only downside about this, is that I wasn't sure I wanted to expose all of this data for external use of the class as some of the internal implementations may change (which is what I'm doing now).  Maybe we need another constant that has just the codes... I think this is fine for now, I may change this in the work I'm doing once I rebase on what you've put in.

adunsulag:
@bradymiller Everything looked good.  I only spotted one bug with the Exception class that should be super easy to fix.  Thanks for going the next step and handling the mapping table!

src/Common/Uuid/UuidMapping.php
bradymiller: done

bradymiller:
thanks for the review @adunsulag , Gonna do a little more testing on this today. Would you rather have me bring this in today or wait for your ongoing vitals work to be completed? (I'm ok with either)

adunsulag:
@bradymiller Go ahead and bring it in.  I can change things if needed on my end.  I'm still toying around with it.  Vitals is embedded in so many places, the many things I want to do are not necessarily what I'm going to be able to do haha.

bradymiller:
testing well and bringing in. the uuid revolution continues :)

PR #4535, opened 2021-07-18
Dynamically resolve host domain Uri for API's and CCDA

sjpadgett:
- fixes for prior PR

Having user set site host address for Api's in globals is problematic IMO. Also we need to be able to resolve domain and ports that may be behind proxies, load balancers etc to the originating requests address.
Since we can't know the many different environments we run under, this is my best attempt.

As a note: Since PHP7 and deprecated in PHP5.4 it is required to provide parentheses when using new for classes.(code sniffer: Parentheses must be used when instantiating a new class).
Patterns like `throw new \RuntimeException()` will error in code sniffer. Generally using namespace paths for new() instantiations.
Even though PHP doesn't seem to mind!

Btw: I added new $GLOBALS['qualified_site_addr'] for $GLOBALS['site_addr_oath'] + webroot eg. https://opensourcedemr.us:8300/openemr

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -21,6 +21,7 @@ if ($response !== true) {
 use OpenEMR\Core\Kernel;
 use OpenEMR\Core\ModulesApplication;
 use Dotenv\Dotenv;
+use OpenEMR\Services\VersionService;
 
 // Throw error if the php openssl module is not installed.
 if (!(extension_loaded('openssl'))) {
@@ -85,6 +86,31 @@ if (preg_match("/^[^\/]/", $web_root)) {
 //  set manually here:
 //   $webserver_root = "/var/www/openemr";
 //   $web_root =  "/openemr";
+
+$ResolveServerHost = static function () {
+    $scheme = $_SERVER['REQUEST_SCHEME'] . "://";
+    $possibleHostSources = array('HTTP_X_FORWARDED_HOST', 'HTTP_HOST', 'SERVER_NAME', 'SERVER_ADDR');
+    $sourceTransformations = array(
+        "HTTP_X_FORWARDED_HOST" => function ($value) {
+            $elements = explode(',', $value);
+            return trim(end($elements));
+        }
+    );
+    $host = '';
+    foreach ($possibleHostSources as $source) {
+        if (!empty($host)) {
+            break;
+        }
+        if (empty($_SERVER[$source])) {
+            continue;
+        }
+        $host = $_SERVER[$source];
+        if (array_key_exists($source, $sourceTransformations)) {
+            $host = $sourceTransformations[$source]($host);
+        }
+    }
+    return rtrim(trim($scheme . $host), "/");
+};
 
 // Debug function. Can expand for longer trace or file info.
 function GetCallingScriptName()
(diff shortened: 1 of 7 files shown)


library/globals.inc.php
```diff
@@ -2976,7 +2976,7 @@ function gblTimeZones()
         'site_addr_oath' => array(
             xl('Site Address (required for OAuth2, FHIR and CCDA)'),
             'text',
-            'https://' . $_SERVER['HTTP_HOST'] . $GLOBALS['webroot'],
+            $ResolveServerHost(), // anonymous function in globals.php
```
bradymiller, line 2979: Another option here is making this blank by default and changing the wording of the setting to be an override in the case that the automatic method does not work. Then the automatic method will get much more attention over the longer term (since will be used on every globals.php call) and can be adjusted (in addition to correctly configuring apache/nginx settings) as it gets exposed to more use cases.
Maybe something like
quick desc - Site Address Manual Override
long desc - Site Address Manual Override. Only need to set this if the automatic method does not work for OAuth. An example setting would be https://localhost:8300 .

interface/globals.php
```diff
@@ -86,6 +87,31 @@
 //   $webserver_root = "/var/www/openemr";
 //   $web_root =  "/openemr";
 
+$ResolveServerHost = static function () {
```
bradymiller, line 90: tenor99

library/globals.inc.php
sjpadgett: okay. that may be less confusing.

sjpadgett:
@adunsulag While I don't really need a service for ccda documents I started one awhile back that:
- fetch latest CDA uuid from repository
- fetch CDA from repository by uuid
- generate and return a CCD as html or xml(xml allows for imports and html is browser readable)

If you can use for FHIR I'll finish it up and include in this PR.

adunsulag:
@sjpadgett If you have it available, even if its not fully finished I can polish it up when I need to integrate it.  Up to you on whether you want to bundle it in this PR or put it in another one.

sjpadgett:
@adunsulag 
I went ahead and used the new service in portal to give an example of use case.
I haven't tested the core document methods yet but they should work.

sjpadgett:
Unsure why I can't resume a session from a call to CCM dispatch from service. 
I did add a button for the ONC CCD view(as example for new service) in demographics reports. Eventually will remove old ccr/ccd out of date documents.
image

sjpadgett:
This gives me a start to ensure CCD can be generated in various ways. The new service will need work and i'm researching replacing the npm libxmljs library to see if I can get the compiled size down to something more manageable. Currently around 56MB.
Either we bite the bullet and compile at build time or find a way to ensure ccdaservice is installed.

@bradymiller I want to merge so @adunsulag can take a look at service and see where we need improvements for his needs.
The main thing here was to sort out the host pathing for api's/ccda.

bradymiller:
@sjpadgett , looks fine to me on a quick review and feel free to bring in. Will look at the session passing stuff in more detail after its in the codebase.

sjpadgett:
@bradymiller When you build releases for windows, are they built from a windows machine? For dependencies i'd think the target machine is important.

bradymiller:
hi @sjpadgett ,
No, the builds are all on linux. At this point, this has worked fine on windows since just basically grabbing js packages. This will likely be an issue on the ccda npm stuff though since doing more complicated things (ie. running a nodejs server).

sjpadgett:
Easiest way may be to include a builder in our installer and upgrade.

bradymiller:
If the builds are going to be architecture dependent for ccda engine then will likely need to pass that requirement downstream from the package (except for the docker; since that environment is known where we are already building the ccda engine). And will be tough to make it a requirement since it requires access to shell, npm, specific environment. Probably best we can do is not make it required (ie. OpenEMR should not break if it is not installed) along with instruction sets to build it. And perhaps somebody could do builds with specific architectures in mind but guessing that would be difficult to consistently sustain over time (ie. may end up like the ubuntu packages :) ).

bradymiller:
quick aside on a cool post build functionality. can incorporate removal of directories after the build (in case figure out how to get ccda engine in builds) for the build/compile/src files (https://github.com/openemr/openemr/blob/master/build.xml#L92-L96)

sjpadgett:
Yep that part I get. The installer is now a bone I can't let go of!:)

bradymiller:
tenor (68)

interface/globals.php
```diff
@@ -227,16 +253,29 @@ function GetCallingScriptName()
...
+    if (!mkdir($concurrentDirectory = $GLOBALS['MPDF_WRITE_DIR'], 0755, true) && !is_dir($concurrentDirectory)) {
+        throw new RuntimeException(sprintf('Directory "%s" was not created', $concurrentDirectory));
+    }
+}
+
+if (empty($GLOBALS['site_addr_oath'])) {
+    $GLOBALS['site_addr_oath'] = $ResolveServerHost();
+}
+if (empty($GLOBALS['qualified_site_addr'])) {
+    $GLOBALS['qualified_site_addr'] = rtrim($GLOBALS['site_addr_oath'] . trim($GLOBALS['webroot']), "/");
```
bradymiller, line 278: Shouldn't the above 2 global sets be done after the main globals are brought in (after line 524)?
(or else $GLOBALS['site_addr_oath'] will get overwritten with the set site_addr_oath global and $GLOBALS['qualified_site_addr'] will no longer match it)

sjpadgett: thought I did ole eagle eye! are you fixing or me?

bradymiller: i got it. done :)

bradymiller: will push the PR sometime today

src/Services/CDADocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+        ]);
+        $response = $httpClient->request('GET', $url, [
+            'query' => [
+                'combination' => $pid,
+                'recipient' => 'self',
+                'view' => '1',
+                'site' => $_SESSION ['site_id'],
+                'sent_by_app' => 'core_api',
+                'me' => session_id(),
+                'XDEBUG_SESSION' => '1'
```
bradymiller, line 95: looks like `XDEBUG_SESSION` has weaseled itself back into the codebase :)

stephenwaite: make-it-look-easy

src/Services/CDADocumentService.php
```diff
@@ -0,0 +1,177 @@
...
+                    $content = fread($fccda, filesize($row['ccda_data']));
+                }
+                fclose($fccda);
+            }
+        }
+
+        return $content;
+    }
+
+    public function generateCCDHtml($pid): string
```
bradymiller, line 80: would it make sense to bypass the need to send all these "outside" requests? I'm happy to look into how to go through encounterccdadispatch without needing it to be external (ie. need to send a session_id).

src/Services/CDADocumentService.php
bradymiller: i'll remove it on my pr :)

src/Services/CDADocumentService.php
sjpadgett: I think it's fine as is. It's not like the id is a secret(in cookie) and is no different than globals opening the session which it still does anyway to do auth. Besides, this is a normal way to handle the sessions in api calls via http.


Up to you but i'd like to avoid it being too cumbersome to call.

PR #4538, opened 2021-07-20
minor stuff

bradymiller:
minor stuff
(mostly php8 warnings)


stephenwaite:
make-it-look-easy

bradymiller:
omg @sjpadgett , just checked out the ccda views and they don't look like their from the 1970's anymore. here we come 2021 :)

PR #4539, opened 2021-07-21
ccda service updates

sjpadgett:
- add nullFlavor for empty sections


interface/forms/care_plan/new.php
```diff
@@ -26,8 +26,8 @@
 $returnurl = 'encounter_top.php';
 $formid = (int)($_GET['id'] ?? 0);
 if (empty($formid)) {
-    $sql = "SELECT id, encounter FROM `form_care_plan` WHERE pid = ? AND encounter = ?  LIMIT 1";
```
adunsulag, line 29: @sjpadgett  Isn't the id and the forms.form_id the same?  I can see where you are checking to only grab ones that are not deleted, but just curious here.

sjpadgett: they are the same it's just we don't show deletes in the form_care_plan or form_clinical_notes when deleting from encounter, only the master forms table.

interface/forms/clinical_notes/save.php
```diff
@@ -89,8 +89,9 @@
             // the note).
             $record['user'] = $_SESSION["authUser"];
         }
-//        $record['note_related_to'] = parse_note($record['description']);
-        $record['note_related_to'] = $record['description'];
+        // this is for related issues to the note
+        $record['note_related_to'] = parse_note($record['description']);
+        //$record['note_related_to'] = $record['description'];
```
sjpadgett, line 94: @adunsulag This is important. not sure why you removed.

adunsulag: @sjpadgett  I'm really confused.  I remember doing work in here and I can't for the life of me remember why I would have commented this out as I do know its important.  My only thought is somehow I was in the middle of a refactor and forgot to save or got pulled into something else.  Sorry.

sjpadgett: no biggy, just wanted to ensure it wasn't breaking api...

bradymiller (approved):

sjpadgett:
Okay here is the completed 2015 Cures HL7 Rel 2.1 ToC Document. All required templates are finished(fingers crossed) and seem to be testing well. Merging.

image

bradymiller:
2DV

stephenwaite:
wow, looks great!

simpsons-cheering

adunsulag:
@sjpadgett Great job, it looks awesome!

PR #4541, opened 2021-07-22
Fix uuid for procedure_result for DiagnosticReport

adunsulag:
Minor pull request to fix the uuid's for procedure_result when a document is uploaded and tied to a procedure.


bradymiller:
@adunsulag , looks good and bringing in. thanks for the fix!

PR #4542, opened 2021-07-26
The great forms migration

bradymiller:
work in progress

plan:
1. migrating all contrib forms to interface/forms
2. test all forms (throw out broken ones)
3. php8 updates

progress:
next form to test: DAP

form issues (ie. forms on the chopping block):
None yet


sunsetsystems:
That's a lot of forms! Nothing for LBF? :)

mdsupport:
@sunsetsystems is correct. Majority of contrib forms are pure input/output - probably specific to a situation - that can now be handled with lbf or any templating engine. Instead of burdening future developers to maintain these rarely used forms, when need arises the time may be better spent in converting these forms into LBF import files that can be optionally installed by admins based on documentation.

bradymiller:
hi,
Figured if they are in the codebase, then they should work and be easy to start using.
Another option is simply removing them.
Happy to go either way on this (LBF conversion not option for me since would require much more effort).

mdsupport:
The only value I see in most of contrib forms is some domain knowledge - e.g. some pulm provided input in selecting what he/she needed for recording 30+ [bronchitis](https://github.com/openemr/openemr/blob/e508a9076359a30fd5ee9d090d0ec4332f67808e/interface/forms/bronchitis/save.php) observations.

Making this a standard form forces someone working on e.g. BS conversion to spend time fixing/reworking these forever when lbf or a template can handle the same list of fields with minimal incremental efforts. Or the FHIR project will be forced to spend time in mapping the bronchitis observations to FHIR standard(?). A few such objects may not be a problem but I worry if someone drops a good, project-wide idea because it will mean reworking 200 scripts.

As I suggested, actual conversion of any script should be taken up only when the need arises. From housekeeping perspective it would be better to find an 'archive' home for these. That can be a contrib branch containing each form as a commit that can be pulled up directly into interface/forms if and when needed. Keeping these in master means some fixes get applied mechanically without any real testing.

adunsulag:
Brady and I talked a great deal about the forms last week.  If they are in the master branch codebase it should be working code that doesn't break when users attempt to use it.  If the code doesn't work it should be removed in order to mitigate any security attack vectors and just poor user experience.  

My suggestion was putting the code into some kind of archive on the website.  Now perhaps another branch as @mdsupport mentioned could work as well that people could pull from.  Unfortunately, I don't see anyone but developers doing something like pulling from a branch, so if we intend clinicians or doctors to have access to, or use these forms a separate branch that wouldn't work that great.

The intent of making the forms standard forms is to make sure whatever is in the codebase is brought up to standards with the rest of the project, or the code should be removed.  I already ran into problems with FHIR because of attempting to upgrade standard forms and having contrib forms that were in conflict (we ended up removing the history_data table form in contrib that was conflicting). 

Brady had mentioned that we likely have a number of users using some of the legacy forms and so just removing them will cause problems if they aren't in the codebase.  That said if the pain outweighs the benefit of keeping those forms, I'm just fine with removing them.

bradymiller:
closing this since now migrating contrib forms to another repo

PR #4543, opened 2021-07-26
Patient Results correct units type fix

sjpadgett:
Modify query to select unit title instead of id.

Fixes #

Forum issue:
https://community.open-emr.org/t/results-unit-id-shows-instead-of-text-value/16930


stephenwaite:
make-it-look-easy

bradymiller:
tenor99

PR #4545, opened 2021-07-27
Openemr fhir procedures

adunsulag:
Implemented the Procedures resource for OpenEMR.  This includes surgeries and any procedures where the procedure type is not 'laboratory_test'.   Also implements a framework for supporting additional procedures as needed.  Procedures expect CPT4 or SNOMED codes.  If a code is not provide it will attempt to use the LOINC standard code set in the compendium.  If that is missing as well it returns a Data Absent Reason for the code.

image

Finishing this resource leaves just MedicationRequest as the last large FHIR resource endpoint outstanding followed by minor fixes in Encounter, Organization, CarePlan, and Encounter.  

Provenance also needs to be built on the endpoint but that one shouldn't be too tough since nearly everything is pointing back to the organization resources.


bradymiller:
hi @adunsulag , Code looks good. Just let me know if this is ready to bring in and then I'll bring it in. thanks!

adunsulag:
@bradymiller It should be ready to bring in.

bradymiller:
2DV

PR #4546, opened 2021-07-27
clean up the uuid upgrade mechanism

bradymiller:
clean up the uuid upgrade mechanism


sql_patch.php
```diff
@@ -53,6 +54,16 @@
     upgradeFromSqlFile('patch.sql');
     flush();
 
+    echo '<p style="font-weight:bold; text-align:left; color:green">' . "Going to update UUIDs (this could take some time)" . '</p>\n';
+    flush_echo();
+    $updateUuidLog = UuidRegistry::populateAllMissingUuids(false, true);
```
bradymiller, line 59: will make this a true, true (then will also be recorded in the log)

sql_upgrade.php
```diff
@@ -329,6 +330,16 @@ function pausePoll(othis) {
             }
             flush();
 
+            echo "<br /><p class='text-success'>Going to update UUIDs (this could take some time)<br />\n";
+            flush_echo();
+            $updateUuidLog = UuidRegistry::populateAllMissingUuids(false, true);
```
bradymiller, line 335: will make this a true, true (then will also be recorded in the log)

stephenwaite:
clapping

sql_upgrade.php
```diff
@@ -202,8 +203,36 @@ function progressStatus(msg = '') {
     let eventList = document.getElementById('status-message');
     let progressEl = document.getElementById('progress');
 
-    progressEl.style.width = processProgress + "%";
-    progressEl.innerHTML = processProgress + "%" + " v" + currentVersion;
+    if (currentVersion == "UUID") {
```
bradymiller, line 206: @sjpadgett , thought you'd like this status for the uuid (it basically just increments slower and slower as time moves on making it look like it is progressing without having any idea about the actual time it will take; it never gets to 99%, but when it's done it gets set to 100% and moves on :) ).
uuid creation is way faster now with @adunsulag 's changes (for example, just took a minute or so for couple hundred thousand uuids on my rough testing in upgrade with 10000 random patients)

bradymiller:
A mere 100K patient upgrade that created a mere 2 million or so uuids (this took about 1.5-2 hours).

image

bradymiller:
testing well and bringing in

stephenwaite:
hooray

PR #4550, opened 2021-07-29
Openemr fhir encounters

adunsulag:
Implemented FHIR encounters endpoint.  I had to change a couple of minor things in the encounter service, but the biggest change involved adding a discharge_disposition to form_encounters and showing the dropdown on the encounters form.  FHIR requires we at least support the property and have some resources flagged with it, fortunately its optional and doesn't have to be filled in for historical data.  

Here is all the happy green checkmarks from Inferno.
image

Here is what the discharge dropdown looks like:
image

Available list options are here.  Note I specifically called it Discharge Disposition for the list as the Appointment FHIR resource (not ONC required) also has a dischargeDisposition property which would likely use this same list.
image

### DIFF
--- a/interface/forms/newpatient/common.php
+++ b/interface/forms/newpatient/common.php
@@ -554,7 +554,7 @@ $ires = sqlStatement("SELECT id, type, title, begdate FROM lists WHERE " .
                                 $dischargeDisposiitons = $dischargeListDisposition->getOptionsByListName('discharge-disposition') ?? [];
                                 foreach ($dischargeDisposiitons as $dispositon) {
                                     $selected = $result['discharge_disposition'] == $dispositon['option_id'] ? "selected='selected'" : "";
-                                ?>
+                                    ?>
                                 <option value="<?php echo $dispositon['option_id']; ?>" <?php echo $selected; ?> ><?php echo $dispositon['title']; ?></option>
                                 <?php } ?>
                             </select>
--- a/interface/forms/newpatient/common.php
+++ b/interface/forms/newpatient/common.php
@@ -20,8 +20,9 @@ use OpenEMR\Common\Acl\AclExtended;
 use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Core\Header;
-use OpenEMR\Services\UserService;
 use OpenEMR\Services\FacilityService;
+use OpenEMR\Services\ListService;
+use OpenEMR\Services\UserService;
 use OpenEMR\OeUI\OemrUI;
 
 $facilityService = new FacilityService();
@@ -550,12 +551,12 @@ $ires = sqlStatement("SELECT id, type, title, begdate FROM lists WHERE " .
                         <div class="col-sm">
                             <select name='discharge_disposition' id='discharge_disposition' class='form-control'>
                                 <?php
-                                $dischargeListDisposition = new \OpenEMR\Services\ListService();
+                                $dischargeListDisposition = new ListService();
                                 $dischargeDisposiitons = $dischargeListDisposition->getOptionsByListName('discharge-disposition') ?? [];
                                 foreach ($dischargeDisposiitons as $dispositon) {
                                     $selected = $result['discharge_disposition'] == $dispositon['option_id'] ? "selected='selected'" : "";
                                     ?>
-                                <option value="<?php echo $dispositon['option_id']; ?>" <?php echo $selected; ?> ><?php echo $dispositon['title']; ?></option>
+                                <option value="<?php echo attr($dispositon['option_id']); ?>" <?php echo $selected; ?> ><?php echo text($dispositon['title']); ?></option>
                                 <?php } ?>
                             </select>
                         </div>
--- a/interface/forms/newpatient/common.php
+++ b/interface/forms/newpatient/common.php
@@ -542,6 +542,26 @@ $ires = sqlStatement("SELECT id, type, title, begdate FROM lists WHERE " .
                             ?>
                         </div>
                     </div>
+                    <!-- Discharge Disposition -->
+                    <div class="form-row align-items-center mt-2">
+                        <div class="col-sm-2">
+                            <label for='facility_id' class="text-right"><?php echo xlt('Discharge Disposition'); ?>:</label>
+                        </div>
+                        <div class="col-sm">
+                            <select name='discharge_disposition' id='discharge_disposition' class='form-control'>
+                                <?php
+                                $dischargeListDisposition = new \OpenEMR\Services\ListService();
+                                $dischargeDisposiitons = $dischargeListDisposition->getOptionsByListName('discharge-disposition') ?? [];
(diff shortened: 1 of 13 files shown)


stephenwaite:
so having an office visit always results in a discharge from that facility? seems like hospital creep on the ambulatory clinics

adunsulag:
@stephenwaite It defaults to 'Home' as the option.  Unfortunately its mandated that I support the field.  It doesn't have to be filled out but we do have to support the option.  If there is a better way we can address the requirement, I'm fine with that.

adunsulag:
@stephenwaite Funny enough this is actually reported in FHIR under the Encounter->hospitalization category and the HL7 comment for the field is the following:
https://www.hl7.org/fhir/us/core/StructureDefinition-us-core-encounter-definitions.html#Encounter.hospitalization
```
An Encounter may cover more than just the inpatient stay. Contexts such as outpatients, community clinics, and aged care facilities are also included.

The duration recorded in the period of this encounter covers the entire scope of this hospitalization record.
```

stephenwaite:
pretty silly, what if it's only displayed in the form if the api is turned on?

adunsulag:
Who will turn off the API?  If you want to use an ONC certified install you have to have the API turned on.  I suppose for people outside the USA they could turn off the FHIR api.

stephenwaite:
ok, my discharge disposition from this PR is leaving against advice :)

adunsulag:
Looks like the PHP 8 tests failed but its not a code issue, I know someone mentioned that there is a way to restart them when they haphazardly fail like that, but I don't know how to do that.

stephenwaite:
if you click on `Details` on the failed job, do you see the `Re-run jobs`?

Screenshot from 2021-07-29 16-13-33

adunsulag:
Nope, it must be an administrator function
image

stephenwaite:
looks like your last commit returned all green though so good to go

stephenwaite:
did restart all the jobs just for fun though

interface/forms/newpatient/common.php
```diff
@@ -542,6 +542,26 @@ function cancelClickedOld() {
...
                     </div>
+                    <!-- Discharge Disposition -->
+                    <div class="form-row align-items-center mt-2">
+                        <div class="col-sm-2">
+                            <label for='facility_id' class="text-right"><?php echo xlt('Discharge Disposition'); ?>:</label>
+                        </div>
+                        <div class="col-sm">
+                            <select name='discharge_disposition' id='discharge_disposition' class='form-control'>
+                                <?php
+                                $dischargeListDisposition = new \OpenEMR\Services\ListService();
```
bradymiller, line 553: use the `use` block

interface/forms/newpatient/common.php
```diff
@@ -542,6 +542,26 @@ function cancelClickedOld() {
...
+                                $dischargeDisposiitons = $dischargeListDisposition->getOptionsByListName('discharge-disposition') ?? [];
+                                foreach ($dischargeDisposiitons as $dispositon) {
+                                    $selected = $result['discharge_disposition'] == $dispositon['option_id'] ? "selected='selected'" : "";
+                                    ?>
+                                <option value="<?php echo $dispositon['option_id']; ?>" <?php echo $selected; ?> ><?php echo $dispositon['title']; ?></option>
```
bradymiller, line 558: escaping in 2 places above:
```
<option value="<?php echo attr($dispositon['option_id']); ?>" <?php echo $selected; ?> ><?php echo text($dispositon['title']); ?></option>
```

interface/forms/newpatient/new.php
```diff
@@ -22,6 +22,7 @@
     ($tmp['squad'] && ! AclMain::aclCheckCore('squads', $tmp['squad'])) ||
     !AclMain::aclCheckForm('newpatient', '', array('write', 'addonly'))
 ) {
+    // TODO: why is this reversed?
     echo "<body>\n<html>\n";
```
bradymiller, line 26: agree that this is very odd :)

src/Services/FHIR/FhirLocationService.php
```diff
@@ -136,4 +143,18 @@ public function createProvenanceResource($dataRecord = array(), $encode = false)
     {
         // TODO: If Required in Future
     }
+
+    private function shouldIncludeContactInformationForLocationType($type)
+    {
+        if ($type == 'patient') {
+            // only those with access to a patient's demographic information can get their data
+            return AclMain::aclCheckCore("patients", "demo") !== false;
```
bradymiller, line 151: note that when a "patient" uses the api, there is no acl mechanism. instead are forcing in the sql queries that information pertinent to the patient is available (progressed from the separate api route, to the puuid binding stuff, then to your mechanism are forcing it your queries).

bradymiller: however, if this is a user that is accessing a patient demographics, then this is correct.

src/Services/FHIR/FhirLocationService.php
```diff
@@ -60,43 +66,44 @@ public function parseOpenEMRRecord($dataRecord = array(), $encode = false)
...
 
-        $locationResource->setAddress(array(
-            'line' => [$dataRecord['street']],
-            'city' => $dataRecord['city'],
-            'state' => $dataRecord['state'],
-            'postalCode' => $dataRecord['postal_code'],
-        ));
-
+        // TODO: what is this `s Home thing, seems really wierd... and its not documented
         if (!empty($dataRecord['name'] && $dataRecord['name'] != "`s Home")) {
```
bradymiller, line 70: Agree this does not make sense and don't think it's stored in the database like this, so unsure where this is coming from.

bradymiller:
review complete :)

interface/forms/newpatient/common.php
adunsulag: Should be fixed now.

interface/forms/newpatient/common.php
adunsulag: Should now be fixed.

interface/forms/newpatient/new.php
adunsulag: I left this alone... but I can change it if needed.

src/Services/FHIR/FhirLocationService.php
adunsulag: So I had to put in quite a bit of effort to handle the patient situation because of how locations operate.  You'll see the updates in the code.  Essentially if you are a patient you see only facility addresses and your own patient address.

src/Services/FHIR/FhirLocationService.php
adunsulag: So I found the original code in LocationService was adding the person's name and Home.  I've just set it to be Home Address

interface/forms/newpatient/new.php
bradymiller: no need to change. will preserve for historical purposes :)

bradymiller:
hi @adunsulag , code looks good and bringing in. thanks!

2DV

PR #4561, opened 2021-08-06
Fixes #4560 Support multiple care team records.

### LINKED ISSUE #4560: Support multiple CareTeam records per patient
I implemented and connected the FHIR CareTeam status property inside patient_data.  However, I discovered that the FHIR test suite requires that you support at least two care team records with different status (active, proposed, suspended, etc) for a single patient.

What I am thinking we do to resolve this issue is to create a patient_careteam_history table and anytime the careteam data changes in patient_data save off the record into the patient_careteam_history table.  Anything in the table will be treated as 'inactive'.  

Long term I think the best thing would be to 'normalize' the CareTeam and have it be a separate table since there are a lot of care team properties that could be reported back via FHIR that people may want to add in the long term. That will be a lot of engineering work, creating this table at least will allow us to track careteam changes over time and handle the need for at least two status elements (inactive, and whatever is currently set for the main careteam) per patient.

adunsulag:
Added a history table for the a patient's Care Team.  Any changes to the
care team providers or facilities that are part of the care team are
saved off to the patient_careteam_history table.  Note the status column
in patient_data is specifically for the current Care Team.  Past care
teams in the history table always have a status of 'inactive'.

I changed up the queries and report these changes back via FHIR.

I also made sure the records are cleaned up if a patient is deleted via the deleter.php script.

Long term we probably want to break up the care team properties in patient_data but that is a bigger engineering effort that hopefully can be taken on in the future.

### DIFF
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -585,12 +585,13 @@ INSERT INTO `ccda_sections` (`ccda_sections_id`, `ccda_components_id`, `ccda_sec
 #EndIf
 
 #IfNotRow2D list_options list_id lists option_id Care_Team_Status
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`) VALUES ('lists', 'Care_Team_Status', 'Care Team Status', '1');
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Care_Team_Status','active','Active',10,0,0);
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Care_Team_Status','inactive','Inactive',20,0,0);
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Care_Team_Status','suspended','Suspended',30,0,0);
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Care_Team_Status','proposed','Proposed',40,0,0);
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Care_Team_Status','entered-in-error','Entered In Error',50,0,0);
+INSERT INTO list_options (list_id,option_id,title, seq, is_default, option_value, notes) VALUES ('lists','Care_Team_Status','Care Team Status',0, 1, 0, 'This list originally comes from http://hl7.org/fhir/R4/valueset-care-team-status.html');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','proposed','Proposed',10,0,1, 'The care team has been drafted and proposed, but not yet participating in the coordination and delivery of patient care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','active','Active',20,1,1, 'The care team is currently participating in the coordination and delivery of care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','suspended','Suspended',30,0,1, 'The care team is temporarily on hold or suspended and not participating in the coordination and delivery of care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','inactive','Inactive',40,0,1, 'The care team was, but is no longer, participating in the coordination and delivery of care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','entered-in-error','Entered in Error',50,0,1, 'The care team should have never existed.');
+
 #EndIf
 
 #IfMissingColumn patient_data birth_fname
@@ -850,3 +851,29 @@ ALTER TABLE `form_vitals` ADD `inhaled_oxygen_concentration` float(4,1) DEFAULT 
 #EndIf
 
 UPDATE `list_options` SET `notes` = 'LOINC:11502-2' WHERE `list_options`.`list_id` = 'Clinical_Note_Type' AND `list_options`.`option_id` = 'laboratory_report_narrative';
+
+#IfMissingColumn patient_data care_team_status
+ALTER TABLE patient_data ADD COLUMN care_team_status VARCHAR(100) NULL DEFAULT NULL;
+#EndIf
+
+#IfNotTable patient_careteam_history
+CREATE TABLE `patient_careteam_history` (
+    `id` BIGINT(20) NOT NULL AUTO_INCREMENT
+    , `uuid` BINARY(16) NULL
+    , `care_team_provider` TEXT NULL
+    , `care_team_facility` TEXT NULL
+    , `date` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
+    , `pid` BIGINT(20) NOT NULL
+    , PRIMARY KEY (`id`)
+    , UNIQUE `uuid` (`uuid`)
+) ENGINE = InnoDB;
+#EndIf
+
+#IfNotRow2D layout_options form_id DEM field_id care_team_status
+SET @group_id = (SELECT group_id FROM layout_options WHERE field_id='care_team_provider' AND form_id='DEM');
+SET @backup_group_id = (SELECT group_id FROM layout_options WHERE field_id='DOB' AND form_id='DEM');
+SET @seq = (SELECT MAX(seq) FROM layout_options WHERE group_id = COALESCE(@group_id,@backup_group_id) AND form_id='DEM');
+-- strange that the list is
+INSERT INTO `layout_options` (`form_id`,`field_id`,`group_id`,`title`,`seq`,`data_type`,`uor`,`fld_length`,`max_length`,`list_id`,`titlecols`,`datacols`,`default_value`,`edit_options`,`description`,`fld_rows`)
+        VALUES ('DEM', 'care_team_status', COALESCE(@group_id,@backup_group_id), 'Care Team Status', @seq+1, 1, 1, 0, 0, 'Care_Team_Status', 1, 1, '', '', 'Indicates whether the care team is current , represents future intentions or is now a historical record.', 0);
+#EndIf
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -856,13 +856,13 @@ UPDATE `list_options` SET `notes` = 'LOINC:11502-2' WHERE `list_options`.`list_i
 ALTER TABLE patient_data ADD COLUMN care_team_status VARCHAR(100) NULL DEFAULT NULL;
 #EndIf
 
-#IfNotTable patient_careteam_history
-CREATE TABLE `patient_careteam_history` (
+#IfNotTable patient_history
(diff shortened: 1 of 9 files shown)


sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -850,3 +851,29 @@ ALTER TABLE `form_vitals` ADD `inhaled_oxygen_concentration` float(4,1) DEFAULT
...
+    `id` BIGINT(20) NOT NULL AUTO_INCREMENT
+    , `uuid` BINARY(16) NULL
+    , `care_team_provider` TEXT NULL
+    , `care_team_facility` TEXT NULL
+    , `date` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
+    , `pid` BIGINT(20) NOT NULL
+    , PRIMARY KEY (`id`)
+    , UNIQUE `uuid` (`uuid`)
+) ENGINE = InnoDB;
+#EndIf
```
sjpadgett, line 870: This bothers me. I don't oppose it but, I'd like to see this as a multi use table. First, most users are likely never going to use.  Second, I better review if a requirement in CCDA.:) I didn't see it!

adunsulag: So if you want to make it multi-use that's fine.  I just need to handle at least two care team records for a patient and this seemed like the simplest route to get there.

sjpadgett: providers/users/address/facility would all be linked tables from patient_history the same as for any patient history items.
I get it where you are panning for eventual relationships and this would be the way to go however while thinking about how to handle history of names I reflected back to a convo I had with Robert D. concerning maintaining lots of different histories.
So if not to much trouble could you rename the table patient_data_history or some such name?

sjpadgett:
don't see anything glaring so bringing in so I can continue work on my deal....

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -850,3 +851,29 @@ ALTER TABLE `form_vitals` ADD `inhaled_oxygen_concentration` float(4,1) DEFAULT
 #EndIf
 
 UPDATE `list_options` SET `notes` = 'LOINC:11502-2' WHERE `list_options`.`list_id` = 'Clinical_Note_Type' AND `list_options`.`option_id` = 'laboratory_report_narrative';
+
+#IfMissingColumn patient_data care_team_status
+ALTER TABLE patient_data ADD COLUMN care_team_status VARCHAR(100) NULL DEFAULT NULL;
```
bradymiller, line 856: @adunsulag ,
missing in database.sql

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -850,3 +851,29 @@ ALTER TABLE `form_vitals` ADD `inhaled_oxygen_concentration` float(4,1) DEFAULT
...
+#EndIf
+
+#IfNotTable patient_history
+CREATE TABLE `patient_history` (
+    `id` BIGINT(20) NOT NULL AUTO_INCREMENT
+    , `uuid` BINARY(16) NULL
+    , `date` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
+    , `care_team_provider` TEXT NULL
+    , `care_team_facility` TEXT NULL
```
bradymiller, line 865: @adunsulag ,
TEXT fields should be bare (also need to do in the entry in the database.sql script)

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -850,3 +851,29 @@ ALTER TABLE `form_vitals` ADD `inhaled_oxygen_concentration` float(4,1) DEFAULT
...
+) ENGINE = InnoDB;
+#EndIf
+
+#IfNotRow2D layout_options form_id DEM field_id care_team_status
+SET @group_id = (SELECT group_id FROM layout_options WHERE field_id='care_team_provider' AND form_id='DEM');
+SET @backup_group_id = (SELECT group_id FROM layout_options WHERE field_id='DOB' AND form_id='DEM');
+SET @seq = (SELECT MAX(seq) FROM layout_options WHERE group_id = COALESCE(@group_id,@backup_group_id) AND form_id='DEM');
+-- strange that the list is
+INSERT INTO `layout_options` (`form_id`,`field_id`,`group_id`,`title`,`seq`,`data_type`,`uor`,`fld_length`,`max_length`,`list_id`,`titlecols`,`datacols`,`default_value`,`edit_options`,`description`,`fld_rows`)
+        VALUES ('DEM', 'care_team_status', COALESCE(@group_id,@backup_group_id), 'Care Team Status', @seq+1, 1, 1, 0, 0, 'Care_Team_Status', 1, 1, '', '', 'Indicates whether the care team is current , represents future intentions or is now a historical record.', 0);
```
bradymiller, line 878: @adunsulag ,
missing in database.sql

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -585,12 +585,13 @@ INSERT INTO `ccda_sections` (`ccda_sections_id`, `ccda_components_id`, `ccda_sec
...
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','proposed','Proposed',10,0,1, 'The care team has been drafted and proposed, but not yet participating in the coordination and delivery of patient care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','active','Active',20,1,1, 'The care team is currently participating in the coordination and delivery of care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','suspended','Suspended',30,0,1, 'The care team is temporarily on hold or suspended and not participating in the coordination and delivery of care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','inactive','Inactive',40,0,1, 'The care team was, but is no longer, participating in the coordination and delivery of care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','entered-in-error','Entered in Error',50,0,1, 'The care team should have never existed.');
```
bradymiller, line 593: @adunsulag ,
What is the timeframe of these changes (ie. how long ago were the original entries placed)? And are there any updates in this that are critical (ie. if a user is running of development codebase and does not get these changes into their database)?

bradymiller:
@adunsulag 
To expedite the minor issues i noted (keep the momentum of all the cool stuff coming in :) ), I'll get these fixed and into the codebase this evening.

sql/6_0_0-to-6_1_0_upgrade.sql
bradymiller: another thing will change is using TEXT for this since patient_data has broken things in past because of the size of the fields (this is why everything we add to it now is basically TEXT to prevent breaking size limit especially for legacy users that are using lots more non-TEXT fields).

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -585,12 +585,13 @@ INSERT INTO `ccda_sections` (`ccda_sections_id`, `ccda_components_id`, `ccda_sec
...
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Care_Team_Status','proposed','Proposed',40,0,0);
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Care_Team_Status','entered-in-error','Entered In Error',50,0,0);
+INSERT INTO list_options (list_id,option_id,title, seq, is_default, option_value, notes) VALUES ('lists','Care_Team_Status','Care Team Status',0, 1, 0, 'This list originally comes from http://hl7.org/fhir/R4/valueset-care-team-status.html');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','proposed','Proposed',10,0,1, 'The care team has been drafted and proposed, but not yet participating in the coordination and delivery of patient care.');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity,notes) VALUES ('Care_Team_Status','active','Active',20,1,1, 'The care team is currently participating in the coordination and delivery of care.');
```
bradymiller, line 590: @adunsulag ,
Are you meaning to have active selected by default. This means whenever make a new patient this will be set to Active even if no care team provider(s) or facility(s are selected)?

bradymiller: i ended up removing this default setting for now in the followup PR since it was really odd to have in testing (ie. whenever make a new patient or edit a patient, it will default to active even though most users will never see or edit this when editing the demographics).

sql/6_0_0-to-6_1_0_upgrade.sql
bradymiller: i ended up migrating these changes to the end of the sql upgrade script in the follow up PR

bradymiller:
ended up resolving all issues in the following PR:
https://github.com/openemr/openemr/pull/4563

install and upgrading is working as expected. hopefully causes no issues in the care team functionality (it shouldn't, but i didn't really know how to test that).

sql/6_0_0-to-6_1_0_upgrade.sql
adunsulag: care_team_status is the option_id FROM list_options.option_id for the care team status.  I would think the column types should match there.  I'll add a comment onto the field to clarify what it points to.  Also will add it into database.sql.

sql/6_0_0-to-6_1_0_upgrade.sql
adunsulag: I've been defaulting list options so the user doesn't have to think about what to select and we can provide a value via FHIR.  It makes sense to remove the default here as if there is no care team we don't want a status.  I guess if we wanted to we could add javascript (or on the backend on save) to require the status to be selected if care team is provided.

If there is a CareTeam record we are supposed to have a value for the status.  If there is no value we report it as an unknown code.  According to US Core FHIR spec the unknown code and Data Absent Reason (DAR) is only supposed to be used for legacy data and when importing data from other systems where the value is not known.

adunsulag:
Thanks for the minor fixes, I'll pull the branches in my own stuff and see how it works.

sql/6_0_0-to-6_1_0_upgrade.sql
sjpadgett: actually an active status for no careteam would be correct imo.

sql/6_0_0-to-6_1_0_upgrade.sql
bradymiller: note there are a large amount of option_id's placed in patient_data that are text. The problem with that table is that it has too many columns (back in the day they were all varchar instead of text and it broke openemr since it was too large when users added new entries in layout editor) and legacy users still have this issue (even going to utf8mb4 may of broken their installs). So, decided to just use text (note this table may someday become vertical if @sunsetsystems has his wishes, which means these will all be placed in a text field anyways like the layout based forms are).

sql/6_0_0-to-6_1_0_upgrade.sql
bradymiller: if a user assigns a provider or facility in care team and doesn't set the status (ie. keeps is unassigned), can't we just assume it is 'Active' when used downstream in fhir etc.?

bradymiller: placed a PR that reverts the active to be the default setting (since guessing fhir needs to both support active and unassigned(unknown code)), which also will place active for all patients during the ugprade (one time thing):
https://github.com/openemr/openemr/pull/4565
(just let me know if want to bring this in and then I'll bring it in)

PR #4562, opened 2021-08-06
Openemr fhir provenance

adunsulag:
I took a creative approach to implementing the provenance. Instead of trying to figure out how to tie the provenance to each individual resource I use the resource and the resource uuid to uniquely identify the provenance resource.  Using that ID I can load the corresponding FHIR resource service and re-run all of the queries and grab the correct Provenance exactly as it was generated.  If we ever do need to grab a list of all Provenance records... that is going to be pretty hard to do as we don't track anything globally except for perhaps the uuid_registry which wouldn't be a perfect map for resources but could be implemented to do this...

I still am not sure that only putting the OpenEMR organization as the provenance source is the right approach as I would imagine most people would like to know the specific user who authored / created a resource.  However, based on Brady's and Jerry's recommendation we went with just the organization and it is passing the ONC test tool.

Since us-core doesn't require we list the provenance resources and we just need to handle the read operation I put in a search operation that requires the _id element to be listed.  

I had to modify the route parser so we could handle colon characters inside the route.

image

You can see how the provenance ids are created through this list of GET requests that Inferno fires off.
image

### DIFF
--- a/_rest_routes.inc.php
+++ b/_rest_routes.inc.php
@@ -1128,6 +1128,32 @@ RestConfig::$FHIR_ROUTE_MAP = array(
         RestConfig::apiLog($return);
         return $return;
     },
+    "GET /fhir/Provenance/:uuid" => function ($uuid, HttpRestRequest $request) {
+        if ($request->isPatientRequest()) {
+            // only allow access to data of binded patient
+            $return = (new FhirProvenanceRestController())->getOne($uuid, $request->getPatientUUIDString());
+        } else {
+            // TODO: it seems like regular users should be able to grab authorship / provenance information
+            RestConfig::authorization_check("admin", "super");
+            $return = (new FhirProvenanceRestController())->getOne($uuid);
+        }
+        RestConfig::apiLog($return);
+        return $return;
+    },
+    // NOTE: this GET request only supports requests with an _id parameter.  FHIR inferno test tool requires the 'search'
+    // property to support which is why this endpoint exists.
+    "GET /fhir/Provenance" => function ($uuid, HttpRestRequest $request) {
+        if ($request->isPatientRequest()) {
+            // only allow access to data of binded patient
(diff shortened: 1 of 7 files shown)


_rest_routes.inc.php
```diff
@@ -1128,6 +1128,32 @@
...
+    // NOTE: this GET request only supports requests with an _id parameter.  FHIR inferno test tool requires the 'search'
+    // property to support which is why this endpoint exists.
+    "GET /fhir/Provenance" => function ($uuid, HttpRestRequest $request) {
+        if ($request->isPatientRequest()) {
+            // only allow access to data of binded patient
+            $return = (new FhirProvenanceRestController())->getAll($request->getQueryParams(), $request->getPatientUUIDString());
+        } else {
+            // TODO: it seems like regular users should be able to grab authorship / provenance information
+            RestConfig::authorization_check("admin", "super");
+            $return = (new FhirProvenanceRestController())->getOne($request->getQueryParams());
```
bradymiller, line 1152: is this supposed to be getAll?

bradymiller:
hi @adunsulag , Code looks good to me and bringing in. thanks! -brady

_rest_routes.inc.php
bradymiller: @adunsulag ,
oh goodness, not sure where my brain went tonight (I brought this PR in even though i noted this issue :) ). If this needs fixing, let me know and I can bring it in or just bring in a PR and will bring it in.

adunsulag: @bradymiller I run into a major setback with Provenance on the Bulk-FHIR so I'm going to have to do some work here again anyways.  I'll do a PR and we can fix this.

sjpadgett: @adunsulag I'd like to talk to you about Provenance!

PR #4565, opened 2021-08-08
make active default setting for care_team_status

bradymiller:
@adunsulag and @sjpadgett ,
if need to support both active and unassigned, then reverted to make active the default setting in the list and converted blank items to active for upgrade (this is just a one time thing).


adunsulag:
@bradymiller This is a good change as we need to force it to be active when we have care plans.  I force it in the code to be active when its blank but I think its good to do it in the database as you've done.

bradymiller:
@adunsulag , sounds good and bringing this in. thanks! -brady

PR #4568, opened 2021-08-10
Fix bug in the careplan list endpoint

adunsulag:
Very minor fix with the careplan not working on a list all in a clinician context rather than the patient context.  Also added a comment to CareTeam to typedoc the service.


bradymiller (approved):

bradymiller:
2DV-2

stephenwaite:
sheep

PR #4574, opened 2021-08-11
Fix #4573 Openemr fhir token revocation

### LINKED ISSUE #4573: Access Token / Refresh Token Revocation
Inferno Test tool requires you to pass in a access token and refresh token and then for you to visually demonstrate the revocation of the token.  In order to make this work I've had to extend the Trusted User concept to allow granular revocation at the access and refresh token level.  I also found that every time a refresh token is given the old refresh token is revoked so we have to track that information as well.  

This will have an impact on people who are using the API if they are not storing the new refresh token each time they generate a new access token which apparently is what the spec calls for (I was not aware of this and it sure made API testing via Postman/Insomnia a lot nicer re-using the same refresh token).

I am updating the API Client List tool to allow you to see each trusted user that is connected to the oauth client and to see the individual access and refresh tokens for each trusted user.  You will be able to revoke a trusted user right from the interface as well as revoke individual access and refresh tokens from the interface.

I also will implement a utility screen that administrators can enter in an access / refresh token and get back information about that token as well as be able to revoke the token.

adunsulag:
Fixes #4573 FHIR Inferno requires you to provide an access token and a refresh token
that will be revoked in order to pass the test cases.  The developer has
to visually demonstrate revoking the token and then run the automated
test case.  This doesn't work with the trusted user mechanism we
originally had for token revokation as there isn't a way to tie the
token back to the trusted user given just the JWT.

I implemented the access token and refresh token revocation by adding a
revoked column and implementing the methods in the repositories to
handle token revocation.  I had to implement a separate isTokenRevoked
method for the access token due to the way League and OpenEMR interact
on the token revocation.  Because we store the site id inside the token
we have to allow League to validate a token and ignore the token
revocation validation check because we don't have access to the database
at that point the in execution (no site id, no database connection).

So we add the revocation check further down in the dispatch.php file in
order to handle the revocation check.

I added a lot of user interface to the Client Admin Controller to
display the list of trusted users and allow you to revoke the trusted
user right from the interface.  You can also revoke individual access
tokens. I've also added the ability to revoke individual refresh tokens 
from the admin GUI.

My next piece of work will be to add a Token Utilities button to the
client list screen.  That Token Utilities page will let you put in a
token and get back data about the token and allow you to revoke the
token from the system.  I will add that in as part of this pull request.

### DIFF
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -946,7 +946,7 @@ CREATE TABLE `api_refresh_token` (
     `expiry` DATETIME,
     `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMMENT '1=revoked,0=not revoked',
     PRIMARY KEY (`id`),
-    INDEX `api_refresh_token_token_idx` (`token`),
+    UNIQUE KEY (`token`),
     INDEX `api_refresh_token_usr_client_idx` (`client_id`, `user_id`)
 ) ENGINE = InnoDB COMMENT = 'Holds information about api refresh tokens.';
 #EndIf
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -932,3 +932,21 @@ INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, notes
 INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, notes) VALUES ('medication-request-intent','instance-order','Instance Order',70,0,1, 'The request represents an instance for the particular order, for example a medication administration record.');
 INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, notes) VALUES ('medication-request-intent','option','Option',80,0,1, 'The request represents a component or option for a RequestGroup that establishes timing, conditionality and/or other constraints among a set of requests.');
 #EndIf
+
+#IfMissingColumn api_token revoked
+ALTER TABLE `api_token` ADD COLUMN `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMMENT '1=revoked,0=not revoked';
+#EndIf
+
+#IfNotTable api_refresh_token
+CREATE TABLE `api_refresh_token` (
+    `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
+    `user_id` VARCHAR(40),
+    `client_id` VARCHAR(80),
+    `token` VARCHAR(128) NOT NULL,
+    `expiry` DATETIME,
+    `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMMENT '1=revoked,0=not revoked',
+    PRIMARY KEY (`id`),
+    INDEX `api_refresh_token_token_idx` (`token`),
+    INDEX `api_refresh_token_usr_client_idx` (`client_id`, `user_id`)
+) ENGINE = InnoDB COMMENT = 'Holds information about api refresh tokens.';
+#EndIf
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -940,10 +940,10 @@ ALTER TABLE `api_token` ADD COLUMN `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMME
 #IfNotTable api_refresh_token
 CREATE TABLE `api_refresh_token` (
     `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
-    `user_id` VARCHAR(40),
-    `client_id` VARCHAR(80),
+    `user_id` VARCHAR(40) DEFAULT NULL,
+    `client_id` VARCHAR(80) DEFAULT NULL,
     `token` VARCHAR(128) NOT NULL,
-    `expiry` DATETIME,
+    `expiry` DATETIME DEFAULT NULL,
     `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMMENT '1=revoked,0=not revoked',
     PRIMARY KEY (`id`),
     UNIQUE KEY (`token`),
--- a/src/Common/Auth/OAuth2KeyConfig.php
+++ b/src/Common/Auth/OAuth2KeyConfig.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Oauth2KeyConfig is responsible for configuring, generating, and returning oauth2 keys that are used by the OpenEMR system.
  * @package openemr
@@ -13,7 +14,6 @@
  */
 
 namespace OpenEMR\Common\Auth;
-
 
 use OpenEMR\Common\Crypto\CryptoGen;
 use OpenEMR\Common\Utils\RandomGenUtils;
@@ -42,8 +42,7 @@ class OAuth2KeyConfig
 
     public function __construct($siteDir = null)
     {
-        if (empty($siteDir))
-        {
+        if (empty($siteDir)) {
             // default to our global location
             $siteDir = $GLOBALS['OE_SITE_DIR'];
         }
--- a/src/Common/Auth/OAuth2KeyConfig.php
+++ b/src/Common/Auth/OAuth2KeyConfig.php
@@ -1,0 +1,162 @@
+<?php
+/**
+ * Oauth2KeyConfig is responsible for configuring, generating, and returning oauth2 keys that are used by the OpenEMR system.
+ * @package openemr
+ * @link      http://www.open-emr.org
+ * @author    Jerry Padgett <sjpadgett@gmail.com>
+ * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @author    Stephen Nielson <stephen@nielson.org>
+ * @copyright Copyright (c) 2020 Jerry Padgett <sjpadgett@gmail.com>
+ * @copyright Copyright (c) 2020 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2021 Stephen Nielson <stephen@nielson.org>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Common\Auth;
+
+
+use OpenEMR\Common\Crypto\CryptoGen;
+use OpenEMR\Common\Utils\RandomGenUtils;
+
+class OAuth2KeyConfig
+{
+    /**
+     * @var string encryption key stored in the database
+     */
+    private $oaEncryptionKey;
+
+    /**
+     * @var string OAUTH2 passphrase used for the private key
+     */
+    private $passphrase;
+
+    /**
+     * @var string File location of the oauth2 private key
+     */
+    private $privateKey;
+
+    /**
+     * @var string File location of the oauth2 public key
+     */
+    private $publicKey;
+
+    public function __construct($siteDir = null)
+    {
+        if (empty($siteDir))
+        {
+            // default to our global location
+            $siteDir = $GLOBALS['OE_SITE_DIR'];
+        }
+
+        // Create a crypto object that will be used for for encryption/decryption
+        $this->cryptoGen = new CryptoGen();
+        // verify and/or setup our key pairs.
+        $this->privateKey = $siteDir . '/documents/certificates/oaprivate.key';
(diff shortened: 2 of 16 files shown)


adunsulag:
Just realized I forgot to add the SQL for the initial database stuff.  Since this is a work in progress PR and more to get eyeballs on this I will work on that tomorrow.

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -932,3 +932,21 @@ INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, notes
...
+ALTER TABLE `api_token` ADD COLUMN `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMMENT '1=revoked,0=not revoked';
+#EndIf
+
+#IfNotTable api_refresh_token
+CREATE TABLE `api_refresh_token` (
+    `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
+    `user_id` VARCHAR(40),
+    `client_id` VARCHAR(80),
+    `token` VARCHAR(128) NOT NULL,
+    `expiry` DATETIME,
```
bradymiller, line 946: default date stuff to NULL

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -932,3 +932,21 @@ INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, notes
...
+#IfMissingColumn api_token revoked
+ALTER TABLE `api_token` ADD COLUMN `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMMENT '1=revoked,0=not revoked';
+#EndIf
+
+#IfNotTable api_refresh_token
+CREATE TABLE `api_refresh_token` (
+    `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
+    `user_id` VARCHAR(40),
+    `client_id` VARCHAR(80),
+    `token` VARCHAR(128) NOT NULL,
```
bradymiller, line 945: should the above 2 fields be matching api_token in the default settings (ie. null)?

src/Common/Auth/OpenIDConnect/Repositories/AccessTokenRepository.php
```diff
@@ -71,4 +100,30 @@ public function getNewToken(ClientEntityInterface $clientEntity, array $scopes,
...
+    public function getTokenById($id)
+    {
+        $sql = "SELECT * FROM api_token WHERE id = ? ";
+        $records = QueryUtils::fetchRecords($sql, [$id]);
+        return $records[0] ?? null;
+    }
+
+    public function getTokenByToken($token)
+    {
+        $sql = "SELECT * FROM api_token WHERE id = ? ";
```
bradymiller, line 125: is this query correct or is it supposed to be WHERE `token` = ...?

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -932,3 +932,21 @@ INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, notes
...
+    `expiry` DATETIME,
+    `revoked` TINYINT(1) NOT NULL DEFAULT 0 COMMENT '1=revoked,0=not revoked',
+    PRIMARY KEY (`id`),
+    INDEX `api_refresh_token_token_idx` (`token`),
```
bradymiller, line 949: should token be unique like in api_token table?

src/FHIR/SMART/ClientAdminController.php
```diff
@@ -130,12 +147,68 @@ public function editAction($clientId, $request)
...
+        $trustedUserService = new TrustedUserService();
+        $trustedUsers = $trustedUserService->getTrustedUsersForClient($clientId);
+
+        $usersWithAccessTokens = [];
+        $userService = new UserService();
+        $patientService = new PatientService();
+        foreach ($trustedUsers as $user) {
+            // TODO: do we need to optimize this query?
+            if (UuidRegistry::isValidStringUUID($user['user_id'])) {
+                $registryRecord = UuidRegistry::getRegistryRecordForUuid($user['user_id']);
```
bradymiller, line 159: 2DV

bradymiller: This is a neat use of the uuid registry table. But couldn't you just search directly in the patient_data table uuid column?

adunsulag: @bradymiller Problem is that the user_id can  beither a patient_data.uuid or a users.uuid so I would have to search both tables first.  That's why I hit the registry to figure out what kind of uuid it is.

sql/6_0_0-to-6_1_0_upgrade.sql
adunsulag: I'll make them match though I don't understand why they defaulted to NULL in the first place as things like expiry should always be filled out in my view.

src/Common/Auth/OpenIDConnect/Repositories/AccessTokenRepository.php
adunsulag: good catch, wonder why it was working haha.

adunsulag:
Here are screenshots for this pull request:  First off more Inferno checkmarks!
image

There is now a Token Tools button in the top right corner of the client apps page.
image

Clicking on that opens up the token tools page where you can paste in a refresh or access token to get details on the token and revoke the token if needed:
image
image

Hitting the revoke token will prevent the token from being used by api clients:
image

I also added to each api client a display of all of the authorized users and access/refresh tokens for a api client.  You can revoke individual tokens there as well as revoke trusted users from the interface:
image
image

I also added a few more fields such as registration date, json web keys, uris, etc.
image

sql/6_0_0-to-6_1_0_upgrade.sql
bradymiller: not addressed yet (they were addressed in database.sql though)

sql/6_0_0-to-6_1_0_upgrade.sql
bradymiller: not addressed yet (they were addressed in database.sql though)

src/Common/Auth/OpenIDConnect/Repositories/RefreshTokenRepository.php
```diff
@@ -13,28 +13,83 @@
...
+
+    public function getTokenById($id)
+    {
+        $sql = " SELECT * FROM api_refresh_token WHERE id = ?";
+        $resource = QueryUtils::sqlStatementThrowException($sql, [$id], true);
+        $result = QueryUtils::fetchArrayFromResultSet($resource);
+        return $result ?? null; // if the result set is not empty then its been revoked, otherwise its a good token
+    }
+
+    public function getTokenByToken($token)
```
bradymiller, line 88: this function is the same as the above getRefreshTokenByToken() function

adunsulag:
Hey @bradymiller I believe I've fixed the issues that you identified.  If it looks good to you, I'll bring it in.  Thanks for the second set of eyes!

src/FHIR/SMART/ClientAdminController.php
```diff
@@ -558,4 +843,273 @@ private function renderFooter()
...
+              <a href="<?php echo attr($parts['revoke_link']); ?>"
+                       class="btn btn-sm btn-primary float-right" onclick="top.restoreSession()"><?php echo xlt('Revoke Token'); ?></a>
+            <?php endif; ?>
+        </h3>
+        <ul>
+        <li><?php echo xlt("Token JTI(DB token value)"); ?>: <?php echo $parts['jti']; ?></li>
+        <li><?php echo xlt("Token DB Id"); ?>: <?php echo $databaseRecord['id']; ?></li>
+        <li><?php echo xlt("Token Status"); ?>: <?php echo $parts['status']; ?></li>
+        <li><?php echo xlt("Token Type"); ?>: <?php echo $parts['token_type']; ?></li>
+        <li><?php echo xlt("Token Expiration"); ?>: <?php echo $databaseRecord['expiry']; ?></li>
```
bradymiller, line 1030: above 5 lines need a text()

src/FHIR/SMART/ClientAdminController.php
```diff
@@ -558,4 +843,273 @@ private function renderFooter()
...
+        <li><?php echo xlt("User UUID"); ?>:
+        <a href="<?php echo attr($parts['user_link']); ?>">
+            <?php echo $databaseRecord['user_id']; ?>
```
bradymiller, line 1033: text() in above line

src/FHIR/SMART/ClientAdminController.php
```diff
@@ -558,4 +843,273 @@ private function renderFooter()
...
+        </a>
+        </li>
+        </ul>
+        <pre><?php echo json_encode($parts, JSON_PRETTY_PRINT); ?></pre>
```
bradymiller, line 1037: should wrap above with a text()
(at least I think we should since this is ouput in html context)

src/RestControllers/AuthorizationController.php
```diff
@@ -140,97 +142,17 @@ private function configKeyPairs(): void
...
+            $this->privateKey = $oauth2KeyConfig->getPrivateKeyLocation();
+            $this->publicKey = $oauth2KeyConfig->getPublicKeyLocation();
+            $this->oaEncryptionKey = $oauth2KeyConfig->getEncryptionKey();
+            $this->passphrase = $oauth2KeyConfig->getPassPhrase();
+        } catch (OAuth2KeyException $exception) {
+            $this->logger->error("OpenEMR error - " . $exception->getMessage() . ", so forced exit");
+            $serverException = OAuthServerException::serverError("Security error - problem with authorization server keys.", $exception);
             SessionUtil::oauthSessionCookieDestroy();
-            $this->emitResponse($exception->generateHttpResponse($response));
-            exit;
```
bradymiller, line 233: very cool to see this stuff get refactored into a separate class. Just making sure it will still force an exit from script after catching the exception and emitting the response (noted the `exit;` was removed)

src/RestControllers/AuthorizationController.php
```diff
@@ -655,6 +577,7 @@ public function getAuthorizationServer(): AuthorizationServer
             $responseType->markIsAuthorizationGrant(); // we have specific SMART responses for an authorization grant.
         }
 
+        $this->logger->debug("key log ", ['privateKey' => $this->privateKey, 'passphrase' => $this->passphrase, 'encryptionKey' => $this->oaEncryptionKey]);
```
bradymiller, line 580: I prob wouldn't log these actual values in a log file (since would consider them very sensitive info from security perspective). Since a novice user could have debug on at some point and potentially pass their log somewhere in unsecure fashion.

bradymiller: maybe just show if they are empty or not if helpful for debug purposes

src/RestControllers/AuthorizationController.php
```diff
@@ -1190,55 +1113,42 @@ public function tokenIntrospection(): void
...
-                    $token_hint = 'refresh_token';
-                }
+                $token_hint = $jsonWebKeyParser->getTokenHintFromToken($rawToken);
+//                // determine if access or refresh.
+//                $access_parts = explode(".", $rawToken);
+//                if (count($access_parts) === 3) {
+//                    $token_hint = 'access_token';
+//                } else {
+//                    $token_hint = 'refresh_token';
+//                }
```
bradymiller, line 1126: would remove rather than comment out

src/Common/Auth/OAuth2KeyConfig.php
```diff
@@ -0,0 +1,161 @@
...
+ * @copyright Copyright (c) 2021 Stephen Nielson <stephen@nielson.org>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Common\Auth;
+
+use OpenEMR\Common\Crypto\CryptoGen;
+use OpenEMR\Common\Utils\RandomGenUtils;
+
+class OAuth2KeyConfig
```
bradymiller, line 21: 2DV-2

bradymiller:
hi @adunsulag , noted a couple more minor issues. I promise that is it. After issues are addressed, then bring it in :)

src/RestControllers/AuthorizationController.php
adunsulag: Wow... that was only for my own debugging purposes.  Can't believe I left that in there.  I'm really, really glad you caught that.

PR #4579, opened 2021-08-12
Openemr fhir offline refresh scope

adunsulag:
So we had two main problems that we needed to address with refresh_tokens and the offline_access.  ONC requires that a refresh token is not issued if there is no offline_access scope present or if the client disables the offline_access scope.  We weren't handling that properly so I changed the grants to disable the refresh_token unless the offline_access scope is present.

The other thing is ONC requires that you clearly explain what the offline access enables to end users and give them a clear option to turn off offline access. You can see the warning message I've put here to make that clear to the end user what is happening.

image

I also bundled into this the ability to register a public / private app from the APP registration page.  It hides the system,user, and offline permissions if the public app option is chosen.
image

This satisfies all of the visual attestation checks for Inferno except for the api documentation piece (I still need to verify the bulk data doesn't return more than the scopes but I think we have this covered).
image


adunsulag:
Looks like I broke something :) I'll work on this more tomorrow.

interface/smart/register-app.php
```diff
@@ -225,6 +280,18 @@ function toggleSelectAll(evt) {
...
+                <div class="form-check form-check-inline">
+                    <input type="radio" class="form-check-input" id="appTypeConfidential" name="appType" value="private" checked="checked"/>
+                    <label for="appTypeConfidential" class="form-check-label pr-2"><?php echo xlt('Confidential'); ?></label>
+                    <input type="radio" class="form-check-input" id="appTypePublic" name="appType" value="public"/>
+                    <label for="appTypePublic" class="form-check-label"><?php echo xlt('Public'); ?></label>
+                </div>
+                <div class="row">
+                    <div class="col alert alert-info">
+                        <p><?php echo xlt("system, user, and offline_access scopes require confidential app permissions."); ?></p>
+                        <p><?php echo xlt("Confidential apps are applications that are able to safely and securely store a secret.  Browser based and many mobile applications do not satisfy this security constraint"); ?></p>
```
bradymiller, line 292: in translation constants, need to limit to single spacing (remove the double space after the period here). This is because the google doc sometimes truncates this from 2 spaces to 1 and then breaks the constant.

oauth2/provider/scope-authorize.php
```diff
@@ -76,6 +81,23 @@
                     </div>
                 </div>
             </div>
+            <?php if (true == $offline_requested) : ?>
+                <div class="row mt-3">
+                    <div class="col-12">
+                        <div class="alert alert-warning">
+                            <p>
+                                <?php echo xlt("This application has requested offline access to your data.  This permission will allow the data you authorize below to be accessed for an extended period of time"); ?>
```
bradymiller, line 89: change the double space to a single space in the translation constant

oauth2/provider/scope-authorize.php
```diff
@@ -76,6 +81,23 @@
...
+                            </p>
+                            <p><?php echo xlt("Offline access end date"); ?>: <?php echo $offline_access_date; ?></p>
```
bradymiller, line 91: text() around the $offline_access_date

bradymiller:
Just noted couple minor issues.

2DV

interface/smart/register-app.php
adunsulag: Good to know.  I've always been a two space guy, I guess I'll have to learn to adjust :)

oauth2/provider/scope-authorize.php
adunsulag: This is a hard-coded value but I'll change it anyways as someone may pull it from the database.

bradymiller (approved):

bradymiller:
looks good :)

tenor99

PR #4583, opened 2021-08-15
possible pathing fix

bradymiller:
@adunsulag , I just noticed this when searching the pathing to ensure the MU3 demos will work (the ones with `a`, `b`, and `c` in the paths). Noted this one was missing the webroot in it.


adunsulag:
@bradymiller Great catch and I'm glad we are testing this on the demos to catch things like this!

adunsulag:
Looks good @bradymiller I just ran into this with inferno, so I'll bring it in.
image

bradymiller:
2DV

PR #4584, opened 2021-08-15
fhir endpoints minor fixes

bradymiller:
@adunsulag , starting to plan the documentation stuff and just noted these minor issues while looking through the endpoints.
1. a duplicated endpoint
2. looks like a place where should use getAll instead of getOne

Prob will also be changing the id to uuid in the endpoints (where pertinent) just for consistency sake when documenting.


adunsulag:
@bradymiller I believe my bulk on fhir pull request fixes the duplicate endpoint.  I do still have the getOne bug but I can fix that in that pull request also.

bradymiller:
thanks @adunsulag :)
Closing this PR.

PR #4588, opened 2021-08-17
CQM support

sjpadgett:
- some core services fixups


sjpadgett:
@adunsulag unsure how handled codes in BaseService addCoding($diagnosis). Prob need to have discussion.
I'm use to seeing the code type(system) passed around and we may need to tie them to system OIDs.

src/Services/BaseService.php
```diff
@@ -284,8 +284,8 @@ public function selectHelper($sqlUpToFromStatement, $map)
     /**
      * Build and Throw Invalid Value Exception
      *
-     * @param $message              - The error message which will be displayed
-     * @param $type                 - Type of Exception
+     * @param $message - The error message which will be displayed
+     * @param $type    - Type of Exception
```
sjpadgett, line 288: We need to keep an eye on to ensure proper syntax e.g.
* @param $message - The error message which will be displayed

should be
* @param string $message - The error message which will be displayed

etc.

bradymiller:
CQM! CQM! CQM!

2DV-2

adunsulag:
@sjpadgett I can call you about it in a few hours once I get through my morning routine with my kids.  The addCoding stuff was there before I started working on FHIR.  I used it but never really liked it... I had to add in all of the systems myself  (which a couple are OIDs but most are URIs) at the FHIR service layer.

src/Services/BaseService.php
adunsulag: I will say in my defense... haha.  Good catch.
image

sjpadgett:
@adunsulag Thanks for fixing up coding array. I know it was a pain but will serve us well in future.

sjpadgett:
@adunsulag The previous name is broken in master because I forgot a file in PR. Fixed in this one plus will soon add the name suffix. Silly me!

sjpadgett:
Next PR will start bringing in the CQM support classes, models and dependencies. Still working on populating models from core patient data. Then I will need to decide on how we'll report i.e. the 'e' in eCQM..
@bradymiller Will you check me on new layout upgrade for suffix. Then bring in if I'm not around.
@adunsulag This will have your suffix support. No parsing.

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -1003,3 +1003,19 @@ INSERT INTO `module_acl_sections` (`section_id`, `section_name`, `parent_section
...
+ALTER TABLE `patient_data` ADD COLUMN `name_history` TINYTEXT;
+#EndIf
+
+#IfMissingColumn patient_data suffix
+ALTER TABLE `patient_data` ADD COLUMN `suffix` TINYTEXT;
+#EndIf
+
+#IfNotRow2D layout_options form_id DEM field_id suffix
+SET @group_id = (SELECT group_id FROM layout_options WHERE field_id='fname' AND form_id='DEM');
+UPDATE `layout_options` SET `seq` = `seq`*10 WHERE group_id = @group_id AND form_id='DEM';
```
bradymiller, line 1017: this is clever :)

bradymiller: @sjpadgett , made couple quick changes on this before bringing it in:
- removed the backup group id since lname will always be here and is needed to anchor the suffix
- in database.sql multiplied the seq by 10 there so they stay consistent with the upgrade

bradymiller:
bringing this in

thanks @sjpadgett !

2DV-2

sql/6_0_0-to-6_1_0_upgrade.sql
sjpadgett: Okay, thanks @bradymiller

PR #4591, opened 2021-08-17
Openemr base service coding fix

adunsulag:
We needed more information from the addCoding method for cqm for @sjpadgett  to work on.  I added
the information in so we could have the system,description in addition to the code.  Had to update a bunch of the FHIR services to accommodate the change.


adunsulag:
Well @sjpadgett I was wrong, it was a bigger change than I expected.  A good one, but a bigger change.  This will change a few things for people using the standard API who relied on the code to description mapping but I think that's ok.

bradymiller (approved):

bradymiller:
tenor99

PR #4592, opened 2021-08-18
Fix facility code to be NUCC for FHIR.

adunsulag:
As discussed in this issue #4590 instead of changing the facility column size we needed to fix the FHIR code to always use NUCC for the facility taxonomy when reporting in FHIR.


bradymiller (approved):

bradymiller:

PR #4594, opened 2021-08-18
Openemr fhir data absent reasons

### LINKED ISSUE #4593: Forward slash escaping in API causes issues with some API clients
So after many long hours of investigation into why the ONC Inferno test suite was failing the data absent coding and extension tests I finally figured out what was going on.  The test client does a straight up string search for a the data absent url such as http://terminology.hl7.org/CodeSystem/data-absent-reason in the API response body before it deserializes the response body.  However, since our PHP json_encode is escaping forward slash characters (which is optional in the spec) it couldn't find the URL as the URL ends up being http:\/\/terminology.hl7.org\/CodeSystem\/data-absent-reason.  

From what I've researched the only reason why we would need to escape forward slash characters is to embed JSON inside of a <script>.  Since our API responses are not doing that I believe it should be safe to remove the forward slash escaping and I'll submit a pull request to do that shortly.

adunsulag:
This fixes #4593  which removes escaping for the forward slash as well as adding a data absent reason to the Allergy Intolerance resource.  This was many, many hours of debugging ruby code to figure out what was happening.  I'm glad I finally figured this out.

Hooray for more check marks!
image

Here is my original commit message:
Our api endpoints were escaping the forward slashes in the JSON
responses.  This is optional in the JSON spec and is really only
necessary as far as I can tell when embedding JSON inside of an html
<script> tag.

However, some servers don't handle this forward slash escaping well.  In
the ONC Inferno test suite it does a strict string match for a URL.  The
match fails because of the escaped forward-slashes in the URL.  This is
why the data absent reason codes/extensions were failing.  It took many,
many hours to trace this down and fortunately its a simple fix.

This shouldn't effect any spec compliant clients who are consuming the
API as the forward slash escaping is an optional component and their
deserializers should still work.  Any non-spec compliant API client
could have issues.

### DIFF
--- a/src/Common/Http/HttpRestRouteHandler.php
+++ b/src/Common/Http/HttpRestRouteHandler.php
@@ -91,7 +91,11 @@ class HttpRestRouteHandler
                     if ($return_method === 'standard') {
                         header('Content-Type: application/json');
                         // if we fail to encode we WANT an error thrown
-                        echo json_encode($result, JSON_THROW_ON_ERROR);
+                        // PHP default json_encode will escape forward slash characters '/' so you can embed the JSON
+                        // inside of a <script> tag.  However, since forward slash escaping is optional as part of the
+                        // JSON spec some servers (looking at you ONC FHIR Inferno and your missing data tests) don't
+                        // know how to handle the unescaped slashes so we remove the forward slash escaping.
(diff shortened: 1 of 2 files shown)


bradymiller (approved):

bradymiller:
looks good and thanks for figuring this out!

2DV-2

PR #4599, opened 2021-08-19
Openemr single patient

adunsulag:
Hooray - Single Patient API complete!

This pull request implements some fixes on MedicationRequest as well as Procedures.  This also implements the patient previous, careteam, suffix and ethnicity.

We'll need to get the MU3 demos on the latest code, but if the data is setup on the instance the API now comes back 100% for single patient.  Hooray!

image

### DIFF
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -60,6 +60,10 @@
 --        arguments: table_name colname value colname2 value2 colname3 value3
 --        behavior:  If the table table_name does have a row where colname = value AND colname2 = value2 AND colname3 = value3, the block will be executed.
 
+--  #IfRowIsNull
+--    arguments: table_name colname
+--    behavior:  If the table table_name does have a row where colname is null, the block will be executed.
+
 --  #IfIndex
 --    desc:      This function is most often used for dropping of indexes/keys.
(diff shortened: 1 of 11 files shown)


sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -1003,3 +1007,9 @@ INSERT INTO `module_acl_sections` (`section_id`, `section_name`, `parent_section
 SET @group_id = (SELECT `id` FROM `gacl_aro_groups` WHERE `value` = 'admin' LIMIT 1);
 INSERT INTO `module_acl_group_settings` (`module_id`, `group_id`, `section_id`, `allowed`) VALUES (@module_id, @group_id, @section_id+1, 1);
 #EndIf
+
+#IfRowIsNull patient_history history_type_key NULL
```
bradymiller, line 1011: can `NULL` be removed from above line?

bradymiller:
@adunsulag , Only noted a minor issue in the upgrade script. Otherwise looks great and congrats on all the green checkmarks!

2DV-2

bradymiller:
@adunsulag , Actually, gonna bring this in and will remove that extraneous null on a proceeding commit. Then when the MU3 demos reset tonight (with updated snapshots from this evening of course :) ), they will have this in their codebase.

PR #4601, opened 2021-08-22
some fixes

bradymiller:
some fixes


stephenwaite:
popeye

adunsulag:
@bradymiller I looked through and everything looked good.  One question I had though is whether we are ever going to need translations with the setup.php?  It can probably be done at another time, but that was the only thing I saw from the review.

bradymiller:
Thanks for checking this PR out @adunsulag . Translating setup.php would definitely be a feat since no database yet. Agree, would be neat if we could support this but not sure if worth effort (unless somebody is interested in pursuing this).

Bringing in this PR.

PR #4602, opened 2021-08-23
Openemr swagger

adunsulag:
Brady this is just a very rudimentary implementation of an API code documentation generator using php-swagger and the OpenAPI specification (formerly called swagger).  I've gone through and documented a few APIs and generated an initial swagger file.  I mentioned that I would send this to you so I'm just putting this out there.

The advantages of this approach is that the API documentation should stay much more in sync with the actual codebase as the definitions come from the codebase itself an is auto-generated.  The next benefit is that swagger has the ability to run as an OAuth2 client and I believe with some additional setup it can be used to register an OpenID Connect client and use that registered client for the OAuth2 code flows.

Here are some sample screenshots:
image

image

image

So this is just a thought of something we could do.  It will be a bit of work to document things but in the end it'd be a nice addition.  I don't know if we actually want to go that route... There will be the difficulty of how to deal with multi-site installations and people having to learn the specific OpenAPI doctrine annotations just to document the API.  

Looking at how the swagger file is implemented I could see with a bit of effort us just generating a swagger file using our ROUTE_APIs automatically w/o having to use the annotation route.  I think it will be difficult to auto-generate specific API parameter descriptions but its possible that could be extracted from the code comments as well but that might be a big project.

This requires that we use the php-swagger project to generate the swagger file.  It leverages php's Doctrine Annotations project but that is only required on development machines whenever we make an API change and need to regenerate the API documentation.  End user machines would not need to have the doctrine/php library installed at all.

The api documentation can be run off a docker install which would let us stay in sync with any changes / updates needed to the html swagger viewer or it can also be run as an NPM package and hosted using Node.JS.  I wasn't sure if we'd really want that so I just use the static html viewer.


adunsulag:
BTW, I don't mind if we just say nope, doesn't work for us, too much work or anything like this.  I see pros and cons on both approaches.  I had done some work down this route and added a little bit more to give a rough prototype but I'm not vested in whether we want to do this or not.  A lot of developers have come to rely on OpenAPI so I can see uses, but I know Brady has put a lot of work already into our API documentation so I'm fine either way.

bradymiller:
@adunsulag , Very neat stuff! I'll play around with this a bit to get an idea of the work it will involve.

PR #4603, opened 2021-08-24
Add provenance author to code commits

adunsulag:
This doesn't really change the inferno test cases but adds more detailed author information if we have access to it for provenance.


bradymiller:
hi @adunsulag , looks great and bringing in. thanks for this bonus feature! -brady

PR #4631, opened 2021-09-17
Openemr events scope vitals

adunsulag:
Added an event to handle adding additional scopes via event listeners.
If a module adds to the REST api the api is pretty useless if there is
not an accompanying scope added with the api.

Also added some helper methods to the http request object and some
more api logging.  I needed this for a module I am building and it
will be useful for the community to have it in core.


bradymiller (approved):

bradymiller:
hi @adunsulag , looks great. thanks for the improvements!

tenor99

Issue #4614, opened 2021-09-05
Main tab bar is not mobile-friendly


sunsetsystems:
### Describe the bug

The main tab bar is not responsive to small devices. That is, with several tabs open on a narrow screen it does not collapse nor otherwise do anything to prevent a horizontal scrollbar. This means that content within a tab cannot be properly laid out for the device and is an obstacle to responsive behavior of the entire application.

### To Reproduce

1. Open a few tabs.
2. Make the main window narrow.
3. Note that the tabs remain in one row.

### Expected behavior

The tab bar items should stack vertically as required.

### Notes

This is built with Knockout JS and has some nice features such as the ability to "lock" a tab onscreen, but is not mobile-friendly. I'm thinking it needs a rewrite with Bootstrap classes that sacrifices this feature, but perhaps there's an easier/better way?

sunsetsystems:
Just a note that I've been working on bootstrapping LBF data entry, so this issue will become more relevant.

mdsupport:
@sunsetsystems Just a simple feedback from past 5-6 years of support. Tab in jquery-ui world to Nav in bootstrap world is relatively easy migration if it is driven by database tables. In fact we even had some javascript reformatter that redid current screen layout. However that only addresses overall navigation. The content / forms that work well on desktop simply cannot be reformatted to work even a small tablet and definitely not on 6" phone screen.
Our response to that problem was "App" feature added to login. That lets mobile users have all things they need for their role and suitable for their device form factor. As an example mobile user will never fully use an text area for data entry. But they can have 10s of checkboxes that builds a nice list in that text area.
Our main problem with Layouts is standard library functions do not follow model and view principle. Hope as you work on LBFs you make it layered where data layer builds an field object that can be processed in very few common view routines. That will make future bootstrap 4 to bootstrap 5 or 6 migration trivial.

sunsetsystems:
@mdsupport Thanks for the input. It won't be so fancy this time around, mainly I'm just converting tables to BS grids and making them stack for small devices and maintain sensible widths and spacing, without hurting the PC experience.

Some LBFs will present special problems for small screens and missing keyboards. For those the developer may want to look into LBF plug-ins, where custom form-specific code can be implemented on both server and client sides.

There's certainly much that can be done to make the code more modular and consistent. I'm pleased to see your interest in that.

mdsupport:
[...]

PR #4635, opened 2021-09-20
Bootstrapping LBF data entry

sunsetsystems:

This converts LBF data entry forms to use Bootstrap classes. Layout stacks to a single column at narrow widths.

There are a couple of things not addressed here that still get in the way of a full solution:
1. The top menu bar with search widget and logged-in user is too wide for some screen widths.
2. Issue #4614.

I'm relatively new to Bootstrap -- suggestions for improvement will be welcomed!

below added by bradymiller:
Can test this PR at following Up For Grabs demo: https://www.open-emr.org/wiki/index.php/Development_Demo#Epsilon_-_Up_For_Grabs_Demo


interface/patient_file/summary/demographics_full.php
```diff
@@ -619,303 +575,406 @@ function end_group()
...
+            <div class="col-md-3 pt-1">
+              <span class='label_custom required'><?php echo xlt('Subscriber Employer (SE)'); ?>:</span>
+            </div>
+            <div class="col-md-9">
+              <input type='entry' class='form-control form-control-sm' size='25'
+               name='i<?php echo attr($i); ?>subscriber_employer'
+               value="<?php echo attr($result3["subscriber_employer"] ?? ''); ?>"
+               onchange="capitalizeMe(this);" />
+               <span class='small mb-1'><br /><?php echo xlt('if unemployed enter Student'); ?>,
+               <?php echo xlt('PT Student, or leave blank'); ?></span>
```
sunsetsystems, line 661: Yes this should be one string, but I decided to leave the original language constants so that existing translations will still work. Don't feel strongly about it though.

bradymiller:
Can test this PR at following Up For Grabs demo: https://www.open-emr.org/wiki/index.php/Development_Demo#Epsilon_-_Up_For_Grabs_Demo

interface/patient_file/summary/demographics_full.php
bradymiller: sounds good to keep as is

sunsetsystems:
If no objections I'll merge this late tonight (US time).

bradymiller:
something seems off on this:
image

bradymiller:
also noted same issue as above on this screen (somehow the input field is getting wrapped to a new line):
image

bradymiller:
Also this screen is odd, where need to scroll to see it all:
image

image

sunsetsystems:
Thanks Brady. These are issues with the layouts.
1. Marital Status and Religion data columns were set to 3 but should be 1, so they wrapped to the next row.
2. Tables will auto-size column widths whereas bootstrap does not, so some layouts will look better if rearranged with more columns and more attention to the number of columns for each item. This PR does not attempt to do that... we do not want to break changes to the standard layouts.
3. In History and Lifestyle, Exam Results is still generated as a table to keep the items together, and this seems to work when columns are stacked for small devices. However they are still relatively wide and just don't fit for larger screens in the existing layout. Bottom line, the layout should be changed as noted in (2), or to put Exam Results in a separate set of rows.

I guess users will need to be warned about layout tweaks needed.

sunsetsystems:
Oh and I should at least tweak database.sql accordingly, for new sites!

sunsetsystems:
Also changed the UFG demo site layouts with these tweaks.

bradymiller:
Sounds like this is ready then :)
As an aside, in the future, we'll likely need to do some work on the name fields since can see folks putting names in wrong places (this has resulted from several different changes to the codebase; ie. not one change to blame :) )

sunsetsystems:
Probably want to add support for the "placeholder" attribute for that.
https://www.w3schools.com/TAGS/tryit.asp?filename=tryhtml5_input_placeholder

PR #4639, opened 2021-09-22
fix module name bug

bradymiller:
All the modules in unregistered section were using the same name if there is a no mal module being listed (they all use the normal module name).
image

This PR fixes this.
image

Will also bring this into rel-600


bradymiller:
regarding rel-600, on an interesting note, this fix is not needed there (guessing a php8 fix brought in this bug in master; and guessing it was me :) ).

PR #4644, opened 2021-09-24
Openemr encounters twig kernel

adunsulag:
Add empty option for encounter discharge.  Right now there is no way to opt out of setting an encounter's hospital discharge status.  Looking at the ONC FHIR encounter spec it looks like this is a MUST Support requirement and not a Required attribute.  Adding the optional value allows people to not fill this encounter in.

Made it so you can add additional search directories for twig.   This makes it so modules can use twig templates inside their own folders.

Added a helper method to the EncounterService to be able to get the most recent encounter for a patient based on the encounter date.

Modified the search method in EncounterService so you can have order by and limit conditions.

Removed spam log from patient demographics.

### DIFF
--- a/src/Services/EncounterService.php
+++ b/src/Services/EncounterService.php
@@ -77,15 +77,14 @@ class EncounterService extends BaseService
      * @param $pid The unique public id (pid) for the patient.
      * @return array|null Returns the encounter if found, null otherwise
      */
-    public function getMostRecentEncounterForPatient($pid) : ?array
+    public function getMostRecentEncounterForPatient($pid): ?array
     {
         $pid = new TokenSearchField('pid', [new TokenSearchValue($pid, null)]);
         // we discovered that most queries were ordering by encounter id which may NOT be the most recent encounter as
         // an older historical encounter may be entered after a more recent encounter, so ordering be encounter id screws
         // this up.
         $result = $this->search(['pid' => $pid], true, '', ['limit' => 1, 'order' => '`date` DESC']);
-        if ($result->hasData())
-        {
+        if ($result->hasData()) {
             return array_pop($result->getData());
         }
         return null;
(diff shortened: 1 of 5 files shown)


src/Services/EncounterService.php
```diff
@@ -234,7 +252,17 @@ class_code,
         try {
             $whereFragment = FhirSearchWhereClauseBuilder::build($search, $isAndCondition);
             $sql .= $whereFragment->getFragment();
-            $sql .= " ORDER BY fe.eid DESC";
+
+            if (empty($options['order'])) {
+                $sql .= " ORDER BY fe.eid DESC";
+            } else {
+                $sql .= " ORDER BY " . $options['order'];
```
bradymiller, line 259: where is $options coming from and what are some examples of what it will look like?
(just trying to figure out if it needs to be escaped)

src/Services/EncounterService.php
```diff
@@ -72,20 +72,38 @@ public function getUuidFields(): array
...
+     * @param $pid The unique public id (pid) for the patient.
+     * @return array|null Returns the encounter if found, null otherwise
+     */
+    public function getMostRecentEncounterForPatient($pid): ?array
+    {
+        $pid = new TokenSearchField('pid', [new TokenSearchValue($pid, null)]);
+        // we discovered that most queries were ordering by encounter id which may NOT be the most recent encounter as
+        // an older historical encounter may be entered after a more recent encounter, so ordering be encounter id screws
+        // this up.
+        $result = $this->search(['pid' => $pid], true, '', ['limit' => 1, 'order' => '`date` DESC']);
```
adunsulag, line 86: @bradymiller Here is the example where I use the options table.  I pass in a query limit and an order field.  I don't believe I need to escape it, but I can add that in if you'd like.

bradymiller: hi @adunsulag , agree no need to escape since hard-coded.

bradymiller (approved):

bradymiller:
tenor99

PR #4647, opened 2021-09-24
Adding billing to soap

juggernautsei:
Fixes #


stephenwaite:
excited_elephant

interface/forms/soap/C_FormSOAP.class.php
```diff
@@ -38,6 +40,13 @@ function default_action()
         return $this->fetch($this->template_dir . $this->template_mod . "_new.html");
     }
 
+    function hasBilling()
+    {
+        $sql = "SELECT id FROM billing WHERE encounter = ? AND activity = 1";
+        $billingEntered = sqlQuery($sql, [$_SESSION['encounter']]);
+        return $billingEntered['id'];
```
adunsulag, line 47: recommend doing ``` return $billingEntered['id'] ?? null;``` otherwise this will throw an array index error if there is no billing.

interface/forms/soap/templates/general_new.html
```diff
@@ -2,13 +2,43 @@
...
+        function set_related(codetype, code, selector, codedesc) {
+            if (code) {
+                if (codetype == 'ICD10') {
+                    document.getElementById('dxcode').value += code + " - " + codedesc + " ,, \r\n";
+                }
+                if (codetype == 'CPT4') {
+                    document.getElementById('cpt').value += code + " - " + codedesc + " ,, \r\n";
+                }
+                let xhr = new XMLHttpRequest();
+                xhr.onload = function() {
```
adunsulag, line 17: This seems odd, was this just for debugging?  Or were you still working on this?

src/Billing/AutoBilling.php
```diff
@@ -0,0 +1,320 @@
...
+    /**
+     * @param $pid
+     * @return mixed
+     */
+    public function getDiagnosis()
+    {
+        /**
+         * get the most recent diagnosis
+         */
+        $sql = "select title, diagnosis, type from lists where pid = ? ORDER BY id DESC LIMIT 1";
```
adunsulag, line 216: We store medications in lists.  Will this create problems for you here that you aren't excluding types such as medication or immunization?

adunsulag:
Saw a few minor things @juggernautsei.  Great job!

interface/forms/soap/templates/general_new.html
juggernautsei: If you are referring to the console log, not sure why that is there either other than I wanted to see a return during the build process. I am removing it. It's not needed.

src/Billing/AutoBilling.php
juggernautsei: I have used this code in other projects and that particular function is not used by anything at the moment. It is a dormant code. It can be enhanced to bring back specific types of list information. However, right now it does nothing.

juggernautsei: Looking this over, It seems that the hasBilling() function can be moved to the AutoBilling class. What are you thoughts on that?
Thanks for the compliment. Trying to work my way onto your team.

adunsulag: You're talking about the hasBilling in C_FormSOAP? Yea it makes more sense to put it in the AutoBilling class.

juggernautsei: It produces this when executed. This will work well for providers that don't use a lot of codes and use the SOAP note. 
image
This way they can do the note and billing on one page!

src/Billing/AutoBilling.php
```diff
@@ -0,0 +1,327 @@
...
+     * @param $enc
+     * @param $text
+     * @param $fees
+     * @param $icd
+     * @param $codetype
+     * @param $modifier
+     */
+    private function insertCPTBilling($event_date, $code, $pid, $provider, $userid, $enc, $text, $fees, $icd10, $codetype, $modifier)
+    {
+        $sql = "REPLACE INTO billing SET " .
```
mdsupport, line 160: REPLACE is a MySQL extension to the SQL standard. So for a project that avoids universal things like SQL views and SQL functions, it is best to implement Add/Change logic rather than this practice of clobbering existing record when the function is clearly called insertXXX.

This will apply to all REPLACE statements.

src/Billing/AutoBilling.php
```diff
@@ -0,0 +1,327 @@
...
+     * @param $provider
+     * @param $userid
+     * @param $enc
+     * @param $text
+     * @param $fees
+     * @param $icd
+     * @param $codetype
+     * @param $modifier
+     */
+    private function insertCPTBilling($event_date, $code, $pid, $provider, $userid, $enc, $text, $fees, $icd10, $codetype, $modifier)
```
mdsupport, line 158: Future developers will thank you if you avoid more than 3-4 parameters for an function. Specially in this case you clearly are providing database record properties and the function is executing an record method. So why not have the caller give you table name and column name/value pairs? Then you could even make these inserts generic.

juggernautsei: @mdsupport as a junior developer, I am still learning the ways to do things. I was focused on getting the feature to work. There are a few columns that are not needed for every entry. I could void those. I will study the other code in the system and figure out how to do what you suggested. I see @adunsulag do that all the time. I will learn.

src/Billing/AutoBilling.php
juggernautsei: "Add/Change logic" enlighten me, please. Show me Add/Change logic in the system, please.

sjpadgett: I think he means insert/update.

src/Billing/AutoBilling.php
mdsupport: @juggernautsei As stated, suggestion will benefit future developers as there is nothing wrong with your code.  Suggestion is to increase flexibility by passing e.g. a single assoc array (col name/value pairs) rather than just dropping parameters and hard coding their defaults - that would make the function even less usable in future. Benefit for the caller is if they have a blank or existing record, their code just fills/changes needed fields and call storage method (your function).

On the other hand if you anticipate this is highly specific situation, all the overhead is wasted effort.

Best.

mdsupport: @juggernautsei Maybe you copied some of the code from existing codebase. In that case reviewers will need to decide if fields that already have default values still need to be initialized to the same (default) values in the code.

mdsupport:
@juggernautsei If you have some flexibility, consider a select2 based approach to code selection. Your users will like that. To get a feel, try entering an immunization where CVX is selected using popup and the lot number is chosen by select2.

src/Billing/AutoBilling.php
adunsulag: Its a private method inside the class, not sure it really matters too much.  For better programming, I think the suggestion is good.  I personally prefer to keep my parameters at no more than 4-5 at most.  I prefer strongly typed objects to associative arrays but I've stuck with the associative array pattern in most places because we use it in the codebase everywhere.  It does increase the burden of documentation (where if there is no documentation of the parameters you have to violate information encapsulation by just diving into the method anyways).

juggernautsei:
@mdsupport and @adunsulag this is just what I need to take my skills to the next level. So, here goes. 

So, @mdsupport change this 

     private function insertCPTBilling($event_date, $code, $pid, $provider, $userid, $enc, $text, $fees, $icd10, $codetype, $modifier)

to this 

        //Build data array
        $billing_data_array = [];
         $billing_data_array['eventdate'] = $event_date;
                 $billing_data_array['code'] = $code;
                    $billing_data_array['pid'] = $this->pid;
            $billing_data_array['provider'] = $provider;
                $billing_data_array['userid'] = $this->userid;
          $billing_data_array['encounter'] = $enc;
                    $billing_data_array['text'] = $text;
                    $billing_data_array['fees'] = $fees;
                  $billing_data_array['icd10'] = $icd10;
            $billing_data_array['codetype'] = $codetype;
              $billing_data_array['modifier'] = $modifier;


          $this->insertCPTBilling($billing_data_array);

Is this a good start?

juggernautsei:
I thought that REPLACE INTO was the smart choice because it takes care of both an insert and an update call. So instead of coding out two different functions to insert and to update a record. The replace into can do both.

mdsupport:
@juggernautsei  Idea of assoc array ($a) is the generic store should not care about specific columns.

```
$fields=array_keys($a);
$values=array_values($a);
$fieldlist=implode(',',$fields); 
$qs=str_repeat("?,",count($fields)-1);
$sql="insert into billing($fieldlist) values(${qs}?)";
sqlStatement($sql, $values);
```

Now the caller has flexibility to pass 1 to max fields in $a as needed by its logic. All missing fields will use database default for initial values. Make table name as a variable and you have a generic inserter for any table.

But as @adunsulag said your original code is just fine per project standards. Didn't mean to waste your time.

juggernautsei:
@mdsupport education is never a waste of time.

juggernautsei:
@mdsupport I see that the sqlStatement method smarter than I know it to be from the past. I will try this out and see what happens. Did I get the associative array correct?

sjpadgett:
Sherwin,
Let PHP parser know the depth of array on init. Also, prevent errors,
warnings and sometimes memory issues.
What if a code description is not passed in? Use the
CodeTypesService->resolveCode() to ensure a validated code with description
is set.

$billing_data_array = array(
    'date' => $event_date ?? null,
    'code' => $code ?? null,
    'pid' => $this->pid ?? null,
    'provider_id' => $provider ?? null,
    'user' => $this->userid ?? null,
    'encounter' => $enc ?? null,
    'code_text' => $text ?? '',
    'fee' => $fees ?? null,
    'code_type' => 'CPT4', // can't really say these code types will
always be ICD10 or CPT4
    'modifier' => $modifier ?? null,
    'justify' => 'ICD10|' . ($icd10 ?? '') . ':'
);

juggernautsei:
@sjpadgett thanks for your comments and I will make sure to add this to my future coding. However, can you enlighten me on why  sqlStatement will not save the entry or what will save the entry?

juggernautsei:
$codeType = new CodeTypesService();
        $this_code_type = $codeType->getCodeTypeForCode($code);  //I can't use this yet because the instance I am working in does not have the latest FHIR code in it. Is that coming in patch 3?

        //Build data array
        $billing_data_array = [
                 'date' => $event_date ?? null,
                 'code' => $code ?? null,
                  'pid' => $this->pid ?? null,
          'provider_id' => $provider ?? null,
                 'user' => $this->userid ?? null,
            'encounter' => $enc ?? null,
            'code_text' => $text ?? null,
                  'fee' => $fees ?? null,
            'code_type' => $this_code_type ?? "CPT4",
             'modifier' => $modifier ?? null,
              'justify' => "ICD10|" . $icd10 ?? null . ":"
              ];

Produces this array:
        array(11) { 
            ["date"]=> string(19) "2021-10-17 00:53:41" 
            ["code"]=> string(5) "90792" 
            ["pid"]=> int(2123)
            ["provider_id"]=> string(1) "1" 
            ["user"]=> string(1) "1" 
            ["encounter"]=> int(5577) 
            ["code_text"]=> string(39) " New patient psychiatric evaluation " 
            ["fee"]=> string(6) "220.00" 
            ["code_type"]=> NULL 
            ["modifier"]=> string(0) "" 
            ["justify"]=> string(12) "ICD10|F31.30" 
        } 

@mdsupport 

        $sql = "INSERT INTO billing($fieldslist) VALUES(${qs}?)";
var_dump($billing_data_array, $fields, $fieldslist, $values)

result

string(81) "date,code,pid,provider_id,user,encounter,code_text,fee,code_type,modifier,justify" 

array(11) { [0]=> string(19) "2021-10-17 01:28:24" [1]=> string(5) "90837" [2]=> int(2123) [3]=> string(1) "1" [4]=> string(1) "1" [5]=> int(5587) [6]=> string(44) " Individual psychotherapy for 60 minutes " [7]=> string(6) "150.00" [8]=> string(4) "CPT4" [9]=> string(0) "" [10]=> string(12) "ICD10|F31.30" }

sjpadgett:
Sherwin I tested this on my dev system and it saves even an empty array so  don't know. I'm PHP8 and Maria 10.4.
No plans to patch FHIR or code types.

mdsupport:
Once again suggesting normal debugging - dump $sql and $values or whatever sqlStatement is going to get. Turn on error logging. Run that dumped sql in mysql with values to make sure it is valid etc.. These old eyes are not meant to be remote php compiler.

Best.

juggernautsei:
I would like to keep this alive. I will have to revisit this code in a couple of weeks. I did make the suggested changes. I figured out why it was not saving. Human issues. So, I will see if this can be brought in after the next release.

PR #4569, opened 2021-08-10
Building appointment event dispatcher

juggernautsei:
Fixes #

PR #4652, opened 2021-09-25
Adding appointment event dispatchers

bradymiller:
Adding appointment event dispatchers

Originated from prior PR:
https://github.com/openemr/openemr/pull/4569


bradymiller:
bringing this into master and rel-600

PR #4654, opened 2021-09-29
ensure always a unique installation id when need it

bradymiller:
ensure always a unique installation id when need it


src/Common/Uuid/UniqueInstallationUuid.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+/**
+ * uniqueInstallationUuid.php class
```
bradymiller, line 4: fix:
```
UniqueInstallationUuid class
```

PR #4659, opened 2021-10-03
bug fix

bradymiller:
another bug fix


templates/documents/general_view.html
```diff
@@ -248,7 +248,7 @@ <h4>{$file->get_name()|text}
                   {xlt t='Date:'} {$note->get_date()|text}
                   {$note->get_note()|text}
                   {if $note->get_owner()}
-                    &nbsp;-{user_info id=$note->get_owner()|text}
```
bradymiller, line 251: The problem with this line was that smarty was html escaping the parameter and not the actual output of the function.

bradymiller:
testing well and bringing in

PR #4661, opened 2021-10-03
QRDA Export and Value sets

### LINKED ISSUE #4662: Tracking external providers!
### Describe the bug
We create them in imports however, don't give any permissions so new imports can be included in charting, reporting etc. Why?
This may be okay for CCD's where imported data is generally viewed but, not so for QRDA where imported data needs to be available for data manipulations and export.

In realities IMO on a CDA type import the responsible party and responsible organization should be create as external user with associated default facility(responsible organization) and the provider flag set.

Anything wrong with me putting users in view to charting by setting the user provider and active flag on external providers?

@bradymiller @sunsetsystems Would appreciate some guidance here.

Also need to add the 2021 and 2022 reporting period CQM value sets to external loads table.

sjpadgett:
- add external loads for the 2021 and 2022 CQM reporting periods

Fixes #4662 

More CQM structural work. QRDA export, import, data loads & report bugs.

### DIFF
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -1061,3 +1061,10 @@ INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_re
 #IfMissingColumn audit_master is_qrda_document
 ALTER TABLE `audit_master` ADD `is_qrda_document` BOOLEAN NULL DEFAULT FALSE;
 #EndIf
+
+#IfNotRow4D supported_external_dataloads load_type CQM_VALUESET load_source NIH_VSAC load_release_date 2020-05-07 load_filename ep_ec_only_cms_20200507.xml.zip
+INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_release_date`, `load_filename`, `load_checksum`) VALUES
+('CQM_VALUESET', 'NIH_VSAC', '2020-05-07', 'ep_ec_only_cms_20200507.xml.zip', '02dc0b497da979e336c24b0b5c6e1ccb');
+INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_release_date`, `load_filename`, `load_checksum`) VALUES
+( 'CQM_VALUESET', 'NIH_VSAC', '2021-05-06', 'ep_ec_eh_cms_20210506.xml.zip', '6455da86e269edb6d33288e72b467373');
+#EndIf
--- a/sql/6_0_0-to-6_1_0_upgrade.sql
+++ b/sql/6_0_0-to-6_1_0_upgrade.sql
@@ -1068,3 +1068,11 @@ INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_re
 INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_release_date`, `load_filename`, `load_checksum`) VALUES
 ( 'CQM_VALUESET', 'NIH_VSAC', '2021-05-06', 'ep_ec_eh_cms_20210506.xml.zip', '6455da86e269edb6d33288e72b467373');
 #EndIf
+
+#IfMissingColumn form_encounter encounter_type_code
+ALTER TABLE `form_encounter` ADD `encounter_type_code` VARCHAR(31) NULL DEFAULT NULL, ADD `encounter_type_description` TEXT NULL;
+#EndIf
+
+#IfMissingColumn users billing_facility
+ALTER TABLE `users` ADD `billing_facility` TEXT NULL, ADD `billing_facility_id` INT(11) NOT NULL DEFAULT '0';
+#EndIf
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -8520,6 +8520,10 @@ INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_re
 ('ICD10', 'CMS', '2020-10-01', 'Code-Descriptions.zip', 'f22e7201fa662689d85b926a32359701');
 INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_release_date`, `load_filename`, `load_checksum`) VALUES
 ('ICD10', 'CMS', '2020-10-01', 'Zip File 5 2021 ICD-10-PCS Order File (Long and Abbreviated Titles).zip', '6a61cee7a8f774e23412ca1330980bbb');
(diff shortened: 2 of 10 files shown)


sjpadgett:
Added default billing facility to users. I will mod encounter form before next release or soon after @matrix-amiel fixes existing bug or fixes in master and I'll port back to v6 patch 3.

sql/6_0_0-to-6_1_0_upgrade.sql
```diff
@@ -1068,3 +1068,11 @@ INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_re
 INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_release_date`, `load_filename`, `load_checksum`) VALUES
 ( 'CQM_VALUESET', 'NIH_VSAC', '2021-05-06', 'ep_ec_eh_cms_20210506.xml.zip', '6455da86e269edb6d33288e72b467373');
 #EndIf
+
+#IfMissingColumn form_encounter encounter_type_code
+ALTER TABLE `form_encounter` ADD `encounter_type_code` VARCHAR(31) NULL DEFAULT NULL, ADD `encounter_type_description` TEXT NULL;
+#EndIf
+
+#IfMissingColumn users billing_facility
+ALTER TABLE `users` ADD `billing_facility` TEXT NULL, ADD `billing_facility_id` INT(11) NOT NULL DEFAULT '0';
```
stephenwaite, line 1077: You prob have this but don't forget this in `database.sql`

sjpadgett: Thanks Steven, I did forget! will get next push. y'all keep an eye on me....

sjpadgett:
Our standard services have become so tightly tied to UUID they are difficult to use in core.
Need more helper methods in BaseService to resolve UUID's.

src/Services/EncounterService.php
```diff
@@ -201,6 +202,8 @@ class.notes as class_title,
                                class_code,
                                facility_id,
                                discharge_disposition,
+                               encounter_type_code,
+                               encounter_type_description,
```
sjpadgett, line 206: @adunsulag check me here.

sql/database.sql
```diff
@@ -1831,6 +1831,8 @@ CREATE TABLE `form_encounter` (
   `shift` varchar(31) NOT NULL DEFAULT '',
   `voucher_number` varchar(255) NOT NULL DEFAULT '' COMMENT 'also called referral number',
   `discharge_disposition` varchar(100) NULL DEFAULT NULL,
+  `encounter_type_code` VARCHAR(31) NULL DEFAULT NULL COMMENT 'not all types are categories',
+  `encounter_type_description` TEXT NULL,
```
sjpadgett, line 1835: @kchapple 
Note that I've added the encounter type code where they will be needed when populating patient models. Cypress relies heavy on these codes as they don't populate diagnosis in template that I've seen. However, QRDA IG allows but I have covered in import where diagnosis becomes an attached issue/problem to encounter.

sjpadgett:
Procedure codes for imports are all over the place as seen below. Supposedly Cypress can be configured for coding system preference but, I've been unable to get to work. Still though, QRDA IG doesn't limit except where a translated code type is available so, that'd be my next parsing mod to test the template for specific code translations e.g a code type snomed-pr may have translation sub template for cpt4 or visa versa. Also be nice if cypress sent code display name like everyone else. For CQM purposes though, the code is important and key to calculations.
But for patient models we can use the installed value set where coding system oid can be retrieved from CodeTypesService for valueset lookup.

However I look at it, we need to do some work on handling codes throughout core and get some consistency.

`
<html><body>


  | Edit | Copy | Delete | 8 | 1 | SNOMED-CT:307521008 | Intravenous infusion of thrombolytic | 1 |   | 0 | procedure | NULL | NULL
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  | Edit | Copy | Delete | 27 | 1 | SNOMED-CT:302497006 | Hemodialysis procedure | 1 |   | 0 | procedure | NULL | NULL
  | Edit | Copy | Delete | 28 | 1 | CPT4:90945 |   | 1 |   | 0 | procedure | NULL | NULL
  | Edit | Copy | Delete | 29 | 1 | CPT4:36820 |   | 1 |   | 0 | procedure | NULL | NULL
  | Edit | Copy | Delete | 30 | 1 | SNOMED-CT:302497006 | Hemodialysis procedure | 1 |   | 0 | procedure | NULL | NULL
  | Edit | Copy | Delete | 31 | 1 | SNOMED-CT:108241001 | Dialysis procedure | 1 |   | 0 | procedure | NULL | NULL
  | Edit | Copy | Delete | 32 | 1 | HCPCS:S2065 |   | 1 |   | 0 | procedure | NULL | NULL


</body>
</html>
`

src/Services/EncounterService.php
adunsulag: @sjpadgett You'll need to add encounter_type_code & encounter_type_description to the outer select on line 157 if you want to actually see this data.  Because of the way searching is done and how we have duplicate columns for 'pid' on the joins on everything we have the multi-select.  This is where I wish we and an ORM.

bradymiller:
2DV

bradymiller:
@sjpadgett , you ok with me doing a quick commit to change the `TEXT NULL` stuff to `TEXT` (i missed this in my review and can adjust this in a couple seconds)?

sjpadgett:
nope, I keep forgetting!

bradymiller:
@sjpadgett , Also had a quick question on the Elvis Operator (ie `?:`) versus Null Coalescing Operator (ie. `::`). When I have used the Elvis Operator in the past I noticed that it still throws the PHP8 warnings (when variable is not yet set) where the Null Coalescing Operator does not. Is this something you are noting also? And if so, what is best way to get rid if the PHP8 warning when using Elvis?

sjpadgett:
@bradymiller I went back and forth on the ternary versus coalescing. You're prob right they should be addressed which I'll do for my next PR.
Matter of fact, I think I was going to go back and add a coalesce to the right argument condition e.g
$a ?: ($b ?? null); or
$a ?? $b may be acceptable to squash warnings which I always have turned off.:)

Your choice however way I will address next PR.

bradymiller:
hi @sjpadgett ,
I'd just go with `$a ?? $b` when possible
unless need to use elvis in the use case, then would do `$a ?: ($b ?? null)`

bradymiller:
actually, i think it needs to be `($a ?? null) ?: $b`
(definitely test to ensure the php8 warnings go away for the elvis workaround)

stephenwaite:
#4641

PR #4666, opened 2021-10-07
Appointment api enhancements.

adunsulag:
Added the patient pid & uuid to the appointment so people can get an
exact match on who the appointment is for.

Added the provider's id & uuid as part of the pc_aid fields so people
can know the exact provider connected with the appointment from the
standard api.

Added validation on the provider and patient ids for the appointment
create.


bradymiller:
2DV

PR #4667, opened 2021-10-08
CCM and QRDA Import

sjpadgett:
- new class and CCM import refactor
- some import fixes and changes

Fixes #

I'm gradually refactoring CCM import into separate classes so eventually can create another controller class to use CDA imports elsewhere other than CCM.


sjpadgett:
This import has three different code systems for encounters and all code types are turned off in lists. A worry gone and just need to mod CCDA generator to use this need class.

image

PR #4670, opened 2021-10-10
allow picking any type of code for codes in list editor

bradymiller:
early christmas present for @sjpadgett :)

(allows picking code of any codetype to populate Code in the list editor)


sjpadgett:
Thank you sir!
I'm close to not needing to have snomeds, cpt4, loinc etc installed and just using our value sets.
I'm going to add valueset as external table to dynamic code picker or maybe you could add while here.
I can send you the table to import or I think I sent you the valueset csv to import via external loads.

sjpadgett:
I've tested this okay

PR #4677, opened 2021-10-13
Enhance user service to return uuids

adunsulag:
Small change to make it so the user service returns uuids as part of its record retrieval on a number of methods.  Chose the least impact methods so we didn't have to deal with a ton of refactoring.


bradymiller:
signal-2020-11-19-220815

PR #4679, opened 2021-10-13
Fixes #4658 change uuid mapping to be on save.

### LINKED ISSUE #4658: UuidMapping registry for vital records creates extremely slow API requests
Working with 100K of patients the UUID mapping for the vitals service (where a uuid is mapped to each individual column for each vital forms) is creating API responses that vary between 45 seconds to 1.5 minutes.

So it executes the following query for each of the 19 vitals observation codes which is what is creating the long delay.  I'm looking at what can be done to optimize the query.
```
SELECT `form_vitals`.`uuid`
                       FROM `form_vitals`
                       LEFT OUTER JOIN `uuid_mapping` ON `form_vitals`.`uuid` = `uuid_mapping`.`target_uuid` AND `uuid_mapping`.`resource` = 'Observation'  AND `resource_path` = 'category=vital-signs&code=3150-0' WHERE (`uuid_mapping`.`uuid` IS NULL OR `uuid_mapping`.`uuid` = '' OR `uuid_mapping`.`uuid` = '\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0')
                       AND (`form_vitals`.`uuid` IS NOT NULL AND `form_vitals`.`uuid` != '' AND `form_vitals`.`uuid` != ''\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0')
                       LIMIT 1000
```

adunsulag:
Closes #4658 
Created a FHIR module that listens to the vitals, patient, and social history save events and populates the uuid mapping tables whenever those records are saved.  This removes the populate uuid mechanisms from being triggered on every single api requests and instead the uuids are populated at the time of save.

Added events to the social history service that triggers whenever a history data record is saved.

We didn't go with the database function / trigger implementation for two reasons.  1) Database portability between MySQL and MariaDB and 2) the UUID mechanism requires us following the v1 UUID which is time based so we can have better indexing of our uuids.  The default UUID mechanism in both MySQL/MariaDB uses UUIDv4 and so we'd have to roll our own UUIDv1 implementation.  

@bradymiller If you get a chance to review this so we can push out a new docker image that'd be great.  Thanks.


bradymiller:
hi @adunsulag ,
Code looks good and assuming testing well for you, so bringing this in (then can expedite the docker build tonight). Note that an issue that needs to be addressed is:
image

(Although FHIR is supposed to be already working by default it is still showing up as unregistered. Sounds like this is just a appearance thing, which can fix by doing what we did for the Care Coordination module to have the shown has installed/enabled by default; see here https://github.com/openemr/openemr/pull/4575/files )

bradymiller:
Or better yet on the above FHIR module, does it make more sense to not even show it since it is a core module (like the other core modules) and it would just be confusing since it can not be disabled?

bradymiller:
@adunsulag , after a mere 1.7 hours of building, the 6.1.0 (ie. dev) dockers are updated and on dockerhub :)
https://hub.docker.com/r/openemr/openemr/tags?page=1&ordering=last_updated

PR #4689, opened 2021-10-18
More CQM Work

sjpadgett:
- minor fix portal document refresh
- modify find_code_popup.php to honor valueset codes if installed
- change various encounter forms(care plan etc) to allow any code types.
- fix escaping error code finder

### DIFF
--- a/interface/forms/care_plan/new.php
+++ b/interface/forms/care_plan/new.php
@@ -113,7 +113,7 @@ endforeach;
                     checkId = 1;
                 }
                 document.getElementById('clickId').value = checkId;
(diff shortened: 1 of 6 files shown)


sjpadgett:
Couple biggies, Assessment Performed observation however, will be used for other observations like mental health(PHQ,GAD), social, and various assessment tests.
Also planned activities(care plan intervention for one)

Observation imports work to DB but still need to work on the encounter form. However, this will allow continuing on with CQM models and calculations.
localhost-127-0-0-1-openemr-form_observation-phpMyAdmin-5-1-1

stephenwaite:
#4641

PR #4697, opened 2021-10-27
Twigify the About Page

robertdown:
Twigify the About Page, add several new Branding globals to improve white labeling

* New Globals section: `Branding` now appears below Appearance in list, several items moved into that section
* New Globals options:
    * Display Main Menu Logo - Hide the top left branding logo (defaults to true)
    * User Manual Link - Override the user manual link (was previously hard-coded) (defaults to https://www.open-emr.org/wiki/index.php)
    * Display links to acknowledgements page (defaults to true)
    * Display the Review link on the about page (defaults to true)
    * Display the donations link on the about page (defaults to true)
* Removed "Display Acknowledgements Link on Login Page" in favor of more abstract global described above

If there's no online support link, user manual link, or support phone number those blocks are hidden from the about page.

Also added a function in the VersionService to return a fully compounded version string (i.e. 6.1.0-dev)

image

image


robertdown:
Ok @bradymiller I resolved the issues and created a check so if the online manual is left blank on globals it will auto-generate the correct url and inject that, otherwise it'll use the globals setting. I also added a new Twig global called `assetVersion` which enables the `$v_js_includes` variable for Twig use (this global is run through attr() in the TwigExtension function and does **not** need to be escaped in template files.

TwigExtension now requires a GlobalsService variable on construct to leverage the $GLOBALS variable

library/globals.inc.php
```diff
@@ -464,9 +464,10 @@ function gblTimeZones()
         ),
 
         'user_manual_link' => [
-            xl('User Manual Link'),
+            xl('User Manual Link Override'),
             'text',
-            'https://www.open-emr.org/wiki/index.php'
+            '',
+            xl("Point to a custom user manual. Leave blank for the default, aut0-generated URL for specific version of application"),
```
bradymiller, line 470: just a minor mispell here with the zero in aut0

bradymiller:
I'd prob call it assetVersionEscaped (or some other memory aid in the function name that it is escaped) or else I will likely forget this in a month or so and recommend escaping it in my reviews
(note it wouldn't be the end of the world though since double escaping number won't break anything :) )

PR #4714, opened 2021-11-07
Created calendar extension points for modules.

adunsulag:
Made it so module writers can add javascript and stylesheet files to the
calendar from their module using the script filter and syle filter
event.

Added extension points to the calendar for modifying the event data
structure that comes back for the day/week/month format.  Users can
override the style on each calendar appointment.  Also added the
appointment id and pid information to each calendar event so frontend
javascript work can be done on the calendar appointments.

Added some security method to the modules application class to make sure
module authors don't include files outside the modules directory.

### DIFF
--- a/interface/main/calendar/modules/PostCalendar/pnadmin.php
+++ b/interface/main/calendar/modules/PostCalendar/pnadmin.php
@@ -32,8 +32,6 @@ use OpenEMR\Common\Acl\AclExtended;
 use OpenEMR\Core\Header;
 use OpenEMR\Events\Core\ScriptFilterEvent;
 use OpenEMR\Events\Core\StyleFilterEvent;
-use Symfony\Component\EventDispatcher\GenericEvent;
-use OpenEMR\Core\ModulesApplication;
 
 //=========================================================================
 //  Load the API Functions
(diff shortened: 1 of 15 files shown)


interface/main/calendar/modules/PostCalendar/pnadmin.php
```diff
@@ -30,6 +30,10 @@
 use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\Common\Acl\AclExtended;
 use OpenEMR\Core\Header;
+use OpenEMR\Events\Core\ScriptFilterEvent;
+use OpenEMR\Events\Core\StyleFilterEvent;
+use Symfony\Component\EventDispatcher\GenericEvent;
+use OpenEMR\Core\ModulesApplication;
```
bradymiller, line 36: would order above alphabetically (just the line above needs reordering)

interface/main/calendar/modules/PostCalendar/pnuserapi.php
```diff
@@ -57,6 +59,8 @@
 unset($pcModInfo, $pcDir);
 
 use OpenEMR\Events\Appointments\CalendarFilterEvent;
+use Symfony\Component\EventDispatcher\GenericEvent;
+use OpenEMR\Events\Appointments\CalendarUserGetEventsFilter;
```
bradymiller, line 63: Would it make sense to combine these 2 separate use blocks?

bradymiller:
hi @adunsulag , Looks great. Only noted 2 very minimal issue.

sjpadgett:
Nice stuff @adunsulag I can't wait for our next release.

PR #4716, opened 2021-11-07
Move patient date methods into service

adunsulag:
Needed to leverage a number of patient date formatting methods which
were hard to get to w/ them being in the library.  Moved them into the
patient service method so we could access them from the service.


bradymiller:
hi @adunsulag , Looks great.

PR #4718, opened 2021-11-07
Make fhir search service classes easier to use

adunsulag:
Made it so the fhir search classes will convert strings and numbers into
an exact string match.  This makes it so that search methods can pass in
simple array values without having to create the search field objects each time.  This
was a request from @sjpadgett to make the classes a little easier to
use.

### DIFF
--- a/src/Services/Search/FhirSearchWhereClauseBuilder.php
+++ b/src/Services/Search/FhirSearchWhereClauseBuilder.php
@@ -34,7 +34,12 @@ class FhirSearchWhereClauseBuilder
             foreach ($search as $key => $field) {
                 if (!$field instanceof ISearchField) {
                     // developer logic error
-                    throw new \BadMethodCallException("Method called with invalid parameter.  Expected SearchField object for parameter '" . $key . "'");
+                    // treat the field as an exact string match if they send us a primitive
+                    if (is_string($field) || is_numeric($field)) {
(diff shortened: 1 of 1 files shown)


bradymiller:
hi @adunsulag . looks good.

sjpadgett:
lol! So @adunsulag, is it really that easy. I simply request or suggest something and it magically appears!
Let me think umm......

src/Services/Search/FhirSearchWhereClauseBuilder.php
```diff
@@ -34,7 +34,12 @@ public static function build($search, $isAndCondition = true): SearchQueryFragme
             foreach ($search as $key => $field) {
                 if (!$field instanceof ISearchField) {
                     // developer logic error
-                    throw new \BadMethodCallException("Method called with invalid parameter.  Expected SearchField object for parameter '" . $key . "'");
+                    // treat the field as an exact string match if they send us a primitive
+                    if (is_string($field) || is_numeric($field)) {
+                        $field = new StringSearchField($key, $field, SearchModifier::EXACT);
+                    } else {
+                        throw new \BadMethodCallException("Method called with invalid parameter.  Expected SearchField object for parameter '" . $key . "'");
+                    }
```
sjpadgett, line 42: Expected SearchField object for parameter

In this case an object is not expected so dev may start scratching their head.

sjpadgett (approved):

PR #4719, opened 2021-11-07
Encounter Appointment enhancements

adunsulag:
Added the user and group data elements to the EncounterService so we can
specify what user and group is creating the encounter form.

Changed the rest API to use the user and group from the session for the encounter.

Enhanced the appointment service to return an address as a string.  Useful when displaying addresses in a html **pre** tag or w/ the **address** tag.

Added a number of helper methods to AppointmentService
getCalendarCategories()
isCheckInStatus()
isCheckOutStatus()
updateAppointmentStatus()
getEncounterForAppointment()
createEncounterForAppointment()

The last two methods will go off the current date of Y-m-d which is not super precise but we have no other linking mechanism right now and we currently do that in the auto-create appointment piece in the appointments service.

Added to the EncountersService
getEncountersForPatientByPid()
getPatientEncounterListWithCategories()

The getPatientEncounterListWithCategories method returns the needed data format that is used by the frontend knockout.js to populate the encounter list dropdown.  This is useful in any kind of frontend rendering / AJAX application for updating that list (using left_nav.setPatientEncounter)

Fixed the getUser method in UserService to return the stringified uuid.


library/patient_tracker.inc.php
```diff
@@ -141,27 +143,15 @@ function fetch_Patient_Tracker_Events($from_date, $to_date, $provider_id = null,
...
-
-    $row = sqlQuery("SELECT toggle_setting_2 FROM list_options WHERE " .
-    "list_id = 'apptstat' AND option_id = ? AND activity = 1", array($option));
-    if (empty($row['toggle_setting_2'])) {
-        return(false);
-    }
-
-    return(true);
+    $service = new AppointmentService();
+    return $service->isCheckInStatus($option);
```
bradymiller, line 154: should this be isCheckOutStatus?

src/Services/AppointmentService.php
```diff
@@ -199,4 +242,125 @@ protected function createResultRecordFromDatabaseResult($row)
...
+
+        return(true);
+    }
+
+    /**
+     * Updates the status for an appointment.  TODO: should be refactored at some point to update the entire record
+     * @param $eid The id of the appointment event
+     * @param $status The status the appointment event should be set to.
+     */
+    public function updateAppointmentStatus($eid, $status)
```
bradymiller, line 294: Check out the manage_tracker_status() function. That mechanism ensure that status changes are logged which is used in reporting mechanism (and any other data analytics a practice may want to do).

library/patient_tracker.inc.php
adunsulag: Great catch.  That definitely should be fixed.

adunsulag:
@bradymiller or @sjpadgett I did some refactoring in order to incorporate brady's suggestion of having appt status updates update the patient track.  I had to extract a bunch of methods from the patient tracker lib file into its own service class to make that happen.  So doing another review of this I think would be wise.  Thanks.

sjpadgett:
@adunsulag 
I very much like where you are going with this but will leave with Brady to look into. I do know tracker can be tricky on reporting side.

bradymiller:
@adunsulag , very nice stuff!

Looks like flow board is broken (nothing is showing in the php error log, but browser is getting 500 error and stuck on the spinner):
image

Also, is there a way to have this new module (and the FHIR module) not show up on this screen (since they are installed by default):
image

sjpadgett:
Most likely in the sql where exception isn't percolating

adunsulag:
@bradymiller Ouch!  I'll look into it.

adunsulag:
@bradymiller To skip the modules not showing up I'll have to add the modules as being installed to the database via the upgrade script. I'll go ahead and do that as part of the fixes with the flow board.

bradymiller:
Do the Installer and Acl modules need to be added like that also? I was hoping there was just a filter somewhere that would completely ignore these autoinstalled modules in the Installer gui.

adunsulag:
@bradymiller So found the missing method that was blowing up the flowboard.  It should have thrown a fatal error and be logged in PHP. No idea why that didn't happen.  Also, I should have seen the error too :)  I think trying to break everything up into separate pull requests messed me up haha.  Oh well, I was trying to make it easier to review for you on a piece by piece basis.

I also found (once again) where those modules are being filtered out and I added in code to filter out the FHIR and PatientFlowBoard module.  Unfortunately right now there are two places you have to edit for the core modules, I'm going to have to figure out how to centralized that code but it appears I've forgotten a bit about the zend framework so I'll investigate that another time.

bradymiller:
@adunsulag , flow board related stuff is testing well, but getting white screen in the encounter which is giving a error in php log (looks like will be a simple fix):

image

php error:
```
[Wed Nov 10 02:19:36.130627 2021] [php:error] [pid 908] [client 172.27.0.1:34870] PHP Fatal error:  Uncaught Error: Class "OpenEMR\\Services\\FormService" not found in /var/www/localhost/htdocs/openemr/src/Common/Session/EncounterSessionUtil.php:32\nStack trace:\n#0 /var/www/localhost/htdocs/openemr/library/encounter.inc(10): OpenEMR\\Common\\Session\\EncounterSessionUtil::setEncounter()\n#1 /var/www/localhost/htdocs/openemr/interface/forms/newpatient/save.php(159): setencounter()\n#2 {main}\n  thrown in /var/www/localhost/htdocs/openemr/src/Common/Session/EncounterSessionUtil.php on line 32, referer: http://localhost:8300/interface/forms/newpatient/new.php?autoloaded=1&calenc=
```

adunsulag:
@bradymiller I think if I rebase this it will fix that problem.  That was an error from a different pull request and I already committed it to master.  I'll go ahead and rebase and verify it resolves the issue.

bradymiller:
If it goes away, then I think it's ready to come in :)

2DV

PR #4724, opened 2021-11-09
Openemr fix 4723 module install script uses OpenEMR sql upgrade service

### LINKED ISSUE #4723: OpenEMR Modules install script fails if string contains semicolon
If you have a table.sql or install.sql in a custom module and the sql statement has a semi-colon inside a string then the module installer will fail.  The installer script dies due to it attempting to explode lines by semicolons in the InstModuleTable->installSQL() method.  This obviously doesn't work if you have a string with a semicolon in.  Here is an example statement:

```sql
INSERT INTO `openemr_postcalendar_categories` (
    `pc_constant_id`, `pc_catname`, `pc_catcolor`, `pc_catdesc`,
   `pc_recurrtype`, `pc_enddate`, `pc_recurrspec`, `pc_recurrfreq`, `pc_duration`,
   `pc_end_date_flag`, `pc_end_date_type`, `pc_end_date_freq`, `pc_end_all_day`,
   `pc_dailylimit`, `pc_cattype`, `pc_active`, `pc_seq`, `aco_spec`
)
VALUES (
    'telehealth_new_patient', 'Telehealth New Patient', '#a2d9e2'
    , 'New Patient Telehealth appointments', '0', NULL
    , 'a:5:{s:17:"event_repeat_freq";s:1:"0";s:22:"event_repeat_freq_type";s:1:"0";s:19:"event_repeat_on_num";s:1:"1";s:19:"event_repeat_on_day";s:1:"0";s:20:"event_repeat_on_freq";s:1:"0";}'
    , '0', '1800', '0', NULL, '0', '0', '0', 0, '1', '10', 'encounters|notes'
);
```

My suggestion is to refactor the installSQL method to use the same code logic in the sql_upgrade.php file rather than trying to write / fix our own parser.  However, one downside of this approach is that it assumes that you will use a single statement per line.  If module authors have been writing modules with multiple sql statements per line this will break.  I noticed that @yurige-matrix's example PatientFilterEventHookTest has the install script all on one line.  My proposed fix in the pr I will put up shortly will leave the zend modules using the old parsing mechanism until I hear back from Matrix on how we can support their modules.

adunsulag:
Fix module semicolon install bug for custom module
    
Closes #4723 . I migrate the logic in the sql_upgrade_fx file into a
SQLUpgradeService object that can be reused in other places.  I change
up the InstallerController for both the upgrade sql and the install sql
process to use the SQLUpgradeService.

I also fix a bug in the uninstall custom module that was throwing an
error due to a recent laminas pdo change with the result objects.  This
prevented the screen from refreshing when a custom module was
uninstalled.

I left in place the sql_upgrade_fx script which calls into the
SQLUpgradeService just to mitigate any potential code problems.
Eventually that file can be removed.


adunsulag:
@matrix-amiel @yurige-matrix you should not be impacted right now, but I think you will want to review this to double check.  Also, I would like to change this code in the future which will impact your laminas modules.  I would like to have the install workflow use the same upgradeFromSqlFile() function as is currently used for version updates.  It makes the code more consistent in my opinion and makes it so any sql file parsing we do has the same logic across the entire system.  If you can look over this code and let me know if you see any problems that would be great.

bradymiller (approved):

bradymiller:
@adunsulag , nice stuff! sql upgrade testing well. After you bring this in, I will then enjoy zapping the sql_upgrade_fx script from the codebase :)

adunsulag:
@bradymiller Should we give Matrix time to review this?  I tested it well and you tested it also, long term I think it could impact them but not sure it should hold up bringing this in.

bradymiller:
Agree shouldn't impact them short/medium term, so would bring it in to keep things moving forward :)

bradymiller:
2DV-2

PR #4726, opened 2021-11-10
Fix #4725 user events, uuid generation

### LINKED ISSUE #4725: UserCreatedEvent fired on error with no user created
If you attempt to create a user with too short of a password or some other data validation error the system still fires the UserCreatedEvent.  The event should only be fired if the user is successfully created in the system.  Going to submit a pr that fixes this issue.

adunsulag:
Fixes #4725 
Made it so the create user event only fires if the user successfully
saves.  Also made it so creating a user generates a uuid for it so we
have it at the time of the save user event.  Added some helper methods
to the user created and user updated event to make them more useful.


bradymiller (approved):

bradymiller:
hi @adunsulag , looks great (ready when styling fixed)!

signal-2020-11-19-220815

PR #4728, opened 2021-11-10
Openemr render events

adunsulag:
Added events to OpenEMR main tabs screen for outputting html at the end of the body tag and at the start of the body tag.
Added event to filter appointments in the portal home page.
Added event to render output at the end of the portal sections and before the scripts of the portal.
Refactored the appointments into a separate twig file that module writers can overwrite.

### DIFF
--- a/templates/portal/appointment-item.html.twig
+++ b/templates/portal/appointment-item.html.twig
@@ -1,0 +1,15 @@
+<div class="card p-2">
+    <div class="card-header">
+        <a href="#" onclick="editAppointment({{ appt.mode | attr_js }},{{ appt.pc_eid | attr_js }})" title="{{ appt.etitle | attr }}">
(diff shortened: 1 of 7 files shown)


adunsulag:
@sjpadgett Here are the portal events I use for the telehealth stuff.  I also add a single javascript file at the end of the main tabs body using these events which controls all of my telehealth messaging as well as displaying / hiding of the telehealth video screen .

sjpadgett (approved):
Should be okay. 
I immediately start thinking of firing notification events for chat and secure messaging..
Thanks Stephen. This will be well used I'm sure.

templates/portal/appointment-item.html.twig
```diff
@@ -0,0 +1,15 @@
+<div class="card p-2">
+    <div class="card-header">
+        <a href="#" onclick="editAppointment({{ appt.mode | attr_js }},{{ appt.pc_eid | attr_js }})" title="{{ appt.etitle | attr }}">
```
bradymiller, line 3: 2DV

bradymiller (approved):

PR #4730, opened 2021-11-10
Adding support for Sphere payment processing

bradymiller:
This supports (Sphere tool that was integrated is in parenthesis):
1. Payment capture from front office (Trustee Premier for MOTO)
2. Payment capture from patient portal (Trustee Premier for Ecomm)
3. void/credit of payments (Trustee API)

Main advantage of this approach is that all the sensitive info (cc number) is captured on Sphere's web server and not done on OpenEMR's server. Note there are 3 different Sphere tools that are utilizing, which are listed above, and is why there are several global credentials required for this. Note the front office uses a card reader (or the machine can enter in the cc number); btw, the machines for front office basically leverage native keyboard drivers so super easy to work (basically just focus the field on web page and then read card or enter data into machine).

Also added a generic mechanism to support encrypted hash in globals since needed a pin mechanism to ensure void/credit is authorized for user within OpenEMR itself. Using this for the `Sphere Void/Credit Confirmation PIN` global.

I considered a module (which is why the code is operating like a module by inserting the code via one liners), but was really awkward because of the current payment processing options (am also hoping this becomes the main payment processing that is used since the OpenEMR foundation will actually get funding when this payment system is used).

Pertinent globals:
- Connectors->Accept Credit Card transactions from Front Payments
- (added Sphere option) Connectors->Select Credit Card Payment Gateway->Gateway for Sphere Payments
- Connectors->Set Gateway to Production Mode (don't turn this on until testing mode is working)
- (new) Connectors->Connectors->Sphere Clinicfront (MOTO) Transaction CustID
- (new) Connectors->Sphere Clinicfront (MOTO) Transaction CustID License Key
- (new) Connectors->Sphere MOTO TC Link Password
- (new) Connectors->Sphere Patientfront (Ecomm) Transaction CustID
- (new) Connectors->Sphere Patientfront (Ecomm) Transaction CustID License Key
- (new) Connectors->Sphere Ecomm TC Link Password
- (new) Connectors->Sphere Void/Credit Confirmation PIN
- Portal->Allow Online Payments

Some things that also need to ensure for this to work:
- Need to be using https
- The global Connectors->Site Address Override needs to be correct

Posted screenshots below for all 3 workflows.

---

# Screenshots

## Screenshots of the front office

image

image

image

image

image

---

## Screenshots for patient portal

image

image

image

image

---

## Screenshots for void/credit

image

image

image

image

image

---

### DIFF
--- a/sphere/initial_response.php
+++ b/sphere/initial_response.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * initial_response.php
+ *
+ * Special script to allow callback from Sphere to avoid cross origin breakage.
+ * Csrf security is maintained.
+ * Call to top.restoreSession() happens to ensure directed to correct session.
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2021 Brady Miller <brady.g.miller@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+if (!empty($_GET['revert'])) {
+    // processing a void or credit (will go to process_revert_response.php)
+    $cancel = '';
+    $front = $_GET['front'];
+    $csrf = $_GET['csrf_token'];
+    $transaction = $_GET;
+    $transaction['querystring'] = $_SERVER['QUERY_STRING'];
+    $revert = 1;
+} elseif (!empty($_GET['cancel']) && ($_GET['cancel'] == 'cancel')) {
+    // user cancelled the payment (will go to process_response.php)
+    $cancel = 'cancel';
+    $front = $_GET['front'];
+    $patientIdCc = $_GET['patient_id_cc'];
+    $csrf = $_GET['csrf_token'];
+    $transaction['ticket'] = $_GET['ticket'];
+    $revert = 0;
+} elseif (!empty($_GET['transaction'])) {
+    // user submitted the payment (will go to process_response.php)
+    $cancel = '';
+    $transaction = $_GET['transaction'];
+    $customFields = json_decode($transaction['transactioncustomfield'], true);
+    $front = $customFields[1];
+    $patientIdCc = $customFields[2];
+    $csrf = $customFields[3];
+    unset($transaction['transactioncustomfield']);
+    $revert = 0;
+} else {
+    exit;
+}
+?>
+
+<html>
+    <head>
+    </head>
+    <body>
+        <?php if ($revert === 1) { ?>
+            <form id="myForm" method="post" onsubmit="return top.restoreSession()" action="process_revert_response.php?front=<?php echo htmlspecialchars(urlencode($front), ENT_QUOTES); ?>&csrf_token=<?php echo htmlspecialchars(urlencode($csrf), ENT_QUOTES); ?>">
+        <?php } else { ?>
+            <form id="myForm" method="post" onsubmit="return top.restoreSession()" action="process_response.php?front=<?php echo htmlspecialchars(urlencode($front), ENT_QUOTES); ?>&cancel=<?php echo htmlspecialchars(urlencode($cancel), ENT_QUOTES); ?>&patient_id_cc=<?php echo htmlspecialchars(urlencode($patientIdCc), ENT_QUOTES); ?>&csrf_token=<?php echo htmlspecialchars(urlencode($csrf), ENT_QUOTES); ?>">
+        <?php } ?>
+        <?php
+        foreach ($transaction as $a => $b) {
+            echo '<input type="hidden" name="' . htmlspecialchars($a, ENT_QUOTES) . '" value="' . htmlspecialchars($b, ENT_QUOTES) . '">';
+        }
+        ?>
+        </form>
+        <script>
+            document.getElementById('myForm').submit();
+        </script>
+    </body>
+</html>
--- a/sphere/process_response.php
+++ b/sphere/process_response.php
@@ -1,0 +1,86 @@
+<?php
+
+/**
+ * process_response.php
+ *
+ * Receives the middleman (initial_response.php) script from Sphere to avoid cross origin breakage.
+ * Csrf prevention is maintained.
+ * Works in both core and portal.
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2021 Brady Miller <brady.g.miller@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+// Will start the (patient) portal OpenEMR session/cookie.
+require_once(__DIR__ . "/../src/Common/Session/SessionUtil.php");
+OpenEMR\Common\Session\SessionUtil::portalSessionStart();
+
+$isPortal = false;
+if (isset($_SESSION['pid']) && isset($_SESSION['patient_portal_onsite_two'])) {
+    $pid = $_SESSION['pid'];
+    $ignoreAuth_onsite_portal = true;
+    $isPortal = true;
+    require_once(__DIR__ . "/../interface/globals.php");
+} else {
+    OpenEMR\Common\Session\SessionUtil::portalSessionCookieDestroy();
+    $ignoreAuth = false;
+    require_once(__DIR__ . "/../interface/globals.php");
+}
+
+use OpenEMR\Common\Crypto\CryptoGen;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Core\Header;
+use OpenEMR\PaymentProcessing\PaymentProcessing;
+
+if (!CsrfUtils::verifyCsrfToken($_GET["csrf_token"], 'sphere')) {
+    CsrfUtils::csrfNotVerified();
+}
+
+if ($GLOBALS['payment_gateway'] != 'Sphere') {
+    die(xlt("Feature not activated"));
+}
+?>
+
+<html>
+<head>
+    <?php
+    Header::setupHeader('opener');
+
+    $auditData = ['get' => $_GET, 'post' => $_POST];
+
+    if ($_GET['front'] == 'patient') {
+        // will not show error descriptions in patient front
+        $description = '';
+    } else { // $_GET['front'] == 'clinic'
+        $description = (!empty($_POST['description'])) ? ' (' . $_POST['description'] . ')' : '';
+    }
+
+    if (!empty($_GET['cancel']) && ($_GET['cancel'] == 'cancel')) {
+        PaymentProcessing::saveAudit('sphere', $_GET['patient_id_cc'], 0, $auditData, $_POST['ticket']);
+        echo "<script>opener.sphereNotSuccess(" . xlj("Transaction Cancelled") . ");dlgclose();</script>";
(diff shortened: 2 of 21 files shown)


sphere/process_response.php
```diff
@@ -0,0 +1,83 @@
...
+<html>
+<head>
+    <?php
+    Header::setupHeader('opener');
+
+    $auditData = ['get'=>$_GET, 'post'=>$_POST];
+
+    if ($_GET['front'] == 'patient') {
+        // in future, will always make this blank in patient front since should not be displayed in patient front
+        $description = (!empty($_POST['description'])) ? ' (' .  $_POST['description'] . ')' : '';
```
bradymiller, line 53: oops, forgot to do this

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+     * @return string
+     */
+    private function renderSphereJsPatientFront(): string
+    {
+        return "
+            function sphereSuccess(encData) {
+                let oForm = document.forms['payment-form'];
+                oForm.elements['mode'].value = 'Sphere';
+
+                let inv_values = JSON.stringify(getFormObj('invoiceForm'));
```
adunsulag, line 140: Do you care if someone inserts a javascript value that can't be stringified here?  I usually put these in try catch blocks.

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+                    method: 'POST',
+                    body: new FormData(oForm)
+                }).then(function(response) {
+                    if (!response.ok) {
+                        throw Error(response.statusText);
+                    }
+                    return response.text();
+                }).then(function(data) {
+                    if(data !== 'ok') {
+                        alert(data);
```
adunsulag, line 160: I'd change this to be
```
throw data;
```
so it goes to the same catch block.

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+                        return;
+                    }
+                    alert(chargeMsg);
+                    window.location.reload(false);
+                }).catch(function(error) {
+                    alert(error)
```
adunsulag, line 166: Do you not want to do a reload here? in the above then() clause if you hit an error you still reload the page.

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * Return clinic front specific sphere code in javascript block.
+     *
+     * @return string
+     */
+    private function renderSphereJsClinicFront(): string
+    {
+        return "
+            function sphereSuccess(authCode, transId) {
+                $('#check_number').val(transId);
```
adunsulag, line 181: document.getElementById would be better here.

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+                alert(" . xlj('Success') . " + '\\n' + 'Auth: ' + authCode + ' TransId: ' + transId);
+                // core payment process
+                $(\"[name='form_save']\").click();
```
adunsulag, line 184: You can use here:
```
document.querySelector("[name='form_save']").click();
```
or
```
document.querySelector("[name='form_save']").dispatchEvent(new Event('click'));
```

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+     * Return core sphere code in javascript block.
+     *   Will employ the front and production status in the rendering.
+     *
+     * @return string
+     */
+    private function renderSphereJsCore(): string
+    {
+        return "
+            function sphereNotSuccess(message) {
+                alert(message);
```
adunsulag, line 199: Are we logging these messages anywhere?  Or does Sphere backend log these messages for the provider/IT go and look them up?  

I've always found that good payment processing contains a log of the errors so you can quickly fix and resolve issues.

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+            }
+            function randomString(length, chars) {
+                let result = '';
+                for (let i = length; i > 0; --i) result += chars[Math.floor(Math.random() * chars.length)];
+                return result;
+            }
+            function paySphere() {
+                let total = $(\"[name='form_paytotal']\").val();
```
adunsulag, line 207: more jQuery that can be changed to use vanilla js.

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+                //  the user makes more than 1 payment from one screen, which would be very unusual to do).
+                let backupTicket = " . js_escape(RandomGenUtils::createUniqueToken(12)) . ";
+                let ticket = randomString(12, '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ');
+                if ((ticket == 0) || (ticket == '') || (ticket.length != 12)) {
+                    error.log('Dynamic javascript ticket creation failed, so using backup ticket.');
+                    ticket = backupTicket;
+                }
+
+                let responseUrl = " . js_escape($this->serverSite) . " + '/sphere/initial_response.php';
+                let cancelUrl = " . js_escape($this->serverSite) . " + '/sphere/initial_response.php?cancel=cancel&front=' + " . js_escape($this->front) . " + '&patient_id_cc=' + " . js_escape($this->patientIdCc) . " + '&csrf_token=' + " . js_escape(CsrfUtils::collectCsrfToken('sphere')) .  ";
```
adunsulag, line 230: Do we need to add any HMAC signature to make sure the variables were not modified / tampered with here? If another javascript was on this page someone could leave the CSRF alone but modify the other variables.  Maybe I'm being too paranoid here?

src/PaymentProcessing/SpherePayment.php
```diff
@@ -0,0 +1,257 @@
...
+    {
+        return '<button type="button" class="btn btn-success btn-transmit mx-1" onclick="paySphere()">' . xlt("Credit Card Pay") . '</button>';
+    }
+
+    /**
+     * Audit log entry for transaction.
+     */
+    public static function sphereAuditLog(int $pid, int $success, array $auditData, ?string $ticket = null, ?string $transactionId = null, ?string $actionName = null): void
+    {
+        $cryptoGen = new CryptoGen();
```
adunsulag, line 252: So this would be a good place if the transaction fails to store the results in the audit log (see javascript errors above).

sphere/initial_response.php
```diff
@@ -0,0 +1,53 @@
...
+ * Call to top.restoreSession() happens to ensure directed to correct session.
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2021 Brady Miller <brady.g.miller@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+if (!empty($_GET['cancel']) && ($_GET['cancel'] == 'cancel')) {
```
adunsulag, line 17: Does Sphere have any kind of signature validation for you to validate that the request originated from them?  We probably only want to process requests if they originated from sphere.

adunsulag (commented):
Hey @bradymiller Did an initial pass through here.  I haven't pulled it to play around with it but just some thoughts here.

adunsulag:
@bradymiller BTW this might make a great live demonstration of turning code into a module for our Nov 20th training.  You wrote this pretty well and it will go over fairly quickly into a module.  I may go ahead and do that next week.

bradymiller:
Just add a payment processing "report" where can audit things and where plan to support void/credit feature. As everything else, is still a work in progress. Next goal is to get void/credit working.

bradymiller:
@adunsulag , Thanks for reviewing the code! All very good points and will begin to address them after get void/credit working.

bradymiller:
btw @adunsulag , definitely feel free to modularize this if will be a good example for the Nov 20th training. Even though it's not complete, whatever you end up doing to make it a module, I can integrate it in future.

sphere/process_response.php
```diff
@@ -0,0 +1,82 @@
...
+    OpenEMR\Common\Session\SessionUtil::portalSessionCookieDestroy();
+    $ignoreAuth = false;
+    require_once(__DIR__ . "/../interface/globals.php");
+}
+
+use OpenEMR\Common\Crypto\CryptoGen;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Core\Header;
+use OpenEMR\PaymentProcessing\PaymentProcessing;
+
```
bradymiller, line 37: exit if Sphere is not turned on

sphere/process_revert_response.php
```diff
@@ -0,0 +1,84 @@
...
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once(__DIR__ . "/../interface/globals.php");
+
+use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Core\Header;
+use OpenEMR\PaymentProcessing\PaymentProcessing;
+use OpenEMR\PaymentProcessing\Sphere\SphereRevert;
```
bradymiller, line 22: exit if Sphere is not turned on

sphere/token.php
```diff
@@ -0,0 +1,52 @@
...
+ * @copyright Copyright (c) 2021 Brady Miller <brady.g.miller@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once(__DIR__ . "/../interface/globals.php");
+
+use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\PaymentProcessing\PaymentProcessing;
+use OpenEMR\PaymentProcessing\Sphere\SphereRevert;
```
bradymiller, line 20: exit if Sphere is not turned on

bradymiller:
This is now working and ready for review. I updated the initial PR message with details and screenshots.

@adunsulag , Converted to vanilla js and followed most other js recommendations. Pretty much logging everything (including unsuccessful transactions). Regarding the issue of bad actor modifying javascript, if a correctly authenticated user/patient decides to muck around with things by modifying javascript and directly making get/post requests then as with other places in OpenEMR things can be messed which would be logged (note that in the patient portal, changes require verification by a user in OpenEMR core), however, they will not be able to acquire sensitive info such as credit card numbers since this is done off of OpenEMR's server. Also the processing of callback requests from Sphere will require an open, active session and custom csrf token matching, so as above a correctly authenticated user/patient would need to decide to muck around with things that would be logged; as an aside the mechanism for the void/credit uses a more detailed sphere toolset where we get a token from sphere on backend, then direct to sphere with the token and get params (which need to match params sent to collect token), we then check hmac hash of the callback query from sphere (on the backend), and if that matches we then send a request to complete the transaction (on the backend).

bradymiller:
Working on wiki documentation here:
https://www.open-emr.org/wiki/index.php/Sphere_Payment_Processing_Module

Still working out details with Sphere on how accounts will be created etc. Then will get out a 6.0.0 patch (patch 4) with this feature :)

sjpadgett:
Nice work @bradymiller

sunsetsystems:
Thanks Brady, looks great!

stephenwaite:
southpark_cheering

PR #4733, opened 2021-11-10
Fix twig and appointment bugs.

adunsulag:
The fireEvent was missing the twig kernel.  No idea how that somehow got
left out of the code as it was working when I tested the fireEvent
stuff.

Also the manageTracker needed a bug fix on the appt status update.  It
was sending in the entire encounter instead of the encounter id.


bradymiller:
2DV

PR #4734, opened 2021-11-10
Prevent top nav being hidden

### LINKED ISSUE #4732: Broken Scrolling
### Describe the bug
Some point recently, OpenEMR's ability to scale within a browser window was broken.

When scaling the browser window between break-points, the content renders off screen, and there is no way to scroll to the right to see the hidden content. Previously, if the main menu (top menu) had many elements, it would stretch the content window beyond the field of view of the browser, but the user was able to scroll to the right to view hidden content.

### To Reproduce

Old EMR (ability to scroll right)

Screen Shot 2021-11-10 at 9 32 07 AM

New EMR (cannot scroll, content not visible and cannot access User's "logout" button)

Screen Shot 2021-11-10 at 9 24 46 AM
### Expected behavior

User should be able to scroll to see hidden content

### Client configuration

- Browser: FF
- OpenEMR version: Master
- Operating system: Mac Big Sur

sjpadgett:
move About to Administration drop down
adjust padding menu items

Fixes #4732


sjpadgett:
we now correctly collapse through all breakpoints. Right Admin User drop margin was preventing correct collapse breakpoint for some reason. That did the most to fix issue. @1220 px
image

sjpadgett:
Bringing in if no objections

sjpadgett:
I see @robertdown . You already cleaned up some of the redundant css classes and removed the brand icon. I didn't have latest and same with Ken. So i'll just fine tune a little more. Thanks

bradymiller:
Pretty sure admin is only permitted by certain acl groups.

bradymiller:
Also, to clarify, are we controlling this bug by limiting length of menu? What happens if a user has the Recall Board entry show up in the menu? Will users still then have the bug Ken brought up?

robertdown:
My changes were from a commit a few weeks ago, including the option to show/hide the logo. I _thought_ I hadn't pushed the PR but turns out I did.

sjpadgett:
That's okay. I'm just going to consolidate some classes from light-compact/full css into tabs_style_*.scss so have just the two place to adjust nav menu. 
I've never tried to do imm/problems from LBF except in NN.

sjpadgett:
FYI let's be careful when moving inline styles to compiled themes. Sometimes inlines are needed or serve a specific purpose.
A ckeditor style ended up in scss causing much grief trying to figure out why new instance editor would not size! (this was from 2008 and was good intentioned). Just beware..

sjpadgett:
This should work however, more can be done. currently top nav collapses @ 1200px. @kchapple

stephenwaite:
southpark_cheering

PR #4738, opened 2021-11-11
Use Description as Placeholder Option

### LINKED ISSUE #4737: LBF Description as Placeholder instead of actual Text
One of the options in the Layout Builder is "Description is Default Text" which injects the description as the text of the input field. What is the collective thought on changing that functionality to injecting the description as the _placeholder_ instead of the actual text. 

If the consensus is static default text should remain I'll just add another option, either "Use Label as Placeholder" or "User Description as Placeholder" and go from there.

I'm leaning towards adding another option, I just worry about option bloat.

@bradymiller @sjpadgett @sunsetsystems

robertdown:
Fixes #4737 

Adds a new Option for Layout Based Forms to use the description field as the placeholder on input types of text and textfield. Initial tests on usage on Text-date types failed so they are excluded from use at this time.

Initial pass is testing well.


bradymiller:
2DV-2

bradymiller:
yep, just have to wait one more year for union types :)

PR #4741, opened 2021-11-12
Encounter Screen UI Improvements

robertdown:
This PR moves the button-based, centered form-navigation on the Encounter screen to an bootstrap navbar. Additionally, it moves btn-primary buttons to more appropriate btn-secondary classes. Some small changes to the form title as well as prepping some inline code for a twig overhaul. The new navbar leverages twig. Fun fact, you don't have to migrate the ENTIRE page to twig in one go, it's flexible enough to take a phased approach.

### DIFF
--- a/src/Common/Twig/TwigExtension.php
+++ b/src/Common/Twig/TwigExtension.php
@@ -207,8 +207,8 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
 
             new TwigFilter(
                 'xlFormTitle',
-                function ($string, $mode = 'r', $prepend = '', $append = '') {
-                    return xl_form_title($string, $mode, $prepend, $append);
+                function ($string) {
+                    return xl_form_title($string);
                 }
             )
         ];
--- a/src/Common/Twig/TwigExtension.php
+++ b/src/Common/Twig/TwigExtension.php
@@ -203,6 +203,13 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
                 function ($string) {
                     return xl_document_category($string);
                 }
+            ),
+
+            new TwigFilter(
+                'xlFormTitle',
+                function ($string, $mode = 'r', $prepend = '', $append = '') {
+                    return xl_form_title($string, $mode, $prepend, $append);
+                }
             )
         ];
     }
--- a/templates/encounter/forms/navbar.html.twig
+++ b/templates/encounter/forms/navbar.html.twig
@@ -13,37 +13,37 @@
             {% for category, details in menuArray %}
                 {% if details.children is iterable %}
                     <li class="nav-item dropdown">
-                        <a class="nav-link dropdown-toggle" href="#" id="category_{{ category|xla }}" role="button" data-toggle="dropdown" aria-expanded="false">
-                            {{ category }}
+                        <a class="nav-link dropdown-toggle" href="#" id="category_{{ category|e('html_attr') }}" role="button" data-toggle="dropdown" aria-expanded="false">
+                            {{ category|e }}
                         </a>
-                        <div class="dropdown-menu" aria-labelledby="category_{{ category|xla }}">
+                        <div class="dropdown-menu" aria-labelledby="category_{{ category|e('html_attr') }}">
                             {% for item in details.children %}
                                 {% set formURL = rootdir ~ "/patient_file/encounter/load_form.php?formname=" ~ item.directory|url_encode %}
-                                <a href="#" class="dropdown-item" onclick="openNewForm('{{ formURL }}', {{ item.displayText|xlFormTitle|attr_js }})">{{ item.displayText|xlFormTitle|text }}</a>
+                                <a href="#" class="dropdown-item" onclick="openNewForm({{ formURL|json_encode|e('html_attr') }}, {{ item.displayText|xlFormTitle|json_encode|e('html_attr') }})">{{ item.displayText|xlFormTitle|e }}</a>
                             {% endfor %}
                         </div>
                     </li>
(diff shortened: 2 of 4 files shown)


robertdown:
# Before
image

# After
image

I admit, this page is still far away from what I'd like it to be, but it's certainly a step in the right direction

src/Common/Twig/TwigExtension.php
```diff
@@ -203,6 +203,13 @@ function ($string) {
                 function ($string) {
                     return xl_document_category($string);
                 }
+            ),
+
+            new TwigFilter(
+                'xlFormTitle',
+                function ($string, $mode = 'r', $prepend = '', $append = '') {
+                    return xl_form_title($string, $mode, $prepend, $append);
```
bradymiller, line 211: those last 3 parameters are not needed (legacy stuff), and would remove them in above 2 lines.

templates/encounter/forms/navbar.html.twig
```diff
@@ -0,0 +1,63 @@
...
+    <span class="navbar-brand">
+        {{ encounterDate|text }} {{ (groupEncounter == true) ? "Group Encounter"|xlt : "Encounter"|xlt}} {{ "for"|xlt }} {{ patientName|text }}
+    </span>
+
+    <div class="collapse navbar-collapse" id="navbarSupportedContent">
+        <ul class="navbar-nav mx-auto">
+            {% for category, details in menuArray %}
+                {% if details.children is iterable %}
+                    <li class="nav-item dropdown">
+                        <a class="nav-link dropdown-toggle" href="#" id="category_{{ category|xla }}" role="button" data-toggle="dropdown" aria-expanded="false">
```
bradymiller, line 16: guessing you meant to use attr instead of xla (also noted this in several other places below where category is shown in a attribute where prob meant to use attr instead of xla)

templates/encounter/forms/navbar.html.twig
```diff
@@ -0,0 +1,63 @@
...
+                            {{ category }}
```
bradymiller, line 17: text

bradymiller: also looks like couple places below where missing text for category

templates/encounter/forms/navbar.html.twig
```diff
@@ -0,0 +1,63 @@
...
+                        </a>
+                        <div class="dropdown-menu" aria-labelledby="category_{{ category|xla }}">
+                            {% for item in details.children %}
+                                {% set formURL = rootdir ~ "/patient_file/encounter/load_form.php?formname=" ~ item.directory|url_encode %}
+                                <a href="#" class="dropdown-item" onclick="openNewForm('{{ formURL }}', {{ item.displayText|xlFormTitle|attr_js }})">{{ item.displayText|xlFormTitle|text }}</a>
```
bradymiller, line 22: for formURL above, would remove the surrounding single quotes and pipe it through attr_js

templates/encounter/forms/navbar.html.twig
```diff
@@ -0,0 +1,63 @@
...
+            {% endfor %}
+            {% for category, details in moduleMenuArray %}
+                {% if details.children is iterable %}
+                    <li class="nav-item dropdown">
+                        <a class="nav-link dropdown-toggle" href="#" id="category_{{ category|xla }}" role="button" data-toggle="dropdown" aria-expanded="false">
+                            {{ category }}
+                        </a>
+                        <div class="dropdown-menu" aria-labelledby="category_{{ category|xla }}">
+                            {% for item in details.children %}
+                                <a href="#" class="dropdown-item" onclick="openNewForm('{{ item.href }}', {{ item.menu_name|xlFormTitle|attr_js }})">{{ item.menu_name|xlFormTitle|text }}</a>
```
bradymiller, line 40: for item.href above, would remove the surrounding single quotes and pipe it through attr_js

bradymiller:
@robertdown , very neat stuff! Just noted some escaping issues and some minor function legacy parameters that can go.

tenor (67)

robertdown:
@bradymiller @adunsulag @sjpadgett 

It looks like Twig has [escape functions ](https://twig.symfony.com/doc/3.x/filters/escape.html)that are far more comprehensive than ours. I would like to evaluate the use of the `escape()` filter for use. Currently we do **not** auto-escape and I think that practice should continue. However, my gut says we may get more mileage by leverage Twig's built in escape function versus our own. The built in function has support for all the key contexts including HTML, Javascript, CSS and HTML Attributes.

We don't have a ton of Twig templates right now so it would be worth it to make a decision. I feel like we've had this discussion before and we decided it was ok, but I can't quite remember.

[The actual EscapeExtension implementation](https://github.com/twigphp/Twig/blob/3.x/src/Extension/EscaperExtension.php)

bradymiller:
@robertdown ,
Strongly recommend pretending twig escaping functions do not exist. Having wrappers/functions controlled by the OpenEMR developers and not native php or a third party function have been extremely helpful (ie. very easy to adjust/fix things codebase wide if needed; for example, if a certain strategy like js escaping is problematic (ie. mangles something) or a vulnerability comes up, we can just fix it in the entire codebase with 1 fix). Current functions are roadtested (no reported vulnerabilities) for at least several years while also not mangling things. And having same function nomenclature across frameworks also make it way easier for developers (ie. when in native php do this, when in twig do this, when in so and so do this, will not go over well; was hard enough getting the current six or so escaping functions into developers habits). And from a selfish perspective since I am basically the escaping police, I do not want to deal with a entire new category of escape policing :)

bradymiller:
2DV

PR #4744, opened 2021-11-14
couple fixes in edit layout

bradymiller:
hi @robertdown , couple very minor fixes in edit layout and noted some things that I couldn't figure out how to fix.

First want to make sure this will work for when a translation is shown for the group name (note this only happens when not using english where it shows translated group name; just made it green to match the translations that are done below that):
image

I was unable to fix odd looking big buttons that show up after clicking the 'Add Field' and 'Rename Group' buttons:
image

image

### DIFF
--- a/interface/super/edit_layout.php
+++ b/interface/super/edit_layout.php
@@ -1472,9 +1472,9 @@ function myChangeCheck() {
 
 <?php if ($layout_id) { ?>
 <div class="btn-group ml-auto">
-    <button type='button' class='btn btn-secondary btn-sm' onclick='edit_layout_props("")'><?php echo xla('Layout Properties'); ?></button>
-    <button type='button' class='btn btn-secondary btn-sm addgroup' id='addgroup'><?php echo xla('Add Group'); ?></button>
-    <button type='button' class="btn btn-primary btn-save btn-sm" name='save' id='save'><?php echo xla('Save Changes'); ?></button>
+    <button type='button' class='btn btn-secondary btn-sm' onclick='edit_layout_props("")'><?php echo xlt('Layout Properties'); ?></button>
+    <button type='button' class='btn btn-secondary btn-sm addgroup' id='addgroup'><?php echo xlt('Add Group'); ?></button>
+    <button type='button' class="btn btn-primary btn-save btn-sm" name='save' id='save'><?php echo xlt('Save Changes'); ?></button>
 </div>
 <br>
     <?php echo xlt('With selected');?>:&nbsp;
@@ -1488,7 +1488,7 @@ function myChangeCheck() {
 <input type='button' class='btn btn-secondary btn-sm' value='<?php echo xla('Tips'); ?>' onclick='$("#tips").toggle();' />&nbsp;
 <input type='button' class='btn btn-secondary btn-sm' value='<?php echo xla('Encounter Preview'); ?>' onclick='layoutLook();' />
 <?php } else { ?>
-<button type='button' class='btn btn-primary btn-sm btn-add btn-new' onclick='edit_layout_props("")'><?php echo xla('New Layout'); ?></button>
+<button type='button' class='btn btn-primary btn-sm btn-add btn-new' onclick='edit_layout_props("")'><?php echo xlt('New Layout'); ?></button>
 <?php } ?>
 
 <div id="tips" class="container tips">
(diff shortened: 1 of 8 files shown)


robertdown:
Acknowledged, I'm guessing an errant `</div>` somewhere, I will prioritize this and address today or tomorrow

bradymiller:
@robertdown , thanks and sounds good. I'll bring this in to avoid merge conflicts and then you can just work from master.

bradymiller:
@robertdown, or actually, looks like you can just work from this PR directly. Just let me know whats best for you and will do it.

robertdown:
Will work from your PR

robertdown:
Ok I've been able to reproduce this and have identified the problem, it has to do with the `btn-group` class. I _could_ just remove that class but I'm trying to refactor a bit to improve the UI of the nav as I'm just not happy with it yet. Expecting to wrap tomorrow because I have to get moving on the LBF Subform stuff

robertdown:
Ok @bradymiller that was a headache and a half. I pushed the changes to your branch. Can you pull down and test - it's working for me but I've been staring at it for far too long today

robertdown:
@sjpadgett check out of some of these UI changes to the LBF Editor, I think you'll appreciate them

sjpadgett:
Okay @robertdown I'll pull down in the morning, test and get back to ya. Looks good from what I see at this point.

robertdown:
I have a feeling @bradymiller will get to it tonight. I was just giving you a heads up that some small changes have been applied and I think it makes a big difference :)

sjpadgett:
I was going to go ahead and pull it down but your latest commits won't show until Brady pulls into his branch.

robertdown:
Hm, interesting. I thought I successfully pushed to @bradymiller's branch? The PR retested

sjpadgett:
woops sorry, forgot to fetch! I got it now...

robertdown:
I think I forgot to wire up the copy mechanism

sjpadgett:
Looks so much better. So happy we got the fixed top nav back. Thanks for doing this and I'm inspired again to maybe work on some of the rendering styling soon.

robertdown:
I think my favorite change on this PR is the group header now having `stick-top` applied. Makes life so much easier. And I've started work on the Sub-Form process!

sjpadgett:
The sub form will be very useful I believe.

interface/super/edit_layout.php
```diff
@@ -1764,16 +1690,20 @@ function myChangeCheck() {
    <td><input type='text' class='form-control' name="newcodes" id="newcodes" value="" onclick='select_clin_term_code(this)' size='10' maxlength='255' /> </td>
   </tr>
   <tr>
-   <td colspan="9">
-    <input type="button" class="btn btn-primary btn-sm savenewfield" value='<?php echo xla('Save New Field'); ?>' />
-    <input type="button" class="btn btn-secondary btn-sm cancelnewfield" value='<?php echo xla('Cancel'); ?>' />
+   <td colspan="15">
+       <div class="w-100 justify-content-end d-flex">
+           <button type="button" class="btn btn-link btn-sm cancelnewfield"><?php echo xla('Cancel'); ?></button>
+           <button type="button" class="btn btn-primary btn-save btn-sm savenewfield"><?php echo xla('Save New Field'); ?></button>
```
bradymiller, line 1696: xlt to above 2 lines

templates/admin/lbf/edit_lbf/partials/layout_toolbar.html.twig
```diff
@@ -0,0 +1,19 @@
+<div class="btn-toolbar ml-auto" role="toolbar" aria-label="{{ "Layout toolbar"|xla }}">
+    {% if layoutID %}
+    <div class="mr-2 btn-group btn-group-sm">
+        <button type="button" class="btn btn-secondary" onclick="edit_layout_props('')">
+            <i class="fa fa-cog"></i>&nbsp;{{ "Layout Properties"|xlt }}
+        </button>
+        <button type="button" class="btn btn-secondary btn-add" data-toggle="modal" data-target="#newgroup" onclidk="AddGroup()">{{ "Add Group"|xlt }}</button>
```
bradymiller, line 7: mispell in onclick

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
```diff
@@ -0,0 +1,37 @@
+<!-- template DIV that appears when user chooses to add a new group -->
+<div class="modal" id="{{ id }}group" tabindex="-1" aria-labelledby="{{ id }}GroupModal" aria-hidden="false">
```
bradymiller, line 2: Would pipe all the id outputs to attr in this block

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
```diff
@@ -0,0 +1,37 @@
...
+    <div class="modal-dialog modal-lg modal-dialog-scrollable">
+        <div class="modal-content">
+            <div class="modal-header">
+                {% set modalTitle = (id == "rename") ? "Rename Group" : "Add New Group" %}
+                {% set btnSaveValue = (id == "rename") ? "Rename Group" : "Save New Group" %}
```
bradymiller, line 7: note i do not know the solution, but the 2 lines of code above will be impossible to capture in the translation pipeline (those 3 constants). Prob not a big deal hear since i bet these 3 constants are already in the translation list. But generally always best to partner a constant directly with an xl function so it gets captured in the translation pipeline (ie. a perl script that seeks them out; btw, reminds me that we will need to update that script to find the new formatted translation constants in twig templates; shouldn't be tough).

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
```diff
@@ -0,0 +1,37 @@
...
+                <div class="form-group">
+                    <label for="{{ id }}groupname" class="form-label">{{ "Group Name"|xlt }}</label>
+                    <input type="text" size="20" maxlength="30" name="{{ id }}groupname" id="{{ id }}groupname" class="form-control">
+                </div>
+                <div class="form-group">
+                    <label for="{{ id }}groupparent" class="form-label">{{ "Parent"|xlt }}</label>
+                    <select name="{{ id }}groupparent" class="select2 form-control" id="{{ id }}group_parent">
+                        <option value="">{{ "None{{group}}"|xlt }}</option>
+                        {% for group in groups %}
+                            <option value="{{ group.id|attr }}" {{ group.default }}>{{ group.title|text }}</option>
```
bradymiller, line 26: might as well pipe group.default through attr (guessing this is just selected or something, but won't hurt)

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
```diff
@@ -0,0 +1,37 @@
...
+                        {% endfor %}
+                    </select>
+                </div>
+            </div>
+            <div class="modal-footer">
+                <button type="button" class="btn btn-link cancel{{ id }}group" data-dismiss="modal">{{ 'Cancel'|xl }}</button>
+                <button type="button" class="btn btn-primary btn-save save{{ id }}group">{{ btnSaveValue|xl }}</button>
```
bradymiller, line 33: xlt in above 2 lines

bradymiller:
@robertdown , Just noted some minor stuff (note the translation constant comment is more something to think about in future coding and do not need to address that in this PR). Wasn't able to test, but have faith :)

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
robertdown: I don't think we need to escape in this context, we are in total control of the variable and it's either "new" or "rename"

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
robertdown: Noted, I'll look into a solution

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
robertdown: Is there really any benefit to running it through the attribute function though?

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
bradymiller: could just think about it this way.
It will do no harm (will not break any html) placing it there, will only be a couple keystrokes, and then never, ever need to worry about it again (maybe in 2 years somebody decides to allow some sort of user input to get into id and we are not around to remember this).

templates/admin/lbf/edit_lbf/partials/modify_group.html.twig
bradymiller: if somehow, some way somebody modifies code in future for group.default to be user defined and get javascript in there, then yes (I honestly don't see how it would be possible, but since these code elements are now further removed 
 from each other maybe a future coding mistake or something).
As above discussed, though, no harm done, easy to do, and then never worry about again.

robertdown:
bump

robertdown:
bump

PR #4746, opened 2021-11-16
LBF Improvement

robertdown:
Minor UI changes to the LBF Edit Properties UI which Save/Cancel buttons to the end, turns them into proper buttons, and de-emphasized the Cancel button


robertdown:
Oops, pushed the wrong thing

PR #4751, opened 2021-11-17
Encounter Menu Hook

robertdown:
Support for `EncounterMenuEvent`. Currently only support for `MENU_RENDER` which is used to populate the Twig variable for output.

Forms.php builds traditional forms menu and zend hooks. Both of those were wrapped in event listeners to streamline a future project to move those out of forms.php and into a better location. (Just means technically this page both adds 2 listeners and then fires the dispatch()).

Is testing well

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,7 @@
         "symfony/config": "4.4.33",
         "symfony/dependency-injection": "4.4.33",
         "symfony/event-dispatcher": "4.4.30",
+        "symfony/event-dispatcher-contract": "2.4.0",
         "symfony/http-foundation": "4.4.33",
         "symfony/yaml": "4.4.29",
         "twig/twig": "3.3.3",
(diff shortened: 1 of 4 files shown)


robertdown:
Note I also brought in `Symfony\Contracts\EventDispatcher\Event` in favor of `Symfony\Component\EventDispatcher\Event` which has been deprecated as of 4.3. We need to migrate over, but for now thought it best to not introduce new code with deprecated dependencies.

bradymiller:
tenor99

adunsulag:
@robertdown I reviewed your code and didn't see any problems.

DennisonWilliams:
This is great.  It will help me pull out some label printing code I am maintaining in core (in a clone of openemr/openemr).  When is this scheduled for release?

robertdown:
Generally this kind of thing doesn't get included in patches, just the next release. @bradymiller thoughts? @DennisonWilliams Expect to see quite a few new hooks coming into high-touch points areas in the coming months

bradymiller:
Yep, will be in next release (6.1.0), which is likely gonna be in about 2 months or so.

PR #4756, opened 2021-11-20
Patient portal appt fix for preferred categories.

adunsulag:
If you had a category other than office_visit (catid=5) then a patient couldn't find any provider available appointments to schedule with on the patient portal.  So if additional categories were added etc, a patient could never make the appointment.

Fixed this issue so now the finder will use the category for the search just like it does with the provider.


bradymiller (approved):

bradymiller:
2DV

PR #4757, opened 2021-11-20
Add appointment service status methods.

adunsulag:
Added the duration field to the getAppointment method.  Added two helper methods for grabbing appointment status lists and checking if an appointment status is valid.  Minor documentation improvement as well.


bradymiller (approved):

bradymiller:
2DV-2

PR #4762, opened 2021-11-22
Fix portal fatal errors due to not found class.

### LINKED ISSUE #4761: Portal Secure Chat is throwing fatal errors due
These two lines are missing the namespace separator '\' before the OpenEMR namespace.  Which is creating fatal errors.

Relevant code here:
https://github.com/openemr/openemr/blob/8bf70dd541b65226a187233e5d51a6abb935444b/portal/messaging/secure_chat.php#L54
https://github.com/openemr/openemr/blob/8bf70dd541b65226a187233e5d51a6abb935444b/portal/messaging/secure_chat.php#L57

Will push out a PR shortly to fix this.

adunsulag:
Fixes #4761

Recent portal changes introduced a bug with the SessionUtil class.  This post adds the namespace escape for the OpenEMR namespace.  Essentially 2 character changes so going to bring this in.

### DIFF
--- a/portal/messaging/secure_chat.php
+++ b/portal/messaging/secure_chat.php
@@ -51,10 +51,10 @@ use OpenEMR\PatientPortal\Chat\ChatController;
 define('C_USER', IS_PORTAL ?  IS_PORTAL : IS_DASHBOARD);
 
 if (isset($_REQUEST['fullscreen'])) {
-    OpenEMR\Common\Session\SessionUtil::setSession('whereto', '#messagescard');
+    \OpenEMR\Common\Session\SessionUtil::setSession('whereto', '#messagescard');
     define('IS_FULLSCREEN', true);
 } else {
-    OpenEMR\Common\Session\SessionUtil::setSession('whereto', '#messagescard');
+    \OpenEMR\Common\Session\SessionUtil::setSession('whereto', '#messagescard');
     define('IS_FULLSCREEN', false);
 }
 


sjpadgett:
Why doesn't it break on Windows

adunsulag:
@sjpadgett I do not know.  I would have thought it would.

sjpadgett:
this should prevent errors and from what I see the includes are there
require_once(__DIR__ . "/../src/Common/Session/SessionUtil.php");

adunsulag:
@sjpadgett The autoloader on my machine as well as on an AWS EC2 instance failed to find the class when it was missing the backslash in front of the namespace.  I would have expected it to fail on windows as well, not sure why it doesn't.

sjpadgett:
yes but it's not auto loaded but is included at top of script
https://github.com/adunsulag/openemr/blob/b8d431efb161a01597dc173e7f4954ca8ab182cd/portal/messaging/secure_chat.php#L20

sjpadgett:
Don't worry about what you have is fine. Now though, I'm concerned of other places in core!

adunsulag:
@sjpadgett Even if its included at the top of the script without the backslash, PHP treats it as a completely different class definition.  We probably haven't seen it that much in core because we usually use the **use <class>** statement instead of specifying the Fully Qualified Class Name (FQCN).

sjpadgett:
Okay i'm confused. Because the class is included(same as if auto loaded) and it's a static method then I always assumed 
`OpenEMR\Common\Session\SessionUtil` is same as self::
by adding the slash the interrupter assumes a direct class path and included class is ignored.
All the portal session start and destroy have been called this way for a couple years.
Still it's prob not best practice and most likely will refactor these next PR to use the included class as self::.

Or I may be using the PHP Vjerry instead of V8:)

PR #4764, opened 2021-11-22
Changing event-dispatcher-contracts dependency

### LINKED ISSUE #4763: event-dispatcher-contracts package name incorrect and conflicts with OpenEMR packages
I was doing some work in the composer.json for a custom module and trying to update my custom module composer file and it threw errors with the latest code from master.

The event-dispatcher-contracts package had a type in it and has the singular name instead of the plural.  Also the 2.4.0 conflicts with the [event dispatcher dependency](https://github.com/symfony/event-dispatcher/blob/4.4/composer.json) version 4.4 which has a hard requirement of ^1.1
```
"symfony/event-dispatcher-contracts": "^1.1",
```

It looks like the best fix is to downgrade the version to 1.9 instead of the 2.4 which is what is already explicit in the lock file.  I did some testing and didn't see anything breaking in the system so I'll submit my PR.  @robertdown if you can take a look at this and let us know what you needed with the 2.4.0 release that would be great.  We'll need to upgrade the event dispatcher to get there and I haven't looked into that to see what / if any problems we'd have in doing that.

adunsulag:
Made this to match the lock file.  Fixed the mispelling and the version.

Fixes #4763.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
         "symfony/config": "4.4.33",
         "symfony/dependency-injection": "4.4.33",
         "symfony/event-dispatcher": "4.4.30",
-        "symfony/event-dispatcher-contract": "2.4.0",
+        "symfony/event-dispatcher-contracts": "1.1.9",
         "symfony/http-foundation": "4.4.33",
         "symfony/yaml": "4.4.29",
         "twig/twig": "3.3.3",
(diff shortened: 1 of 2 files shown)


adunsulag:
Since master is broken I'm going to let the tests run and if its looking good, I'll pull this in.

bradymiller:
2DV

PR #4767, opened 2021-11-24
Standard table alert for layout editor

mdsupport:
Warn users to avoid deletion of standard columns.

### DIFF
--- a/interface/super/edit_layout.php
+++ b/interface/super/edit_layout.php
@@ -401,7 +401,7 @@ $layout_id = empty($_REQUEST['layout_id']) ? '' : $_REQUEST['layout_id'];
 $layout_tbl = tableNameFromLayout($layout_id);
 // Setup alert for table layouts
 $layout_alert = "";
-if ($layout_tbl == "") {
+if (($layout_id !== "") && ($layout_tbl !== "")) {
     $layout_alert = sprintf(
         '<div class="alert alert-warning alert-dismissible fade show" role="alert">
             <strong>%s!</strong> "%s" %s.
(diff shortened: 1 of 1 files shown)


sjpadgett:
Rod brought in a feature to prevent users from deleting sensitive items or rename id's from LBF.
Fyi we have an promised alert, confirm etc i.e 'dialog.alert(message).then(...)'
or timed with promise

(async (time) => {
       await asyncAlertMsg(rtn, time, 'success', 'lg');
                })(2000).then(rtn => {
                    //document.edit_form.submit();
                });

mdsupport:
> prevent users from deleting sensitive items

At least this user likes to be warned upfront.

> promised alert

Do we want alerts identified at server to be handled by javascript calls on ready?

sjpadgett:
All the alerts and popups can be called in any context. Above is not on ready event. What you have is fine. I was just pointing out we have several popups available of which return a promise and very handy.

interface/super/edit_layout.php
```diff
@@ -398,6 +398,18 @@ function encodeModifier($jsonArray)
 
 // The layout ID identifies the layout to be edited.
 $layout_id = empty($_REQUEST['layout_id']) ? '' : $_REQUEST['layout_id'];
+$layout_tbl = tableNameFromLayout($layout_id);
+// Setup alert for table layouts
+$layout_alert = "";
+if (($layout_id !== "") && ($layout_tbl !== "")) {
+    $layout_alert = sprintf(
+        '<div class="alert alert-warning alert-dismissible fade show" role="alert">
+            <strong>%s!</strong> "%s" %s.
```
bradymiller, line 407: place the exclamation mark in the actual constant below

bradymiller:
@mdsupport , Code looks good to me. If this test well for you and the psr12 styling issue is addressed and passes styling test, then will bring this in.

bradymiller:
thanks @mdsupport ! bringing in

2DV

PR #4768, opened 2021-11-25
Venom dx codes

robertdown:
Adds support for the [VeNom Coding Group](http://venomcoding.org/) codes as an external data load. Users still need to obtain the codes from the coding group but like other codes, once the file is placed in the proper location it can be installed.


robertdown:
Ok, I'm gonna turn this back into a draft - it is not nearly as comprehensive as I would like. I'm thinking instead of doing this as an external data load, I'll use a custom module approach which will provide way more support.

robertdown:
Deleting

PR #4779, opened 2021-12-02
make the insane dev environment less insane

bradymiller:
still need to test this


bradymiller:
These changes will get the insane dev environment down to a very reasonably insane 33 dockers :)

btw, here's a screenshot showing PHP 8.2-dev in action (due to be released about a year from now :) ):
image

bradymiller:
testing well and bringing in

PR #4781, opened 2021-12-03
first php 8.2 fix

bradymiller:
Related to an interesting deprecated error in future PHP 8.2 (noted it while was updating the insane dev environment) that is discussed by php folks here:
https://wiki.php.net/rfc/deprecate_dynamic_properties

Notably, the deprecation errors do not actually go away with this, but assuming this is simply because of a bug in php 8.2 since very early in development.

Makes sense to do this for CryptoGen class since it is using dynamic properties to cache keys (much more efficient doing this).

Will reengineer the EventAuditLogger class a tiny bit since really don't need a dynamic property for this. Prob just initiate it to null and then create the new CryptoGen call if the variable is not instance of CryptoGen class (or something like that). (btw caching that CryptoGen isntance also makes things much more efficient since will then cache the keys on the potential numerous encryptions done while logging)

### DIFF
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -30,6 +30,7 @@ namespace OpenEMR\Common\Crypto;
 
 use OpenEMR\Common\Utils\RandomGenUtils;
 
+#[AllowDynamicProperties]
 class CryptoGen
 {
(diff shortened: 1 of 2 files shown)


bradymiller:
testing well and bringing in

PR #4782, opened 2021-12-03
Adding support for another fax module

juggernautsei:
Fixes #
support for fax module being built
The change to the menu to show the fax/scan feature if this module is enabled

Add documofax_enable to the globals req in the menu to show the fax/scan

### DIFF
--- a/interface/main/tabs/menu/menus/standard.json
+++ b/interface/main/tabs/menu/menus/standard.json
@@ -2100,7 +2100,8 @@
         ],
         "global_req": [
           "enable_hylafax",
-          "enable_scanner"
+          "enable_scanner",
+          "oedocumofax_enable"
         ]
       },
       {


juggernautsei:
This is where the build takes place. 
https://github.com/juggernautseinc/oe_module_batch_creditcard/tree/main/interface/modules/custom_modules/oe-module-documo-fax

interface/main/tabs/menu/menus/standard.json
```diff
@@ -2100,7 +2100,8 @@
         ],
         "global_req": [
           "enable_hylafax",
-          "enable_scanner"
+          "enable_scanner",
+          "oedocumofax_enable"
```
adunsulag, line 2104: Why can't your module listen to the MenuEvent::MENU_UPDATE event and add the global_req change you need there?  Similar to what I show in my bootstrap example:
https://github.com/adunsulag/oe-module-custom-skeleton/blob/00a7d067da09594a524b2f8d23274cb05f5c46b9/src/Bootstrap.php#L245

```php
    foreach ($menu as $item) {
		if ($item->menu_id == 'fax') {
		    $item->global_req[] = "oedocumofax_enable";
		    break;
		}
	}
```

juggernautsei: I can do that. I did it at first but the system places the new item at the end of the list :(. So, I thought why not reuse what is there. But there are tradeoffs everywhere I turn. I can do it the other way no problem. Just trying to reuse what is there with minimum codebase changes. I can go for no codebase changes also.

juggernautsei:
Done, updated module to add Fax to the misc menu.

interface/main/tabs/menu/menus/standard.json
adunsulag: I'm confused here.  Aren't you just adding a check to make sure the globals flag has been set for the menu item to show up? How does adding that global flag check inside your module for the menu item put the item at the end of the list?  I don't see how it could be doing that at all if you use the modified snippet I put up above.

PR #4785, opened 2021-12-06
Care Coordination Import Fix

kchapple:
Fix importer PHP issue where string evaluated to 0 and caused medical problems to not be imported
@sjpadgett 

Fixes #


interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/CarecoordinationTable.php
```diff
@@ -477,7 +477,7 @@ public function insert_patient($audit_master_id, $document_id)
                 $arr_prescriptions['lists3'][$b]['provider_state'] = $newdata['lists3']['provider_state'];
                 $arr_prescriptions['lists3'][$b]['provider_root'] = $newdata['lists3']['provider_root'];
                 $b++;
-            } elseif ($table == 'lists1' && $newdata['lists1']['list_code'] != 0) {
+            } elseif ($table == 'lists1' && !empty($newdata['lists1']['list_code'])) {
```
sjpadgett, line 480: Makes sense. Thanks. Still strange works okay on windows though!
I'll bring in once actions clear...

PR #4789, opened 2021-12-11
Allowing the system to fire patient report/document event always

juggernautsei:
Fixes #
These events were wrapped in if statements checking if oefax is installed
I am building a fax module. I need to use the current event dispatcher that is in the system. 

Remove the if statements so that the event fire always and not only when the faxsms module is installed.

### DIFF
--- a/interface/patient_file/report/patient_report.php
+++ b/interface/patient_file/report/patient_report.php
@@ -40,12 +40,12 @@ $auth_relaxed  = AclMain::aclCheckCore('encounters', 'relaxed');
 $auth_med      = AclMain::aclCheckCore('patients', 'med');
 $auth_demo     = AclMain::aclCheckCore('patients', 'demo');
 
-$oefax = !empty($GLOBALS['oefax_enable']) ? $GLOBALS['oefax_enable'] : 0;
 /**
  * @var EventDispatcherInterface $eventDispatcher  The event dispatcher / listener object
  */
 $eventDispatcher = $GLOBALS['kernel']->getEventDispatcher();
 ?>
+<!DOCTYPE>
 <html>
 <head>
 <title><?php echo xlt("Patient Reports"); ?></title>
@@ -269,9 +269,9 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                         <button type="button" class="genpdfrep btn btn-primary btn-download btn-sm" value="<?php echo xla('Download PDF'); ?>" ><?php echo xlt('Download PDF'); ?></button>
 
                         <?php
-                        if ($oefax) {
+
                             $eventDispatcher->dispatch(PatientReportEvent::ACTIONS_RENDER_POST, new GenericEvent());
-                        }
+
                         ?>
                         <input type='hidden' name='pdf' value='0' />
                         <br />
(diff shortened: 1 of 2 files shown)


juggernautsei:
Also, I installed a new tool for me in my PHPStorm. It is called SonarLint. It made a lot of code compliance suggestions. I only made a few of those changes in this pull request. I did not want to overstep. I have attached a screenshot of some of the suggestions made. 

image

bradymiller:
Looks good to me. @adunsulag , @sjpadgett , I think there was discussion regarding removing this global stuff previously, just making sure you are ok with this.

adunsulag:
@bradymiller So I prefer we remove the if conditionals.  We will need to make sure we update @sjpadgett fax module to have the global conditional in there.  I probably won't get to that for a while so I can see it breaking people who rely on the module and are using master... but hopefully people aren't doing that or if @sjpadgett has extra cycles he can get in and fix it.

juggernautsei:
The global is in @sjpadgett  module.
https://github.com/openemr/oe-module-faxsms/blob/62e61d2b4f97e38076e846f57b89fa3edd7a3752/openemr.bootstrap.php#L125
It is not in the codebase.

adunsulag:
@juggernautsei So @sjpadgett creates the global in his module but people still have to turn the global on with the current behavior in order to run his fax module.  Right now, the fax module that's installed, but has not been enabled will always render the javascript and html in his module because the global flag check happens in core instead of in the module.  That will need to be fixed, though the odds of someone having his module installed but NOT enabled are probably pretty low so its more of a minor bug rather than a major bug.

juggernautsei:
@adunsulag are you suggesting the event listeners be wrapped in a conditional?

adunsulag:
@juggernautsei No, I'm not suggesting any change with what you've done here in core.  I'm saying that @sjpadgett's fax module will need to be fixed at some point when this comes in because there's a small use case where it will break for users due to this change.

sjpadgett:
I don't see an issue where the only reason I wrapped the dispatch was I didn't want to fire unless module was installed for resource reasons. i.e. why fire if not needed. Seem this is marginal in grand scheme of things.
If issue comes up, I'll take care of.
Bringing in unless more to be done @juggernautsei

interface/patient_file/report/patient_report.php
```diff
@@ -40,12 +40,12 @@
 $auth_med      = AclMain::aclCheckCore('patients', 'med');
 $auth_demo     = AclMain::aclCheckCore('patients', 'demo');
 
-$oefax = !empty($GLOBALS['oefax_enable']) ? $GLOBALS['oefax_enable'] : 0;
 /**
  * @var EventDispatcherInterface $eventDispatcher  The event dispatcher / listener object
  */
 $eventDispatcher = $GLOBALS['kernel']->getEventDispatcher();
 ?>
+<!DOCTYPE>
```
bradymiller, line 48: hi @juggernautsei , just noted this `<!DOCTYPE>` here. Why was this added?

bradymiller: I found a couple other places where this is used in codebase, but they contain html `<!DOCTYPE html>`

juggernautsei: I installed a new tool for me in my PHPStorm. It is called SonarLint. It made a lot of code compliance suggestions. I only made a few of those changes in this pull request.

PR #4791, opened 2021-12-12
php 8.2 fix

bradymiller:
php 8.2 fix


bradymiller:
testing well and bringing in

PR #4798, opened 2021-12-18
testing removal of a line of code in kernel class

bradymiller:
just seeing what happens if remove this line of code on ci testing
@robertdown @adunsulag odd to have this include in a class, but guessing there is a reason for it

### DIFF
--- a/src/Core/Kernel.php
+++ b/src/Core/Kernel.php
@@ -8,8 +8,6 @@
 
 namespace OpenEMR\Core;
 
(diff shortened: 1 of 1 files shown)


robertdown:
Hm, I'm not sure. Especially because the Kernel gets injected into $GLOBALS in the globals file

bradymiller:
Looked through all the code that uses this and can't see any reason why this line should be here. Gonna remove this for now. Really need to keep control of where this globals.php file is included since it is critical for connecting cookie/session/site in addition to the other hundred things it does :)

PR #4800, opened 2021-12-20
Hook into code types

robertdown:
Adding support for external codes with an event listener. The actual installation happens in a custom module. However, the event listener isn't actually working as expected. @adunsulag can you take a look at this? Looks like the EventDispatcher in the code_types file isn't the right ED and I can't tell why


adunsulag:
@robertdown I'm not seeing anything just looking at the code that would create the problem.  It's possible there's a bug with the compatibility layer in PHP that the Symfony project put in to backport Symfony 4.3 from the 5.4 release.  They added a compatibility layer to handle using event names(Symfony 4.3) still instead of the event object directly (Symfony 5.4) and there might be something there that is giving you issues.

I'd try using an event name before passing the class and see if that then works.  You may have already tried that, but in case you didn't I'd try this:
```
$eventDispatcher->dispatch(ExternalCodesCreatedEvent::EVENT_HANDLE, $externalCodesEvent);
```

The other possibility is that in the place you are expecting your event to fire the code_types.inc.php file is being included before globals.php.  I've seen a couple places in the codebase where code_types is included without the globals.php file being included so that would mess up the firing order also as your event listeners would be registered after the event had fired.  Youd need to look at the file include hierarchy where you expect the event to be dispatched and see if that is the case (if you have xdebug you could always step through this and see whats going on).

If none of that works I'll need to pull the code down and play with it.  Unfortunately, I'm not sure if I will get to it this week as I'm doing a lot of traveling for family holiday events.

custom/code_types.inc.php
```diff
@@ -47,11 +47,18 @@
...
  * @copyright Copyright (c) 2006-2010 Rod Roark <rod@sunsetsystems.com>
  * @copyright Copyright (c) 2019 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2021 Robert Down <robertdown@live.com>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+use OpenEMR\Events\Codes\ExternalCodesCreatedEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+
+require_once dirname(__FILE__) . "/../interface/globals.php";
```
bradymiller, line 57: requiring this line in a library generally means something is wrong in the script that is using this library. Guessing that whatever is requiring this line in your module is potentially causing your issue.

custom/code_types.inc.php
```diff
@@ -47,11 +47,18 @@
...
 require_once(__DIR__ . "/../library/csv_like_join.php");
 
+global $GLOBALS;
```
bradymiller, line 60: related to note above, also odd that need to make $GLOBALS global since that is already implicit for that array

.gitignore
```diff
@@ -12,6 +12,10 @@ node_modules
 !/public/assets/modified
 /.docs/.out/*
 
+# Custom modules
+interface/modules/custom_modules/**
+!interface/modules/custom_modules/README.md
```
bradymiller, line 17: would probably do this instead in a separate commit/PR that also supported a interface/modules/core_modules/ as discussed on saturday call. if have time over next couple days, i may take a stab at that functionality :)

bradymiller: just realized you probably have your custom module in there now, so rec just keeping this in place then.

robertdown:
@adunsulag ok, still no dice. If I dump the event dispatcher in `code_types.inc.php` I get this:
```php
Symfony\Component\EventDispatcher\EventDispatcher Object
(
    [listeners:Symfony\Component\EventDispatcher\EventDispatcher:private] => Array
        (
        )

    [sorted:Symfony\Component\EventDispatcher\EventDispatcher:private] => Array
        (
        )

    [optimized:Symfony\Component\EventDispatcher\EventDispatcher:private] => Array
        (
        )

)
```

however, if I dump the event listener from my Bootstrap file in the module, I get this:
```php
Symfony\Component\EventDispatcher\EventDispatcher Object
(
    [listeners:Symfony\Component\EventDispatcher\EventDispatcher:private] => Array
        (
            [menu.update] => Array
                (
                    [0] => Array
                        (
                            [0] => Array
                                (
                                    [0] => Application\Listener\ModuleMenuSubscriber Object
                                        (
                                        )

                                    [1] => onMenuUpdate
                                )

                            [1] => Array
                                (
                                    [0] => ProvidenceHealthTech\Venom\Bootstrap Object
                                        (
                                            [eventDispatcher:ProvidenceHealthTech\Venom\Bootstrap:private] => Symfony\Component\EventDispatcher\EventDispatcher Object
 *RECURSION*
                                            [service:ProvidenceHealthTech\Venom\Bootstrap:private] => ProvidenceHealthTech\Venom\Service\Setup Object
                                                (
                                                    [externalCodeTables:ProvidenceHealthTech\Venom\Service\Setup:private] => Array
                                                        (
                                                            [venom_dx] => VeNom Diagnoses
                                                            [venom_dx_test] => VeNom Diagnostic Tests
                                                            [venom_proc] => VeNom Procedures
                                                            [venom_admin] => VeNom Admin Tasks
                                                        )

                                                )

                                        )

                                    [1] => addVenomMenuItem
                                )

                        )

                )

            [menu.restrict] => Array
                (
                    [0] => Array
                        (
                            [0] => Array
                                (
                                    [0] => Application\Listener\ModuleMenuSubscriber Object
                                        (
                                        )

                                    [1] => onMenuRestrict
                                )

                        )

                )

            [service.save.post] => Array
                (
                    [0] => Array
                        (
                            [0] => Array
                                (
                                    [0] => OpenEMR\ZendModules\FHIR\Listener\UuidMappingEventsSubscriber Object
                                        (
                                            [fhirVitalObservationResourcePaths:OpenEMR\ZendModules\FHIR\Listener\UuidMappingEventsSubscriber:private] => Array
                                                (
                                                )

                                            [fhirSocialObservationResourcePaths:OpenEMR\ZendModules\FHIR\Listener\UuidMappingEventsSubscriber:private] => Array
                                                (
                                                )

                                        )

                                    [1] => onServicePostSaveEvent
                                )

                        )

                )

            [patient.created] => Array
                (
                    [0] => Array
                        (
                            [0] => Array
                                (
                                    [0] => OpenEMR\ZendModules\FHIR\Listener\UuidMappingEventsSubscriber Object
                                        (
                                            [fhirVitalObservationResourcePaths:OpenEMR\ZendModules\FHIR\Listener\UuidMappingEventsSubscriber:private] => Array
                                                (
                                                )

                                            [fhirSocialObservationResourcePaths:OpenEMR\ZendModules\FHIR\Listener\UuidMappingEventsSubscriber:private] => Array
                                                (
                                                )

                                        )

                                    [1] => onPatientCreatedEvent
                                )

                        )

                )

        )

    [sorted:Symfony\Component\EventDispatcher\EventDispatcher:private] => Array
        (
        )

    [optimized:Symfony\Component\EventDispatcher\EventDispatcher:private] => Array
        (
        )

)
```

Also, I checked the call stack and globals is being required before anything else.

robertdown:
And for reference, https://github.com/providencehealthtech/mod-venom-coding

adunsulag:
@robertdown 

So I pulled your branch, installed your module and it looks like the code_types.inc.php is being included before the module loader.  I added the following code to code_types.inc.php at the top of the file to track down where it was being included:
```
var_dump($GLOBALS['kernel']->getEventDispatcher());
echo "<pre>";
var_dump(debug_print_backtrace());
echo "</pre>";
```

Which gave me the following:
```
object(Symfony\Component\EventDispatcher\EventDispatcher)#140 (3) { ["listeners":"Symfony\Component\EventDispatcher\EventDispatcher":private]=> array(0) { } ["sorted":"Symfony\Component\EventDispatcher\EventDispatcher":private]=> array(0) { } ["optimized":"Symfony\Component\EventDispatcher\EventDispatcher":private]=> array(0) { } }
#0  require_once() called at [/var/www/localhost/htdocs/openemr/src/Services/BaseService.php:26]
#1  include(/var/www/localhost/htdocs/openemr/src/Services/BaseService.php) called at [/var/www/localhost/htdocs/openemr/vendor/composer/ClassLoader.php:480]
#2  Composer\Autoload\includeFile(/var/www/localhost/htdocs/openemr/vendor/composer/../../src/Services/BaseService.php) called at [/var/www/localhost/htdocs/openemr/vendor/composer/ClassLoader.php:346]
#3  Composer\Autoload\ClassLoader->loadClass(OpenEMR\Services\BaseService) called at [/var/www/localhost/htdocs/openemr/src/Services/VersionService.php:15]
#4  include(/var/www/localhost/htdocs/openemr/src/Services/VersionService.php) called at [/var/www/localhost/htdocs/openemr/vendor/composer/ClassLoader.php:480]
#5  Composer\Autoload\includeFile(/var/www/localhost/htdocs/openemr/vendor/composer/../../src/Services/VersionService.php) called at [/var/www/localhost/htdocs/openemr/vendor/composer/ClassLoader.php:346]
#6  Composer\Autoload\ClassLoader->loadClass(OpenEMR\Services\VersionService) called at [/var/www/localhost/htdocs/openemr/interface/globals.php:568]
#7  require_once(/var/www/localhost/htdocs/openemr/interface/globals.php) called at [/var/www/localhost/htdocs/openemr/interface/modules/zend_modules/public/index.php:60]
```
So that include of [line 568 of globals.php](https://github.com/robertdown/openemr/blob/68c498d26eb6564cd8d9f96528ddf4c77e34ed79/interface/globals.php#L568) is using the VersionService.php which does a require_once on the code types before [ModulesApplication runs on line 636](https://github.com/robertdown/openemr/blob/68c498d26eb6564cd8d9f96528ddf4c77e34ed79/interface/globals.php#L636).  

There's a number of ways to fix this.  One would be refactoring BaseService to not include the code_types dependency which could require modifying a number of the service classes to fix that problem, another would be to change VersionService to not extend BaseService and to run its methods manually, there are probably other approaches that would fix this as well.

robertdown:
I removed the Version Service from globals, as best I can tell it's only being leveraged to define $GLOBALS['openemr_version'] which is only referenced in the eRX_xml file. I have modified __that__ file to just reference the VersionService directly. This removes the code_types dependency from globals.php and the event is now firing as expected.

interface/globals.php
```diff
@@ -565,25 +565,6 @@ function GetCallingScriptName()
...
-    }
-
-    $openemr_version = $version['v_major'] . "." . $version['v_minor'] . "." . $version['v_patch'];
-    $openemr_version .= $version['v_tag'] . $patch_appending;
-} else {
-    $openemr_version = xl('Unknown version');
-}
-$GLOBALS['openemr_version'] = $openemr_version;
-
 $srcdir = $GLOBALS['srcdir'];
```
sjpadgett, line 568: $GLOBALS['openemr_version'] is used in portal footer to keep up with core versioning.
Used in `sql_patch.php`,  `portal/patient/templates/_Footer.tpl.php`, `src/Services/ProductRegistrationService.php`

sjpadgett: recm just grabbing from version table and format there in globals.

robertdown:
Ok, querying the version directly in the globals file to ensure `openemr_version` remains in the `$GLOBALS` array

interface/globals.php
```diff
@@ -565,24 +565,10 @@ function GetCallingScriptName()
...
-    $openemr_version = $version['v_major'] . "." . $version['v_minor'] . "." . $version['v_patch'];
-    $openemr_version .= $version['v_tag'] . $patch_appending;
-} else {
-    $openemr_version = xl('Unknown version');
-}
-$GLOBALS['openemr_version'] = $openemr_version;
+$versionSql = "SELECT * FROM `version`";
+$versionRes = sqlQuery($versionSql);
+$versionString = "{$versionRes['v_major']}.{$versionRes['v_minor']}.{$versionRes['v_patch']}";
+$GLOBALS['openemr_version'] = $versionString;
```
bradymiller, line 571: this is not incorporating the `v_tag` or `v_realpatch`, when applicable (see above code)
would also use sqlQueryNoLog() function since will be called on every globals and does not need to be logged

bradymiller: Would it be possible to drop this completely from globals.php (like you planned to do) and use the VersionService in the patient portal instead?

bradymiller: just realize there are several other places. I could probably make quick work of this (migrate all use to VersionService so do not need to build in global.php); will try to make a PR tonight for this.

bradymiller: bye bye $GLOBALS['openemr_version'] :)
https://github.com/openemr/openemr/pull/4809

bradymiller:
Just brought following into codebase to remove the $GLOBALS['openemr_version'] variable:
#4809

custom/code_types.inc.php
```diff
@@ -47,9 +47,14 @@
...
  * @copyright Copyright (c) 2006-2010 Rod Roark <rod@sunsetsystems.com>
  * @copyright Copyright (c) 2019 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2021 Robert Down <robertdown@live.com>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+use OpenEMR\Events\Codes\ExternalCodesCreatedEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+
+require_once dirname(__DIR__) . "/interface/globals.php";
```
bradymiller, line 57: should not need the above line since this script should never be included outside context of globals.php

robertdown:
Ok I resolved the questions. I'm still not 100% convinced my module is behaving as expected but I'm guessing it's something in the module, not in this PR. Going to bring this in.

bradymiller:
2DV

PR #4805, opened 2021-12-24
Wrap Globals save in a transaction.

### LINKED ISSUE #4804: Occasional Globals corruption on save.
### Describe the bug
@adunsulag first saw this a month or so ago when I then encountered a couple days later.
I believe has something to do with session timeout/corruption.
The result appears that data is deleted but never written and seems to happen somewhere in the globals save. 
The symptom after is user can't log in causing user to embark on a mission to overwrite secure password. A frustrating adventure no doubt!
The only solution to issue is import globals from backup or anew from default database.

My solution will be to a least prevent the corruption to table by wrapping global save in a transaction to revert.
This of course relies on an exception, redirect or fatal error that breaks out of the save routine.

sjpadgett:
See issue

Fixes #4804

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -207,7 +207,9 @@ if (array_key_exists('form_save', $_POST) && $_POST['form_save'] && !$userMode) 
 
   // Get all the globals from DB
     $old_globals = sqlGetAssoc('SELECT gl_name, gl_index, gl_value FROM `globals` ORDER BY gl_name, gl_index', false, true);
-
+    // start transaction
+    sqlStatementNoLog('SET autocommit=0');
+    sqlStatementNoLog('START TRANSACTION');
     $i = 0;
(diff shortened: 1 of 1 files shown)


interface/super/edit_globals.php
```diff
@@ -267,6 +269,9 @@ function checkBackgroundServices()
             ++$i;
         }
     }
+    // end of transaction
+    sqlStatementNoLog('COMMIT');
+    sqlStatementNoLog('SET autocommit=1');
```
bradymiller, line 274: nice!

2DV

sjpadgett:
Bringing in. My xmas gift to community!:)

Seriously, @adunsulag recommended awhile back and I concur, we need to go through core and add transactions where appropriate. Namely some of our more DB intensive queries and those with inserts and updates etc.. Will speed up the query by more than a 100%....

PR #4812, opened 2022-01-06
Use BS Alerts instead of gritter

mdsupport:
Replacing package used to display messages by Bootstrap's alerts.

### DIFF
--- a/portal/index.php
+++ b/portal/index.php
@@ -527,7 +527,9 @@ if (!(isset($_SESSION['password_update']) || isset($_GET['requestNew']))) {
         let strongMsg = document.createElement("strong");
         strongMsg.innerHTML = msg;
         divAlert.prepend(strongMsg);
-        setTimeout(()=>{$('.alert').alert('close');}, 5000);
+        setTimeout(()=>{
+            document.querySelector("div.alert").remove();
+        }, 3000);
     }
     </script>
 </body>
--- a/portal/index.php
+++ b/portal/index.php
@@ -401,6 +401,14 @@ if (!(isset($_SESSION['password_update']) || isset($_GET['requestNew']))) {
         </div><!-- div wrapper -->
                 <?php } ?> <!--  logon wrapper -->
 
+<div id="alertStore" class="d-none">
+    <div class="h6 alert alert-warning alert-dismissible fade show my-1 py-1" role="alert">
+      <button type="button" class="close my-1 py-0" data-dismiss="alert" aria-label="Close">
+        <span aria-hidden="true">&times;</span>
+      </button>
+    </div>
+</div>
+
     <script>
         var tab_mode = true;
         var webroot_url = <?php echo js_escape($GLOBALS['web_root']) ?>;
--- a/portal/index.php
+++ b/portal/index.php
@@ -522,9 +522,11 @@ if (!(isset($_SESSION['password_update']) || isset($_GET['requestNew']))) {
         }
 
     function bsAlert(msg) {
-        let htm = $($("#alertStore").html());
-        $('<strong>'+msg+'</strong>').insertBefore(htm.find("button:first"));
-        htm.insertBefore("form");
+        let divAlert = document.getElementById("alertStore").querySelector("div.alert").cloneNode(true);
+        document.querySelector("form").prepend(divAlert);
+        let strongMsg = document.createElement("strong");
+        strongMsg.innerHTML = msg;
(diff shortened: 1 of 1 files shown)


portal/index.php
```diff
@@ -523,6 +520,13 @@ function callServer(action, value, value2, last, first) {
                 alert(message);
             });
         }
+
+    function bsAlert(msg) {
+        let htm = $($("#alertStore").html());
+        $('<strong>'+msg+'</strong>').insertBefore(htm.find("button:first"));
+        htm.insertBefore("form");
+        setTimeout(()=>{$('.alert').alert('close');}, 5000);
```
bradymiller, line 528: Would be really nice to be able to drop the gritter library.
Are you able to use standard js in above lines 525, 526, 528 rather than jquery (we are trying to use standard js in new code rather than jquery)?

mdsupport: Had the developer do this conversion to test productivity impact.

While using standard js is nice objective, project risks introducing unstable code scattered throughout codebase unless a suitable library of jquery like functions or a complete framework is specified.

robertdown: I agree with @bradymiller here, would prefer this be standard JS instead of jQuery. The use of jQuery does not mitigate against unstable code.

mdsupport: 1. As stated, let's stipulate use standard js is nice objective. Then there is matter of limited money and time. Without any suitable alternative, every dev is going to learn yet another way of making these minor changes on our dime. That leads to ..
2. Completely disagree with stability assessment. Any modular, mature development is more stable than new, ad-hoc block of code. Leading towards obsolescence, yes but not unstable. 
3. To be clear, organizations use external components to leverage (free) fruits of efforts of additional developers with free testing and maintenance to boot. Replacing jQ is much bigger lift that jQUI that is embedded here. A bigger and more organized approach is necessary if that is to happen.

sjpadgett: Let's be clear, refactoring jquery with javascript will never happen imo.  Besides, what's the big deal with using jquery. It's just JS wrapped code anyway no diff than any other framework. Are we next going to tackle refactoring Symfony functions to pure php because it's not pure enough?

robertdown: The conversation at hand is not whether or not to refactor every line of jQuery in OpenEMR with vanilla javascript. Instead, I think we should strive to continually embrace current best-practices; if we don't take small steps forward, the codebase doesn't progress.

sjpadgett: Keyword here is 'current'. What is considered best practices today may not be tomorrow and wasn't yesterday perhaps!
Good engineers evaluate and use available tools based on the design goals for a project. Team talent resources play a role helping to determine what is the best practice for a project.

Pure Javascript is most times cumbersome.. This is why we have libraries that wrapper Javascript. While you and I may not have difficulty writing Javascript, there are many that only use jquery.

I thought we decided that if the core code being modified was jquery then we'd continue using jquery for that feature/module. In the end, I agree we should promote Javascript for new code however for consistency, we follow the original design intent when doing modifications.

bradymiller: I pointed this out since it was a new block of code, so figured best to go with pure js.

To simplify this discussion.
For completely new code, should we be using jquery or pure js?

portal/index.php
```diff
@@ -433,23 +441,12 @@ function restoreSession(){
...
             <?php } ?>
             <?php if (isset($_GET['w'])) { ?>
-            var unique_id = $.gritter.add({
-                title: '<span class="red">' + <?php echo xlj('Oops!');?> +'</span>',
-                text: <?php echo xlj('Something went wrong. Please try again.'); ?>,
-                sticky: false,
-                time: '5000',
-                class_name: 'my-nonsticky-class'
-            });
+                // mdsupport - Would be good to include some clue about what went wrong!
```
bradymiller, line 444: From a security perspective, this message should probably be generic no matter the login issue (could make it more helpful by providing a contact option when unable to login though).

bradymiller: Also not sure if telling them to keep trying to login makes sense. Maybe the message should simply state that login was not successful (along with contact option if that is set)

portal/index.php
robertdown: I would say new code should use pure JS;.

bradymiller:
@mdsupport , testing well and going to bring this in. thanks for the contribution! after get this into the codebase, will then completely remove gritter package from OpenEMR's dependencies :)

PR #4813, opened 2022-01-06
Fixed Layout Based Form Validation Failure #4735

### LINKED ISSUE #4735: Layout Based Form Validation Failure
### Describe the bug

An LBF that requires a Text-date field will save with an empty field

### To Reproduce

1. Create a new LBF Layout
2. Add a Group
3. Add a new field
4. Set UOR to Required
5. Set Data Type to Text-date
6. Save form
7. Open form
8. Leave text field blank
9. Save form

### Expected behavior
Form is not submitted

### Client configuration

- Browser: Firefox
- OpenEMR version: Master
- Operating system: Flex Docker

richcow2018:
Fixes #4735

Fixed Layout Based Form Validation Failure

### DIFF
--- a/interface/forms/LBF/new.php
+++ b/interface/forms/LBF/new.php
@@ -658,7 +658,7 @@ if (
                 top.restoreSession();
             }
 
-            return true;
+            return errMsgs.length == 0;
         }
 
         // Called to open the data entry form of a specified encounter form instance.
--- a/library/options.inc.php
+++ b/library/options.inc.php
@@ -700,7 +700,7 @@ function generate_form_field($frow, $currvalue)
         if (!$agestr) {
             echo " title='$description'";
         }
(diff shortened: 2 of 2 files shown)


bradymiller:
It appears this bug is likely happening in all layout based fields that are set as required (tested to happen in both text fields and the date-text).
To see an example of how a required item is supposed to work, recommend making a new patient and leaving the fname, lname, gender, and/or DOB (note this is a text-date field) fields to see how it is supposed to function.
In the layout based fields, will note that it "tries" to do something (field turns red, but then the form gets submitted anyways).

richcow2018:
OK, understand. I will take a look.

bradymiller:
hi @richcow2018 , Testing well and bringing in. Thanks for the fix! -brady

2DV-2

PR #4814, opened 2022-01-06
Rolling back #4767

robertdown:
Rolls back #4767 

The dialog does not actually indicate _what_ the standard fields are, it just says do not delete them. This leads to confusion for the user as they are being told to not do something but are not being given enough information to safely act on the info given. A better approach would be to identify the standard fields identify them on a row-level in the UI. Additionally, some javascript could be used to check if any standard field is checked, if so, disable the delete button. Additionally, the alert clashes with UI changes made in #4744


sjpadgett:
Agree. The array of disallowed id's are in edit_layouts.php L-215
`function isColumnReserved($tablename, $field_id)`

robertdown:
Going to bring this PR in and notate the location of the reserved list and bring the recommendations in at a later date.

PR #4632, opened 2021-09-19
Enhancing Login Page Customizations and Twig Migration

robertdown:
#### Changes proposed in this pull request:
This PR adds customization to the login page by creating 3 new global options in the Appearance section. First, a boolean that will hide the tagline. Second, a textbox to customize the tagline if you wanted (The default is now what the previously hard-coded text was). Third, a boolean that hides the Acknowledgement link. The defaults of these new settings match current state to minimize disruption.

The login.scss file was updated doing away with the gradient behind the logo and instead uses the primary theme color of the theme as defined by the bootstrap theme-colors map.

The entire login.php page was moved over to a Twig template to improve separation of business logic from display logic.

The labels on the username and password were hidden by default using `.sr-only` as they already have a Placeholder attribute, effectively resulting in duplicate directions. 

The login icon was removed from the Login button.

PR #4821, opened 2022-01-08
Rollback PR 913

mdsupport:
Remove changes introduced by PR 913 in login.php and main.php.


bradymiller:
hi @mdsupport , Is this removed since feature no longer works or is helpful? (Also recall there being a list_options that went with this.)

mdsupport:
https://github.com/openemr/openemr/pull/4632 removed way for an user to choose an app. So this PR cleans up extraneous code that is likely to confuse future developers. 

We do use the approach on all mobile devices but not with this code or database records. However @sjpadgett uses a single entry in the list for signature panel. So refrained from touching database records. If needed, we could add those deletions to the PR.

sjpadgett:
@mdsupport I'm no longer using the list so can remove if you wanted. Thanks.

mdsupport:
Done.

robertdown:
This may have been an oversight during the Twig conversion. If there is still a need, we can modify the Twig template to support the App choice. If it's old code no longer being used, then we can remove. @mdsupport let me know which direction you'd like to go

sql/patch.sql
```diff
@@ -47,3 +47,8 @@
 
 --  #EndIf
 --    all blocks are terminated with and #EndIf statement.
+
+#IfRow2D list_options list_id lists option_id apps
+DELETE FROM list_options WHERE list_id="lists" AND "option_id"="apps";
+DELETE FROM list_options WHERE list_id="apps";
+#EndIf
```
sjpadgett, line 54: I'll be removing the `portal/sign/assets/signit.php` script that relied on the app feature in my current portal PR. Just a heads up and no big deal because I deprecated this feature last year.

mdsupport:
Apps is only option if anyone wants a single script to run in authenticated mode but without any tabs - something we find indispensable for simplified interface scripts designed for running on smartphones. Another use we have (with a minor routing change) is unifying portal and emr login to a single screen as part of permitting a single point of authentication for the application.

It is admin choice to keep or remove this option. Accordingly the PR can be closed.

mdsupport:
1. Looks like the login template already made provision for apps. So added the variable to pass select box.
2. Made paths relative to root instead of '/interface' directory to avoid confusion.

interface/login/login.php
```diff
@@ -36,6 +36,85 @@
...
+            'display' => $opt,
+            'value' => $opt,
+            'selected' => 'selected',
+        ];
+    }
+
+    $appOpts[$urlPart[0]] = $opt;
+}
+
+$div_app = '';
```
robertdown, line 89: Please move any HTML creation into a Twig template.

interface/login/login.php
```diff
@@ -36,6 +36,85 @@
...
+    $opt = array_values($appOpts);
+    $div_app = sprintf(
+        '<input type="hidden" name="appChoice" value="%s">',
+        (is_array($opt) ? attr($opt['value']) : attr($opt))
+    );
+} else {
+    $optHtm = '';
+    foreach ($appOpts as $opt) {
+        $optHtm .= sprintf(
+            '<option value="%s" %s>%s</option>',
```
robertdown, line 100: Same as above

interface/login/login.php
```diff
@@ -36,6 +36,85 @@
...
+            (is_array($opt) ? attr($opt['value']) : attr($opt)),
+            (is_array($opt) ? attr($opt['selected']) : ''),
+            (is_array($opt) ? text($opt['display']) : attr($opt)),
+        );
+    }
+    $div_app = sprintf(
+        '<div id="divApp" class="form-group">
```
robertdown, line 107: Same as above

robertdown (changes requested):

sjpadgett:
@bradymiller I wasn't paying attention here. Do I have time to to finish cleaning up with PR Sunday

PR #4825, opened 2022-01-10
Fixes the color dependency infinite loop problem.

adunsulag:
The Node.JS colors developer introduced an infinite loop into the colors
project which breaks this project on dev environments.  We have to lock the
dependency until we figure out a replacement package.

More details can be found here: https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,7 +55,7 @@
       },
       "devDependencies": {
         "autoprefixer": "^10.4.0",
-        "colors": "^1.4.0",
+        "colors": "1.4.0",
         "del": "^6.0.0",
         "glob": "^7.2.0",
         "gulp": "^4.0.2",
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   ],
   "devDependencies": {
     "autoprefixer": "^10.4.0",
-    "colors": "^1.4.0",
+    "colors": "1.4.0",
     "del": "^6.0.0",
     "glob": "^7.2.0",
     "gulp": "^4.0.2",


sjpadgett:
Okay by me. I don't know how we use. I will note that Nodejs < v16 is only version that works with CCDA/QRDA as FYI

bradymiller (approved):

bradymiller:
this looks good to me
(btw, looks like is used in eye form, list editor, and globals editor)

bradymiller:
actually, i am completely wrong. colors is not used in OpenEMR UI. I think it maybe was added by Tyler in past to support the colors in the build output.

bradymiller:
gonna bring this in since gonna do build update (removing gritter)

bradymiller:
@adunsulag , When I update the packages, I am thinking of pinning all the dev js packages (we already do this for the js packages used by openemr). It won't really add any work (since need to check them anyways when do an update) and then no surprises when do updates (for example, when did update to remove gritter, it also updated packages that were not pinned). is this a good idea?

adunsulag:
@bradymiller I think we definitely should pin the packages.  In the past I've always wanted to allow npm packages to have minor updates due to security fixes, but this last problem with the colors.js piece was introduced as a bug fix / security fix and so we just can't rely on auto-updates for security fixes and need to move to a manual vetting.

The downside I've seen with the way node specifically works is that even if we pin our dependencies, our pinned dependencies may pull in sub-dependencies that are not pinned.  From what I can see this means we run the risk of still having this problem on a deeper hierarchical level.  I currently don't know how we can solve that problem outside of having an NPM mirror for OpenEMR repositories which I'm not sure we have bandwidth to manage.

PR #4827, opened 2022-01-12
Sphere payment processing update

bradymiller:
Needed to differentiate between office payments that were taken
over the phone (MOTO) versus office payments that were taken
in person (RETAIL).


bradymiller:
testing well and bringing. also updated instruction set:
https://www.open-emr.org/wiki/index.php/Sphere_Payment_Processing_Module

PR #4831, opened 2022-01-13
Improve GAD-7 form UI

robertdown:
Improves the UI of the GAD7 New.php form. Would like to go ahead and merge and submit a second PR that improves the edit form UI.

### DIFF
--- a/interface/forms/gad7/new.php
+++ b/interface/forms/gad7/new.php
@@ -6,8 +6,9 @@
  * @package   OpenEMR
  * @link      http://www.open-emr.org
  * @author    Ruth Moulton <moulton ruth@muswell.me.uk>
+ * @author    Robert Down <robertdown@live.com>
  * @copyright Copyright (c) 2021 ruth moulton <ruth@muswell.me.uk>
- *
+ * @copyright Copyright (c) 2022 Robert Down <robertdown@live.com>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
--- a/interface/forms/gad7/new.php
+++ b/interface/forms/gad7/new.php
@@ -13,204 +13,182 @@
 
 require_once("gad7.inc.php"); //common strings, require_once(globals.php), other includes  etc
 
-use OpenEMR\Common\Csrf\CsrfUtils;    // security module
+use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Core\Header;
 ?>
 <html>
 <head>
-    <title><?php echo text($string_form_title); ?> </title>
+    <title><?php echo text($string_form_title); ?></title>
     <?php Header::setupHeader(); ?>
+    <script>
+    var no_qs = 8; // number of questions in the form
+    var gad7_score = 0; // total score
+    </script>
+
+    <script src="<?php echo $rootdir;?>/forms/gad7/gad7_javasrc.js?v=<?php echo $GLOBALS['v_js_includes'];?>"></script>
 </head>
 <body class="body_top">
 
+<div class="container">
+
+    <form method="post" action="<?php echo $rootdir;?>/forms/gad7/save.php?mode=new" name="my_form" onSubmit="return(check_all());" >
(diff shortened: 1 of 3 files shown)


bradymiller:
Looks great! Would rec doing both new and edit in the same PR since they should look the same. Otherwise run risk of having different looking forms in distant future if something comes up and it doesn't get completed.
@ruthkonyn will prob want to check out this PR also

interface/forms/gad7/new.php
```diff
@@ -6,211 +6,190 @@
...
+        </div>
+        <div class="form-group row">
+            <div class="col-sm-12 col-md-6 offset-md-6 my-3">
+                <span id="show_gad7_score"><?php echo xlt("Total GAD-7 score"); ?>:</span>&nbsp;<span id="score" class="font-weight-bold"></span>
+            </div>
+        </div>
+        <div class="form-group row mt-3">
+            <div class="col-sm-12 col-md-6 offset-md-6">
+                <button type="submit" class="btn btn-primary btn-save"><?php echo xl("Save Form");?></button>
+                <button type="button" class="btn btn-link btn-cancel" onclick="top.restoreSession();return( nosave_exit());"><?php echo $str_nosave_exit;?></button>
```
bradymiller, line 157: text()

interface/forms/gad7/new.php
```diff
@@ -6,211 +6,190 @@
...
+                    <option selected value="undef" ><?php echo text($str_default); ?></option>
+                    <option value="0"><?php echo text($str_not); ?></option>
+                    <option value="1"><?php echo text($str_several); ?></option>
+                    <option value="2"><?php echo text($str_more); ?></option>
+                    <option value="3"><?php echo text($str_nearly); ?></option>
+                </select>
+            </div>
+        </div>
+        <div class="form-group row d-none" id="q8_container">
+            <label for="" class="col-sm-12 col-md-6 col-form-label"><?php echo $str_q8; ?>.<br><span class="small italic"><?php echo $str_q8_2;?></span></label>
```
bradymiller, line 138: text() for both $str_q8 and $str_q8_2

ruthkonyn:
thanks @robertdown for the improvements, looks good, i apologise for not looking at this sooner (i've been a bit busy). I'd like to run the code and look at the code in detail, and am also checking with our team.  i do intend to do that by wednesday.

robertdown:
No problem, I only bumped it to keep our Stale Bot from closing the issue 🙂

ruthkonyn:
Hi @robertdown - we really like your version of the gad-7 - thanks for making the improvements. There are one or two  comments/suggestions that we'd like to make:

- add a question mark to the leading question 'Over the last 2 weeks, how often have you been bothered by the following problems?'
- the 8th question (that is optional and displays when score >0) - we would prefer if it were separated from the other questions more clearly - e.g. in my version it was under a line, but this isn't the only way to do it.
- we feel it would be clearer if the Total score stood out a little more so that it was more obvious

as @bradymiller says, it would be good (i'd even say necessary) to modify the 'edit' code too so that the two versions of the form are consistent.  I'd be happy to have a go at this if you'd like, although it might be quicker for you... 

but we'll definitely go along with your version and hope it's adopted into the release of OpenEMR soon.

robertdown:
I will add that question mark to the lead sentence. Is there a specific reason you want the 8th question separated? It's hidden unless the score > 0. I can add some padding above the Total score to give it some separation from the questions, but the score is already bolded, there's not much more I would want to do to make it standout.

I'll get the edit form looking the same. The plan will be to point both new and edit functions to the same templates (similar to how the Vitals form manages this). Ideally, the new and edit forms should _always_ be rendered from the same template, reducing duplicated code and increasing consistency.

ruthkonyn:
thanks robert

the  reason we made the suggestion for question 8 is that in many other examples of the gad-7 form question 8 is separated from the earlier questions, at the bottom of the page. This also draws attention to the fact that it's different from questions 1-7 in that it's not compulsory and only relevant if score > 0 - when i tried out your form i was concentrating on the earlier questions so did not notice question 8 quietly appear. maybe even a little padding  would help. 

Of course the text is there that says it's not compulsory, and practioners will be used to this, or become used to it early on in it's use - so this is really not a big deal. As I said, we're very happy with whatever you decide.

PR #4839, opened 2022-01-18
Add new layouts directive to database upgrade

sjpadgett:
#4835 
An easy way to modify layout form edit_options in mass

Add new directive 
``` 
#IfUpdateEditOptionsNeeded
 desc: Change Layout edit options.
 arguments: mode(add or remove) layout_form_id the_edit_option comma_seperated_list_of_field_ids
 ```

There's probably a more elegant way to do this and some further error checking needs to be done but I thought i'd get some input.

### CODE REVIEW

**Comment by sjpadgett** [comment]
Add DAP

image

Remove DAP

image

File: src/Services/Utils/SQLUpgradeService.php [hunk 1]
```diff
@@ -1243,4 +1250,48 @@ private function convertLayoutProperties()
             } // end group
         } // end form
     }
+
+    private function updateLayoutEditOptions($mode, $form_id, $add_option, $values)
+    {
+        $subject = explode(',', str_replace(' ', '', $values));
+
+        $sql = "SELECT `field_id`, `edit_options`, `seq` FROM `layout_options` WHERE `form_id` = ? ";
```
bradymiller, line 1258: would only select the field_id's that are in subject here (this will prevent touching things not related; ie. decrease scope of weird things happening to unrelated entries).

src/Services/Utils/SQLUpgradeService.php
```diff
@@ -1243,4 +1250,48 @@ private function convertLayoutProperties()
...
+        $result = sqlStatementNoLog($sql, array($form_id));
+
```
bradymiller, line 1260: should have a simple flag/mechanism to follow if anything actually gets changed.
then when call this function, if something is changed, return true
and is something is not changed, then return false
then can incorporate this return in the place where function is called to show if the call is "skipped" or not

src/Services/Utils/SQLUpgradeService.php
bradymiller: actually, i see you are filtering for field_id already below, which will also suffice to protect the unrelated entries (ie. disregard my post above :) )

bradymiller:
tenor99

sjpadgett:
image

sjpadgett:
Anybody see a field id I missed for description is placeholder option?

sjpadgett:
Okay boys and girls, bringing in and will await/promise my atta boys and/or promotion from VP of Stuff to SVP of Stuff!

adunsulag:
Why not just make you CTO of stuff?  Go right to the top.

sjpadgett:
I thought that was Brady's job! Besides, i'd forget the title by tomorrow!:)

bradymiller:
will let the masses decide :)
CTO of stuff! CTO of stuff! CTO of stuff!

image

bradymiller:
lol, sent a screenshot on phimail instead of a gif :)
here's the correct picture:

packed-crowd-into-it

sjpadgett:
> lol, sent a screenshot on phimail instead of a gif :)

Well there goes any promotion you are up for!:)

PR #4841, opened 2022-01-19
minor upgrade adjustments for escaping and skipping

bradymiller:
@sjpadgett , thanks for this new upgrade feature! On my review I forgot to point out escaping (this miss in addition to my screenshot mistake merits demotion or a 1 minute ban from github :) ). Adding escaping here. Also changed it to only show verbage if it is not "skipped". testing well.


sjpadgett:
@bradymiller btw: sorry on escape as for some reason I thought our escaping functions weren.t available but that's for somewhere else maybe logs in certain scopes.

PR #4844, opened 2022-01-20
Improved User Group UI

robertdown:
Use `table-sm` on the table for users. Drop the `.alert` div in favor of styling the `<td>` of the password expiration column. Only draw attention to the row if they are in grace or expired.


bradymiller:
2DV-2

PR #4845, opened 2022-01-21
Move Patient Education URLs into a List

robertdown:
Move hardcoded URLs into the list_options table which can be managed by the List UI. Using the notes column as the URL holder.


robertdown:
I stuck with the default 3 search options for now, but this will allow users to further control the resources. @bradymiller This is testing well, but a second set of eyes would be appreciated, want to make sure I got the DB changes completed properly

interface/reports/patient_edu_web_lookup.php
```diff
@@ -28,11 +31,19 @@
...
+$websites = [];
+foreach ($options as $opt) {
+    $websites[$opt['title']] = $opt['notes'];
+}
+
+// $websites = array(
+//     'Medline'   => 'http://vsearch.nlm.nih.gov/vivisimo/cgi-bin/query-meta?v%3Aproject=medlineplus&query=[%]&x=12&y=15',
+//     'eMedicine' => 'http://search.medscape.com/reference-search?newSearchHeader=1&queryText=[%]',
+//     'WebMD'     => 'http://www.webmd.com/search/search_results/default.aspx?query=[%]&sourceType=undefined'
+// );
```
bradymiller, line 46: would drop above commented out lines

interface/super/edit_list.php
```diff
@@ -1238,97 +1230,104 @@ function lister() {
...
             if ($GLOBALS['translate_lists'] && $_SESSION['language_choice'] > 1) {
                 echo "<th class='font-weight-bold'>" . xlt('Translation') . "<span class='help' title='" . xla('The translated Title that will appear in current language') . "'> (?)</span></th>";
             } ?>
-            <th class="font-weight-bold"><?php echo xlt('Style'); ?></th>
-            <th class="font-weight-bold"><?php echo xlt('Force Show'); ?></th>
-            <th class="font-weight-bold"><?php echo xlt('Access Control'); ?></th>
+            <th><?php echo xlt('Style'); ?></th>
+            <th><?php echo xlt('Force Show'); ?></th>
+            <th><?php echo xlt('Access Control'); ?></th>
+        <?php elseif ($list_id == "external_patient_education"): ?>
```
bradymiller, line 1293: Would just place this if below where the Notes label is then only need to use 'URL' instead of 'Notes' at that place (and no need need to touch other colums). Then all the other columns will function like standard lists do (such as translations do) and then it won't break if any changes are made to standard columns.
(also note should use th and not td in initial tag)

sql/database.sql
```diff
@@ -6547,6 +6547,12 @@ INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity) VALUE
...
 INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity) VALUES ('discharge-disposition','oth','Other',110,0,1);
+
+-- External Patient Education
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,option_value) VALUES ('lists', 'external_patient_education', 'External Patient Education', 0, 1, 0);
+INSERT INTO list_options (list_id,option_id,title,notes,seq,is_default,activity) VALUES ('external_patient_education', 'emedicine', 'eMedicine', 'http://search.medscape.com/reference-search?newSearchHeader=1&queryText=[%]', 10, 1, 1);
+INSERT INTO list_options (list_id,option_id,title,notes,seq,is_default,activity) VALUES ('external_patient_education', 'medline', 'Medline', 'http://vsearch.nlm.nih.gov/vivisimo/cgi-bin/query-meta?v%3Aproject=medlineplus&query=[%]&x=12&y=15', 20, 1, 1);
+INSERT INTO list_options (list_id,option_id,title,notes,seq,is_default,activity) VALUES ('external_patient_education', 'webmd', 'WebMD', 'http://www.webmd.com/search/search_results/default.aspx?query=[%]&sourceType=undefined', 30, 1, 1);
```
bradymiller, line 6555: is_default should be 0 in above entried

interface/super/edit_list.php
```diff
@@ -1419,6 +1418,35 @@ function lister() {
...
+
+            for ($i = 0; $i < 3; ++$i) {
+                $results[] = ['line_no' => $opt_line_no];
+                ++$opt_line_no;
+            }
+
+            $vars = [
+                'rows' => $results,
+            ];
+            echo $twig->render('super/lists/external_patient_education/partials/row.html.twig', $vars);
```
bradymiller, line 1448: shouldn't need this block of code. since all doing is changing Notes label to URL label, can just leverage code that is already being used.

sql/database.sql
bradymiller: this also needs to go in the upgrade script

robertdown:
@bradymiller Resolved all of your comments

bradymiller:
2DV-2

PR #4850, opened 2022-01-24
Some portal and general bug fixes

sjpadgett:
portal site id url checks and adds
CarecoordinationTable php empty fixes


stephenwaite:
nice fixins for kkappiah!
southpark_cheering

portal/index.php
```diff
@@ -384,10 +384,10 @@ function validate_new_pass() {
                 <div class="row">
                     <div class="col-12">
                         <?php if ($GLOBALS['portal_onsite_two_register']) { ?>
-                            <button class="btn btn-secondary float-left" onclick="location.replace('./account/register.php')"><?php echo xlt('Register'); ?></button>
+                            <button class="btn btn-secondary float-left" onclick="location.replace('./account/register.php?site=<?php echo urlencode($_SESSION['site_id']); ?>')"><?php echo xlt('Register'); ?></button>
                         <?php } ?>
                         <?php if ($GLOBALS['portal_two_pass_reset'] && isset($_GET['w']) && (isset($_GET['u']) || isset($_GET['p']))) { ?>
-                            <button class="btn btn-danger ml-2" onclick="location.replace('./index.php?requestNew=1')"><?php echo xlt('Reset Credentials'); ?></button>
+                            <button class="btn btn-danger ml-2" onclick="location.replace('./index.php?requestNew=1&site=<?php echo urlencode($_SESSION['site_id']); ?>')"><?php echo xlt('Reset Credentials'); ?></button>
```
bradymiller, line 390: for the above 4 urlencode() calls, would instead use the attr_url() function

portal/index.php
```diff
@@ -484,7 +484,7 @@ function callServer(action, value = '', value2 = '') {
                 data: data
             }).done(function (rtn) {
                 if (action === "cleanup") {
-                    window.location.href = "./index.php" // Goto landing page.
+                    window.location.href = "./index.php?site=<?php echo urlencode($_SESSION['site_id']); ?>" // Goto landing page.
```
bradymiller, line 487: would use the fancy-schmancy js_url() function here:
```
window.location.href = "./index.php?site=" + <?php echo js_url($_SESSION['site_id']); ?>; // Goto landing page.
```

portal/index.php
sjpadgett: nothing wrong with using urlencode!

bradymiller: the attr_url is running urlencode (for the url encoding), but then also runs attr (for the html escaping).

as an aside, from a escaping geek perspective, this is a pretty neat use case where could actually do triple escaping (html/js/url) since this falls within a /html/js/url context. I think it may be overkill in this situation, but it would look something like:
```
<button class="btn btn-danger ml-2" onclick="location.replace('./index.php?requestNew=1&site=' + <?php echo attr_js(urlencode($_SESSION['site_id'])); ?>)"><?php echo xlt('Reset Credentials'); ?></button>
```

portal/index.php
sjpadgett: I don't mind changing....

sjpadgett:
image

sjpadgett:
New directive that allows to require a signature before patient is allowed to submit.

image

sjpadgett:
Time to bring but no doubt once I do someone will find a bug. Any bets?

bradymiller:
2DV

portal/import_template_ui.php
```diff
@@ -701,7 +701,11 @@ function popGroupsDialog() {
                                 }
                                 $next_due = $templateService->showTemplateFromEvent($file, true);
                                 if ($next_due > 1) {
-                                    $audit_status['denial_reason'] = xl('Scheduled');
+                                    if ($audit_status['denial_reason'] === 'In Review') {
+                                        $audit_status['denial_reason'] = xl('Scheduled') . ' ' . xl('but Needs Review');
```
sjpadgett, line 705: this seems silly could have been
 $audit_status['denial_reason'] = xl('Scheduled but Needs Review.')

oh well, hindsight...

PR #4851, opened 2022-01-24
Use external site content for login screen

mdsupport:
PoC to insert login form in external content.

Many of the practices already have public websites. This PoC provides simple changes needed to integrate emr login with that content.

This PR inserts inline login form into open-emr.org landing page. If installed, it will show the potential although some issue that completes the login process for this change will need to be researched.

Fixes #


mdsupport:
The login button leading to the inline form is added to the top right as shown here - 

image

robertdown:
Are you requesting this PR be accepted or just showcasing an idea?

mdsupport:
PR to get feedback about potential issues that others may have encountered - or if they used any different approach. Already had to address some CORS hurdles for open-emr.org content. But something that keeps coming up - specially for patient portal. 

Plan to close this in a day or so.

adunsulag:
One thing to bear in mind if you don't know already is with iOS devices and ITP you have to have a shared TLD if you are going to do cross origin logins that issue a cookie if you plan on staying on your original domain origin.  

If you just plan on having the login screen do a form post directly to the OpenEMR application login (which then takes them either into patient portal or the admin portal) then you're probably find, though I think we still need to use a nonce as part of the login which I didn't see in your PR (maybe I'm not following the flow completely of what you are attempting).

interface/login/login.php
```diff
@@ -252,4 +252,72 @@ function getLanguagesList(): array
...
     'showLabels' => $GLOBALS['show_labels_on_login_form'],
 ];
-echo $t->render("login/login_core.html.twig", $viewArgs);
+
+// mdsupport - Integrate external content
+$publicUrl = "https://www.open-emr.org/";
+$loginHtml = file_get_contents($publicUrl);
+if ($loginHtml) {
+    // Fix during tests
+    $cssFix = "https://www.open-emr.org/style.min.a401b847525db22d3414c418e139dd793430539520659bb5ea12b3f144b3fc65.css";
```
robertdown, line 261: All of this should be in its own Twig file, not in the login.php screen itself.

interface/login/login.php
```diff
@@ -252,4 +252,72 @@ function getLanguagesList(): array
...
+    ';
+} else {
+    $loginHtml = $t->render("login/login_core.html.twig", $viewArgs);
+}
+
+echo $loginHtml;
+
+// Following code relies on browsers moving html stuff inside the body
+?>
+<div id='inline-form-login' class='d-none'>
```
robertdown, line 278: Can't be in this file, has to be in a Twig file

interface/login/login.php
```diff
@@ -252,4 +252,72 @@ function getLanguagesList(): array
...
+          </form>
+          <div class="dropdown-divider"></div>
+          <a class="dropdown-item" href="#">New around here? Sign up</a>
+          <a class="dropdown-item" href="#">Forgot password?</a>
+        </div>
+    </li>
+</div>
+<script>
+$(() => {
+    $('#inline-form-login > li label').addClass('text-white');
```
robertdown, line 320: Use `document.querySelector()` instead of jQuery here.

interface/login/style.min.css
```diff
@@ -0,0 +1,4 @@
+/*!* Bootstrap v4.3.1 (https://getbootstrap.com/)
```
robertdown: Why another CSS file?

interface/login/login.php
```diff
@@ -252,4 +252,72 @@ function getLanguagesList(): array
     'formArea' => $formarea,
     'showLabels' => $GLOBALS['show_labels_on_login_form'],
 ];
-echo $t->render("login/login_core.html.twig", $viewArgs);
+
+// mdsupport - Integrate external content
+$publicUrl = "https://www.open-emr.org/";
```
robertdown, line 257: Needs to be configurable by the user, not hard-coded here

robertdown (changes requested):

robertdown:
I strongly believe this should be created as a module, and not introduced into core.

sjpadgett:
I like this concept but agree with Robert, perfect for a module.
We are currently talking about setting up a custom module install pattern where module code stays with core and is turned on similar to how we do CCM/Zend modules now.

adunsulag:
@robertdown @sjpadgett I'm pretty sure this is a Proof of Concept asking for opinions on how to allow 3rd party websites to have a single click login from the external website into OpenEMR.  I'm guessing there isn't any intention in this current iteration to integrate this PR into the codebase (@mdsupport mentioned taking the PR down in a day or two).  The hard coded examples were all illustrative right @mdsupport?

Is this something that could be exposed as an open-ended API endpoint @mdsupport to grab a login form with a nonce? You could potentially embed it in an iframe with the sandbox attribute and the 'allow-top-navigation-by-user-activation' attribute.  If you expose the API this would be great as a module.

mdsupport:
Basic need here is an external website acting as gateway for emr. It becomes all the more critical for patient portal. If you have control of the external(not emr) site, best to handle an oAuth there and pass token to emr for smooth transition without login.php ever showing up.

The existing login.php is ui for initial setup and then choice of last resort for practices that do not have a web presence.

Thx all.

PR #4852, opened 2022-01-25
Medical Records Dashboard improvements

robertdown:
* Move the delete patient button to the right column, at the bottom.
* Move the unauthorized portal alert to the right column, reduce verbiage
* Create a Patient Portal card on the right side for creating/resetting credentials and assigning documents
* Move the deceased alert to the left column, restyle to draw attention.

Before and after photos below


sjpadgett:
nice Robert! However, can we make the portal panel red, white with blue font color?:)

sjpadgett:
> But here is a red, white, and blue American flag for you :)

Okay, that'd satisfy my zeal for the day. Really though, glad to see all those buttons gone....

juggernautsei:
@sjpadgett thanks for looping me in. I have been watching @robertdown work for a while now. I knew this was coming. All of my telehealth buttons are hardcoded for the section where the portal credentials button is in a row there. I will just have to recode them as what Robert has done. I set up a site with these updates. I will do a new pull to see if anything else has changed before I start planning my client mergers into the new codebase.

robertdown:
> > But here is a red, white, and blue American flag for you :)
> 
> Okay, that'd satisfy my zeal for the day. Really though, glad to see all those buttons gone....
> 
> btw, if I recall someone was using an event dispatch for telehealth or SMS there. Maybe @juggernautsei

Here's a thought...

We could introduce an event for the patient card. Then you could hook into that and we just render out the event details.

robertdown:
I am considering a pretty heavy endeavor. I want the ability to create dynamic Pages which can contain Blocks which can be tied into core with ease. Drupal does it, Moodle does it, it would absolutely elevate our feature set. This allows each practice to create a view specific to their workflow. For instance, the info I need as a nurse is different than what Brady needs as an MD. So my default encounter view would look pretty different.

It's a big lift. And I want to fix a lot of the LBF UI stuff, which is also pretty detailed.

bradymiller:
@robertdown , didn't test but code and screenshots look great!

bradymiller:
tenor99

sjpadgett:
> We could introduce an event for the patient card. Then you could hook into that and we just render out the event details.

Possibly but we have to keep legacy modules in mind. However, if there aren't any events already there then it'd be a go or in addition to.



> I will do a new pull to see if anything else has changed

Okay @juggernautsei but i'd do it sooner than later before it gets too crazy with changes.


> It's a big lift. And I want to fix a lot of the LBF UI stuff, which is also pretty detailed.

I have some thoughts regarding LBF where we'd start by creating a template for each of our LBF datatypes that gets plugged in in our LBF render engine.

robertdown:
I'm going to bring this in and submit the event hooks in a separate PR. I have the events working, I just foresee a bigger convo around implementation

stephenwaite:
there was a styling issue

robertdown:
> there was a styling issue

Can you clarify - I'll fix right now

stephenwaite:
you can see it in the `Checks` tab above

Screenshot from 2022-01-25 15-25-16

PR #4853, opened 2022-01-25
Hook into patient cards

robertdown:
Initial pass at creating events for the Cards found on the patient summary screen. Pass `cardID` to the new Event to define exactly which card is being rendered.

The Event expects new content as an array with the keys `template` and `vars` which represent a Twig template and any specific variables the template needs.

@adunsulag what are your thoughts on the approach to including other twig templates into the event?

If this approach is agreed upon, I can introduce this into all of the cards and move the logic of a lot of the cards itself into a class that adds itself as a listener. Would make the entire page easier to manage.


templates/patient/partials/portal.html.twig
```diff
@@ -13,6 +13,9 @@ The Patient Portal card for the Medical Record Dashboard
     {% if portalAuthorized == false %}
         {{ "Patient Portal Not Authorized"|xlt }}
     {% else %}
+        {% for row in content %}
+            {{ include(row.template, row.vars) }}
```
adunsulag, line 17: so if someone wants to include content from within their module, they would need to add an additional path / file loader to the twig environment right?

interface/patient_file/summary/demographics.php
```diff
@@ -1223,12 +1224,29 @@ function setMyPatient() {
                     endif;
 
                     if ($GLOBALS['portal_onsite_two_enable']):
+                        /** @var EventDispatcher */
+                        $d = $GLOBALS['kernel']->getEventDispatcher();
+                        $e = new CardRenderEvent('portal');
```
adunsulag, line 1229: Are you thinking of firing a CardRenderEvent for each one of the cards?  Portal here being just one example?  Just trying to make sure I'm following the flow here.

interface/patient_file/summary/demographics.php
```diff
@@ -1223,12 +1224,29 @@ function setMyPatient() {
...
+
+                        $result = $d->dispatch(CardRenderEvent::EVENT_HANDLE, $e);
+
                         echo $twig->getTwig()->render('patient/partials/portal.html.twig', [
                             'portalAuthorized' => portalAuthorized($pid),
                             'portalLoginHref' => $portal_login_href,
                             'title' => xl('Patient Portal'),
                             'id' => 'patient_portal',
                             'initiallyCollapsed' => (getUserSetting($id) == 0) ? false : true,
+                            'content' => $result->getContent(),
```
adunsulag, line 1249: Would these be better renamed as 'templates'? When I think of content I'm thinking of string / html content.

src/Events/Patient/Summary/Card/RenderEvent.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $this->card = $card;
+    }
+
+    public function getContent()
+    {
+        return $this->content;
+    }
+
+    public function setContent(array $content)
```
adunsulag, line 54: Do we want to typesafe this with a class to represent the Tuple for template & vars?

interface/patient_file/summary/demographics.php
robertdown: Exactly, each card would have a new CardRenderEvent with the card ID being passed to it. CardRenderEvent is generic to all cards. The portal use is just a proof for the discussion

interface/patient_file/summary/demographics.php
robertdown: Sure, content could be better. Maybe `data` instead? Because `getContent()` returns a template string and potentially variable data

src/Events/Patient/Summary/Card/RenderEvent.php
robertdown: Yep, will update

templates/patient/partials/portal.html.twig
robertdown: Correct, they would need to add whatever path they are using into the loader environment. Which, since the loader is going to get crowded, perhaps we should make it best practice to namespace all templates? That would help easy any future change in filesystem paths (https://twig.symfony.com/doc/3.x/api.html#built-in-loaders).

robertdown:
Also, @adunsulag should we consider [sandboxing](https://twig.symfony.com/doc/3.x/api.html#sandbox-extension) these templates?

robertdown:
@adunsulag Ok, lots of improvements.

* RenderInterface now ensures the Event can access the template file name and variables.
* RenderModel is a RenderInterface compliant class that can be used for simple use-cases. 
* card.macro.twig holds the render function to keep things DRY
* The twig templates are going to be tightly coupled to the RenderInterface (See card.marco.twig)

robertdown:
This is testing well for me. The proof of concept is on the portal card which is working, I haven't tested every card, but I copy/pasted the model from portal, no reason they shouldn't work. Pulling the proof of concept from the code and once the final tests run, I'll merge

templates/patient/partials/testing.html.twig
```diff
@@ -0,0 +1,2 @@
+{# THIS IS FOR POC ONLY - REMOVE BEFORE SUBMITTING PR #}
+{{ var1 }}
```
bradymiller, line 2: hi @robertdown , looks like this should not of been included in the code that went into codebase?

PR #4856, opened 2022-01-26
Vitals Form Improvements

robertdown:
Add the ability to show/hide fields based on a yaml file. Additionally, the order of the fields in the yaml file now dictate the order of fields displayed to the user (on add new and edit forms). The yaml file is buried in the forms folder - however, I included an Options Resolver so if the user wants to customize the display or ordering they can copy the file to `config/forms/form_vitals.yaml` (or .yml) and set the defaults there. Here is the yaml file

```yaml
form_vitals:
    measurements:
        pulse_rate: true
        blood_pressure: true
        respiratory_rate: true
        temperature: true
        oxygen_saturation: true
        oxygen_flow_rate: true
        inhaled_oxygen_concentration: true
        growth_meausrements: true
        height: true
        weight: true
        head_circumference: true
        waist_circumference: true
        body_mass_index: true
        pediatric_weight_percentile: true
        pediatric_bmi_percentile: true
        pediatric_head_percentile: true
```

Set any of those to false and they will be hidden from display on the input form. Move them around, the order changes.

I would love a full, UI-Based config option, but that requires more work and I think this solution solves the problem. Will get to the UI-based config one day :)gi


robertdown:
@sjpadgett thought you might like this

sjpadgett:
@robertdown I am wondering if we couldn't centralize the YAML option file for other forms or is it best to keep it with the forms scope. Even though you're not planning full implementation now, perhaps putting options in library now will save that step later.
Also, you never know whom might pick this up for somewhere else. Um, i've been known to do as much!:)

I do like concept and may be able to remove some duties from globals.

robertdown:
@sjpadgett actually @bradymiller argued pretty strongly against doing any file-based config. So I am working on a different solution that is DB-based

sjpadgett:
Makes sense and agree. I should have learnt my lesson with the portal templates!:)

PR #4859, opened 2022-01-28
Fix percentage discount

### LINKED ISSUE #1805: Percentage discount should be on total amount not on "Form_amount"
https://github.com/openemr/openemr/blob/9e42523c4eaa0c3ff3a420df54d5dc9338391a76/interface/patient_file/pos_checkout.php#L587

surfacehazard:
Fixes #1805

### DIFF
--- a/interface/patient_file/pos_checkout_normal.php
+++ b/interface/patient_file/pos_checkout_normal.php
@@ -580,7 +580,7 @@ function generate_receipt($patient_id, $encounter = 0)
             if ($GLOBALS['discount_by_money']) {
                 $amount  = sprintf('%01.2f', trim($_POST['form_discount']));
             } else {
-                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount/(100-(trim($_POST['form_discount']))));
+                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount / (100 - (trim($_POST['form_discount']))));
             }
             $memo = xl('Discount');
             sqlBeginTrans();
--- a/interface/patient_file/pos_checkout_normal.php
+++ b/interface/patient_file/pos_checkout_normal.php
@@ -580,7 +580,11 @@ function generate_receipt($patient_id, $encounter = 0)
             if ($GLOBALS['discount_by_money']) {
                 $amount  = sprintf('%01.2f', trim($_POST['form_discount']));
             } else {
-                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount / (100 - (trim($_POST['form_discount']))));
+                $form_discount = trim($_POST['form_discount']) ?? 0;           
+                if ($form_discount < 100) {                                
+                    $total_discount = $form_discount * $form_amount / (100 - $form_discount);
+                    $amount = sprintf('%01.2f', $total_discount);
+                }
             }
             $memo = xl('Discount');
             sqlBeginTrans();
--- a/interface/patient_file/pos_checkout_normal.php
+++ b/interface/patient_file/pos_checkout_normal.php
@@ -580,8 +580,8 @@ function generate_receipt($patient_id, $encounter = 0)
             if ($GLOBALS['discount_by_money']) {
                 $amount  = sprintf('%01.2f', trim($_POST['form_discount']));
             } else {
-                $form_discount = trim($_POST['form_discount']) ?? 0;           
-                if ($form_discount < 100) {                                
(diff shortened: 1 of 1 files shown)


interface/patient_file/pos_checkout_normal.php
```diff
@@ -580,7 +580,7 @@ function markTaxes($taxrates)
             if ($GLOBALS['discount_by_money']) {
                 $amount  = sprintf('%01.2f', trim($_POST['form_discount']));
             } else {
-                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount / 100);
+                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount/(100-(trim($_POST['form_discount']))));
```
adunsulag, line 583: I thought the bug was that the discount needs to be on the total amount instead of the form amount.  I'm not seeing how this fixes the original problem as the form_amount is still being used here.  Also there is a potential division by zero error here if form_discount is set to 100 so you'll need to check and make sure that never happens.

@stephenwaite You probably know this file better than I do, is there a better variable that @surfacehazard should use here instead of form_amount (after he's fixed the divide by zero bug).

interface/patient_file/pos_checkout_normal.php
```diff
@@ -580,7 +580,7 @@ function markTaxes($taxrates)
             if ($GLOBALS['discount_by_money']) {
                 $amount  = sprintf('%01.2f', trim($_POST['form_discount']));
             } else {
-                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount / 100);
+                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount / (100 - (trim($_POST['form_discount']))));
```
sjpadgett, line 583: I strongly recommend pre processing any math so you can coalesce arguments. Consider type casting also.

interface/patient_file/pos_checkout_normal.php
surfacehazard: The bug is that the discount is being REapplied to the form_amount on this line. The updated calculation gets you back to the total amount before applying the discount. I can fix the division by zero issue for this code, but it is already broken somewhere else in the file (i.e. if you give a 100% discount, the receipt will show the whole balance outstanding).

surfacehazard: After typing that, thought occurred to me that perhaps no calculation should happen here at all.  `$amount = sprintf('%01.2f', $form_amount);` might do the same thing.

surfacehazard: @adunsulag total amount = discounted amount / (100 - discount percentage) * 100

stephenwaite: not that familiar with it actually since it interferes with insurance billing if it's used for copays as @sunsetsystems notes
```
(4) On checkout, the billing and drug_sales table entries are marked
 *     as billed and so become unavailable for further billing.
```

interface/patient_file/pos_checkout_normal.php
```diff
@@ -580,7 +580,11 @@ function markTaxes($taxrates)
             if ($GLOBALS['discount_by_money']) {
                 $amount  = sprintf('%01.2f', trim($_POST['form_discount']));
             } else {
-                $amount  = sprintf('%01.2f', trim($_POST['form_discount']) * $form_amount / (100 - (trim($_POST['form_discount']))));
+                $form_discount = trim($_POST['form_discount']) ?? 0;           
+                if ($form_discount < 100) {                                
+                    $total_discount = $form_discount * $form_amount / (100 - $form_discount);
```
sjpadgett, line 585: check your math. 10% discount 100.00 charge
100.00*10/100-10 = 1000 / 90 = 11.1

sjpadgett:
umm, looking over more closely I see the discounts are pre calculated in the form.
When submitting the invoice the form amount is already calculated so why the calculations before submit.
Apparently at some point this was changed either in the form or the backend payment post. Looks like we need to factor out the post calculations.
image

sjpadgett:
Never mind. I missed the intent of calculation. Kind of a crazy way to do this and the form post should be passing the total before any calculations for validation reasons.
Anyway, this is fine @surfacehazard sorry for my misunderstanding. However, I do see there needs to be some attention paid to the form styling and validation which I'll probably address in my next fixes PR.

Congratulations on your first PR. We always appreciate and thanks.
Bringing in...

bradymiller:
2DV

PR #4862, opened 2022-01-29
Update ORDataObject persist method

mdsupport:
Use metaColumns function to get field objects.


mdsupport:
Just a suggestion - extensive testing needed since it is probably used by legacy code.

The optional block allows for 'id' in tables like document to be auto_incremented rather than use of global 'sequences'.

PR #4873, opened 2022-02-02
service throws eception instead should be graceful

### LINKED ISSUE #4869: FacilityService Fatal Error
FacilityService getById dies with empty id  

### To Reproduce
Go to demo
Modify demographics
Pick one of the user fields
set type to Facilities
Save
Select a patient
Dashboard is not displayed.

### To Fix
Line 144 of facilityservice can log the error but should return null.

sjpadgett:
Fixes #4869 #4874

removed exception.
added fix for patient delete

### DIFF
--- a/interface/orders/list_reports.php
+++ b/interface/orders/list_reports.php
@@ -207,19 +207,14 @@ function doWait(e){
                     <select name='form_lab_id' id='form_lab_id' class='form-control'>
                         <option value="0"><?php echo xlt('All Labs'); ?></option>
                         <?php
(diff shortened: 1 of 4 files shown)


bradymiller:
2DV

sjpadgett:
will leave open for a bit where i'll probably go a head and add some lab resulting fixes.

sjpadgett:
I changed lab resulting to consider the FS(ascii control char file separator)(end of block for HL7) an end of file so we still result these blocks so users can still see result.

PR #4880, opened 2022-02-06
Bug fixes

### LINKED ISSUE #4878: Rule throws and error when trying to create a new rule
### Describe the bug
When creating a new rule or attempting to update an existing rule with a DB version higher than 5.x of MariaDB or MySQL. 

write a concise description of what the bug is.
https://github.com/openemr/openemr/pull/2928

### To Reproduce

image

1. Select Administration
2. Select practice
3. Select rules
4. Select any rule

### Expected behavior

Work correctly to enter a rule. 

### Client configuration

- Browser:
- OpenEMR version: 6.0.0 (3)
- Operating system: Ubuntu

### LINKED ISSUE #4879: Procedure Configuration Escaping
### Describe the bug
This has bugged me(forgive the pun) for a few months. Started in v6 worked in v5.0.2.
Found because this order `FISH, BURKITT'S/NHL/ALL` wasn't being escaped correctly with text()(would work if echoed direct to html). Failed to escape `'` single quote;

Problem is that the configuration is rendered with ajax call returning a jquery.html() so entire result should be escaped with a js_escape()(json_encode()).

sjpadgett:
consolidated php rendered html and escaped.

Fixes #4879
also Fixes #4878 
Fix procedure configuration escaping.
see issue

### DIFF
--- a/interface/orders/types_ajax.php
+++ b/interface/orders/types_ajax.php
@@ -18,7 +18,9 @@ $id = (isset($_GET['id']) ? $_GET['id'] : '') + 0;
 $order = (isset($_GET['order']) ? $_GET['order'] : '') + 0;
 $labid = (isset($_GET['labid']) ? $_GET['labid'] : '') + 0;
 
-echo "$('#con" . attr($id) . "').html('<table class=\"table\">";
+$render = '';
+
+$render .= "'<table class=\"table\">";
 // Determine indentation level for this container.
 for ($level = 0, $parentid = $id; $parentid; ++$level) {
     $row = sqlQuery("SELECT parent FROM procedure_type WHERE procedure_type_id = ?", [$parentid]);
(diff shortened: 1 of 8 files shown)


juggernautsei (commented):
Jerry, when I implemented this change. 
The view only dumps the SQL out to the screen without rendering the view. Trying to track down the view.

interface/orders/types_ajax.php
```diff
@@ -121,18 +123,21 @@
...
-    echo "</tr>";
+    $render .= "</td>";
+    $render .= "</tr>";
 }
 
-echo "</table>');\n";
-echo "nextOpen();\n";
+$render .= "</table>'\n";
+$rendered = "$('#con" . attr($id) . "').html(";
+$rendered .= js_escape($render) . ");";
```
bradymiller, line 141: 2DV-2

sjpadgett:
> I can't delete this is not an issue it fixed

@juggernautsei The functionality works for me on master. Although appears to be some over escaping causing html to appear in inputs etc.
I don't want to get started fixing CDR as it would require several days.
My concern is the deprecated sql function where the best I can tell works with the SHA1().

sjpadgett:
Actually CDR is pretty nicely done from structural POV and the age of project. Styling needs to catch up. Completely don't understand why bootstrap was specifically excluded in header class! Looks like we have jquery-ui and jquery libraries included in project as separate dependencies directory. Suppose I could maybe do a little work on project. If anything, for peace of mind:)

juggernautsei:
One day I will be able to flow in and out of code as you do.

sjpadgett:
What, trying to take my job there Sherwin.:)

juggernautsei:
Some day. I can dream. It has been said if you don't aim for something. You hit nothing.

sjpadgett:
check it out
image

image

juggernautsei:
That looks better. Did you have to add the container div?

sjpadgett:
actually container-xl so goes 100% width after 1220px breakpoint.

sjpadgett:
bringing in and will do separate PR for rules UI

PR #4891, opened 2022-02-09
Rules UI Restyle

### LINKED ISSUE #4890: Fix CDR Rules UI and restyle
### Describe the problem
Numerous styling issues with the Rules UI and some possible bugs.
I'm not going to completely restyle however try to at least give a better UX than is currently.
Eventually would like to remove the jquery-ui dependency and replace with new sortable drag and drop library. Also rid us of interact.

Updated final restyles:

image

image

image

image

you get the idea....

sjpadgett:
Fixes #4890

Clean up rules engine UI

### DIFF
--- a/interface/super/rules/base/template/basic.php
+++ b/interface/super/rules/base/template/basic.php
@@ -1,34 +1,35 @@
 <?php
 
- // Copyright (C) 2010-2011 Aron Racho <aron@mi-squred.com>
- //
- // This program is free software; you can redistribute it and/or
- // modify it under the terms of the GNU General Public License
- // as published by the Free Software Foundation; either version 2
- // of the License, or (at your option) any later version.
+// Copyright (C) 2010-2011 Aron Racho <aron@mi-squred.com>
+//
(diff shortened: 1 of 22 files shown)


sjpadgett:
@juggernautsei 
I strongly advise, since you're having issues with rules, ya go check all your rules tables for corruptions.
Some item and category labels where returning un-wanted html from the options.inc datatypes where an empty result would return the now annoying fa fa-explanation icon wrapped in a span. The way the rules views render these returned html would populate inputs and thus get saved to tables!

sjpadgett:
Geesh! More involved than I wanted to do but it looks much better and fixed a few bugs too boot..
@juggernautsei will bring in before dev demo reset if you'll jump up on demo farm and test to see if some of your issues are resolved...

juggernautsei:
Sure, I will. LEt me know when

bradymiller:
2DV-2

juggernautsei:
@sjpadgett the changes work well.

PR #4897, opened 2022-02-10
Portal Updates

sjpadgett:
fix chat scroll
remove return home in portal messages

### DIFF
--- a/portal/import_template.php
+++ b/portal/import_template.php
@@ -21,7 +21,7 @@ $patient = json_decode($_POST['upload_pid'] ?? '');
 
 $template_content = null;
 
-if ($_POST['mode'] === 'save_profiles') {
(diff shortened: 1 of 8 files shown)


sjpadgett:
@robertdown I think your last merge broke test!

portal/import_template.php
```diff
@@ -293,7 +293,13 @@ function renderProfileHtml()
     <!DOCTYPE html>
     <html>
     <head>
-        <?php Header::setupHeader(['opener', 'sortablejs']); ?>
+        <?php
+        if (empty($GLOBALS['openemr_version'] ?? null)) {
```
bradymiller, line 297: is this if block meant for rel-601 or master?

sjpadgett: both for now. I want to be able to move portal between versions and not worry with custom to any one version.

PR #4905, opened 2022-02-13
Capture layout display data

mdsupport:
Provide an option to capture display data for a layout.

### DIFF
--- a/library/options.inc.php
+++ b/library/options.inc.php
@@ -3491,12 +3491,16 @@ function extract_layout_data($layoutid, $result1, $result2 = '')
     foreach ($objDoc->getElementsByTagName('td') as $objTD) {
         foreach (['lbl', 'data'] as $chk) {
             $extract = $objTD->getAttribute("data-fld$chk");
-            if (empty($extract)) continue;
-            if (!isset($layout_data[$extract])) $layout_data[$extract] = [];
+            if (empty($extract)) {
+                continue;
+            }
+            if (!isset($layout_data[$extract])) {
+                $layout_data[$extract] = [];
+            }
             // Group messes up label-data pairs. So second label is treated as data.
             // stm, Never overwrite data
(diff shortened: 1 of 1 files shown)


mdsupport:
@sjpadgett, Here is one of the approaches to get label/data pairs for layouts. This allows portal or other scripts to render layouts like any other data.

library/options.inc.php
```diff
@@ -3475,6 +3475,47 @@ function getLayoutProperties($formtype, &$grparr, $sel = "grp_title")
     }
 }
 
+// mdsupport - No easy way to get just display data for a layout.
+// This function attempts to use data-fld to get text data for each field
+// Unlike base function, this function is designed for a single form.
+//
+function extract_layout_data($layoutid, $result1, $result2 = '')
+{
```
sjpadgett, line 3483: Not sure I understand where this gets called from.

mdsupport: @sjpadgett , this PR provides a standardized function to scrape layout data without waiting for it to become responsive.

As an example, currently portal relies on display_layout_rows call in portal/get_profile.php to display demographics. That binds portal's responsiveness to code put out by the standard function. The table/row layout also leads to fields appearing randomly thrown out based on length of contents of label and/or data columns. 

By getting a label/data pair, the caller gets to decide how the layout data will be displayed. If you decide to stick with rows layout, you can now let your own css control the responsiveness. For us, we found masonry layout works best.

Here is an example in desktop mode:
image


Here is that same page with smaller size - notice the insurance data as how the existing table code will show -
image

sjpadgett:
Hi @mdsupport Okay, I get it. I'm looking into where I may want to do some additional refactoring.
I'll get back to you in a day or so. Thanks for the suggestion.

sjpadgett:
Reopening so once I get freed up, I'll take another go at it.

PR #4901, opened 2022-02-11
Portal backport patch4

sjpadgett:
#### Short description of what this resolves:
Bringing some LBF stuff back. Also various events support for telehealth

#### Changes proposed in this pull request:
added sortablejs to portal directory tree.
still testing

below added by bradymiller:
Up For Grab demo for this PR is here: https://www.open-emr.org/wiki/index.php/Development_Demo#Zeta_-_Up_For_Grabs_Demo

PR #4911, opened 2022-02-16
Bring back portal from master to patch 4

sjpadgett:
Porting new portal features to patch 4
This PR replaces PR #4901


sjpadgett:
Still testing from new install through patch 4.

sjpadgett:
Setup new install, patched up to patch3 then applied current patch 4.
Testing well. @bradymiller I'm comfortable with bringing in if you need or can wait another day.

portal/home.php
```diff
@@ -50,7 +52,8 @@
     }
 }
 if ($newcnt > 0 && $_SESSION['portal_init']) {
-    $whereto = $_SESSION['whereto'] = '#secure-msgs-card';
+    $whereto = '#secure-msgs-card';
+    OpenEMR\Common\Session\SessionUtil::setSession('whereto', '#secure-msgs-card');
```
bradymiller, line 56: this doesn't look right unless using this outside the portal for core openemr session. in portal can set sessions (there is no write block) and when use this setSession it will actually create a core (OpenEMR) session.

portal/add_edit_event_user.php
```diff
@@ -487,8 +487,9 @@
     $user = sqlQueryNoLog("SELECT users.username FROM users WHERE authorized = 1 And id = ?", array($_POST['form_provider_ae']));
     $rtn = addPnote($_POST['form_pid'], $note, 1, 1, $title, $user['username'], '', 'New');
 
-    $_SESSION['whereto'] = 'appointmentcard';
-    header('Location:./home.php#appointmentpanel');
+
+    OpenEMR\Common\Session\SessionUtil::setSession('whereto', '#appointmentcard');
```
bradymiller, line 491: same issue as noted below. if these setSession calls are meant for use in portal, then would change these all to direct session changes or else will be creating a new core OpenEMR session:
```
$_SESSION['whereto'] = '#appointmentcard';
```

portal/lib/paylib.php
```diff
@@ -37,7 +37,7 @@
 use OpenEMR\Common\Crypto\CryptoGen;
 
 if ($_SESSION['portal_init'] !== true) {
-    $_SESSION['whereto'] = '#paymentcard';
+    OpenEMR\Common\Session\SessionUtil::setSession('whereto', '#paymentcard');
```
bradymiller, line 40: This is another place where will be creating a Openemr core session inside the portal with this function. This script works in both openemr and portal and performance not an issue. so would directly set the session variables (ie. go back to line that was removed) and then use flag prior to the globals.php include that turns off the session lock write so direct setting of session will also work when this script is used in openemr core.

portal/lib/persist.php
```diff
@@ -38,8 +38,8 @@
 }
 
 if (!empty($data['where'] ?? null)) {
-    $_SESSION['whereto'] = $data['where'];
+    OpenEMR\Common\Session\SessionUtil::setSession('whereto', $data['where']);
```
bradymiller, line 41: same issue as i above paylib.php comment

sql/patch.sql
```diff
@@ -147,3 +147,106 @@ KEY (`success`)
...
+#EndIf
+
+SET @ai_exist =(SELECT COUNT(*) FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = 'document_templates' AND COLUMN_NAME = 'id' AND 'EXTRA' LIKE '%auto_increment%' LIMIT 1);
+SET @query = IF (@ai_exist = 0, 'ALTER TABLE `document_templates` CHANGE `id` `id` BIGINT(21) UNSIGNED NOT NULL AUTO_INCREMENT', '');
+PREPARE statement FROM @query;
+EXECUTE statement;
```
bradymiller, line 185: can't this change just be made directly in the above document_templates table definition?

sql/patch.sql
```diff
@@ -147,3 +147,106 @@ KEY (`success`)
...
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Patient_Groupings','group_3','Group III',30,0,0);
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Patient_Groupings','group_4','Group IV',40,0,0);
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`) VALUES ('Patient_Groupings','group_5','Group V',50,0,0);
+#EndIf
+
+#IfMissingColumn document_templates send_date
+ALTER TABLE `document_templates` CHANGE `exclude_portal` `send_date` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP, CHANGE `exclude_dashboard` `end_date` DATETIME DEFAULT NULL;
+ALTER TABLE `document_templates` CHANGE `profile` `profile` VARCHAR(63) NOT NULL, CHANGE `category` `category` VARCHAR(63) NOT NULL;
+ALTER TABLE `document_templates` DROP INDEX `location`, ADD UNIQUE `location` (`pid`, `profile`, `category`, `template_name`);
+#EndIf
```
bradymiller, line 234: can't this change just be made directly in the above document_templates table schema?

sql/patch.sql
```diff
@@ -147,3 +147,106 @@ KEY (`success`)
...
+
+#IfMissingColumn document_template_profiles member_of
+ALTER TABLE `document_template_profiles` ADD `member_of` VARCHAR(64) NOT NULL;
+ALTER TABLE `document_template_profiles` ADD `active` TINYINT(1) NOT NULL DEFAULT '0';
+ALTER TABLE `document_template_profiles` ADD `recurring` TINYINT(1) NOT NULL DEFAULT '0',ADD `event_trigger` VARCHAR(31) NOT NULL, ADD `period` INT(4) NOT NULL;
+ALTER TABLE `document_template_profiles` CHANGE `profile` `profile` VARCHAR(64) NOT NULL, CHANGE `category` `category` VARCHAR(64) NOT NULL, CHANGE `template_name` `template_name` VARCHAR(255) NOT NULL;
+ALTER TABLE `document_template_profiles` DROP INDEX `location`, ADD UNIQUE `location` (`profile`, `template_id`, `member_of`);
+#EndIf
```
bradymiller, line 242: can't this change just be made directly in the above document_template_profiles table schema?

sql/patch.sql
```diff
@@ -147,3 +147,106 @@ KEY (`success`)
 ) ENGINE=InnoDB;
 #EndIf
```
bradymiller, line 149: for the document_templates and document_template_profiles could simplify by just using same tables from database.sql (assuming those are the most updated tables) and not need to then also then modify them later on in this script

bradymiller:
code review done (didn't test). just couple issues noted.
1. couple database issues
2. setSession issue

portal/home.php
sjpadgett: @bradymiller unsure why I changed this. maybe a senior moment or I was tracking a bug. still, can't be right either way so thanks.

sql/patch.sql
sjpadgett: yes that'd make sense however, I have to track all the pre testing patches I have in the wild so treat this as upgrade...

portal/home.php
sjpadgett: yeah, this all started going back to when I converted the portal to a single page app plus, portal scripts are shared between just a portal session and an openemr session for dashboard.
Brady separated these out when he did the new session manager classes. You're right we had an issue with persisting where patient last viewed across refresh however, I fixed by adding a persist function on menu clicks. But session wasn't being saved correctly thus I added the sessionWrite() which made it work not realizing I was writing to an openemr session. Weird thing with php sessions(not cookies) the whereto session writes still appear in the $_SESSION super as if the same cookie!

Too much for my fragile brain!:)

In the end I had a bug in my persist function where the whereto was immediately being overwritten which caused me to embark on the sessionWrite changes in the first place.

So long story short, this is Brady's fault is what I'm going with.:)

stephenwaite:
Wow!
southpark_cheering

PR #4913, opened 2022-02-16
Openemr fix 4888 document whitelist

### LINKED ISSUE #4888: Secure Direct Messaging fails due to Document whitelisting when a document is transmitted.
Currently there is a bug in the EMRDirect receive process which fails if an unknown mime type is found for a Direct document.  Default OpenEMR installations by default have a restricted set of mime types allowed that is contained in the Administration -> Forms -> Lists -> Files type white list.  However, for receiving Summary of Care documents and other referral type documents (ccr, ccda, xdm, etc) via Direct these files automatically fail as they are not in the allowed mime types.  There is no error reported and the entire direct receive fails using EMRDirect with no ability for a user to diagnose why they are not getting their direct secure emails.

There are a couple ways we can fix this:

- Skip the mime type checking when a file is received via EMRDirect.  @matrix-amiel added the white list code into here so perhaps they can chime into the security ramifications of pursuing this route.  This option assumes that a file received by EMRDirect is safe and can be used inside the EMR.
- Allow the files to be received but put them into some kind of quarantine until they've been reviewed / approved by a user or admin.  This would be necessary if we think that the EMR could be compromised by a file received by a direct user.  OpenEMR doesn't do any virus / file scan checking anyways so going solely off the mimetype can be decieving as a file's mime type can be masked if an attacker so chose to do so.
- Another option is to add specifically the mime types that are required for ONC ETT testing (xml, xdm zip file, etc) to the approved list but that doesn't solve the problem of EMRDirect failing when the mime type is not known.
- Reject the Direct message and send a message to the user or site admin that the file is not in the accepted white list and to review the file in their EMRDirect web portal and if appropriate add the correct mime type to the mime type white list.

The easiest approach will likely be to just skip the mime type filter for EMRDirect and trust that the file is allowed by the EMR which is the approach we are pursuing for certification.  If this is a problem, @matrix-amiel or others please chime in.

adunsulag:
Fixes #4888

I added an event to the sanitize.inc.php::isWhiteFile method called
the IsAcceptedFileFilterEvent that allows event consumers to modify the
whitelist on the fly or at the initial one time function invocation.

I put the event into the direct_message_check.inc::phimail_store
function that adds a listener to whitelist the direct mime type and then
removes the listener once the file has been uploaded.  This ensures that
no further php processing that happens after the function can sneek in a
file of the same mime-type into the system.

It looks like direct message files are already prefixed with a 'dm' in
the filename that's uploaded into OpenEMR so I'm not sure if we need to
do anything else more obvious to dennote this is a Direct file.  We
could stick the file into its own category but I'm not sure if that will
confuse users.

I originally was going to bring in a much bigger PR with a much larger refactor of the direct messaging that includes debugging, more modularity, and a number of enhancements.  However, I realized that was going to slow ONC certification down so I've pushed that code out into its own branch here: https://github.com/adunsulag/openemr/tree/openemr-fix-4888-document-whitelist

If time permits I'll try and come back to that larger branch and bring that code into OpenEMR as I think it will greatly enhance the Direct code (and support vendors other than EMRDirect for CCDA/CDA/CCR sending).


library/direct_message_check.inc
```diff
@@ -31,6 +31,7 @@ require_once(dirname(__FILE__) . "/gprelations.inc.php");
 use OpenEMR\Common\Crypto\CryptoGen;
 use OpenEMR\Common\Logging\EventAuditLogger;
 use PHPMailer\PHPMailer\PHPMailer;
+use OpenEMR\Events\Core\Sanitize\IsAcceptedFileFilterEvent;
```
bradymiller, line 34: alphabetize

library/direct_message_check.inc
```diff
@@ -430,6 +431,9 @@ function phimail_close($fp)
 
 function phimail_logit($success, $text, $pid = 0, $event = "direct-message-check")
 {
+    if (!$success) {
+        (new \OpenEMR\Common\Logging\SystemLogger())->errorLogCaller($event, ['success' => $success, 'text' => $text, 'pid' => $pid]);
```
bradymiller, line 435: place in use block above so can use shortened form

library/direct_message_check.inc
```diff
@@ -517,22 +521,65 @@ function phimail_service_userID($name = 'phimail-service')
...
+    try {
+        // Import the document
+        $phimail_direct_message_check_allowed_mimetype = $mime_type;
+        $filesize = filesize($fn);
+        $return = addNewDocument($name, $mime_type, $fn, 0, $filesize, $user, 'direct');
+        if (is_array($return)) {
+            $return['filesize'] = $filesize;
+        }
+    } catch (\Exception $exception) {
+        (new \OpenEMR\Common\Logging\SystemLogger())->errorLogCaller($exception->getMessage(), ['name' => $name, 'mime_type' => $mime_type, 'fn' => $fn]);
```
bradymiller, line 569: place in use block above so can use shortened form

bradymiller:
2DV

library/direct_message_check.inc
adunsulag: Fixed

library/direct_message_check.inc
adunsulag: Fixed

library/direct_message_check.inc
adunsulag: Fixed

stephenwaite:
southpark_cheering

bradymiller (approved):

PR #4914, opened 2022-02-16
Fixes #4912 direct message patient update fails

### LINKED ISSUE #4912: OpenEMR received Direct messages fail when trying to update patient for message
When attempting to update a message received via Direct, the patient id of 0 is treated as a valid patient id and sent as a patient to interface/main/multi_patient_finders.php file.  The multi select shows a 'false' row which you can't delete or get rid of.  When you attempt to select a patient and hit the OK button the messages screen dies and you have to close the OpenEMR window tab completely (not the browser, the tabbed pane inside OpenEMR) in order to load the messages up again.

Looking over the code I couldn't seeing anything that has changed since @lcmaas originally put the code dealing with patient selection.  Near as I can tell it must be due to a php type error conversion where a pid integer value of 0 would type cast to an empty string '".  I will put up a PR shortly that specifically checks against this condition and fixes the patient selection update for direct messages.

I also noticed that the multi select doesn't give any error messages and fails if you don't hit the buttons in the right order.  I've added in user error messages to help with that also.

adunsulag:
Fixes #4912 Made it so you can update the patient for a direct message.

There's been a lot of hands in the messages.php file and it really needs to be refactored into multiple files due to how many branches and its cyclomatic complexity.  I'd greatly appreciate people taking a review of it to make sure I'm not breaking something else here.

Also added helper error messages to the multi select so you know that you need to add at list one patient to the selection before hitting the ok button.

### DIFF
--- a/interface/main/finder/multi_patients_finder.php
+++ b/interface/main/finder/multi_patients_finder.php
@@ -232,7 +232,7 @@ $('#add-to-list').on('click', function (e) {
     e.preventDefault();
 
     if($('#by-name').val() == '') {
-        alert("<?php echo xls("You must provide a patient name or id to add to the list"); ?>");
+        alert(<?php echo xlj("You must provide a patient name or id to add to the list"); ?>);
         return;
     }
 
@@ -246,7 +246,7 @@ $('#add-to-list').on('click', function (e) {
         if (patient.pid == currentResult.pid) exist = true;
     })
     if(exist){
-        alert("<?php echo xls("This patient has already been added to the list"); ?>");
+        alert(<?php echo xlj("This patient has already been added to the list"); ?>);
         return;
     }
 
@@ -280,11 +280,11 @@ function removePatient(pid) {
 //send array of patients to function 'setMultiPatients' of the opener
 function selPatients() {
     if (!(patientsList && patientsList.length)) {
-        alert("<?php echo xls("You must add a patient to the list before hitting ok"); ?>");
+        alert(<?php echo xlj("You must add a patient to the list before hitting ok"); ?>);
         return false;
     }
     if (opener.closed || ! opener.setMultiPatients) {
-        alert("<?php echo xls('The destination form was closed; I cannot act on your selection.'); ?>");
+        alert(<?php echo xlj('The destination form was closed; I cannot act on your selection.'); ?>);
     } else {
         opener.setMultiPatients(patientsList);
     }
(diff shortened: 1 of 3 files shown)


interface/main/finder/multi_patients_finder.php
```diff
@@ -231,7 +231,10 @@
 $('#add-to-list').on('click', function (e) {
     e.preventDefault();
 
-    if($('#by-name').val() == '')return;
+    if($('#by-name').val() == '') {
+        alert("<?php echo xls("You must provide a patient name or id to add to the list"); ?>");
```
bradymiller, line 235: use xlj instead (note the outside quotes are also removed) (this function basically uses json_encode to escape in js context which adds the quotes):
```
alert(<?php echo xlj("You must provide a patient name or id to add to the list"); ?>);
```

interface/main/finder/multi_patients_finder.php
```diff
@@ -242,7 +245,10 @@
     $.each(patientsList, function (key, patient) {
         if (patient.pid == currentResult.pid) exist = true;
     })
-    if(exist)return;
+    if(exist){
+        alert("<?php echo xls("This patient has already been added to the list"); ?>");
```
bradymiller, line 249: use xlj and drop outside quotes

interface/main/finder/multi_patients_finder.php
```diff
@@ -273,10 +279,15 @@ function removePatient(pid) {
 
 //send array of patients to function 'setMultiPatients' of the opener
 function selPatients() {
-    if (opener.closed || ! opener.setMultiPatients)
+    if (!(patientsList && patientsList.length)) {
+        alert("<?php echo xls("You must add a patient to the list before hitting ok"); ?>");
```
bradymiller, line 283: use xlj and drop outside quotes

library/pnotes.inc
```diff
@@ -495,7 +495,9 @@ function updatePnotePatient($id, $patient_id)
     $activity = $assigned_to ? '1' : '0';
 
     $pid = $row['pid'];
+
     if ($pid != 0 || (int)$patient_id < 1) {
+        (new \OpenEMR\Common\Logging\SystemLogger())->errorLogCaller("invalid operation", ['id' => $id, 'patient_id' => $patient_id, 'pid' => $pid]);
```
bradymiller, line 500: place in use block and use short form

bradymiller:
2DV (1)

adunsulag:
@bradymiller all the suggestions should be fixed now.  Let me know if it all looks good and I'll merge it.

PR #4916, opened 2022-02-17
Openemr fix 4889

### LINKED ISSUE #4889: Embed EMRDirect public certificates into the EMR at the time of installation
Currently users wanting to use EMRDirect for Secure Direct Messaging have to download a certificate and store it in the sites directory of the OpenEMR installation which is cumbersome for less technically savvy users.  In order to make it easier for installation we should download the public & testing certificate at the time of installation.  We should have a global flag to set Direct to use test mode or the production flag.

adunsulag:
Fixes #4889 

Added in the public certificates to EMRDirect.  Made the initial setup for Direct messaging use the test certificate system.  

Added a flag to globals **phimail_testmode_disabled** to turn off test mode and use the production certificate when doing direct.

Added more error logs / debugging for system logger when there are problems.


bradymiller (approved):

bradymiller:
signal-2020-11-19-220815

Issue #4930, opened 2022-02-20
weno_log_sync.php


mdsupport:
@juggernautsei ,
One of the installations has this error. Are you sure globals is required in a library with just a function? Caller of your function should already have globals called.

`require_once(../interface/globals.php): failed to open stream: No such file or directory in xxx/library/weno_log_sync.php on line 11`

sjpadgett:
yep I noticed this also. Plan to fix in my current open PR.

sjpadgett:
@juggernautsei 
Most definitely this functionality wasn't working however even after fatal fix, I have doubts this is working the way you'd expect.
Suggest you carefully test and fix.

juggernautsei:
Ok, I am remembering why I put that there and it was to be called by the background services.

sjpadgett:
I believe you have some closure issues instantiating some of your classes! Probably why you added globals thinking classes weren't auto loading. I may be wrong but again, suggest you fire up and test.

juggernautsei:
You are right.

sjpadgett:
Maybe take a look at your constructors!

juggernautsei:
Thanks for the hint but I have parted ways with this particular project. If anyone would like to pick up the work on this project. Be my guest.

sjpadgett:
huh, that opens up a ton of questions. But, not me for sure. Issue is solved and will close when I bring in my PR. Thanks @mdsupport for taking the time to bring this up.

PR #4932, opened 2022-02-20
Updates concerning CCDA and portal

### LINKED ISSUE #4925: Fix CCD view in patient custom reports.
### Describe the problem
We need to replace the old Rel C32 HL7 CCD download with the current CCD from our ccda service in our patient custom reports. On Hold!
Also, verify CCD generation in portal and whether we have to send via Direct from portal. works now and added a download.

Unsure what to do about old CCR. Leave or remove? Does CCR even apply any longer?

sjpadgett:
add CCD download to portal menu
update some icons
fix CDADocumentService
fix some CCM module CDA weirdness
fix new CCD view in patient report
change Your Documents to Select Documents in portal toolbar

Fixes #4925
Portal inactivity #4933 
Fix for weno log issue #4930 Removed global required.
Add javascript translation capability to portal #4939

### DIFF
--- a/src/Services/CDADocumentService.php
+++ b/src/Services/CDADocumentService.php
@@ -37,6 +37,10 @@ class CDADocumentService extends BaseService
         $this->serverUrl = $GLOBALS['qualified_site_addr'];
     }
 
+    /**
+     * @param $pid
+     * @return array|false|null
+     */
     public function getLastCdaMeta($pid)
     {
         $query = "SELECT cc.uuid, cc.date, pd.fname, pd.lname, pd.pid FROM ccda AS cc
@@ -47,6 +51,10 @@ class CDADocumentService extends BaseService
         return sqlQuery($query, array($pid));
     }
 
+    /**
+     * @param $id
+     * @return false|string
+     */
     public function getFile($id)
     {
         $query = "select couch_docid, couch_revid, ccda_data, encrypted from ccda where uuid=?";
@@ -77,6 +85,10 @@ class CDADocumentService extends BaseService
         return $content;
     }
 
+    /**
+     * @param $pid
+     * @return string
+     */
     public function generateCCDHtml($pid): string
     {
         $url = $this->serverUrl . "/interface/modules/zend_modules/public/encounterccdadispatch";
@@ -100,6 +112,10 @@ class CDADocumentService extends BaseService
         return $response->getContent();
     }
 
+    /**
+     * @param $pid
+     * @return string
+     */
     public function generateCCDXml($pid): string
     {
         $url = $this->serverUrl . "/interface/modules/zend_modules/public/encounterccdadispatch";
@@ -123,6 +139,10 @@ class CDADocumentService extends BaseService
         return $response->getContent();
     }
 
+    /**
+     * @param $pid
+     * @return string
+     */
     public function portalGenerateCCD($pid): string
     {
         $url = $this->serverUrl . "/interface/modules/zend_modules/public/encounterccdadispatch";
@@ -145,6 +165,10 @@ class CDADocumentService extends BaseService
         return $response->getContent();
     }
 
+    /**
+     * @param $pid
+     * @return string
+     */
     public function portalGenerateCCDZip($pid): string
     {
         $parameterArray = array(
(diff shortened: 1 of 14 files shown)


sjpadgett:
Home-OpenEMR-Portal(1)


image

sjpadgett:
inactivity portal log out:

image

sjpadgett:
I think 10 minutes is an appropriate timeout for portal inactivity. Maybe 5 minutes would be better so any suggestions now'd be the time to speak up.
I just don't think we need another global to set however, any disagreement?

sjpadgett:
@bradymiller 
What happened to register and password reset? not displaying.

bradymiller:
@sjpadgett ,
Need to set up google recaptcha for them to work (2 new settings in globals portal section). This will be part of the patch special instructions.

sjpadgett:
Okay thanks but I imagine getting some push back.
I really like your release wiki. Very nicely done!

bradymiller:
Regarding timeout, although hate to make it a global, would do it. Same issue came up with timeout in core OpenEMR (currently defaults to 7200 seconds in globals, which would prob do same on portal); regulations are different everywhere and in different practices, so can't really have a standard timeout time. If clinic's have large forms for patient's to fill out, then a patient may keep going back and forth to it over a long time, so too short a time out would cause issues there.

bradymiller:
if do the global, then would have 0 be no timeout

sjpadgett:
Okay, can do however, don't you think it should be up to the patient? Two hours seems a long time etc.

bradymiller:
definitely go with whatever you think is best for the default (ie. 5 or 10 minutes ok by me) since it can be set to whatever practice wants. Prob would make it a clinic level global setting rather than patient level; since most laypeople don't know what a timeout is (ie. i've never seen this setting in options for bank websites etc.).

src/Services/CDADocumentService.php
```diff
@@ -173,4 +197,40 @@ public function portalGenerateCCDZip($pid): string
...
+            'combination' => $pid,
+            'components' => 'allergies|medications|problems|immunizations|procedures|results|plan_of_care|vitals|social_history|encounters|functional_status|referral|instructions|medical_devices|goals',
+            'downloadccda' => 'download_ccda',
+            'latestccda' => '0',
+            'send_to' => 'download_all',
+            'sent_by_app' => 'core_api',
+            'ccda_pid' => [0 => $pid],
+            'view' => 0,
+            'recipient' => 'self',
+            'site' => $_SESSION ['site_id'],
```
bradymiller, line 220: i think there is a space between $_SESSION and the `[` bracket

sjpadgett: hmm wonder if that would still work

sjpadgett:
An end to another chapter in the continuing saga that is the portal!
portal default timeout = 30 minutes or 15 minutes if user selects under 5 minutes in globals.
a warning is displayed 30 seconds to timeout.

templates/portal/home.html.twig
```diff
@@ -34,6 +36,10 @@
             return true;
         }
 
+        function xl(string) {
```
bradymiller, line 39: isn't this essentially a function that does not do what it is supposed to do :)

would just use xlj via php/twig instead rather than doing this

templates/portal/home.html.twig
```diff
@@ -54,12 +60,52 @@
...
+            window.onmousedown = resetLogOutTimer;
+            window.ontouchstart = resetLogOutTimer;
+            window.ontouchmove = resetLogOutTimer;
+            window.onclick = resetLogOutTimer;
+            window.onkeydown = resetLogOutTimer;
+            window.addEventListener('scroll', resetLogOutTimer, true);
+
+            function logoutWarn() {
+                const warnMsg = xl('Your current log on session will expire in 30 seconds.') +
+                    '<br />' + xl('Please take action to prevent being logged out!');
```
bradymiller, line 81:
```
const warnMsg = {{ 'Your current log on session will expire in 30 seconds.' | xlj }} +
                    '<br />' + {{ 'Please take action to prevent being logged out!' | xlj }};
```
then can remove the xl() stub function above

bradymiller:
2DV-2

templates/portal/home.html.twig
sjpadgett: Nope I did on purpose. I need to have the xl() js so I can use my js utilities alerts etc. especially when I start doing notifications. I planned to add the functionality just not now....

bradymiller: Is it necessary/used in current codebase (if fix below use of the xl)? If not needed, then would just wait until have the functionality. Otherwise the 6.1.0 release will have xl calls that can not be translated.

bradymiller: answered my own question, looks like you are using utility.js function below, so guessing will break if don't have the xl(). We will need to prioritize getting xl() js translation mechanism in portal to work for 6.1.0.

sjpadgett: @bradymiller For an additional 2 atta boys I'll go ahead and add it now.:)
However, you should know me now by now. I actually do come back to stuff..

bradymiller: no prob if want to come back (plan release 6.1.0 in 2 weeks :) )
here are the atta boys though :)

2DV

sjpadgett: Inflation and all. Next week I'll need 3 for any basic work!:)

sjpadgett:
@mdsupport 
Have you added MFA for your portal? If so, what are you using?

sjpadgett:
It seems I'm on a role here. How about allow users to select theme for portal?
Or maybe MFA using e-mail since it should already be setup if users use portal.

templates/portal/home.html.twig
```diff
@@ -28,17 +28,28 @@
...
-
+        
         function restoreSession() {
             return true;
         }
 
-        function xl(string) {
-            return string;
-        }
+        i18next.init({
```
bradymiller, line 40: tenor (67)

sjpadgett:
I'm done! bringing in once actions lets me.... Maybe I'll go pick on something else!

sjpadgett:
@bradymiller I've run actions three times with the phpunit file not loading as a result.
I need to quit for the night but wanted to bring this in before demo resets so, will you bring in later? Thanks.

bradymiller:
@sjpadgett , done :) (i'll bring it over to rel-610 also)

mdsupport:
> MFA for your portal

Sorry No MFA. Just HybridAuth.

Issue #4928, opened 2022-02-19
Patient demographics header


mdsupport:
Demographics header currently takes more than 20-30% of tablet view and offers no additional/usable information for experienced users. While looking at some reorganization, noticed the following:

https://github.com/openemr/openemr/blob/efa882d5cdd7aacfe61afe85bd08a4bf1616fb6f/interface/patient_file/summary/demographics.php#L903

But the class constructor takes no arguments
https://github.com/openemr/openemr/blob/efa882d5cdd7aacfe61afe85bd08a4bf1616fb6f/src/Menu/PatientMenuRole.php#L33

Not sure about the benefits of Twig which is a template engine and this project seems to be adding "templates" that are and can be used only once. It would be great if developers can pass different data to a template and are spared of writing the verbose bootstrap code or having to translate every literal. But hard coding page specific content in a twig or smarty template does not help anyone.

robertdown:
There is inherent benefit in using Twig if for nothing more than separation of business logic from display logic. However, Twig conversion is still relatively early on and as infrastructure continues to improve, a more flexible scheme will unfold, abstracting away a lot of the redundant code.

PR #4935, opened 2022-02-21
Patient Record layout update 1

mdsupport:
Updates patient data display as a follow up to #4928.

1. Patient menu can provide patient record data for use by javascript.
2. The content from nav-link targets is now loaded in a common div eliminating need for each script to render menu.


mdsupport:
This change enables transfer of links from current scripts to single sub-menu.

An example included moves the patient delete button from form to Action -> Delete. Use of forms in dropdowns can also simplify workflow where staff currently is expected to 
1. fetch patient record, 
2. wait for all ajax data to load, 
3. Click edit button, 
4. Clicks the specific tab
5. Updates a field
6. Save the form.

Sample images
image


image

sjpadgett:
@robertdown How does this fit with the refactors you're doing?
I kind of like the idea of making some of the workflows, actions. Anything to speed load times etc...

robertdown:
* During load, the menu gets displayed, then, after the js runs it gets repainted in the final rendering. This is jarring and moves elements on the page which will lead to errors.
* The styling of the page sacrifices the UX of the computer user for the UX of the mobile user. While I appreciate the need for mobile compatibility, we should not trade off the computer user for the mobile user. White space and padding has value (even on mobile)
* The UI of the menu does not flow with the rest of the styling
* Inactive tooltip isn't displaying the right data in my test

`displayHorizNavBarMenu()` hasn't been moved to Twig yet, but if we are going to refactor it, we should go ahead and use a proper Twig template. Other comments in a proper review

interface/patient_file/summary/demographics.js
```diff
@@ -0,0 +1,62 @@
+/* Reformat EMR Patient Record Display
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    MD Support<mdsupport@users.sourceforge.net>
+ * @copyright Copyright (c) 2021-2022 MD Support<mdsupport@users.sourceforge.net>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+*/
+
+$('body')
```
robertdown, line 9: Why is all of this occurring in jQuery and not just added to the HTML class?

robertdown (changes requested):

interface/patient_file/summary/demographics.js
```diff
@@ -0,0 +1,62 @@
...
+.addClass('text-body py-1')
+.removeClass('text-white');
+
+// Replace generic Dashboard label by patient name
+try
+{
+    let objPt = $('#dashboard').data('ptdata');
+    let disp = objPt.lfname;
+    if (objPt.deceased_date !== null) {
+        disp += `<i class="fas fa-exclamation-triangle ml-1 bg-light text-warning" data-toggle="tooltip" title="${objPt.inactive_days}"></i>`;
```
robertdown, line 34: bg-light is breaking the UI here

Issue #4855, opened 2022-01-25
Logout from expired session


mdsupport:
The (in)famous 'Site ID' message was addressed in #1534. 

To reproduce, let session expire or get closed. Then try to use 'Logout' dropdown option from menu.

Noticed that logout.php does get called but user gets no response possibly because changes to auth library resulting in login.php never getting called.

This was a decent compromise to users having to enter emr link in address bar.

mdsupport:
Thought composer/PSR-4 autoloader corruption was the issue. But looks like it is not.

To reproduce in demo setup:
1. Without prior logins, go to : https://one.openemr.io/a/openemr/interface/logout.php
2. You will get no response (because of `PHP Fatal error:  Uncaught Error: Class 'OpenEMR\Common\Logging\SystemLogger' not found in /src/Common/Session/SessionUtil.php:133`)
3. Now login to demo installation
4. Enter exact browser request from Step 1 : https://one.openemr.io/a/openemr/interface/logout.php
5. Logout will be processed and you will get login screen.
6. Now enter same request : https://one.openemr.io/a/openemr/interface/logout.php
7. Again login screen is presented.

Problem with cookie management part of SessionUtil?

Changes in #4938 do resolve the issue although as pointed out by @adunsulag they do not make sense since autoloader does have the class entry.

mdsupport:
Just a suggestion in case someone wants to address this.

Root cause is in this specific situation, SessionUtil is getting called before/without globals including the autoloader.

In our experience autoloader has no preconditions and is almost always the first statement.

Here is the start of our globals.php that resolved this issue.

```
 * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
 */

/**
 * mdsupport - Customization
 * 
 * 1. Centralize standard directory structure pointers based on this __FILE__
 * 2. Make directory variables that are friendly for new developers.
 */
$osProjRoot = dirname(__FILE__, 2);
// Project Directory structure
$osProjInterfaceDir = "$osProjRoot/interface";
$osProjLibraryDir = "$osProjRoot/library";
$osProjPublicDir = "$osProjRoot/public";
$osProjSitesDir = "$osProjRoot/sites";
$osProjPSR4Dir = "$osProjRoot/src";
$osProjVendorDir = "$osProjRoot/vendor";

// Pull in autoloader
require_once ("$osProjVendorDir/autoload.php");

[...]

PR #4938, opened 2022-02-21
Resolve Logger error for destroyed sessions

mdsupport:
No logging possible when script does not have any way to establish db connection.

### DIFF
--- a/src/Common/Session/SessionUtil.php
+++ b/src/Common/Session/SessionUtil.php
@@ -130,10 +130,12 @@ class SessionUtil
 
     public static function coreSessionDestroy(): void
     {
-        (new SystemLogger())->debug(
-            "Destroying core session cookie",
-            ['name' => session_name(), 'cookieParams' => session_get_cookie_params()]
-        );
+        if (class_exists('SystemLogger')) {
+            (new SystemLogger())->debug(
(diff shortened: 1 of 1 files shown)


mdsupport:
Resolves #4855.

src/Common/Session/SessionUtil.php
```diff
@@ -185,10 +185,13 @@ public static function apiSessionStart($web_root): void
 
     public static function apiSessionCookieDestroy(): void
     {
-        (new SystemLogger())->debug(
-            "Destroying api session cookie",
-            ['name' => session_name(), 'cookieParams' => session_get_cookie_params()]
-        );
+        // mdsupport - No current session > No db connection > No log entry
+        if (class_exists('SystemLogger')) {
```
adunsulag, line 189: I'm confused here.  Is the PSR4 autoloader not working here in this class?  How is there a class where the SystemLogger class won't exist?  Also, class_exist is going to look for a a \SystemLogger class, if the line is truly needed I believe you will need to use the following:
```
if (class_exists('OpenEMR\Common\Logging\SystemLogger')) {
```

mdsupport:
Looks like autoloader corruption issue. Composer dump resolved the error.
Thx @adunsulag

PR #4942, opened 2022-02-22
Portal login checkUserName() undefined error quick fix

sjpadgett:
removed as is not needed for login


bradymiller:
2DV

sjpadgett:
was driving me nuts!:)

image

PR #4943, opened 2022-02-22
Add basic E2E test suite and framework

coineagan:
### Test Suite
I added 120 E2E tests. 

Most of these tests focus on iframe navigation. Others test validating required fields in forms and checking for the correct error feedback.

### Test Framework
The goal is to reduce code duplication and the need to manage a lot of Xpaths. There are three folders: ui, pages, and tests.

The ui folder is organized by components and aims to simplify how Xpaths are handled. For example, rather than specifying hundreds of button Xpaths spread out in all of your tests, you could write:

`$button->clickById($session, 'form_save');`

Another common action is opening tabs. Before you would have to specify Xpaths for the navbar menu, and each of the navbar items you have to interact with. To open the users tab would require at least three Xpaths. Now you can write:

`$tab->open($session, 'Administration', 'Users');`

The pages folder is for standardizing actions used in tests. Instead of writing the same code to check for validation errors multiple times in the same test file, the code is written once in the corresponding page file and called upon in the test file. For example:

`$user->findCreateUserValidationError($session, $data, 'Required field missing: Please enter the User Name');`

Files in the test folder are organized by tab or page with the exception of the routes tests. Tests draw from easily readable functions in the ui and pages folders. A basic test to see if the correct tab was opened would look something like:

```
$tab->open($session, 'Administration', 'Users');
$tab->isActive($session, 'User / Groups');
```

### Future Development
The goal is to expand on this to include more complex tests. I wrote some tests that added and removed records from the database, as well as tested the presence of new records in the UI, but I removed them for now to keep things more focused.


adunsulag:
Hey @coineagan awesome PR and great job for a first time PR as well.  We're always excited to see people contribute more to our test cases.  It looks like a couple of your tests are failing on some of our environments.  I don't know that its environment specific though and it could be the case that we've had issues with flaky e2e tests in the past.  Looking at your code you look like you have some expertise in this area and could possibly help us in figuring this out, especially if you are going to contribute more e2e tests in the future.

We do have a styling guide and if you are on docker there is an easy utility to reformat your code to match our checkstyle here: https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md#dev_tools_tests

If you're not on docker we use php_codesniffer and the ruleset is in the ci/phpcs.xml folder in your OpenEMR installation.

bradymiller:
hi @coineagan , agreed with @stephenwaite that very cool to see all these tests! Current ci is running e2e testing 3 times (since was flaky and only couple tests), which I modified a little but in this commit that i just brought into codebase #4945 . So it will then only repeat the testing runs if needed up to 3 times, which will markedly speed things up as long as pass testing. Of course, this doesn't explain why the e2e testing is flaky, which hoping you will have some insight into.

coineagan:
Thank you both for your comments.

And thanks @bradymiller for #4945. That should speed things up a bit.

All of the logs from the different environments should help me figure out why these tests aren't working consistently and maybe if there are any bigger issues.

PR #4946, opened 2022-02-22
Openemr fix #4941 direct send ccda

### LINKED ISSUE #4941: Direct Send does not send messages from Care Coordination Module
At some point the Direct Send from the Care Coordination Module broke.  We need to make it so that the CCDA gets sent via the Care Coordination module.  The original Care Coordination module created a duplicate code pathway for calling into Direct which needs to be consolidated and use the same functions as the existing transmitCCD that the Direct code works.  We need to make sure in refactoring this that we still handle the AMC tracking that occurs in the CareCoordination module.

This is part of the ONC work needed for certification listed in #4887.

adunsulag:
Fixes #4941 

The direct send was not working with the CareCoordination module as it
had a duplicate code of the ccr/transmitCCD.  Working with EMRDirect
found that the Direct community expects one message per patient instead
of bundling multiple patients into a single message per recipient.  I
changed up the code flow to make this happen as well as removed the
duplicate code.  I was able to extend the ccr/transmitCCD.php file to
handle pdf and html files.

Also pulled out some of the constants in the transmitCCD function into
a class with better explanations.  This made the code more legible.

Also documented some of the AMC stuff with the care coordination.

Per request from EMRDirect we add in the version information so they can
have better debugging facilities on their end when OpenEMR has issues.

Note @sjpadgett I refactored the HTML generation in EncountermanagerController to forward the request on to the EncountermanagerTable class since that code was being duplicated and the XSL was not updated when you made your refactor.  From my tests it looks like everything is all working but you may want to just look at those two files and make sure I didn't break anything for you.

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncountermanagerController.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncountermanagerController.php
@@ -14,6 +14,7 @@
 namespace Carecoordination\Controller;
 
 use Application\Listener\Listener;
+use Carecoordination\Model\EncountermanagerTable;
 use DOMDocument;
 use Laminas\Filter\Compress\Zip;
 use Laminas\Hydrator\Exception\RuntimeException;
@@ -24,6 +25,9 @@ use XSLTProcessor;
 
 class EncountermanagerController extends AbstractActionController
 {
+    /**
+     * @var \Carecoordination\Model\EncountermanagerTable
+     */
     protected $encountermanagerTable;
     protected $listenerObject;
 
@@ -140,17 +144,7 @@ class EncountermanagerController extends AbstractActionController
 
     public function buildCCDAHtml($content)
     {
-        $xml = simplexml_load_string($content);
-        $xsl = new DOMDocument();
-        // cda.xsl is self contained with bootstrap and jquery.
-        // cda-web.xsl is used when referencing styles from internet.
-        $xsl->load(__DIR__ . '/../../../../../public/xsl/cda.xsl');
-        $proc = new XSLTProcessor();
-        $proc->importStyleSheet($xsl); // attach the xsl rules
-        $outputFile = sys_get_temp_dir() . '/out_' . time() . '.html';
-        $proc->transformToURI($xml, $outputFile);
-
-        return file_get_contents($outputFile);
+        return $this->getEncountermanagerTable()->getCcdaAsHTML($content);
     }
     public function downloadAction()
     {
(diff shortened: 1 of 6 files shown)


interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncountermanagerController.php
```diff
@@ -140,17 +144,7 @@ public function indexAction()
...
-        // cda.xsl is self contained with bootstrap and jquery.
-        // cda-web.xsl is used when referencing styles from internet.
-        $xsl->load(__DIR__ . '/../../../../../public/xsl/cda.xsl');
-        $proc = new XSLTProcessor();
-        $proc->importStyleSheet($xsl); // attach the xsl rules
-        $outputFile = sys_get_temp_dir() . '/out_' . time() . '.html';
-        $proc->transformToURI($xml, $outputFile);
-
-        return file_get_contents($outputFile);
+        return $this->getEncountermanagerTable()->getCcdaAsHTML($content);
```
sjpadgett, line 147: No problem I can see. As long as the same indexaction from api route are followed which this doesn't bother, my service is good.

sjpadgett (approved):

src/Common/DirectMessaging/ErrorConstants.php
```diff
@@ -0,0 +1,52 @@
...
+ * @copyright Copyright (c) 2022 Stephen Nielson <snielson@discoverandchange.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Common\DirectMessaging;
+
+class ErrorConstants
+{
+    // Message that is sent back to the user (translated) if messaging is unavailable
+    const MESSAGING_DISABLED = "Direct messaging is currently unavailable.";
```
bradymiller, line 17: if do something like this, then would also place the constant in a comment wrapped in xl() (this then will get picked up by the translation scripts when collecting new constants).
```
// xl("Direct messaging is currently unavailable.")
```

src/Common/DirectMessaging/ErrorConstants.php
```diff
@@ -0,0 +1,52 @@
...
+
+    // used in the message sent back to the user when something fails
+    const ERROR_CODE_ABBREVIATION = "EC";
+
+    // translated message sent back to the user when a Direct Address is invalid or message delivery not allowed
+    // Details of the reason is attached to this message (which is not translated).
+    const RECIPIENT_NOT_ALLOWED = "Delivery is not allowed to the specified Direct Address: ";
```
bradymiller, line 24: this constant will need to have the ending space removed to ensure translated in current engine (would place the space in the place where using this constant)

bradymiller:
hi @adunsulag , looks good! just noted couple very minor translation issues.

bradymiller (approved):

stephenwaite:
creepin

PR #4948, opened 2022-02-23
Patient Finder to show deceased patients at end

mdsupport:
Deceased patients to appear at end of find results

### DIFF
--- a/interface/main/finder/dynamic_finder_ajax.php
+++ b/interface/main/finder/dynamic_finder_ajax.php
@@ -64,14 +64,14 @@ $searchMethodInPatientList = isset($_GET['searchType' ]) && $_GET['searchType' ]
 
 // Column sorting parameters.
 //
-$orderby = '';
+$orderby = 'ORDER BY (IF(IFNULL(deceased_date,0)=0, 0, 1)) asc';
 if (isset($_GET['iSortCol_0'])) {
     for ($i = 0; $i < intval($_GET['iSortingCols']); ++$i) {
         $iSortCol = intval($_GET["iSortCol_$i"]);
         if ($_GET["bSortable_$iSortCol"] == "true") {
             $sSortDir = escape_sort_order($_GET["sSortDir_$i"]); // ASC or DESC
             // We are to sort on column # $iSortCol in direction $sSortDir.
-            $orderby .= $orderby ? ', ' : 'ORDER BY ';
+            $orderby .= (!empty($orderby) ? ', ' : '');
             //
             if ($aColumns[$iSortCol] == 'name') {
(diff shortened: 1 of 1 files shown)


bradymiller:
From a workflow perspective, there are times when i need to go into a deceased patient's chart (death certificate info etc or unfinished paperwork). And if no longer alphabetically sorted for that use case, could that confuse users?

mdsupport:
If Doe, Jane is deceased and Doe, John is not, both will still show up during search - except Jane will be towards end. More of UX choice where choices likely to be more relevant in general are presented first. 

Unfortunately no experience or interest in datatables - otherwise a visual indicator like muted row will make it obvious for the users.

sjpadgett:
I had two goals for release 6.1.
- Split Billing
- Archiving Patients

unfortunately I haven't been able to get to them. However for archiving patients I plan to modify patient Finder to search by archived patients so user can view or bring back to active. Would not a deceased patient be archived? or could add as search param while i'm in the datatables. I too find datatables tedious.:)

mdsupport:
Archiving is a big long term solution. But from experience with other systems, it opens a can of worms because project does not implement referential integrity at data level. In any case, most practices would keep these records around until billing is completed. That could be for several months. Malpractice guys will like it there for several more years. During that time, what would be percent of searches where inactive patient would have equal relevance as active patients? In any case, when search returns more than 10-20 results user will find it expedient to keep refining rather than scroll.

Had this mod for years and no complaints.  Did look at putting an asterisk in the name but have no time in addressing any issues other parts could create.

Ideally the project should support concept of 'inactive' patient. We currently use translation as a workaround and treat deceased_date as date when inactivity started. As an example, if patient requested transfer of medical records, doesn't that record have same retention requirement as a deceased patient?

sjpadgett:
I guess archive is a misnomer on my part. Inactive would be a better way to put it. The way I plan to do it is not really a big lift!
It's a matter of removing the inactive patient from views and reports. Is that not the goal of marking patient inactive. Or I may be wrong not having much clinical experience.
All I know is that this has been asked for in the community for years and the only reason we haven't done it up to this point is we have to touch a lot of core.

robertdown:
I agree with Brady re: the workflow and usability concern. We know deceased and inactive patients are causing performance issues and @sjpadgett seems to have a plan. This PR would be negated by those improvements and could cause confusion to the end user in the intermediate.

sjpadgett:
I am going to take this on now. Hopefully in next day or two I'll put up a working PR so you guys can give input and guidance.
I really would like to get this feature into next release.

mdsupport:
@sjpadgett , In your design please consider adding and using the active flag to patient_data as specified in [FHIR model](https://build.fhir.org/patient-definitions.html). Also specifies link between patient.active and patient.deceased.

PR #4952, opened 2022-02-23
Promote Patient Finder to top-level menu

robertdown:
Move the Patient Finder menu item from underneath Patient/Client to the top level as the second menu item.

### DIFF
--- a/interface/main/tabs/menu/menus/standard.json
+++ b/interface/main/tabs/menu/menus/standard.json
@@ -16,6 +16,18 @@
     ]
   },
   {
+    "label": "Patient Finder",
+    "menu_id": "fin0",
+    "target": "fin",
+    "url": "/interface/main/finder/dynamic_finder.php",
+    "children": [],
+    "requirement": 0,
+    "acl_req": [
+      "patients",
+      "demo"
+    ]
+  },
+  {
     "label": "Flow Board",
     "menu_id": "pfb0",
     "target": "flb",
@@ -64,18 +76,6 @@
     "label": "Patient/Client",
     "menu_id": "patimg",
     "children": [
-      {
-        "label": "Patients",
-        "menu_id": "fin0",
-        "target": "fin",
-        "url": "/interface/main/finder/dynamic_finder.php",
-        "children": [],
-        "requirement": 0,
-        "acl_req": [
-          "patients",
-          "demo"
-        ]
-      },
       {
         "label": "New/Search",
         "menu_id": "new0",
--- a/interface/main/tabs/menu/menus/standard.json
+++ b/interface/main/tabs/menu/menus/standard.json
@@ -16,7 +16,7 @@
     ]
   },
   {
-    "label": "Patient Finder",
+    "label": "Finder",
     "menu_id": "fin0",
     "target": "fin",
     "url": "/interface/main/finder/dynamic_finder.php",
@@ -28,7 +28,7 @@
     ]
   },
   {
-    "label": "Flow Board",
+    "label": "Flow",
     "menu_id": "pfb0",
     "target": "flb",
     "url": "/interface/patient_tracker/patient_tracker.php?skip_timeout_reset=1",
@@ -44,7 +44,7 @@
     ]
   },
   {
-        "label": "Recall Board",
+        "label": "Recalls",
         "menu_id": "pfb0",
         "target": "rcb",
         "url": "/interface/main/messages/messages.php?go=Recalls",
@@ -73,7 +73,7 @@
     ]
   },
   {
-    "label": "Patient/Client",
+    "label": "Patient",
     "menu_id": "patimg",
     "children": [
       {
@@ -753,7 +753,7 @@
     ]
   },
   {
-    "label": "Administration",
+    "label": "Admin",
     "menu_id": "admimg",
     "children": [
       {
--- a/interface/main/tabs/menu/menus/standard.json
+++ b/interface/main/tabs/menu/menus/standard.json
@@ -1294,14 +1294,6 @@
           "admin",
           "super"
         ]
-      },
-      {
-        "label": "About",
-        "menu_id": "abo0",
-        "target": "msc",
-        "url": "/interface/main/about_page.php",
-        "children": [],
-        "requirement": 0
       }
     ],
     "requirement": 0,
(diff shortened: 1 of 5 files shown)


sjpadgett:
The top level menu is already over crowded. With Brady adding back the About page and several features turned on then we'll have to start collapsing menu on laptops views or col-xl. I recm turning on all possible globals that appear in top level and move some if needed.
Also user can turn on patient finder as one of their two default tabs if finder use is needed to be prominent.
anyway, this is 1280px width. I don't think we break till 1200 for some reason. That needs to be fixed. Also, there is too much padding used on top menu items. We currently need to collapse at 1280px.

image

bradymiller:
Also, a quick way to get to patient finder is to click the search box

robertdown:
The patient finder is probably a high-traffic screen, I think we should really consider making it top-level. It's arguably as important in a clinical workflow as the calendar, flow board, and recall board (if not more important than the recall board)

robertdown:
Also, agree there is too much padding. I'll work on reducing the padding (and maybe nudging down the font-size of the menu a bit)

robertdown:
Possible idea would also be to have this menu customizable by user and by role (more so than it already is).

sjpadgett:
> The patient finder is probably a high-traffic screen,

Oh I agree Robert! I'm just pointing out something needs to be done because either we fix the collapse breakpoint to break at 1280px or some items have to go from top level. There are a lot of folks working with 1280x800 views and it becomes a poor UX when menu is collapsed as there default view. I don't know what the tradeoff should be.

sjpadgett:
> (and maybe nudging down the font-size of the menu a bit)

is already too small on light theme.

bradymiller:
Maybe change:
`Flow Board` to `Flow`
`Recall Board` to `Recall`
`Patient Finder` to `Finder`
`Patient\Client` to `Patient`
`Administration` to `Admin`
`Miscellaneous` to `Misc`

robertdown:
I think all of those are good options (and low-hanging fruit). The only bummer is we lose the translation constants.

adunsulag:
Can we use the CSS text-overflow: ellipsis property on the menu and give the menu items a constrained width?  If we make sure the full title is in the aria and title properties this may be good enough as they can still see the titles on hover?  That way you don't lose translations.

robertdown:
Do we really have a large user base using 1200x768? Wouldn't 1280x768 be a more likely resolution? Or even 1366x768? We can also look at hiding the menu logo on smaller screens

sjpadgett:
> Wouldn't 1280x768

That's what I was pointing out. Currently we collapse at 1200px should be 1280px.

robertdown:
Ok, I reduced the X-padding down some (Probably can't reduce it anymore and maintain readability), dropped "Board" from the menu items and here is a screenshot on a 1200x768 view

image

sjpadgett:
@bradymiller Did you already move the About page? Did you want that on top level?
Also Robert you may want to move newcrop or turn on to check menu with it on. I don't know of any other features that get on top level if feature is turned on besides Therapy Groups.

robertdown:
Think we could put the About link under the User menu? That way it is still reachable by everyone but is not top level

bradymiller:
Under User menu would work also since will always show up as option (will be clear where it is after User uses OpenEMR a couple times; would prob need to call it `About OpenEMR` so User does not think it is about themselves :) ) when user logs out (goal is to make it as accessible as possible since it has a donate button).

bradymiller:
also, looks like all tests are failing; guessing a simple issue in menu change.

sjpadgett:
Maybe could also stick it under provider pull for logout and password reset. It'd get seen there a lot...

robertdown:
**IMPORTANT** - I committed on this branch branch and pushed it before realizing my mistake; I was on the wrong branch. A commit which was intended for https://github.com/openemr/openemr/pull/4955 was accidentally committed to this PR.

I did a local reset and did a force push to return this PR to the proper state.

robertdown:
Dropped user's name from the menu. Here is the new User menu option.

image

interface/main/tabs/templates/user_data_template.php
```diff
@@ -18,17 +18,24 @@
...
+                        <li class="menuLabel">
+                            <span class="font-weight-bold"><span data-bind="text:fname"></span> <span data-bind="text:lname"></span></span>
+                        </li>
+                        <li class="menuLabel" data-bind="click: editSettings"><i class="fa fa-fw pr-2 text-muted fa-cog"></i>&nbsp;<?php echo xlt("Settings");?></li>
+                        <li class="menuLabel" data-bind="click: changePassword"><i class="fa fa-fw pr-2 text-muted fa-lock"></i>&nbsp;<?php echo xlt("Change Password");?></li>
+                        <li class="menuLabel" data-bind="click: changeMFA"><i class="fa fa-fw pr-2 text-muted fa-key"></i>&nbsp;<?php echo xlt("MFA Management");?></li>
+                        <div class="dropdown-divider"></div>
+                        <li class="menuLabel" data-bind="click: function() {navigateTab('/interface/main/about_page.php', 'About');activateTabByName('msc');}"><i class="fa fa-fw pr-2 text-muted fa-info"></i>&nbsp;<?php echo xlt("About");?> <?php echo $GLOBALS['openemr_name']; ?></li>
```
bradymiller, line 35: text() on $GLOBALS['openemr_name']

interface/main/tabs/menu/menus/standard.json
```diff
@@ -16,7 +16,19 @@
     ]
```
bradymiller, line 16: Would remove the About menu item in this menu (it is in Admin)

interface/main/tabs/menu/menus/standard.json
```diff
@@ -16,7 +16,19 @@
...
   },
   {
```
bradymiller, line 18: When this file is ready, also should copy it to sites/default/documents/custom_menus/Custom.json
(this is basically a copy of standard.json, which folks can use as a starting point to create their own custom menus)

bradymiller:
screen shot of user menu is beautiful!

2DV-2

bradymiller:
Looks like all function e2e tests are failing at `Check add user` test. 
should be able to fix this by changing `Administration` to `Admin` at line 13 of tests/Tests/E2e/Pages/MainPage.php

PR #4953, opened 2022-02-24
Keyboard shortcut support

robertdown:
Not the most extendable solution, but this is a start of supporting keyboard shortcuts. For now it support `alt+1` (Go to calendar), `alt+2` (Go to Patient Finder), and `alt+shift+w` (Clear the patient). Would like to also support `alt+r` as a refresh of the current tab, but was having some issues getting that working.

I think this is a good starting point. Some basic navigation. Eventually I'd love to have more robust support but for now, let's run with this.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -243,3 +243,7 @@ assets:
     sortablejs:
       basePath: '%assets_static_relative%/sortablejs/'
       script: Sortable.min.js
+    hotkeys:
+        basePath: '%assets_static_relative%/hotkeys-js/'
+        script:
+            - dist/hotkeys.min.js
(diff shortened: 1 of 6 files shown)


sjpadgett:
I like this and is something I thought was lacking when I first was introduced to openemr.

bradymiller:
looks very cool! would document this on a quick new wiki page so it does not get lost (guessing folks will be adding to this over time) and will be a very nice new feature in next release

bradymiller (approved):

robertdown:
Will definitely get a wiki page up

stephenwaite:
excited_elephant

PR #4954, opened 2022-02-24
Fixes #4888 Direct Message Whitelist

### LINKED ISSUE #4888: Secure Direct Messaging fails due to Document whitelisting when a document is transmitted.

adunsulag:
Fixes #4888 

I discovered that the document white listing fails if the local
operating system's mime type detection on the document is
different than the mime type received via direct. This was
discovered with a CCDA xml file. EMRDirect was sending a mime type
of application/xml mime type and the local OS detected it as
text/xml. This breaks our current solution as we temporarily allow
the EMRDirect mime type if it matches the document mime type and
then we remove the temporary bypass.

Since we don't want to build a file quarantine process I went
with our original design approach and bypass the mime type check
altogether during the document storage process.
The files are protected on the drive as is so this should be fine.

### DIFF
--- a/library/direct_message_check.inc
+++ b/library/direct_message_check.inc
@@ -592,8 +592,6 @@ function phimail_store($name, $mime_type, $fn)
 {
     global $phimail_direct_message_check_allowed_mimetype;
 
-    (new SystemLogger())->debug("Attempting to store phimail attachment in OpenEMR", ['name' => $name, 'mime_type' => $mime_type]);
-
     $allowMimeTypeFunction = 'phimail_allow_document_mimetype';
     // we bypass the whitelisting JUST for phimail documents
     if (isset($GLOBALS['kernel'])) {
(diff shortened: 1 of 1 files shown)


bradymiller (approved):

bradymiller:
2DV

PR #4955, opened 2022-02-24
Normalize buttons, improve readability of patient name

robertdown:
A bit more work on the recent changes to the patient banner bar. Since the new encounter button is not the primary action button, swapped over to `btn-secondary`, moved the View History button to btn-secondary as well to remain consistent. Instead of the btn-info on the patient name, I dropped the button styling (It's more of a secondary navigation element versus a button with action) and turned it into an `<h3>` - this promotes it, making it easier to see (and, by extension, easier to click). The Close Patient button had a btn-warning class on it, which we need to reserve for critical actions or for things that need our attention (think the way we display the delete patient button).

We should really strive to avoid using red, yellow, and orange colors in places that do not require immediate clinician attention. I have seen many charts (not in OpenEMR) that have full-width, high-impact color background banner alerts. One or two grabs my attention, but it's not uncommon to see 7 of them stacked on each other... it gets to be overwhelming.

image

### DIFF
--- a/interface/main/tabs/templates/patient_data_template.php
+++ b/interface/main/tabs/templates/patient_data_template.php
@@ -59,11 +59,12 @@ switch ($search_any_type) {
             </div>
             <div class="form-group">
                 <!-- ko if: patient -->
-                <a class="ptName btn btn-small btn-info" data-bind="click:refreshPatient,with: patient" href="#" title="<?php echo xla("To Dashboard") ?>">
-                    <span data-bind="text: pname()"></span>
-                    (<span data-bind="text: pubpid"></span>)
-                </a>&ensp;
-                <a href="#" class="btn btn-sm btn-warning" data-bind="click:clearPatient" title="<?php echo xla("Clear") ?>">
+                <h4 class="d-inline">
+                    <a class="ptName" data-bind="click:refreshPatient,with: patient" href="#" title="<?php echo xla("To Dashboard") ?>">
+                        <span data-bind="text: pname()"></span>&nbsp;<small class="text-muted">(<span data-bind="text: pubpid"></span>)</small>
+                    </a>
+                </h4>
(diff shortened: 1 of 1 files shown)


stephenwaite:
I prefer to see buttons but understand you feel really strongly about color overload. How about neutral buttons for the name and clear?

robertdown:
Here is an updated screenshot. I moved the close button to the left of the name, this way it is in the same place, no matter the size of the name. I also added some margin to the bottom of the buttons to give the DOB and Encounter text a bit of breathing room.

@stephenwaite would you be ok with a new globals section called "Patient Banner" where we can add an option to have the name displayed as a button or large text? There's actually a LOT of potential uses in the patient banner that would need customizations. I'll put together a quick POC for some of the other things the PBB could be used for.

image

robertdown:
Just to get this out of my head and on "paper" here is a direction the PBB could go. All these data points are pretty important in daily workflow for inpatients, I'm sure there's outpatient data points as well.

image

robertdown:
Ok, there's now a global option to manage the display. Three options exist.

1. `As Text Link` will display with essentially no styling.
2. `As Large Text Link` will display as an `<h3>` element
3. `As Button` will display as a button group

Here are the corresponding screenshots

image

image

image

robertdown:
I also swapped the badge on the messages from red to blue and removed the parenthesis for that badge

robertdown:
I also declared the patient photo height and width in the actual image tag which helps with sizing the box before the image loads. It's already 75x75, but by explicitly defining it, the image element can be painted before the image actually loads

stephenwaite:
nice! although the uniformity of the clear X is better it makes the user go across the name to get to it, also the X could be called close instead of clear

robertdown:
I'll move the X back and update the title to "Close Patient Record" from "Clear"

bradymiller:
2DV

stephenwaite:
yes-gif-5

Issue #4964, opened 2022-02-26
ONC § 170.315(b)(1) testing issues and results


sjpadgett:
### Describe the problem

A place to track b(1) testing.
SMTP Test 8, 14 and 18 Send pass.

sjpadgett:
Appears CCM auto Import from Direct export is working for CCD.
![image](https://user-images.githubusercontent.com/1859985/155860884-3f9c3393-471b-4294-a383-255f7ddae907.png)

sjpadgett:
Yeah! Good work @adunsulag for a Georgia boy!:)

![image](https://user-images.githubusercontent.com/1859985/155906497-e11b6ab1-a316-481e-8ea4-15e0b2fa9343.png)

sjpadgett:
Send test working except we're not giving the exported CCDA xml a name it appears. I have a couple IG conformance errors dealing with name prefix and suffix data types.

![image](https://user-images.githubusercontent.com/1859985/155910866-fe2e4710-b198-41d7-acf8-190b3516ecc9.png)

@bradymiller and @adunsulag , I say we can call b(1) complete. I'm going to do a quick run at h(1) since we've changed some things with our Direct account since Stephen tested.

sjpadgett:
@adunsulag 
Looks like we do have to use two different usernames for testing.
b(1) testing uses openemr-edge2015@directtest.interopengine.com the  where h(1) requires us to use openemr@test.directproject.net sandbox username.

![image](https://user-images.githubusercontent.com/1859985/155912991-2662e9d0-d706-4fdb-a5ca-b0ae31c45853.png)

I'll leave to you if we want to add additional globals for usernames and hook in code if Direct is in dev mode. Otherwise we'll need to change as we test. However, I did a few of the same testing you've done with good results.

bradymiller:
@sjpadgett , @adunsulag ; Hopefully can make the globals more centric towards productions users, who will likely just need to turn the service on and provide a username and password.
Would probably turn off the test mode by default.
And is the phiMail Server Address needed (ie. is there a separate one for production and testing that can be hard-coded or does it need to be flexible where need to enter it in like we have it?)

sjpadgett:
We're leaving globals alone. b(1) and h(1) are green. Closing this.

PR #4971, opened 2022-02-27
Direct message empty content text/plain and text/html rejected

### LINKED ISSUE #4970: Direct Send with empty message body fails with M7 Error
For ONC certification on (b)(1) we need to be able to handle a Direct message that has a subject but an empty body with attachments.  Right now it just returns an M7 error message (which @lcmaas) introduced for better error reporting.

sjpadgett:
See issue.
Basically we would reject Direct mime types text/plain and text/html received message with empty body.

Fixes #4970 #4964

### DIFF
--- a/library/direct_message_check.inc
+++ b/library/direct_message_check.inc
@@ -68,11 +68,11 @@ function phimail_connect(&$phimail_error)
         case "tcp":
         case "http":
             $server = "tcp://" . $phimail_server['host'];
-              $phimail_secure = false;
+            $phimail_secure = false;
             break;
         case "https":
             $server = "ssl://" . $phimail_server['host']
-               . ':' . $phimail_server['port'];
+                . ':' . $phimail_server['port'];
             break;
         case "ssl":
         case "sslv3":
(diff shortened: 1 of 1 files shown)


sjpadgett:
I think we need to address resolving incoming CCDA documents a little better.
Looks like we store empty messages as docs. Need to address.
Need to inform user to go to CCM to import if CCDAs. Maybe a link.
For now making mods to get through § 170.315(b)(1) testing. Going well so far and Direct seems to be working very well once you unscramble setting up!:)

Still trying to understand our Direct directory in sites/documents.

bradymiller:
Hopefully you don't need to understand how stored in sites/default or else there is likely an issue with the standard mechanism to get the document. This is because getting a document should be abstracted out (ie. collecting the document via the standard mechanism to get the document will get you a unencrypted form of the document no matter how it was imported and whether it is stored on couchdb instead).
fyi, if it is stored to drive (and not couchdb), then it will be stored in `sites/*/documents/direct/<random-directory-from-0-10000/<uuid-name-of-document>`

sjpadgett:
> Hopefully you don't need to understand how imported direct documents are stored in sites/*/documents

I'm pretty sure we have an issue here however I didn't look at too closely and maybe not. On one test I wasn't able to open a html document after moving to patient and category.

bradymiller:
Another thing i should of mentioned is that documents never get moved anymore on the drive. So they stay wherever they initially get imported. And changes to document name, patient, or category are just lightweight changes (ie. changes in the sql entry in the documents table).

sjpadgett:
> Still trying to understand our Direct directory in sites/documents.

Retested this a.m and appears documents are being moved to patient and category correctly when done from documents.

sjpadgett:
Testing well. Bringing in when can.

image

sjpadgett:
@adunsulag Any questions before I bring in?

adunsulag:
Thanks @sjpadgett It all looks great!

PR #4972, opened 2022-02-28
make SessionUtil class self-sufficient

### LINKED ISSUE #4855: Logout from expired session

bradymiller:
fixes #4855

@adunsulag , this seemed like the simplest solution since this class is called before the class autoloader in scripts that need to support both portal and core OpenEMR. Since not critical functionality and just debugging stuff figured easier to remove it than complicate code by needing to bring in class autoloader in these instances.

### DIFF
--- a/src/Common/Session/SessionUtil.php
+++ b/src/Common/Session/SessionUtil.php
@@ -3,6 +3,9 @@
 /**
  * start/destroy session/cookie for OpenEMR or OpenEMR patient-portal or OpenEMR oauth2
  *
+ * Note that keeping this class self-sufficient since it is used before the class autoloader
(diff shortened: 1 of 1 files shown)


adunsulag:
@bradymiller This is fine, someday I would probably like to switch this to some kind of class factory mechanism that swaps out the SessionUtil for a SessionUtilDebug class or something so we can trace session problems when we have them.

adunsulag (approved):

PR #4978, opened 2022-03-04
complete e3

bradymiller:
complete e3

yep, all of this to simply support a "clickable link" in pnotes to pass e3 :)
https://www.open-emr.org/wiki/index.php/Patient_health_information_capture_(MU3))

### DIFF
--- a/library/pnotes.inc
+++ b/library/pnotes.inc
@@ -531,3 +531,13 @@ function deletePnote($id)
     sqlStatement("UPDATE pnotes SET deleted = '1', update_by = ?, update_date = NOW() WHERE id=?", array($_SESSION['authUserID'], $id));
     return true;
 }
+
+function patientNoteConvertLinks($note) {
+    $noteActiveLink = preg_replace('!((https://)[-a-zA-Z()0-9@:%_+.~#?&;//=]+)!i', '<a href="$1" target="_blank" rel="noopener">$1</a>', $note);
+    if (empty($noteActiveLink)) {
+        // something bad happened (preg_replace returned null) or the $note was empty
+        return $note;
+    } else {
+        return $noteActiveLink;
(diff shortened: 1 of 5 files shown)


library/pnotes.inc
```diff
@@ -531,3 +531,14 @@ function deletePnote($id)
     sqlStatement("UPDATE pnotes SET deleted = '1', update_by = ?, update_date = NOW() WHERE id=?", array($_SESSION['authUserID'], $id));
     return true;
 }
+
+function patientNoteConvertLinks($note)
```
bradymiller, line 535: will change this name to something smaller like pnoteConvertLinks

library/pnotes.inc
```diff
@@ -531,3 +531,14 @@ function deletePnote($id)
...
+{
+    $noteActiveLink = preg_replace('!((https://)[-a-zA-Z()0-9@:%_+.~#?&;//=]+)!i', '<a href="$1" target="_blank" rel="noopener">$1</a>', $note);
```
stephenwaite, line 537: squirrel

adunsulag, line 537: Do we have anywhere in the system a standard way for link sanitation? I'm guessing we don't care about supporting unicode characters for external links right (which is a good way to cut down on a lot of spam links) but will create problems for language support.

bradymiller: @adunsulag , no standard on this and yep what i am using is pretty much as simple as possible and english character centric. Figured that is safest way to proceed since this is supposed to make active links to support onc2015 and likely will be a minimally used feature. We can definitely expand on it to be more flexible down road if needed.

bradymiller:
bringing this in, now on to g7 :)

stephenwaite:
jumping

PR #4986, opened 2022-03-06
removal of ^M line endings

bradymiller:
hi @kchapple and @sjpadgett ,
The ^M line endings ended up in composer.json and all the mustache files. I only addressed composer.json in this PR (this is something we needed to address a long time ago since was causing issues and will figure out how to do the mustache files later). Ensure repos are correctly set for line endings here: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings (if this is set correctly, then the ^M character won't be included in the commits)


stephenwaite:
endoftheline

PR #4990, opened 2022-03-08
QRDA cat i measure section rendering

kchapple:
Renders the measure section (which measures are being run) to the QRDA cat i XML document export.

```
<section>
        <!-- 
          *****************************************************************
          Measure Section
          *****************************************************************
        -->
        <!-- This is the templateId for Measure Section -->
        <templateId root="2.16.840.1.113883.10.20.24.2.2"/>
        <!-- This is the templateId for Measure Section QDM -->
        <templateId root="2.16.840.1.113883.10.20.24.2.3"/>
        <!-- This is the LOINC code for "Measure document". This stays the same for all measure section required by QRDA standard -->
        <code code="55186-1" codeSystem="2.16.840.1.113883.6.1"/>
        <title>Measure Section</title>
        <text>
          <table border="1" width="100%">
            <thead>
              <tr>
                <th>eMeasure Title</th>
                <th>Version specific identifier</th>
              </tr>
            </thead>
            <tbody>
              <tr>
                <td>Percentage of children 2 years of age who had four diphtheria, tetanus and acellular pertussis (DTaP); three polio (IPV), one measles, mumps and rubella (MMR); three or four H influenza type B (Hib); three hepatitis B (Hep B); one chicken pox (VZV); four pneumococcal conjugate (PCV); one hepatitis A (Hep A); two or three rotavirus (RV); and two influenza (flu) vaccines by their second birthday</td>
                <td>2C928085-7198-38EE-0171-9D6E75580676</td>
                <td/>
              </tr>
            </tbody>
          </table>
        </text>
        <!-- 1..* Organizers, each containing a reference to an eMeasure -->
        <entry>
          <organizer classCode="CLUSTER" moodCode="EVN">
            <!-- This is the templateId for Measure Reference -->
            <templateId root="2.16.840.1.113883.10.20.24.3.98"/>
            <!-- This is the templateId for eMeasure Reference QDM -->
            <templateId root="2.16.840.1.113883.10.20.24.3.97"/>
            <id extension="480fe5cf-20d2-4865-85bf-5f951af4066c" root="1.3.6.1.4.1.115"/>
            <statusCode code="completed"/>
            <!-- Containing isBranch external references -->
            <reference typeCode="REFR">
              <externalDocument classCode="DOC" moodCode="EVN">
                <!-- SHALL: This is the version specific identifier for eMeasure: QualityMeasureDocument/id it is a GUID-->
                <id extension="2C928085-7198-38EE-0171-9D6E75580676" root="2.16.840.1.113883.4.738"/>
                <!-- SHOULD This is the title of the eMeasure -->
                <text>Percentage of children 2 years of age who had four diphtheria, tetanus and acellular pertussis (DTaP); three polio (IPV), one measles, mumps and rubella (MMR); three or four H influenza type B (Hib); three hepatitis B (Hep B); one chicken pox (VZV); four pneumococcal conjugate (PCV); one hepatitis A (Hep A); two or three rotavirus (RV); and two influenza (flu) vaccines by their second birthday</text>
                <!-- SHOULD: setId is the eMeasure version neutral id  -->
                <setId root="B2802B7A-3580-4BE8-9458-921AEA62B78C"/>
              </externalDocument>
            </reference>
          </organizer>
        </entry>
      </section>
```

### DIFF
--- a/src/Services/Qrda/Helpers/Cat1View.php
+++ b/src/Services/Qrda/Helpers/Cat1View.php
@@ -152,15 +152,19 @@ trait Cat1View
         }
 
         $oid = $result['system'] ?? $result['codeSystem'];
-        $system = $this->get_code_system_for_oid($oid);
-        if (!empty($result['code'])) {
-            return "<value xsi:type=\"CD\" code=\"" . $result['code'] . "\" codeSystem=\"" . $oid
-                . "\" codeSystemName=\"" . $system . "\"/>";
-        } elseif (!empty($result['value'])) {
-            return "<value xsi:type=\"PQ\" value=\"" . $result['value'] . "\" unit=\"" . ($result['unit'] ?: "UNK") . "\"/>";
+        if (!empty($oid)) {
+            $system = $this->get_code_system_for_oid($oid);
+            if (!empty($result['code'])) {
+                return "<value xsi:type=\"CD\" code=\"" . $result['code'] . "\" codeSystem=\"" . $oid
+                    . "\" codeSystemName=\"" . $system . "\"/>";
+            } elseif (!empty($result['value'])) {
+                return "<value xsi:type=\"PQ\" value=\"" . $result['value'] . "\" unit=\"" . ($result['unit'] ?: "UNK") . "\"/>";
+            } else {
+                // TODO: @sjpadgett, @adunsulag, @ken.matrix the ruby code didn't handle this case... what happens here?
+                // no result so template shouldn't show.
+                return "";
+            }
         } else {
-            // TODO: @sjpadgett, @adunsulag, @ken.matrix the ruby code didn't handle this case... what happens here?
-            // no result so template shouldn't show.
             return "";
         }
     }
--- a/src/Services/Qrda/Helpers/View.php
+++ b/src/Services/Qrda/Helpers/View.php
@@ -14,6 +14,7 @@ namespace OpenEMR\Services\Qrda\Helpers;
 
 use Mustache_Context;
 use Ramsey\Uuid\Rfc4122\UuidV4;
+use function Clue\StreamFilter\fun;
 
 trait View
 {
@@ -21,12 +22,42 @@ trait View
 
     public function measures()
     {
-        // TODO: we don't know if this is the correct implementation or not, depends on how the measures are sent.
-        $hqmf_id = $this->_measures['hqmf_id'] ?? null;
-        $hqmf_set_id = $this->_measures['hqmf_set_id'] ?? null;
-        $description = $this->_measures['description'] ?? null;
-        return json_encode(['hqmf_id' => $hqmf_id, 'hqmf_set_id' => $hqmf_set_id, 'description' => $description]);
+//        // TODO: we don't know if this is the correct implementation or not, depends on how the measures are sent.
+//        $hqmf_id = $this->_measures['hqmf_id'] ?? null;
+//        $hqmf_set_id = $this->_measures['hqmf_set_id'] ?? null;
+//        $description = $this->_measures['description'] ?? null;
+//        return [
+//            'hqmf_id' => $hqmf_id,
+//            'hqmf_set_id' => $hqmf_set_id,
+//            'description' => $description
+//        ];
+
+        return array_map(function ($measure) {
+            return [
+                'hqmf_id' => $measure['hqmf_id'] ?? null,
+                'hqmf_set_id' => $measure['hqmf_set_id'] ?? null,
+                'description' => $measure['description'] ?? null
+            ];
+        }, $this->_measures);
+
+        //return $this->_measures;
     }
+
+//    public function hqmf_id(Mustache_Context $context)
+//    {
+//        $hqmf_id = $context->find('hqmf_id');
+//        return $hqmf_id;
+//    }
+
+//    public function hqmf_set_id(Mustache_Context $context)
+//    {
+//        return $context->find('hqmf_set_id');
+//    }
+//
+//    public function description(Mustache_Context $context)
+//    {
+//        return $context->find('description');
+//    }
 
     public function random_id()
     {
--- a/src/Services/Qdm/MeasureService.php
+++ b/src/Services/Qdm/MeasureService.php
@@ -2,6 +2,9 @@
 
 namespace OpenEMR\Services\Qdm;
 
+use GuzzleHttp\Psr7\LazyOpenStream;
+use GuzzleHttp\Psr7;
+
 class MeasureService
 {
     public static $measure_sources = [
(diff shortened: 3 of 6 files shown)


kchapple:
@sjpadgett Here's the code that pulls in the measure section, for whatever measure you are running against. Currently in our tool, you can only do one, but this also supports many. 

Still have no clue what the external document reference is within the measure sections. Maybe that's a Cat iii doc?

src/Services/Qrda/Helpers/Cat1View.php
```diff
@@ -152,15 +152,19 @@ public function result_value_as_string($result)
...
         $oid = $result['system'] ?? $result['codeSystem'];
-        $system = $this->get_code_system_for_oid($oid);
-        if (!empty($result['code'])) {
-            return "<value xsi:type=\"CD\" code=\"" . $result['code'] . "\" codeSystem=\"" . $oid
-                . "\" codeSystemName=\"" . $system . "\"/>";
-        } elseif (!empty($result['value'])) {
-            return "<value xsi:type=\"PQ\" value=\"" . $result['value'] . "\" unit=\"" . ($result['unit'] ?: "UNK") . "\"/>";
+        if (!empty($oid)) {
+            $system = $this->get_code_system_for_oid($oid);
+            if (!empty($result['code'])) {
```
sjpadgett, line 157: This won't work in many resulting cases(labs).  There are results like urine color 'YELLOW' e.t.c. that will not have an oid.

src/Services/Qrda/Helpers/Cat1View.php
```diff
@@ -152,15 +152,19 @@ public function result_value_as_string($result)
...
+                return "<value xsi:type=\"CD\" code=\"" . $result['code'] . "\" codeSystem=\"" . $oid
+                    . "\" codeSystemName=\"" . $system . "\"/>";
+            } elseif (!empty($result['value'])) {
+                return "<value xsi:type=\"PQ\" value=\"" . $result['value'] . "\" unit=\"" . ($result['unit'] ?: "UNK") . "\"/>";
+            } else {
+                // TODO: @sjpadgett, @adunsulag, @ken.matrix the ruby code didn't handle this case... what happens here?
+                // no result so template shouldn't show.
+                return "";
+            }
```
sjpadgett, line 166: This prob will never happen but I left it in anyway.

src/Services/Qrda/Helpers/View.php
```diff
@@ -21,11 +21,17 @@ trait View
...
+        return array_map(
+            function ($measure) {
+                return [
+                    'hqmf_id' => $measure['hqmf_id'] ?? null,
+                    'hqmf_set_id' => $measure['hqmf_set_id'] ?? null,
+                    'description' => $measure['description'] ?? null
+                ];
+            },
+            $this->_measures
+        );
```
sjpadgett, line 34: I assume this will filter what QDMs get reported in patient section. If not, we need to filter populating the patient section with only data appliable to the measure(s) being reported.

src/Services/Qrda/Helpers/Cat1View.php
kchapple: If it's allowed to be empty, then we can't pass an $oid to get_code_for system, and the rest of this code needs to be checked. It was crashing my system.

src/Services/Qrda/Helpers/View.php
kchapple: No that's not how it works. The QDMs for patient are all populated (diagnosis, observations, allergies, etc), then you run the measures based on the data. At that point we don't even really know what data we'll need top calculate the measures. 

You don't then remove patient data that wasn't needed to calculate the measures.

The measure section contains one or more measures that are being reported, and lists their Description and ID, and the results are presumably contained in linked "external documents" which may be Cat iii, I'm not sure, may need to get clarification from CMS.

src/Services/Qrda/Helpers/Cat1View.php
sjpadgett: Then I guess filter the oid with null coalesce or condition as long as a valid result is not filtered out. This came up several times in import.

src/Services/Qrda/Helpers/View.php
sjpadgett: I'm pretty sure. The patient data section should only be populated with the data relative to the measure being reported. Not all possible qdms.
My understanding is that the qdms are resolved based on the measures QDEs and are generally only positive results.

```
eMeasure Data Elements
• eMeasures are comprised of the data elements that are
required to compute them
• Identify QDEs in the referenced eMeasure(s)
• Example: Medication Administered - Aspirin
Find matching data elements in a patient chart
• Specification calls this “scooping”
• Example: Find all medications administered for a patient
Filter the data to the relevant elements
• Example: Only place administrations of Aspirin in the
QRDA- I document
• This happens by matching clinical codes
• Specification calls this “smoking gun”
Repeat for all QDEs in the referenced
eMeasure(s)
```

I don't know what would happen in testing the way we have now though, I'm with you, why not send all.

Perhaps we need a phone call to clarify with each other...

sjpadgett: From a Lantana doc for MU2 available at https://ecqi.healthit.gov/system/files/qrda_basics_08_12_2015_a_508.pdf.
It just gives an outline on QRDA.
I'm looking over the IG now.

sjpadgett: @kchapple If  I read this correctly then appears you may be correct. From HL7 IG
```
6.2.3 How Much Data Should be Sent?
QDM-based QRDA adheres to a "scoop and filter" philosophy, whereby data from an 
EHR are scooped up and filtered, and the remaining content is packaged into the 
instance.
When the recipient of the instance has access to no other EHR data, it is important that 
the instance include data elements relevant to computing eCQM criteria, as well as the 
other data elements defined in an eCQM —for stratification, for risk adjustment, etc. 
Every data element present in the EHR that is required by the referenced eCQM(s), not 
just those needed to compute criteria, shall be included in the QRDA document.
The EHR may have more data than are relevant to the referenced eCQM(s) and more 
data than are needed to compute the criteria. For instance, a patient who has been in 
the Intensive Care Unit undergoing continuous blood pressure monitoring will have 
reams of blood pressure observations. QDM-based QRDA adheres to a "smoking gun"
philosophy where, at a minimum, the conclusive evidence needed to confirm that a 
criterion was met shall be included in the instance.
At the very least, the QRDA document should include:
• For each data element in each referenced eCQM, smoking gun data that offer 
confirmatory proof, where a patient has met the criterion—For disjunctive criteria
Page 45 HL7 CDA R2:QRDA I, Release 1, STU Release 5 US Realm, Vol. 1 — Introductory Material
© 2017 Health Level Seven, International. All rights reserved. December 2017
(i.e., where a criterion can be satisfied by one of multiple data elements) include 
minimal smoking gun data for at least one data element.
• Stratification variables, supplemental data elements, risk adjustment variables, and 
any other data element specified in the referenced eCQM(s)
A quality program implementing QRDA will often provide prescriptive guidelines that 
define additional data, outside the smoking gun, that may or must be sent (such as the 
complete problem or medication list). Where such prescriptive guidelines exist, those 
take precedence over the more general guidance provided here. In other words, the 
“smoking gun” heuristic ensures that the minimum is present in the QRDA, and does 
not preclude inclusion of additional data.
```

kchapple: This is my takeaway:

> 
The EHR may have more data than are relevant to the referenced eCQM(s) and more data than are needed to compute the criteria.

kchapple: To me, it sounds like they need "at least" the relevant data, which they call the "smoking gun", not ONLY the relevant data.

sjpadgett: agreed Ken. going by Cypress sample docs they only include 'smoking gun' which makes it easier to eval the measure visually. Our html version of the doc will look very similar across patients and measures however, if ONC/CMS/HL7 doesn't care then I suppose we shouldn't either.

Still need our organizational/author/addresses/phone etc in our context.
CCM has all these queries if want to reuse them. Unsure how you want to populate in patient data class/context.

I'm working on Cat I export.

PR #4998, opened 2022-03-13
Changing the order from ascending to descending

juggernautsei:
Fixes #

### DIFF
--- a/interface/billing/billing_tracker.php
+++ b/interface/billing/billing_tracker.php
@@ -94,7 +94,7 @@ use OpenEMR\OeUI\OemrUI;
                     { "data": "created_at" },
                     { "data": "updated_at" },
                 ],
-                "order": [[4, 'asc']] // Order by 'Date Created' with newest first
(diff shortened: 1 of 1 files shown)


juggernautsei:
bump

juggernautsei:
here we are again. Keep at it right?

interface/billing/billing_tracker.php
```diff
@@ -94,7 +94,7 @@
                     { "data": "created_at" },
                     { "data": "updated_at" },
                 ],
-                "order": [[4, 'asc']] // Order by 'Date Created' with newest first
+                "order": [[4, 'desc']] // Order by 'Date Created' with newest first
```
bradymiller, line 97: does the comment need to change?

juggernautsei: Yes it does.

juggernautsei:
This is already in the codebase.
https://github.com/openemr/openemr/blob/52773322af5a42b2622f0ce285433be6a1247323/interface/billing/billing_tracker.php#L97

PR #5001, opened 2022-03-14
Feature Build to import modules

juggernautsei:
https://youtu.be/s4993FNcObI

This feature allows none technical users of the program to install modules. A developer can publish a module on the OpenEMR site and provide a link to install a module. 

The owner of the /custom_modules folder needs to be changed to www-data.

### DIFF
--- a/interface/modules/zend_modules/module/Installer/src/Installer/Controller/ModuleImport.php
+++ b/interface/modules/zend_modules/module/Installer/src/Installer/Controller/ModuleImport.php
@@ -1,0 +1,102 @@
+<?php
+
+/*
+ *  package OpenEMR
+ *  link    https://www.open-emr.org
+ *  author  Sherwin Gaddis <sherwingaddis@gmail.com>
+ *  Copyright (c) 2022.
+ *  license https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace Installer\Controller;
+
+
+/**
+ * the purpose for this class is to allow users to import modules to the system with no technical knowledge
+ * All they need is a zip file location from the module author
+ */
+class ModuleImport
+{
+    private $url;
+    private $name;
+    public $importDir;
+
+    /**
+     * @param $url
+     * @param $name
+     * @param $import_dir
+     */
+    public function __construct($url, $name, $import_dir)
+    {
+        $this->url = $url;
+        $this->name = $name;
+        $this->importDir = $import_dir;
+        return self::download();
+    }
+
+    /**
+     * @return mixed|string
+     */
+    private function download()
+    {
+        $zipResource = fopen($this->importDir . $this->name, "w");
+        // Get The Zip File From Server
+        $ch = curl_init();
+        curl_setopt($ch, CURLOPT_URL, $this->url);
+        curl_setopt($ch, CURLOPT_FAILONERROR, true);
+        curl_setopt($ch, CURLOPT_HEADER, 0);
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+        curl_setopt($ch, CURLOPT_AUTOREFERER, true);
+        curl_setopt($ch, CURLOPT_BINARYTRANSFER,true);
+        curl_setopt($ch, CURLOPT_TIMEOUT, 10);
+        curl_setopt($ch, CURLOPT_SSL_VERIFYHOST, 0);
+        curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, 0);
+        curl_setopt($ch, CURLOPT_FILE, $zipResource);
+        $page = curl_exec($ch);
+        if(!$page) {
+            return "Error :- ".curl_error($ch);
+        } else {
+            $status = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+        }
+        curl_close($ch);
+        return $status;
+    }
+
+    /**
+     * @return string
+     */
+    public static function createImportDir(): string
+    {
+        $import_dir = dirname(__DIR__, 8) . DIRECTORY_SEPARATOR . "sites" . $_SESSION['site_id'] .
+            DIRECTORY_SEPARATOR . "documents" . DIRECTORY_SEPARATOR . 'imports/';
+        if (!is_dir($import_dir)) {
+            $import_dir = dirname(__DIR__, 8) . DIRECTORY_SEPARATOR . "sites" .
+                DIRECTORY_SEPARATOR . $_SESSION['site_id'] .
+                DIRECTORY_SEPARATOR . "documents" . DIRECTORY_SEPARATOR;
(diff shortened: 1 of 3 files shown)


adunsulag:
I like the idea and the concept.  I'm trying to think through the security ramifications of this change.  I've wanted for quite some time to introduce a module installer into the codebase for non-technical users.  The challenge is to avoid introducing malicious code execution from a 3rd party.

If I'm following your code correctly, this will put the module on the file system but it will not activate the module.  Currently the only thing in our code base that will actually scan and parse a file in the downloaded module is the info.txt where the name of the module is parsed in the modules installer.  No code is executable unless the module has been activated or if its paired with another OpenEMR exploit that would locally include the file.

I know we may not want to do this exercise, but I think we need to carefully do a risk assessment on potential threats and threat actors for bringing this code in.

Threats:
1. Non-Admin user calling the 3rd party module installer directly to download and activate a malicious module/script.
2. Privilege escalation attack allowing non-admin user to become admin user and installing malicious modules.
3. Supply chain attack on 3rd party module coupled with local file inclusion attack to include malicious script inside OpenEMR
4. Others??

Mitigation
1. Threat 1 - All installer actions need to be restricted to an administrator ACL to prevent non-admin users from installing / activating modules.
2.Threat 2 - Prevent privilege escalation attacks in OpenEMR.  If an admin user account is compromised the entire OpenEMR installation data and system can be compromised due a 3rd party module installed.  Allowing this code change means a privilege escalation attack must be treated as an entire system compromise event as module code can be both installed and removed from the system.
3. Threat 3 - Any include statements or local code inclusion statements in php must be heavily sanitized to prevent loading code stored in the custom_modules directory.
4. Threat 3 - Should we encrypt the 3rd party module code and flag in the database the code is encrypted?  We do this for file uploads in the documents directory to prevent local file inclusion attacks.  It will slow down 3rd party modules but we could decrypt the module and store it in memory during page load.  Its been a while but I believe this is how Java Enterprise handles some of their 3rd party modules.  However, they store the code in memory in the servlet which is persistent and thus can save on page loads whereas php wouldn't have this advantage.  We could store the module in a redis cache or something similar.  

I'm sure there are other things to consider as I've just spent a bit of time thinking through the threat assessment.  Granted if the end user wants to install a module is this any different then them blindly doing a 'composer require <module-name>'.  Our assumption is that users capable enough of doing something like that knows a bit more to not just blindly trust a link.  Plus there's no chance for privilege escalation attacks here.  @bradymiller and @robertdown likely have some additional input here.

On a minor aside, this code doesn't appear to update the composer autoloader so the only kinds of modules that will work here are ones that directly include/require their sub files instead of using the system autoloader.

juggernautsei:
@adunsulag I was waiting on this discussion.  Go check out Drupal's module installer. It works (on the surface) very similarly to this.
To respond to this statement:
  "this code doesn't appear to update the composer autoloader so the only kinds of modules".

The way that I packaged up the file is not different than the way OpenEMR is packaged into a zip file (in my opinion). All of the composer stuff has been taken care of before creating the zip file so on install. No composer work is needed. Be kind to my limited view of things.

I agree with your concerns. When we look at WordPress or any other system out there that allows third-party modules. There are modules that are certified and those that are not. I don't know all the processes that make a module on any of the certified or not.

It ultimately becomes the responsibility of the user to investigate whether they should install the module or not.

I looked at the code for the module installer and ACL has been added to the module installer that it can only be done by Admin. 

Can we come up with a safety system to have modules that are considered safe to install? You bet your sweet bippy we can.

I do agree with all the safety concerns. They are all valid and need to be addressed. I just thought it was time to get the ball rolling. 

Who's kicking it next?

@mdsupport, I like to hear from you.

adunsulag:
@juggernautsei So when OpenEMR is installed all of the composer autoload functionality is already present and accounted for.  Using this module installer mechanism, new code will be brought into the codebase that will not be in the composer autoload code structure (kept inside the vendor/composer) directory.  That code will need to be inserted into the main project composer.json file and then do a  ```composer dump auto-load``` if we want it to work with the autoloader. 

There are other ways we could get around doing this.  The Laminas modules have their bootstrap mechanism that extends the autoloader but this will require us to do a php file include for every module in the system to pick up autoload definitions.  Wordpress does this with their plugin bootstrap files.  However, in a healthcare context I'm not sure that I feel comfortable auto-loading automatically files before the user has hit the install/activate button.  

We could go with some kind of sha1 hash or a public / private key verification from OpenEMR similar to other OS package distribution mechanisms (I'm most familiar with debian packages in this regard).  To verify and validate zip file contents but I'm just uncertain on how much we try to protect OpenEMR installations vs letting the user do whatever they want.

Reading up on the Drupal Module installation best practices they have a trusted_sites setting that only allows downloading modules from trusted domain name locations.
https://www.drupal.org/node/2410395

Granted if someone does a DNS poisoning attack this security mechanism could be bypassed, but I think the target OpenEMR installation at that point would have bigger problems.

juggernautsei:
@adunsulag when I built my modules, I followed the suggestion of not polluting the OpenEMR namespace. So, when I run the composer autoload it creates the vendor folder and all of the classes that are in the module autoload for the module. It seems to connect fine to the OpenEMR namespace and all the use statements work like use Header. I have installed the module on three different installs. The module worked on each without having to run composer dumpautoload.

juggernautsei:
I have done some more testing today. The update feature works. I updated the prior auth module. Created a new zip file for it. Posted new link on the server. I pasted the link into the module uploaded. It deleted the current installation and installed the updated version of the program. The new features were now available in the installation.

Let's get this out for others to use!

bradymiller:
Can't go into details here, but the current module code in codebase requires significant security improvements before this could get in. At this time there would be several avenues that any authenticated user could bring in code from any url they want via module_import and then run that code on the server. Until current codebase could support something like this, would rec the zip method (ie. a user downloads the zip onto their server and places it in the module directory) or composer method for bringing in modules.

juggernautsei:
@bradymiller Let's have an private conversation about the security issues. I have sent you an email expressing this desire.

juggernautsei:
Learning the Luminas way

mdsupport:
@juggernautsei What is the target ("non technical" users) here? Apart from the different layers of security concerns, there is operational aspect you will need to address. The current module framwork is predicated on code being in standard and one repository. Any piece of code outside the openemr repo as you are proposing opens up a dependency hell. Since you have been through few upgrades, please carefully reconcile standard upgrade instructions for a site with multiple custom modules.

On a related note, current structure and its reliance on a specific directory are before composer based installation procedure. If you are going to rely on your own classes, treat yourself as a vendor and retain full control of your code without concern for what standard package does. We grappled with this aspect for several years before arriving at [this framework](https://community.open-emr.org/t/new-features-improvements/12089/16) for adding or altering standard application across multiple versions and multiple installations.

Best.

juggernautsei:
@mdsuport The target is the none tech user. I am looking at this as proof of concept. I  am working on Luminis classes.

PR #5004, opened 2022-03-14
missed files previous PR

sjpadgett:
These were supposed to be in already!

### DIFF
--- a/src/Services/Qdm/MeasureService.php
+++ b/src/Services/Qdm/MeasureService.php
@@ -30,6 +30,12 @@ class MeasureService
         return $options;
     }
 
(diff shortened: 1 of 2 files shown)


sjpadgett:
@kchapple We may need to change our reporting period to match the mustache templates. e.g
ruby project is using recent cda template version such as: 
`<templateId extension="2021-08-01" root="2.16.840.1.113883.10.20.24.3.23"/>`

where the 2020 performance period test patient qrda is using:
`<templateId extension="2019-12-01" root="2.16.840.1.113883.10.20.24.3.23"/>`

note extension date. this is really a template version date.
Soo much!

bringing in

PR #5009, opened 2022-03-16
update package.json for new oe-cqm-parser version

sjpadgett:
updated our npm for oe-cqm-parser version to 5.6.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
         "magic-wand-js": "1.0.0",
         "moment": "2.29.1",
         "numeral": "2.0.6",
-        "oe-cqm-service": "1.0.1",
+        "oe-cqm-service": "1.0.2",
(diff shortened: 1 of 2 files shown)


sjpadgett:
@bradymiller needs to be tested for demos. work okay for me on windows but we know how that goes.

sjpadgett:
@bradymiller  I now have the issue of having to change the new eCQM 2021 Performance Period list. This made me apprehensive in the first place for this very reason.
Should I delete the list then insert a new one? Just modify the existing list inserts in upgrade as if it was never installed. The only folks affected could just delete their list the rerun update.

sjpadgett:
looks like I need to re publish oe-cqm-service and bump it's version to reflect oe-cqm-parser bump! nutz

sjpadgett:
going to bring this in because demo reset may break if not. I think I got it right.

PR #5010, opened 2022-03-16
Update for 2022 CQM measures

### LINKED ISSUE #5005: QRDA reporting versioning required.
Currently all our QRDA templates are based on the CDAR2 IG R1 STU Release 5.3 2021 which is slated for the 2022 performance period.

For ONC testing we have two choices: 2020 or 2021 periods. Our current measures and patient models are for the 2020 period which was our plan for testing however, we need to rethink IMO and test for 2021.

This leads to the need to version our templates to take into account new CDAR STU releases.
Since our patient models for the calculator are created from the reporting period measures, I don't see that as being too much of an issue. Templates though, will be.

For this testing go around I suppose we could simply pull in the correct templates and for future testing we update our template tree with the newly needed template changes. One of the projects that Cypress is based maintains these templates in a forward looking way so access to new templates will not be an issue that I can currently see.

However this would mean no legacy reporting throughout OpenEMRs product release lifecycle. Meaning the reporting period for a particular OpenEMR release would need to be fixed. In many ways, this bothers me.

For now I think we need to regenerate workflow for the 2021 performance period. We can updates our templates and not do the work to report for 2021 and be okay for testing but, I feel it is best to do this now.

@kchapple @adunsulag @bradymiller Awaiting thoughts on this matter if so incline to chime in.:)

sjpadgett:
change list for ecqm_2022_reporting
bump DB version
Fixes #5005


sjpadgett:
only one new measure 'CMS646v2','Intravesical Bacillus-Calmette-Guerin for non-muscle invasive bladder cancer'. Doesn't pertain to our use.

sjpadgett:
@kchapple 
I have all the services/dependencies updated(I hope). I was able to install module as normal and the new measures pulled into your tool.
I was able to regenerate the new patient models(nice work btw) however, I'll not commit and let you have a look see for any quirks I've not noticed.  You should have enough to work with. I'll bring this in once Brady has a chance to look at.

bradymiller (approved):

bradymiller:
2DV

PR #5011, opened 2022-03-16
Added 5.6 QDM models, and a readme how to generate new models

kchapple:
Updating models and tested generating patient JSON, measure execution via CQM calculator, and generating QRDA Cat I templates (via CQM tools and Care Coordination)

Everything looks good for me to continue fine-tuning the templates, and populating the Cat III template, including the download of QRDA Cat I via Care-coordination module. 

The "View QRDA" doesn't seem to work, but I think that's a known issue.

The "View CCDA" Gives me an error, but I think it's a "me" problem because I see this on the screen:
```
Message:
    Please Enable C-CDA Alternate Service in Global Settings
```


kchapple:
We can sit on this PR, I'll work from this branch to work on templates.

sjpadgett:
Hi @kchapple 
You need this PR #5010 for QRDA reporting to work because I added measure names to a list so user can select what default measures for different practice disciplines.
You have to install the ccda service and/or turn on for ccda to work.

sjpadgett:
Also @kchapple You'll need to change performance period in Globals->CDR to 2022 from 2021. I forgot in PR and code default.

image

QRDA downloads are in CCM:

image

kchapple:
There appears to be a problem, and I'm not sure but I think it may be related to the way we're not using namespaced helpers (not using helpers at all actually) and presenting functions for the template variables in the global context. For example, every code on the patient _record_target template is generating an element for every other code:

```
<administrativeGenderCode code="21112-8" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC"/>
<administrativeGenderCode code="1006-6" codeSystem="2.16.840.1.113883.6.238" codeSystemName="CDC Race"/>
<administrativeGenderCode code="" codeSystem="2.16.840.1.113883.10.20.28.3.56" codeSystemName="Unknown"/>
<administrativeGenderCode code="F" codeSystem="2.16.840.1.113883.5.1" codeSystemName="AdministrativeGender"/>
```

```
<raceCode code="21112-8" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC"/>
<raceCode code="1006-6" codeSystem="2.16.840.1.113883.6.238" codeSystemName="CDC Race"/>
<raceCode code="" codeSystem="2.16.840.1.113883.10.20.28.3.56" codeSystemName="Unknown"/>
<raceCode code="F" codeSystem="2.16.840.1.113883.5.1" codeSystemName="AdministrativeGender"/>
```

I'm looking into it.

kchapple:
@sjpadgett this is the latest.

src/Services/Qrda/Helpers/Date.php
```diff
@@ -46,7 +46,7 @@ public function performance_period_end(Mustache_Context $context)
 
     public function current_time(Mustache_Context $context)
     {
-        return DateHelper::format_datetime_gmdate(date('Y-m-d H:i'));
+        return DateHelper::format_datetime(date('Y-m-d H:i'));
```
sjpadgett, line 49: @kchapple again, date format needs to be in the form xxxxxxxx not xxxx-xx-xx
will not pass IG setup this way.

kchapple: The format_datetime function does that, I could probably just remove the parameter

adunsulag:
@kchapple @sjpadgett Did y'all know that projectcypress/cypress latest tagged version (6.2.1) uses version 3.1.8 of the cqm-reports project? Version 3.1.8 has a number of changes in their mustache templates https://github.com/projecttacoma/cqm-reports/tree/v3.1.8/lib/qrda-export/catIII-r2-1 which I'm bringing into my branch as part of the ruby conversion.  I think that is all correct.  If for some reason we actually should be using the version in cqm-reports master branch (which is what is in this PR originally) please let me know and I'll revert those changes.

sjpadgett:
@adunsulag This is concerning! Did you do a diff on the two versions? Also I need to check the Cat 1 I guess

kchapple:
I thought I checked all the versions, but that may have been for the old reports, or maybe I accidentally pulled the wrong templates? Thanks @adunsulag using the versions specified in the latest cypress 6 GemFile, which I believe are the versions you mentioned, sounds right.

sjpadgett:
Umm hold up. I'm pretty certain @adunsulag you're working with an older Cypress container because we had to goto a newer QDM version to support 2021 reporting period. Templates in v3.1.8 are behind 2021 reporting period. The version of  Cypress I'm working with show using this gem: gem `'cqm-reports', '~> 3.1.9'` it came from master.

I better run setup and run a test on the public demo but it should be the same

sjpadgett:
@kchapple I pulled your changes and cat1 export is broken. Is it working for you?

kchapple:
> @kchapple I pulled your changes and cat1 export is broken. Is it working for you?

Yes, looks like it's working from CQM Tools at least. I don't think I touched anything in that pipeline. You may need the latest oe-module-cqm?

kchapple:
It’s possible I changed something and forgot to update the QRDAController that you wrote to handle exports to care cooordination.

kchapple:
Yep, that was it @sjpadgett sorry. I changed some things around and forgot to update the CareCoordination call. The last commit should fix it.

sjpadgett:
Thank you sir. I'm going to verify this version mess. I'm pretty sure the version in now is correct!

sjpadgett:
Hey, look what I stumbled into:
image

kchapple:
@sjpadgett I'm not exactly sure what that means, but I assume green is good! I just pushed address and phone additions to patient model.

sjpadgett:
> but I assume green is good!

No I was using wrong cat1 report.

kchapple:
ahh, that makes more sense!

kchapple:
Closed in favor of #5046

PR #5016, opened 2022-03-17
Some Import fixes

sjpadgett:
fix ethnicity not parsing correctly

More to come like mass import add patient


sjpadgett:
@kchapple FYI
Added ability to clear all pending imports before committing to chart. Useful if you just want to checkout how the import parses and/or view imported doc details then clear imported.
Also, added button to auto chart all the pending imports instead of have to import to chart one by one.
Note that if a patient match is found the auto import will skip so user can review and approve any possibly changes.

image

@bradymiller I plan to bring in in morning if you want to look over. thanks.

bradymiller (approved):

bradymiller:
2DV

sjpadgett:
@bradymiller Thanks for adding the test rerun on only failed test. Very nice to have with a lot less aggravation when in a hurry.

PR #5020, opened 2022-03-19
Openemr fix 5003 group calculation

### LINKED ISSUE #5003: AMC § 170.315 (g)(2) 2015 ONC - Group Calculation Method
We need to support the group calculation method for aggregating AMC results under a TIN for multiple provider NPIs.  The way we discussed doing this is to have multiple billing facilities and track the patient to provider NPI relationship via the encounters for that patient.  If a patient has a single encounter during the performance period (usually the entire year) with a provider linked to a TIN via the billing facility we will count them as a patient for that encounter. 

In our report we need to display essentially this kind of data for a group calculation:
image
image

You can see the group and transitive calculation differences between Mai and Tom here.  Tom is seen at Billing Facility A (TIN A) by provider 1 and Tom is seen at Billing Facility B (TIN B) by provider 2.  Provider 1 & provider 2 are part of two groups TIN A, and TIN B.  In the calculation Provider 1 with TIN B does not get credit at the individual EP/EC level for Tom, but they do get credit under the EP/EC Group Calculation method.  The same for Provider 2, no credit for TIN A under individual EP/EC calculation, but they do get credit under the Group Calculation method.

So we don't have to display the data exactly as these screenshot reports, but the ATL tester needs to be able to clearly see the calculations and see the reports of the individuals for numerator / denominator for each calculation.

I'm still going through designs of how to do this group calculation but I may just have a Group Calculation dropdown that lists each billing facility.  The Group Calculation will display each individual provider in the group and their patient breakdown and then there will be a summary for the entire Group.

When we submit our reports to ONC we will need to collate the individual Group Calculations with each of the individual calculations which should display all the requisite data without having to drastically alter our reports.  At least that is my thoughts.

adunsulag:
Fixes #5003 
Keeping this as a draft, but want people to be able to see what I've been doing here with the group calculations.  Had to go through 4 different iterations to finally get something that would work without breaking all the other rules for CQM/alerts/patient_reminders (as well as working within our batching framework for large patient numbers).

I moved all of the report results into twig to 1) simplify the code logic and 2) make it extendable so I could add the group functionality easily (and support future types if we need it for the reports).

I had to heavily modify the buildPatientArray to handle the billing facility connection.

I originally had it reporting individual provider breakdowns underneath the groups.  I may still add that in as I'm not sure if we need it for ONC testing or not.  The test data shows individual provider breakdown by group TIN but the actual MIPS group calculation doesn't mention it and only talks about numerator / denominator percentage.

The ItemTracker classes simplify the checking for individual item reporting and I originally was leveraging them to tally up provider group breakdowns.  I may still use them if I have to have the individual provider breakdown.

Couple screenshots of the Group AMC (The tests show failing correctly in the screenshot as I haven't generated the patient API credentials here to pass the test)
image

image

Note this is what the Test Script shows for the baseline test as you can see the provider by TIN breakdown.  Again not sure if its necessary after reading the testing instructions many, many times. 
image

I pulled out the provider breakdown as it was just failing repeatedly but I will keep working on it to see if I can make it work.


sjpadgett:
It'd probably take me a week to figure out what the heck you are talking about!:)
Lots of words though so, must be cool!

bradymiller:

adunsulag:
@bradymiller The Group Calculation PR should be ready to be reviewed now for the AMC.  I've been a bit sick so I may have missed something (which I'm sure Brady will call me out on) but feel free to attack it :)

library/classes/rulesets/Amc/library/AbstractAmcReport.php
```diff
@@ -20,15 +20,25 @@
...
 
+require_once(dirname(__FILE__) . "/../../library/RsFilterIF.php");
 require_once('AmcFilterIF.php');
 require_once(dirname(__FILE__) . "/../../../../clinical_rules.php");
 require_once(dirname(__FILE__) . "/../../../../amc.php");
 
+use OpenEMR\Reports\AMC\Trackers\AMCItemTracker;
+use OpenEMR\Reports\AMC\Trackers\AMCItemSkipTracker;
+
+use OpenEMR\Common\Logging\SystemLogger;
```
bradymiller, line 35: alphabetize use block

library/clinical_rules.php
```diff
@@ -21,6 +21,8 @@
 
 use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\ClinicialDecisionRules\AMC\CertificationReportTypes;
+use OpenEMR\Reports\AMC\Trackers\AMCGroupItemTracker;
+use OpenEMR\Common\Logging\SystemLogger;
```
bradymiller, line 25: alphabetize

library/clinical_rules.php
```diff
@@ -589,8 +596,159 @@ function test_rules_clinic_batch_method($provider = '', $type = '', $dateTarget
     }
 }
 
+function rules_clinic_get_providers($billing_facility, $pat_prov_rel)
+{
+    $results = [];
+    // TODO: @adunsulag do we need to batch the provider processing here?  Its unlikely to have more than 50-100 providers...
```
bradymiller, line 602: no need to batch for now; the batching for patients was done since it literally took days for when patients over 10000 because it kept getting slower with each patient. The patient batching removed this issue.

library/clinical_rules.php
```diff
@@ -589,8 +596,159 @@ function test_rules_clinic_batch_method($provider = '', $type = '', $dateTarget
...
+            'pat_prov_rel',
+            'start',
+            'batchSize',
+            'user'],
+            func_get_args()
+        )
+    );
+
+    $results = [];
+    $facilityService = new \OpenEMR\Services\FacilityService();
```
bradymiller, line 672: place in use block

library/clinical_rules.php
```diff
@@ -658,12 +804,13 @@ function test_rules_clinic($provider = '', $type = '', $dateTarget = '', $mode =
...
 
-  // If set organize-mode to plans, then collects active plans and run through this
-  // function recursively and return results.
+    // If set organize-mode to plans, then collects active plans and run through this
+    // function recursively and return results.
     if ($organize_mode === "plans") {
         // First, collect active plans
         $plans_resolve = resolve_plans_sql($plan, $patient_id);
         // Second, run through function recursively
+        error_log(var_export($plans_resolve, true));
```
bradymiller, line 813: guessing debug that needs to be removed or placed in system logger function

library/clinical_rules.php
```diff
@@ -1126,6 +1350,101 @@ function buildPatientArray($patient_id = '', $provider = '', $pat_prov_rel = 'pr
...
+        $sql .= " WHERE `billing_facility` = ? ";
+        $binds[] = $billing_facility;
+    }
+    if (!empty($provider_id)) {
+        $sql .= " AND provider_id = ? ";
+        $binds[] = $provider_id;
+    }
+    $sql .= "ORDER BY `pid`";
+    if (!($start == null || $batchSize == null || $onlyCount)) {
+        $sql .= "LIMIT ?,?";
```
bradymiller, line 1382: limits can's use binding. i escape with the escape_limit() function (always available and in formdata.inc.php), but intval is ok also since this is what escape_limit basically does

library/clinical_rules.php
```diff
@@ -1126,6 +1350,101 @@ function buildPatientArray($patient_id = '', $provider = '', $pat_prov_rel = 'pr
...
+        $binds[] = $billing_facility;
+    }
+    if (!empty($provider_id)) {
+        $sql .= " AND (`provider_id`=? OR `supervisor_id`=?) ";
+        $binds[] = $provider_id;
+        $binds[] = $provider_id;
+    }
+    $sql .= "ORDER BY `pid`";
+    if (!($start == null || $batchSize == null || $onlyCount)) {
+        $sql .= "LIMIT ?,?";
```
bradymiller, line 1430: same issue as above where can not bind limits

library/report_database.inc
```diff
@@ -292,11 +292,35 @@ function collectItemizedRuleDisplayTitle($report_id, $itemized_test_id, $numerat
...
+
+    // for our group calculation reports we want to display the group and the provider label information
+    if ($report_view['provider'] == 'group_calculation') {
+        $display_group_provider_info = true;
+    }
     foreach ($dataSheet as $row) {
+        if ($display_group_provider_info) {
+            if (isset($row['is_provider_group'])) {
+                $group_provider_label = xl('Group') . ': ' . $row['name'] . ' ( ' . xl('TIN') . ' '
+                    . $row['federaltaxid'] . ' )';
```
bradymiller, line 306: need to html escape this, 
xlt text xlt text

library/report_database.inc
```diff
@@ -292,11 +292,35 @@ function collectItemizedRuleDisplayTitle($report_id, $itemized_test_id, $numerat
...
+            } else if (isset($row['is_provider'])) {
+                $provider_label = ' ' . xl('Provider') . ': ' . $row['prov_fname'] . ',' . $row['prov_lname']
```
bradymiller, line 308: need to html escape this,
xlt text text xlt text

bradymiller:
nice stuff! code review complete. didn't test, but have faith :)

library/report_database.inc
bradymiller: Good point and not sure why i did escaping in this function rather than on the echo of it (maybe i was using html in the past which have since removed). In that case would be best to strip escaping in the other items in this function.

adunsulag: I was wrong here, this is escaping into the patient_finder and not into the cqm output so I do need to escape these values.

bradymiller (approved):

bradymiller:
code looks good to me (didn't test, but have faith :) )

tenor (67)

PR #5024, opened 2022-03-23
Add ccda meets amc numerator for med,allergies and problems

sjpadgett:
@adunsulag 
Hope this is what is needed. Result flag is map_id true/false. Didn't know if you need to map to document or even the transaction map_id but, you can change as you need.

localhost-127-0-0-1-openemr_qrda-amc_misc_data-phpMyAdmin-5-1-1

Bring in if I'm not around.


adunsulag:
@sjpadgett Thanks Jerry!

PR #5028, opened 2022-03-23
updated to mariadb 10.7 for ci and docker dev environments

bradymiller:
updated to mariadb 10.7 for ci and docker dev environments


bradymiller:
actually, this PR does not make sense since mariadb 10.7 is a very short term release (https://en.wikipedia.org/wiki/MariaDB), so will stick with 10.6 at standard testing version for now.

PR #5030, opened 2022-03-24
Bug fixes that will go to rel-610

bradymiller:
bug fixes:
- advanced directive broken query bug fix
- fix for print LBF form


interface/forms/LBF/printable.php
```diff
@@ -97,10 +97,10 @@
         'default_font' => '',
         'margin_left' => $GLOBALS['pdf_left_margin'],
         'margin_right' => $GLOBALS['pdf_right_margin'],
-        'margin_top' => $GLOBALS['pdf_top_margin'],
-        'margin_bottom' => $GLOBALS['pdf_bottom_margin'],
-        'margin_header' => '',
-        'margin_footer' => '',
+        'margin_top' => $GLOBALS['pdf_top_margin']*2,
+        'margin_bottom' => $GLOBALS['pdf_bottom_margin']*2,
```
bradymiller, line 101: note had to multiple this by 2 so the text in the header/footer did not overlap with the main text

interface/forms/LBF/printable.php
```diff
@@ -97,10 +97,10 @@
...
+        'margin_header' => $GLOBALS['pdf_top_margin'],
+        'margin_footer' => $GLOBALS['pdf_bottom_margin'],
```
bradymiller, line 103: note had to multiple margin-top/margin-bottom by 2 so the text in the header/footer did not overlap with the main text

bradymiller: @sjpadgett , @stephenwaite ,
Another option that works is setting the margin_header and margin_footer to 0 (and then do not need to multiple the margin_top/bottom by 2). Am thinking that would be best although would be great to get input.

sjpadgett: I wouldn't hard code a scaler value. Might confuse user. Change global default.
Glad we didn't have to revert Mpdf. I saw this awhile back and it seem to coincide with th version bump so, that where my mine went.

bradymiller: There is no global settings for margin_header/margin_footer. Setting them to zero makes them too close to the end of page and then do not get printed. By making the margin_header the value of global pdf_top_margin works well and then set the margin_top to 2 times this (need to do this or the text overlaps), which is testing well. Gonna go with this for now so will now work. Can always revisit in future to improve (especially if start using headers/footers in other places where mPDF is used).

stephenwaite:
nice

PR #5038, opened 2022-03-26
Introduce machine_name as a global option

robertdown:
New global for the machine name of the application. Used to identify the EMR when working with HL7 and other systems that may want this information.


bradymiller (approved):

adunsulag:
@robertdown Don't we already have the unique_installation_id global GUID for this that uniquely identifies the OpenEMR installation?  Wouldn't that serve this purpose?

robertdown:
It could, though it's not controllable. This would allow an actual OID to be used.

PR #5045, opened 2022-03-27
fix race import

sjpadgett:
Fixes #

fix race code not populate patient data
Note @kchapple


sjpadgett:
@kchapple @adunsulag 
Pretty cool. Testing using the Cypress CVU validation mode is mush easier for quick test. You can have a patient from the patient download for the test that covers all the measures under test. Nice because in shows the calculator results. This is the CMS69v10 measure CMS Cat1 test

image

PR #5061, opened 2022-03-30
Revert Cat1 mustache to cqm-report version 3.1.8

sjpadgett:
@adunsulag @kchapple 
So what happen. Thought these template were to be brought to same cqm-reports version as the Cat3 template 3.1.8 with Stephens push.

Bring in projects 3.1.8 version templates to match cypress testing.

### DIFF
--- a/src/Services/CodeTypesService.php
+++ b/src/Services/CodeTypesService.php
@@ -368,4 +368,17 @@ class CodeTypesService
         );
         return $value;
     }
+
+    public function dischargeOptionIdFromCode($formatted_code)
+    {
+        $listService = new ListService();
+        $ret = $listService->getOptionsByListName('discharge-disposition', ['codes' => $formatted_code]) ?? '';
+        return $ret[0]['option_id'];
+    }
+
+    public function dischargeCodeFromOptionId($option_id)
+    {
+        $listService = new ListService();
+        return $listService->getListOption('discharge-disposition', $option_id)['codes'] ?? '';
+    }
 }
--- a/src/Services/Qrda/Helpers/Cat1View.php
+++ b/src/Services/Qrda/Helpers/Cat1View.php
@@ -123,9 +123,8 @@ trait Cat1View
 
     public function result_value(Mustache_Context $context)
     {
-        // TODO Do patient models ever return a string for results?
         $result = $context->find('result');
-        if (empty($result['value'])) {
+        if (empty($result)) {
             return "<value xsi:type=\"CD\" nullFlavor=\"UNK\"/>";
         }
         if (is_array($result)) {
--- a/src/Services/Qrda/Helpers/Cat1View.php
+++ b/src/Services/Qrda/Helpers/Cat1View.php
@@ -123,11 +123,11 @@ trait Cat1View
 
     public function result_value(Mustache_Context $context)
     {
+        // TODO Do patient models ever return a string for results?
         $result = $context->find('result');
-        if (empty($result)) {
+        if (empty($result['value'])) {
             return "<value xsi:type=\"CD\" nullFlavor=\"UNK\"/>";
         }
-
         if (is_array($result)) {
             // indexed array
             if (array_key_exists(0, $result)) {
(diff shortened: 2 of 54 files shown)


sjpadgett:
Darn Phpstorm profiler is broken at least on windows. This used to work well but now profile result files compiled to gzip are corrupted. Tried everything.
Every new release is worst! I'm now on the EAP 2022.1Will have to use VSCode to profile!!!

sjpadgett:
Okay fixed result_values helper I think. It's also used in observation and labs where I'm not sure I've covered all the scenarios possible or tested yet. Check me @kchapple 

CMS22v10 is passing IG validation.

src/Services/Qrda/Helpers/Cat1View.php
```diff
@@ -123,11 +123,11 @@ public function dose_quantity_value(Mustache_Context $context)
 
     public function result_value(Mustache_Context $context)
     {
+        // TODO Do patient models ever return a string for results?
```
sjpadgett, line 126: @kchapple Still need to verify various patient models QDM result types for observations and labs. You may want to change the way I did helper but this is close.

adunsulag:
@sjpadgett My bad Jerry.  I forgot to commit the 3.1.8 cat1 files.  Glad you got it taken care of.

sjpadgett:
> I forgot to commit the 3.1.8 cat1 files

No problem @adunsulag. I thought I did something in my git chain.

sjpadgett:
@kchapple I've tested CMS22v10, 69v10, 122v10 and 124v10 which you can log in to my server and view the results under the 2022 CQM Product. 
So far using the result_value Lambda changes in this PR we generally pass validation with some minors like reverse low/high effective dates(new I think as don't recall seeing before) and missing author etc. I'm starting to see a light to end.
However, below is new from my previous testing and hope you can shed some light. Every cat1 seems to have one or more from cypress calculator. Let me know if you want to have a call so I can give you some tips on cypress tool. Thanks.

image

kchapple:
Thanks @sjpadgett 
It’s hard to tell “why” without comparing the cat 1 XML, but those are denominator exclusions. I suspect that there is some model, or perhaps a status attribute we’re not including.

sjpadgett:
Thanks Ken, I think I found it because I wasn't resolving the PQ datatype correctly. It's strange that the Site public demo cypress gave different result to same test different patients.

lol, my brain hates inconsistencies like these. There's other strange things like a patient download is 75 patients but test only wants 72! Could be they send invalid cat1 and I'm suppose to throw out. Oh well just another little gotcha...

sjpadgett:
So why would we have to handle an inpatient encounter. I don't look for or import sdtc:dischargeDispositionCode
Where would I put?

```
<text>Encounter Inpatient</text>
      <statusCode code="completed"/>
        <!-- QDM Attribute: Relevant Period -->
        <effectiveTime><low value='20200808080000'/><high value='20200820081500'/></effectiveTime>
        <!-- QDM Attribute: Discharge Disposition -->
        <sdtc:dischargeDispositionCode code="428371000124100" codeSystem="2.16.840.1.113883.6.96" CodeSystemName="SNOMEDCT"/>
```

sjpadgett:
Looks like @adunsulag added the encounter disposition to encounters so yeah!
needs codes though!

src/Services/CodeTypesService.php
```diff
@@ -368,4 +368,17 @@ public function lookupFromValueset($code, $codeType, $codeSystem)
...
+    public function dischargeOptionIdFromCode($formatted_code)
+    {
+        $listService = new ListService();
+        $ret = $listService->getOptionsByListName('discharge-disposition', ['codes' => $formatted_code]) ?? '';
+        return $ret[0]['option_id'];
+    }
+
+    public function dischargeCodeFromOptionId($option_id)
+    {
+        $listService = new ListService();
```
sjpadgett, line 381: @kchapple this is where you'll get the discharge code to populate the
QDM Attribute: Discharge Disposition sdtc:dischargeDispositionCode. 
I still have work to populate form_encounter with code instead of option_id if not in list plus add all the codes to the list. ugh
It's populated in encounter:
image

sjpadgett:
Give me an hour so I can push the codes for discharge to lists or i'll just bring in.
Sometime this morning when I get this PR up, we can talk and i'll run you through cypress.

sjpadgett:
@kchapple I'm going to bring into master so the 3.1.8 templates are in...

sjpadgett:
Okay Ken, don't forget to run sql_upgrade

PR #5072, opened 2022-04-01
registration email verify fix when language menu is off

bradymiller:
registration email verify fix when language menu is off


bradymiller:
Will also bring this into next 6.1.0 patch

adunsulag (approved):
I haven't executed it on my machine but looking over the changes this looks fine to me.

PR #5080, opened 2022-04-01
Fixes #5078 CCM Direct Timing

### LINKED ISSUE #5078: Carecoordination timing issue sends to EMRDirect out of order before CCDA generation
The CCDA generation and EMR Direct ajax post requests in Care Coordination get fired off at the exact same time.  If the send to EMR data depends on the prior CCDA request (which one would assume it would) it will fail (or happily send an earlier generated CCDA).  The code flow needs to be structured so that ajax requests are serial rather than concurrent.

adunsulag:
Fixes #5078 Make the ccda generation process first and then do the transmitCCDA function in CCM.


sjpadgett (approved):
yep looks like most logical way around.

bradymiller:
2DV-2

PR #5090, opened 2022-04-04
Fixes #5089 Remove unused CDR menu items.

### LINKED ISSUE #5089: Remove unused menu options for CDR: AMC Tracking and CQM
Since we are generating the CQM in another location and we are not using / certifying the AMC Tracking for (g)(1) (which uses old rulesets anyways), we should remove the CQM and AMC Tracking menu options under the Reports -> Clinic menus.

adunsulag:
Fixes #5089 Removes the CQM menu since that is now done in carecoordination module.
Removes the AMC tracking since it uses old rule sets and we aren't
certifying against it.

### DIFF
--- a/interface/main/tabs/menu/menus/standard.json
+++ b/interface/main/tabs/menu/menus/standard.json
@@ -1474,19 +1474,6 @@
             "global_req": "enable_cdr"
           },
           {
-            "label": "Quality Measures (CQM)",
-            "menu_id": "rep0",
-            "target": "rep",
-            "url": "/interface/reports/cqm.php?type=cqm",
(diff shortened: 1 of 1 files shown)


bradymiller:
copy the modified interface/main/tabs/menu/menus/standard.json to sites/default/documents/custom_menus/Custom.json and then this PR is ready to come in
(we keep the Custom.json the same as the standard.json)

2DV-2

adunsulag:
@bradymiller Made the change, and if that looks good I'll bring this in.

bradymiller (approved):

PR #5108, opened 2022-04-07
MedEx Tweaks

ophthal:
Apply hipaa_override to all message group.
Delete MedEx Preference "Assume patients permit Messaging" from the interface.
Will need an additional PR to remove the corresponding field from DB.


stephenwaite:
Hi Ray @ophthal , guessing something broke on `v7`. Can you checkout this forum thread or fill us in on anything you can please? Thanks

ophthal:
Looks like these changes are not in your master and probably not the user’s either.
Please confirm and if so  I think it just needs to be reviewed and added to a patch.

stephenwaite (approved):
thanks @ophthal

bradymiller (approved):

stephenwaite:
nice @ophthal , just let us know when to pull the trigger :)

ophthal:
There is also one change needed in save.php in interface/main/messages.  I’ll send that later today.

PR #5109, opened 2022-04-07
CQM Matched Patient Ids with calculator

adunsulag:
Tacoma project CQM Calculator uses Mongoose ObjectId for its id properties. For the CQM calculation we can't just pass straight through the patient id in order to get the matched patient ids back.  Instead we have to pass in a stringified format for the patient id that matches the serialized BSON format for the ObjectId.  In this case its a 24 character string in hexidecimal format that represents the pid.

Fortunately the bigint data type is only 8 bytes and ObjectId is a 12 byte format so it fits within the serialization.

@kchapple @sjpadgett This should fix the matching of patient data in the ResultsCalculator.  I noticed that the codes property are empty in my QDM models in ResultsCalculator:L44-L56.  So gender, race, ethnicity, payer are not getting codes populated on my machine.  Not sure if I need to do something specific for the data setup.

### DIFF
--- a/src/Services/Qdm/CqmCalculator.php
+++ b/src/Services/Qdm/CqmCalculator.php
@@ -31,7 +31,8 @@ class CqmCalculator
         $this->client = CqmServiceManager::makeCqmClient();
     }
 
-    private function convertToObjectIdBSONFormat($id) {
+    private function convertToObjectIdBSONFormat($id)
+    {
         $hexValue = dechex($id);
         // max bigint size will fit in 16 characters so we will always have enough space for this.
         return sprintf("%024x", $hexValue);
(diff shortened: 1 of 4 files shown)


kchapple:
Thanks Stephen, I can fix that, I think no know why. I was also looking at that today! I’ll pull this down in the morning and add my changes to populated the codes.

sjpadgett:
Stephen, don't know if you were in the email chain when I sent this to Ken.
Hopefully helps
[image: cypress_results.png]

adunsulag:
@sjpadgett The image didn't upload.

sjpadgett:
[image: cypress_results.png]

sjpadgett:
how about now

adunsulag:
@kchapple Fixed the styles here in this commit.  I also stepped through the supplemental data and identified another bug that was preventing them from rendering.  They are now being populated in the cat3 reports.  Once you fix the gender, race, ethnicity, payer code values that will work (I left a big ALL CAPS todo in there for you to remove my supplemental_data override once you fix that data).

adunsulag:
@kchapple I believe you brought in your own PR with these changes?  Are we ok to just close this one, or do you still need this code?

kchapple:
Thanks for checking @adunsulag 
It can be closed.

PR #5112, opened 2022-04-08
Openemr fix 5095 fhir resources

adunsulag:
Not sure what happened but it looks like provider_uuid got stomped on in
CarePlan and Immunization.  Adding that in to our uuid mappings fixes
the problem with the resources returning 500.

For Patient it looks like the patient/* context was broken in the
Patient resource endpoint when we refactored the parameter name for
documentation purposes.  I've updated the $id parameter in the patient
route to be $uuid.


adunsulag:
I already did this merge.   Wow... mind is not working clearly today lol

PR #5113, opened 2022-04-08
Openemr cqm supplemental data

kchapple:
This builds on commits by @adunsulag to add the aggregate counts to the QRDA Cat III xml document.


adunsulag:
@kchapple the changes look fine to me.

PR #5114, opened 2022-04-08
Export Cat3 addition CCM

sjpadgett:
@kchapple


sjpadgett:
image

If All Measures All Patients is selected then a directory cat3_report is created in document directory and an xml for each measure with all patients selected for date range in export is created.

If a measure is selected then only that measure with all patients unless a patient is selected from list. If more than one patient is selected then report reverts to all patient in the report. i.e either all patients or one patient report can be generated for now.

I'll fix this up later but wanted to get in so cat3 testing can move forward.
Sorry Ken took so long. Moving slow today!

I can bring this in now or wait until I have payer done. Up to you Ken.

sjpadgett:
I have to do this:
[ALTER](http://localhost/phpmyadmin/url.php?url=https://dev.mysql.com/doc/refman/8.0/en/alter-table.html) [TABLE](http://localhost/phpmyadmin/url.php?url=https://dev.mysql.com/doc/refman/8.0/en/alter-table.html) `insurance_companies` CHANGE `ins_type_code` `ins_type_code` [TINYINT](http://localhost/phpmyadmin/url.php?url=https://dev.mysql.com/doc/refman/8.0/en/numeric-types.html)(4) NULL [DEFAULT](http://localhost/phpmyadmin/url.php?url=https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_default) NULL;
From tinyint(2)
Also, insuranceservice insert is broken.

sjpadgett:
@adunsulag you might want to take a look at InsuranceService insert. Don't see how that can work for API. The $this->doesInsuranceTypeHaveEntry($pid, $type) is at issue.

adunsulag:
@sjpadgett well that definitely looks broken on the standard api.  Looks like that first iteration of them standardizing the API for FHIR broke it back in January of 2021.  InsuranceService isn't actually used by FHIR but the standard API for sure is broken.

sjpadgett:
Okay, will put in a quick fix that should work for both.

sjpadgett:
Jeez what a pain!
I wonder if anyone is ever going to fix the duplicate companies in UI? I thought I opened an issue on this last year or has anybody else seen this? Maybe corrupt DBs!
image

bradymiller:

PR #5116, opened 2022-04-09
Various QRDA Changes

sjpadgett:
add last chance lookup for system OID from code type

Missed OID for HCPCS causing validation errors for these codes

@kchapple note.

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php
@@ -89,10 +89,20 @@ class EncounterccdadispatchController extends AbstractActionController
         $this->latest_ccda = $this->getRequest()->getQuery('latest_ccda') ?: $this->params('latest_ccda');
         $hie_hook = $this->getRequest()->getQuery('hiehook') || 0;
 
+        // @TODO uncomment below for production.
+        /*$qrda_options = [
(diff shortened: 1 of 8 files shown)


sjpadgett:
Now also save cat1 xml's to local file system for easier viewing and troubleshooting.

image

sjpadgett:
@kchapple 
We have an issue with generated extensions.
This <id extension="00000000000004190000000000000000" root="1.3.6.1.4.1.115"/> extension generally should be unique for components that reference the root(patientRole). If they're not, parsing will see the first extension and skip the next if they're the same.

In CDA we take something unique from the component data(like pid+encounter etc) then create a SHA to use for the extension.
This is a big issue for the calculator because QDMs are missed on parse. You'd think cypress would complain instead of keeping it a big secret. It's a plot started years ago just to aggravate little ole me!:)

kchapple:
Wow, nice catch. Will jump on payer cat 1 and 3 in the am.

sjpadgett:
@kchapple I just found out also that we can't go by just reason code for negation so I'm changing to looking at if OID is part of code. After this I'll bring in this PR so you can have in morning. Thanks

sjpadgett:
Ken I'm going to leave the negation fix to you. To tired tonight but my thought is something like:
```
if (!empty($record['reason_code'])) {
     if (str_starts_with($record['code'], 'OID:') === true) {
         $model->negationRationale = $this->makeQdmCode($record['reason_code']);
     }
 }
```

I'm not sure how to pass a non negated reason code in context to populate the reason template.

PR #5117, opened 2022-04-09
allow options (reporting period) to be passed down to cat i export

kchapple:
@sjpadgett I think this will allow you to pass the options into the cat i export. It was supported by the template's view-model, but we didn't have a path to get the options down to it.

Something like this will need to be added in order to get the data from the UI or database, into the exporter:

L45 of QrdaReportController.php
```
$options = [
    'performance_period_start' => 'some date',
    'performance_period_end' => 'some date'
];

$document = $this->reportService->generateCategoryIXml($pid, $measures_resolved, $options);
```

### DIFF
--- a/src/Services/Qrda/ExportCat1Service.php
+++ b/src/Services/Qrda/ExportCat1Service.php
@@ -18,13 +18,28 @@ class ExportCat1Service
         $this->request = $request;
     }
 
-    public function export($measures = [])
+    /**
+     * @param array $measures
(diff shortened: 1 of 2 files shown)


sjpadgett:
see https://github.com/openemr/openemr/pull/5116

sjpadgett:
@kchapple this is what's missing in your other current PR. I'm closing this and will add to my PR

Issue #5122, opened 2022-04-11
FHIR Inferno Single Patient Tests Failing


adunsulag:
The newest version of FHIR Inferno is failing our single patient test suite.  Its strange because it appears like some code got stomped on, or never committed, yet we had our demo servers fully passing w/ inferno so very odd how a number of these issues w/ broken code have cropped up.

There are a number of issues that I'll link out to individual sub-issues.

- #5123 
- #5124
- #5125
- #5127
- #5137 
- #5168

adunsulag:
Closing this as it is now all working.

PR #5129, opened 2022-04-11
Openemr fhir single patient api #5122

### LINKED ISSUE #5123: FHIR Invalid Ids for Goals,CarePlans,Provenance
Inferno updated their validation suite to force FHIR resource identifiers to confirm to the updated specification in FHIR R4 4.0.1.

FHIR 4.0.1 id columns put a constraint on ids such that: 
Ids can be up to 64 characters long, and contain any combination of upper and lowercase ASCII letters, numerals, "-" and ".".  Logical ids are opaque to the resource server and should NOT be changed once they've been issued by the resource server

Up to OpenEMR 6.1.0 patch 0 we used underscores as our UUID separator for Careplan and Goals due to the way the underlying data structures work.  We will switch this id to be -Sk- (meaning Surrogate Key) to separate the encounter uuid and the careplan form id.  We could go with a uuid approach like we do with vitals but that would require a rewrite of the careplan form and data structure to mirror what we do with vital signs.  Resource constraints prohibit that decision so we will change the surrogate key identifier to comply with the 4.0.1 specification.

Provenance resources originally used the format of <FHIR Resource Type>:<Resource Logical Id>, we are switching the format to <FHIR Resource Type>-PSK-<Resource Logical ID>. Where PSK stands for Provenance Surrogate Key.

However, in order to make sure we deal with existing FHIR data in OpenEMR installations we need to check the creation date of these resources and handle the old data format.  

This does get a bit tricky since we are handling older data formats as we need to use only new ids for our MU3 demos in order to pass inferno validation.  For the MU3 demos we will need to update the data to make sure they have newer timestamps on their FHIR resources so only the new logical id formats work.

### LINKED ISSUE #5124: FHIR Procedures were breaking due to SurgeryService relying on an invalid username query
Attempting to call the SurgeryService via FHIR breaks the query due to a missing 'username' column that was used for Provenance.

### LINKED ISSUE #5125: FHIR ClinicalNotes and Immunization resources were not returning provenance
Calling Clinical Notes and Immunization resource endpoints with the _revinclude=Provenance:target parameter would not return provenance resources due to the way the author resources were being populated.

adunsulag:
This PR holds the fixes to many of the sub issues in #5122.  

I will keep it as a draft for now, but it does the following:

- Fixes #5123 
- Fixes #5124 
- Fixes #5125 

I add the username to the SurgerService.

I fix all the provenance queries for DiagnosticReports,DocumentReference,Immunization,Procedure,CarePlan,Goals.

I switch Careplan and Goal to use -SK- for the uuid separator instead of an underscore(_).

I switch Provenance to use -PSK- for the id separator (since a CarePlan/Goal can be inside a Provenance logical id).  

I put in code to handle old resources for the ids to handle consumers of the API prior to this fix.

### DIFF
--- a/src/Services/CarePlanService.php
+++ b/src/Services/CarePlanService.php
@@ -30,7 +30,17 @@ use Twig\Token;
 
 class CarePlanService extends BaseService
 {
-    const SURROGATE_KEY_SEPARATOR = "_";
+    // Note: FHIR 4.0.1 id columns put a constraint on ids such that:
+    // Ids can be up to 64 characters long, and contain any combination of upper and lowercase ASCII letters,
+    // numerals, "-" and ".".  Logical ids are opaque to the resource server and should NOT be changed once they've
+    // been issued by the resource server
+    // Up to OpenEMR 6.1.0 patch 0 we used underscores as our separator
+    const SURROGATE_KEY_SEPARATOR_V1 = "_";
+    // use the abbreviation SK for Surrogate key and hyphens.  Since Logical ids are opaque we can do this as long as
+    // our UUID NEVER generates a two digit hyphenated id which none of the standards currently do.
+    // the best approach would be to completely overhaul Careplan but for historical reasons we aren't doing that right now.
+    const SURROGATE_KEY_SEPARATOR_V2 = "-SK-";
+    const V2_TIMESTAMP = 1649476800; // strtotime("2022-04-09");
     private const PATIENT_TABLE = "patient_data";
     private const ENCOUNTER_TABLE = "form_encounter";
     private const CARE_PLAN_TABLE = "form_care_plan";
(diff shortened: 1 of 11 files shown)


adunsulag:
I meant to put this as a draft PR, but accidentally made it a full PR.  I'll just keep adding to it then and let people know when its ready.

adunsulag:
@bradymiller So the Single Patient Inferno tests are now all passing once again on my local machine.   When we bring this into the demos we will verify on the public inferno instance which is what SLI uses for their tests.    For the public inferno there are a couple of data changes we will need to make:
- We need to update all of the forms table's date_created column WHERE name="Care Plan Form" to be NOW() so that we use the latest version of the FHIR logical ids.
- For Patient Alice we need to update her most recent vital forms to have 0 for height, weight, bps/bpd/temperature
- We need to set in issues Medication for Alice to link one of the issues (such as 1 ML Darboepoetin) to an encounter (2015 for example).
- We need to update the inactive Cetrifaxone Intent to be **Plan**
- We need to add one new medication (can just use Cetrifaxone) and set the Intent to **Plan**.  It needs a start date, but leave the end date empty
- We need to make sure that the first facility is set to be the primary billing facility

Once this code is merged in, I will update the demo machine and then we should take another snapshot.

bradymiller:
2DV-2

bradymiller (approved):

PR #5139, opened 2022-04-11
Fixes #5127 mapped uuids not saving in registry

### LINKED ISSUE #5127: New FHIR Vitals Observations not retrievable due to missing UUID Registry records
When we refactored the UUID mapping logic we broke retrieving new FHIR vitals observations as the mappings are generated but the uuid_registry value is not inserted into the database. This was discovered in the inferno api specifically because it attempted to retrieve a specfic observation with a data absent reason. 

We need to make it so we populate all of these missing values in the database as well as fixing the mapping uuid generation process.

Original Ticket Description:
When a vitals form is created UUIDMapping records are generated for every single column in the vitals form.  If the vitals property has a 0 value for columns such as blood pressure, etc, its supposed to generate a uuid for it. These empty values are then reported as missing with a Data Absent Reason (DAR).  However, we have a number of DAR being reported in inferno but when the resource is attempted to be fetched it returns as not found.

adunsulag:
Fixes #5127  Made it so the uuids for mapped resources such as vitals are being populated in the uuid registry.  Users will need to run the sql_upgrade.php for any existing installation in order for any new vitals to be retrievable via FHIR for the single resource GET operation.

### DIFF
--- a/src/Common/Uuid/UuidMapping.php
+++ b/src/Common/Uuid/UuidMapping.php
@@ -91,6 +91,8 @@ class UuidMapping
                 sqlStatementNoLog($insertStatement, $bindValues, true);
                 $index++;
             }
+            // now insert the mapped uuids into the registry
+            $uuidRegistry->insertUuidsIntoRegistry($uuids);
         }
         return $uuids;
     }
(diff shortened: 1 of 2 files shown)


bradymiller:
bonus: the populateAllMissingUuids function is actually called by an OpenEMR service every 4 hours, so should take care of the missing uuid's silently without the user even knowing it :)

bradymiller (approved):

PR #5145, opened 2022-04-12
Request of abbreviated bullets in the patient's history

jl1dvg:
Fixes #


library/ajax/prescription_drugname_lookup.php
```diff
@@ -39,10 +39,10 @@
     } else {
         $sql = "SELECT `name`, `drug_code` as rxnorm FROM `drugs` WHERE `name` LIKE ? GROUP BY `drug_code` ORDER BY `name` LIMIT 100";
     }
-    $val = array($term . '%');
+    $val = array('%' . $term . '%');
     if ($is_rxcui) {
         $code_type = sqlQuery("SELECT ct_id FROM `code_types` WHERE `ct_key` = ? AND `ct_active` = 1", array('RXCUI'));
-        $val = array($term . '%', $code_type['ct_id']);
```
adunsulag, line 45: Have you tested how this impacts index performance for lookups?  If someone has both SNOMED & RXCUI loaded into the table how does performance behave.  Depending on the code_types indexes this could dramatically impact performance when doing a full key substring scan compared to a key prefix scan.  

I like the change, just thinking about this from a performance standpoint since that table has a lot of entries.

PR #5146, opened 2022-04-12
Providers List Being Searchable Implementation #5019

### LINKED ISSUE #5019: Providers list being searchable
**Describe the problem**

The Providers list under the calendar and within the appointment scheduling can get out of hand when you have 50+ providers listed there. With no way to manage or sort through them quickly.

https://community.open-emr.org/t/provider-list-search-function/18037

**Describe the solution you'd like**

Is there anyway to make this provider list into a searchable with a search box.

infeg-khaleela:
Fixes #5019 
- Providers list made searchable (using select2) in add appointment popup window
- Providers list made searchable (using select2) in calendar left side provider selection


monte213 (commented):
Thanks again for this.

The add_edit_event code works nicely. This is exactly what was needed to be able to add providers efficiently.

The calendar display is not quite as useful though. For non-authorized (non-provider?) roles, the default functionality loads all the providers as selected so when we have many, the display is not ideal. Also with the select2 in place, if you want to display the calendar of just one person, you would then need to un-select all the other providers.

Ideally, having the option of the Calendar initially displaying with no providers selected for non-authorized roles might be the ideal solution because then the select2 fits in perfectly as a lookup.

I don't know PHP or the openemr codebase well enough to come up with an elegant solution but added at line 123 of ajax_template.html to check that if the provider array was had a length over 5 (arbitrary number) then set it to empty. 

if (count($providers) > 5) $providers = [];

This helped handle that initial display for non-authorized users by displaying an empty Calendar and then the select2 worked nicely here.

I'm not sure if that type of solution is on the right track where the display max number could be parameterized or something. Or if there was otherwise another way to not initially populate the providers array on login for non-authorized users, that might be better. I'm not sure.

Hope these comments were helpful! Thanks again

PR #5147, opened 2022-04-12
Remove unneeded globals.php from libraries

bradymiller:
Remove unneeded globals.php requires from libraries
(in process of testing)

Did search of entire codebase for both the content_parser.php and api.inc and global.php is always required before these libraries are used (this is the standard pattern in openemr where globals.php is required and then the libraries after that in scripts).

I did note one place where globals.php was not brought in prior to api.inc in track_anything/report.php , which I added. However, this points out a good issue and that is that all the report.php files in all the forms should not need to require globals.php , but that is code cleanup for another day :)

### DIFF
--- a/interface/forms/CAMOS/content_parser.php
+++ b/interface/forms/CAMOS/content_parser.php
@@ -12,7 +12,6 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
-require_once(dirname(__FILE__) . "/../../globals.php");
 require_once(dirname(__FILE__) . "/../../../library/api.inc");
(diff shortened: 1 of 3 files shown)


bradymiller:
testing well

stephenwaite:
Nice! Sounds like a good plan.

bradymiller:
reminder to self:
bring in following commit into rel-610 after #5153 is brought into rel-610:
https://github.com/openemr/openemr/commit/c28caceca1ddc59193dd890900a4f703a1414925

stephenwaite:
adding c28caceca1ddc59193dd890900a4f703a1414925 to my `610` PR

bradymiller:
@stephenwaite , thanks!

PR #5150, opened 2022-04-12
api updates related to recent transactions endpoints

bradymiller:
api updates related to recent transactions endpoints

- drop the transaction_type since not functional and would need a scope
- rename from /api/patient/transaction/{tid} PUT endpoint to /api/transaction/{tid} since there is no patient info given (ie. no pid or puuid mapping)
- added 'body' element to swagger POST/PUT examples
- added the new scopes to other places (examples, docker stuff)


bradymiller:
endpoints are testing well on swagger

PR #5151, opened 2022-04-13
Fix patient scopes when used by users

### LINKED ISSUE #5157: FHIR patient context scope not working when used non-patients
So we discovered that we are not handling the required use case where a non-patient wants to have their SMART on FHIR application bound to a patient context.  So if a Clinician wants to use an application that is only intended to be a single patient application (IE only looks at one patient's data at a time) they cannot currently do so.  A single patient application will only use patient/* scopes instead of user/* scopes.  When we originally built OpenEMR's FHIR implementation we assumed that apps intended for providers would use user/* scopes and patient apps would use patient/* scopes.  

However, going through the ONC test suite we discovered we misinterpreted the specification and provider's can/will use applications that use patient/* scopes.  When a patient/* scope is used it needs to be bound to the patient that is currently selected in OpenEMR.  If no patient is currently selected (and the Application uses the launch/patient scope) then OpenEMR needs to provide a patient selector and make sure all resources governed by a patient/* scope are only allowed to see that specific patient's data.

We need to make sure we still keep the patient bound context when renewing access tokens with a refresh token (when the offline_scope is present for an application to get a refresh_token).

adunsulag:
Fixes #5157 

We need to handle in the standalone application context the ability to use patient scopes while logged in as a practitioner.  We originally restricted patient scopes to JUST patients and user scopes to JUST users.  However, latest inferno and v2 of smart on FHIR is expecting a practitioner to be able to select a user and have their scopes
restricted to JUST that user.

Currently I have this working when used for a single patient.  I did this by overriding the http request's isPatientRequest and getPatientStringUUID properties to populate these values IF we are in a user/system context and the launch/patient scope is present (meaning the user has requested to have a patient context bound) and the resource endpoint being requested has a patient scope (but no overriding privilege of a user/system scope).

I made it so that each access token has a uniquely bound session context rather than using a shared session context that was populated from the trusted user session cache.  Now SMART applications can have different patients open using different access tokens for the same application without stomping on each other's data.

Modified refresh token grant so we can pull our access token context and repopulate it when a new access token is issued from the refresh token.

Refactored IdTokenSmartResponse to pull the Access Token response context variables into its own class called SMARTSessionTokenContextBuilder that is used to build the context information.  This is needed to deal with patient bound contexts and contexts that are re-issued via refresh tokens.

Changed up AccessTokenRepository to save off the access token context information (such as a selected patient, encounter, etc) that needs to be retrieved / filtered for each access token (or reissued when a refresh token
generates a new access token).

Changed up AuthorizationController,SMARTAuthorizationController to set the session pid to be the actual patient pid and not the uuid of the patient. Added the puuid to the session.

Modified dispatch.php to call RestConfig to populate the access token context instead of grabbing it from the trusted_user session_cache.

### DIFF
--- a/src/RestControllers/SMART/SMARTAuthorizationController.php
+++ b/src/RestControllers/SMART/SMARTAuthorizationController.php
@@ -124,8 +124,8 @@ class SMARTAuthorizationController
      */
     public function needSMARTAuthorization()
     {
-        if (empty($_SESSION['pid']) && strpos($_SESSION['scopes'], SmartLaunchController::CLIENT_APP_STANDALONE_LAUNCH_SCOPE) !== false) {
-            $this->logger->debug("AuthorizationController->userLogin() SMART app request for patient context ", ['scopes' => $_SESSION['scopes'], $_SESSION['pid']]);
+        if (empty($_SESSION['puuid']) && strpos($_SESSION['scopes'], SmartLaunchController::CLIENT_APP_STANDALONE_LAUNCH_SCOPE) !== false) {
+            $this->logger->debug("AuthorizationController->userLogin() SMART app request for patient context ", ['scopes' => $_SESSION['scopes'], 'puuid' => $_SESSION['puuid'] ?? null]);
             return true;
         }
         return false;
@@ -167,7 +167,7 @@ class SMARTAuthorizationController
             // throws access denied if user doesn't have access
             $foundPatient = $searchController->getPatientForUser($patient_id, $user_uuid);
             // put PID in session
-            $_SESSION['pid'] = $patient_id;
+            $_SESSION['puuid'] = $patient_id;
 
             // now redirect to our scope-authorize
             $redirect = $this->smartFinalRedirectURL;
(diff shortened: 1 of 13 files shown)


src/RestControllers/SMART/SMARTAuthorizationController.php
```diff
@@ -162,9 +162,10 @@ public function patientSelectConfirm()
...
         // set our patient information up in our pid so we can handle our code property...
         try {
-            $patient_id = $_POST['patient_id'];
+            $patient_id = $_POST['patient_id']; // this patient_id is actually a uuid.. wierd
             $searchController = new PatientContextSearchController(new PatientService(), $this->logger);
-            $searchController->getPatientForUser($patient_id, $user_uuid);
+            // throws access denied if user doesn't have access
+            $foundPatient = $searchController->getPatientForUser($patient_id, $user_uuid);
             // put PID in session
             $_SESSION['pid'] = $patient_id;
```
bradymiller, line 170: this inconsistency will likely be problematic in backend openemr stuff if the session pid is set to the puuid. For things to work (backend code), this should be the pid and not the puuid.

sjpadgett: Probably should store both if session is referenced for the uuid at some point in transaction. Not sure why that'd be the case though....

bradymiller (approved):

PR #5172, opened 2022-04-16
fixed sanitization vulnerability in users form

VarunVeginati:
Fixes #


bradymiller:
hi @VarunVeginati , thanks for the PR!
Is this fixing a vulnerability?
Meaning is there anything that can be brought in without the filter_var() sanitization that causes a actual vulnerability in OpenEMR?
OpenEMR generally does the html escaping when outputting to html rather than before storing in the database (since never know where data is going to be used; ie. html, xml, json, straight database collection).

PR #5179, opened 2022-04-18
Fixes #5178 FHIR Medication endpoint

### LINKED ISSUE #5178: FHIR Medication resource needed to pass Medication.read whitelisted scopes
Our scope whitelisting prevents the Medication.read resource from being used because there is no corresponding Medication api endpoint.  We need to implement a basic version of the FHIR Medication endpoint so that Inferno will handle the scope correctly.

adunsulag:
Implements the Medication endpoint for FHIR.


bradymiller (approved):

bradymiller:
back from the grave :)

zombie-hand

PR #5182, opened 2022-04-19
Add 2 digit language codes for FHIR

### LINKED ISSUE #5183: OpenEMR use 2 digit ISO-639-1 when available instead of 3 digit ISO 639-2
Inferno validation failed for patient language communication value set.  'eng' wasn't in the allowed code sets and digging into this we are supposed to use 2 digit codes from ISO 639-1 if they are available and fall back to the 3 digit ISO 639-2 codes if we don't have a two digit code.  This is what IANA  officially requires and inferno validates against.  This is also what CCDA 2.1 uses as well.

We will change up our list_options in the notes column to deal with these changes.

adunsulag:
Fixes #5183 
Add in the 2 digit language codes to the codes column so we can pass
ccda and FHIR validation. 2 digit is required to be used if one
available otherwise we use 3 digit language codes.


sql/6_1_0-to-7_0_0_upgrade.sql
```diff
@@ -315,3 +315,196 @@ ALTER TABLE `form_care_plan` CHANGE `date` `date` DATETIME NULL DEFAULT NULL;
...
+('lat','la'), 
+('lav','lv'), 
+('lim','li'), 
+('lin','ln'), 
+('lit','lt'), 
+('lub','lu'), 
+('ltz','lb'), 
+('mac(B)|mkd(T)','mk'), 
+('mlg','mg'), 
+('may(B)|msa(T)',''),
```
bradymiller, line 425: this is odd that would make it blank

bradymiller: looks like should be ms

adunsulag: My bad, apparently I missed one... its why I shouldn't merge my own PRs haha.  It should be 'ms'.  I get all the values from here: https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry where I put in the subtag.

adunsulag: I'll push another PR to fix it.

PR #5184, opened 2022-04-19
Openemr fix language codes - Missing Malay Language

### LINKED ISSUE #5183: OpenEMR use 2 digit ISO-639-1 when available instead of 3 digit ISO 639-2

adunsulag:
Fixes #5183 Malay language code needs to be set as well.


bradymiller:
don't worry, i'll add the ms to my pr coming up soon (no need to add another if since the fix is very close to when it got into codebase)

adunsulag:
@bradymiller haha, well I didn't get your message in time.

PR #5186, opened 2022-04-19
Fixes #5181 CareTeams,Minor FHIR bug fixes

### LINKED ISSUE #5181: FHIR CareTeam role uses old value set
US Core changed the value set that can be used for role.  It no longer uses NUCC and instead uses the Care Team Member Function( https://vsac.nlm.nih.gov/valueset/2.16.840.1.113762.1.4.1099.30/expansion) which is supposed to bring it in alignment with the CCDA format.  Change is noted here: https://jira.hl7.org/browse/FHIR-27769.

It comes up as a warning on Inferno but I'm going to guess its going to fail testing if we don't fix it.

adunsulag:
Changed up the care team value sets since we can't use NUCC anymore and
instead need to use a SNOMED-CT valuset for the care team provider.
Oddly there is no 'valid' value for facilities as the role is all
provider centric.  However, SNOMED-CT still has values for health care
organization so we use that for our code.

Added some fixes to get rid of warnings with the smart session context
builder.

Also Changed up the group export to use getOne which is what we really
want instead of hitting getAll.

### DIFF
--- a/src/Common/Auth/OpenIDConnect/SMARTSessionTokenContextBuilder.php
+++ b/src/Common/Auth/OpenIDConnect/SMARTSessionTokenContextBuilder.php
@@ -25,7 +25,7 @@ class SMARTSessionTokenContextBuilder
     private $sessionArray;
     public function __construct($sessionArray = array())
     {
-        $this->sessionArray = !empty($sessionArray) ? $sessionArray : $_SESSION;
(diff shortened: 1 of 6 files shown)


bradymiller (approved):

bradymiller:
2DV

PR #5189, opened 2022-04-19
Fixes #5182 set correct patient language system

adunsulag:
The system property was not set correctly for the patient language.
Went and found inferno's sample patients and changed the system property
to be what it has listed.
Fixes #5182


bradymiller (approved):

bradymiller:
tenor99

PR #5199, opened 2022-04-20
Fixes #5190 DocumentReference categories

### LINKED ISSUE #5190: FHIR DocumentReference invalid system of openemr
So Inferno doesn't like the fact that we fall back to an OpenEMR default category type if we can't find anything else.  Although it states the code type is expandable it just doesn't like this.  

Here is  a sample error:
```
Line #10: Unknown Code System: /fhir/ValueSet/openemr-document-types
```

I think what we do to solve this problem is add either to the categories table or to the documents table itself a codes column that users can then provide a coded value that we can return.  We can still fall back to our default OpenEMR value set but this way we can give users the options of provide better structured data.  My recommendation for our MU3 demos is we use the code sets from Loinc's document ontology system here: https://loinc.org/document-ontology/current-version/  I'm personally in favor of a lot of the Kind of Document codes, but if we give a code selector than users can pretty much use any code system (SNOMED, LOINC, etc).

Later today I'll give it a test run and hard code it to a LOINC category and see if that passes validation, if it does then we can do the code selector.

adunsulag:
I have added a codes column to the category table for documents so you
can set a specific LOINC document category code to a documents category.

The LOINC codes come from https://loinc.org/document-ontology/current-version/

I made it so the codes column is optional.  I also fixed up a bit of the
CategoryTree code to decouple itself from the CategoryTree class.

I set a couple of different LOINC codes for categories.  There are more
that fit but this at least gets us through MU3 validation.

I had to completely drop the openemr type and instead report a Data
Absent Reason if a document category doesn't have a LOINC code.

In order to get the codes selector to work I switched the document
categories over to a twig file.  Couldn't get the javascript to execute
properly in smarty and it was a simple enough file to port.  So one less
smarty file to deal with.


sql/6_1_0-to-7_0_0_upgrade.sql
```diff
@@ -513,6 +513,12 @@ DROP TABLE lang_updates_610;
 UPDATE list_options SET notes='ms' WHERE list_id='language' AND option_id='malay';
 #EndIf
 
+#IfMissingColumn categories category_codes
+ALTER TABLE categories ADD COLUMN `codes` varchar(255) NOT NULL DEFAULT '' COMMENT 'Category codes for documents stored in this category';
+UPDATE categories SET codes='LOINC:LP173418-7' WHERE name='Advance Directive';
+UPDATE categories SET codes='LOINC:LP173421-1' WHERE name='FHIR Export Document';
+UPDATE categories SET codes='LOINC:LP173394-0' WHERE name='Reviewed';
+#EndIf
```
bradymiller, line 521: would put this block at the bottom of this file (then avoid issues in ordering of the database changes)

templates/document_categories/general_list.html
```diff
@@ -1,62 +0,0 @@
-{literal}
- <style>
```
bradymiller, line 2: bye bye smarty!

tenor (4)

templates/document_categories/general_list.html.twig
```diff
@@ -0,0 +1,82 @@
+<style>
```
bradymiller, line 1: hello twig!

2DV-2

sql/6_1_0-to-7_0_0_upgrade.sql
adunsulag: Interesting, the rebase must have stuck it in there, because it originally was at the bottom of the file :)

bradymiller:
looks great! only note minor issue in sql upgrade.

bradymiller (approved):

templates/document_categories/general_list.html.twig
```diff
@@ -0,0 +1,82 @@
...
+</script>
+<script src="{{ webroot }}/library/js/CategoryTreeMenu.js?v={{ V_JS_INCLUDES }}"></script>
+<table>
+    <tr>
+        <td height="20" valign="top">{{ "Document Categories"|xlt}}</td>
+    </tr>
+    <tr>
+        <td valign="top">{{ tree_html}}</td>
+        {% if message %}
+        <td valign="top">{{ message|text}}</td>
```
stephenwaite, line 41: awesome work! does this need a space after `|text }}`?

PR #5200, opened 2022-04-20
Openemr fix #5191 refresh token

### LINKED ISSUE #5191: FHIR Refresh Token set patient context
Inferno is giving a warning saying that refresh token is missing context.  Need to populate session context for the refresh token even if the user provides no scopes or a subset of the scopes.

We were keying off the launch/patient scope in the access token to renew scopes in the refresh token.  However, if a refresh token is requested and its missing the launch/patient scope apparently per inferno we still need to bind up the context values into the access token.

adunsulag:
Fixes #5191 

Made it so the refresh_grant will provide patient context and other
SMART context variables if scopes are omitted or if a subset of scopes
are sent.

Had a few resources that didn't have scope permission messages setup which this commit
fixes.


bradymiller (approved):

bradymiller:
tenor (67)

PR #5210, opened 2022-04-20
QRDA Report Service work

sjpadgett:
- implement node start service for cat3 report
- cleanup unused admin functions. They need to be in an admin class.

@kchapple note


sjpadgett:
still more validation needs to be done. Hopefully w/o needing a schematron

PR #5218, opened 2022-04-21
fix hex conversion of PID

kchapple:
@sjpadgett code was double-converting to hex, so all IDs were the same. This is the fix.


sjpadgett:
well if one conversion is good then two must be better!:) but guess not...

PR #5233, opened 2022-04-23
Fixes #5232 android oauth2 authorization issues.

### LINKED ISSUE #5232: FHIR Android Native App User Display and nonce scope
Android native apps using the Oauth2 Native App best practices specification utilizes the nonce scope that is bundled into the identity token and send as part of the refresh token.  However, our ScopeRepository fails the nonce scope check on the receiving side.

Also on the Android native apps the browser view gets distorted with the flexbox and h-100 class styles. It essentially cuts off half of the page and those scopes are unselectable. Have to rework the CSS so it looks good on native mobile and for larger screen sizes.

adunsulag:
#5232
Fixes the scope nonce validation so native apps can authorize using the
nonce from the identity token.

Fixes the css view on mobile native devices so we can see the scope
authorization screen without half of it cut off.

### DIFF
--- a/oauth2/provider/scope-authorize.php
+++ b/oauth2/provider/scope-authorize.php
@@ -67,15 +67,15 @@ ksort($scopesByResource);
     <title><?php echo xlt("OpenEMR Authorization"); ?></title>
     <?php Header::setupHeader(); ?>
     <script src="<?php echo $GLOBALS['webroot'] ?>/library/js/u2f-api.js"></script>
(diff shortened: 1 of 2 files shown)


adunsulag:
This is a very minor change and shouldn't impact any of the other codebase.  I've already run through inferno on my local box to make sure this works.  If the tests pass (which they should), I"ll just go ahead and merge this in so we can get it into the MU3 tests.  That way I can point the android native app application against the MU3 demos.

bradymiller:
2DV-2

PR #5261, opened 2022-04-29
LBF  Bugfix - Allow Label Size 0 in First Column

### LINKED ISSUE #5240: Incorrect Display of LBF Used by Demographics
### Describe the bug
For the demographics LBF layout, in a field:
If you set label columns to 0 and it is in the first column, the label still uses the first column and the data starts in the second column.  

However, this only occurs if you display the demographics.  If you edit demographics, it is displayed properly.

### To Reproduce
–insert a field of datatype textbox into demographics layout
–leave label blank
–label cols = 0
–data cols = 4
–options = “Jump to next row”
-save changes
–go to demographics, enter info into new textbox, and click save.

In display mode, the textbox data is in column 2 instead of column 1. (wrong)
In edit mode, the textbox is in column 1. (correct)

I also, recreated the problem on the development demo, changing the layout of social security number to label cols = 0 and adding option jump to next row.

### Expected behavior
In display mode, the textbox data should be in column 1, not column 2.

### Client configuration
- Browser:  Firefox
- OpenEMR version:  Latest Development Version

The problem is with the display of LBF used by demographics, likely in the function "display_layout_tabs_data" in the file /library/options.inc.

psoas:
Fixes #5240

Currently, a label is always in the first column.  If you specify label size of 0 columns (i.e. do not have label), the code changes the size to 1 column.  

The fix allows the label size to remain at 0 if specified in the LBF.

I accidentally created an openemr module in my commit and I could not figure out how to remove it.  Please remove it for me.

### DIFF
--- a/library/options.inc.php
+++ b/library/options.inc.php
@@ -3841,10 +3841,6 @@ function display_layout_tabs_data($formtype, $result1, $result2 = '')
                     echo "<tr>";
                 }
 
-                if ($item_count == 0 && $titlecols == 0) {
-                    $titlecols = 1;
-                }
-
                 // Handle starting of a new label cell.
                 if ($titlecols > 0) {
                     disp_end_cell();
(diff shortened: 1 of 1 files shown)


sjpadgett:
we can't remove your submodule. there's not reason for you to commit any composer or package files or even your entire openemr dev tree.

I don't trust these changes because handling column and row 0 has been around forever. I don't recall anyone changing recently to break but could be wrong.

@sunsetsystems Would you look this over for me and give your opinion. That is, if you're not on a walk in the out back!:)

sunsetsystems:
The design is that when you specify 0 for label cols (or data cols), it continues to use the current column. It does not mean that you have no label. To omit a label make it an empty string, then specify 0 for the following data cols field. Specifying 0 for the first width in the layout is not valid.

sjpadgett:
> The design is that when you specify 0

Thanks Rod @sunsetsystems for helping clarify. This feature should not be altered unless for a very good reason. The main concern is legacy user custom LBF and of course, are own.

David, Modifying the normal row/column parsing to make your new template render the way you need it to is not the answer.
Unsure what the answer is and unless you open a draft working PR with your new code so we can get context to help, I don't think this PR will get traction.

This is not a bug and works as designed the best I can tell w/o a deep dive.

sunsetsystems:
I can see how this area is confusing. It would be great if there's someone who wants to work on documentation improvements.

psoas:
Thanks Rod @sunsetsystems and Jerry @sjpadgett:

Rod's suggestion to specify an empty string for the label and make data cols 0, made me think of trying the following:
--empty string for label
--label cols 4
--data cols 0

And, this allowed me to do what I want to do to properly display my custom data type.
(Previously, I tried label cols 1 and data cols 0, which did not work.)

So, I no longer feel that I "need" to do my modification of the LBF code in this pull request.

HOWEVER:
I still don't see why "Specifying 0 for the first width in the layout is not valid."  

Does allowing this break something elsewhere?  Is it incompatible with a particular datatype?

And, I did do a deep dive into the code and my change is so minor, that I can't see it affecting anyone's LBF's, unless they specify label size 0 (and my change will help, albeit in a minor way, by preventing a small indent in front of the data field and empty lines at the top of the form that should not be there.)

My modification of the code only does the following:
--deletes the part that automatically changes label cols from 0 to 1.
--moves code that echo's the second half of the label related html into the "if" statement that echo's the first half (which is where it should have been in the first place).  
FYI: The need to change label cols from 0 to 1 was to avoid display errors caused by not having all the label display code within the "if" statement.

No new code was created in my commit, so I doubt it would cause errors in anyone's custom LBF forms.

I wouldn't mind if Rod looks at this and considers it.  

And, I also wouldn't mind more explanation why my change would cause harm and why label cols 0 would not be valid.

BUT AGAIN:  I am not wedded to my modification despite the fact that I think it is appropriate.

sunsetsystems:
Hi David, my main point is that label cols = 0 is for continuing the label in the previous column, and for the first field there is no such column yet. It sounds like my suggested solution is imperfect for you because a tiny amount of space is allocated for the empty label? I haven't studied your change in detail but I think you'd want a general solution for empty labels at the beginning of a column, not just the first one in the layout.

PR #5262, opened 2022-04-29
Prcedures and Other

sjpadgett:
add conditional to cvx_code in service

### DIFF
--- a/src/Services/Qrda/QrdaReportService.php
+++ b/src/Services/Qrda/QrdaReportService.php
@@ -27,6 +27,8 @@ class QrdaReportService
     protected $builder;
     protected $calculator;
     protected $patientJson;
+    protected $effectiveDate;
+    protected $effectiveDateEnd;
     public $measuresPath;
 
     public function __construct()
@@ -35,15 +37,25 @@ class QrdaReportService
         $this->client = CqmServiceManager::makeCqmClient();
         $this->client->start();
         $this->builder = new QdmBuilder();
-        $this->calculator = new CqmCalculator($this->builder);
+        $this->calculator = new CqmCalculator();
         $this->measuresPath = MeasureService::fetchMeasuresPath();
         $this->patientJson = "";
+        $this->effectiveDate = trim($GLOBALS['cqm_performance_period'] ?? '2022') . '-01-01 00:00:00';
+        $this->effectiveDateEnd = trim($GLOBALS['cqm_performance_period'] ?? '2022') . '-12-31 23:59:59';
+        if (empty($this->client->getHealth()['uptime'] ?? null)) {
+            $msg = xlt("Can not complete report request. Node Service is not running.");
+            throw new \RuntimeException($msg);
+        }
     }
 
+    /**
+     * @param $scope
+     * @return array
+     */
     function fetchCurrentMeasures($scope = 'active'): array
     {
         $measures = [];
-        $year = trim($GLOBALS['cqm_performance_period'] ?? '2022');
+        $year = trim($GLOBALS['cqm_performance_period'] ?: '2022');
         $list = 'ecqm_' . $year . '_reporting';
         $active = $scope == 'active' ? 1 : 0;
         $results = sqlStatement("SELECT `option_id` as measure_id, `title`, `activity` as active FROM `list_options` WHERE `list_id` = ? AND `activity` >= ?", array($list, $active));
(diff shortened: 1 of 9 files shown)


sjpadgett:
Retested CMS127v10 to verify changes to service and  tested green for cat1 and cat3.

I still can't figure out why cypress send good patients and yet not expect them back. Passes all de-dupl ckecks and XSD and visual for errors.
Until this is figured out I may open blocking issue for testing... Others must have seen this but so far I don't see an issue on Jira

stephenwaite:
for 127v10 there are 2 "good" patients in the list who don't meet the age criteria, 
see how Vickie Hart is only 12 and the min age is 66
Screenshot from 2022-04-29 17-10-16

sjpadgett:
That may be true @stephenwaite however, those become an exclusion in the calculator of which is part of the test i.e patients with various negations that may/will exclude the patient from the overall population of patients for which the analytics are concerned.

sjpadgett:
Although the other patient i'm not to include in returned patient is Allison, Waters who also doesn't meet age requirement.

This can't be true we have to qualify calculator requirements for patients against what measure is being reported. But you may be onto something however, if this is true, requires much thought on how we do this.

sjpadgett:
@kchapple Is there a way to qualify patients based on measure being reported. For example, picking what patients from the EHR get chosen for report.
Thinking about this it makes prefect sense and may answer my question. Thanks @stephenwaite !

kchapple:
If they are imported via cat I, then there is a measure section. We could pull data from that and we could stick the measure in the database somewhere per patient. That doesn’t help cat I validation though… unless some patients don’t have a measure section, or if we run the calculation on import and validation that they meet initial patient population criteria.

sjpadgett:
> they meet initial patient population criteria

That'd have to be the solution for import validation unless we hard code the various conditionals required per measure. Still hard code may be only way to be able to select patients to report on in emr.

kchapple:
There is a RequestSome implementation of RequestInterface that uses MySQL “in” to calculate on a set of patients, but you need to know which patients before you pass them in, so you’d still need some importing of the measure data from cat 1.

kchapple:
There’s RequestOne, RequestAll and RequestSome

sjpadgett:
@kchapple @adunsulag Somewhere in cypress project the below is parsed out. So far I haven't been able to find but perhaps one of you guys can take a look see. There must be a way short of running all patients against a measure to determine if patient should be included. Perhaps the conditions are part of the measure json object. Anything short of hard coding.

image

kchapple:
Those are calculator results. They come back from the cqm service. If the patient qualifies for IPP, then it will have a value of 1 (or maybe 2) for IPP in the result. It would have 0 if doesn’t qualify. Then the statement results have all the gritty details of which sections the patient pass/fail.

sjpadgett:
So if I send in all measures against a patient I could use the IPP to qualify the patient for measure? Maybe?

sjpadgett:
So I'm thinking I should be able to grab calculator result then iterate through the relevance(population and/or statement results) for any trues and verify true/false result to identify measure validity to patient.  
What do you think @kchapple?

adunsulag:
So I'm trying to read up on the specification for this as it sure looks like the CQL already does all of the patient filtering (which is where you get the IPP, DENOM, NUM (and corresponding stratifications) and you would just send the entire patient population through the calculation if there is a provider / clinic relationship with the patient (either through encounter or primary care provider designation).  

If I can find anything useful I'll post about it.

kchapple:
@sjpadgett that sounds the only option, unless they were imported via cat 1 and we import the measure section along with the patient record.

sjpadgett:
Thanks guys, I think I have it and will commit a start you guys can review soon.

sjpadgett:
I wish we would have used the shift dates so our reporting periods matched what globals reflect i.e 2022. this way wouldn't have to worry with passing darn dates around from import dates!

adunsulag:
@sjpadgett Is that still an option?  SLI won't create the product until 6.3.0 is released.

sjpadgett:
Yep since she hasn't set up test yet we should be able to but i'm going to set up for production where I already have a global for reporting period. All I need to do is change default to 2020 and we can keep existing code. I'll just use global where we can change to meet whatever period we need. That way I only have one place to change for production and not touch core code.
I guess this makes sense!:)

sjpadgett:
Turned out to work with at least CMS127. On export I run patient through calculator and tee off IPP.
I need to create a log trace to give back to user so they know what patients get excluded.
I also refactored all period dates down to where they get used and have everything setup to turn on production.
This will mean we'll request SLI to shift patients which I'm pretty sure she'll not have an issue with due to we're still a week away from cypress upgrade.
@bradymiller will you make that request or do you want me to do?
@stephenwaite and @adunsulag I'll create a new product with shifted records for you guys to test with.

Bottom line, after I make this PR whole we won't have to worry about doing changes to code after testing to go to production.

I started from scratch with required download, de-dup'ed and using our core patient duplicate and merge, ran test on CMS127.

image

sjpadgett:
@adunsulag where does SystemLogger logs get stored and can I use as independent log for just cqm?

sjpadgett:
image

sjpadgett:
and spot check cat3

image

sjpadgett:
Check this out folks. Start to finish CMS22v10 about 15 mins with shifted records and refactored code.
Just need to figure out logging.

image

adunsulag:
@sjpadgett System logs is currently configured to store in the error_log file.  We can modify it to add handlers / replace existing handlers if you want to store the logs in the database, or have it log to a separate file.  Just look at the _construct method.  Its very similar to winston in Node.JS if you are familiar with that.

sjpadgett:
Can now search and qualify for patients in export.

image

Resulting unqualified patients against selected measure reports in ren and won't have a select checkbox for export.

image

Someone please get me out of this darn CCM!:) 
Anyway just need to do an end report of test data to give to SLI and once we get c(1), we should be able to go to test! This has all been a beast....

sjpadgett:
bringing in

src/Services/Qrda/QrdaReportService.php
```diff
@@ -86,23 +98,56 @@ function resolveMeasuresPath($measures): array
...
+        } else {
+            $request = new QdmRequestAll();
+        }
+
+        $exportService = new ExportCat3Service($this->builder, $this->calculator, $request);
+        $result = $exportService->export($measures, true);
+        $include = array_shift($result);
+        if ((int)$include[0]->IPP === 0) {
+            error_log(errorLogEscape(xlt('Patient did not qualify') . ' pid: ' . $pid . ' Measures: ' . text(basename($measures[0]))));
+            return false;
```
stephenwaite, line 139: excited_elephant

PR #5319, opened 2022-05-11
Validation service

### LINKED ISSUE #5317: Install mechanism for our Node Services
Our latest development cycle has sort of forced us into having to develop a post install for our ever growing number of node services.

Currently for production needs we have:
- oe-cqm-service used by our QRDA III calculator.
- ccda service used to generate various ToC C-CDA documents
- oe-schematron-service a new service I'm currently working on.

For the future I'd expect this list will grow. For oe-cqm-service note it is currently called in following. Also note that even though we do copy the service to public on build, all the other dependencies to support are not:
```
class CqmServiceManager
{
    public static function makeCqmClient()
    {
        $servicePath = $GLOBALS['fileroot'] . DIRECTORY_SEPARATOR .
            'node_modules' . DIRECTORY_SEPARATOR .
            'oe-cqm-service' . DIRECTORY_SEPARATOR .
            'server.js';
        $client = new CqmClient(
            new System(),
            $servicePath,
            'http://localhost', // should be 127.0.0.1 loopback
            '6660' // ccda service runs on 6661
        );

        return $client;
    }
}
```

sjpadgett:
Fixes #5317 

We are required to run our imports through a schematron and be able to report to the user any errors found.
Though this requirement is mainly for b(1) C-CDA I've design for QRDA reports as well. This is automatic based on document type OIDs.

I was able to companion the new schematron service into the ccda service NPM install where they run in the same directory.
Ccda service on port 6661 and Schematron on port 6662. 

We also have the oe-cqm-service that requires node modules and runs on port 6660. I'm looking at pulling this package from core package json and maybe including with ccda service install as well but, I must work in baby steps!:)


sjpadgett:
@bradymiller The vox.xml file is 65mb raw but zips down to less than 6mb. This file is important to schematron and is mainstay of detecting wrong or missing mood codes, template type codes and various required value sets. 

I'm half tempted to make all this a module or a separate service but I'm running out of time.

sjpadgett:
reverted the new schematron service and will create a service in our repository. Mostly to avoid the large files required and may have to fork the validator javascript project because of deprecated dependencies. This all nuts.

sjpadgett:
@bradymiller My last commit has the interesting stuff.
I consolidated all our required services into the ccdaservice directory. This makes it one npm install for all the services.
I plan to move new services off my repo to openemr and create a new packages in our npm account after more testing and before release.

The schematron is optional after testing I've been told so we may decide not to release.
I've retested cat1 and cat3 with green so should be good to go.
Let me know your thoughts.

ccdaservice/package.json
```diff
@@ -12,6 +12,12 @@
         "lodash": "4.17.21",
         "net": "1.0.2",
         "uuid": "8.3.2",
-        "xmljson": "0.2.0"
+        "xmljson": "0.2.0",
```
bradymiller, line 15: @sjpadgett , super cool stuff! and completely agree with keeping the services in the separate ccdaservice node_modules.
can the oe-cqm-service entry from main package.json now be removed?
at some point will need to create another official npm package for oe-schematron-service :) 

2DV-2

sjpadgett: yep will remove from core package. Have to register new npm packages because most times user won't have git installed.
Now I just need to remember how!:)

sjpadgett:
@bradymiller 
This is ready if you want to do a once over for me.
I created oe-schematron-service and oe-cda-schematron repo's and published to NPM. Did npm remove of our current oe-cqm-service and added the install to ccdaservice install.

Installs surprisingly fast however, I'm curious to see if dev demo handles. If late, bring in if approved.

sjpadgett:
@adunsulag 
image

bradymiller:
nice stuff!

PR #5325, opened 2022-05-13
Openemr fix 5284 fhir docref - WIP

adunsulag:
Added the infrastructure for the $docref operation.  Added scopes, rest
endpoint, and service classes.

Moved the Export operation into a sub folder where we can keep our future operations that we support.

Implemented generating a ccda via an event in the system so fhir can create
the ccda for the $doc-ref FHIR operation.

Refactored the ccda generation pieces from the Encounterccdadispatchcontroller
into separate classes so we can call it from the event cleanly without having
to deal with a ton of POST / controller logic code.

Refactored the ccda document generation so that it uses the document class for
encryption, couchdb storage, and retrieval.  This means a generated ccda now
generates a file in the patient CCD/CCDA folder.

Implemented a view ccda event that allows the Carecoordination module to sort
and truncate the ccda sections for the ccda that is being viewed.  This is
necessary in order to meet the b1 requirements of a user being able to set
how many sections they want to see as well as the sort order of the sections.

Began the work of adding a new globals field for displaying a list of options
and adding that list to the globals.  The options are saved in the globals
table as a semi-colon delimited string.

Added the beginnigs of the ccda-sections list options that people can select
from in the globals user settings table.

Modified the FHIR Document Binary api to use uuids instead of ids for the
document. Made the ids backwards compatable so old Documents using the id columns will
still work.

Added a method to the CodesTypeService to parse codes into codeable concepts
so we can use them in FHIR.

Updated the CCR/display.php file to use the user preferences and ccda view
event so that we display CCDs with the newest stylesheet that @sjpadgett
implemented.

Added to the CCR/display.php checks to make sure the document exists and that
the ACLs are being properly checked.

Added documentation to some of the Carecoordination modules.

Added to the GlobalsService the addUserSpecificTab method in order to be able
to add additional tabs to the user preferences section.


adunsulag:
@sjpadgett Here is my PR.  Let me know if anything here is going to break what you are doing.

ccr/display.php
```diff
@@ -1,68 +1,68 @@
...
-        $log_content .= date('Y-m-d H:i:s') . " ==> CouchDB Document Id: " . $couch_docid . "\r\n";
-        $log_content .= date('Y-m-d H:i:s') . " ==> CouchDB Revision Id: " . $couch_revid . "\r\n";
-        $log_content .= date('Y-m-d H:i:s') . " ==> Failed to fetch document content from CouchDB.\r\n";
-        //$log_content .= date('Y-m-d H:i:s')." ==> Will try to download file from HardDisk if exists.\r\n\r\n";
-        $this->document_upload_download_log($d->get_foreign_id(), $log_content);
-        die(xlt("File retrieval from CouchDB failed"));
+
+try {
+    $twig = new TwigContainer(null, $GLOBALS['kernel']);
+    if (!$d->can_access($_SESSION['user_id'])) {
```
bradymiller, line 30: check out this PR (brought into codebase about a week ago) where clarified how to use the can_access function:
https://github.com/openemr/openemr/pull/5235

(basically just don't include a parameter and then the acl inner functions will default to $_SESSION['authUser'] )

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -2192,62 +2197,88 @@ public function getCCDAComponents($type)
     * @param    integer     $status
     * @return   None
     */
-    public function logCCDA($pid, $encounter, $content, $time, $status, $user_id, $view = 0, $transfer = 0, $emr_transfer = 0)
+    public function logCCDA($pid, $encounter, $content, $time, $status, $user_id, $document_type, $view = 0, $transfer = 0, $emr_transfer = 0)
     {
         $content = base64_decode($content);
```
bradymiller, line 2202: if goal here is to use documents mechanism, then would not try to recreate it (it is complicated, considering couchdb especially, encryption, and uuid, and also then need to updated it when the core document mechanism is updated). Instead should just run it through the document function (createDocument()), which does it all via the centralized function.

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -2192,62 +2197,88 @@ public function getCCDAComponents($type)
...
+            throw new RuntimeException("document category id does not exist in system");
+        }
+
+        $binaryUuid = (new UuidRegistry(['table_name' => 'ccda']))->createUuid();
+        $file_name = UuidRegistry::uuidToString($binaryUuid);
+        $mimeType = "text/xml";
+
+        $higherLevelPath = "";
+        $pathDepth = 1;
+        $owner = $_SESSION['user_id'];  // userID of who is creating the document...
```
bradymiller, line 2225: this session only appears to be set in oauth stuff and is the user uuid

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
bradymiller: oh wait, you are using createDocument. I was thrown off by the creation of the uuid which looks like you are doing for the ccda table entry

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
bradymiller: to clarify, owner should be the id

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -2192,62 +2197,88 @@ public function getCCDAComponents($type)
...
+        $expirationDate = null;
+
+        try {
+            \sqlBeginTrans();
+
+            // set the foreign key so we can track documents connected to a specific export
+            $result = $document->createDocument(
+                $pid,
+                $categoryId,
+                $file_name,
```
bradymiller, line 2236: Just checking to make sure you want to use a uuid for the human readable name of the document?

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -2192,62 +2197,88 @@ public function getCCDAComponents($type)
...
+                $mimeType,
+                $content,
```
bradymiller, line 2238: To simplify code, can stop after $content with the parameters. All the rest will work via default ($owner will be set to $_SESSION['authUserID'] within the function).

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -2192,62 +2197,88 @@ public function getCCDAComponents($type)
...
+                $document->set_encounter_id($encounter);
+            }
+            $document->persist(); // save the updated references here.
+            \sqlCommitTrans();
+        } catch (\Exception $exception) {
+            \sqlRollbackTrans();
+            // TODO: @adunsulag do we need to clean up the file if we fail to commit the transaction here?
+            throw $exception;
+        }
+        return new GeneratedCcdaResult($moduleInsertId, UuidRegistry::uuidToString($binaryUuid), $content);
```
bradymiller, line 2281: looks like 2nd parameter could be $file_name

interface/super/edit_globals.php
```diff
@@ -730,6 +730,8 @@ function checkBackgroundServices()
                                                 }
 
                                                 echo "  </select>\n";
+                                            } else if ($fldtype == GlobalSetting::DATA_TYPE_MULTI_SORTED_LIST_SELECTOR) {
+                                                include 'templates/field_multi_sorted_list_selector.php';
```
bradymiller, line 734: 2DV

sql/6_1_0-to-7_0_0_upgrade.sql
```diff
@@ -656,3 +656,13 @@ DROP TABLE `temp_table_one`;
...
 #EndIf
+
+#IfNotRow2D categories name CCD codes LOINC:34133-9
+Update categories SET codes='LOINC:34133-9' WHERE name='CCD';
+#EndIf
+
+#IfNotRow2D list_options list_id lists option_id ccda-sections
+INSERT INTO list_options (list_id,option_id,title, seq, is_default, option_value) VALUES ('lists','ccda-sections','CCDA Sections',0, 1, 0);
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, codes) VALUES ('ccda-sections','allergies','Allergies, Adverse Reactions, Alerts',10,0,1, 'oid:2.16.840.1.113883.10.20.22.2.6.1');
+INSERT INTO list_options (list_id,option_id,title,seq,is_default,activity, codes) VALUES ('ccda-sections','medications','History of Medication Use',10,0,1, 'oid:2.16.840.1.113883.10.20.22.2.1.1');
```
bradymiller, line 667: make this seq 20

sql/6_1_0-to-7_0_0_upgrade.sql
```diff
@@ -656,3 +656,13 @@ DROP TABLE `temp_table_one`;
 #IfNotRow2D list_options list_id issue_subtypes option_id diagnosis
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`) VALUES ('issue_subtypes','diagnosis','Diagnosis',40);
```
bradymiller, line 657: this is just a draft, so figured you haven't added to database.sql yet

src/RestControllers/FHIR/FhirDocumentRestController.php
```diff
@@ -110,4 +117,15 @@ public function addMimeTypeHandler($mimeType, IDocumentDownloader $handler)
         }
         $this->mimeTypeHandlers[$mimeType] = $handler;
     }
+
+    private function findDocumentForDocumentId(string $documentId)
+    {
+        if (Uuid::isValid($documentId)) {
+            $document = \Document::getDocumentForUuid($documentId);
+        } else {
+            // use our integer values
```
bradymiller, line 126: if allowing integers, then prob need to remove the `string` from parameter

bradymiller:
@adunsulag , even in draft form, awesome code!
(even though was in draft form, did a code review to hopefully speed things up on the review side)

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
adunsulag: That is what ccda was doing before... I'm not sure what is the best to do here?  Maybe PatientName-CCDA-YYYY-MM-DD.xml?

bradymiller: If this is what was happening before, then would just leave it be :)

sql/6_1_0-to-7_0_0_upgrade.sql
adunsulag: I should now have all the changes in the database.sql file.

adunsulag:
@bradymiller I switched this from draft to a full review.  I believe I've addressed all your issues with your earlier review. There are some additional adjustments I'll need to make in a future PR to how the dates are handled but this code is actually working as the ccd is generated via the api.

I have a number of b1 stuff that will build off this PR.

bradymiller (approved):

bradymiller:
nice! bring on the munchkins!

munchkin-wizard-of-oz

PR #5341, opened 2022-05-15
couple optimizations for the node services

### LINKED ISSUE #5340: Carecoordination module not working
Related to nodejs service. I have a fix for this which will post as a PR (not actually breaking, just need to wait for the service to start up).

image

bradymiller:
fixes #5340

also removed node_binary global since not used in most service and is a security issue.

TODO (for me):
make it so that it only breaks when the service is needed (for example, Care Coordination module does not need it for import of ccda)
(I am going to work on this TODO in future after get more acquainted with the node services)


stephenwaite:
excited_elephant

PR #5363, opened 2022-05-19
Attempt to fix QdmRequestSome implementation of QDMRequestInterface

kchapple:
@sjpadgett I haven't tested this update, but this should get you close. I don't think this was used previously in the system anywhere, so never got a full workout.

### DIFF
--- a/src/Services/Qdm/QdmRequestSome.php
+++ b/src/Services/Qdm/QdmRequestSome.php
@@ -15,10 +15,6 @@ use OpenEMR\Services\Qdm\Interfaces\QdmRequestInterface;
 
 class QdmRequestSome implements QdmRequestInterface
 {
-    protected $pids = [];
-
-    protected $pidString = "";
-
     protected $filter = null;
 
     /**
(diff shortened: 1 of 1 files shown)


src/Services/Qdm/QdmRequestSome.php
```diff
@@ -28,39 +24,18 @@ class QdmRequestSome implements QdmRequestInterface
      */
     public function __construct(array $pids)
     {
-        $this->pids = $pids;
-
-        if (is_array($pids)) {
-            $this->pidString = implode(",", $pids);
-        }
-
+        $filterClause = str_replace("?,", count($pids));
```
sjpadgett, line 27: @kchapple Doesn't make sense. Missing subject in str_replace.

kchapple: sorry @sjpadgett is supposed to be str_repeat(), auto-complete beat me this time.

sjpadgett: That worked so thank you sir. I'll bring in once you update PR.

sjpadgett (approved):
Sorry just noticed you already update. Thanks again and merging once test complete.

PR #5378, opened 2022-05-25
Openemr fix 5377

### LINKED ISSUE #5377: CQM Measure 147 fails with fatal error on cat1 imports
Measure 147 is currently failing on a cat1 import when dealing with a negated immunization section.  It throws an error because we are not providing a negation reason when there is a SNOMED-CT code in the c1 file for a patient refusal.  The cypress parser looks for the negation timestamps and dies on import. 

It looks like there is a problem between the qrda import files that use "SNOMED CT" as its codesystem type and the codes we use in our systems for SNOMED-CT when dealing with the immunization negation reason code.

adunsulag:
Fixes #5377 updates the reason code lookup for immunization to pass Measure 147 for Clinical Quality Measures.

Also added some debugging functions.  

Fixed a negation rationale error that was being thrown.

### DIFF
--- a/src/Services/Cda/CdaTemplateImportDispose.php
+++ b/src/Services/Cda/CdaTemplateImportDispose.php
@@ -797,8 +797,11 @@ class CdaTemplateImportDispose
             }
             $option = '';
             if (!empty($value['reason_code'])) {
+                // TODO: check with @sjpadgett why the codes are "SNOMED CT:<number>" instead of "SNOMED-CT" for the immunization reason code
+                $reason_code = str_replace("SNOMED CT", "SNOMED-CT", $value['reason_code']);
+
                 $listService = new ListService();
-                $option = $listService->getOptionsByListName('immunization_refusal_reason', ['codes' => $value['reason_code']])[0];
+                $option = $listService->getOptionsByListName('immunization_refusal_reason', ['codes' => $reason_code])[0];
             }
 
             if (!empty($value['extension'])) {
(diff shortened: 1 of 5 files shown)


src/Services/Cda/CdaTemplateImportDispose.php
```diff
@@ -797,8 +797,11 @@ public function InsertImmunization($imm_array, $pid, CarecoordinationTable $care
             }
             $option = '';
             if (!empty($value['reason_code'])) {
+                // TODO: check with @sjpadgett why the codes are "SNOMED CT:<number>" instead of "SNOMED-CT" for the immunization reason code
+                $reason_code = str_replace("SNOMED CT", "SNOMED-CT", $value['reason_code']);
```
sjpadgett, line 801: Because SNOMED CT is the correct name for the code type. OpenEMR add the dash for the proper reason of search/parse etc. Still Ken handles this in the services and unclear why the current version works for me and not you.
Even still this is an appropriate fix so, yeah!

src/Services/Qdm/CqmCalculator.php
```diff
@@ -73,7 +73,7 @@ public function calculateMeasure($patients, Measure $measure, $effectiveDate, $e
             foreach ($patient->dataElements as $dataElement) {
                 // We need to look up OIDs and add the first code concept from the value set.
                 // We do this first in case it's in a dataElement that we need to clone for the calculator below
-                if ($dataElement->negationRationale !== null) {
+                if (isset($dataElement->negationRationale) && $dataElement->negationRationale !== null) {
```
sjpadgett, line 76: what's wrong with !empty($dataElement->negationRationale)?

adunsulag: Nothing, that'd work too.  Just did the first thing that jumped to my mind :)

bradymiller (approved):

src/Services/Qdm/CqmCalculator.php
sjpadgett: Oh well, I just jump at any opportunity to nag!:)

PR #5397, opened 2022-05-30
Improve Patienf Finder UI

robertdown:


interface/main/finder/dynamic_finder.php
```diff
@@ -286,7 +286,7 @@
         if (!empty($GLOBALS['gbl_pt_list_new_window'])) {
             echo ' checked';
         }
-        ?> /><?php echo xlt('Open in New Window'); ?></label><label for='setting_search_type' id='setting_search_type_label'><input type='checkbox' name='setting_search_type'  id='setting_search_type' onchange='persistCriteria(this, event)' value='<?php echo attr($patient_finder_exact_search); ?>'<?php echo text($patient_finder_exact_search); ?>/><?php echo xlt('Search with exact method'); ?></label></form>");
+        ?> /><?php echo xlt('Open in New Window'); ?></label>&nbsp;<label for='setting_search_type' id='setting_search_type_label'><input type='checkbox' name='setting_search_type'  id='setting_search_type' onchange='persistCriteria(this, event)' value='<?php echo attr($patient_finder_exact_search); ?>'<?php echo text($patient_finder_exact_search); ?>/><?php echo xlt('Search with exact method'); ?></label></form>");
```
sjpadgett, line 289: can we just use margin classes instead of `&nbsp`. 
`
$("div.mytopdiv").html("<form name='myform'><label for='form_new_window' id='form_new_window_label'><input class='mx-1' type='checkbox' id='form_new_window' name='form_new_window' value='1'<?php
        if (!empty($GLOBALS['gbl_pt_list_new_window'])) {
            echo ' checked';
        }
        ?> /><?php echo xlt('Open in New Window'); ?></label><label for='setting_search_type' id='setting_search_type_label'><input type='checkbox' name='setting_search_type' class='mx-1'  id='setting_search_type' onchange='persistCriteria(this, event)' value='<?php echo attr($patient_finder_exact_search); ?>'<?php echo text($patient_finder_exact_search); ?>/><?php echo xlt('Search with exact method'); ?></label></form>");
`

interface/main/finder/dynamic_finder.php
```diff
@@ -374,11 +374,11 @@ function persistCriteria(el, e) {
             <?php if (AclMain::aclCheckCore('patients', 'demo', '', array('write','addonly'))) { ?>
                 <button id="create_patient_btn1" class="btn btn-primary btn-add" onclick="top.restoreSession();top.RTop.location = '<?php echo $web_root ?>/interface/new/new.php'"><?php echo xlt('Add New Patient'); ?></button>
             <?php } ?>
-            <div class="jumbotron mt-3 p-4">
+            <div>
                 <div id="dynamic"><!-- TBD: id seems unused, is this div required? -->
                     <!-- Class "display" is defined in demo_table.css -->
-                    <div class="table-responsive">
```
sjpadgett, line 380: Need to keep table-responsive class so margins and padding are maintained on small screens. Otherwise will scroll on x-axis most times.

interface/main/finder/dynamic_finder.php
robertdown: Resolved with `.form-check-inline` which is a bootstrap utility class

interface/main/finder/dynamic_finder.php
robertdown: Added back

sjpadgett: thanks Robert

PR #5401, opened 2022-05-30
Fixes #5400 cat 1 memory usage

### LINKED ISSUE #5400: QRDA Cat 1 export uses 11MB of memory for each patient causing low memory php servers to crash.
Had an issue where the qrda cat 1 export was crashing on my linux docker image with 256MB for the php application.  Using phpstorm and xdebug I profiled the memory usage using memory_get_usage() and saw everytime we did a json_decode on our patient for the cat1 export that the memory usage would climb by 11MB.  For whatever reason despite several objects, function calls etc being popped off the stacks (which should have triggered php's garbage collection), memory usage kept climbing.

We need to manually trigger the garbage collector after each patient is processed in the cat 1 zip export so we can minimize our memory usage.  I was only able to get the cat 1 reports to work on my local machine and the demo farms by bumping the memory usage up to 512MB for php to run.  If we can clear out the memory then the zip exports will still work on our default configuration of 256MB.

adunsulag:
Manually forces a garbage collection cycle after each cat1 patient is
exported.  We trade cpu cycles for memory usage so that we can run on
low memory environments (such as our default dockers).

### DIFF
--- a/src/Cqm/QrdaControllers/QrdaReportController.php
+++ b/src/Cqm/QrdaControllers/QrdaReportController.php
@@ -143,7 +143,13 @@ class QrdaReportController
                     }
                     file_put_contents($file, $content);
                     file_put_contents($file_local, $content);
+                    // in order to deal with our zip files we are going to force our garbage collector to run
+                    // json_decode stores a LOT of memory with our measures so we need to collect it.  Otherwise
+                    // we end up exceeding our memory usage. ~60 patients exceeds over 256MB of memory...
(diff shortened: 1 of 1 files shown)


adunsulag:
@sjpadgett This was a fun and weird bug!  If it looks good to you, let's bring it in lol.

bradymiller (approved):

sjpadgett (approved):
Seems like I brought this up awhile back.

PR #5402, opened 2022-05-30
ci and docker updates to begin to support alpine 3.16 and mariadb 10.8

bradymiller:
ci and docker updates to begin to support alpine 3.16 and mariadb 10.8


bradymiller:
btw, sticking with alpine 3.15 for the standard flex (easy dev docker environment) and for the planned 7.0.0 release since don't want to break anything for the release. After the 7.0.0 release, will transition standard flex and future 7.0.1 to use alpine 3.16 (php 8.1).

PR #5406, opened 2022-06-01
Fix multi code type with code.

sjpadgett:
- procedure import was adding additional code type to code on code validation.


sjpadgett:
up to you guys if you want this in our test branch.
What was happening is a combined code type and code(CPT4:72107) is sent to import  from parse where sometimes they may be just a type and code to be combined so call to method that combines type and code was adding an additional type to code(CPT4CPT4:72107) causing mayhem.:)
Fortunately the back end qrda services checked for this and fixed. Only viewing in UI showed error.

sjpadgett:
You're welcome. So merging with master.

PR #5417, opened 2022-06-04
Fixes #5416 CCDA Zip Import

### LINKED ISSUE #5416: Carecoordination Import Support zip file imports
We should support importing the zip files that OpenEMR generates for our ccd/toc/referral Carecoordination exports. Right now the zip exports are in a format of TimestampDirectory/PatientName/PatientName_ccd.xml.  We need to be able to grab all of the xmls that have been exported in a zip file and be able to import the entire zip into OpenEMR.

Right now OpenEMR can't actually use any of the zip files we export w/o the user manually extracting their OpenEMR exports and grabbing all of the xml files by hand to upload to another OpenEMR installation.

Careful consideration needs to be paid to the security of working with zip files.  The security issues I'm aware of are the following:
1. We have to deal with the zip file archive comments to mitigate any buffer overflow attacks in the archive.
2. We need to avoid file overflow attacks where an attacker attempts to flood the hard drive with millions/billions of uncompressed 0-1 byte files.  
3. We need to avoid recursive decent attacks where attackers have folder structures that are a thousands/millions deep
4. We need to make sure uncompressed files are jail rooted to their target folder and things such as "../etc/passwd" are not allowed when extracting to a tmp folder.

Mitigation of this threats would be done as follows:
1. Threat 1 - Set the zip file archive comment to an empty string
2. Threat 2 - Limit the upload to a maximum number of patients (make this configurable) a good default would be 500 patients.  Since patient names are the second folder name we can track the number of unique patients this way.  
3. Threat 2 -  Limit the number of parsed ccd files that a single patient can have (make this configuration) a good default would be 5-10 records.
4. Threat 3 - use explode on the '/' delimiter and set a limit to 3 so we don't parse any farther on the zip file.
5. Threat 4 - Remove any entries that do not explicitly match the OpenEMR zip file format of TimestampDirectory/PatientName/ccd.xml which will remove entries attempting to escape the jail root.

At some point we would like to bundle the patient documents as part of a export / transfer of care as an option so this mechanism should be extensible so that use case can be supported.

One possible mechanism is to avoid extracting files directly to the disk at all and just grab the specific ccds we are wanting to import from the zip file.   This would mitigate a lot of dangers that can occur with zip files.  We know the ccds are xml format so we can grab mime type information directly w/o having to extract the contents.

The downside of not extracting the zip file is if we ever want to use the mechanism to migrate patient documents alongside their ccd we wouldn't be able to get mime type information and pass it through our upload filters (which restricts the kinds of mime types allowed in the system).

adunsulag:
We make it so we can import a zip file of CCDA files in the format that we export them via OpenEMR.

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/CarecoordinationController.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/CarecoordinationController.php
@@ -26,6 +26,7 @@ use Carecoordination\Model\CarecoordinationTable;
 use C_Document;
 use Document;
 use CouchDB;
+use OpenEMR\Common\Logging\SystemLogger;
 use OpenEMR\Services\Cda\CdaValidateDocuments;
 use xmltoarray_parser_htmlfix;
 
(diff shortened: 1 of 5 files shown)


interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/CarecoordinationController.php
```diff
@@ -909,4 +914,152 @@ private function getRevAndApproveAuditArray($audit_master_id, $table_name)
...
+
+            // now we need to do our document import for our ccda for this patient
+            if ($componentCount == 3) {
+                // let's process the ccda
+                $file_name = basename($stat['name']);
+
+                $pid = '00';
+                $ob = new Document();
+                $contents = $z->getFromIndex($i);
+                $ret = $ob->createDocument($pid, $catId, $file_name, 'text/xml', $contents, '', 1, 0);
```
bradymiller, line 1053: would drop the last 3 parameters of this function call since are just the default settings

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncountermanagerController.php
```diff
@@ -318,6 +325,10 @@ public function downloadallAction()
             header("Content-Transfer-Encoding: binary");
             readfile($zip_dir . $zip_name);
 
+            // we need to unlink both the directory and the zip file once are done... as its a security hazard
+            // to have these files just hanging around in a tmp folder
+            unlink($zip_dir . $zip_name);
```
bradymiller, line 330: 2DV

bradymiller:
@adunsulag , looks good! just noted minor issue and there's also a psr12 styling issue.

bradymiller (approved):

PR #5437, opened 2022-06-08
Fixes #5434 - Check New Messages button

### LINKED ISSUE #5434: Direct Mail Messages - Check New Messages button
Right now there can be a 2-3 minute delay between when a direct message is sent and it shows up in the Messages screen inside OpenEMR.

When we know that a message has been sent we want to be able to tell the system to pull in new messages right way.  We should have a button on the Messages screen for check new messages.  This button should only appear if the direct messaging feature is turned on.

adunsulag:
Added a check new messages button for direct messaging that triggers the background service to run immediately.

### DIFF
--- a/interface/main/messages/messages.php
+++ b/interface/main/messages/messages.php
@@ -884,7 +884,6 @@ if (!empty($_REQUEST['go'])) { ?>
             $('.trusted-messages-force-check').click(function() {
                 window.top.restoreSession();
                 request = new FormData;
-                request.append("skip_timeout_reset", "1");
                 request.append("ajax", "1");
                 request.append("csrf_token_form", <?php echo js_escape(CsrfUtils::collectCsrfToken()); ?>);
                 request.append("background_service", "phimail");
@@ -904,7 +903,7 @@ if (!empty($_REQUEST['go'])) { ?>
                     }
                 }).catch(function(error) {
                     console.log('Background Service refresh failed: ', error);
-                    alert(window.xl("Check new messages failed.  Check the server logs for more information."));
+                    alert(window.xl("Check new messages failed. Check the server logs for more information."));
                 });
             });
             <?php endif; ?>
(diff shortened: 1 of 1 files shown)


interface/main/messages/messages.php
```diff
@@ -878,6 +880,35 @@ function deselectRow(row) {
                 ,minDate : 0 //only future
             })
 
+            <?php if ($GLOBALS['phimail_enable']) : ?>
+            $('.trusted-messages-force-check').click(function() {
+                window.top.restoreSession();
+                request = new FormData;
+                request.append("skip_timeout_reset", "1");
```
bradymiller, line 887: if this function is being run manually, then can remove above line. If this function is done automatically on a timer, then would leave in.
(this basically bypasses the timeout mechanism (logout user that is not active for a certain amount of time), so that automated scripts do not impact the timeout (where the timeout would never happen))

adunsulag: ok, no timeout, I just grabbed the sample from what we were doing in the main window frame for running the background service.

interface/main/messages/messages.php
```diff
@@ -878,6 +880,35 @@ function deselectRow(row) {
...
+                request.append("ajax", "1");
+                request.append("csrf_token_form", <?php echo js_escape(CsrfUtils::collectCsrfToken()); ?>);
+                request.append("background_service", "phimail");
+                request.append("background_force", "1");
```
bradymiller, line 891: super cool. forgot about the forced call of background service :)

bradymiller:
@adunsulag , looks great! just noted couple minor issues. otherwise ready to bring in.

2DV-2

interface/main/messages/messages.php
adunsulag: Fixed.

interface/main/messages/messages.php
```diff
@@ -878,6 +880,35 @@ function deselectRow(row) {
...
+                        console.log('Background Service refresh failed. Status Code: ' + response.status);
+                    } else {
+                        // we've refreshed give them time to reload the page
+                        setTimeout(function() {
+                            window.location.reload();
+                        }, 500);
+                    }
+                }).catch(function(error) {
+                    console.log('Background Service refresh failed: ', error);
+                    alert(window.xl("Check new messages failed.  Check the server logs for more information."));
```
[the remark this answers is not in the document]
adunsulag: Fixed.

PR #5438, opened 2022-06-08
Fixes #5436 Messages datetime

### LINKED ISSUE #5436: Messages Screen - Show Messages with time component
We need to modify the messages screen to show both the date and the time that a message is received.  If a lot of messages are received in the interface on the same day, it is difficult to figure out the newer messages.

adunsulag:
Made it so the date and time of the message received is shown.


bradymiller (approved):

stephenwaite:
bradys-panda

PR #5448, opened 2022-06-09
New external MDHT validation

### LINKED ISSUE #5447: Need different CCDA validation for ONC negative testing
### Describe the problem
Though recommended by C-CDA I.G, our current schematron will not catch many potential conformance errors.
Solution is to implement the same validator that is used by the ONC testing environment.
The validator is a Java project which can be ran in a local Tomcat instance or via a rest API call to ONC public server.

Will set up globals to enable and address location.

sjpadgett:
- add functionally to validation service.
- fix referral from direct to be empty
- new globals for validation server

Fixes #5447

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -3509,6 +3509,20 @@ $GLOBALS_METADATA = array(
             '',
             xl('USPS Web Tools API Username')
         ),
+
+        'mdht_conformance_server_enable' => array(
+            xl('Use MDHT External Validation Service'),
+            'bool',                           // data type
+            '0',
+            xl('Enable CCDA conformance and validation API service')
+        ),
+
+        'mdht_conformance_server' => array(
+            xl('CCDA MDHT Validation API Server Address'),
+            'text',                           // data type
+            'http://ccda.healthit.gov',
+            xl('CCDA conformance and validation API service URL. Default is a public server mainly used for testing. Production sites should spin up your own.')
+        ),
     ),
 
     'Rx' => array(
--- a/src/Services/Cda/CdaValidateDocuments.php
+++ b/src/Services/Cda/CdaValidateDocuments.php
@@ -12,14 +12,29 @@
 
 namespace OpenEMR\Services\Cda;
 
+use CURLFile;
+use CURLStringFile;
 use DOMDocument;
 use Exception;
 use OpenEMR\Common\System\System;
+use stdClass;
 
 class CdaValidateDocuments
 {
+    public $externalValidatorUrl;
+    public $externalValidatorEnabled;
+
     public function __construct()
     {
+        $this->externalValidatorEnabled = !empty($GLOBALS['mdht_conformance_server_enable'] ?? false);
+        $this->externalValidatorUrl = null;
+        if ($this->externalValidatorEnabled) {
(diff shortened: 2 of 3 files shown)


sjpadgett:
image

sjpadgett:
I plan to merge as quickly as possible so @adunsulag can have for his development.

library/globals.inc.php
```diff
@@ -3509,6 +3509,20 @@ function gblTimeZones()
...
+            xl('Use MDHT External Validation Service'),
+            'bool',                           // data type
+            '0',
+            xl('Enable CCDA conformance and validation API service')
+        ),
+
+        'mdht_conformance_server' => array(
+            xl('CCDA MDHT Validation API Server Address'),
+            'text',                           // data type
+            'http://ccda.healthit.gov',
```
adunsulag, line 3523: can we assume the file will always have the /referenceccdaservice/ path?  Or would it be better to specify the full server address here?  That leaves the option for people to customize / change this if they never need it to.

src/Services/Cda/CdaValidateDocuments.php
```diff
@@ -12,14 +12,29 @@
...
 {
+    public $externalValidatorUrl;
+    public $externalValidatorEnabled;
+
     public function __construct()
     {
+        $this->externalValidatorEnabled = !empty($GLOBALS['mdht_conformance_server_enable'] ?? false);
+        $this->externalValidatorUrl = null;
+        if ($this->externalValidatorEnabled) {
+            $this->externalValidatorUrl = trim($GLOBALS['mdht_conformance_server'] ?? null) ?: 'http://ccda.healthit.gov';
```
adunsulag, line 32: Does trim work on a null?

src/Services/Cda/CdaValidateDocuments.php
```diff
@@ -30,14 +45,54 @@ public function __construct()
...
+     * @param $xml
+     * @return array|mixed
+     */
+    public function ettValidateCcda($xml)
+    {
+        try {
+            $result = $this->ettValidateDocumentRequest($xml);
+        } catch (Exception $e) {
+            $e = $e->getMessage();
+            error_log($e);
```
adunsulag, line 71: I would change this to (new SystemLogger())->errorLogCaller($e->getMessage(), ["trace" => $e->getTraceAsString()]);  This will escape any user data that could have been passed back from the exception since you are doing a web request here.

adunsulag (commented):
Hey @sjpadgett Just noted a couple things, one minor security fix with the error_log.  Other than that it looks great!

src/Services/Cda/CdaValidateDocuments.php
sjpadgett: yes

library/globals.inc.php
sjpadgett: all arguments are specific to this server.

adunsulag:
Looks good to me, feel free to bring it when you're ready @sjpadgett.

sjpadgett:
thanks for the review Stephen.

sjpadgett:
Hitting the public server is kind of slow so you may want to try against my server Stephen if it seems tedious to you. I'll send the web address in email.

bradymiller:
2DV

PR #5451, opened 2022-06-10
quick fix prior commit

sjpadgett:
- use CURLStringFile is only valid for PHP >= 8.1
- remove unused use

Nuts is about all I have to say about this!:)


adunsulag:
Well that's awful.  Sorry for the bad advice!

sjpadgett:
not on you. CURLFile is okay. CURLStringFile is new where I was using it but forgot to remove the use when I switched.
so, my boo boo! I'll store it away though and blame you for something else when it comes up.

PR #5452, opened 2022-06-10
Openemr fix 5450 ccda viewer display options

### LINKED ISSUE #5449: ccda viewer missing content
Our implementation of the ccda viewer is missing the softwareDevice information that ONC checks for in their html output comparisons.  We are also missing timezone information. We are also missing the OID information for the patient ids.

### LINKED ISSUE #5450: ccda viewer section display options
We need to be able to view a single section of ccda at a time.  Ideally we could toggle all the sections to be turned off, toggle a single section at a time to be on, and turn on and off sections.

This needs to work with our initial section preferences that are available in globals and user preferences.  IE, the section limiting and section re-ordering.

adunsulag:
Fixes #5449 
Fixes #5450

Removed the duplicate cda-web.xsl file that was just a copy of the
cda.xsl file.

Added better support for assignedAuthor.softwareName
Added back the timezone component display
Added the oid display for patient ids
Disabled section sorting, Adding a single section display component

Setting the stage for a ccda toggle button on the top left.

[Integrated single/multi section hide/display toggles](https://github.com/openemr/openemr/commit/96731eda3fd19efdaa020a22614a6dd03d818a75) 

Still need to handle the overall clinical sections toggle

Screenshots:

Toggle Navigation options on ccda view
image

Toggle multiple sections hidden
image

Single Section View
image


sjpadgett:
Looks great Stephen.
The way ONC is going it won't be long until we have to have smiley faces on a good visit!:)

bradymiller:
tenor99

adunsulag:
Added a section if everything is hidden to indicate no clinical information is present.
image

Got the sorting working in this latest push so sections can now be dragged and dropped if someone wants to change up the display order at the time of print.  They can re-arrange sections different than what their global preferences are.  I thought it'd be nice if we could just save the preferences right there, but we use the same CDA.xsl for generating our offline version so I didn't worry about it.

One unfortunate caveat with using the Lantana version is that it doesn't have any of the language translations that the HL7 V4 CDA has.  We may want to look at doing translations for a future improvement.  One thought would be to make the CDA.xsl a php file that would then generate the CDA.xsl file.  That would enable us to get translations based on the user's preferences.  It'd be quite a bit of work though so I'm not going to tackle it before our ONC certification.

adunsulag:
@bradymiller this is now ready for review.

bradymiller (approved):

bradymiller:
@adunsulag , looks great!

PR #5469, opened 2022-06-12
flow board report php8 fixes

### LINKED ISSUE #5468: flow board report php8 errors
flow board report php8 errors

bradymiller:
fixes #5468 

@psoas , this will fix the errors you noted on the flow board report during today's call


adunsulag:
That's a really odd bug.  Since line 26 has the facility service already defined, it should have already populated it in globals.

bradymiller:
yep, agreed. remember seeing something like in the distant past when having methods call other methods in the gacl stuff that relied on a globals.

PR #5481, opened 2022-06-13
Fixes #5479 Change reason code empty constant name

### LINKED ISSUE #5479: Rename ReasonStatusCodes::EMPTY to be NONE due to reserved PHP keyword
Updated php code sniffer does not like using EMPTY as a constant value.  We need to rename this to be something like NONE to signify an empty string value, IE there is no reason code status.

adunsulag:
Fixes #5479 change the reason code from EMPTY to NONE to fix the php keyword problem.


bradymiller:

PR #5490, opened 2022-06-14
Fixes #5455 Document Type Display Order

### LINKED ISSUE #5455: Change carecoordination display order preferences to be grouped by document type.
We built the care coordination global / user preferences to be able to sort and set the display order of each section. 

However, reading through the SLI testing script they state that we must be able to set the display order preferences for EACH document type.  The carecoordination document types being tested for ONC certification are a CCD document and a referral document.

It makes sense as someone may want to see a different order for sections for a referral document than they would for a ccd.  It does mean the current way we implemented this isn't going to work.

adunsulag:
Fixes #5455 
After talking this over with @sjpadgett we decided to implement this by having multiple sort order sections for the four document types we support in CCD-A.  Its not ideal, but the alternative will take quite a bit of time to architect.  

I wrote a new preview option for CCD-A that examines the XML document to figure out the correct CCDA Template document type so we can know what sort order to use to display for the correct document.

Added a deprecation notice to the ccr/display.php in case anyone is using it.  We no longer reference it from the documents pages so we can probably remove it at some point.

I also added a new filter event to the patient document tree viewer so people can modify / update the html tree rendering on the documents if they so choose.  This made it easy to loosely couple the tree with the Carecoordination module.

One thing I did was leave in place where we give the CCDA preview only to documents that are in the CCR/CCD/CCDA category folders.  Another option would be to key off the LOINC codes of any category.  That way people could have whatever categories they want... but I don't know if people would really use that and if it just makes more sense to go with the defaults.  The downside is if anyone deletes those categories then the CCDA previewers will stop working.

Another option would be just to have the preview on pretty much any XML document but I'm not sure we want to do that either.  If you have any thoughts here, let me know.

### DIFF
--- a/ccr/display.php
+++ b/ccr/display.php
@@ -8,6 +8,7 @@
  * @author    Ajil P.M <ajilpm@zhservices.com>
  * @author    Brady Miller <brady.g.miller@gmail.com>
  * @author    Stephen Nielson <snielson@discoverandchange.com>
+ * @deprecated 7.0.0 People should use the /interface/modules/zend_modules/public/encountermanager/previewDocument?docId=<id> REST action instead of this file
  * @copyright Copyright (c) 2011 Z&H Consultancy Services Private Limited <ajilpm@zhservices.com>
  * @copyright Copyright (c) 2013 Brady Miller <brady.g.miller@gmail.com>
  * @copyright Copyright (c) 2022 Discover and Change <snielson@discoverandchange.com>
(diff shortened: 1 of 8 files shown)


adunsulag:
Screenshot of the four sortable document types
image

Referral note sorted sections
image

CCD sorted sections
image

Updated document window launcher
image

sjpadgett:
very nice work @adunsulag thanks.

PR #5494, opened 2022-06-15
Fixes #5473 Metric Precision fixes

### LINKED ISSUE #5473: Vitals Form with metric display setting does not show correct values from saved input due to imprecise precision saving
On the vitals form if you have your Global settings of Locale->Units for Visit Forms set to **Show metric only**, and you put in a vitals observation value that has an Imperial (United States customary units) conversion with more than two digits of precision, it will not display as the same value when you go back to edit the form.

This occurs because we store all OpenEMR vital observation values as imperial units in the database even though we display them as metric.  For example if you put in a value of 177 cm for the height observation, save the value and come back and edit the vitals form again, you will see the height value is now 177.01 cm.  In the database the value is saved as 69.69 inches when it should be saved as 69.685 inches for a correct conversion.  A celsius temperature of 38, when edited again shows, 37.99.  

We either need to store more precision values in the database or store the actual metric value in the database to avoid these kinds of errors.

adunsulag:
Fixes #5473 
This was a doozy of a fix.  Went through several database restores and upgrades to make sure this wasn't destroying data.  It does need to be run in off hours as there's potential for data loss due to the dropping of columns.  Straight up converting the columns did not work for what I was testing.  Though perhaps @bradymiller has a better solution on handling the data conversions. 

Increased the decimal precision from 2 digit to 6 digits in the database
for form_vitals.  Our metric conversion was not giving back the same
values to the user as what they were entering in.  Much of this was due
to the conversion precision of just 2 digits.

Also MySQL is deprecating the float(X,N) data type and so we changed to
the decimal format to make this work.  Note that upgrades will need to
be done in off hours here as we are dropping the database columns to
handle the conversion.

Changed up the vitals conversion javascript so we could have a hidden
input that contains the 6 digit precision for metric / us conversion
purposes but only shows the two digit value to the user to keep the
interface clean and less confusing.


bradymiller (approved):

bradymiller:
@adunsulag , beautiful stuff!

tenor (67)

interface/forms/vitals/vitals.js
```diff
@@ -227,17 +264,35 @@ function convCtoF(name) {
...
+    let bmiNode = document.getElementById("BMI_input");
+    if (!bmiNode) {
+        console.error("Failed to find node with id BMI_input");
+        return;
+    }
+
+    let precision = vitalsGetPrecision(bmiNode, 2);
+
+    let heightNode = document.getElementById("height_input");
+    if (!bmiNode) {
```
stephenwaite, line 276: thinking should be
if (!heightNode) {

interface/forms/vitals/vitals.js
```diff
@@ -227,17 +264,35 @@ function convCtoF(name) {
...
+        console.error("Failed to find node with id height_input");
+        return;
+    }
+    let weightNode = document.getElementById("weight_input");
+    if (!bmiNode) {
```
stephenwaite, line 281: and like above
if (!bmiNode) {

interface/forms/vitals/vitals.js
```diff
@@ -39,6 +39,51 @@
...
+            return;
+        }
+        let inputConv = document.getElementById(targetInputConv);
+        if (!inputConv) {
+            console.error("Failed to find node with data-target-input-conv of ", targetInputConv);
+            return;
+        }
+
+        if (value != "") {
+            let convValue = convUnit(system, unit, value);
```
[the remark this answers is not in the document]
adunsulag: This is fixed by removing the kg.toFixed() inside the convUnit function.  I thought I had removed them all, but apparently I hadn't.  Everything now comes back as a number from the function.

interface/forms/vitals/vitals.js
adunsulag: Good catch!  Thanks!

interface/forms/vitals/vitals.js
adunsulag: Good catch!  Thanks!

PR #5518, opened 2022-06-20
mariadb 10.2 is end of life, so dropping support and ci testing

bradymiller:
mariadb 10.2 is end of life, so dropping support and ci testing


adunsulag:
Hooray for faster builds!

PR #5525, opened 2022-06-22
Openemr fix 5508 ccd date ranges

### LINKED ISSUE #5508: G9 api connect start and end date for ccd generation
Currently g9 only is generating the most current ccd document for a patient.  

We need to make sure we can generate a document on a specific date (start and end date are the same day) and that we can generate a CCD document with a date range.  For ONC we only have to meet those two requirements.

If we want to be IG spec compliant with HL7 FHIR g9 we need to be able to handle the datetime modifiers (eq, lt,gt,eb,ap, etc) with portions of dates (only year, year & month, year & day) all the way down to second granularity.  This would require passing down our Date search field objects into the CCDA generator and using their more powerful search expressiveness to generate our date filtering.  If I end up doing this, we would need to support a string date like we have now and the option of the date search field object which should accomodate what we need from both the UX and the API endpoints.

For background reading if anyone cares here is the the $docref specification for HL7 and here is the search modifier documentation:
http://hl7.org/fhir/us/core/OperationDefinition-docref.html

https://build.fhir.org/search.html#date

### LINKED ISSUE #5517: Additional Addresses - Dates Not Being Saved
When you enter start date and end date, it is not being saved in the database.

You can see this if you:
Click on "Save" after entering the address information.
Then, click on the edit icon next to demographics.

Screen Shot 2022-06-20 at 1 05 56 AM

Also, if you look at the mysql columns where dates were entered, the date columns are:  0000-00-00 00:00:00.

If you do not enter a date, the date columns are:  NULL.

adunsulag:
Fixes #5508 
Fixes #5517

Changes g9 $docref to be a POST api since we generate a new document
every time.

Fixes ISO8601 date problems with php since php's date constant is
incorrect.  Switch to ATOM constant.

Fixed the ORDataObject saving incorrectly on some database instance
types due to the timezone and microsecond formatting.

Added __clone helper methods to the search field classes so we can copy
them and modify properties w/o modifying the references.

Added brief, terse documentation for g9 generation with docref.  Needs
to be fleshed out in the fhir readme file.

Updated the swagger file to point to the g9 documentation.

Removed the remote xsl stuff as I found out chrome actively blocks
remote xsl's from rendering unless everything is in the same folder,
port, origin, and schema.

Changed up the filtering as careplan was still breaking due to its
requirement on an encounter.  Changed up the filtering to be encounter
based so that any of the sections related to an encounter are filtered
for the CCD.  I left non-ccd document types alone since g9 only requires
the ccd filtering.

Removed unused referral pieces in the care plan section as they were all
being skipped due to the missing required codes in the transactions
table.  node side was skipping it as well so I left it off.

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
@@ -55,6 +55,7 @@ class EncounterccdadispatchTable extends AbstractTableGateway
     public $searchFromDate;
     public $searchToDate;
     public $searchFiltered = false;
+    private $encounterFilterList = [];
 
     public function __construct()
     {
@@ -64,12 +65,13 @@ class EncounterccdadispatchTable extends AbstractTableGateway
      * @param $options
      * @return void
      */
-    public function setOptions($options)
+    public function setOptions($pid, $encounter, $options)
     {
         // we keep from and to dates in order to handle the transaction table where we have to manually convert the dates
         // since they are stored as strings.
         $this->searchFromDate = null;
         $this->searchToDate = null;
+        $this->encounterFilterList = [];
 
         $dateValues = [];
 
@@ -104,7 +106,11 @@ class EncounterccdadispatchTable extends AbstractTableGateway
         if (!empty($dateValues)) {
             $this->searchDateField = new DateSearchField('search_date', $dateValues, DateSearchField::DATE_TYPE_DATETIME, true);
             $this->searchFiltered = !empty($options['filter_content'] ?? false);
+            $this->encounterFilterList = $this->getEncounterListForDateRange($pid, $encounter);
         } else {
+            if (!empty($encounter)) {
+                $this->encounterFilterList = [intval($encounter)];
+            }
             $this->searchFiltered = false;
         }
 
@@ -665,10 +671,9 @@ class EncounterccdadispatchTable extends AbstractTableGateway
     */
     /**
      * @param $pid
-     * @param $encounter
      * @return string
      */
-    public function getAllergies($pid, $encounter)
+    public function getAllergies($pid)
     {
         $allergies = '';
         $query = "SELECT l.id, l.title, l.begdate, l.enddate, lo.title AS observation,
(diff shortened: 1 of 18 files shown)


adunsulag:
@sjpadgett and @bradymiller when you get a chance if you can look this over and let me know what you think?  I'll run through more test scenarios tomorrow but I wanted to get this out there for the two of you to look at it.

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -18,6 +18,7 @@
 use Application\Listener\Listener;
 use Application\Model\ApplicationTable;
 use Carecoordination\Model\CarecoordinationTable;
+use Couchbase\SearchQuery;
```
adunsulag, line 21: Just realized this was here, not sure how it got in here.  I'll remove it tomorrow.

bradymiller (approved):

bradymiller:
@adunsulag , Looks great! I will have time over next couple days to expand on the documentation.

2DV

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -55,11 +65,66 @@ public function __construct()
...
+        // we keep from and to dates in order to handle the transaction table where we have to manually convert the dates
+        // since they are stored as strings.
+        $this->searchFromDate = null;
+        $this->searchToDate = null;
+        $this->encounterFilterList = [];
+
+        $dateValues = [];
+
+        if (!empty($options['date_start'])) {
+            $searchFromDate = strtotime($options['date_start']);
```
sjpadgett, line 79: This concerns me a bit. Should this be coalesced to null. May be okay as I think $options is qualified before here unless there is another path calling this class from elsewhere, portal maybe.
Dates always get me a flutter but I trust you and will help test them once merged.
Looks good to me and thanks for the tedious work. Oh to be young again!:)

sjpadgett (approved):
Thanks Stephen. I still want to take another look at referral note and clinical notes.
I think there should be an encounter pointing at transaction to be definitive that the transaction is the referral for the document and not the latest referral. Or should all referral history be shown in an organizer. This is one of those items my gut tells me to pay attention.

We can still branch after this merge if you want or do tomorrow after some testing today.

adunsulag:
@sjpadgett Right now we don't track in the database any linkage between the encounter and the referral do we?  I think we were debating adding a direct linkage back when we did CQM but we decided to not do that if I recall.

As far as branching, I'd like to do it soon, but if we need to hold off until tomorrow I think that's fine.  I'd like to have a backup demo machine just in case something catastrophic goes on with my local environment and we'll need to give @bradymiller enough time to do that.

sjpadgett:
Okay to branch as soon as you want.
Concerning referral linkage: At the time, I thought we had an encounter id in transactions. Still there may have been a good reason that an encounter was never attached in transactions perhaps because transaction is more clinically attached to the patient rather than an encounter. I'll check the spec and ensure the template is being generated correctly or is more appropriate in a section rather than just a note.

PR #5541, opened 2022-06-23
Fixes #5539 Toggle Messages Received Confirmation

### LINKED ISSUE #5539: EMR Direct Toggle for Verify Message Received check
ONC requires that we support the Direct protocol for message disposition notification.  This allows us to ensure that a message was delivered to the intended recipient.  

Without the flag checked, EMR Direct will deliver a message to the recipients HISP (essentially their direct email provider) but does not guarantee that the recipient has received and accepted the message (security rules, network outage, etc).  

If the flag is checked then EMR Direct will ensure that the message was received and accepted by the recipient.  According to EMR Direct's documentation.  Some DIrect mail providers may not support this feature, or may have it disabled for security reasons.  If the flag is turned off then the message is automatically marked as failed since a confirmation that the message was received could not be confirmed.  In this case individuals will need to communicate with the direct recipient and determine whether to resend the message with the verify flag turned off.

By default I will have the flag turned off.  I'll set it up as a global setting that the flag will default from and I will use it as well in the care coordination module since there is no easy way to bundle yet another checkbox in that GUI.

adunsulag:
Fixes #5539 

We were not supporting the message disposition notification check with EMRDirect.  This is an ONC requirement in order to support immediate confirmation that a recipient received the message.  I added a globals setting and a flag on the messages page to support this use.

I changed up the transmitCCD and transmitMessages function to use SET FINAL flag for EMR Direct to trigger the message disposition notification check.

The carecoordination will use whatever the globals setting is set to.

@lcmaas you may be interested in reviewing this to make sure we are handling everything your company needs.

From my testing this passes the ONC 2015 (b)(1) MT 46 Disposition Notification Options test.

### DIFF
--- a/ccr/transmitCCD.php
+++ b/ccr/transmitCCD.php
@@ -38,9 +38,10 @@ use OpenEMR\Common\DirectMessaging\ErrorConstants;
  * server, the script will return "SUCCESS", otherwise it will return an error msg.
  * @param string message The message to send via Direct
  * @param string recipient the Direct Address of the recipient
+ * @param bool Whether to force receipt confirmation that the message was delivered.  Can cause message delivery failures if recipient system does not support the option.
  * @return string result of operation
  */
-function transmitMessage($message, $recipient)
+function transmitMessage($message, $recipient, $verifyFinalDelivery = false)
 {
 
     $reqBy = $_SESSION['authUser'];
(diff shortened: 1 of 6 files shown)


bradymiller:
getting about time to pray/meditate/beg to the MU3 gods for no more surprises :)

2DV-2

lcmaas:
Looks good.
FYI, the formal name for this is "final delivery notification". I would note that these notifications aren't necessarily immediate; the industry standard is to allow up to one hour for the receiving system to confirm that final delivery has occurred.

PR #5564, opened 2022-06-27
Fixes #5563 ccda provenance timezones

### LINKED ISSUE #5563: ccda lab results need timezones, fix timezone offset
Right now the time zone offset for ccda is not taking into account the locale setting for timezone.  Also the lab results author provenance timestamp is missing the timezone offset.

adunsulag:
Fix the timezone offset to handle locale settings.

Fix the lab results provenance timestamp to have timezone settings.

Also made it so the encompassingEncounter piece only shows up for episodic care notes (ie referral note, or other things that are encounter based rather than entire patient summary record based).

### DIFF
--- a/ccdaservice/serveccda.js
+++ b/ccdaservice/serveccda.js
@@ -2384,12 +2384,16 @@ function populateParticipant(participant) {
 function populateHeader(pd) {
     // default doc type ToC CCD
     let name = "Summarization of Episode Note";
+    let isEpisodicDocumentType = false;
     let docCode = "34133-9";
     let docOid = "2.16.840.1.113883.10.20.22.1.2";
     if (pd.doc_type == 'referral') {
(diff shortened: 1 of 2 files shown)


ccdaservice/serveccda.js
```diff
@@ -1201,8 +1223,8 @@ function getResultSet(results) {
             {
                 "date_time": {
                     "point": {
-                        "date": fDate(tResult.date_ordered),
-                        "precision": getPrecision(fDate(tResult.date_ordered))
+                        "date": fDate(authorDateTime),
+                        "precision": "tz"
```
sjpadgett, line 1227: at this point authorDateTime has already been through fDate(). Won't hurt though to ensure in case global formatting is removed in future.

sjpadgett (approved):
Okay with me.

PR #5566, opened 2022-06-27
Fixes #5565 referral fallback, style fixes

### LINKED ISSUE #5565: ccda referral provider fall back display, provider phone numbers
We need to display the referring provider information in a received CCDA that does not specify the REF or REFB participant.  We need to fallback to the encompasingEncounter assignedPerson if we have have one.  Otherwise we need to display the Primary Care Provider or Primary Care Physician as the referring / transitioning provider.  We also need to display their office contact information (address + phone number).

adunsulag:
Implemented the referral fallback by using the encompassing encounter provider information if we have one in the displayed CCDA, or falling back to the Primary Care Provider (PP) or the Primary Care Physician (PCP) if we have one.

In order to grab the phone and contact information I had to fix up some issues in the care team and provider templates.  Those are still displaying properly in the documentationOf and in the Care Team sections.

Everything is passing in the ETT validator and looking at the xml it looks like everything is populating correctly for care team and the documentationOf.

Fixed up a number of style issues in the cda to make the format a bit more condensed while still preserving readability.  Did this by moving address info all into one line.  Found a couple of places where headers were being pushed onto two lines that I tidied up.

### DIFF
--- a/ccdaservice/oe-blue-button-generate/lib/headerLevel.js
+++ b/ccdaservice/oe-blue-button-generate/lib/headerLevel.js
@@ -187,51 +187,12 @@ var provider = exports.provider = [{
                 attributes: leafLevel.code,
                 content: [{key: "originalText", text: "Care Team Member"}],
                 dataKey: "type"
-            }, {
-                key: "addr",
-                attributes: {
-                    use: leafLevel.use("use")
-                },
-                content: [{
-                    key: "country",
-                    text: leafLevel.inputProperty("country")
-                }, {
-                    key: "state",
-                    text: leafLevel.inputProperty("state")
-                }, {
-                    key: "city",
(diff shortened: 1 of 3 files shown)


adunsulag:
Tidied up styles
image

image

Received document with fallback of Primary Care Physician
image

adunsulag:
@sjpadgett This should be the last code fix that I'm aware of.  I've gone line by line through the  *_r21 _*.xml files  received html files and everything looks accurate.  I'm going to go through the *_r11 _*.xml files now and make sure that's correct.  If so, then we should be done and ready for the parse and display components of b1 receive. 

@sjpadgett You don't have anything else that needs to get in right?  If there's anything else outstanding lets talk.

sjpadgett (approved):
Well, good work and fingers cross:)

PR #5570, opened 2022-06-27
Openemr fix 5569 5568 5567 b1 issues

### LINKED ISSUE #5567: ccda referral provider for ccd document not populated if there is no referral transaction
So in our SLI test case Jeremy has no referral transaction created and I didn't notice until now that it means he isn't getting a participant created with a REFB participant type for the referring provider.  We need to make it fall back to the primary care physician when we create a document if there is no referral transaction.

### LINKED ISSUE #5568: ccda phone number has spaces in it for providers
Couldn't figure out where the space is coming from in the phone number for the primary care provider in our ccda generation.  Phone numbers were being created as **tel: +1(555)-555-1002** instead of **tel:+1(555)-555-1002**.

### LINKED ISSUE #5569: Medical Problems status column needs to show as Completed instead of Inactive
Just noticed that our medical problems column needs to have a status value of 'Completed' instead of 'Inactive' for the visual inspection for ONC b1 certification. 
image

adunsulag:
Fixes #5569 
Fixes #5568 
Fixes #5567 

Change status from inactive to be Completed for medical problems

Fixed the referral provider being populated as the primary care
physician for ccd documents / referral documents when there is no
referral transaction created.

Fixed the telecom phone numbers having an extra space

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
@@ -505,12 +505,19 @@ class EncounterccdadispatchTable extends AbstractTableGateway
         $participant = '';
         $records = $this->getReferralRecords($pid);
         if (empty($records[0]['refer_from']) || !is_numeric($records[0]['refer_from'])) {
-            return $participant;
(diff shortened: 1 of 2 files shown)


stephenwaite (approved):

adunsulag:
Thanks @stephenwaite !

PR #5577, opened 2022-06-29
Fixes #5576 CCDA fix null sections

### LINKED ISSUE #5576: Health Concerns and Reason for Referral break on empty section content
Our latest ccda changes add in the author provenance for #5575 but break the health concerns and reason for referral sections when they have no content (IE they don't get the NULL flavor settings).

adunsulag:
Health concerns and referral were breaking when empty due to the author provenance changes.

### DIFF
--- a/ccdaservice/oe-blue-button-generate/lib/entryLevel/careTeamEntryLevel.js
+++ b/ccdaservice/oe-blue-button-generate/lib/entryLevel/careTeamEntryLevel.js
@@ -1,6 +1,6 @@
 "use strict";
 
-var fieldLevel = require('../headerLevel');
+var headerLevel = require('../headerLevel');
 var fieldLevel = require('../fieldLevel');
 var leafLevel = require('../leafLevel');
 var condition = require("../condition");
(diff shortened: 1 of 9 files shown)


sjpadgett:
@adunsulag I guess I'll leave with you now unless you need me for something.
I figure it's best you fix as you run through testing so we don't step on each other.

adunsulag:
So I found a few more provenances that were missing and now they all show up when I generate the file directly.  However, some of the provenances are not showing up when they CCD/Referral is sent via Direct to the ETT site. I'm investigating why.

adunsulag:
It helps if I look at the right downloaded file.  Everything is working, populated, and passing the testing tools as well as the visual inspections.  I'm going to bring this in and pull it over to the mu3 branch so we can have the demo server populated with the code as a fail-safe for my local machine.  

I'm also going to play around with using for the author code the provider's taxonomy code which is a SHOULD recommendation for the right value set for CCDA, but I'm doubtful that will be required for tomorrow so I may just skip it.

bradymiller:
2DV-2

PR #5578, opened 2022-06-30
Mu3 3 author local

adunsulag:
I'll fix the commits and rebase them on master but these changes make it so the last modified provenance information comes from the individual data pieces.  I made the changes off the mu3-3 branch just to keep things safe for the SLI test.  If the hie_force_latest_encounter_provenance_date flag is set in the care coordination module it will force all of the provenance dates for the entire document to use the encounter provenance date.

I added creator and modifier ids in the USCDI core data tables where we were missing them.  Everything is passing on the ETT tools. 

My gut was getting at me that we wanted to make sure we had all of the provenance information working for the non ETT test case and any changes we make in the future would be minor bug fixes.  So here it is.

This is what I'll be using to test against this morning with SLI.


stephenwaite:
wow! 
southpark_cheering

adunsulag:
Been traveling for family events, I hope to get this updated on Tuesday.

adunsulag:
Had all kinds of issues with this working off the mu3-3 branch.  Created a new PR with all of the provenance changes and some ccda bug fixes which can be found here: https://github.com/openemr/openemr/pull/5588

Closing this PR since everything is in the other PR.

PR #5584, opened 2022-07-03
Bug fix span entire row layout option

### LINKED ISSUE #5585: New layout option span entire row bug
### Describe the bug
Looks like I forgot to catch echoing title span after not echoing the td. So empty title was being echoed outside table container causing unwanted row at top of tab view.

@stephenwaite Deduct one atta boy from my atta boy bank!:)

sjpadgett:
- was writing span outside table layout.

Fixes #5585 

Failed to prevent echo span for title column after preventing the table td being echoed for span entire row layout option causing an empty span outside the table. i.e. Appears as extra padding at top of form.


sjpadgett:
Demographics view style fixup. Gave 100% width and some weight to labels.

image

image

sjpadgett:
Since we add a mb-1 to all smallform style for selects and inputs we also need to add to Add button otherwise appended buttons will be larger:

image

sjpadgett:
@psoas Ya may want to pull this stuff in as it has a lot of fixes and cleanup for LBF.

bradymiller:
2DV

PR #5587, opened 2022-07-04
sdoh portal form dashboard fix

### LINKED ISSUE #5582: Portal encounter form audit chart form locks up page
### Describe the bug
save.php?mode=update&id=4:21 Uncaught TypeError: parent.closeTab is not a function
    at save.php?mode=update&id=4:21:8
(anonymous) @ save.php?mode=update&id=4:21

Not only page but entire session. Windows FF and Chrome and can not recover.

bradymiller:
fixes #5582


sjpadgett (approved):
Working for me Brady, thanks. I was looking in the wrong spot!:)
I like this a lot....

I need to spend a little time on sprucing up template parsing and styles but we've come a long way.

sjpadgett:
Wanted to test adding to the form via template so if we want we could add a signature required to force the form to be signed. I want to add an electronic signature with canned disclaimer for signing  with a click.

image

PR #5588, opened 2022-07-05
Mu3 3 author local cherry pick

adunsulag:
Was having all kinds of weird merge issues with my prior PR on the mu3-3 branch.  I decided to create a new branch off of the latest master, cherry-picked these commits and it is all working much nicer.  I've tested against ETT with our latest mu3 demo data as well as with a blank database.

- Add created_by and updated_by columns to patient_history, patient_data, and history_data as well as modify dates and created dates so we can at least track the last user who modified a piece of information and the user that created a piece of information for provenance purposes. A more detailed history can then be derived from the audit table log as we'll be able to see what user modified the data on each patient record.
- Change author role code to be the ccda recommended
- Added in support to populate medical problems with the provenance information from the lists table.
- Made the medical device and allergies provenance information come from the lists table with the last modifydate and username being the provenance information.
- Made it so the prescriptions table keeps track of the user that created the prescription and the user that last modified the prescription.
- Updated provenance information to use the last modified date and the last user that modified the prescription for the provenance information. For historical data the assigned provider id is treated as the provenance author.
- Left the new table columns as nullable so we don't break any legacy application use.
- Made the ccda use the provenance information from the immunizations table instead of the encounter date.
- Made the birth sex social history provenance information point to the patient_data table which is where gender is being pulled from.
- Made the social history smoking, alcohol records pull from the history_data table instead of from the encounters.
- Made it so provenance care team populates from the patient_data table if the primary care provider is set, or if the care coordination primary provider is being overridden the care team is populated from the module config settings.
- Fix allergy error for no known allergies
- Make the assessment and clinical notes templates use the provenance information from the form, and form_clinical_notes tables.
- Made goals, health concerns and care plan use the local provenance information from the form_care_plan form.
- Made the provenance information for the vitals come from the form_vitals instead of the encounter
- Made procedures use provenance from billing table instead of encounters
- Made the provenance for lab results local instead of encounter
- Add provenance db changes to new database sql, also escaped the column names in the upgrade file
- Fix ccda crashes w/ missing primary care provider.  If the primary care provider was missing the ccda crashes.  This was breaking on a new database install with a new patient who didn't have the PCP set.
- Fixed undefined array key errors.  Fixed a number of issues in the ccda generator where undefined array keys were being thrown.  Re-tested on ETT and that all is working.

### DIFF
--- a/ccdaservice/oe-blue-button-generate/lib/entryLevel/socialHistoryEntryLevel.js
+++ b/ccdaservice/oe-blue-button-generate/lib/entryLevel/socialHistoryEntryLevel.js
@@ -97,7 +97,7 @@ exports.genderStatusObservation = {
             required: true,
             dataKey: "gender"
         }
-        ,fieldLevel.author
+        ,[fieldLevel.author, contentModifier.dataKey("gender_author")]
     ],
     existsWhen: function (input) {
         return input && input.gender;
--- a/ccdaservice/serveccda.js
+++ b/ccdaservice/serveccda.js
@@ -1919,44 +1919,7 @@ function getMentalStatus(pd) {
 function getAssessments(pd) {
     return {
         "description": cleanText(pd.description),
-        "author": {
-            "code": {
-                "name": all.author.physician_type || '',
-                "code": all.author.physician_type_code || '',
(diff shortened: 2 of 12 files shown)


bradymiller (approved):

bradymiller:
done :)

PR #5595, opened 2022-07-12
prep dockers for auto upgrade

bradymiller:
prep dockers for auto upgrade


bradymiller:
looks like php 8.2 ci testing is broken, which is unrelated to us (guessing it will start working again in near future)

PR #5611, opened 2022-07-18
Contact ID Fix 2

psoas:
Fixes Contact Table Not Being Updated Correctly #5609.

Fixes bug where every time an address was modified or added for a patient, a new contact table row was added.  There is supposed to be only one contact table row per patient.

This fix is an evolution of Brady's fix.

My understanding of ORDataObject is that it is not supposed to have queries in it.  That goes against its purpose and design.  The queries should be in Services--in this case ContactService.  My fix moves the query that Brady created to ContactService.

I thoroughly tested this solution and traced it with XDebug.

### DIFF
--- a/src/Common/ORDataObject/Contact.php
+++ b/src/Common/ORDataObject/Contact.php
@@ -37,7 +37,7 @@ class Contact extends ORDataObject
     const CONTACT_TYPE_PATIENT = 'Patient';
     const CONTACT_TYPES = [self::CONTACT_TYPE_PATIENT];
 
-    public function __construct($id = "")
+    public function __construct($id)
     {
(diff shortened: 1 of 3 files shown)


bradymiller:
tenor99

2DV

sjpadgett (changes requested):
Still broken.
When there are not any additional addresses and user adds two+ new addresses then a contact reference is added for each new address to contact table.

localhost-127-0-0-1-openemr_dev-contact_address-phpMyAdmin-5-1-1

This is because we generate the records for save in memory as an array of records. This issue is most likely need to be sorted in the contact service.

I'm going to test Bradys latest and if it works then I say move on!

PR #5626, opened 2022-07-24
bug fix

bradymiller:
bug fix

### DIFF
--- a/interface/main/backup.php
+++ b/interface/main/backup.php
@@ -500,8 +500,8 @@ if ($form_step == 0) {
 }
 
 if ($form_step == 1) {
-    $form_status .= xla('Dumping OpenEMR database') . "...<br />";
-    echo nl2br($form_status);
+    $form_status .= xl('Dumping OpenEMR database') . "...||br-placeholder||";
+    echo brCustomPlaceholder(text($form_status));
     if (file_exists($TAR_FILE_PATH)) {
         if (! unlink($TAR_FILE_PATH)) {
             die(xlt("Couldn't remove old backup file:") . " " . text($TAR_FILE_PATH));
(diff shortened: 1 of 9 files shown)


interface/main/backup.php
```diff
@@ -500,8 +500,8 @@ function export_submit(step) {
 }
 
 if ($form_step == 1) {
-    $form_status .= xla('Dumping OpenEMR database') . "...<br />";
-    echo nl2br($form_status);
+    $form_status .= xl('Dumping OpenEMR database') . "...||br-placeholder||";
+    echo brCustomPlaceholder(text($form_status));
```
sjpadgett, line 504: I don't see anything messing with windows unless you do. still I'll try to test this tomorrow sometime.

sjpadgett:
I still need to test on windows. Just let me know just before you're ready to merge and I'll test.

bradymiller:
above testing well. @sjpadgett , let me know if windows passes on the backup script (i would be very surprised if those changes caused any issues in windows since substituting `<br>` rather than on os dependent newlines, but never know with windows :) ) and then I'll bring this in.

sjpadgett:
I won't get to 'til later tonight. Go ahead and merge  and I'll test tonight or tomorrow then fix if anything is amiss...

sjpadgett:
@bradymiller Where is a good place for my questionnaire_template.php default form wrapper to live in dir tree?
I was thinking openemr/templates/questionnaires or in the src/Services/DocumentTemplates or library somewhere...

bradymiller:
hi @sjpadgett , pretty sure the templates directory only holds smarty and twig templates and the src directory should only hold psr4 classes. I think @adunsulag has recommended php templates to go into a templates directory in the directory where the main script is.

PR #5628, opened 2022-07-24
Add huntr.dev to SECURITY.md

huntr-helper:
As requested through the platform by @null, this will point your security policy to [huntr.dev](https://huntr.dev/repos/openemr/openemr})

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Reporting a Vulnerability
+
+If you discover a security vulnerability in openemr please disclose it via [our huntr page](https://huntr.dev/repos/openemr/openemr/). Bounty eligibility, CVE assignment, response times and past reports are all there.
+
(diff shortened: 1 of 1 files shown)


bradymiller:
thoughts on this @stephenwaite and other admins? this would be a good option, but would also be good to have the option of sending via email (more flexible in cases of multiple issue reports that some security firms would rather use) which i think we have in another security document somewhere (in that case thinking of directing users to hello@open-emr.org address with a pgp key that the admin's can share).

SECURITY.md
```diff
@@ -0,0 +1,5 @@
+# Reporting a Vulnerability
```
sjpadgett, line 1: Sound like a good idea to me though I'm not up on security like you.

bradymiller:
updated this in new PR at #5635 so will close this

PR #5642, opened 2022-07-28
Openemr fix 5641 erx javascript

### LINKED ISSUE #5641: Refactor erx_javascript.inc.php to be a plain javascript file
Now that we have language translations available on the client side, we don't need the erx_javascript.inc.php to be a php file.  It breaks the DOM document by outputting script tags before the HTML element and makes it harder to twigify OpenEMR pages.

We should refactor it as a javascript asset and let it be loaded like our other js assets using the Header class.

adunsulag:
Fixes #5641 changed out the erx_javascript.inc.php file to be an erx_javascript.js file that can be included in the header.  This fixes some DOM issues where the script tag is placed above the html tag breaking the DOM.  It also makes it easier for users to include the script by just adding 'erx' to the header.  Finally it facilitates a future migration of the user_admin pages to twig.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -250,3 +250,9 @@ assets:
     reason-code-widget:
         basePath: '%webroot%/library/js/'
         script: reasonCodeWidget.js
+    erx:
+        basePath: '%webroot%/library/js/'
+        script: erx_javascript.js
+    checkpwd_validation:
+        basePath: '%webroot%/interface/usergroup/'
+        script: checkpwd_validation.js
--- a/interface/new/new_comprehensive.php
+++ b/interface/new/new_comprehensive.php
@@ -84,9 +84,8 @@ $fres = getLayoutRes();
 <!DOCTYPE html>
 <html>
 <head>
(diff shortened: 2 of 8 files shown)


bradymiller (approved):

config/config.yaml
```diff
@@ -249,4 +249,10 @@ assets:
...
         basePath: '%webroot%/library/js/'
-        script: reasonCodeWidget.js
\ No newline at end of file
+        script: reasonCodeWidget.js
+    erx:
+        basePath: '%webroot%/library/js/'
+        script: erx_javascript.js
+    checkpwd_validation:
+        basePath: '%webroot%/interface/usergroup/'
+        script: checkpwd_validation.js
```
bradymiller, line 258: just noted this checkpwd_validation, which doesn't look like is being used anywhere

bradymiller:
2DV

config/config.yaml
adunsulag: Yea, I plan on using it in a subsequent commit.

PR #5644, opened 2022-07-28
Fixes #5643 ModulesClassLoader autoload module code without composer install

### LINKED ISSUE #5643: Custom Modules support autoloading without needing composer command
Currently its cumbersome for someone to install a module that uses namespaces and autoloading.  Module writers currently have the option of using requires on all of their code to deal with the autoloading issue, or having to instruct their clients to add the module to the composer.json and run the ```composer dump-autoload``` command.  This requires the right version of composer to be installed, for the vendors and composer.lock file to be writeable and a number of other issues.  

What we'd like to have happen is a mechanism in place that allows module writers to just unzip a compressed module file into the OpenEMR installation and have the system be able to load the module classes without having to run any composer command.  We need to make sure this mechanism does not enable any PHP code execution until the end user has installed the module from the OpenEMR Manage Modules page (IE don't change the current mechanism of reading from the info.txt file to grab the module name).

There's been several debates about allowing OpenEMR users to autodownload modules and install them automatically on OpenEMR.   However, there's all kinds of security risks to allowing something like that to happen.  This issue is only to cover the use case of an end-user that has control of their server and can extract a zip file onto the filesystem to mitigate any remote code exploitation problems.  

If we figure out a way to securely auto-install a module we can handle that in a different issue.

adunsulag:
Implemented a modules class loader that allow module writers to register a namespace if its not already been registered.  This makes it so the ModuleWriter doesn't have to use requires everywhere and a module can be just dropped into the filesystem and execute once its been registered and installed from the Manage Modules interface.  

This is a proposed solution to the problem in #5643.  If it looks good to everyone then hopefully we can bring this in.

```php
<?php
// openemr.bootstrap.php
namespace Acme\OpenEMR\Modules\MyUniqueModule;
use OpenEMR\Core\ModulesClassLoader;

// @global ModulesClassLoader $classLoader
$classLoader->registerNamespaceIfNotExists('Acme\\OpenEMR\\Modules\\MyUniqueModule\\', __DIR__ . DIRECTORY_SEPARATOR . 'src');

// run any other custom module code needed here.
```

### DIFF
--- a/src/Core/ModulesClassLoader.php
+++ b/src/Core/ModulesClassLoader.php
@@ -35,7 +35,6 @@ namespace OpenEMR\Core;
 
 use Composer\Autoload\ClassLoader;
 
-
 class ModulesClassLoader
 {
     /**
@@ -59,10 +58,10 @@ class ModulesClassLoader
      *
      * @return void
      */
-    public function registerNamespaceIfNotExists($namespace, $paths) {
+    public function registerNamespaceIfNotExists($namespace, $paths)
+    {
         $prefixes = $this->classLoader->getPrefixesPsr4();
-        if (empty($prefixes[$namespace]))
-        {
+        if (empty($prefixes[$namespace])) {
             $this->classLoader->addPsr4($namespace, $paths);
             return true;
         }
@@ -75,7 +74,8 @@ class ModulesClassLoader
      *
      * @return void
      */
-    public function registerClassmap($classMap) {
+    public function registerClassmap($classMap)
+    {
         $this->classLoader->addClassMap($classMap);
     }
 }
--- a/src/Core/ModulesClassLoader.php
+++ b/src/Core/ModulesClassLoader.php
@@ -1,0 +1,81 @@
+<?php
+
+/**
+ * ModulesClassLoader is made available in every custom_module's openemr.bootstrap.php file as the variable $classLoader.
+ * Modules can use the ModulesClassLoader to register their namespace in order to auto load their classes.  This facilitates
+ * modules being dropped into the filesystem without having to go through a composer install / dumping of the autoloader.
+ *
+ * The class loader will check to make sure a namespace has not already been registered in case the module is installed
+ * via composer and is already in the namespace heirarchy.  Note this means each module needs to have its own DISTINCT
+ * namespace and should not share a namespace as the very first module that registers with the namespace is the one
+ * made available to OpenEMR.
+ *
+ * For performance reasons, if a module writer wishes to avoid hitting the filesystem for class discovery, they can use
+ * the registerClassmap function to optimize their runtime performance.
+ *
+ * An example use in an openemr.bootstrap.php file is as follows:
+ * <example>
+ * // openemr.bootstrap.php
+ * namespace Acme\OpenEMR\Modules\MyUniqueModule;
+ * use OpenEMR\Core\ModulesClassLoader;
+ * // @global ModulesClassLoader $classLoader
+ * $classLoader->registerNamespaceIfNotExists('Acme\\OpenEMR\\Modules\\MyUniqueModule\\', __DIR__ . DIRECTORY_SEPARATOR . 'src');
+ * // run any other custom module code needed here.
+ * </example>
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ *
+ * @author    Stephen Nielson <stephen@nielson.org>
+ * @copyright Copyright (c) 2019 Stephen Nielson <stephen@nielson.org>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Core;
+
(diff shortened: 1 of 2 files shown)


adunsulag:
@juggernautsei I know you've been interested in simplifying the module installation process.  I thought you'd be interested in looking this over.  Give me your thoughts.  It doesn't solve the auto-install problems, but hopefully will make it easier to install modules as you can just unzip the module in the custom_modules directory.

sjpadgett:
@adunsulag I don't like it!

adunsulag:
@sjpadgett Lol, my email response only showed your first line and not the response.  I clicked the link with trepidation and saw the rest of your response :)  Had my heart pounding there for a second.

sjpadgett:
@adunsulag I thought that'd wake you up. I have a solution for chucking data at least on windows. Kind of weird but I get it in the end why diff on linux.

src/Core/ModulesClassLoader.php
```diff
@@ -0,0 +1,81 @@
...
+ * For performance reasons, if a module writer wishes to avoid hitting the filesystem for class discovery, they can use
+ * the registerClassmap function to optimize their runtime performance.
+ *
+ * An example use in an openemr.bootstrap.php file is as follows:
+ * <example>
+ * // openemr.bootstrap.php
+ * namespace Acme\OpenEMR\Modules\MyUniqueModule;
+ * use OpenEMR\Core\ModulesClassLoader;
+ * // @global ModulesClassLoader $classLoader
+ * $classLoader->registerNamespaceIfNotExists('Acme\\OpenEMR\\Modules\\MyUniqueModule\\', __DIR__ . DIRECTORY_SEPARATOR . 'src');
```
bradymiller, line 22: 118382341-53d9d600-b5c2-11eb-8b7e-220bbab8468a

bradymiller (approved):

juggernautsei:
> @adunsulag I don't like it! You're suppose to say "Why not?" so I can say "Because I didn't think of it!"
> 
> You left me hanging!

I am with Jerry on this one that I don't like it. 
In my opinion, the module should not from my experience, and does not need to interact with the core composer. I have never had to make changes to the core composer. That to me is bad designing if the module has to change anything about the core composer. The module should be completely self-contained and only talk to the core system to exchange information and act upon. 

Sorry, I am late to the party. I see that Brady has already brought it into the codebase, @bradymiller, I implore you to reverse that decision. I think it will lead to more complications in the long-term of module building.

adunsulag:
@juggernautsei the code is not merged in.... Nor is it interacting with core composer as its following a standard PSR-4 namespace autoloading function (its a PHP thing not a composer thing).  Its interacting with OpenEMR's autoloader which implicitly uses core composer but does not need to use it.  This decouples OpenEMR from having anything to do with Composer if we ever decide to move away from composer.

I hadn't considered the use case of module writers just bundling a nested composer autoloader which I suppose satisfies dropping a module into the system.  The nice thing is it lets the module owner stay in control of their namespace loading and class map optimizations.  The downsides is having multiple autoloaders which can lose out on any benefits that the overall OpenEMR can implement (APC caching, HHVM, namespace collisions, code signing for authenticity, etc, ). 

Also, if you go with the embedded autoloader, I'm guessing that prevents people from installing the module via composer if its available.  Otherwise you'll end up with duplicate class / namespace definitions, correct? Since it will use the OpenEMR classloader to load your PSR4 definition, and then the openemr.boostrap.php will load the definitions again with the embedded namespace autoloader.  It seems like if you go with the embedded composer you will forced to distribute it only as a zip file and never via composer.  Please correct me if I'm mistaken here.

I'm not seeing the active harm it creates, if anything its another mechanism for module writers to do namespace registration without having to rely on a nested composer.  I'm open to being persuaded otherwise though.

juggernautsei:
@adunsulag I will consider the things you presented and get back to you in a few hours.

juggernautsei:
I'm going to address this.

"Also, if you go with the embedded autoloader, I'm guessing that prevents
people from installing the module via composer if its available"

Using composer to install a module is not prevented if the module has a
nested composer instance.

juggernautsei:
@adunsulag 
I actually took time to read and digest the code proposal better. I see your design desire to bring the namespace of the module into the OpenEMR autoloader. I still don't think it is necessary. I do all of my autoloading from within the module. I still go back to your suggestion to use Juggernauts as my namespace hence there will be no collision with other namespaces in the program. I do know in programming is better safe than sorry. So, have away to manage namespace is what I see your changes are doing. 
But what happens when there is a namespace collision?

sjpadgett (approved):
Very nice and useful IMO! 
To me it's just an additional option available to module writers to help manage namespace.
Seems pretty clear namespace collisions are handled if module writer is checking result of the registration.
I suppose we could force an exception to registerNamespaceIfNotExists() to break workflow and force developer to pay attention but, up to @adunsulag if necessary.

adunsulag:
If there is a namespace collision it will only utilize the first namespace.  I think that's fine.

PR #5650, opened 2022-08-01
Fixes for saving History data in LBFs and for reading History data wi…

### LINKED ISSUE #5649: LBFs and Patient History Data
### Describe the bug

When a LBF instance referencing multiple history fields is saved, multiple history_data table rows are created with identical timestamps.

Also, if the most recent history_data rows for the same patient have duplicate timestamps for the above reason or any other reason, the most recent row might not be retrieved when getting current history values.

I will create a PR to fix this.

### To Reproduce

1. Create a LBF layout with multiple History source items.
2. Create an instance of the LBF in a patient visit, entering data for these multiple fields.
3. Query the history_data table to observe that multiple records are created with the same timestamp.

### Expected behavior

Only one record should be created.

### Client configuration

Not relevant.

sunsetsystems:
…th identical timestamps.

Fixes #5649

### DIFF
--- a/interface/forms/LBF/new.php
+++ b/interface/forms/LBF/new.php
@@ -8,7 +8,7 @@
  * @author    Rod Roark <rod@sunsetsystems.com>
  * @author    Brady Miller <brady.g.miller@gmail.com>
  * @author    Jerry Padgett <sjpadgett@gmail.com>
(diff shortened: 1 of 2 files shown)


sunsetsystems:
Simple fix + successful testing = bringing in. :)

bradymiller:
@sunsetsystems , thanks for the fix! i just brought it into rel-700

2DV

PR #5658, opened 2022-08-03
Manage CCDA data transmission to service

### LINKED ISSUE #5657: CCDA generation fails for large documents
### Describe the bug
Due to an apparent Windows issue, we had to limit streaming content to ccda service to 128Kb.
Solution is to manage socket buffer size and chuck send array content

### Client configuration

- Browser: All
- OpenEMR version: 7.0.0
- Operating system: windows linux

sjpadgett:
- CCM chunk send data to service
- add data stacking buffer to service
- honor EOT(file separator character) in service
- replace deprecated js substr with substring

Fixes #5657 

Allows sending ccda content xml array for content larger than 128kb. Theoretically this will allow any array size.
Co-Authored with @adunsulag

### DIFF
--- a/ccdaservice/serveccda.js
+++ b/ccdaservice/serveccda.js
@@ -26,6 +26,39 @@ var webRoot = "";
 var authorDateTime = '';
 var documentLocation = '';
 
+class DataStack {
+    constructor(delimiter) {
+        this.delimiter = delimiter;
+        this.buffer = "";
+    }
+
+    endOfCcda() {
+        return this.buffer.length === 0 || this.buffer.indexOf(this.delimiter) === -1;
+    }
+
+    pushToStack(data) {
+        this.buffer += data;
+    }
+
+    fetchBuffer() {
+        const delimiterIndex = this.buffer.indexOf(this.delimiter);
+        if (delimiterIndex !== -1) {
+            const bufferMsg = this.buffer.slice(0, delimiterIndex);
+            this.buffer = this.buffer.replace(bufferMsg + this.delimiter, "");
+            return bufferMsg;
+        }
+        return null
+    }
+
+    returnData() {
+        return this.fetchBuffer();
+    }
+
+    clearStack() {
+        this.buffer = "";
+    }
+}
+
 function trim(s) {
     if (typeof s === 'string') return s.trim();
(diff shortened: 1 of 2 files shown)


ccdaservice/serveccda.js
```diff
@@ -26,6 +26,39 @@ var webRoot = "";
...
+
+    pushToStack(data) {
+        this.buffer += data;
+    }
+
+    fetchBuffer() {
+        const delimiterIndex = this.buffer.indexOf(this.delimiter);
+        if (delimiterIndex !== -1) {
+            const bufferMsg = this.buffer.slice(0, delimiterIndex);
+            this.buffer = this.buffer.replace(bufferMsg + this.delimiter, "");
```
adunsulag, line 47: couldn't you get the same effect by just doing 
```this.buffer = "";```
Not sure why you need the this.buffer.replace here.

sjpadgett: In current use case this may be true.  But if I wanted to read the buffer before delimit, I don't want to destroy stack.
This made sense to me. Do you see a potential issue?

adunsulag: So you've already grabbed the bufferMsg in line 46.  Line 47, if I'm reading it correctly just sets this.buffer to be an empty string.  I didn't see why you need the replace operation when you can just assign this.buffer = ""; on line 47.

sjpadgett: Just because I'm not currently validating data doesn't mean I won't in future. The intent is to be able to read the current stack at anytime but not destroy stack until the end of file is received.
Being lazy, I probably should have added a new function to scan buffer to avoid confusion.
I could add if you feel strongly

sjpadgett: Say I want to fetch by different delimiter like new line and process received data line by line. I could just return out of receive to a function that pulls from the stack line by line by resetting delimiter from eof to new line i.e FIFO.
Even though this is not the way it is used currently doesn't mean it won't be in future either here or a different workflow using service.

adunsulag (approved):
Looks good.

sjpadgett:
Will pick to v7 branch for patch

PR #5662, opened 2022-08-03
insane dev docker update to default to alpine 3.16

bradymiller:
insane dev docker update to default to alpine 3.16


stephenwaite:
jumping

PR #5663, opened 2022-08-04
Dynamic Encounter Questionnaires

### LINKED ISSUE #5661: Implement Questionnaire results management.
### Describe the problem
This is the second phase of the FHIR Questionnaire and assessments feature. Need to add the resulting answers resource save and retrieve generic across all possible use cases i.e. from API Questionnaire/QuestionnaireResponse, SMART or other rendered forms.

I'm going to do this part at the same time I'm developing an encounter form implementation of questionnaires retrieved through API lookup by LOINC codes or description or from inputted FHIR server URL.

I think this will be a very nice feature. I also wonder if there is a clinical need to render questionnaires from demographic? Perhaps integrate with the Easipro workflow in demographics to allow doing questionnaires from there.

I'm taking suggestions for some default questionnaires/assessments to offer as core items in OpenEMR too.

Wish me luck!:)

sjpadgett:
- new table
- uuid registry

Fixes #5661 

This feature allows importing new LOINC panels or questionnaires from NIH as encounter forms. Once imported, user may simply fill and save imported panel and save for that encounter or register import as a new questionnaire form similar to using Forms Administration.


sjpadgett:
I wanted to put up the table schema for questionnaire response in case anyone wants to comment or offer suggestions.

sjpadgett:
Decided to collapse the AMC questions in encounter. Annoying and to much view real estate..

image

image

sjpadgett:
Implemented a disclaimer as to openemr's responsibility for imported forms. @adunsulag @bradymiller how's content?

image

sjpadgett:
I think I have this pretty well worked out and will merge once peer review is approved.

Next phase will be working out the portal questionnaire management and user audit. I also need to add a report to this encounter feature but I'm waiting until I build methods in service to parse answers which will be somewhat involved. 

Portal will use questionnaire_response table where hopefully structure will be compatible with upcoming FHIR Questionnaires API's.

I think this, coupled with portal questionnaire implementation, will be a great addition to openemr.

interface/forms/questionnaire_assessments/questionnaire_assessments.php
```diff
@@ -0,0 +1,191 @@
...
+                parent.closeTab(window.name, false);
+            }
+            document.getElementById('questionnaire').value = JSON.stringify(qFhir);
+            document.getElementById('lform').value = JSON.stringify(data);
+            document.getElementById('form_name').value = jsAttr(formName);
+            document.getElementById('code_type').value = jsAttr('LOINC');
+            document.getElementById('code').value = jsAttr(data.code);
+            if (typeof data.copyrightNotice !== 'undefined' && data.copyrightNotice > '') {
+                document.getElementById('copyright').value = jsAttr(data.copyrightNotice);
+                document.getElementById('copyrightNotice').innerHTML = jsText(data.copyrightNotice);
```
bradymiller, line 87: agree when inserting html via innerHTML that is needs to be html escaped, however when setting `.value`'s above, pretty sure do not need to escape that.
(honestly, though, I don't have a 100% grasp on when need to escape js for html)

interface/forms/questionnaire_assessments/questionnaire_assessments.php
```diff
@@ -0,0 +1,191 @@
...
+            <div class="form-group">
+                <div class="input-group isNew">
+                    <label for="loinc_item" class="font-weight-bold mt-2 mr-1"><?php echo xlt("Search and Select a LOINC form") . ': '; ?></label>
+                    <input class="form-control search_field" type="text" id="loinc_item" placeholder="<?php echo xla("Type to search"); ?>" autocomplete="off" role="combobox" aria-expanded="false">
+                </div>
+                <div>
+                    <p class="text-center"><?php echo "<span class='font-weight-bold'>" . xlt("Note") . ": </span><i>" . xlt("LOINC form definitions are subject to the LOINC"); ?>
+                        <a href="http://loinc.org/terms-of-use" target="_blank"><?php echo xlt("terms of use.") . "</i>"; ?></php></a></p>
+                    <p id="copyrightNotice">
+                        <?php echo text($form['copyright'] ?: ''); ?>
```
bradymiller, line 155: would only use elvis operator when it is needed since it will still throw a php warning when empty (unlike use of `::` which will not throw a php warning when empty)

interface/forms/questionnaire_assessments/report.php
```diff
@@ -0,0 +1,12 @@
+<?php
```
bradymiller, line 1: would ensure all the scripts have a header (copyright/license stuff)

interface/forms/questionnaire_assessments/report.php
```diff
@@ -0,0 +1,12 @@
...
+
+require_once(dirname(__FILE__) . '/../../globals.php');
```
bradymiller, line 3: drop above line. not needed (I realized this when it was causing issues on the form to portal work i did in past). at some point gonna remove this line for all the forms.

interface/forms/questionnaire_assessments/report.php
```diff
@@ -0,0 +1,12 @@
...
+require_once($GLOBALS["srcdir"] . "/api.inc");
+
+function questionnaire_assessments_report($pid, $encounter, $cols, $id)
+{
+    $count = 0;
+    $data = formFetch("form_questionnaire_assessments", $id);
+    if ($data) {
```
bradymiller, line 10: guessing haven't gotten to showing the questionaire results yet.

interface/forms/questionnaire_assessments/table.sql
```diff
@@ -0,0 +1,22 @@
+--
```
bradymiller, line 1: wouldn't even include this file since this is installed by default (then avoid needing to remember to update tables in 2 places when make changes in schema).

sql/7_0_0-to-7_0_1_upgrade.sql
```diff
@@ -128,3 +128,60 @@ CREATE TABLE `questionnaire_repository` (
...
+  `form_name` varchar(255) DEFAULT NULL,
+  `code` varchar(31) DEFAULT NULL,
+  `code_type` varchar(31) DEFAULT "LOINC",
+  `questionnaire` longtext,
+  `questionnaire_response` longtext,
+  `lform` longtext,
+  `lform_response` longtext,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB;
+INSERT INTO `registry` (`name`, `state`, `directory`, `id`, `sql_run`, `unpackaged`, `date`, `priority`, `category`, `nickname`, `patient_encounter`, `therapy_group_encounter`, `aco_spec`, `form_foreign_id`) VALUES ('New Questionnaire\n', 1, 'questionnaire_assessments', 28, 1, 1, '2022-08-04 14:45:15', 0, 'Questionnaires', '', 1, 0, 'encounters|notes', NULL);
```
bradymiller, line 186: remove the `\n` at the end of the name entry

bradymiller: should not be hard-coding the id; should let the insert take care of that.

bradymiller:
@sjpadgett , Are we letting providers look up and bring in forms? If so, this could be really problematic (the clinic will not have standard forms and providers will just ignore any copyright issues (I generally ignore most popups/modals/warnings especially if not clinically relevant). Probably better to only allow clinic admins to bring in forms via the forms gui and patient portal templates gui.

interface/forms/questionnaire_assessments/questionnaire_assessments.php
sjpadgett: lol aw come on now!

interface/forms/questionnaire_assessments/questionnaire_assessments.php
sjpadgett: All coalesce does is prevent undefined (strongly typed in c) from throwing. A purposeful setting to null should not throw. I don't want null sent to interface thus the conditional! copyright in DB is default null.

interface/forms/questionnaire_assessments/report.php
sjpadgett: yep prob not needed.

interface/forms/questionnaire_assessments/report.php
sjpadgett: I put in post comment why...

interface/forms/questionnaire_assessments/table.sql
sjpadgett: True. I pondered to leave or not but your reason is a good one. will remove.

sql/7_0_0-to-7_0_1_upgrade.sql
sjpadgett: Just followed what is currently there. A value must be populated or will error using the current format of values only to insert.

sjpadgett:
> Are we letting providers look up and bring in forms?

good point. I'll add an ACL when called to search.
Thanks for the review Brady.

interface/forms/questionnaire_assessments/questionnaire_assessments.php
sjpadgett: I think I agree that normally an escape wouldn't be needed but when I look to escape I look at the source of the data. If from outside of my control say from web as in this case, I tend to be leery and escape. Although, we know I'm not know as an escaping king by no means.
I went back and forth on this and tested both ways then decided to err on being safe. If you feel strongly I can remove.

interface/forms/questionnaire_assessments/questionnaire_assessments.php
sjpadgett: however you're correct here because a new questionnaire is undefined form rows. sorry, I appear to not be with it lately!
I did notice that we are in for some trouble soon with deprecated casting in areas e.g
`[09-Aug-2022 10:08:19 America/New_York] PHP Deprecated:  Automatic conversion of false to array is deprecated in C:\xampp\htdocs\openemr\src\Gacl\Gacl.php on line 569`

soon we may need to get all of together to turn on deprecated warnings and errors and mass fix these.

I/we have a theme issue with lforms as it is themed to BS3. I'd have to fork and refactor lform project and contribute back to project or we maintain our version of lforms. I know how to do if my brain would cooperate but unsure if I want to tackle just now!

sjpadgett:
Best way I can thin of just now unless I go into menu auth system.
image

sjpadgett:
> patient portal templates gui.

@bradymiller 
Must be admin to do anything in portal. This needs to change because other than admin needs ability to use secure messaging, audits, chat etc.
When did this change?

PR #5682, opened 2022-08-11
Fixes #5681 Header events for js/css

### LINKED ISSUE #5681: Provide event hooks to add javascript or styles to anywhere the Header class is used
Instead of adding event hooks for specific pages across OpenEMR to inject javascript or stylesheets in from a module we should add event listeners to the main Header class that can add additional Javascript / Stylesheet assets as part of our header asset bundling process.  The event hooks should pass in the name of the currently executing script so that module writers can add / inject assets conditionally based upon the page they are on or inject them on every single page where the Header class is utilized.

adunsulag:
Fixes #5681 
Added the ability to inject javascript or css into any page inside
OpenEMR using the Header class specifically the setupHeader method which
is used in smarty, twig, and our regular php files all over the place to
bring in javascript and css assets.  This allows module / event writers
to inject content on every page or on specific pages based on the script
name / page name.

### DIFF
--- a/src/Core/Header.php
+++ b/src/Core/Header.php
@@ -9,6 +9,8 @@
 namespace OpenEMR\Core;
 
 use OpenEMR\Common\Logging\SystemLogger;
+use OpenEMR\Events\Core\ScriptFilterEvent;
+use OpenEMR\Events\Core\StyleFilterEvent;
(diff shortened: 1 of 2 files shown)


bradymiller (approved):

bradymiller:
2DV-2

PR #5684, opened 2022-08-11
Fixes #5683 user admin events

### LINKED ISSUE #5683: Add events to User Admin pages to render html content
I'd like to add events to the user admin pages to render html content from within a module.  It would be nice to have events to display content before the user fields and after the user fields when editing or creating a user.  The content should be contained inside of a div to allow us to move things around if we decide to drop the user table structure for responsiveness.

adunsulag:
Fixes #5683 adds events to user_admin and usergroup_admin_add page in
order to be able to render content on those pages from a module.

### DIFF
--- a/interface/usergroup/user_admin.php
+++ b/interface/usergroup/user_admin.php
@@ -29,6 +29,7 @@ use OpenEMR\Menu\MainMenuRole;
 use OpenEMR\Menu\PatientMenuRole;
 use OpenEMR\Services\FacilityService;
 use OpenEMR\Services\UserService;
+use OpenEMR\Events\User\UserEditRenderEvent;
 
 if (!empty($_GET)) {
     if (!CsrfUtils::verifyCsrfToken($_GET["csrf_token_form"])) {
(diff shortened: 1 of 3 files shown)


interface/usergroup/user_admin.php
```diff
@@ -605,6 +617,19 @@ class="form-control" value="<?php echo attr($frow['id']); ?>"><?php echo text($f
...
         </td>
     </tr>
+    <tr>
+        <td colspan="4">
+            <?php
+            // TODO: we eventually want to move to a responsive layout and not use tables here.  So we are going to give
+            // module writers the ability to inject divs, tables, or whatever inside the cell instead of having them
+            // generate additional rows / table columns which locks us into that format.
+            $postRenderEvent = new UserEditRenderEvent('user_admin.php', $_GET['id']);
+            error_log("rendering event");
```
bradymiller, line 627: remove above debug line before bringing in

bradymiller (approved):

interface/usergroup/user_admin.php
adunsulag: Thanks for the catch!

PR #5692, opened 2022-08-16
Rel 610 Bring in  #5643 #5681 #5683 #5685

adunsulag:
Fixes https://github.com/openemr/openemr/issues/5643 ModulesClassLoader

Implemented a modules class loader that allow module writers to register a namespace if its not already been registered.  This makes it so the ModuleWriter doesn't have to use requires everywhere and a module can be just dropped into the filesystem and execute once its been registered and installed from the Manage Modules interface.

Style fixes

Fixes https://github.com/openemr/openemr/issues/5681 Added the ability to inject javascript or css into any page inside OpenEMR using the Header class specifically the setupHeader method which is used in smarty, twig, and our regular php files all over the place to bring in javascript and css assets.  This allows module / event writers to inject content on every page or on specific pages based on the script name / page name.

Fixes https://github.com/openemr/openemr/issues/5685 add provider calendar methods

Fix spelling, improve sql escaping 
Fixes https://github.com/openemr/openemr/issues/5683 user admin events

Fixes https://github.com/openemr/openemr/issues/5683 adds events to user_admin and usergroup_admin_add page in order to be able to render content on those pages from a module.

Remove debug line

### DIFF
--- a/interface/main/calendar/modules/PostCalendar/pnuserapi.php
+++ b/interface/main/calendar/modules/PostCalendar/pnuserapi.php
@@ -326,13 +326,14 @@ function postcalendar_userapi_buildView($args)
 
                 //==================================
                 //FACILITY FILTERING (CHEMED)
+        $userService = new UserService();
         if ($_SESSION['pc_facility']) {
-            $provinfo = getProviderInfo('%', true, $_SESSION['pc_facility']);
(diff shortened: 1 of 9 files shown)


adunsulag:
I've tested this and it appears to be working correctly.  I'll leave it up for a day and if no one has any objections, I'll bring it into the 6.1.0 release.

PR #5706, opened 2022-08-29
Openemr fix 5704 5705 pkce oauth2 clients

### LINKED ISSUE #5704: Enable PKCE support for OAuth2 clients
We should enable PKCE support for the oauth2 clients, both public apps and confidential apps.  We need to adjust both our smart well-known configuration to broadcast support for PKCE as well as change up our authorization_grant settings to support the PKCE functionality.

### LINKED ISSUE #5705: Change OAuth2 approval settings to Auto Approve patient standalone apps
I discovered that ONC does NOT allow any vetting of a patient's chosen 3rd party app as long as the app is restricted to soley being a patient standalone app and meets the security requirements of all OpenEMR patient standalone apps.  As an admin has to pretty much auto-approve these applications we should simplify the administrative burden on OpenEMR administrators and change our Oauth2 3rd party app approval process to be auto approved for patient standalone apps.

Currently a public app is auto-approved in OpenEMR, whereas all confidential apps are required to be manually approved.  We should change up the logic so that if a confidential app requests ONLY patient/* scopes and not user/* or system/* scopes the app is auto approved.  

We should also add a global setting if a user wishes to change the logic to require manual approval of all apps.  Note that if a user goes this route and they must comply with US healthcare laws then they MUST approve the app within 48 hours according to a lecture I attended from the SMART on FHIR group.  I'm still trying to track down the specific requirement as it looks like it might be a MIPS thing but I'm still waiting on details for that.

As part of this change we should allow public apps to have the offline_access scope if it is a Patient Standalone scope.  A public app must have a way of securely storing the refresh token in their application in a secure enclave if the public app is a native mobile application.  Patients will be notified that the app is requesting offline access and should be careful in granting this permission.

I will be posting a preliminary PR but I'd love any feedback from the community on any security issues they see with this approach.

adunsulag:
Fixes #5704 
Fixes #5705 

Put in support for PKCE and restricted it to only support the S256
code_challenge_method per the SMART on FHIR requirements.

Added to our smart configuration the PKCE support as well as the
capability that we only support the v1 smart scopes at this point in
time.

Added a global flag that controls how app authorizations are done.  US
jurisdictions need to use the auto-enabling of patient standalone apps
or need to manually approve patient apps within 48 hours.  Other
jurisidictions can turn off all app approval if they wish to have
tighter control of their security.

Fixed some bugs with the OAuth2 authorization_code, password, and
refresh grants.

Allowed offline_access scope for public patient standalone apps.
Patients will be warned when offline_access is granted.

Updated API documentation per ONC requirements for PKCE information.


bradymiller:
code looks good to me :)
(note api testing is failing on the ci stuff)

adunsulag:
So the API tests failing were actually a good sign that things were working :)  Fixed the API clients to enable the password grant clients.

PR #5717, opened 2022-08-31
Openemr fix 5704 5705 pkce oauth2 clients (#5706)

adunsulag:
* #5704, #5705 PKCE, OAuth2 app approvals

Put in support for PKCE and restricted it to only support the S256
code_challenge_method per the SMART on FHIR requirements.

Added to our smart configuration the PKCE support as well as the
capability that we only support the v1 smart scopes at this point in
time.

Added a global flag that controls how app authorizations are done.  US
jurisdictions need to use the auto-enabling of patient standalone apps
or need to manually approve patient apps within 48 hours.  Other
jurisidictions can turn off all app approval if they wish to have
tighter control of their security.

Fixed some bugs with the OAuth2 authorization_code, password, and
refresh grants.

Allowed offline_access scope for public patient standalone apps.
Patients will be warned when offline_access is granted.

Updated API documentation per ONC requirements for PKCE information.

* Fix stlye issues

* Fix API Tests use enabled client

Discovered the api tests were failing which is GREAT as it means some of
our bug fixes are working.  Disabled clients using the password grant
were being rejected.  We have to enable the client in order for the
tests to communicate via the api.


bradymiller:
2DV

PR #5712, opened 2022-08-30
Rel 700

omegasystemsgroup:
Fixes #5711 

#### Short description of what this resolves:
Users couldn't dispense drugs using the charges panel.

#### Changes proposed in this pull request:
Dispense the drug using the current/selected encounter.

PR #5722, opened 2022-09-02
Fix Drugs dispensing in Charges Panel

omegasystemsgroup:
The current encounter was not used in dispensing drugs. Default encounter value of 0 was used. So we take the encounter from the session and use it accordingly.

Fixes #5712

Internal error: the referenced encounter no longer exists. 1 0

Get the current encounter and use this in dispensing a drug via the charges panel

### DIFF
--- a/interface/drugs/dispense_drug.php
+++ b/interface/drugs/dispense_drug.php
@@ -46,6 +46,7 @@ $prescription_id = $_REQUEST['prescription'];
 $quantity        = $_REQUEST['quantity'];
 $fee             = $_REQUEST['fee'];
 $user            = $_SESSION['authUser'];
+$encounter       = $_SESSION['encounter']
 
 if (!AclMain::aclCheckCore('admin', 'drugs')) {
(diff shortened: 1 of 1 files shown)


interface/drugs/dispense_drug.php
```diff
@@ -46,6 +46,7 @@ function send_email($subject, $body)
 $quantity        = $_REQUEST['quantity'];
 $fee             = $_REQUEST['fee'];
 $user            = $_SESSION['authUser'];
+$encounter       = $_SESSION['encounter'] :: 0;
```
bradymiller, line 49: (this is my fault since I gave you wrong command during the conference call. Should use this command instead of above):
$encounter = $_SESSION['encounter'] ?? 0;

omegasystemsgroup: Yeah, should've known better. Thanks. I created a new commit. Should create a new PR?

bradymiller:
hi @omegasystemsgroup , Looks good and bringing this in. Thanks for the contribution!

2DV

omegasystemsgroup:
Thanks.

PR #5727, opened 2022-09-05
deal with legacy smarty

bradymiller:
Time to deal with legacy smarty issues (legacy smarty will break in php 9 which guessing will be released in little more than a year)
Plan:
- Migrate legacy smarty to within openemr (Smarty_Legacy class) and use for the calendar (since this is in process of being converted to twig by @adunsulag ).
- Use modern smarty for core openemr and gacl.

Future plan that @adunsulag is considering:
- calendar to twig (ongoing)
- gacl and core (including forms) to twig (give form users a version warning that smarty support in forms will be dropped)


sjpadgett:
I believe @adunsulag has a local project and is gradually converting calendar to twig.

stephenwaite:
bradys-panda

PR #5729, opened 2022-09-06
Update SASS Packages

### LINKED ISSUE #5726: NPM Dependencies
I have rewrote this issue as some things can't be fixed until we upgrade to BS5.

I'll be updating some of the dependencies for SASS in an upcoming PR. This includes swapping out the old dart-sass dependency with the new one that is more up to date.

tywrenn:
Fixes #5726 

This updates dart-sass to the new way it needs to run and updates NPM dependencies. Note that this may increase warnings, but this gives us more of a reason to upgrade to Bootstrap 5.


bradymiller:
just making sure this will work for the arm docker build (i just recall there being issues with sass in past for arm)

tywrenn:
I'll check my raspberry pi for ARM issues...

bradymiller:
just so you don't spin your wheels too much no this, note that only arm64 dockers are now being built (in addition to the amd64), since arm7 broke awhile ago and was discussed here:
https://github.com/openemr/openemr-devops/issues/321

tywrenn:
@bradymiller arm64 might be supported on Raspberry Pi, just can't use 32 bit operating systems to build your code. I'll see if I can build it.

bradymiller:
hi @tywrenn , one good option is that i can just bring this into the codebase now and then see if the flex-edge build works (at this time, i basically build them with the click of a github action) and if the build break can revert it :)

PR #5730, opened 2022-09-06
Fix Navigation Spacing

### LINKED ISSUE #5723: Navigation Bar Has Inconsistent Spacing
Hey 🙂 It's been awhile since I've looked at the UI of this, but it seems like the navigation bar has inconsistent spacing throughout all themes you switch through. Were we trying to get rid of some spacing on the navigation bar at all? If so, we did it incorrectly. Please advise on what was done as I might make some fixes to this in my free time.

tywrenn:
Fixes #5723

This fixes the navigation spacing issue we had as discussed in my issue request. It also creates/starts a README file in the themes folder so that developers can understand what things are where in this folder as it seems that hasn't been written down.

This is an ongoing PR that I'm not finished yet with -- i'll let you all know when I am done.


tywrenn:
Closing issue due to new findings, new issue will be created soon.

PR #5733, opened 2022-09-10
Fix dispense drug from charges panel

### LINKED ISSUE #5711: Unable to Dispense in OPENEMR 6.1 and 7.0
### **Internal error: the referenced encounter no longer exists. 1 0** 

The logged in user cannot dispense drugs using the charges panel under Fees-> Charges.

In older versions, the drug label is what shows in a dialog, followed by printing. But this is not the case in V6.1 and 7.0

omegasystemsgroup:
Fixes #5711

### DIFF
--- a/interface/drugs/dispense_drug.php
+++ b/interface/drugs/dispense_drug.php
@@ -46,6 +46,7 @@ $prescription_id = $_REQUEST['prescription'];
 $quantity        = $_REQUEST['quantity'];
 $fee             = $_REQUEST['fee'];
 $user            = $_SESSION['authUser'];
+$encounter            = $_SESSION['encounter'] :: 0;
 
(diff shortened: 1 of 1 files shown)


interface/drugs/dispense_drug.php
```diff
@@ -46,6 +46,7 @@ function send_email($subject, $body)
 $quantity        = $_REQUEST['quantity'];
 $fee             = $_REQUEST['fee'];
 $user            = $_SESSION['authUser'];
+$encounter            = $_SESSION['encounter'] :: 0;
```
bradymiller, line 49: (this is my fault since I gave you wrong command during the conference call. Should use this command instead of above):
$encounter = $_SESSION['encounter'] ?? 0;

bradymiller:
this is duplicate PR as https://github.com/openemr/openemr/pull/5722

PR #5757, opened 2022-09-16
Questionnaire encounter report

### LINKED ISSUE #5677: Handle FHIR QuestionnaireResponse Resource
### Describe the feature
As I muddle my way through this whole FHIR Questionnaire implementation, my destination of import is capturing the answer response along with making available the Questionnaire to the FHIR API leading to an example SMART app.

Currently I've implemented FHIR Questionnaire in Portal and Encounters. The encounter implementation saves the answer response to a  forms table rather than the questionnaire response table. I wanted to keep the workflow contained to the normal encounter form workflow and patterns for consistency.
For the portal though, I want to follow the same response management that the FHIR API would follow killing two birds one might say.
The hardest part of portal is the user audit. I hope to design a pattern that will be available for reporting to and audit by the original requester of the questionnaire for the API as well. At that point the response is historically saved using the completed status(in-progress | completed | amended | entered-in-error | stopped) and user meta data.

For those interested in the FHIR API, SMART and these features in general, I'd hope they would chime in if having concerns for my approach or questions. The more folks involved the better the outcome IMO.

sjpadgett:
- two versions of reports. Column and flex row reports. Column is default
- fix if missing lForm form data then it is created from the questionnaire json or visa versa.
- encounter can now create new form from just a FHIR questionnaire.

Fixes #5677 

See issue for example of reports

### DIFF
--- a/interface/forms/questionnaire_assessments/questionnaire_assessments.php
+++ b/interface/forms/questionnaire_assessments/questionnaire_assessments.php
@@ -271,9 +271,9 @@ $q_list = $service->getQuestionnaireList(true);
     <div class="container-xl my-2">
         <div class="title"><h3><?php if ($mode != 'new_form' && $mode != 'update') {
                     echo xlt("Create Encounter Questionnaires");
-                } else {
-                    echo xlt("Edit Questionnaire");
-                } ?></h3></div>
+                               } else {
+                                   echo xlt("Edit Questionnaire");
+                               } ?></h3></div>
         <?php if (!$is_authorized) { ?>
             <div class="d-flex flex-column w-100 align-items-center">
                 <?php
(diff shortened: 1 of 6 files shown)


sjpadgett:
Implemented feature to select and create new encounter registry form or do a one time questionnaire from our questionnaire repository.
I defy anyone that says this isn't way cool stuff! Also I'm running low on atta boys.:)

image

image
continued bottom of form
image

image

New form

image

sjpadgett:
I'd like to merge before dev demo reset. 
Perhaps I can record a quick tutorial on Sat call using the demo server.
Hoping to get some testers going for feedback before I move on to portal audit and API endpoints

bradymiller:
2DV

PR #5740, opened 2022-09-11
drop support for php 7.4

bradymiller:
Preparing to drop support for php 7.4. First need to deal with following composer issues:
```
  Problem 1
    - Root composer.json requires paragonie/multi-factor 0.2.2 -> satisfiable by paragonie/multi-factor[v0.2.2].
    - paragonie/multi-factor v0.2.2 requires php ^7 -> your php version (8.0; overridden via config.platform, actual: 8.1.9) does not satisfy that requirement.
  Problem 2
    - laminas/laminas-console 2.6.0 requires php ^5.5 || ^7.0 -> your php version (8.0; overridden via config.platform, actual: 8.1.9) does not satisfy that requirement.
    - laminas/laminas-console[2.7.0, ..., 2.8.0] require php ^5.6 || ^7.0 -> your php version (8.0; overridden via config.platform, actual: 8.1.9) does not satisfy that requirement.
    - laminas/laminas-mvc-console 1.3.0 requires laminas/laminas-console ^2.6 -> satisfiable by laminas/laminas-console[2.6.0, 2.7.0, 2.8.0].
    - Root composer.json requires laminas/laminas-mvc-console 1.3.0 -> satisfiable by laminas/laminas-mvc-console[1.3.0].
```

PR #5763, opened 2022-09-18
update mfa

bradymiller:
update mfa

needed for dropping php 7.4 (see #5740) in future (also bypassing saving of qr code to drive (even though should just be temporary) and needing the Imagick php extension and removing a stale package)

testing well


stephenwaite:
jumping

PR #5769, opened 2022-09-20
Fixes Left and Right Login Layout Color Not Fully Showing

### LINKED ISSUE #5759: Left and Right Login Layout Color Not Fully Showing
In the centered layout position, everything works fine, but the left and right login color layouts are broken. I am testing this while using a custom site logo and it glitches out. This seems to extend to OpenEMR 7.0.0 but I am not sure if it affects other versions. Definitely something that needs to be fixed.

tywrenn:
Fixes #5759 

Fixes the right and left login layout not fully showing up as explained in the issue.

NOTE: This also needs to be merged in as a patch to version 7.0.0 and version 6.1.0 as it is affected by this bug. Login layouts will look distorted if this fix isn't implemented.


bradymiller:
@tywrenn , thanks and bringing in! I will update the rel-610 and rel-700 branches with this fix so they go out in next patches.

PR #5784, opened 2022-09-27
Fix for graphing of LBF data entered retroactively.

### LINKED ISSUE #5783: LBF graphs using data entry dates instead of visit dates
### Describe the bug

Currently LBF graphing is based on the data entry timestamp, which makes sense when you might have multiple instances of the form in a visit and are measuring something like blood pressure. However it's more common to have just one instance of a form in a visit, and in that case the visit date should be used in case values are entered retroactively.

### To Reproduce

1. Select a patient with a prior day's visit and add a LBF with graphable data.
2. Create a new visit and add another instance of that LBF.
3. Click the Show Graph button in the LBF.
4. Observe that all dates in the graph are today.

### Expected behavior

All dates should be the date of the visit. When the data entry timestamp is the same as the visit date, use the time from that timestamp as well.

I will be submitting a PR for this.

sunsetsystems:
Fixes #5783

As described in the issue.

### DIFF
--- a/library/ajax/graphs.php
+++ b/library/ajax/graphs.php
@@ -74,15 +74,17 @@ function graphsGetValues($name)
         $values = sqlStatement(
             "SELECT " .
             "ld.field_value AS " . add_escape_custom($name) . ", " .
-            "f.date " .
-            "FROM forms AS f, lbf_data AS ld WHERE " .
+            // If data was entered retroactively then cannot use the data entry date.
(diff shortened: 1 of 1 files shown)


sjpadgett (approved):
Looks good to me Rod.

sunsetsystems:
Thanks Jerry!

stephenwaite:
nice

PR #5791, opened 2022-09-28
updated easy dev documentation with video links

bradymiller:
updated easy dev documentation with video links


stephenwaite:
jumping

PR #5792, opened 2022-09-28
Medex9.28.22

ophthal:
Remove a MedEx Preference.
Reset background_services on Preference change.

### DIFF
--- a/interface/main/messages/save.php
+++ b/interface/main/messages/save.php
@@ -45,7 +45,7 @@ if ($_REQUEST['go'] == 'sms_search') {
 if ($_REQUEST['go'] == 'Preferences') {
     if (AclMain::aclCheckCore('admin', 'super')) {
         $sql = "UPDATE `medex_prefs` SET `ME_facilities`=?,`ME_providers`=?,`ME_hipaa_default_override`=?,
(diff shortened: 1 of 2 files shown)


stephenwaite:
hi @ophthal , is this ready to bring in? thanks

ophthal:
Only Ralf is using the latest openemr version and I haven’t heard from him in a while, so the answer must be yes for MedEx.

However, a cursory look at the demo and I see lots of formatting errors associated with the Recall Board due to MedEx for non MedEx users.  Haven’t looked at the Flow Board.  Eye Form is loaded with new errors too.  

I think it may be time for me to submit PRs to remove MedEx from parts of OpenEMR like the Recall Board - very few MedEx people use the Recall Board…

stephenwaite:
Ok, thanks @ophthal. We'll bring this in and get into the next patch too.

We've been fixing up php8 errors along the way and hopefully that's all there is to address in those places.

ophthal:
Maybe.  The advanced features can only really be appreciated when used in production over time - it continually adapts for each provider as data is entered and also for individual patient charts as new visits are added, so I understand how they can be missed.  It is not a simple form to debug since it wasn’t written by a programmer ;)

PR #5806, opened 2022-10-03
Questionnaire feature port to v7 patch

sjpadgett:
- portal source
- questionnaire_assessments encounter form
- services
- moded forms.php
- UuidRegistry

Fixes #


sjpadgett:
working on database patch

adunsulag:
I've been looking over the code. I'm going to try and pull this down to look at things as it will make more sense to me with all of the moving pieces.  Gotta get to a clean state with what I am doing with the calendar stuff as everythings broken right now lol.

adunsulag:
Huge amount of work and effort here though @sjpadgett, good job!

sjpadgett:
Suddenly see an issue that new questionnaires from LOINC(PHQ9) are not calculating after converting LH-form to questionnaire R4! 
And thanks for spending time looking over. I do know I need to do some logging however the exceptions pattern is working well.
We have issues with Exceptions e.g CCM where exceptions aren't percolating to somewhere that gets to UI. i.e The exception is caught however next level up is also looking for exception and never sees the original triggered. Hope that makes sense.

sjpadgett:
New questionnaire import utility that renders, converts lforms to questionnaire and visa versa.
Also, new manual mode to paste and create questionnaires.

image

And whenever a json form(lform or questionnaire) is selected for upload util pops up where it renders form that was imported and converts to othe formats. Then user can test both formats from conversions.

image

sjpadgett:
Going forward it is safe to cherry pick these scripts from master for any changes.

PR #5562, opened 2022-06-27
final version for new form phq9

ruthkonyn:
also Fixes #5310 

#### Short description of what this resolves:
formatting and numbering the questions

#### Changes proposed in this pull request:
contribute a new form for PHQ9 questions.

PR #5810, opened 2022-10-04
Phq9form final version

### LINKED ISSUE #5809: PHQ9 form
Have a PHQ-9 Patient Depression Questionnaire .

ruthkonyn:
Fixes  #5809
also Fixes https://github.com/openemr/openemr/issues/5310

makes available a  PHQ-9 patient depression questionnaire 

work had been being done in PR #5562 - but this had been from a master clone. We've moved it into a branch from which this new PR has been made.


interface/forms/phq9/phq9_javasrc.js
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// function update_score - display new total score - check if question 8 should be displayed
+// @param int index  question being answered, is 'undef' if we simply want to display the score, e.g. on startup
+// @param int new_score is 'undef' if it's from clicking 'please select an answer' in a new form - treat as zero.
+// @return true|false
+function update_score(index, new_score)
+{
+  //index is the number of the question, score it's score
+    var score = new_score;
```
stephenwaite, line 76: can you make this
```
    var score = parseInt(new_score);
```
and then make the suggested changes to use `js_escape()` instead of `attr()` to keep the use of this function consistent?

ruthkonyn: hi @stephenwaite - i'm afraid the use of parseInt at this point does not fix the problem introduced by using js_escape. If i use parseInt when i obtain the value, i.e. where i've just used js_escape, then I get a numeric value.This works as the numeric score (the sum of the individual scores for each question) is calculated by adding the number obtained each time. by the time update_score is called if new_score is already a string of digits then  parseInt is simply converting the whole of this string into a, rather large, integer. 

e.g. if new_score is '134' instead of 8, then  parseInt   returns the integer 134 - a score off the scale (which is why the explanation is ' Severe depressive disorder' - the highest it can be.

 It would help me enormously if I understood why it's important to use js_escape(), rather than, say, text() -  can you explain or point me at documentation,  

meanwhile, if attr() is depricated here, it seems the choice is 
 var i = <?php echo js_escape($obj['interest_score']); ?>; //the value from last time
                            phq9_score += parseInt(i);
or
 var i = <?php echo text($obj['interest_score']); ?>; //the value from last time
                            phq9_score += i;
which is more straight forward.  
both work correctly. 

The alternative would be to rewrite the logic so it copes with strings of numbers - but that is much more work
thanks

stephenwaite: hi @ruthkonyn , [escaping functions](https://github.com/openemr/openemr/blob/master/library/htmlspecialchars.inc.php)

ruthkonyn: thanks @stephenwaite - i had already found this file. what i don't understand is why i would want to get a json string rather than an integer straight off, which is what i need. What is the problem with my using text() at this point? If it's important i'm happy to change all these instances, as indicated in the first code snippet above, i'm just very curious as to why, so i can write better code in the future. Put it down to my inexperience with writing web applications... thanks

stephenwaite: looks like we'll need a new function like
```
/**
 * Escape a javascript literal for maths.
 */
function js_escape_num($text)
{
    return json_encode($text, JSON_NUMERIC_CHECK);
}

Comment (sjpadgett) [reply interface/forms/phq9/phq9_javasrc.js:76]: > function js_escape_num($text)
> {
>     return json_encode($text, JSON_NUMERIC_CHECK);
> }

Wait! Why? 
From what I see all values being scored are strings so json_encode is doing its job. Otherwise if a score was truly an integer then json_encode would handle and pass the integer.
The problem I see is you're trying to do math with an intermix of string and integers in javascript where javascript would determine to use string concatenation based on if any argument in expression is a string.

You need to do the correct math in your scoring function as shouldn't need new escaping routines

Comment (ruthkonyn) [reply interface/forms/phq9/phq9_javasrc.js:76]: hi @sjpadgett - my original code used a series of lines like 
 ```
 var i = <?php echo text($obj['interest_score']); ?>; //the value from last time
                            phq9_score +=  i;
                            
```
to build a score from a number of answers on a form. It was suggested to use attr() rather than text, then js_escape() - conversations with stephen and brady above and on earlier pr at #5562. then stephen suggested the solution above, which works nicely. 

yes, using js_escape generates string concatanation, whereas text() or attr() seem to generate integer values.

the form answers are strings, but they are associated with numeric values. 

i guess i can accept a concatenation of numbers as strings and then write code to deal with this in the scoring function - it's just a bit more work. 

please could you agree with @stephenwaite and @bradymiller what you'd like me to do, and let me know, then we can get this form submitted, many thanks

library/htmlspecialchars.inc.php
```diff
@@ -20,6 +20,15 @@ function js_escape($text)
     return json_encode($text);
 }
 
+/**
+ * Escape a javascript literal for maths.
+ */
+function js_escape_num($text)
+{
+    return json_encode($text, JSON_NUMERIC_CHECK);
```
sjpadgett, line 28: This needs to be used carefully. It may work in this case but create hidden issues in others.
If json_encode is passed a true integer then it should encode as an integer.
However, each version of PHP seems to be unpredictable with issues like this.
We ran into this encode issue awhile back I think PHP 5.6 or 7 and determined the encoding was working correctly.
Here I think we believe an integer is being passed but it's a string in reality.

I guess I don't have an issue with this solution but I'm wary it will get misused. You might check out some of the issues this can cause.

stephenwaite: thanks @sjpadgett for weighing in, sounds like it's not too much trouble to use `parseInt()` when @ruthkonyn is calculating the score and revert to `js_escape()`

sjpadgett: If Brady is okay with this solution then I'm as well. I am not the best at keeping up with escaping.
Personally I never trust passed in values in javascript and handle accordingly. This goes for Booleans and math.
I tend to test my variables and cast depending on their use case. In the case of javascript I think in terms of strict typing just to stay out of trouble. In terms of PR reviews, seems I can't stay out of trouble!:)

I though the parseInt() would have been the logical solution. But, this works too.

ruthkonyn: thanks you @sjpadgett  - i'd rather not introduce a potential misuse issue - so happy to find another solution. 
what are your thoughts on my using the following to pick up the integer value of the score? 

 $obj = formFetch("form_phq9", $_GET["id"]);

```
  var i = <?php echo text($obj['interest_score']); ?>; //the value from last time
                           phq9_score +=  i;
```

ruthkonyn: thanks all, i'll go with @stephenwaite's js_escape/parseInt solution, i've already tried it out and it does work. I'll resubmit tomorrow. 
thanks for weighing in  @sjpadgett !!- i have found the discussion  very useful.

interface/forms/phq9/new.php
```diff
@@ -0,0 +1,18 @@
...
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Ruth Moulton <moulton ruth@muswell.me.uk>
+ * @author    Jerry Padgett <sjpadgett@gmail.com>
+ * @copyright Copyright (c) 2021 ruth moulton <ruth@muswell.me.uk>
+ *
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once(__DIR__ . "/../../globals.php");
```
bradymiller, line 15: remove above line

interface/forms/phq9/common.php
```diff
@@ -0,0 +1,341 @@
...
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Ruth Moulton <moulton ruth@muswell.me.uk>
+ * @copyright Copyright (c) 2021 ruth moulton <ruth@muswell.me.uk>
+ *
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once("phq9.inc.php"); //common strings, require_once(globals.php), other includes  etc
+require_once("../../globals.php");
```
bradymiller, line 15: put the require_once globals.php line above the require once phq9.inc.php line

interface/forms/phq9/common.php
```diff
@@ -0,0 +1,341 @@
...
+require_once("$srcdir/api.inc");
```
bradymiller, line 16: note this will be changing to following sometime today (.php added to api.inc) :)
```
require_once("$srcdir/api.inc.php");
```

interface/forms/phq9/report.php
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Ruth Moulton <moulton ruth@muswell.me.uk>
+ * @copyright Copyright (c) 2021 ruth moulton <ruth@muswell.me.uk>
+ *
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once(dirname(__FILE__) . '/../../../library/api.inc');
```
bradymiller, line 15: another change that will be happening today :) (adding the .php to end of api.inc)
```
require_once(dirname(__FILE__) . '/../../../library/api.inc.php');
```

interface/forms/phq9/report.php
```diff
@@ -0,0 +1,80 @@
...
+
+    if ($data) {
+        print "<table><tr>";
+        foreach ($data as $key => $value) {
+// include scores_array and total for backward compatibility
+            if ($key == "id" || $key == "pid" || $key == "user" || $key == "groupname" || $key == "authorized" || $key ==  "activity" || $key == "date" || $value == "" || $key == "scores_array" || $key == "total" || $value == "0000-00-00 00:00:00") {
+                continue;
+            }
+            if ($key == "difficulty") {
+                print "<td><span class=bold>" . text($str_issues[$key]) . ": </span><span class=text>" . text($str_difficulty_values [$value]) . "</span></td>";
```
bradymiller, line 42: remove the whitespace in `$str_difficulty_values [$value]` (should be `$str_difficulty_values[$value]`)

interface/forms/phq9/save.php
```diff
@@ -0,0 +1,71 @@
...
+ * @link      http://www.open-emr.org
+ * @author    Ruth Moulton <moulton ruth@muswell.me.uk>
+ * @copyright Copyright (c) 2021 ruth moulton <ruth@muswell.me.uk>
+ *
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once("../../globals.php");
+require_once("$srcdir/api.inc");
+require_once("$srcdir/forms.inc");
```
bradymiller, line 16: change above to (these are changes that will be happening in codebase today):
```
require_once("$srcdir/api.inc.php");
require_once("$srcdir/forms.inc.php");
```

library/classes/InsuranceNumbers.class.php
```diff
@@ -21,12 +21,13 @@ class InsuranceNumbers extends ORDataObject
                                                         "1D" => "Medicaid Provider Number","1G" => "Provider UPIN Number","1H" => "Champus Identification Number","1J" => "Facility ID Number",
```
bradymiller, line 21: if this is not part of the PR topic, then revert this script. easiest way to do this is to simply replace this file from the file in master on github:
wget https://raw.githubusercontent.com/openemr/openemr/master/library/classes/InsuranceNumbers.class.php

stephenwaite:
really nice review, thanks @bradymiller

library/classes/InsuranceNumbers.class.php
ruthkonyn: many thanks brady, i've made the changes and will do the PR very shortly

bradymiller:
hi @ruthkonyn , Looks good and bringing in! thanks for the contribution!

2DV

bradymiller:
(note I reverted the changes in InsuranceNumbers.class.php before bringing this in since that should be in a separate PR)

stephenwaite:
jumping

ruthkonyn:
@bradymiller and @stephenwaite and @sjpadgett - thank you so much for your help  - and brilliant that the form is in!!

PR #5822, opened 2022-10-11
Fix vital display format and CCDA bugs

### LINKED ISSUE #5821: Vitals conversions to display format errors.
### Describe the bug
Using an invalid conversion where rounding after converting to display string.
Needs to change from:
`round(number_format($value * 2.54, 2), 1);`

To:
`number_format(round($value * 2.54, 1), 2);`

In several places in core.

### LINKED ISSUE #5826: CCDA creates an encounter for each encounter diagnosis
### Describe the bug

Thankfully the ONC tests never had more that one diagnosis in an encounter however the issue becomes glaring when importing.

### To Reproduce
Generate a CCD with multiple encounter diagnosis/issue/problems. Results in an encounter for each problem.

### Expected behavior
Each problem should be wrapped in an activity template under an encounter template.
So:
```
<enc>Encounter 1
   <act>Problem 1</act>
   <act>Problem 2</act>
</enc>
```

### Client configuration

- Browser:
- OpenEMR version: v7 and master

### LINKED ISSUE #5835: CDA encounter types not being coded in template.
### Describe the bug
Currently we're using a default code for all encounters where the display text tells the encounter type.  This is because the code for category is unreliable.
Thankfully @adunsulag added the encounter type to the form.
This requires rework for CCDA build of encounters.

Since I'm bringing the vitals PR back to v7 I'm just going to add the CCDA work to it.

### LINKED ISSUE #5848: Add import/export of all patient documents via CCDA.
I found a reference template with an external document template I may be able to use.
However I can't find a Section/Component to wrap it.
Most examples I've found use the Problem list section to add an external document for a problem reference.

For me the sensible section would be the Clinical Notes section of a CCD.
So I'll add there and run the document through ONC test jig and see if it complains. If not great otherwise, I'll add a flag for openemr to openemr import/export and allow user to decide.

- How it will work. 

I'll add the ref template for each of the patient documents. The template will have the documents GUID and the patients UUID which imports as the patient external Id. Then will copy and zip up the exported documents to attach somehow to the new exported CCD's zip file. Or maybe a better idea is to use our document API and move them in that fashion.

Hope this makes sense and recm's are welcomed as I may be missing a better way.

sjpadgett:
- round before formatting
- several CCDA enhancements and fixes

Fixes #5821
Fixes #5835 
Fixes #5848 
Fixes #5826 

Display errors


sjpadgett:
This should fix vitals once and for all. Testing well for me...

interface/modules/zend_modules/module/Carecoordination/view/carecoordination/carecoordination/upload.phtml
```diff
@@ -262,6 +262,13 @@ echo $this->headScript()->prependFile($this->basePath() . '/js/autosuggest/autos
     async function handleFileUpload(files, obj) {
         let fnames = '';
         let uploadURL = "<?php echo $this->basePath();?>/carecoordination/upload";
+        // flash to let user know we're working.
+        let f = document.getElementById('uploadFlash');
+        f.innerText = <?php echo js_escape(xlt("Starting Upload Process.")); ?>;
```
bradymiller, line 267: would do (xlj for escaping in javascript context and then the jsText to escape since it is going to be printed into the html context):
```
f.innerText = jsText(<?php echo xlj("Starting Upload Process."); ?>);
```

sjpadgett: can't use jsText or would have. not in scope.

adunsulag: Since its in javascript land can't you just use 
```
top.xlj("Starting Upload Process.");
```
Since the translation functions are in javascript?

adunsulag: I think your right about the header class issue. I remember having that problem before, attempted to fix it and walked angrily away after reverting my code.

sjpadgett:
ho hum!

Issue #5834, opened 2022-10-12
Handle scopes for DELETE request apis in standard api


adunsulag:
Currently our capability statement and scope validation is not handling any of our DELETE api routes for scopes.  We need to make sure we open the <resource>.write scope if there is a DELETE api route.

Issue #5833, opened 2022-10-12
Patient portal api fails unless fhir scope is requested


adunsulag:
When attempting to request scopes for just api:port and any patient portal related scopes the request fails.  In order for the standard api scopes to be added into our oauth2 request the system scope repository requires the api:fhir scope to be present which is incorrect.  We need to make it so standard and portal scopes can stand alone independent of the fhir scopes requested.

Issue #5831, opened 2022-10-12
FHIR Capability Statement Smart capabilities update passthrough contexts


adunsulag:
ONC Inferno is now failing our smart capabilities document which is bizarre and requiring us to use different names for the context-banner and context-style values.  In the .well-known smart document it will use the original values and in the Capability documents these must be called context-passthrough-banner and context-passthrough-style.  Very annoying and not sure when this changed, but latest Inferno ONC is requiring it.

Issue #5830, opened 2022-10-12
FHIR Api add document title to DocumentReference endpoint


adunsulag:
Currently the document title is not available via the FHIR api when hitting the DocumentReference resource endpoint.  We need to expose that property in the api via the attachment.title property.

PR #5836, opened 2022-10-12
Openemr inferno fixes for #5827 #5828 #5829 #5830 #5831 #5833 #5834

adunsulag:
Fixes https://github.com/openemr/openemr/issues/5831 - capabilities statement add passthrough statement
Fixes https://github.com/openemr/openemr/issues/5830 - add document title to DocumentReference endpoint
Fixes https://github.com/openemr/openemr/issues/5829 - enable patients to download their own documents.
Fixes https://github.com/openemr/openemr/issues/5828 - OperationDefinition endpoint, $bulkdata-status updates
Fixes https://github.com/openemr/openemr/issues/5827 - Migrate FHIR doc binary download to root Binary endpoint
Fixes https://github.com/openemr/openemr/issues/5833 - Patient portal scopes independent of fhir scopes
Fixes https://github.com/openemr/openemr/issues/5834 - Add delete interaction in capability statement to handle
write scopes correctly.


library/classes/Document.class.php
```diff
@@ -283,6 +283,16 @@ public function has_expired()
         return false;
     }
 
+    public function can_patient_access($pid)
```
adunsulag, line 286: @bradymiller If you can look at this from a security point of view and let me know if you see any problems.  Its called in the FhirDocumentRestController.

bradymiller: would just make sure $pid is also not empty in the if (overkill but might as well). otherwise looks good.

src/Common/Auth/OpenIDConnect/Repositories/ScopeRepository.php
```diff
@@ -768,19 +774,21 @@ public function getCurrentStandardScopes(): array
...
                 }
             }
         }
+        $oidc = array_combine($this->oidcScopes(), $this->oidcScopes());
         $scopes_api = array_merge($scopes_api, $scopes_api_portal);
 
         $scopesSupported = $this->apiScopes();
         $scopes_dict = array_combine($scopesSupported, $scopesSupported);
-        $scopesSupported = null;
+        $scopesSupported = null; // this is odd, why do we have this?
```
bradymiller, line 791: maybe some sort of loop is happening here where scopesSupported is used to build the $scopes_dict

bradymiller (approved):

bradymiller:
2DV-2

library/classes/Document.class.php
adunsulag: @bradymiller Maybe I'm tired but I believe the first check for empty on the foreignID makes sure the $pid is also never empty since foreignID has to equal the $pid.

bradymiller: that is true. is overkill, on security related stuff, I overuse !empty() and am explicit so don't need to think  :)
can leave as is.

PR #5854, opened 2022-10-18
Fixes #5838 encounter resource identifier change

### LINKED ISSUE #5838: ONC Inferno  V.0.3.12 fails Encounter resource system property
The encounter system property which was set to the URI for RFC 3986 which includes the uuid classification is failing in the inferno validator.  I'm going to switch the system to be a urn:uri:<uuid> property which is an example given in the URI definition for FHIR.  This appears on preliminary testing to pass the new version of inferno.

adunsulag:
Changed the identifier to be a uuid and removed the RFC 3986 reference as that is failing the updated inferno test suite.


bradymiller (approved):

bradymiller:
2DV

PR #5863, opened 2022-10-21
Openemr rel700 5827 5828 5830 5831 5833 5834 5854 5832 5853 5858,

### LINKED ISSUE #5827: Migrate fhir Document Binaries endpoint to new FHIR endpoint
ONC Inferno is now filing the /fhir//document/<uuid>/Binary endpoint that we were using for our bulk-data exports as well as for our DocumentReference endpoints.  Looking at the FHIR 3.1.1 documentation it no longer shows a /document/<uuid>/Binary endpoint and only lists a /Binary endpoint which is odd.  That said, we need to migrate the endpoint so we can pass this off with ONC's newest Inferno testing suite for ongoing compliance with ONC certification.

### LINKED ISSUE #5828: Add FHIR OperationDefinition endpoints to support $bulkdata-status endpoint
Inferno's latest ONC test suite is failing our $bulkdata-status custom operation definition.  We need to add a /OperationDefinition endpoint to FHIR that will hold our custom server operations.  We then need to add the $bulkdata-status as an OperationDefinition in our FHIR capability statement.

### LINKED ISSUE #5830: FHIR Api add document title to DocumentReference endpoint

### LINKED ISSUE #5831: FHIR Capability Statement Smart capabilities update passthrough contexts

### LINKED ISSUE #5832: FHIR api expose document categories via DocumentReference endpoint

### LINKED ISSUE #5833: Patient portal api fails unless fhir scope is requested

### LINKED ISSUE #5834: Handle scopes for DELETE request apis in standard api

### LINKED ISSUE #5853: Expose OpenEMR document categories heirarchy via FHIR api

### LINKED ISSUE #5858: DocumentReference failing on date search with inferno test suite
Inferno is failing the document reference endpoint on date searches because the Clinical Notes date field is only year-month-day whereas the fhir is treating it as a date time field and the search is failing.  We need to change the datatype on the fhir search field to be a date search and not a datetime.

adunsulag:
Bringing back the following issue fixes into Rel-700

Fixes #5827
Fixes #5828
Fixes #5830
Fixes #5831
Fixes #5833
Fixes #5834
Fixes #5854
Fixes #5832
Fixes #5853
Fixes #5858


library/classes/Document.class.php
```diff
@@ -283,6 +283,16 @@ public function has_expired()
         return false;
     }
 
+    public function can_patient_access($pid)
```
bradymiller, line 286: will be good to get this PR asap since I just realized i just brought code into rel-700 that uses this function :)

bradymiller (approved):

PR #5871, opened 2022-10-23
bug fixes

bradymiller:
Ready to go after a little more testing.

### DIFF
--- a/controllers/C_Document.class.php
+++ b/controllers/C_Document.class.php
@@ -16,6 +16,7 @@ require_once(__DIR__ . "/../library/patient.inc");
 use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\Common\Crypto\CryptoGen;
 use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Logging\SystemLogger;
 use OpenEMR\Common\Twig\TwigContainer;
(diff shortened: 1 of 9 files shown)


library/classes/Document.class.php
```diff
@@ -264,7 +264,7 @@ public function get_categories()
         . "WHERE `ctd`.`document_id` = ? ";
         $resultSet = sqlStatement($categories, [$this->get_id()]);
         $categories = [];
-        while ($category = sqlGetAssoc($resultSet)) {
+        while ($category = sqlFetchArray($resultSet)) {
```
bradymiller, line 267: @adunsulag ,
Note this may impact your code that checks for access. For some reason the sqlGetAssoc call was always empty, which means that the  can_access call was always true no  matter what (since that function returns true when categories is empty). This change fixes the can_access call to work since the categories is now populated.

controllers/C_Document.class.php
```diff
@@ -202,6 +202,16 @@ public function upload_action_process()
 
         if (is_numeric($_POST['category_id'])) {
             $category_id = $_POST['category_id'];
+        } else {
+            $category_id = 1;
+        }
+
+        // ensure user has access to the category that is being uploaded to
```
bradymiller, line 209: likely need logic to skip this when services (ie. Direct) are uploading through this function

PR #5873, opened 2022-10-24
fix for  insertEncounter, always creates an encounter with the curren…

### LINKED ISSUE #5872: Encounter Post API - Date is saving as current date instead of data passed]
### Describe the bug

I have been working with the API for Encounter, the date always defaults to today instead of the one passed on the POST API. please help

For example:
{
“date”: “2020-11-10”,
“onset_date”: “”,
“reason”: “Pregnancy Test”,
“facility”: “Owerri General Hospital”,
“pc_catid”: “5”,
“facility_id”: “3”,
“billing_facility”: “3”,
“sensitivity”: “normal”,
“referral_source”: “”,
“pos_code”: “0”,
“external_id”: “”,
“provider_id”: “1”,
“class_code”: “AMB”
}

the date that is being saved to the db is the current date instead of “date”: “2020-11-10”.

### To Reproduce
more info here https://community.open-emr.org/t/encounter-post-api-date-is-saving-as-current-date-instead-of-data-passed/18550

### Client configuration

OpenEMR Version 
I’m using OpenEMR version 7.0

Browser:
I’m using: brave/chrome

Operating System
I’m using: WINDOWS

reymondko:
Issue:[Encounter Post API - Date is saving as current date instead of data passed](https://community.open-emr.org/t/encounter-post-api-date-is-saving-as-current-date-instead-of-data-passed/18550)

Fixes #5872

this will replace the current date with the date that is being transferred via API

### DIFF
--- a/src/Services/EncounterService.php
+++ b/src/Services/EncounterService.php
@@ -307,7 +307,7 @@ class EncounterService extends BaseService
         $encounter = generate_id();
         $data['encounter'] = $encounter;
         $data['uuid'] = UuidRegistry::getRegistryForTable(self::ENCOUNTER_TABLE)->createUuid();
-        $data['date'] = date("Y-m-d");
+        // $data['date'] = date("Y-m-d");
         $puuidBytes = UuidRegistry::uuidToBytes($puuid);
         $data['pid'] = $this->getIdByUuid($puuidBytes, self::PATIENT_TABLE, "pid");
         $query = $this->buildInsertColumns($data);


src/Services/EncounterService.php
```diff
@@ -307,7 +307,7 @@ public function insertEncounter($puuid, $data)
         $encounter = generate_id();
         $data['encounter'] = $encounter;
         $data['uuid'] = UuidRegistry::getRegistryForTable(self::ENCOUNTER_TABLE)->createUuid();
-        $data['date'] = date("Y-m-d");
+        // $data['date'] = date("Y-m-d");
```
adunsulag, line 310: I recommend changing this to set the date to be the current date IF $data['date'] is empty.  That will make sure we maintain at least some backwards compatibility with API callers who have not been sending the encounter date.

If you add in that fix then the AppointmentService will still work as this insertEncounter function is called there.
https://github.com/openemr/openemr/blob/46c3f55c8dfe860106ca416f1cc3c9089e88e897/src/Services/AppointmentService.php#L388-L405

adunsulag (changes requested):
Thanks for the pull request.  Just a minor fix is needed to maintain backwards compatability.

PR #5879, opened 2022-10-26
bug fixes

bradymiller:
bug fixes

### DIFF
--- a/interface/forms/LBF/new.php
+++ b/interface/forms/LBF/new.php
@@ -201,7 +201,7 @@ if (isset($LBF_SERVICES_SECTION) || isset($LBF_PRODUCTS_SECTION) || isset($LBF_D
 }
 
 if (!$from_trend_form) {
-    $fname = $GLOBALS['OE_SITE_DIR'] . "/LBF/$formname.plugin.php";
+    $fname = $GLOBALS['OE_SITE_DIR'] . "/LBF/" . check_file_dir_name($formname) . ".plugin.php";
     if (file_exists($fname)) {
         include_once($fname);
     }
--- a/setup.php
+++ b/setup.php
@@ -4,6 +4,18 @@
  *
  * Installation script.
  *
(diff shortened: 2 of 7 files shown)


interface/forms/LBF/new.php
```diff
@@ -201,7 +201,7 @@ function end_row()
 }
 
 if (!$from_trend_form) {
-    $fname = $GLOBALS['OE_SITE_DIR'] . "/LBF/$formname.plugin.php";
+    $fname = $GLOBALS['OE_SITE_DIR'] . "/LBF/" . check_file_dir_name($formname) . ".plugin.php";
```
stephenwaite, line 204: jarosiewicz-mr-bean

stephenwaite (commented):
simpsons-cheering

a lot of fixes, nice!

setup.php
```diff
@@ -244,12 +251,76 @@ function recursive_writable_directory_test($dir)
...
+    SessionUtil::setupScriptSessionStart();
+    session_regenerate_id(true);
+    $_SESSION = [];
+    CsrfUtils::setupCsrfKey();
+} else {
+    // start up session and check csrf
+    SessionUtil::setupScriptSessionStart();
+    $state = (int)$state;
+    $verifyCsrf = false;
+    if (($state > 0) && ($state < 8)) {
```
adunsulag, line 287: Anyway we can put these numbers as constants somewhere?  Makes it hard to follow the logic of what is going on with 0, 8, 3, etc.

bradymiller: i could, but then I'd begin to consider Seppuku :)
i'll do the next best thing, though, and put some comments about what each one is

adunsulag: lol, well let's not get hasty haha.

bradymiller: luckily no sword handy, but there is always amazon prime :)

bradymiller: added more description of the state(s) to the script header comments

bradymiller:
testing well and bringing in

PR #5891, opened 2022-10-28
Fixes #5708 Add Appointment FHIR,Portal endpoints (#5709)

### LINKED ISSUE #5708: Add endpoints for Appointments to FHIR and patient portal apis

adunsulag:
Backports fix in master for #5708 to rel-700.


adunsulag:
If this looks good I'll bring this in as it's blocking my other PR.

version.php
```diff
@@ -28,7 +28,9 @@
 // is a database change in the course of development.  It is used
 // internally to determine when a database upgrade is needed.
 //
-$v_database = 470;
+
+$v_database = 472;
```
sjpadgett, line 32: Are you skipping because of my PR? 471

sjpadgett (approved):
looks okay to me. Ya been a busy bee my friend!

version.php
sjpadgett: Don't worry about my backport as I can fix the conflicts from this.

sql/database.sql
```diff
@@ -7310,9 +7310,11 @@ CREATE TABLE `openemr_postcalendar_events` (
   `pc_sendalertemail` VARCHAR( 3 ) NOT NULL DEFAULT 'NO',
   `pc_billing_location` SMALLINT (6) NOT NULL DEFAULT '0',
   `pc_room` varchar(20) NOT NULL DEFAULT '',
+  `uuid` binary(16) DEFAULT NULL,
   PRIMARY KEY  (`pc_eid`),
   KEY `basic_event` (`pc_catid`,`pc_aid`,`pc_eventDate`,`pc_endDate`,`pc_eventstatus`,`pc_sharing`,`pc_topic`),
-  KEY `pc_eventDate` (`pc_eventDate`)
+  KEY `pc_eventDate` (`pc_eventDate`),
+  UNIQUE KEY `uuid` (`uuid`)
```
bradymiller, line 7317: also needs corresponding sql upgrading elements to sql/patch.php

version.php
bradymiller: @sjpadgett , @adunsulag 
note no need to increment this in the patching branches. it's ok to do it, but not used, so also ok not to do it :)

adunsulag: Had a conflict so I just bumped it to the higher version of the conflict.  Sounds like it doesn't really matter.

sql/database.sql
adunsulag: I added the database changes to patch.php.

bradymiller (approved):

PR #5898, opened 2022-11-01
minor adjustments

bradymiller:
@sjpadgett , also gotta remember to bring this into rel-700 after your ccda fixes go there

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncountermanagerTable.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncountermanagerTable.php
@@ -339,7 +339,7 @@ class EncountermanagerTable extends AbstractTableGateway
         $query = "SELECT cc.id, pd.fname, pd.lname, pd.pid FROM ccda AS cc
 		    LEFT JOIN patient_data AS pd ON pd.pid=cc.pid
 		    WHERE cc.pid = ?
-		    ORDER BY cc.id DESC LIMIT " . escape_limit($limit);
+		    ORDER BY cc.id DESC LIMIT $limit";
         $res = $appTable->zQuery($query, array($pid));
         foreach ($res as $row) {
             $res_cur[] = $row;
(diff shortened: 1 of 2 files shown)


bradymiller:
yep, you are correct, the limit was escaped in several lines above. will revert that fix. my eagle eyes are dimming :)

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncountermanagerTable.php
```diff
@@ -339,7 +339,7 @@ public function getFileID($pid, $limit = 1)
         $query = "SELECT cc.id, pd.fname, pd.lname, pd.pid FROM ccda AS cc
 		    LEFT JOIN patient_data AS pd ON pd.pid=cc.pid
 		    WHERE cc.pid = ?
-		    ORDER BY cc.id DESC LIMIT $limit";
+		    ORDER BY cc.id DESC LIMIT " . escape_limit($limit);
```
sjpadgett: escaped above

bradymiller: yep, double escaped :)
maybe should do following to really be sure :)
```
ORDER BY cc.id DESC LIMIT ";
for ($i=0; $i<1000000000; $i++) {
    $limit = escape_limit($limit);
}
$query .= $limit;
```

sjpadgett: I tend to try to set variables before using.

src/Services/Cda/CdaTemplateImportDispose.php
```diff
@@ -2033,14 +2033,13 @@ public function findClosestEncounter($item_date, $item_pid)
      */
     public function findClosestEncounterWithForm($item_date, $item_pid, $form = '')
     {
-        $form = add_escape_custom($form);
```
bradymiller, line 2036: @sjpadgett , let me know if you want me to change above to (would keep with your style of escaping variables prior):
```
$form = escape_table_name($form);
```
(and then I would remove the line below)

sjpadgett: No you're fine.

PR #5902, opened 2022-11-02
By default, clear the existing CAMOS table data before trying to create a new table (#5901)

### LINKED ISSUE #5901: Enabling the CAMOS form is slightly broken
### Describe the bug

*(As a quick note to start, this work is being driven by my employer, [Altispeed Technologies](https://altispeed.com/). I'm actively working on this issue. Any screenshots shown use test data, and are not associated with any real patient.)*

### To Reproduce

 1. Install the CAMOS form using the expected workflow
 2. Import a database dump from another EMR instance that does not have CAMOS enabled or installed in the DB
 3. Repeat step 1

image

### Client configuration

- Browser: Firefox Nightly
- OpenEMR version: tip of master
- Operating system: Ubuntu Devel (currently 23.04)

tsimonq2:
Fixes #5901

This fixes the issue I filed, but does not strike me as scalable.

When a user is enabling the CAMOS module, should it account for existing data in the database? It would otherwise show as "DB installed" if the user used the expected workflow (enable CAMOS before populating the data).

I would be interested in hearing ideas for how this could be implemented in a more scalable way. For now, here's what works on my end.

### DIFF
--- a/interface/forms/CAMOS/table.sql
+++ b/interface/forms/CAMOS/table.sql
@@ -1,4 +1,5 @@
-CREATE TABLE IF NOT EXISTS `form_CAMOS` (
+DROP TABLE IF EXISTS `form_CAMOS`;
+CREATE TABLE `form_CAMOS` (
 id bigint(20) NOT NULL auto_increment,
 date datetime default NULL,
 pid bigint(20) default NULL,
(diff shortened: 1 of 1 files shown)


bradymiller:
this solution would lead to potential data loss. will post some comments on the issue

PR #5904, opened 2022-11-03
ccda import fix

### LINKED ISSUE #5903: CCDA import appears broken
If try to import the 'Virginia Klein' ccda (this worked during saturday's call) in care coordination module in most recent master, and it does not work (just goes back to the upload screen in care coordination module) and showed following javascript message (nothing is showing up in the php error log):
image

Additionally, the easy docker random synthea command line ccda importing is also not working. It runs and looks like it does the ccda imports (but the import is way to fast and is not actually happening). Note this import generally shows lots of php warning while it is working but none are showing up which makes me think that the import is breaking even before it starts the import process.

bradymiller:
fixes #5903 

@sjpadgett , the silent fail was happening at the creation of the instance `$xmltoarray = new XmlExtended();`. No constructor is involved, so figured that the class it extended was bringing in something that XmlExtended didn't. So, if create a dummy Xml class `$dummyXml = new XML();` before creating the XmlExtended() class then it works (gui import of xml and zip with documents and the command line synthea import also works)
:)
I think something that XML brings in at https://github.com/laminas/laminas-config/blob/3.7.0/src/Reader/Xml.php#L5-L21 that is not brought in by XmlExtended is causing the fail, but even if I copy that into the XmlExtended class script, it still fails.

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/CarecoordinationTable.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/CarecoordinationTable.php
@@ -193,6 +193,9 @@ class CarecoordinationTable extends AbstractTableGateway
         // 10/27/2022 sjp Extended base reader Laminas XML class using provided interface class
         // Needed to add LIBXML_COMPACT | LIBXML_PARSEHUGE flags because large text node(>10MB) will fail.
         try {
+            // For unknown reasons, need to create a prior dummy XML() instance for the XMLExtended() instance to work.
+            //  Otherwise it fails silently in linux (notably, this is not needed for windows).
+            $dummyXml = new XML();
             $xmltoarray = new XmlExtended();
             $xml = $xmltoarray->fromString($xml_content_new);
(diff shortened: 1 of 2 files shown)


stephenwaite:
think the issue is the `case`

Screenshot from 2022-11-03 10-36-06

sjpadgett:
Good point @stephenwaite and nice catch

stephenwaite:
pushed to [Brady's branch](https://github.com/bradymiller/openemr/commit/44c9ce6e4a29b096e51d55e3e75f1e6ca6f385f5) but it didn't sync here

stephenwaite:
pushed to [Brady's branch](https://github.com/bradymiller/openemr/commit/44c9ce6e4a29b096e51d55e3e75f1e6ca6f385f5) but it didn't sync here

stephenwaite:
now it shows

yes-gif-5

sjpadgett (approved):
thanks Brady and Stephen

bradymiller:
nice @stephenwaite !

2DV-2

PR #5917, opened 2022-11-05
Additional CCDA import fixes.

### LINKED ISSUE #5916: CCDA import/export needs improvement and bug fixes
### Describe the bug
Using a Synthea patient generated example CCD found that there are differences from ONC test requirement documents and the example CCD. 
Many sections within CCDA spec can be presented in different template types. As example encounters may use an encounter performed template or an encounter activity template.

Another issue is we only export procedures associated with procedures that have results. We should report pending as well.

All these issues only serve to point out a fraction of potential ways CCDA's can be generated/received where we can only try to support the most common types of CCDA documents in use always keeping in mind future ONC testing requirements which for import will be coming soon I expect.

Current import of test document is getting close:

image

### LINKED ISSUE #5921: Procedures not reported in CCDA unless billed
### Describe the bug
As title says, if procedure is not in billing table then currently we exclude all other procedures.
This can not endure....

sjpadgett:
- new procedure parse to capture organizer batches i.e assessments(PHQ-9)
- demo language and ethnicity fixes

Fixes #5916 
Fixes #5921


sjpadgett:
for those keeping up with Virginia.

image

sjpadgett:
Will bring in soon for dev demo reset so can test.

sjpadgett:
merging. testing well

PR #5920, opened 2022-11-05
Fix #5919 datetime timezone.

### LINKED ISSUE #5919: FHIR DateTime search with timezone component fails to adjust to local server time
The DateTime search using a timezone component fails if the server and the requestor are on a different timezone.  We need to make sure we adjust the search queries to use the local server time instead of the requestor timezone.

For example a query such as this: 
```
/apis/default/fhir/DiagnosticReport?category=LAB&date=2015-06-22T04:00:00%2B00:00
```
Will fail to return any data if the local server timezone is set to **America/New York** and the database procedure order dates are saved as **2015-06-22 00:00**.  

We need to parse the timezone information on the incoming request and we need to adjust the timezone at the time we generate the FHIR search queries.

adunsulag:
Handles the timezone requests on the date search field to parse out the requestor's timezone.

Handles the adjustment of the requestor's timezone on the search field statement resolver to be the local time.
Fixes #5919

### DIFF
--- a/src/Services/Search/DateSearchField.php
+++ b/src/Services/Search/DateSearchField.php
@@ -131,8 +131,8 @@ class DateSearchField extends BasicSearchField
             throw new \InvalidArgumentException("Invalid comparator found for value " . $value);
         }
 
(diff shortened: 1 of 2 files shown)


bradymiller (approved):

bradymiller:
tenor99

Issue #5924, opened 2022-11-06
A way to resend batched claims


juggernautsei:
### Describe the problem

Using the SFTP feature the auto sends batches of claims to a clearinghouse. If the batches fail due to login-error. The batches have to be rebatched to put them into waiting status again. Time consuming rework that needs to be and can be avoided. 

### Describe the solution you'd like

The system should auto detect the SFTP connection has been restored and automatically resend claims that are marked as login-error.

PR #5925, opened 2022-11-06
A way to resend batched claims 

juggernautsei:
Fixes #
#5924
Trying to solve the issue of sending claims after a failed login. 

Test the SFTP connection and when the connection is available/good. Change claim status to waiting to be sent.

### DIFF
--- a/src/Billing/X12ClaimRepost.php
+++ b/src/Billing/X12ClaimRepost.php
@@ -19,8 +19,7 @@ class X12ClaimRepost
         $username,
         $password,
         $port = 22
-    )
-    {
+    ){
         $connection = new SFTP($host, $port);
         if (false === $connection->login($username, $password)) {
             return 'failed';
@@ -47,4 +46,3 @@ class X12ClaimRepost
         sqlQuery("UPDATE `x12_remote_tracker` SET status = 'waiting', messages = NULL WHERE `status` = 'login-error'");
     }
 }
-
--- a/src/Billing/X12ClaimRepost.php
+++ b/src/Billing/X12ClaimRepost.php
@@ -44,10 +44,7 @@ class X12ClaimRepost
 
     public static function updateStatus(): void
     {
-        sqlQuery("UPDATE `x12_remote_tracker` SET 
-                                status = 'waiting',
-                                messages = NULL
-                            WHERE `status` = 'login-error'");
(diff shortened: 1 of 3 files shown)


src/Billing/X12ClaimRepost.php
```diff
@@ -0,0 +1,50 @@
...
+        return sqlQuery('SELECT x12_sftp_login FROM x12_partners');
+    }
+    public static function x12Password()
+    {
+        return sqlQuery('SELECT x12_sftp_pass FROM x12_partners');
+    }
+
+    public static function updateStatus(): void
+    {
+        sqlQuery("UPDATE `x12_remote_tracker` SET status = 'waiting', messages = NULL WHERE `status` = 'login-error'");
```
stephenwaite, line 48: hi @juggernautsei, this might be too broad of a query. If there was a way to give a button some capability on each line of the billing tracker to call this repost then it wouldn't risk resending a file with a batch of claims that may have been regenerated from the billing manager and successfully sent in place of the file that failed.

juggernautsei: I understand what you are asking. The thought process behind the build was that this would run automatically and not be a manual process. That is why I put the file in the library to run it in the background manager. I added it to the claim tracker page so that when the claim tracker page is looked at and the connection is restored. The query would reset all the claims that have been marked with the login error (which was our case). The users were not regenerating any of the batches. In fact they did not want to go through all the hassle of regenerating batches because of the failed login. They just wanted the batches to go out once the connection has been restored with no effort on their part. That was the driving force behind this feature.

If a batch is resent, the clearinghouse will just reject the claims. The claim won't be processed again to my knowledge. We have done that. Also, the billing manager does not auto batch so why would there be a duplicate batch that has successfully gone through while the SFTP connect is down?

If someone while the SFTP connection is down manually uploaded a file the clearinghouse and the batch failed. When the connection returns the batch will be the same one that was uploaded manually during the time that the connection was down.  

So, compromise. Make it a global to resend all claims that are marked login-error.

stephenwaite: hi @juggernautsei, since it should be fixed for more than 1 sftp account then having a button on each line would be a good solution

the button could also help fix other errors like bad parameters etc

juggernautsei:
Ok, I will keep working on the feature. 
Is this the vision you have in mind?
image

juggernautsei:
Closing this branch. It is not acceptable as is. Maybe I will circle back in the future.

PR #5959, opened 2022-11-15
fix enter in medical device gui

### LINKED ISSUE #5947: Add Enter button support for the Process UDI when adding implantable device
(part of video tutorial)
Add Enter button support for the Process UDI when adding implantable device (if focused on the udi text field, then enter button should be same as Process UDI button).

Machine-OpenEMR:
fixes #5947


bradymiller:
bringing this in
congrats on first commit @Machine-OpenEMR :)

PR #5984, opened 2022-11-29
Switching from smarty to twig templating

### LINKED ISSUE #5983: Convert from smarty to twig
### Describe the problem

There is no problem that this solves other than to move the codebase forward with migration away from smarty to twig 

The soap note has been converted from smarty as the template engine to twig. It is tested and working.

juggernautsei:
Fixes #5983

moving from smarty to twig 

stop using smarty in the soap note and use twig


interface/forms/soap/C_FormSOAP.class.php
```diff
@@ -1,40 +1,44 @@
...
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
 require_once($GLOBALS['fileroot'] . "/library/forms.inc.php");
 require_once("FormSOAP.class.php");
+require_once("Bootstrap.php");
 
 use OpenEMR\Common\Csrf\CsrfUtils;
 
 class C_FormSOAP extends Controller
```
bradymiller, line 20: Can the `extends Controller` now be remove?

interface/forms/soap/C_FormSOAP.class.php
```diff
@@ -1,40 +1,44 @@
...
+        $this->template = $twig->twigEnv();
     }
 
     function default_action()
     {
-        $form = new FormSOAP();
-        $this->assign("data", $form);
-        return $this->fetch($this->template_dir . $this->template_mod . "_new.html");
+         $form = new FormSOAP();
+        return $this->template->render(
```
bradymiller, line 33: hopefully can do it all with following one liner (or something similar), so then don't need all the bootstrap stuff):
```
        return (new TwigContainer($GLOBALS['fileroot'] . "/interface/forms/soap/templates", $GLOBALS['kernel']))->getTwig()->render(
            'soap_form.twig',
            [
                "FORM_ACTION" => $GLOBALS['web_root'],
                "DONT_SAVE_LINK" => $GLOBALS['form_exit_url'],
                "CSRF_TOKEN_FORM" => CsrfUtils::collectCsrfToken(),
                "data" => $form
            ]
        );
```

then just need the following use at top of this script:
```
```

That 1st parameter in TwigContainer seems a bit odd though since it seems to add the path to an array with the original path, so probably makes sense to ensure the .twig files have more specific names, such as interface/forms/soap/templates/soap_form.twig

interface/forms/soap/C_FormSOAP.class.php
```diff
@@ -1,40 +1,44 @@
...
+            'general_new.twig',
+            [
+                "FORM_ACTION" => $GLOBALS['web_root'],
+                "DONT_SAVE_LINK" => $GLOBALS['form_exit_url'],
+                "CSRF_TOKEN_FORM" => CsrfUtils::collectCsrfToken(),
```
bradymiller, line 38: note there is a twig function to get the csrf token called `csrfTokenRaw`

interface/forms/soap/C_FormSOAP.class.php
```diff
@@ -45,9 +49,15 @@ function view_action($form_id)
...
 
-        $this->assign("data", $form);
-
-        return $this->fetch($this->template_dir . $this->template_mod . "_new.html");
+        return $this->template->render(
+            'general_new.twig',
+            [
+                "FORM_ACTION" => $GLOBALS['web_root'],
+                "DONT_SAVE_LINK" => $GLOBALS['form_exit_url'],
+                "CSRF_TOKEN_FORM" => CsrfUtils::collectCsrfToken(),
```
bradymiller, line 57: note there is a twig function to get the csrf token called `csrfTokenRaw`

interface/forms/soap/templates/general_new.twig
```diff
@@ -1,55 +1,55 @@
...
 <body>
     <div class="container mt-3">
         <div class="row">
             <div class="col-12">
-                <h2>{xlt t='SOAP'}</h2>
-                <form name="soap" method="post" action="{$FORM_ACTION}/interface/forms/soap/save.php" onsubmit="return top.restoreSession()">
-                    <input type="hidden" name="csrf_token_form" value="{$CSRF_TOKEN_FORM|attr}" />
+                <h2>{{'SOAP'|xlt}}</h2>
+                <form name="soap" method="post" action="{{ FORM_ACTION | attr }}/interface/forms/soap/save.php" onsubmit="return top.restoreSession()">
+                    <input type="hidden" name="csrf_token_form" value="{{ CSRF_TOKEN_FORM | attr }}" />
```
bradymiller, line 12: so for csrf could just do:
```
<input type="hidden" name="csrf_token_form" value="{{ csrfTokenRaw()|attr }}" />
```

interface/forms/soap/C_FormSOAP.class.php
juggernautsei: I will find out..

juggernautsei: No, here is my theory on why. The FormSOAP.class.php file still depends on the parent::populate(). @adunsulag would need to confirm that. I have been far enough down that path of how the populate works. It reads the table columns. Using that list of columns it auto matches the column to the getter and setter method. Depending on the mod as the trigger. Until the dependency on the populate method. The extends controller has to stay.

interface/forms/soap/C_FormSOAP.class.php
juggernautsei: I agree with you but you have to remember that the C_FormSOAP.class.php file is called by the new.php file. To get away from the bootstrap file, the TwigContainer would have to injected through the new.php file. I chose not to go that route right now.

interface/forms/soap/C_FormSOAP.class.php
adunsulag: What @juggernautsei is saying is correct.  Still have to do the extends here.

juggernautsei:
@adunsulag I was able to disconnect from the controller. I tracked it down and I think it was you that built a bridge between the old and the new. I disconnected and tested. All was working the last time I tested. That is me calling it a bridge. The dependency on the controller has been resolved so that the populate works outside of the controller. I am not saying it right but I hope you follow the code and see.

bradymiller:
hi @juggernautsei , Code looks good. If this is testing well, then can being in. Just let us know that is testing well.

interface/forms/soap/C_FormSOAP.class.php
```diff
@@ -1,40 +1,46 @@
...
+    public function __construct()
     {
-        parent::__construct();
-        $this->template_mod = $template_mod;
-        $this->template_dir = dirname(__FILE__) . "/templates/";
-        $this->assign("FORM_ACTION", $GLOBALS['web_root']);
-        $this->assign("DONT_SAVE_LINK", $GLOBALS['form_exit_url']);
-        $this->assign("STYLE", $GLOBALS['style']);
-        $this->assign("CSRF_TOKEN_FORM", CsrfUtils::collectCsrfToken());
+        $path = self::getTemplatePath();
```
bradymiller, line 24: actually, just realized that using wrong call here (this is for a static function). Should be:
```
$path = $this->getTemplatePath();
```

interface/forms/soap/C_FormSOAP.class.php
```diff
@@ -65,10 +76,22 @@ function default_action_process()
...
+            );
             $_POST['process'] = "";
         }
-
-        return;
+    }
+    /**
+     * @return string
+     */
+    public function getTemplatePath(): string
```
bradymiller, line 93: change public to private

juggernautsei: @adunsulag in the Controller.class.php file you have this note // TODO: @adunsulag move these into src/
If I can shadow you on making this move. That would help me as a developer.

juggernautsei: @bradymiller I will begin testing today or tomorrow or soon. =^)8

juggernautsei:
Staying alive

adunsulag:
@juggernautsei so is this ready to bring in?

juggernautsei:
It is ready. I just tested it under the current version and it works well.

bradymiller (approved):

bradymiller:
hi @juggernautsei , Code looks good and bringing in. Thanks for the improvements!

PR #5987, opened 2022-11-30
Render event java demographics

### LINKED ISSUE #5986: Injecting JavaScript into the patient demographics page
### Describe the problem

We need a way to inject java scripting into the patient demographics page

Add a render event to the src/Events/PatientDemographics/RenderEvent.php class

juggernautsei:
Fixes #5986

allows the ability to inject java script into the demographics 

adding java event to demographics page

### DIFF
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -1844,6 +1844,12 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
             ];
             checkSkipConditions();
 
+            <?php
+            $GLOBALS["kernel"]
+                ->getEventDispatcher()
+                ->dispatch(new RenderEvent($pid), RenderEvent::EVENT_RENDER_POST_PAGELOAD, 10);
+            ?>
+
             var isPost = <?php echo js_escape($showEligibility ?? false); ?>;
             var listId = '#' + <?php echo js_escape($list_id); ?>;
             $(function() {
@@ -1853,11 +1859,6 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                     $("#eligibility").get(0).scrollIntoView();
                 }
             });
-            <?php
-                $GLOBALS["kernel"]
-                    ->getEventDispatcher()
-                    ->dispatch(RenderEvent::EVENT_RENDER_JAVA, new RenderEvent($pid), 10);
-            ?>
         </script>
 </body>
 
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -1853,6 +1853,11 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                     $("#eligibility").get(0).scrollIntoView();
                 }
             });
+            <?php
+            $GLOBALS["kernel"]
+                ->getEventDispatcher()
+                ->dispatch(RenderEvent::EVENT_RENDER_JAVA, new RenderEvent($pid), 10);
+            ?>
         </script>
 </body>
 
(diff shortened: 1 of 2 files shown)


robertdown:
I'd recommend making the name of this even more abstract, something like ```EVENT_RENDER_POST_RENDER``` and ```patientDemographics.render.post_render``` (I know it sounds redundant, but it's following the previously used logic of ```EVENT_RENDER_``` prefix. Making this a post render event makes it easier over time to understand where in the cycle the event is being dispatched and can be used in unknown future states that may wish to inject non-javascript based code.

It'd be nice to toss in a ```EVENT_RENDER_PRE_RENDER``` event hook in this PR.

But on the whole, this is a good PR.

As an aside, I'd avoid using "java" as a shorthand to describe javascript :)

interface/patient_file/summary/demographics.php
```diff
@@ -1853,6 +1853,11 @@ function setMyPatient() {
                     $("#eligibility").get(0).scrollIntoView();
                 }
             });
+            <?php
+                $GLOBALS["kernel"]
+                    ->getEventDispatcher()
+                    ->dispatch(RenderEvent::EVENT_RENDER_JAVA, new RenderEvent($pid), 10);
```
sjpadgett, line 1859: Method signature has changed.
Should be `dispatch(new RenderEvent($pid), RenderEvent::EVENT_RENDER_JAVA, 10)`

Soon we'll have to refactor all our events as current argument placement is deprecated.
"Calling the "Symfony\Component\EventDispatcher\ " method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead."

juggernautsei:
@adunsulag what you wrote is good to know just to inject java script anywhere. I had a specific need to load the java script after the buttons in the dashboard ribbon has loaded. It is so much easier to just place the java script at the bottom. 
@robertdown the  name is changed to reflect what the dispatch is to be used for so in the future. It can still be clear. @sjpadgett the order has been rearranged to meet the new requirement.

src/Events/PatientDemographics/RenderEvent.php
```diff
@@ -40,7 +40,7 @@ class RenderEvent extends Event
      * It allows event listeners to render additional functionality after the page
      * has loaded.
      */
-    const EVENT_RENDER_JAVA = 'patientDemographics.render.javascript';
+    const EVENT_RENDER_POST_PAGELOAD = 'patientDemographics.render.javascript';
```
robertdown, line 43: Change `patientDemographics.render.javsacript' to `patientDemographics.render.post_pageload`

bradymiller:
2DV

interface/patient_file/summary/demographics.php
```diff
@@ -1844,6 +1844,12 @@ function setMyPatient() {
             ];
             checkSkipConditions();
 
+            <?php
```
robertdown, line 1847: Since the event is now more content agnostic, this event needs to fire at the outermost part of the page DOM. I would move it after the closing `</body>` tag. If it is kept here, a user could use the event to inject CSS or HTML but it would be rendered inside of a `<script>` tag (which starts at line 1840).

Also, line 72 creates `$ed` which is a helper variable, pointing to `$GLOBALS['kernel']->getEventDispatcher()` so you could consolidate this to just be `$ed->dispatch(new RenderEvent($pid), RenderEvent::EVENT_RENDER_POST_PAGELOAD, 10);` though it is not necessary.

robertdown (dismissed):
One last change and we can bring this in

robertdown:
Heads up @juggernautsei I made the last change I requested and pushed to your branch. I can bring this in once it passes tests

juggernautsei:
My question is that moving the dispatch outside of the script tags. Does that mean that script tags will need to be sent from the listener?

robertdown:
Correct, it becomes the responsibility of the listener to setup whatever they want. We don't want to force the event into a javascript tag on the very real chance this event gets used to inject non-javascript

robertdown:
Thanks for the contrib, @juggernautsei

PR #6004, opened 2022-12-10
add yet another mariadb short term release to ci

bradymiller:
add yet another mariadb short term release to ci


bradymiller:
Good news, though, is that after 4 short term releases, the next release (10.11) will be a long term release:
https://mariadb.org/mariadb-10-11-is-lts/

PR #6006, opened 2022-12-11
composer packages update

bradymiller:
composer packages update


bradymiller:
Needed to fork the particle/validator project since there was a fix required to work with PHP 8.2 and it appears that project is stale. I did place a PR in upstream so you never know (if it gets fixed, then can have composer collect again from the upstream project): https://github.com/particle-php/Validator/pull/197

bradymiller:
testing well and bringing in

PR #6011, opened 2022-12-13
update composer particle/validator package to 2.3.5

bradymiller:
update composer particle/validator package to 2.3.5


bradymiller:
upstream package particle/validator package has been fixed/updated, so can go back to pulling in the original package: https://github.com/particle-php/Validator/pull/197

PR #6014, opened 2022-12-13
Custom addon patient nav menu

### LINKED ISSUE #6013: Need a way to add custom navigation to the patient dashboard
### Describe the problem

We want to add a button to the patient navigation bar. 

The idea would be to create a module that injects java script into the demographics page. That would allow us to populate a div with whatever button or custom logic we see fit.

juggernautsei:
Fixes #6013

This solves being able to add custom logic from a module into the patient dashboard. 

A change to the patient navigation ribbon by adding a div on the end that can be populated by a module.


juggernautsei:
@robertdown  pointed out a way to do this on the fly. No need to change the code base.

PR #6016, opened 2022-12-13
Fix #6009 send telehealth user role

### LINKED ISSUE #6009: Telehealth send user role flag in registration request
For the comlink telehealth service we want to send the role of the telehealth user (provider or patient) when they register with the comlink service.

adunsulag:
Send the user role in the telehealth registration request.

Fixes #6009 in release 700.


bradymiller:
@adunsulag , going to bring this in, which will now pass ci since removed php 8.2 checking in rel-700

PR #6019, opened 2022-12-14
fix ci for php 8.2 in rel-700

bradymiller:
fix ci for php 8.2 in rel-700


bradymiller:
@adunsulag , turns out can't support php 8.2 in 7.0.0 (mostly secondary to third party package issues) so removed 8.2 ci testing from rel-700/

adunsulag:
@bradymiller I guess this will need to be something on our radar for the future right due to ONC stuff we can't just bump people up to another release and not worry about this right?

bradymiller:
@adunsulag , not uncommon that this happens when there is a significant gap between an OpenEMR release and a PHP release (PHP releases are yearly and usually in november). I just made it clear on the download page that PHP 8.2 is not supported. This means we should shoot for a 7.0.1 release in a month or so to deal with this. Since will be 7.0.*, will not impact MU, but the quarterly attestation after that will need to list any pertinent changes.

PR #6020, opened 2022-12-14
Update npm packages

bradymiller:
- Was unable to update dwv package and migrate related deprecated i18next-xhr-backend package to i18next-http-backend package (this will require much more time to do)
-  Was unable to update to most recent del package since it now requires pure ESM (this will also require much more time to do)


bradymiller:
testing well and bringing in. will turn the above 2 things I was unable to do into issues.

PR #6026, opened 2022-12-15
Modernizing after care form

### LINKED ISSUE #6025: Update the After Care form
### Describe the problem

The form look is out dated. 

Adding bootstrap to the after care form to modern form look and feel

juggernautsei:
Fixes #6025 

Adding bootstrap to the after care form to modern the look and remove hard coded entries on the form. 

Updated look.

### DIFF
--- a/interface/forms/aftercare_plan/new.php
+++ b/interface/forms/aftercare_plan/new.php
@@ -60,12 +60,8 @@ $obj = $formid ? formFetch("form_aftercare_plan", $formid) : array();
                 <h1 class="forms-title"><?php echo xlt('After Care Planning'); ?></h1>
 
         <div class="col-md-12 mt-4">
-            <?php
-            echo "<form method='post' name='my_form' " .
-                "action='$rootdir/forms/aftercare_plan/save.php?id=" . attr_url($formid) . "'>\n";
-            ?>
+            <form method="post" name="my_form" action="<?php echo $form_action;?>">
             <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken()); ?>" />
-
             <table class="table">
                 <tr>
                     <td  class="forms">
@@ -124,28 +120,28 @@ $obj = $formid ? formFetch("form_aftercare_plan", $formid) : array();
                 <tr>
                     <td class="forms">1.</td>
                     <td colspan="3">
-                        <textarea name="goal_a_acute_intoxication" rows="2"
+                        <textarea name="goal_a_I" rows="2"
                                   class="form-control"
                                   cols="80" wrap="virtual name"><?php
-                                    echo text($obj["goal_a_acute_intoxication"]);?></textarea>
+                                    echo text($obj["goal_a_I"]);?></textarea>
                     </td>
                 </tr>
                 <tr>
                     <td class="forms">2.</td>
                     <td colspan="3">
-                        <textarea name="goal_a_acute_intoxication_I"
+                        <textarea name="goal_a_II"
                                   class="form-control"
                                   rows="2" cols="80" wrap="virtual name"><?php
-                                    echo text($obj["goal_a_acute_intoxication_I"]);?></textarea>
+                                    echo text($obj["goal_a_II"]);?></textarea>
                     </td>
                 </tr>
                 <tr>
                     <td class="forms">3.</td>
                     <td colspan="3">
-                        <textarea name="goal_a_acute_intoxication_II"
+                        <textarea name="goal_a_III"
                                   class="form-control"
                                   rows="2" cols="80" wrap="virtual name"><?php
-                                    echo text($obj["goal_a_acute_intoxication_II"]);?></textarea>
+                                    echo text($obj["goal_a_III"]);?></textarea>
                     </td>
                 <tr>
                     <td class="forms-subtitle">
@@ -160,19 +156,19 @@ $obj = $formid ? formFetch("form_aftercare_plan", $formid) : array();
                     <td class="forms">1.</td>
                     <td colspan="3">
                         <textarea
-                                name="goal_b_emotional_behavioral_conditions"
+                                name="goal_b_I"
                                 class="form-control"
                                 rows="2" cols="80" wrap="virtual name"><?php
-                                echo text($obj["goal_b_emotional_behavioral_conditions"]);?></textarea>
+                                echo text($obj["goal_b_I"]);?></textarea>
                     </td>
                 </tr>
                 <tr>
                     <td class="forms">2.</td>
                     <td colspan="3">
-                        <textarea name="goal_b_emotional_behavioral_conditions_I"
+                        <textarea name="goal_b_II"
                                   class="form-control"
                                   rows="2" cols="80" wrap="virtual name"><?php
-                                    echo text($obj["goal_b_emotional_behavioral_conditions_I"]);?></textarea>
+                                    echo text($obj["goal_b_II"]);?></textarea>
                     </td>
                 </tr>
 
@@ -188,19 +184,19 @@ $obj = $formid ? formFetch("form_aftercare_plan", $formid) : array();
                 <tr>
                     <td class="forms">1.</td>
                     <td colspan="3">
-                        <textarea name="goal_c_relapse_potential"
+                        <textarea name="goal_c_I"
                                   class="form-control"
                                   rows="2" cols="80" wrap="virtual name"><?php
-                                    echo text($obj["goal_c_relapse_potential"]);?></textarea>
+                                    echo text($obj["goal_c_I"]);?></textarea>
                     </td>
                 </tr>
                 <tr>
                     <td class="forms">2.</td>
                     <td colspan="3">
-                        <textarea name="goal_c_relapse_potential_I"
+                        <textarea name="goal_c_II"
                                   class="form-control"
                                   rows="2" cols="80" wrap="virtual name"><?php
-                                    echo text($obj["goal_c_relapse_potential_I"]);?></textarea>
+                                    echo text($obj["goal_c_II"]);?></textarea>
                     </td>
 
                 </tr>
--- a/interface/forms/aftercare_plan/new.php
+++ b/interface/forms/aftercare_plan/new.php
@@ -56,11 +56,9 @@ $obj = $formid ? formFetch("form_aftercare_plan", $formid) : array();
 <body class="body_top">
 <div class="container">
     <div class="row">
-        <div class="mt-5">
-            <div class="col-md-12 mt-5">
+
                 <h1 class="forms-title"><?php echo xlt('After Care Planning'); ?></h1>
-            </div>
-        </div>
+
         <div class="col-md-12 mt-4">
             <?php
             echo "<form method='post' name='my_form' " .
@@ -70,17 +68,15 @@ $obj = $formid ? formFetch("form_aftercare_plan", $formid) : array();
 
             <table class="table">
                 <tr>
-                    <td  class="forms" class="forms">
+                    <td  class="forms">
                         <?php echo xlt('Client Name'); ?>:
                     </td>
                     <td class="forms">
                         <label class="forms-data">
-                            <?php if (is_numeric($pid)) {
+                            <?php
                                 $result = getPatientData($pid, "fname,lname,squad");
                                 echo text($result['fname']) . " " . text($result['lname']);
-                            }
-
-                            $patient_name = ($result['fname']) . " " . ($result['lname']);
+                                $patient_name = ($result['fname']) . " " . ($result['lname']);
                             ?>
                         </label>
                         <input type="hidden" name="client_name" value="<?php echo attr($patient_name);?>">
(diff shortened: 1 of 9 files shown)


sjpadgett:
IMO this form should be a contrib form and not part of normal encounter forms.
It is specific to addiction and should be charted using functional/cognitive, clinical notes and care plan as these are the forms our cda and api grabs this type of info.
The mentioned forms should be encouraged and other discouraged. Don't forget certification requirements plus QRDA/Measures.

I don't oppose your PR @juggernautsei but at some point we have to make clear the correct workflows if our other tools are to be useful.
I think we all understand this however, it will not be an easy task for sure man!:)

sjpadgett:
@robertdown and I are looking into a viable option to make available more structure clinical notes and care plans.

juggernautsei:
I see what you are saying and the EU that is using this form did know why the goals were hard coded. We can just drop this request if you want.

sjpadgett:
Nope leave it for EU/international plus legacy for those using in the past. Thanks Sherwin for the update

interface/forms/aftercare_plan/new.php
```diff
@@ -1,20 +1,22 @@
...
  * @copyright Copyright (c) 2017-2019 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2022 Sherwin Gaddis <sherwingaddis@gmail.com>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
 require_once(__DIR__ . "/../../globals.php");
-require_once("$srcdir/api.inc.php");
-require_once("$srcdir/patient.inc.php");
+require_once("$srcdir/api.inc");
+require_once("$srcdir/patient.inc");
```
sjpadgett, line 19: Are these bootstrapped to the .php version?

sjpadgett:
> Are those hard coded goals still needed?

I don't see what you mean hard coded goal. Example please

juggernautsei:
Here is the original form:

https://github.com/openemr/openemr/blob/5b079fa52c42025c846d1a940f3c8360616a66b1/interface/forms/aftercare_plan/new.php#L115

          <td class="forms-subtitle" colspan="4"><B><?php echo xlt('Goal A');?>:</B>&nbsp;<?php echo xlt('Acute Intoxication/Withdrawal'); ?></td>

         <td class="forms-subtitle" colspan="4"><B><?php echo xlt('Goal B');?>:</B>&nbsp;<?php  echo xlt('Emotional / Behavioral Conditions & Complications'); ?></td>

         <td class="forms-subtitle" colspan="4"><B><?php echo xlt('Goal C');?>:</B>&nbsp;<?php  echo xlt('Relapse Potential'); ?></td>

sjpadgett:
I see! Yeah having user input the goal is a more versatile use of the form however again, Care Plan does the same thing but allows adding as many goals as user wants. This form is pretty much a more limited duplicate of careplan but using careplan with the correct code for follow up goal is the same. With this then you would also have the original plan for the follow up that should have be added prior.

juggernautsei:
@sjpadgett I understand what you are saying. Sometimes we have no idea why one form over another is selected. I was surprised this form was enabled by the EU when the Care Plan is more elaborate. And that could be the draw is the simplicity.

sjpadgett:
Yep, clinicians will be drawn to simplicity(well put) however this form should not be included but would be hard to remove because of legacy.
We'll probably have to create a document on the recommended workflow use of encounter forms if reporting with CCDA and API is important to the users.
I'm done rambling and thanks.:)

interface/forms/aftercare_plan/new.php
```diff
@@ -1,20 +1,22 @@
 <?php
 
-/**
```
robertdown, line 3: Dumb, I know, but I think the double asterisks should stay for doc-blcks

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
-    <td><input type='submit' value='<?php echo xla('Save'); ?>' class='button-css' />&nbsp;
-    <input type='button' value='<?php echo xla('Print'); ?>' id='printbutton' class='button-css' />&nbsp;
-    <input type='button' class='button-css' value='<?php echo xla('Cancel'); ?>'
- onclick="parent.closeTab(window.name, false)" /></td>
-    </tr>
-</table>
-</form>
+<div class="container">
+    <div class="row">
+        <div class="mt-5">
```
robertdown, line 59: This div is unnecessary

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+            <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken()); ?>" />
+
+            <table class="table">
+                <tr>
+                    <td  class="forms" class="forms">
+                        <?php echo xlt('Client Name'); ?>:
+                    </td>
+                    <td class="forms">
+                        <label class="forms-data">
+                            <?php if (is_numeric($pid)) {
```
robertdown, line 78: I know it's a hold over from the previous form, but PID is likely always numeric, and if not, things will break way before this. Can probably remove this conditional

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+
+                            $patient_name = ($result['fname']) . " " . ($result['lname']);
+                            ?>
+                        </label>
+                        <input type="hidden" name="client_name" value="<?php echo attr($patient_name);?>">
+                    </td>
+                    <td   class="forms"><?php echo xlt('DOB'); ?>:</td>
+                    <td class="forms">
+                        <label class="forms-data">
+                            <?php if (is_numeric($pid)) {
```
robertdown, line 91: Same for this conditional

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+                                  class = "form-control"
+                                  cols="80" wrap="virtual name"><?php
+                                    echo text($obj["goal_a_acute_intoxication"]);?></textarea>
+                    </td>
+                </tr>
+                <tr>
+                    <td class="forms">2.</td>
+                    <td colspan="3">
+                        <textarea name="goal_a_acute_intoxication_I"
+                                  class = "form-control"
```
robertdown, line 142: Remove spaces around `=`

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+                        <input type="text" class="form-control" name="goal_b" value=""
+                               placeholder="<?php echo xlt('Enter patient second goal'); ?>">
+                    </td>
+                </tr>
+                <tr>
+                    <td class="forms">1.</td>
+                    <td colspan="3">
+                        <textarea
+                                name="goal_b_emotional_behavioral_conditions"
+                                class = "form-control"
```
robertdown, line 169: Remove spaces around `=`

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+                                rows="2" cols="80" wrap="virtual name"><?php
+                                echo text($obj["goal_b_emotional_behavioral_conditions"]);?></textarea>
+                    </td>
+                </tr>
+                <tr>
+                    <td class="forms">2.</td>
+                    <td colspan="3">
+                        <textarea name="goal_b_emotional_behavioral_conditions_I"
+                                  class = "form-control"
```
robertdown, line 178: Remove spaces around `=`

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+                    <td colspan="3">
+                        <input type="text" class="form-control" name="goal_c" value=""
+                               placeholder="<?php echo xlt('Enter patient third goal'); ?>">
+                    </td>
+                </tr>
+                <tr>
+                    <td class="forms">1.</td>
+                    <td colspan="3">
+                        <textarea name="goal_c_relapse_potential"
+                                  class = "form-control"
```
robertdown, line 197: Remove space around `=`

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+                                  rows="2" cols="80" wrap="virtual name"><?php
+                                    echo text($obj["goal_c_relapse_potential_I"]);?></textarea>
+                    </td>
+
+                </tr>
+                <tr>
+                    <td></td>
+                    <td><input type='submit' value='<?php echo xla('Save'); ?>' class='btn btn-primary' />&nbsp;
+                        <input type='button' value='<?php echo xla('Print'); ?>'
```
robertdown, line 215: Convert `input` to `button` and make it `btn-secondary` instead of `btn-warning`

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+                               id='printbutton' class='btn btn-warning' />&nbsp;
+                        <input type='button' class='btn btn-danger' value='<?php echo xla('Cancel'); ?>'
```
robertdown, line 217: Convert `input` to `button` and make this `btn-cancel` instead of `btn-danger`

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,198 @@
...
+                        <textarea name="goal_c_relapse_potential_I"
+                                  class = "form-control"
+                                  rows="2" cols="80" wrap="virtual name"><?php
+                                    echo text($obj["goal_c_relapse_potential_I"]);?></textarea>
+                    </td>
+
+                </tr>
+                <tr>
+                    <td></td>
+                    <td><input type='submit' value='<?php echo xla('Save'); ?>' class='btn btn-primary' />&nbsp;
```
robertdown, line 214: Should probably wrap these in a div with class of `btn-group`

interface/forms/aftercare_plan/new.php
juggernautsei: In phpstorm, for the license block, it now uses a single asterisk. It does not treat it as doc block.

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,197 @@
...
-</form>
+<div class="container">
+    <div class="row">
+
+                <h1 class="forms-title"><?php echo xlt('After Care Planning'); ?></h1>
+
+        <div class="col-md-12 mt-4">
+            <?php
+            echo "<form method='post' name='my_form' " .
+                "action='$rootdir/forms/aftercare_plan/save.php?id=" . attr_url($formid) . "'>\n";
```
robertdown, line 65: Since you're refactoring the form, this is a good opportunity to reduce the amount of HTML being echo'd. This is better as:

```
<?php 
$form_action = "$rootdir/forms/aftercare_plan/save.id=" . attr_url($formId);
</php
<form method="post" name="my_form" action="<?php echo $form_action;?>">
```
not a huge deal, but the less HTML we have embedded in PHP, the better.

interface/forms/aftercare_plan/new.php
bradymiller: need to revert this block back to .php endings

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,193 @@
...
-    </tr>
-</table>
-</form>
+<div class="container">
+    <div class="row">
+
+                <h1 class="forms-title"><?php echo xlt('After Care Planning'); ?></h1>
+
+        <div class="col-md-12 mt-4">
+            <form method="post" name="my_form" action="<?php echo $form_action;?>">
```
bradymiller, line 63: I can not see where $form_action is defined above

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,193 @@
...
+                </tr>
+                <tr>
+                    <td class="forms-subtitle" colspan="4"><strong><?php echo xlt('Goal and Methods');?></strong></td>
+                </tr>
+                <tr>
+                    <td class="forms-subtitle">
+                        <strong><?php echo xlt('Goal A');?>:</strong>
+                    </td>
+                    <td colspan="3">
+                        <input type="text" class="form-control" name="goal_a" value=""
```
bradymiller, line 116: Doesn't changing the name attributes have downstream repercussions when saving?

interface/forms/aftercare_plan/new.php
```diff
@@ -26,153 +28,193 @@
...
-
-        <td class="forms-subtitle" colspan="4"><B><?php echo xlt('Goal and Methods');?></B></td>
-
-    </tr>
-    <tr>
-        <td align="left colspan="3" style="padding-bottom:7px;"></td>
-    </tr>
-    <tr>
-
-        <td class="forms-subtitle" colspan="4"><B><?php echo xlt('Goal A');?>:</B>&nbsp;<?php echo xlt('Acute Intoxication/Withdrawal'); ?></td>
```
bradymiller, line 115: What happens to users that were using the previous form? (ie. when they edit old forms or make new forms looks like there will be changes in labels)

juggernautsei: I can see everything breaking. This is now a totally different form and should be saved under a new name. Once I changed the field names then the old saved data will not pull into the fields. If the fields were reverted back to their previous names at least the fields would populate with the old data. The goals would still be blank because they were hard coded before.
That is why I was just trying to update the look and not go to far as to destroy the backwards compatibility of the form. at this point the backwards compatibility is destroyed.

robertdown:
We now have a different problem. If this now contains breaking changes to the previous form, as you stated, we are now introducing a new form... which gets back to Jerry's point about how this form is not best practice when compared to the Care Plan form.

If this remains a purely styling PR, just getting everything up to our bootstrap standard (which I think is a worthy goal) we are good. But I'm hesitant to bring in a _new_ form into core that has limited functionality.

juggernautsei:
I agree with you. This was a styling mission to begin with. If we roll back to [bf2738d](https://github.com/openemr/openemr/pull/6026/commits/bf2738db8e9549d7cf2c44cd0efd6ef77db29868) and continue from there. We should be able to complete the styling mission. I am not trying to introduce a new form

juggernautsei:
My original idea was to make the form pretty to look at. Now, with the input of @sjpadgett. I have abandoned this project. I  have other fish to fry.

PR #6053, opened 2022-12-23
LHC-Form update v33.1 from 33.0

sjpadgett:
- custom mod to not append score to answer will be an config option in next release.
- project bug fixes and version bump to v33.1

Fixes #


bradymiller:
2DV

PR #6058, opened 2022-12-28
Fixes #6056

### LINKED ISSUE #6056: Dark theme does not change Select2 box
Select2 box does not entirely get inverted when using the Dark theme. Backgrounds remain light while text _does_ invert, leading to white text color on white background. Here is an example from a PR I'm working on. I'll spin off a separate PR that solves just the Select2 issue before submitting the actual PR I am working on

image

robertdown:
Fixes #6056 

Not 100% and I would love to spend more time and make this look _good_ versus just _usable_ but for now, this will have to suffice. Current state is unusable, this PR makes it usable.

### DIFF
--- a/interface/themes/oe-styles/style_dark.scss
+++ b/interface/themes/oe-styles/style_dark.scss
@@ -90,5 +90,13 @@ $fa-font-path: "../assets/@fortawesome/fontawesome-free/webfonts";
   }
 }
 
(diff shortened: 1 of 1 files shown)


adunsulag:
Closing this down as we've had no activity for 3 years.  Feel free to re-open it if you want to keep working on it.

PR #6060, opened 2022-12-28
Better Page Heading UI

robertdown:
Better wrapping for Page Heading - it is now a flex div and pushes icon to the right, sizing down the icons. Previously, they were nested inside the `h2` element, this puts all the icons in a better place.

(Screenshot has a wrench icon... that is a side effect from some other work, just pretend it is not there)

image

### DIFF
--- a/src/OeUI/OemrUI.php
+++ b/src/OeUI/OemrUI.php
@@ -99,7 +99,18 @@ class OemrUI
             $action_icon = $this->actionIcon();
             $help_icon = $this->helpIcon();
             $expandable_icon = $arrexpandIcon[0];
-            $heading = "<h2>$heading $expandable_icon $action_icon $help_icon</h2>";
+            $heading = <<<HTML
+                <div class="d-flex">
+                    <div class="flex-grow-1">
+                        <h2>$heading</h2>
+                    </div>
+                    <div class="pt-2">
+                        $expandable_icon
+                        $action_icon
+                        $help_icon
+                    </div>
+                </div>
+            HTML;
         } else {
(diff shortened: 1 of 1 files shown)


sjpadgett:
I don't see a reason to have title Medical Record Dashboard taking up valuable real estate.
We know the patient and we also know we're in the Dashboard so it serves no purpose.

robertdown:
Oh I agree with you! However, I don't want to fully remove it in this PR. I think it can be better styled, but I've started to focus on small, easy to integrate PRs.

I actually want to move the menu of the MRD into a navbar, which already supports both a page title and menu. But it would require more refactoring than I want to do right now.

However - for the meantime, what if we just make the page heading smaller? Like this

image

instead of 

image

sjpadgett:
Sounds good thanks but if you are coming back to this then I'd just leave as is and fix in the more in-depth styling PRs.

robertdown:
Work is already done :) would prefer to get this in the codebase. Not sure how complicated the refactor will be to move a menu into a different part of the render logic, incremental improvement is better than a long refactor.

sjpadgett:
Okay, thanks for the adjustment. 
lol once you get started styling it seems like it never ends! At least for me....

robertdown:
> Okay, thanks for the adjustment. 
> 
> lol once you get started styling it seems like it never ends! At least for me....

Agreed. That's why I just have to say "ok, this is a good stopping point - I can't fix it all in 1 PR" tomorrow is another day.

robertdown:
I can't react to the label add directly, so I have to do it here

image

bradymiller:
2DV

PR #6064, opened 2022-12-29
Portal document delete exception and New Print button

### LINKED ISSUE #6063: Portal saved document delete throws exception.
### Describe the bug
When deleting a rendered document from history an user exception is thrown.
The cause is when validating the patient passed in PID to the Global set PID for security there is a typing int to string mismatch.
Was using strict conditional so I throw a security exception.

### Client configuration

- Browser:
- OpenEMR version: v7/master
- Operating system: windows/linux

sjpadgett:
- type cast pids to same type, int.

Fixes #6063


bradymiller:
2DV

sjpadgett:
Since I haven't heard from my tester will merge later today.

PR #6074, opened 2023-01-01
Make it easier to have different logos on multi-site installs

### LINKED ISSUE #6073: [FEATURE] Allow custom logo on login page via a no-code solution
### Describe the problem

It is a task to have a different logo for each of the installs for a multi-site
Could this be made easier?

it would be nice to just drop a logo into the sites folder and the system pick up that logo for that site

juggernautsei:
Fixes #6073

dropping in a logo for the different sites in a multi site installation

move the login logo to the site/"site_id"/images folder and point the login page to the location.


stephenwaite:
nice

stephenwaite:
what if you made a new global for it instead?

juggernautsei:
Not sure that would work. Thinking about it the file location would still need to change and the site ID would still be needed. Seems the same either way. I have no preference. I just saw the opportunity while upgrading a multi site were we had done special coding to handle the logo situation.

interface/login/login.php
```diff
@@ -238,7 +240,7 @@ function getLanguagesList(): array
     'googleSigninClientID' => $GLOBALS['google_signin_client_id'],
     'logoArea' => $logoarea,
     'displayExtraLogo' => $GLOBALS['extra_logo_login'],
-    'primaryLogoSrc' => file_get_contents($GLOBALS["images_static_absolute"] . "/login-logo.svg"),
+    'primaryLogoSrc' => file_get_contents($loginlogolocation),
```
bradymiller, line 243: If we move the main logos into this directory then many upgraders won't get updated logos in future.
Rec:
1. keeping main logos where they originally were in public and use them by default
2. use the logo at `$GLOBALS["images_static_absolute"] . "/login-logo.svg"` if it exists (also put a comment above this line describing what it does; ie. if a user placed a file at `site/"site_id"/images/login-logo.svg`, then it will override the logo on the logon screen)

interface/login/login.php
```diff
@@ -223,6 +223,8 @@ function getLanguagesList(): array
     $cookie = json_encode($cookie);
 }
 
+$loginlogolocation = dirname(__DIR__, 2) . "/sites/" . $_SESSION['site_id'] . "/images/login-logo.svg";
```
bradymiller, line 226: btw,
This line is where would put the check for the custom logo file and the comment

portal/index.php
```diff
@@ -439,7 +439,7 @@ function enableVerifyBtn(){
             </form>
         <?php } else {
             ?> <!-- Main logon -->
-        <img class="img-responsive center-block login-image" src='<?php echo $GLOBALS['images_static_relative']; ?>/login-logo.png' />
+        <img class="img-responsive center-block login-image" src='<?php echo "../sites/" . $_SESSION['site_id'] . "/images"; ?>/login-logo.png' />
```
bradymiller, line 442: here also need to check for the directory and the comment

Since this in in 2 places, would recommend doing the file check and comment in globals.php script and setting the logo to variable `$GLOBALS['loginlogopath']` or something like that, which then can use in both the core/portal login scripts.

bradymiller:
note there is also svg logo use at:
interface/smart/register-app.php

After that and realizing that svg is used in core/smart screens and png is use in portal, recommend block of code in globals.php that checks for existence of the custom svg/png logos and then sets 2 globals:
```
$GLOBALS['corelogo']
$GLOBALS['portallogo']
```

juggernautsei:
I am not seeing how that is going to auto select the correct logo for multi site?

bradymiller:
Code block in global.php would be something like:
```
// Comment here explaining how users can customize the logo in core and in the portal
$GLOBALS['corelogo'] = the svg in original public folder
$GLOBALS['portallogo'] = the png in original public folder
if (custom svg file exists) {
   $GLOBALS['corelogo'] = custom svg file
}
if (custom png file exists) {
   $GLOBALS['portallogo'] = custom png file
}
```
Then can use `$GLOBALS['corelogo']` in the core login and smart gui and `$GLOBALS['portallogo']` in the portal login

bradymiller:
also, in terms of ensuring it follows multisite, would just ensure the global.php block of code is after the `$GLOBALS['OE_SITE_DIR']` which could use when checking for existance of the custom svg/png logos

juggernautsei:
I respectfully suggest that the global does not make a no coding solution. 
Why are we depending on the globals to solve this issue. Isn't the globals bloated enough? This is just my opinion and does not rise to the level of anything worth more than a pine bark dime.

bradymiller:
You could put it in non-global, but then the variable wouldn't always be in scope inside classes and functions (this is why there a variables that hold same info like `$v_js_includes` and  `$GLOBALS['v_js_includes']` :) )

You are correct, though, that since this is only used in 3 scripts, would be nice to avoid processing for this only when needed. Could make a very simple class maybe called LogoPath that takes either 'core' or 'portal' as parameter in the constructor that has a function maybe called getPath that returns the logo path. So then the following one-liners will work in all 3 scripts:
`$path = (new LogoPath('core'))->getPath();` for core/smart
and
`$path = (new LogoPath('portal'))->getPath();` for portal

juggernautsei:
=^)
Why make this more complicated than it needs to be. Not everything needs to go in the globals and not everything needs to be globally assessable. In this case this only needs to be used right at run time. To me the only thing that needs to be concern with the login logo are the login pages. Nothing else in the system care where these logos are. In my opinion only.

Keep It Simple S... 
https://www.youtube.com/@ProgramWithGio

bradymiller:
@juggernautsei ,
1. Portal logo, which is a jpg is in 1 script
2. Core logo, which is a svg is in 2 scripts

Those 3 places will reuse same code, so seems like simplest thing is to centralize the code, hence a class/function. This also avoids using different mechanism in same places. And then could be extended in future to support logos elsewhere or different way of customizing logos. And then not tough for folks to figure out how to modify the logos. Note I think this is a helpful feature too since modifying the logo is asked on forums quite frequently and is not super simple to understand and basically impossible for multisite. I'm happy to create the class/feature to support this since would be a nice break from what i'm doing now (bit of kubernetes hell) and I already have a solid idea of how to do it.

juggernautsei:
Let me look into it and see what I can come up with.

interface/login/login.php
juggernautsei: The global would need a space where the name of the file can be entered. Is that what you were thinking?

interface/login/login.php
juggernautsei: The only way to know it exists is that there has to be an entry of that file name.

interface/login/login.php
robertdown: Look at $GLOBALS['OE_SITES_BASE'] instead of dirname()

robertdown (changes requested):
Look at https://github.com/openemr/openemr/blob/master/interface/globals.php#L657

Also, the PNG is the old version of the logo, our current one has the infinity circle and text sized closer together. I'll send it over

juggernautsei:
True. But I was trying to travel down Brady's suggestion to make this a class function. To make this usable by the average user and not make this another hire and engineer thing to do is what I am struggling with at the moment. There need to be two global entries for both locations as to what the file name is. It will need to distinguish file type between png and SVG. Or there needs to be some language in place to say which file type to use. If this is to become a class.  @robertdown am I over thinking it?

robertdown:
Yeah, maybe just a bit. Instead maybe just whitelist of extensions that can sit in sites/site_id/images/login_logo/

Then your conditions be

* there must only be 1 file in that folder
* that file must be an image
* that file must have a file name of "login_logo"
* if there's a file in sites it takes priority

Given our long time position on not allowing a user-uploaded logo, the end user is going to _have to_ manually place the logo in the sites directory.

robertdown:
I don't think we need to make this into $GLOBALS. It's far easier to just say "If you want a custom logo, you **must** follow this schema. Then we build in the logic directly into login. 

Could have the following structure

```
sites/
sites/<site_id>/
sites/<site_id>/images/
sites/<site_id>/images/logos/

sites/<site_id>/images/logos/core/
sites/<site_id>/images/logos/core/login/
sites/<site_id>/images/logos/core/login/primary/
sites/<site_id>/images/logos/core/login/primary/login-logo.png|svg|jpg
sites/<site_id>/images/logos/core/login/mobile/
sites/<site_id>/images/logos/core/login/mobile/login-logo.png|svg|jpg

sites/<site_id>/images/logos/portal/
sites/<site_id>/images/logos/portal/login/
sites/<site_id>/images/logos/portal/login/primary/
sites/<site_id>/images/logos/portal/login/primary/login-logo.png|svg|jpg
sites/<site_id>/images/logos/portal/login/mobile/
sites/<site_id>/images/logos/portal/login/mobile/login-logo.png|svg|jpg
```

This keeps us from messing with GLOBALS while keeping the ability to make a change relatively straightforward. We are trading ease-of-use for security. Ideally, we would just allow the user to securely upload these items from the UI, but that's been a hard no for a long time (We used to allow this, many moons ago).

juggernautsei:
@robertdown I agree. This should be dropped. The new file structure makes it easier for me. I should have put this in a thread on the message board instead.

PR #6090, opened 2023-01-11
Hook for page action buttons

robertdown:
Create a new event and interface for injecting buttons into the page header rendered by OemrUI. Convert the current help icon, action icon, and expand icons into listeners to the new event. Move the pageHeading() rendering to Twig.

### TBD
~1. The Action Icon has some logic I need to still resolve~
~2. Create spot to inject custom Javascript to actually do something with your fancy new button.~


src/OeUI/OemrUI.php
```diff
@@ -14,6 +14,11 @@
 
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Core\Header;
+use OpenEMR\Events\UserInterface\BaseActionButtonHelper;
+use OpenEMR\Events\UserInterface\PageHeadingRenderEvent;
+use OpenEMR\Common\Twig\TwigContainer;
+use OpenEMR\Events\UserInterface\SampleActionButtonClass;
+use Symfony\Component\EventDispatcher\EventDispatcher;
```
bradymiller, line 21: sort alphabetically

templates/oemr_ui/page_heading/partials/page_heading.html.twig
```diff
@@ -0,0 +1,26 @@
...
+<nav class="navbar navbar-light navbar-expand-sm bg-light pt-0">
+    {% block pageHeading %}
+        <span class="navbar-brand mb-0 h1">{{ heading }}</span>
+    {% endblock %}
+
+    {% block actionButtons %}
+        <div class="collapse navbar-collapse">
+            <div class="navbar-nav ml-auto" id="pageHeadingNav">
+                {% for a in buttonList %}
+                    <a href="{{ a.getHref() }}" id="{{ a.getId() }}" title="{{ a.getTitle() }}" {{ a.getAttributes() }} class="nav-link {{ a.getAnchorClasses()|join(' ') }} "><i class="{{ a.getIconClass() }}">&nbsp;</i>{{ a.getDisplayText() }}</a>
```
bradymiller, line 20: Would escape here instead of in the upstream code:
```
<a href="{{ a.getHref()|attr }}" id="{{ a.getId()|attr }}" title="{{ a.getTitle()|attr }}" {{ a.getAttributes() }} class="nav-link {{ a.getAnchorClasses()|join(' ')|attr }} "><i class="{{ a.getIconClass()|attr }}">&nbsp;</i>{{ a.getDisplayText()|text }}</a>
```

templates/oemr_ui/page_heading/partials/page_heading.html.twig
```diff
@@ -0,0 +1,26 @@
...
+@package OpenEMR
+@subpackage OemrUi
+
+@author Robert Down <robertdown@live.com>
+@copyright Copyright (c) 2023 Robert Down
+@copyright Copyright (c) 2023 Providence Healthtech
+#}
+<nav class="navbar navbar-light navbar-expand-sm bg-light pt-0">
+    {% block pageHeading %}
+        <span class="navbar-brand mb-0 h1">{{ heading }}</span>
```
bradymiller, line 13: escape here instead of upstream:
```
<span class="navbar-brand mb-0 h1">{{ heading|text }}</span>
```

src/OeUI/OemrUI.php
```diff
@@ -82,39 +89,60 @@ public function __construct($arrOeUiSettings = array())
...
-                        $expandable_icon
-                        $action_icon
-                        $help_icon
-                    </div>
-                </div>
-            HTML;
+        $pageHeadingEvent = $this->ed->dispatch(new PageHeadingRenderEvent(), PageHeadingRenderEvent::EVENT_PAGE_HEADING_RENDER);
+        $vars = [
+            "buttonList" => $pageHeadingEvent->getActions(),
+            "heading" => text($this->heading),
```
bradymiller, line 136: much better to html escape as far downstream as possible for variables (ie. escape it in twig when it is outputted). This makes the code much easier to maintain security and avoid double escaping. Also then doesn't pigeon hole variables to certain escaping contexts (ie. if the twig template wishes to place the variable in a attr() context or a javascript context). There are several more variable below (such as title which is escaped prior to twig in 1 case and not the other also demonstrating why best to get it as downstream as possibe) which should also do this.

bradymiller:
@robertdown , just noted some escaping stuff. Otherwise code looks great to me :) (btw, I didn't test it)

robertdown:
Updated the way javascript interacts with the interface

The interface dictates `getAttributes()` must return an array of attributes to be injected into the element. This means you could do the following:

```php
public function getAttributes()
{
    return [
        "onclick" => "function(){ alert('do something'); }"
    ];
}
```

which would render the onclick attribute (Though the escaping would be problematic because it gets run through `attr` not a js-specific escape.

OR, you could leverage `getClickHandlerTemplateName()` and `getClickHandlerFunctionName()` which returns the name of the twig template containing your function and the name of the function itself. If you return those, twig will automatically add an event listener to your button.

Still some downsides here, Ideally, the template being included should _only_ contain 1 function, named whatever your pass in getClickHandlerTemplateName(), however, it could be expanded and is now effectively a hook for any javascript you want.  Not sure how to force a better solution though.

sjpadgett:
@robertdown Installing this PR errors javascript with
image
image

And

image

robertdown:
Hmmm, the patient finder page was one I tested against. Let me try to replicate

sjpadgett:
Looks like a dependency scope issue. showActionClass looks like not making it into the DOM. You may need to change how it's declared.

robertdown:
@sjpadgett I could not replicate the error.

Steps I took
1. docker-compose down -v to clear any lingering
2. git checkout master
3. git pull upstream master
4. docker pull openemr/openemr:latest
5. docker-compose up -d from development-easy to bring up a completely fresh instance

sjpadgett:
Okay. I was treating it as a patch which should work and will still see what it takes to work that way. In the end I'll spin up a docker as you did.

sjpadgett:
I think this has to do with the initial hide stats is in for the section and will account for why you're not seeing.
There is also the sql prepare error.
Or it could be my current dev environment so still checking it out

robertdown:
Not sure where the SQL error is coming from, I'm not doing any DB interaction. I built this off of master and I think Brady is planning an actual release, not a patch in the coming days

sjpadgett:
Nope sorry that was me not updating sql. I have a simple fix for the javascript error which I'll add to my existing PR.
With that we should be good. give me a call if need to.

src/OeUI/OemrUI.php
```diff
@@ -455,21 +533,20 @@ private function headerActionJs($arrAction = array())
...
 
             // We display if we remember we're showing it, but we don't intentionally hide it (no else here to hide)
             // Because the hideaway is probably shown by default for a reason like in the billing manager
             if (shouldDisplay) {
-                $('.hideaway').show(500);
-                $('#show_hide').removeClass(showActionClass);
-                $('#show_hide').addClass(hideActionClass);
+                document.querySelector('.hideaway').classList.remove('d-none');
+                elementIcon.classList.remove(showActionClass);
+                elementIcon.classList.add(hideActionClass);
```
sjpadgett, line 545: elementIcon is never defined I can see. It is defined within the show_hide click() handler so it would be out of scope elsewhere.
I don't see how this would work in any case because while  show_hide click will set as soon as routine is exited then you lose scope. elementIcon needs to be assigned/declared somewhere global.

robertdown:
Could you split it into its own PR. Still can't replicate so I want to test further

sjpadgett:
If you're not seeing the error then you are not hitting the routine due to shouldDisplay not being set at around L-544 if (should Display) { ...}
After testing on this issue for awhile somehow something or flag was set that I no longer hit the errant code.
Easier if you give me a call to discuss. I can revert my fix if you want but the changes I made are necessary regardless if that section is never hit in our testing.

robertdown:
I can give you a call Sunday at the earliest. I trust whatever your fix is

PR #6100, opened 2023-01-16
fix: #6097 uuid failing on ccda generation

### LINKED ISSUE #6097: CCDA Office Contact Missing Organization UUID should not crash ccda document
If the ccda global configuration's office contact is missing an organization's uuid it should not crash the ccda document generation.  This can occur when an organization is not connected to a provider, or if the organization somehow is missing a uuid.

adunsulag:
Fixes #6097 

Makes it so we don't blow up the ccda generation when the uuid doesn't generate.  This only happens when you specify an office contact and the organization uuid is missing or the organization for the office contact is missing.

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
@@ -728,7 +728,17 @@ class EncounterccdadispatchTable extends AbstractTableGateway
         if (empty($details)) {
             return '';
         } else {
-            $organization_uuid = UuidRegistry::uuidToString($details['facility_uuid']);
+            if (!empty($details['facility_uuid'])) {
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -728,7 +728,17 @@ public function getOfficeContact($pid, $encounter)
         if (empty($details)) {
             return '';
         } else {
-            $organization_uuid = UuidRegistry::uuidToString($details['facility_uuid']);
```
robertdown, line 731: Would it be easier to reduce the condition wrapping? This just keeps us from existing entirely in an if/then condition.

```php
$org_uuid = $details['facility_uuid'] ?? '';

if ($org_uuid == '') {
    // Process error
}
```

bradymiller:
bringing this over to rel-701 (in process of making sure rel-701 has all the most recent fixes from master)

PR #6104, opened 2023-01-17
Insurance card information adjustments 2

juggernautsei:
Fixes #5954 

From Dr Alexander Berlin, TX
Another important bug fix. In the new system, they got the insurance card wrong, with all insurances, including old insurances showing "until Present".

Here's the fix and it also introduces the concept of Self-Pay display for those times when a patient did not have insurance, both currently and in the past.

But wait there is more.
Unallocated amount. We have the ability to take a pre-payment on the front_payment screen, but it just doesn't show anywhere obvious, so that it could be applied to the account!

Go to patient chart, put in a couple of insurances (primary and secondary) with an effective date of let’s say a year ago. Then add another primary insurance (as a separate test, set to Unassigned for primary) with an effective date of a month ago and set secondary to Unassigned for the same effective date a month ago.
Then refresh patient demographics by clicking on the patient name above. See what you see for insurances and their effective dates (insurance tab on top - Insurance from… until …).

### DIFF
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -1048,8 +1048,11 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                         $patientbalance = get_patient_balance($pid, false);
                         $insurancebalance = get_patient_balance($pid, true) - $patientbalance;
                         $totalbalance = $patientbalance + $insurancebalance;
+                        $unallocated_amt = get_unallocated_patient_balance($pid); //ALB Added this
+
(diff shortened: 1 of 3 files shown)


bradymiller:
@juggernautsei ,
Rec making another PR via the following:
```
git checkout 75b8f1a21c
git checkout -b insurance_card_information_adjustments_3
```
Then push this branch to another PR. And then I can resolve the conflicts on that PR via github and bring it in.

bradymiller:
This feature was brought in via subsequent pull request: https://github.com/openemr/openemr/pull/6105

PR #6105, opened 2023-01-17
Insurance card information adjustments 3

### LINKED ISSUE #5953: old insurances showing "until Present" if latest insurance is Unassigned
### Describe the bug
From Dr Alexander Berlin, TX
Another important bug fix. In the new system, they got the insurance card wrong, with all insurances, including old insurances showing "until Present".

Here's the fix and it also introduces the concept of Self-Pay display for those times when a patient did not have insurance, both currently and in the past.

### To Reproduce

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

juggernautsei:
Fixes #5953 

Prior previous revision pull requests at #5954 and #6104 

From Dr Alexander Berlin, TX
Another important bug fix. In the new system, they got the insurance card wrong, with all insurances, including old insurances showing "until Present".

Here's the fix and it also introduces the concept of Self-Pay display for those times when a patient did not have insurance, both currently and in the past.

But wait there is more.
Unallocated amount. We have the ability to take a pre-payment on the front_payment screen, but it just doesn't show anywhere obvious, so that it could be applied to the account!

Go to patient chart, put in a couple of insurances (primary and secondary) with an effective date of let’s say a year ago. Then add another primary insurance (as a separate test, set to Unassigned for primary) with an effective date of a month ago and set secondary to Unassigned for the same effective date a month ago.
Then refresh patient demographics by clicking on the patient name above. See what you see for insurances and their effective dates (insurance tab on top - Insurance from… until …).


interface/patient_file/summary/demographics.php
```diff
@@ -1119,6 +1123,8 @@ function setMyPatient() {
                         $params[] = $pid;
                         $params = array_merge($params, $insurance_array);
                         $res = sqlStatement($sql, $params);
+                        $prior_ins_type = '';
```
bradymiller, line 1126: super odd that above line got here which was not in original. i will remove it.

bradymiller: oh wait, disregard above, it was there already. moving on :)

bradymiller:
Looking good. i will fix the psr12 issue and then bring this in.

stephenwaite:
excited_elephant

PR #6108, opened 2023-01-17
feat: allergy comments fields in REST API

IvanJov:
Dear Community,  I am representing the project of a Digital health platform developed by Cardo Health AB ( [www.cardohealth.com](http://www.cardohealth.com/)),  through the support of my current employer ( NearForm ). We are pleased to contribute some fixes in OpenEMR on behalf of Cardo Health and look forward to continuing to do so in the future.

In this PR we are exposing `comments` fields for the allergies entity. It's editable in UI and we thought that we can also use it in the API. Thanks!

### DIFF
--- a/src/Validators/AllergyIntoleranceValidator.php
+++ b/src/Validators/AllergyIntoleranceValidator.php
@@ -32,9 +32,10 @@ class AllergyIntoleranceValidator extends BaseValidator
             self::DATABASE_INSERT_CONTEXT,
             function (Validator $context) {
                 $context->required('title')->lengthBetween(2, 255);
-                $context->required('begdate')->datetime('Y-m-d');
+                $context->optional('begdate')->datetime('Y-m-d H:i:s');
                 $context->optional('diagnosis')->lengthBetween(2, 255);
-                $context->optional('enddate')->datetime('Y-m-d');
+                $context->optional('enddate')->datetime('Y-m-d H:i:s');
+                $context->optional('comments');
                 $context->required("puuid", "Patient UUID")->callback(
                     function ($value) {
                         return $this->validateId("uuid", "patient_data", $value, true);
--- a/src/RestControllers/AllergyIntoleranceRestController.php
+++ b/src/RestControllers/AllergyIntoleranceRestController.php
@@ -26,7 +26,8 @@ class AllergyIntoleranceRestController
         'title',
         'begdate',
         'enddate',
-        'diagnosis'
+        'diagnosis',
+        'comments'
     );
 
     public function __construct()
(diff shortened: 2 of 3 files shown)


src/Validators/AllergyIntoleranceValidator.php
```diff
@@ -32,9 +32,10 @@ protected function configureValidator()
             self::DATABASE_INSERT_CONTEXT,
             function (Validator $context) {
                 $context->required('title')->lengthBetween(2, 255);
-                $context->required('begdate')->datetime('Y-m-d');
+                $context->optional('begdate')->datetime('Y-m-d H:i:s');
```
adunsulag, line 35: Can you explain the reason for why you're making these fields be optional now?

adunsulag:
Thanks for the PR and glad to see you contributing to OpenEMR.  Just a minor question on the allergy begin date and everything else looks good.

src/Validators/AllergyIntoleranceValidator.php
IvanJov: Hello! Oh yeah, forgot to mention that 🤦‍♂️ 
It's just the use case that worked for us, I made it optional because that's how the UI works, and also, it's something we needed in the API. It might be a good idea to leave it optional since it's optional in the UI, too. What do you think?

sjpadgett: So if you're just sending an begin date w/o an end date, this is working for you.
Since we're not very granular in our searches and we don't do paging in our returns the required end date was prob used to limit returns.
Though I suppose one could send an end date far in the future.

This doesn't bother me if okay with you Stephen.

bradymiller: Also looks good to me

kurtmilam: High folks, I'm working with Ivan on this project via [NearForm](nearform.com), as well. Just wanted to chime in on this.

The reasoning behind not requiring a starting date for allergies was that most folks who report an allergy they have already been diagnosed with are unlikely to recall the exact date of diagnosis. Cardo's Head of Product, who is also a physician, preferred, in this scenario, having no starting date for allergies rather than an exact start date that was likely incorrect.

adunsulag: So I'm fine with the change.  I started looking more closely at the datetime change and wondering if we will break existing clients.  It looks like Particle\Validator\Chain->datetime(..) forces the format so if any existing clients are using just the 'Y-m-d' format this will break those clients. 

I think that's fine and it should be a fairly small change for users to update the date format especially since that's what we are using now in the GUI.  We'll just need to mention it in the changelog/patch notes that this is a breaking change.

PR #6110, opened 2023-01-17
Prepayment display in front payments

juggernautsei:
Fixes 6109

This is a little workaround, but it works if someone does a pre-payment. You already posted the one about displaying it on the billing card. This is to display and apply on the front payment (when the receptionist charges the patient, she can see that there is a pre-payment and apply it as needed).


library/patient.inc.php
```diff
@@ -1822,3 +1822,17 @@ function updateDupScore($pid)
     );
     return $dupscore;
 }
+
+function get_unallocated_payment_id($pid)
+{
+    $query = "SELECT a.session_id " .
+        "FROM ar_session AS a " .
+        "WHERE a.patient_id = ? AND " .
+        "a.adjustment_code = 'pre_payment' AND a.closed = 0 ORDER BY a.check_date ASC LIMIT 1";
```
sjpadgett, line 1831: Does check date get populated if a cash payment which occurs in some of your use cases?
Also it drives me nut when an alias is used when not necessary i.e ar_session as a!:) It slows down fetch. Just a personal note. Do what works...

sjpadgett: oh never mind check_date isn't pert of query, sorry!

PR #6140, opened 2023-01-29
Expand the PageHeading option to allow for primary menu

robertdown:
Fixes #


bradymiller:
@robertdown , code looks good to me :) (i didn't test it)

PR #6145, opened 2023-01-31
Fixes #6143 support UNION (,) searches

### LINKED ISSUE #6143: FHIR Patient _search is not handling UNION comma (,) operator
The patient _search operation is not handling UNION search option of separating values with a comma (IE MySQL OR).  IE

Doing something like "/apis/default/fhir/Patient/_search?name:contains=bob,jane" will return nothing for a two patients named bob and jane whereas a search for /apis/default/fhir/Patient/_search?name:contains=bob and a search for /apis/default/fhir/Patient/_search?name:contains=jane returns the patient resources correctly.

We have a bug in that the StringSearchField is overriding the isAnd property instead of counting the values like the Token,Date,Reference,Number, etc classes do.  We used this to specially handle the patient **:name** search field which maps to 5 inner fields.

adunsulag:
Union (OR) searches were broken when using String search fields.
Fixes #6143 

I'll need to check this fix against Inferno and make sure we didn't break anything.

### DIFF
--- a/src/Services/PatientService.php
+++ b/src/Services/PatientService.php
@@ -45,6 +45,9 @@ class PatientService extends BaseService
      */
     private $patient_picture_fallback_id = -1;
 
+    /**
+     * @var PatientValidator
+     */
     private $patientValidator;
 
     /**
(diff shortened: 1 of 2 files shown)


bradymiller (approved):

adunsulag:
This passes inferno so bringing in.

stephenwaite:
jumping

PR #6154, opened 2023-02-01
chore: Bootstrap-friendly config page and nav

robertdown:
The list of sections is now a pills-based navigation

Before/After
image

image


adunsulag:
Going to close this unless you want to come back and revisit it Robert.

PR #6160, opened 2023-02-01
Fixes #6158,#6159 Portal Credentials Refactor

### LINKED ISSUE #6158: Pull create_portallogin.php portal credentials code into a service class that can be used elsewhere
I have a need to be able to use the portal credential generation code in another location.  The create_portallogin.php code for creating those credentials as well as sending out the patient invitation email should be abstracted into a re-usable class that can be called elsewhere in the codebase.

### LINKED ISSUE #6159: Portal Credentials send email should allow plus sign (+) in email address
The local part (before the @) of the email address should allow a + sign to be in the email to handle things like gmail labels.  For example an email address example+label@gmail.com currently fails in attempting to send the portal credentials email.  We should allow this to work.

adunsulag:
Fixes #6158
Fixes #6159

Pulled the portal credentials into its own class so it can be re-used elsewhere.

Added to the validEmail check function the ability to have a + sign in the email address to support email labels like example+label@gmail.com.

### DIFF
--- a/interface/patient_file/summary/create_portallogin.php
+++ b/interface/patient_file/summary/create_portallogin.php
@@ -24,81 +24,10 @@ require_once("../../globals.php");
 require_once('../../../library/amc.php');
 
 use OpenEMR\Common\ {
-    Auth\AuthHash,
     Csrf\CsrfUtils,
-    Logging\EventAuditLogger,
-    Twig\TwigContainer,
     Utils\RandomGenUtils,
 };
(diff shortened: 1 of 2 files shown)


sjpadgett (approved):
I don't see anything obvious to me.
I'll try to pull down a little later to test

bradymiller (approved):

bradymiller:
dance-bear

PR #6169, opened 2023-02-03
Allow Form Templates to improve documentation efficienty

robertdown:
Starting the PR for the form templates feature.

My approach? While creating the new form, select the form from a list, load the form like it was being loaded in an encounter into an iframe, add a listener to ensure we don't actually submit the form, instead we push it into the templates table.

- [ ] Fix the insert query to capture the template name in the correct place
- [ ] Render active templates
- [ ] Develop the process for actually injecting the template-based data into a new form entry

Plan on supporting some constant replacement expressions for things like dates, times, user, and CSFR tokens. 

Note: Still a draft, wanted to get it pushed up to GH so the community can start to take a look at the process

### DIFF
--- a/interface/modules/custom_modules/oe-form-templates/index.php
+++ b/interface/modules/custom_modules/oe-form-templates/index.php
@@ -1,0 +1,52 @@
+<?php
+/**
+ *
+ */
+
+use Google\Service\Books;
+use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Common\Twig\TwigContainer;
+use OpenEMR\Modules\FormTemplates\Bootstrap;
+use OpenEMR\Modules\FormTemplates\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Twig\Extension\DebugExtension;
+
+require_once "../../../globals.php";
+require_once "../../../../library/registry.inc.php";
+
+// Control access
+if (!AclMain::aclCheckCore('patients', 'demo')) {
+    echo xlt('Not Authorized');
+    exit;
+}
+
+$templateDir = Bootstrap::getTemplatePath();
+$twigContainer = new TwigContainer($templateDir);
+$twig = $twigContainer->getTwig();
+$twig->addExtension(new DebugExtension());
+$twig->enableDebug();
+
+$request = Request::createFromGlobals();
+
+$r_controller = $request->get('controller', 'default');
+$r_action = $request->get('action', 'index');
+
+$routes = [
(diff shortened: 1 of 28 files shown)


robertdown:
Install the module like normal, refresh the page and you'll see a new menu item under Admin called Form Templates.

sjpadgett:
Could this be a replacement for nation note template?

robertdown:
Your comment does present an interesting idea though. I had originally planned on a passive action, click button = actually posting the template. But there could be another path where clicking the button populates the form but doesn't actually submit. This would allow the template to be further customized.

sjpadgett:
I don't know if you've seen my text templates nation note but you dbl click any text area in encounters and it will popup.
Maybe you can use some of that.

robertdown:
Will take a look.

interface/modules/custom_modules/oe-form-templates/index.php
```diff
@@ -0,0 +1,52 @@
+<?php
+/**
+ *
+ */
+
+use Google\Service\Books;
```
adunsulag, line 6: Where is this google service books used?

interface/modules/custom_modules/oe-form-templates/index.php
```diff
@@ -0,0 +1,52 @@
...
+
+$r_controller = $request->get('controller', 'default');
+$r_action = $request->get('action', 'index');
+
+$routes = [
+    'default'       => 'OpenEMR\Modules\FormTemplates\Controller\ConfigurationController',
+    'configuration' => 'OpenEMR\Modules\FormTemplates\Controller\ConfigurationController',
+];
+
+if (array_key_exists($r_controller, $routes) == true) {
```
adunsulag, line 39: I wonder if you could use PHP 8 attributes here to provide yourself some stronger typing here instead of using ReflectionClass for your frontend controller here (I'm referring to the design pattern, not a JS frontend).

adunsulag:
I think I'm going to have to pull this down and try to play with it.  I'm trying to see how this is different than the work that @sjpadgett has done with Questionnaires which allow us to have all kinds of of form templates for collecting patient information.

interface/modules/custom_modules/oe-form-templates/index.php
robertdown: I will definitely look at this - I know using Reflection isn't the most optimized approach, it's kind of my default controller build right now, which I know can be improved.

adunsulag:
Going to close this, feel free to re-open Robert when you or someone else wants to come back to this.

PR #6177, opened 2023-02-04
Fixes #6171 local api scope checks

### LINKED ISSUE #6171: Fix internal local api to not require scope security checks
The internal local API uses the CSRF, session cookies, and the ACL checks to verify API access.  Somehow we broke this in master.  Surprised no one has reported the issue yet.

adunsulag:
Made it so the local api works with the standard and fhir api calls by skipping the scope checks.

Fixes #6171

### DIFF
--- a/src/Common/Http/HttpRestRouteHandler.php
+++ b/src/Common/Http/HttpRestRouteHandler.php
@@ -275,6 +275,7 @@ class HttpRestRouteHandler
             if (
                 $restRequest->getResource() == 'version'
                 || $restRequest->getResource() == 'product'
+                || $restRequest->isLocalApi() // skip security check if its a local api
             ) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


adunsulag:
@bradymiller if this looks fine to you, let me know.  The only way someone can get that CRSF token is if they have an authenticated cookie session right?  That was my only concern with bypassing the scopes here.

bradymiller:
@adunsulag ,
Yes, would need an authenticated session to get the CSRF token (this prevents Cross site request forgery which could be very problematic with the internal write api calls; ie. a user that is already logged into OpenEMR clicks on an bad actor email link that sends them to a bad link that tries to then submit an internal api call that changes database state). Since this changes with every authorized session, is enough protection against that.
Will this then open every fhir/api call? This is ok as long as we ensure we have the acl's set up for all the routes (pretty sure we do, but would make sure on the ones that allow bulk collection of data; ie. those fancy ones that do the data dumps).

bradymiller:
i.e. regarding above acl. We just don't want a receptionist able to do a complete patient/clinic data dump etc?

adunsulag:
@bradymiller All of the bulk export operations require the admin/users ACL.  The $docref requires demo/users, everything else on FHIR has an ACL associated with it as far as I saw.

adunsulag:
@bradymiller To answer your previous question, yes this would open all of the FHIR and standard api's to the internal api session.

bradymiller:
Then good to go :)

bradymiller (approved):

PR #6189, opened 2023-02-08
Fixes #6186 windows file path problems.

### LINKED ISSUE #6186: Module application script filtering is broken on windows hosted installations, Telehealth module has same problem
The ModuleApplication class that filters script and css injection is not handling windows file pathing when OpenEMR is hosted on a windows installation.  

The telehealth module has this problem as well as it's using a file path that is specific to linux/unix platforms and does not work on windows file systems.

adunsulag:
This makes the module file path filtering work on windows and linux.

Fixes #6186


bradymiller (approved):

bradymiller:
dance-bear

PR #6195, opened 2023-02-08
fix: Remove unnecessary non-breaking whitespace if not displaying the field label

### LINKED ISSUE #6194: Extraneous non-breaking whitespace causing UI issue on new patient LBF
Looks like we are injecting a `&nbsp;` if not displaying the label of a field, this is not necessary and is causing a rendering issue on Firefox on Windows. See screenshot below. Gonna drop a PR that just removes the random non breaking character.

Interestingly, this does not occur on the default form, but I was adding to it and noticed the error.

image

robertdown:
Just remove the second half of the condition, no need to display a non-breaking white space in this situation. Fixes #6194

### DIFF
--- a/interface/new/new_comprehensive.php
+++ b/interface/new/new_comprehensive.php
@@ -406,7 +406,7 @@ $constraints = LBF_Validation::generate_validate_constraints("DEM");
                       <td class="text-left align-top">
                     <?php
                     if ($SHORT_FORM) {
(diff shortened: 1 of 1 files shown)


ash1kn1zar (commented):
Reviewed, good for merge

bradymiller (approved):

PR #6209, opened 2023-02-15
chore: only stale-bot issues that lack info for reproduction or are unassigned

### LINKED ISSUE #6201: process: do not automatically close issues as "stale" because they lack activity
### What's wrong

Issues are being marked as "stale" just because they have no activity and/or are not triaged / picked up by a member of the community. 

This doesn't mean that the bug no longer exists, but that there has been no activity (or interest in picking it up to fix). Not everyone is programmer (I am), some are just reporting issues. I believe this is bad practice, and discourages folks from reporting issues which may not get an immediate fix. Worse, users cannot reopen issues. 

### Expected

- We should not close issues as stale just because they lack activity. 
- We should close issues once triaged and are determined to need _more information_ to reproduce but lack follow up or reproduction steps after a certain amount of time

### Suggested workflow

1. Unlabeled/triaged bugs are never marked stale
2. Triaged bugs without sufficient reproduction steps are marked as `NeedsReproductionSteps` or `NeedsMoreInfo`
3. If the issue goes $X number of days with those tags, go ahead mark as stale and close.
4. Users can re-open issues.

### Related

- https://github.com/openemr/openemr/issues/3544
- https://github.com/openemr/openemr/pull/3556
- https://github.com/openemr/openemr/issues/4470
- https://github.com/openemr/openemr/issues/3877#issuecomment-781629962

And many many many more but I don't want to flood the associations.

eabmd:
Fixes #6201

Nota bene: this requires an additional label `WaitingForInfo` when maintainers decide something should be closed without further input. Feels like a fair tradeoff to me. 

... Hopefully this is the winner. Multi-user git is hard.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -3,6 +3,11 @@
 daysUntilStale: 90
 # Number of days of inactivity before a stale issue is closed
 daysUntilClose: 7
+# Only close triaged issues that lack reproduction / clear next steps
+onlyLabels:
+    - WaitingForInfo
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: true
 # Issues with these labels will never be considered stale
 exemptLabels:
   - pinned


bradymiller (approved):

bradymiller:
@eabmd , looks good and thanks for doing this. will wait a bit to ensure no other input.

bradymiller:
@eabmd , added the WaitingForInfo label and bringing this in. Will see how it works out. Thanks for the improvement!

PR #6231, opened 2023-02-22
fix: Ensure a Facility ID is always sent to a new encounter save event

### LINKED ISSUE #6230: Error Creating New Encounter if "Facility" is hidden
If you try to create/save a new encounter and we have the "Facility" field hidden, we'll get a sql Insert error, saying that the facility_id can't be null.
I've tested it in the online demo on 7.0.0 (2) version.

 To Reproduce this behavior  we simply have to:
         1. Go to Global Settings -> Encounter Form
         2. Set "Show Facility option on Encounters" to "Hide on New and Edit Form" and Save
         3. Select a Patient
         4. New Encounter and fill the fields
         5. Save

1 - Global Settings
2 - New Encounter
3 - Error saving

What we should expected is a successful creation, but instead the sql error will appear.

I check that the value of the field is always with a default value, so this shouldn't happen. 
But has I analyzed it the problem is that if there are fields (or div parents) hidden through 'd-none' class those fields on submit will not passthrough, that's because the 'd-none' class sets the 'display' property to 'none'. 

I think we can use the "visibility:hidden" property instead, but I didn't tested it out.

### Client configuration
- Browser: Edge
- OpenEMR version: 7.0.0(2)
It was on the online demo

robertdown:
Fixes #6230

### DIFF
--- a/interface/forms/newpatient/common.php
+++ b/interface/forms/newpatient/common.php
@@ -41,6 +41,8 @@ $years = array($thisyear - 1, $thisyear, $thisyear + 1, $thisyear + 2);
 
 $mode = (!empty($_GET['mode'])) ? $_GET['mode'] : null;
 
+$viewmode = false;
+
 // "followup" mode is relevant when enable follow up encounters global is enabled
 // it allows the user to duplicate past encounter and connect between the two
 // under this mode the facility and the visit category will be same as the origin and in readonly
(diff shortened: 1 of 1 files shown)


interface/forms/newpatient/common.php
```diff
@@ -84,6 +86,52 @@
 
 $displayMode = ($viewmode && $mode !== "followup") ? "edit" : "new";
 
+function getDefinedFacility()
+{
+    global $viewmode;
+    global $default_fac_override;
+    global $care_team_facility;
+    global $GLOBALS;
```
bradymiller, line 94: fyi, $GLOBALS is always a global (implied)

bradymiller:
thanks for the fix!

2DV(1)

PR #6236, opened 2023-02-25
fix: Silently return if logo cannot be found

### LINKED ISSUE #6235: Fatal Error with logos service on main tabs
### Describe the bug

```
PHP Fatal error:  Uncaught TypeError: OpenEMR\\Services\\LogoService::convertToWebPath():
Argument #1 ($path) must be of type string, null given, called in /var/www/localhost/htdocs/openemr/src/Services/LogoService.php on line 89 and defined in /var/www/localhost/htdocs/openemr/src/Services/LogoService.php:98
Stack trace:
#0 /var/www/localhost/htdocs/openemr/src/Services/LogoService.php(89): OpenEMR\\Services\\LogoService->convertToWebPath()
#1 /var/www/localhost/htdocs/openemr/interface/main/tabs/main.php(30): OpenEMR\\Services\\LogoService->getLogo()
#2 {main} thrown in /var/www/localhost/htdocs/openemr/src/Services/LogoService.php on line 98, referer: http://localhost:8300/interface/login/login.php?site=default`
```

robertdown:
Fixes #6235


robertdown:
The public images fallback image name wasn't correct, it was `menu-logo.svg` instead of `logo.svg`. Further, `findLogo()` could return `null` if the logo could not be found, causing an issue upstream. Now, `findLogo()` will always return a string - if the logo cannot be found, it will just return an empty string to allow a silent error out.

bradymiller:
2DV-2(13)

PR #6261, opened 2023-03-04
Bug fix flow board status change

### LINKED ISSUE #6260: Past flow board entries disappear
The issue is when the appointment status is set to inactive. The flow board will no longer show that appointment when reviewing the flow board in the past that had the now deactivated status. The flow board will not show an appointment that had been completed in the past.

juggernautsei:
Fixes #6260

https://youtu.be/ELIoXcRyTho
Reproduced on the main demo site.

past appointments disappear from the flow board if the status is made inactive. 

change the query to remove the active flag.


adunsulag:
Hey @juggernautsei were you meaning to bring in commits from @robertdown as part of this PR?

Also, you've checked the patient flow board report as well with this change?  That was the only other place I saw that function call being used.  Otherwise it looks fine to me.

robertdown:
@adunsulag That's probably an upstream merge to catch his working branch up to master

robertdown:
I think we should consider the implications of this. If we inactivate an appointment status, it should really disable it everywhere. I understand that if disabled, there's theoretically no way to create a new appointment with the status, so no _new_ events would show up on the flow board. However, this would cause the flow board to show information a user has explicitly stated they don't want shown.

I am not convinced current status is actually incorrect. I understand what you're trying to achieve, but, like you mentioned on the call, this solves a problem only introduced because someone is trying to do something with the flow board that the flow board was not intended to do.

robertdown (changes requested):
Not sure I want to introduce this as it results in a page no longer respecting the activity status of a list item

juggernautsei:
Robert, the end user does not think of the status as something they don't want showing on the flow board. They are looking at the active and inactive only in the light of the appointment scheduler. It is the past appointments that have that status that is being wiped off the flow board so when they go back and look at the flow board to see the statuses for the past time period chosen. They can see what is the status of the appointments and take the appropriate action. The color coding makes it easier than using the appointments report that has no color coding.  Even though the EU does not think that going back in time to look at the flow board status is not pulling a report. It essential is for a report of sorts that helps them to navigate their day. 
As @adunsulag pointed out, this function is not called anywhere else. It does not impact anything but the past history of the flow board. The appointment scheduler will function the proper way and only show the selectable statuses.

juggernautsei:
@adunsulag No I did not mean for all those other commits to come over. 
I have cleaned my master branch.

bradymiller:
hi @juggernautsei , What is the workflow to making an appt inactive (I don't see that as a selection in the status menu)?

juggernautsei:
@bradymiller there is no workflow to make an appointment inactive. This is about the appointment statuses.
Go to Lists -> Appointment Statuses. There is a checkbox column labeled Active. That is where the status can be set to active or inactive. As the EU adds or removes status types. The EU manages which appointment type show in the Status in the appointment scheduler.

juggernautsei:
@robertdown I have come across others that use the flow board to manage if providers and staff have done their job by marking the appointments in the appropriate status. 
Like it or not. The way people use a feature has to take president over the original design of the feature. The feature should change to fit the need of the EU and not constrain them to the past idea.

robertdown:
A middle ground would be a toggle on the flow board which indicates whether inactive statuses should be considered. It would need to be on the flow board itself so that users can toggle it ad-hoc. The state of the toggle should always default to Active Only.

I just don't think it's a good choice to introduce a query that ignores the active setting defined by the implementor.

juggernautsei:
@robertdown I can agree to make it a toggle. The toggle should come in the form of a checkbox. That bit can be used to toggle the query 

if ($bit) {
    $query .= "active = 1";
}

We are going to need some documentation for the checkbox.

robertdown:
Please don't introduce a global option for this, the active indicator should come straight from the Flow Board during the HTTP request.

juggernautsei:
@robertdown I agree this is not going in the globals

juggernautsei:
I decided not to introduce this into the community. It is not a wide spread complaint.

PR #6310, opened 2023-03-29
Fixes #6292, #6303 localization bug vitals unit conv.

### LINKED ISSUE #6292: Vitals form metric conversion fails for non-english languages
There is a bug in the vitals form for non-english languages when calculating the conversion values.  If a language translation for the unit exists, the translated unit used in the metric conversion does not match the english value in the codebase and the conversion fails.  

The code should have a separate value for the unit and the translated unit name to fix this problem.

### LINKED ISSUE #6303: Vitals weight and height can´t be recorded if my local settings are diffent than USA
### Describe the bug
I can´t save weight in pounds or kg, or height in inches or centimeters if I don´t set English(standard) as my language when logging in.
I am testing Spanish (Latin America)
The function convInputElement inside vitals.js in line 77 is trying to save value instead of convValue.

### To Reproduce
Login with Spanish (Latin America) as language, then go to "Paciente" in top menu, then you have to click on Búsqueda, then select any patient, then select any encounter, then you select Clinico and then Vitales (to set a vital signs assessment), then when you type any weight or height, you get in the console "Failed to get valid number for input with id  weight_input_usa  with value  <number you typed>" 
image

### Expected behavior

Let you input weight or height and convert it to US or metrics.

### Client configuration

- Browser: Firefox 110.0
- OpenEMR version: 7.0.1-dev

adunsulag:
Had a bug where the localization of the vitals unit was throwing things off.  Broke out the label from the unit translation which fixes this for international users.

Fixes #6303 
Fixes #6292

### DIFF
--- a/interface/forms/vitals/C_FormVitals.class.php
+++ b/interface/forms/vitals/C_FormVitals.class.php
@@ -105,8 +105,10 @@ class C_FormVitals
                 // eventually we could just grab the raw values...
                 ,'vitalsValue' => "get_weight"
                 ,'vitalsValueMetric' => "get_weight_metric"
-                ,'unit' => xl('lbs')
-                ,'unitMetric' => xl('kg')
+                ,'unit' => 'lbs'
+                ,'unitMetric' => 'kg'
+                ,'unitLabel' => xl('lbs')
+                ,'unitMetricLabel' => xl('kg')
                 ,'precision' => 2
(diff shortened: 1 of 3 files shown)


stephenwaite:
jumping

bradymiller (approved):

bradymiller:
2DV-2(8)

PR #2138, opened 2019-01-17
Dropping pma tables

morg0n:
Deleted pma_ tables

#651

PR #6339, opened 2023-04-08
Dropping pma tables

mdsupport:
Supplements commit #2138 for old installations.


adunsulag:
Looks fine to me!

sql/patch.sql
```diff
@@ -46,4 +46,32 @@
...
+DROP TABLE IF EXISTS `pma_relation`;
+#EndIf
+
+#IfTable pma_table_coords
+DROP TABLE IF EXISTS `pma_table_coords`;
+#EndIf
+
+#IfTable pma_table_info
+DROP TABLE IF EXISTS `pma_table_info`;
+#EndIf
```
bradymiller, line 77: hi @mdsupport , This code should instead be placed in the sql/7_0_1-to-7_0_2_upgrade.sql script
(then would bring it in)

bradymiller:
looks good and bringing in. thanks for the fix!

PR #6340, opened 2023-04-10
General onetime service

### LINKED ISSUE #6341: Develop an authenticated onetime token service
### Describe the problem
The portal is already using onetime tokens for reset password and new registration notification.
However there is a growing need to use onetime elsewhere such as assessment/questionnaire apps and Telehealth.
Will create a service to create and decode tokens to be available from anywhere in core.
Options will include a pin, embedded redirects and ability to create token to enable on a future date or immediately.

sjpadgett:
- new onetime service with redirect

Fixes #6341 

Phase one to create a workflow for using onetime tokens
Also to help start cleaning up portal onetimes for reset and register.

Demo for this PR is at:
https://www.open-emr.org/wiki/index.php/Development_Demo#Gamma_-_Up_For_Grabs_Demo


sjpadgett:
@adunsulag 
This is a beginning with a short stub in portal for initial testing.
Had trouble with module setup to continue integration.
Give me a call tomorrow to discuss.

portal/index.php
```diff
@@ -62,6 +62,22 @@
     unset($_SESSION['password_update']);
 }
 
+if (!empty($_GET['service_auth'] ?? null)) {
+    $token = $_GET['service_auth'];
+    $redirect_token = $_GET['target'] ?? null;
+    $oneTime = new OneTimeAuth();
+    $do_actions = $oneTime->decodePortalOneTime($token, $redirect_token);
+    if (!empty($do_actions['error'])) {
+        (new SystemLogger())->debug("Failed " . $do_actions['error']);
```
adunsulag, line 71: probably want to use (new SystemLogger())->error() instead of debug here.

src/Common/Auth/OneTimeAuth.php
```diff
@@ -0,0 +1,177 @@
...
+        $token_raw = RandomGenUtils::createUniqueToken(32);
+        $pin = RandomGenUtils::createUniqueToken(6);
+        $token_encrypt = (new CryptoGen())->encryptStandard($token_raw);
+        $token_database = $token_raw . $pin . bin2hex($expiry->format('U'));
+        if (!empty($p['pid']) && !empty($token_raw)) {
+            $query_parameters = [$token_database, $p['pid']];
+            sqlStatementNoLog("UPDATE `patient_access_onsite` SET `portal_onetime` = ? WHERE `pid` = ?", $query_parameters);
+            (new SystemLogger())->debug("New onetime token added in database.");
+        } else {
+            (new SystemLogger())->debug("Onetime failed missing PID or token creation failed");
```
adunsulag, line 60: Since the function already throws exceptions, would it be better to just throw an InvalidArgumentException here?

src/Common/Auth/OneTimeAuth.php
```diff
@@ -0,0 +1,177 @@
...
+        // scope = portal/service tasks (reset, register). context = portal, patient etc.
+        $this->scope = $scope;
+        $this->context = $context;
+    }
+
+    /**
+     * @param array $p [
+     *                 'pid' => '', // required for most onetime auth
+     *                 'target_link' => '', // Onetime endpoint
+     *                 'redirect_link' => '', // Where to redirect the user after auth
```
adunsulag, line 38: Where does redirect_link get used?  I didn't see it used anywhere.

sjpadgett:
I keep going back and forth on creating a onetime token table. It would simplify workflow in regards to just sending a token and look up necessary redirects and/or ancillary data instead of encrypting in query.
Still just using current portal table is easier to implement last minute before a release and I feel it is more secure as it would be harder to hack all the encryption.

I created the new service to be compatible with our existing portal onetime so we can cleanup in the index, create an account class and generally smooth out portal login. I wish I had the time before release to do this now as it would need a day or to of testing to ensure I haven't broken anything.

adunsulag:
@sjpadgett Can you check in the table creation for the sql upgrade script?  Was testing out with the new table stuff but it looks like this didn't get checked in.

sjpadgett:
working on now. need a few minutes.

sjpadgett:
@adunsulag done

adunsulag:
@sjpadgett k I'll try it out in 20-30 minutes once the kids give me a break lol.

src/Common/Auth/OneTimeAuth.php
```diff
@@ -0,0 +1,340 @@
...
+                'site' => $site_id
+            ]));
+        } elseif ($this->scope == 'reset_password') {
+            $encoded_link = sprintf($format, attr($site_addr), http_build_query([
+                'forward' => $token_encrypt,
+                'site' => $site_id
+            ]));
+        } else {
+            $encoded_link = sprintf($format, attr($site_addr), http_build_query([
+                'id' => $pid,
```
adunsulag, line 232: Do we need to have the pid in this link?  Not sure why but my gut says we don't want to expose the patient pid as part of the link.

sjpadgett: no, I'll remove

src/Common/Auth/OneTimeAuth.php
```diff
@@ -0,0 +1,340 @@
...
+class OneTimeAuth
+{
+    private $scope;
+    private $context;
+
+    public function __construct($context = 'portal', $scope = 'redirect')
+    {
+        // scope = portal/service tasks (reset, register). context = portal, patient etc.
+        $this->scope = $scope;
+        $this->context = $context;
```
bradymiller, line 33: Something that can increase performance is creating the CryptoGen instance in __construct and storing it as a private variable (this instance will cache the keys as they are used so it doesn't then need to keep collecting and unencrypting the key from drive if done more than once)

src/Common/Auth/OneTimeAuth.php
```diff
@@ -0,0 +1,340 @@
...
+            $site_addr = trim($GLOBALS['portal_onsite_two_address']);
+        }
+
+        $rtn['encoded_link'] = $this->encodeLink($site_addr, $token_encrypt, $passed_in_pid, $redirect_token);
+        $rtn['onetime_token'] = $token_encrypt;
+        $rtn['redirect_token'] = $redirect_token;
+        $rtn['pin'] = $pin;
+        $rtn['email'] = $email;
+
+        $save = $this->insertOnetime($passed_in_pid, $pin, $token_raw, $redirect_raw, $redirect_token, $expiry->format('U'));
```
bradymiller, line 89: Does the redirect_token (encrypted) need to be stored? Would be best if this was not stored and the info was instead collected (and perhaps comfirmed with stored redirect_raw) when decrypting the token.

sjpadgett: agreed

src/Common/Auth/OneTimeAuth.php
sjpadgett: I usually do this but have been lazy. thanks for suggestion.

sjpadgett:
@adunsulag 
I've changed the onetime, by default, to not encrypt and include in link (target query). You may include by  setting flag in `createPortalOneTime(array $p = [], $encrypt_redirect = false)` method signature.
Since I now use a separate table to track onetimes I feel including in link is clutter and serves no purpose.
Let me know if you disagree....

adunsulag:
@sjpadgett Have you pushed that change?  I don't see it.  I'm getting logged in now just fine.  Had to add a few more variables we were missing from the session but overall everything is working when I log.  

I realized that I do give users the option to go back to their patient portal when they close their session.  Hitting a really wierd bug where when they come back to the portal the initial session is there everything works.  When I make ajax requests the session is now empty.  So strange.  Not sure if there's some kind of concurrency bug or something going on.

adunsulag:
@sjpadgett Also, does it make sense to have an option of max usage?  I noticed that the link can be used multiple times which I think can be a nice feature but some people may want to restrict it to just a single use, or to 2 (if dealing with email scanners) or 5, etc.

sjpadgett:
I'll give a call in about 10min

interface/modules/custom_modules/oe-module-claimrev-connect/src/GlobalConfig.php
```diff
@@ -36,6 +36,7 @@ class GlobalConfig
     public const CONFIG_ENABLE_RESULTS_ELIGIBILITY = "oe_claimrev_eligibility_results_age";
     public const CONFIG_ENABLE_AUTO_SEND_ELIGIBILITY = "oe_claimrev_send_eligibility";
     public const CONFIG_X12_PARTNER_NAME = "oe_claimrev_x12_partner_name";
+
```
adunsulag, line 39: Not sure how I ended up adding a space here... oh well.

sjpadgett:
Need to track back where window pathing slashes are being used to build redirect.
`https://opensourcedemr.us/interface/modules/custom_modules/oe-module-comlink-telehealth\public\index-portal.php?action=launch_patient_session&pc_eid=32`
While browser will cope with them php location header doesn't.

I may validate the redirect address in ontime save but prob should fix at the source.

sjpadgett:
@adunsulag I'm running into the session not restoring when I hit the index-portal from email join click.
Same issue you were having which I thought you said you found and fixed.

sjpadgett:
I fixed session issue but the back to portal url is corrupted if the portal globals site address includes the site=??? resulting in:
`https://opensourcedemr.us/portal/index.php?site=default/home.php`

sjpadgett:
So all working now except back to portal dashboard and I still need to track down cross origin connecting to vendor gateway for video/audio.
I may, if time permits, add methods to class to log into portal with onetime and sms/email pin confirm.

interface/modules/custom_modules/oe-module-comlink-telehealth/public/index-portal.php
```diff
@@ -24,7 +24,11 @@
 // need to retain the webroot if we have one
 $landingpage = $basePath . "portal/index.php?site=" . urlencode($_GET['site_id'] ?? '') . "&redirect=" . urlencode($redirect);
 $skipLandingPageError = true;
-
+// test session
+if (!empty($_REQUEST['me'] ?? null) && empty($_SESSION ?? null)) {
+    session_id($_REQUEST['me']);
+    session_start();
+}
```
bradymiller, line 31: is this needed? Would be ideal to avoid passing session ids in request parameters.
(the code in verify_session should capture a portal session, so odd that this is needed)

sjpadgett: agreed @bradymiller I don't know if this is a window issue or a browser issue. Currently testing with chrome engine.
The verify session doesn't work. I had this same issue with the CCDA rest call from portal. This was the solution and has been in code for at least 4 years. Ran into same issue with OAuth but there I needed to store the session in database because of a different origin for request.
It is not unusual to use session id in session management. Still, maybe will encrypt id.
 or store to database to be more comfortable.
Since this is a same origin request the session should be intact or maybe something in my PHP settings.
Will get Stephen to verify he doesn't see the session on his environment.

adunsulag: I'm not seeing this problem anymore with me fixes from yesterday on linux on Chrome.  I'll test this on firefox.  Perhaps its something on Windows and I can try booting up a VM to test this out.

sjpadgett: Let me fire up a docker and test first and maybe will save some time.

portal/index.php
```diff
@@ -62,6 +67,46 @@
...
+        exit();
+    } catch (OneTimeAuthExpiredException $exception) {
+        $logit->portalLog(
+            'onetime login attempt',
+            $exception->getPid() ?? '',
+            ':invalid one time',
+            '',
+            '0'
+        );
+        // do we want a separate message that their token has expired?
```
bradymiller, line 92: separate message in the logging engine is always helpful for debugging purposes

src/Common/Auth/OneTimeAuth.php
```diff
@@ -0,0 +1,388 @@
...
+use OpenEMR\Common\Auth\Exception\OneTimeAuthExpiredException;
+use OpenEMR\Common\Crypto\CryptoGen;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Logging\SystemLogger;
+use OpenEMR\Common\Session\SessionUtil;
+use OpenEMR\Common\Utils\RandomGenUtils;
+use OpenEMR\Services\PatientService;
+use OpenEMR\Services\UserService;
+
+class OneTimeAuth
```
bradymiller, line 29: Beautiful class!

2DV

sjpadgett:
@adunsulag 
Okay I'm up to date and all is working. I'm now doing
- validate and auto create the default portal address from server/proxy supers(like I did for API configs) in config. This url structure needs to be predictable.
- validate/repair uri/url ending slashes.
- either encrypt or save to table session id for passing workflows forward.
- ensure that class exception messages trickle up to user UI.
- 
I've got to say that this is a first class module and great job Stephen! I'm betting it will be a big hit in the community.
Now you're going to need to top this for our next release so get on it!:)

adunsulag:
@sjpadgett just pushed a PR of my changes to your onetime PR.  Hopefully we didn't double up on any work and it merges in nicely with your stuff. I did pull your latest changes into my onetime to make sure it worked.  Gonna take my daughter out for a birthday lunch but should be available in an hour or so.

sjpadgett:
reminder to talk about
```
// note the cache buster is already being populated in Header.php since this script isn't a registered
 // asset
```

not working!

bradymiller:
Demo for this PR is at:
https://www.open-emr.org/wiki/index.php/Development_Demo#Gamma_-_Up_For_Grabs_Demo

(@sjpadgett , let me know when you want the demo updated; it is very tough to do, gotta click a button :) 
image
)

sjpadgett:
very cool. it'd be a couple hours and thanks.

src/Common/Auth/OneTimeAuth.php
```diff
@@ -0,0 +1,388 @@
...
+        $_SESSION['ptName'] = $patient['fname'] . ' ' . $patient['lname'];
+        // never set authUserID though authUser is used for ACL!
+        $_SESSION['authUser'] = 'portal-user';
+        $_SESSION['portal_username'] = $auth['username']; // required by portal/handle_note.php
+        $_SESSION['portal_login_username'] = $auth['login_username']; // required by portal/handle_note.php
+        // Set up the csrf private_key (for the patient portal)
+        //  Note this key always remains private and never leaves server session. It is used to create
+        //  the csrf tokens.
+        CsrfUtils::setupCsrfKey();
+        $auth['redirect'] .= "&me=" . session_id();
```
bradymiller, line 383: what is an example value for `$auth['redirect']` here? (especially interested in the value you use here that is breaking on windows)

sjpadgett: `
https://opensourcedemr.us/interface/modules/custom_modules/oe-module-comlink-telehealth/public/index-portal.php?action=launch_patient_session&pc_eid=35
`

sjpadgett:
adding alert messages for onetime fail
image

sjpadgett:
@bradymiller 
If I copy/paste the link into new chrome window it will work however when clicking from my gmail email client then it breaks.
Weird to say the least. I have click tracking turn off on host.

Give me a call if you want to go into this deeper. Too much to type out...

Putting up all current changes in a few minutes. session id will be disabled

sjpadgett:
I still have some small items I want to address but may not be until tomorrow.
Will work on logo a bit but I'm running out of steam and may not get too far!!!

Brady is this demo reset something I can do?

bradymiller:
just reset the demo
I'll email you later tonight with credentials to demo farm gui, so can reset when needed

sjpadgett:
@adunsulag Setup demo but won't show session icon in calendar. All creds setup so beats me.
I suppose I need to do a new install local. Demo server won't show debug in log.

bradymiller:
Something it could be (and a nice thing about the demos) is unmasking of a path issue (since demos have webroot set). This can also be tested on the easy dev demos with the following `openemr-cmd` commands:
```
webroot-management:
  cwb, change-webroot-blank              Change webroot to be blank (this is default setting of environment)
  cwo, change-webroot-openemr        Change webroot to be openemr
```

sjpadgett:
@bradymiller 
Still won't restore session on easy-dev docker. Works fine if I copy paste the onetime from email to browser.
I can only conclude it is the google client as the cause. In the end the issue leads me to conclude adding the session id for just these cases. 
I know I could hack it sending in a super in the raw so I will either save it to a file or the onetime table. I believe I can include the sql api in verify_session or the index-portal endpoint.
Unless you guys have a better idea.

sjpadgett:
I need to still validate and reformat the portal address from the global.
I'll do separate commits for it and then one for session id.
I'm also finding an issue when using local host and need clarification of the openemr UUID use by video address.

sjpadgett:
going to merge so Stephen can pick it up to continue additions and add a fix for linking out of google client causing a cross site problem for session restore.

sjpadgett:
over to you @adunsulag

bradymiller:
tenor99

PR #6345, opened 2023-04-10
Fixes #6344 API improvements

### LINKED ISSUE #6344: API improvements and origin check fix
In Chrome you cannot have both a Access-Control-Allow-Origin wildcard '*" while also using the Access-Control-Allow-Credentials header. In the HTTPRestRouteHandler we need to actually respond with the correct origin for the header.  Since we already validate the headers using our JWT access token we should respond with the requesting origin header.

If we wanted to be more secure we could cross-validate this against our client app redirect_uris, but I think that's probably overkill right now.

adunsulag:
Fixes #6344 
Centralized the key location in the RestConfig class into the ServerConfig class.

Added some helper methods to the AccessTokenEntity for whether the token is revoked or expired.

Added better error logging in the validate redirect uri so that api consumers can see why their client api request fails if the redirect_uri does not match in the authorization_code grant.

Fixed the chrome CORS security complaint about the Access-Control-Allow-Credentials  being used alongside a wildcard Access-Control-Allow-Origin header.  I use the passed along origin since we've already validated at this point the AccessToken and verified it against a redirect_uri of the client.

Added a helper method for better type safety checking w/ the RestApiScopeEvent.

### DIFF
--- a/src/Common/Auth/OpenIDConnect/Entities/AccessTokenEntity.php
+++ b/src/Common/Auth/OpenIDConnect/Entities/AccessTokenEntity.php
@@ -19,14 +19,14 @@ use League\OAuth2\Server\Entities\Traits\TokenEntityTrait;
 
 class AccessTokenEntity implements AccessTokenEntityInterface
 {
+    use AccessTokenTrait;
+    use TokenEntityTrait;
+    use EntityTrait;
+
     /**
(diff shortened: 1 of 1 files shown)


bradymiller (approved):

bradymiller:
tenor99

PR #6370, opened 2023-04-17
Fix first party cookie redirect with post submit

adunsulag:
We had a bug in gmail and other browser based email programs trying to use the one time login which would have their cookies blocked and the sessions would be lost.

Changed up the process so it takes the sso request and then submits it as a post request automatically via javascript.  User has a button to do the submission as well manually if anything goes wrong.

Also added CSRF protection to this.

### DIFF
--- a/portal/index.php
+++ b/portal/index.php
@@ -75,7 +75,6 @@ if (isset($_GET['woops'])) {
  * and compared to portal credential account id lookup.
  * */
 if (!empty($_REQUEST['service_auth'] ?? null)) {
-
     if (!empty($_GET['service_auth'] ?? null)) {
         // we have to setup the csrf key to preven CSRF Login attacks
         // we also implement this mechanism in order to handle Same-Site cookie blocking when being referred by
@@ -92,8 +91,7 @@ if (!empty($_REQUEST['service_auth'] ?? null)) {
             'images_static_relative' => $GLOBALS['images_static_relative'] ?? ''
         ]);
         exit;
-    }
-    else if (!empty($_POST['service_auth'] ?? null)) {
+    } else if (!empty($_POST['service_auth'] ?? null)) {
         $token = $_POST['service_auth'];
         $redirect_token = $_POST['target'] ?? null;
         $csrfToken = $_POST['csrf_token'] ?? null;
--- a/portal/index.php
+++ b/portal/index.php
@@ -45,6 +45,7 @@ use OpenEMR\Core\Header;
 use OpenEMR\Services\LogoService;
 use OpenEMR\Common\Auth\Exception\OneTimeAuthExpiredException;
 use OpenEMR\Common\Auth\Exception\OneTimeAuthException;
+use OpenEMR\Common\Twig\TwigContainer;
 
 //For redirect if the site on session does not match
 $landingpage = "index.php?site=" . urlencode($_SESSION['site_id']);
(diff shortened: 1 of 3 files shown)


adunsulag:
wierd my local phpcbf didn't find the style errors. I'll have to see if there is an update to the tools or something.

sjpadgett:
you might have missed my email but I suggested this is easier by using below in processOnetime methad.
```
echo '<script>window.location.href="' . $auth['redirect'] . '";</script>';
echo '<noscript><meta http-equiv="refresh" content="0;url=' . $auth['redirect'] . '" /></noscript>';
```

adunsulag:
@sjpadgett you've tested that and it works in gmail? How does it look?  Does the user notice the white flash of content?  I threw the twig message up to let the user know something is happening while the browser is submitting the post.  I guess we could do the same kind of display there in the OneAuth for higher latency environments.

sjpadgett:
yes tested well.
The only thing I noted was if the domain didn't have a valid cert then user would get the request to allow the xfer.
Your way is fine and may be a better user experience. I tried to catch you with suggestion before you got too far into dev your solution.
We can go with this....

portal/index.php
```diff
@@ -73,36 +74,63 @@
...
+            'action' => $GLOBALS['web_root'] . '/portal/index.php',
+            'service_auth' => $_GET['service_auth'],
+            'target' => $_GET['target'] ?? null,
+            'csrf_token' => CsrfUtils::collectCsrfToken('autologin'),
+            'pagetitle' => xl("OpenEMR Patient Portal"),
+            'images_static_relative' => $GLOBALS['images_static_relative'] ?? ''
+        ]);
+        exit;
+    }
+    else if (!empty($_POST['service_auth'] ?? null)) {
```
sjpadgett, line 96: Isn't this always true and valid because any POST will be in REQUEST!

adunsulag: It gets caught in the first if statement with the $_GET request.

adunsulag:
> tried to catch you with suggestion before you got too far into dev your solution.

well, that's what happens when freak events occur and I get knocked off the internet for 3 hours lol.

portal/index.php
sjpadgett: oh shoot sorry, I missed this is in inner conditional. Don't pay attention to the man behind the curtain!:)

adunsulag: I don't know... the man behind the curtain ultimately gets you back to Kansas...

portal/index.php
```diff
@@ -73,36 +74,61 @@
...
+                ':invalid one time',
+                '',
+                '0'
+            );
+            OpenEMR\Common\Session\SessionUtil::portalSessionCookieDestroy();
+            header('Location: ' . $landingpage . '&oi');
+            exit();
+        }
+    } else {
+        // TODO: @adunsulag need to handle an invalid request method... should never get here
```
bradymiller, line 131: prob just exit here to be safe. i think the only thing that could get here is if there was a `$_COOKIE['service_auth']` values

templates/portal/login/autologin.html.twig
```diff
@@ -0,0 +1,54 @@
...
+    <div class="card overflow-auto text-center">
+        <header class="card-header bg-primary text-light">
+            <h1>{{ 'Authenticating Login Session' | xlt }}</h1>
+        </header>
+        <h5 class="m-5">{{ "Please wait while we authenticate your session"|xlt }}</h5>
+        <form method="POST" id="autologinform" action="{{ action }}">
+            <input type="hidden" name="csrf_token" value="{{ csrf_token|attr }}" />
+            <input type="hidden" name="service_auth" value="{{ service_auth|attr }}" />
+            <input type="hidden" name="target" value="{{ target|attr_url }}" />
+            <input class="btn btn-lg btn-primary" type="submit" name="autoLoginSubmit" value="{{ "Go To Destination"|attr }}" />
```
bradymiller, line 37: xla for "Go To Destination"

templates/portal/login/autologin.html.twig
```diff
@@ -0,0 +1,54 @@
...
+        </form>
+        <p>{{ "If you have not been redirected in a few seconds click the button above" }}</p>
```
bradymiller, line 39: xlt above line

bradymiller:
very nice! only noted couple very minor issues.

bradymiller:
oops, sorry about the close/open

portal/index.php
adunsulag: Fixed.

templates/portal/login/autologin.html.twig
adunsulag: Fixed.

PR #6377, opened 2023-04-18
Fixes #6376 verify existing users check

### LINKED ISSUE #6376: Telehealth Verify Settings Existing User Check
There's a bug if the telehealth settings have changed (username / CMSID) and there is already a provisioned user in the comlink telehealth system.  It will skip the validation even though the username / cmsid are now incorrect.  This should rarely if ever happen but I triggered it accidently in my testing of other stuff.  

We need to do a check on the existing user.  Since there isn't a comlink api mechanism to do that right now we should accomplish the same thing by suspending the currently provisioned user and then activating the user.  This will treat it like a newly provisioned user request and verify that the credentials are working properly.

adunsulag:
Need to check for config changes with existing provisioned users. Fixes #6376


sjpadgett (approved):
I don't see anything jumping out at me. Good to go!

PR #6385, opened 2023-04-21
Fix encounter location to provide facility data

### LINKED ISSUE #6384: Fix Inferno Encounter.location.location MustSupport check
The Inferno test suite has changed up the encounter location checks once gain.  We need to figure out why the test is not passing and fix the issues with the location check.  This needs to be done for the 7.0.1 release.

image

adunsulag:
Fixes #6384 

The facility data was being excluded in a patient bound context.  If the access token is bound to a patient context it will return the patient location data when requesting a specific facility location id.  This was due to the OR clause in the patient bound search.

Fixed up the search query binding so patients get their own data or the facility data but it still prevents access to other patient data.


bradymiller (approved):

bradymiller:
tenor99

PR #6390, opened 2023-04-21
Bug openemr fix 6389 telehealth bugs

### LINKED ISSUE #6389: bug: Telehealth fix bugs, onetime auth, suspend/resume, verify settings check
Have a bug where the OneTimeAuth service's isValidPatient fails if the patient has changed their name.  Removing the check that forces the account to be the same as the patient's first name + id fixes this problem.  We should just check to make sure the portal username is not empty.

If you suspend a telehealth user and then resume them, the user stays suspended.  Need to fix this problem.

The verify settings check always returns as invalid if the Telehealth Login link expiration setting has not been saved.

The verify settings check fails if the third party invitations has not been setup.  

If the verify settings check fails and then succeeds the success message is displayed in red when it should be displayed in the success message color.

adunsulag:
Fixes #6389 

Fix telehealth provisioning resume to resume the user not suspend it.

Add debug logging for telehealth resume, suspend, and deactivate actions
in provisioning service.

Add more detailed messages on the verifier to detail whether the config
is not saved vs the portal is not setup.

Made the login time limit setting an optional setting so it doesn't
break the configuration setting.

Fixed a minor issue where if the settings fail validation and then
succeed afterwards, the text color is set to be a success message
instead of a failure.

If the patient's first name changed, isValidPatient would fail.  Fixed
the check so we just verify that there is a valid username and don't try
to match the patient first name with the portal username.

### DIFF
--- a/interface/modules/custom_modules/oe-module-comlink-telehealth/src/Controller/TeleHealthVideoRegistrationController.php
+++ b/interface/modules/custom_modules/oe-module-comlink-telehealth/src/Controller/TeleHealthVideoRegistrationController.php
@@ -281,7 +281,7 @@ class TeleHealthVideoRegistrationController
 
     public function resumeUser(string $username, string $password): bool
     {
-        return $this->remoteService->suspendUser($username, $password);
+        return $this->remoteService->resumeUser($username, $password);
     }
 
     public function deactivateUser(string $username, string $password)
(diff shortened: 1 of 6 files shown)


bradymiller (approved):

bradymiller:
2DV

PR #6393, opened 2023-04-22
Rel 701 backport 6389 6384

### LINKED ISSUE #6384: Fix Inferno Encounter.location.location MustSupport check

### LINKED ISSUE #6389: bug: Telehealth fix bugs, onetime auth, suspend/resume, verify settings check

adunsulag:
Fixes #6389 in rel-701
Fixes #6384 in rel-701

Had to cherry-pick the individual commits because I did a merge into master on the PRs instead of squash and merge like I should have done.

### DIFF
--- a/interface/modules/custom_modules/oe-module-comlink-telehealth/src/Controller/TeleHealthVideoRegistrationController.php
+++ b/interface/modules/custom_modules/oe-module-comlink-telehealth/src/Controller/TeleHealthVideoRegistrationController.php
@@ -281,7 +281,7 @@ class TeleHealthVideoRegistrationController
 
     public function resumeUser(string $username, string $password): bool
     {
-        return $this->remoteService->suspendUser($username, $password);
+        return $this->remoteService->resumeUser($username, $password);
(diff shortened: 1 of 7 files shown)


adunsulag:
Gonna bring this in so we can test against the rel-701 w/ telehealth.

PR #6397, opened 2023-04-22
Fixes #6396 webroot path for safe filter

### LINKED ISSUE #6396: ModulesApplication.filterSafeLocalModuleFiles blocks module script/style files when used with subfolder OpenEMR installation
If OpenEMR is installed in a subfolder or anywhere other than the root of the web server (IE web_root is NOT empty) the safety check we use for filtering module files blocks the module script / style files from being included.  This can easily be seen by running the Telehealth module in a webroot installation.

For example if your OpenEMR installation is at **https://oe-build.somedomain.com/openemr/** then the calendar icons will fail to show as the scripts are not injected into the OpenEMR installation in the header whereas if OpenEMR runs at **https://oe-build.somedomain.com/** it will work fine.

adunsulag:
Fixes #6396

Makes it so the webroot path still functions in the safe filter. Prevents injection of scripts in the header unless the file actually exists in OpenEMR and that its locked down to the module directory.

### DIFF
--- a/src/Core/ModulesApplication.php
+++ b/src/Core/ModulesApplication.php
@@ -189,7 +189,14 @@ class ModulesApplication
                 // scripts that have any kind of parameters in them such as a cache buster mess up finding the real path
                 // we need to strip that out and then check against the real path
                 $scriptSrcPath = parse_url($scriptSrc, PHP_URL_PATH);
-                $realPath = realpath($GLOBALS['fileroot'] . $scriptSrcPath);
+                // need to remove the web root as that is included in the $scriptSrc and also in the fileroot
+                $pos = strpos($scriptSrcPath, $GLOBALS['web_root']);
+                if ($pos !== false) {
+                    $scriptSrcPathWithoutWebroot = substr_replace($scriptSrcPath, '', $pos, strlen($GLOBALS['web_root']));
+                } else {
+                    $scriptSrcPathWithoutWebroot = $scriptSrcPath;
+                }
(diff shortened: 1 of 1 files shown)


src/Core/ModulesApplication.php
```diff
@@ -189,7 +189,14 @@ public static function filterSafeLocalModuleFiles(array $files): array
                 // scripts that have any kind of parameters in them such as a cache buster mess up finding the real path
                 // we need to strip that out and then check against the real path
                 $scriptSrcPath = parse_url($scriptSrc, PHP_URL_PATH);
-                $realPath = realpath($GLOBALS['fileroot'] . $scriptSrcPath);
+                // need to remove the web root as that is included in the $scriptSrc and also in the fileroot
+                $pos = strpos($scriptSrcPath, $GLOBALS['web_root']);
```
sjpadgett, line 193: You think for safety should use stripos? Windows is case sensitive.

adunsulag: @sjpadgett You mean case insensitive right?  Could this introduce subtle errors where the script doesn't really exist on linux due to to a misnamed filename?  Yet, in the same vein could create problems on windows if the web server path is misconfigured when the windows file system doesn't care.  

I'm fine with using stripos, seems like either way we can have different problems.

sjpadgett: You're right Windows doesn't care but window folks get use to using caps in paths and filenames.
Do what you think best.

bradymiller (approved):

bradymiller:
2DV-2

PR #6402, opened 2023-04-24
Telehealth signup credential auto save convience feature.

### LINKED ISSUE #6401: Telehealth signup credential auto save.
### Describe the problem
This is a feature enhancement of @juggernautsei nice user account signup application to automatically save the resulting credentials to the Telehealth setup in Config/globals.

Sava and verify save of credentials.

sjpadgett:
- Create an iframe to serve the welcome home page for signup.
- Accept and process a postMessage of credential data from signup application server.
- Auto save credentials to globals to save user from having to copy/pasted.

Fixes #6401

### DIFF
--- a/interface/modules/custom_modules/oe-module-comlink-telehealth/moduleConfig.php
+++ b/interface/modules/custom_modules/oe-module-comlink-telehealth/moduleConfig.php
@@ -5,13 +5,102 @@
  *  package   OpenEMR
  *  link      http://www.open-emr.org
  *  author    Sherwin Gaddis <sherwingaddis@gmail.com>
+ *  author    Jerry Padgett <sjpadgett@gmail.com>
  *  copyright Copyright (c )2021. Sherwin Gaddis <sherwingaddis@gmail.com>
+ *  copyright Copyright (c) 2023 Jerry Padgett <sjpadgett@gmail.com>
  *  license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  *
  */
 
 require_once dirname(__FILE__, 4) . '/globals.php';
 
+use OpenEMR\Common\Csrf\CsrfUtils;
+
 $module_config = 1;
 
-require_once('welcome.php');
+if (!empty($_GET['setup']) ?? null) {
+    /**
+     * Example expected data structure for received data.
+     * $args = ['ctsiVBData' => [
+     * 'registrationUri' => 'https://sginsts.comlinktelehealth.io:28749/CTSIVB/',
+     * 'videoApiUri' => 'https://sginsts.comlinktelehealth.io:22528',
+     * 'ctsiOrgUid' => 'OPEN5C0D4F',
+     * 'ctsiOrgPwd' => 'DD71C92B-834D-4C',
+     * 'ctsiOrgId' => 'OPENEM280D0D']];
+     *
+     * */
+    if (!CsrfUtils::verifyCsrfToken($_GET["csrf_token_form"])) {
+        CsrfUtils::csrfNotVerified();
+    }
+
+    $content = trim(file_get_contents("php://input"));
+    $credentials = json_decode($content, true);
+
+    $items['comlink_telehealth_registration_uri'] = $credentials['registrationUri'];
+    $items['comlink_telehealth_video_uri'] = $credentials['videoApiUri'];
+    $items['comlink_telehealth_user_id'] = $credentials['ctsiOrgUid'];
+    $items['comlink_telehealth_user_password'] = $credentials['ctsiOrgPwd'];
+    $items['comlink_telehealth_cms_id'] = $credentials['ctsiOrgId'];
+    // Save to globals table.
+    foreach ($items as $key => $credential) {
+        sqlQuery(
+            "INSERT INTO `globals` (`gl_name`,`gl_value`) VALUES (?, ?) ON DUPLICATE KEY UPDATE `gl_name` = ?, `gl_value` = ?",
+            array($key, $credential, $key, $credential)
+        );
+    }
+    // Recall GLOBALS to verify the save.
+    require_once dirname(__FILE__, 4) . "/globals.php";
+    // Just a comparison of one of our saved credential.
+    if (isset($GLOBALS['comlink_telehealth_cms_id']) == $items['comlink_telehealth_cms_id']) {
+        echo xlt("Credentials save is successful.");
+    } else {
+        echo xlt("Credentials save failed.");
+    }
+    // todo we may want a location header here!
+    exit;
+}
+?>
+<html>
+<head>
+    <script>
+        // Process received data from origin.
+        function receiveMessage(e) {
+            let remote = "https://credentials.affordablecustomehr.com";
+            // ensure we accept data of known origin.
+            // TODO Remember! if application moves domain to change here!
+            if (e.origin !== remote) {
+                alert(xl("Invalid source!"));
+                return false;
+            }
+            // fix twig single quote escaping of passed in object.
+            let prepared = e.data.replace(/&quot;/ig, '"');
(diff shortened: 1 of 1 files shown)


interface/modules/custom_modules/oe-module-comlink-telehealth/moduleConfig.php
```diff
@@ -5,13 +5,102 @@
...
+    }
+
+    $content = trim(file_get_contents("php://input"));
+    $credentials = json_decode($content, true);
+
+    $items['comlink_telehealth_registration_uri'] = $credentials['registrationUri'];
+    $items['comlink_telehealth_video_uri'] = $credentials['videoApiUri'];
+    $items['comlink_telehealth_user_id'] = $credentials['ctsiOrgUid'];
+    $items['comlink_telehealth_user_password'] = $credentials['ctsiOrgPwd'];
+    $items['comlink_telehealth_cms_id'] = $credentials['ctsiOrgId'];
```
bradymiller, line 43: do any of these get saved in encrypted format (pretty sure the password was on the prior code reviews of this module, but don't recall)?

sjpadgett: yep you're right. thanks for catching.

interface/modules/custom_modules/oe-module-comlink-telehealth/moduleConfig.php
```diff
@@ -5,13 +5,102 @@
...
+    // Save to globals table.
+    foreach ($items as $key => $credential) {
+        sqlQuery(
+            "INSERT INTO `globals` (`gl_name`,`gl_value`) VALUES (?, ?) ON DUPLICATE KEY UPDATE `gl_name` = ?, `gl_value` = ?",
+            array($key, $credential, $key, $credential)
+        );
+    }
+    // Recall GLOBALS to verify the save.
+    require_once dirname(__FILE__, 4) . "/globals.php";
```
bradymiller, line 52: this is already require_once above, so above line would be ignored. Also shouldn't bring in globals.php twice in same script or weird things can happen. Prob shouldn't need to verify this anyways since you are doing doing this sql insert above (and if wanted to confirm, could just grab it from globals table with a sql query).

bradymiller:
2DV

interface/modules/custom_modules/oe-module-comlink-telehealth/moduleConfig.php
sjpadgett: I meant require.

sjpadgett: I found this interesting. I always suspected require_once does not work with globals in many scopes which this proved me correct. Though I meant to require, it still works. My bet is that since globals has some dynamic components on init it may fool interpreter. It also explains some of the scope issues we see from time to time. Or could be the windows compiled version.
Or could be my aging brain:)

sjpadgett:
Finished end to end test on windows EDGE and FF. Works well.
@juggernautsei I highly recommend you test as well on a linux machine once I merge.

image

adunsulag:
Great work @sjpadgett!

juggernautsei:
I see it a lot in all of my installs.

juggernautsei:
I have seen it in v7.0.0(2)

sjpadgett:
I bet this happens if a global with encrypted type is empty

sjpadgett:
I found how but not why. If the drive keys are regenerated after object is encrypted then fail
if (hash_equals($hmacHash, $calculatedHmacHash))

juggernautsei:
That sounds like the "why" to me. Sound like the new keys are different than the previous keys.
The system has never produced the exact same keys every time the keys are removed and new ones are generated.

sjpadgett:
my understanding was they should but @bradymiller needs to wade in.

stephenwaite:
new keys will be unique, Brady has some good background on this [community thread](https://community.open-emr.org/t/getting-a-white-screen-when-saving-demographics-in-502/15277/6?u=stephenwaite)

bradymiller:
Can't get on github directly, so hopefully this reply via email gets
through. Note that the keys should never be removed.

sjpadgett:
In that case there should be a big red alert saying such when keys are lost. Otherwise folks go on fat and happy with no clue they've lost files or passwords that can't be decrypted...

bradymiller:
This has come up in the forums with i think my most recent
mention/explanation here (with a further explanation in a link in that
post):
https://community.open-emr.org/t/openemr-blank-login-screen/19061/6

adunsulag:
Initial testing for me is working fine and I'm not seeing any credential losses.  Do I need to do something to trigger the keys being lost?

juggernautsei:
@adunsulag I got out of order. I had to move the keys from v5 to v7 and then do the database upgrade.

sjpadgett:
I'll be moving all my encryption from drive to database keys.
IMO the drive keys should be backed up in the database with the database keys for safety.
I'm not sure if the file encryption uses the file or database keys but that is my most concern. If database key then replacing various passwords related to file key is not that big a deal except the annoying errors in log.
I think I may change that log message to give a hint what to look for and perhaps an event trace.

bradymiller:
The keys need to encrypt the opposite medium. If you backed up the drive
key onto database then there would be no point in having them. And there is
also no point to encrypting something stored in database with a key in the
database. Read my description on forum for (link in the post) why this is
done. Again, one should never need to remove keys. If one is in a position
where removing keys is felt to be needed, then much better to fix the
underlying problem.

juggernautsei:
Well, I fellows. I have my order correct. I copied the keys from version 5. I created a new database and imported v5 database. Ran the upgrade script and everything was completed successfully. The login screen will not come up. I just get the white screen. I cannot run the patch script. It won't come up either. I have tried 4 times now. This is the fifth try and then I will revert back to v7.0.0 (2).

sjpadgett:
keys can get lost in moves, upgrade or accident.
I get the keys and why we have the two. I was just under the impression that if we auto regenerate drive keys they would be capable of use for previous encryption i.e generated based on the database key but unique from database key.

Allowing drive keys to be regenerated without acknowledgement by the user the necessity for the new keys, could be deceiving and propagate the problem down the road. Or suddenly modules don't work.

I'm just saying we need a better error description and give me some credit that I'm not going off and breaking encryption!

Sherwin I don't think your issue is caused by keys but could be wrong.

adunsulag:
> Well, I fellows. I have my order correct. I copied the keys from version 5. I created a new database and imported v5 database. Ran the upgrade script and everything was completed successfully. The login screen will not come up. I just get the white screen. I cannot run the patch script. It won't come up either. I have tried 4 times now. This is the fifth try and then I will revert back to v7.0.0 (2).

Hey Sherwin, can you clarify here.  You said you ran the upgrade script and it completed successfully.  However, at the end of your message you said the upgrade script failed with the error you pasted.  Can you help me understand the order of your steps?  That will help in reproducing the problem so we can fix it.

adunsulag:
To clarify, @juggernautsei I'm wondering if we need to add in some backwards compatibility in some of the code to skip doing any uuid checks when we are on version 5 until the code can run properly on a v7 database.

The other option is to say that v5 must be upgraded on a v6 codebase and then upgraded from v6 -> v7.  Just trying to explore options here.

juggernautsei:
@adunsulag I understand. Do you want to get together on this to see what needs to happen? I will go ahead and do the two-step process and let you know the results.

adunsulag:
Hey @juggernautsei are there custom changes to this v5 database you are using? 

I'm working on the telehealth press releases right now, but might be available a little later if you think this is something broken in the release we need to investigate.

adunsulag:
@juggernautsei just to clarify, did you mean you copied the keys from your v5 sites folder, or did you copy the keys from the actual database file?  The keys in the database can't be used for the filesystem keys.  I'm assuming you wrote that as a typo, but if indeed you are using keys from the database in the filesystem that will create problems.

juggernautsei:
There are no customizations in the database that  I know about.

PR #6413, opened 2023-04-26
Openemr fix telehealth signup locale

adunsulag:
Fixed up the verbiage and some problems with the telehealth live payment signup process. It was popping up an alert box with the paypal subscriber id and expected the patient to copy and remember that value for signup.  This is most likely what was causing the multiple signup problems that users of the module were dealing with.

Broke the signup into three steps and provide progress status to the users as they signup.  Better instructions on how to move to the next step are included in each process.

PR #6414, opened 2023-04-26
Openemr fix telehealth signup locale (#6413)

adunsulag:
* Add save and locale telehealth check instructions

Added a check for the locale settings which should be setup to avoid clinic problems.

Added instructions that the user needs to log out and log back in after saving the telehealth settings.

* Fix psr problems.

* Fix telehealth subscription signup.

Include instructions and the subscription id for the user to use when they signup for the credentials.

Add a copy button and switch the alert of the subscription id to actually show up in a success message info box.  Before the change the user had limited understanding on how to proceed to the next steps in the configuration system.

* Improve signup verbiage, add subscription id

To make the signup process easier for users, we break the signup into three steps with instructions.  We also add saving the paypal subscription id so we can check if the paypal process has already completed and been signed up.

We show progress to the user as they signup.

* Fix if conditions, paypal key

Fixed the paypal key coming back from Sherwin to be the right key value. Fixed the conditions on the subscription id.

* Include the free verbiage in the description.

* Updated telehealth plan for 14 day free trial.

* Defaulting values to avoid sql error.

### DIFF
--- a/interface/modules/custom_modules/oe-module-comlink-telehealth/moduleConfig.php
+++ b/interface/modules/custom_modules/oe-module-comlink-telehealth/moduleConfig.php
@@ -16,6 +16,7 @@ require_once dirname(__FILE__, 4) . '/globals.php';
 
 use OpenEMR\Common\Crypto\CryptoGen;
 use OpenEMR\Common\Csrf\CsrfUtils;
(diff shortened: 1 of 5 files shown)


bradymiller:
bringing into rel-701 in case do a package build tonight :)

PR #6416, opened 2023-04-26
Fix institution id to generate if empty

### LINKED ISSUE #6415: Telehealth Config fails to verify settings if globals have not been saved
If a user signs up for the telehealth settings before they have saved anything in their OpenEMR globals config, the process fails as there is no institutionId since the globals value is empty until the Admin -> Config page has been saved.

We need to set the TelehealthGlobalConfig class to use the UniqueInstallationUuid class to retrieve the institution id instead of hitting the globals array.  That way the uuid is generated if it is empty.

adunsulag:
Telehealth settings needs to use the UniqueInstallationUuid instead of using the globals.

Fixes #6415.  This should be pretty rare that someone hasn't configured anything in their OpenEMR installation before setting up the telehealth settings, but this fixes that bug.


bradymiller (approved):

bradymiller:
2DV(1)

PR #6198, opened 2023-02-10
reset fail user fail counter on password update

sorvani:
This fixes #6197

PR #6431, opened 2023-05-04
feat: improve max failed login mechanism

### LINKED ISSUE #5627: Improve user experience of the  "Maximum Failed Login Attempts (0 for no maximum)." setting
Considering expanding upon the maximum failed login attempts to also offer an option where it automatically resets (ie. will work again automatically rather than require manual intervention by admin) after a certain time. Thus if for example, there a 3 failed requests and it shuts down account, it won't be shut out indefinitely if the user sets this new setting to a time interval (ie. after that time interval, the user can then login).

bradymiller:
fixes #5627 

Basically provides ability to set a timeout (from last failed login) where then ok to login.
Also sets default setting of 20 for failed logins (and 1 hour for the timeout above)
Displayed the failed user login counter and last failed date in the Users display.

Would it also be helpful to have ability to reset the failed login counter in the Users display?

### DIFF
--- a/interface/usergroup/usergroup_admin.php
+++ b/interface/usergroup/usergroup_admin.php
@@ -559,6 +559,21 @@ function authorized_clicked() {
  f.calendar.checked  =  f.authorized.checked;
 }
 
+function resetCounter(username) {
+    top.restoreSession();
+    request = new FormData;
+    request.append("function", "resetUsernameCounter");
+    request.append("username", username);
+    request.append("csrf_token_form", <?php echo js_escape(CsrfUtils::collectCsrfToken('counter')); ?>);
+    fetch("<?php echo $GLOBALS["webroot"]; ?>/library/ajax/login_counter_ip_tracker.php", {
+        method: 'POST',
+        credentials: 'same-origin',
+        body: request
+    });
+    let loginCounterElement = document.getElementById('login-counter-' + username);
+    loginCounterElement.innerHTML = "0";
+}
+
 </script>
 
 </head>
@@ -618,6 +633,7 @@ function authorized_clicked() {
                                 echo '<th>' . xlt('Password Expiration') . '</th>';
                             }
                             ?>
+                            <th><?php echo xlt('Failed Login Counter'); ?></th>
                         </tr>
                     <tbody>
                         <?php
@@ -687,6 +703,41 @@ function authorized_clicked() {
                                 }
                                 echo '</td>';
                             }
+                            if (empty($iter["active"])) {
+                                echo '<td>';
+                                echo xlt('Not Applicable');
+                            } else {
+                                echo '<td id="login-counter-' . attr($iter["username"]) .  '">';
+                                $queryCounter = privQuery("SELECT `login_fail_counter`, `last_login_fail`, TIMESTAMPDIFF(SECOND, `last_login_fail`, NOW()) as `seconds_last_login_fail` FROM `users_secure` WHERE BINARY `username` = ?", [$iter["username"]]);
+                                if (!empty($queryCounter['login_fail_counter'])) {
+                                    echo text($queryCounter['login_fail_counter']);
+                                    if (!empty($queryCounter['last_login_fail'])) {
+                                        echo ' (' . xlt('last on') . ' ' . text(oeFormatDateTime($queryCounter['last_login_fail'])) . ')';
+                                    }
+                                    echo ' ' . '<button type="button" class="btn btn-sm btn-danger ml-1" onclick="resetCounter(' . attr_js($iter["username"]) . ')">' . xlt("Reset Counter") . '</button>';
+                                    $autoBlocked = false;
+                                    $autoBlockEnd = null;
+                                    if ((int)$GLOBALS['password_max_failed_logins'] != 0 && ($queryCounter['login_fail_counter'] > (int)$GLOBALS['password_max_failed_logins'])) {
+                                        if ((int)$GLOBALS['time_reset_password_max_failed_logins'] != 0) {
+                                            if ($queryCounter['seconds_last_login_fail'] < (int)$GLOBALS['time_reset_password_max_failed_logins']) {
+                                                $autoBlocked = true;
+                                                $autoBlockEnd = date('Y-m-d H:i:s', (time() + ((int)$GLOBALS['time_reset_password_max_failed_logins'] - $queryCounter['seconds_last_login_fail'])));
+                                            }
+                                        } else {
+                                            $autoBlocked = true;
+                                        }
+                                    }
+                                    if ($autoBlocked) {
+                                        echo '<br>' . xlt("Currently Autoblocked");
+                                        if (!empty($autoBlockEnd)) {
+                                            echo ' (' . xlt("Autoblock ends on") . ' ' . text(oeFormatDateTime($autoBlockEnd)) . ')';
+                                        }
+                                    }
+                                } else {
+                                    echo '0';
+                                }
+                            }
+                            echo '</td>';
                             print "</tr>\n";
                         }
                         ?>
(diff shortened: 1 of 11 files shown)


bradymiller:
Prob also should make the last login fail date null as done in this prior fix:
https://github.com/openemr/openemr/pull/6198

stephenwaite:
excited_elephant

bradymiller:
Added ability to reset counter from the user admin gui.

Also added mechanism to deal with failed logins from a ip address perspective (with ability to also flag ip address as always blocked and even flag to skip the prevent timing attack when always blocking an ip address (skipping the timing attack prevention will increase performance of server since don't need to run a fake hash for every attempt, but this means the bad actor will know that their ip address is being blocked which is good for one server, but in the big scheme of things, ie. the OpenEMR ecosystem of thousands of servers is actually better to not do this)

Working on:
Report to view the ip addresses and reset their counters and set the block and skip timing attack flags
Send email to admin when a user or ip address gets automatically blocked because of too many failed logins

interface/usergroup/usergroup_admin.php
```diff
@@ -687,6 +703,26 @@ function authorized_clicked() {
                                 }
                                 echo '</td>';
                             }
+                            if (AuthUtils::useActiveDirectory($iter["username"]) || empty($iter["active"])) {
+                                // LDAP bypasses failed login counter mechanism
```
bradymiller, line 707: remove above comment since this is no longer true

interface/usergroup/usergroup_admin.php
```diff
@@ -687,6 +703,26 @@ function authorized_clicked() {
                                 }
                                 echo '</td>';
                             }
+                            if (AuthUtils::useActiveDirectory($iter["username"]) || empty($iter["active"])) {
```
bradymiller, line 706: Also remove the above check for useActiveDirectory since this is no longer true

interface/usergroup/usergroup_admin.php
```diff
@@ -687,6 +703,26 @@ function authorized_clicked() {
...
+                            } else {
+                                echo '<td id="login-counter-' . attr($iter["username"]) .  '">';
+                                $queryCounter = privQuery("SELECT `login_fail_counter`, `last_login_fail` FROM `users_secure` WHERE BINARY `username` = ?", [$iter["username"]]);
+                                if (!empty($queryCounter['login_fail_counter'])) {
+                                    echo text($queryCounter['login_fail_counter']);
+                                    if (!empty($queryCounter['last_login_fail'])) {
+                                        echo ' (' . xlt('last on') . ' ' . text(oeFormatDateTime($queryCounter['last_login_fail'])) . ')';
+                                    }
+                                    if (AclMain::aclCheckCore('admin', 'super')) {
+                                        echo ' ' . '<button type="button" class="btn btn-sm btn-danger ml-1" onclick="resetCounter(' . attr_js($iter["username"]) . ')">' . xlt("Reset") . '</button>';
```
bradymiller, line 719: Change to "Reset Counter"

library/ajax/login_counter_ip_tracker.php
```diff
@@ -0,0 +1,38 @@
...
+use OpenEMR\Common\Auth\AuthUtils;
+use OpenEMR\Common\Csrf\CsrfUtils;
+
+require_once(__DIR__ . "/../../interface/globals.php");
+
+if (!CsrfUtils::verifyCsrfToken($_POST["csrf_token_form"], 'counter')) {
+    CsrfUtils::csrfNotVerified(false);
+}
+
+if (!AclMain::aclCheckCore('admin', 'super')) {
```
adunsulag, line 23: Do you think frontend staff are going to need the ability to reset someone from being logged out?  This seems like a fairly common support staff function.  I understand the desire to lock it down to just an admin thing... but it seems like the support staff are going to hit this the most from someone.

library/globals.inc.php
```diff
@@ -2148,10 +2148,31 @@ function gblTimeZones()
...
+            xl('Time (seconds) to Reset Maximum Failed Login Attempts For User'),
+            'num',                            // data type
+            '3600',                           // default to 1 hour
+            xl('Time (seconds) to Reset Maximum Failed Login Attempts Counter For User (0 for no reset).')
+        ),
+
+        'ip_max_failed_logins' => array(
+            xl('Maximum Failed Login Attempts From IP Address'),
+            'num',                            // data type
+            '100',                            // default
```
adunsulag, line 2167: is the reason why you made this 100 due to a worry about NATed IP addresses?

sql/7_0_1-to-7_0_2_upgrade.sql
```diff
@@ -136,3 +136,24 @@ DROP TABLE IF EXISTS `pma_table_coords`;
...
+#EndIf
+
+#IfNotRow globals gl_name time_reset_password_max_failed_logins
+UPDATE `globals` SET `gl_value` = 20 WHERE `gl_name` = 'password_max_failed_logins' AND `gl_value` = 0;
+#EndIf
+
+#IfNotTable ip_tracking
+CREATE TABLE `ip_tracking` (
+`id` bigint NOT NULL auto_increment,
+`ip_string` varchar(255) DEFAULT '',
```
adunsulag, line 151: This table will get large, does it make sense to store the ip string or store it as the 16 byte for ipv6 and 4byte for ipv4?  It will definitely be better for debugging/administration to have the string representation.

src/Common/Auth/AuthUtils.php
```diff
@@ -943,35 +1005,90 @@ private function checkPasswordNotExpired($user)
...
+                return ['pass' => false, 'force_block' => true, 'skip_timing_attack' => true];
+            } else {
+                return ['pass' => false, 'force_block' => true, 'skip_timing_attack' => false];
+            }
+        }
+        if ($query['ip_login_fail_counter'] >= $GLOBALS['ip_max_failed_logins']) {
+            if (!empty($GLOBALS['ip_time_reset_password_max_failed_logins']) &&
+                $GLOBALS['ip_time_reset_password_max_failed_logins'] > 0 &&
+                !empty($query['seconds_last_ip_login_fail']) &&
+                $query['seconds_last_ip_login_fail'] > $GLOBALS['ip_time_reset_password_max_failed_logins']
```
adunsulag, line 1060: seems like you could end up with some wierd bugs if someone were to put a letter in the globals, would it be better to grab an intval of $GLOBALS['ip_time_reset_password_max_failed_logins'] and do the comparison that way?

adunsulag (commented):
Looked it over, and noted a couple things, overall though looks fine to me.

bradymiller:
User gui
image

ip tracking gui
image

library/ajax/login_counter_ip_tracker.php
bradymiller: changed this to admin users acl (same that is needed to get to the admin users gui)

library/globals.inc.php
bradymiller: just picked that number. happy to go higher or lower on this (goal is to not break user installs since this will be default and also prevent bruteforce foolishness)

sql/7_0_1-to-7_0_2_upgrade.sql
bradymiller: note ip_string is not always a straightforward ip address. sometimes is 2 addresses in string form:
https://github.com/openemr/openemr/blob/9dc161d8d5b3fa71508b3e9f23b7c3e4aea49b97/library/sanitize.inc.php#L31

src/Common/Auth/AuthUtils.php
bradymiller: yep, prob good idea to go through the globals that using in this pr and ensure are int
TODO for me

PR #6448, opened 2023-05-09
Fixes #6446, #6447, query helpers

### LINKED ISSUE #6446: Add transaction methods to QueryUtils
It would be helpful to add methods for starting/committing/rolling back transactions into the QueryUtils class.

### LINKED ISSUE #6447: Enhance UserService to return username in queries
Extend the UserService methods to return the username in queries when toggled.  In many of our queries that get passed down into the APIs and frontends we don't want to expose the username.  However, in same cases in the UserService we want to expose sensitive fields.  We should add a way of modifying the UserService to return the username as needed.

adunsulag:
Fixes #6446
Fixes #6447

Adds transaction methods to the QueryUtils class.

Adds a toggleSensitiveFields method to the UserService for returning data such as the username.

### DIFF
--- a/src/Common/Database/QueryUtils.php
+++ b/src/Common/Database/QueryUtils.php
@@ -228,4 +228,30 @@ class QueryUtils
     {
         return \generate_id();
     }
+
+    public static function startTransaction()
+    {
+        \sqlBeginTrans();
+    }
+
+    public static function commitTransaction()
+    {
+        \sqlCommitTrans();
+    }
+
+    public static function rollbackTransaction()
+    {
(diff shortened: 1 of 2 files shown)


adunsulag:
``openemr-cmd pf``` let's me down once again...

bradymiller:
Odd on the styling stuff since both github actions and the docker are using same version 3.7.2 of PHP_CodeSniffer . I wonder if there are permission issues in your docker environment (ie. the docker is unable to modify stuff since the host permissions is not letting it).

bradymiller (approved):

PR #6450, opened 2023-05-11
First attempt at refresh for the patient portal

### LINKED ISSUE #6449: Refresh patient port look and feel
### Describe the problem
Update the look of the patient portal 

Write a concise description of the problem 
The patient portal is dated 

Cleaning the log in page to simplify the look

Stated above

juggernautsei:
Fixes #6449

Refreshing the look and feel of the patient portal

### DIFF
--- a/portal/index.php
+++ b/portal/index.php
@@ -314,8 +314,19 @@ if (!(isset($_SESSION['password_update']) || (!empty($GLOBALS['portal_two_pass_r
 <head>
     <title><?php echo xlt('Patient Portal Login'); ?></title>
     <?php
-    Header::setupHeader(['no_main-theme', 'datetime-picker', 'patientportal-style', 'patientportal-base', 'patientportal-register']);
+    Header::setupHeader(['no_main-theme', 'shrink-to-fit']);
     ?>
+    <style>
+        .sign-in-container {
+            display: block;
+            width: 50%;
+            margin: 0   auto;
+            position: absolute;
+            top: 50%;
+            -ms-transform: translateY(-50%);
+            transform: translateY(-50%);
+        }
+    </style>
     <script>
         function checkUserName() {
             let vacct = document.getElementById('uname').value;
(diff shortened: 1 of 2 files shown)


juggernautsei:
The next steps are to make this responsive.

portal/index.php
```diff
@@ -314,8 +314,19 @@
 <head>
     <title><?php echo xlt('Patient Portal Login'); ?></title>
     <?php
-    Header::setupHeader(['no_main-theme', 'datetime-picker', 'patientportal-style', 'patientportal-base', 'patientportal-register']);
+    Header::setupHeader(['no_main-theme', 'shrink-to-fit']);
```
sjpadgett, line 317: Put back datepicker unless you're sure it's scope is not needed by upstream!

portal/index.php
```diff
@@ -526,39 +537,40 @@ function enableVerifyBtn(){
             </form>
         <?php } else {
             ?> <!-- Main logon -->
-        <img class="img-fluid login-logo" src='<?php echo $logoSrc; ?>'>
-        <form class="text-center" action="get_patient_info.php" method="POST" onsubmit="return process()">
-            <?php if (isset($redirectUrl)) { ?>
-                <input id="redirect" type="hidden" name="redirect" value="<?php echo attr($redirectUrl); ?>" />
-            <?php } ?>
-            <fieldset>
-                <legend class="bg-primary text-white pt-2 py-1"><h3><?php echo xlt('Patient Portal Login'); ?></h3></legend>
```
sjpadgett, line 535: Fieldset and legend are important for accessibility and screen readers. We need to keep and even add throughout.

portal/index.php
```diff
@@ -526,39 +537,40 @@ function enableVerifyBtn(){
...
-                <div class="jumbotron jumbotron-fluid px-5 py-3">
-                    <div class="form-row my-3">
+        <img class="img-fluid w-25" src='<?php echo $logoSrc; ?>'>
+        <div class="form-container sign-in-container w-50">
+            <form class="patient-form" action="get_patient_info.php" method="POST" onsubmit="return process()" style="padding-left: 250px">
```
sjpadgett, line 542: How does padding in a style add to responsiveness?
If anything we should remove the old portal theme, implement openemr dark or light theme selectable from Config and  include BS in header.

sjpadgett:
The current login is already responsive so I'm trying to understand the goal here.
If you're going to restyle throughout all of portal, which is fine btw, then should start out on a surer footing by stripping out old theming and starting from our current themes dark and light only.

Also you need to be careful when trying to predict any dependency scope. The portal is based on the SPA(single page app) principle where components may be contained in iframe or a card. I wouldn't be surprised after all the different restyles over the years we don't have some misbehaving dependencies.

Also I'd like to see some sort of plan as to the extent and goal of the project. Styling is so subjective where appearance should take a backseat to functionality. 
So in that vain we'd want to ensure useability, accessibility and responsiveness are our primary goal. 

I'm not trying to rain on the parade and am very happy you want to improve portal however I think we all should agree on changes so hopefully this is the last time we visit portal style restructuring.

Be aware though I have plans to restructure documents where it would appear to be a separate app accessible from portal or standalone. Whichever way I go the goal is to lose the top menu getting in the way and getting some real estate back.

If you need help with introducing BS dependency let me know and I'll ask someone else:) kidding, let me know.....

juggernautsei:
@sjpadgett I don't have the background for planning out how things will go. I am still learning. Trial and error. Those are my friends when developing. I was not trying to change the functionality but "modernize" the look of the portal. I don't feel you are raining on my parade because it is dark and wet and cold where I am so I can't see a parade of anything. I have to do a lot of reading comprehension to get as far as I have. 
I have had many requests to make the portal look more inviting to patients. This was my attempt to do so. I had some free time while waiting on other projects to materialize. I thought I would try my hand at making this look like 
https://secure.simplepractice.com/users/sign_in
That is the extent of my planning. There is nothing out here in the world of online learning that can teach nuts and bolts in a short period of time. I would have to go through weeks of classes to find out they are not going to tell me what I need to know in order to build what I am trying to achieve. I am not put off by this fact but accept the challenge. If you would like to school me. I have no issue with it. It beats reading through five miles of text to find one command to move the box to the middle of the screen.
All of coding is subjective. One says do it this way and another says something different. I applaud the brilliance of the people that can pick this stuff up and run with it and create marvelous code. I struggle on. 
My feeling aside. Call me when you get a chance.

portal/index.php
juggernautsei: That was a desperate effort to get the screen to behave how I wanted it to. It was the only way I could get the form in the center of the screen. 
As I said, this is the first draft so you could see where I was headed.

src/Core/Header.php
```diff
@@ -87,8 +87,13 @@ public static function setupHeader($assets = [], $echoOutput = true)
         $output = "\n<meta charset=\"utf-8\" />\n";
         // Makes only compatible with MS Edge
         $output .= "<meta http-equiv=\"X-UA-Compatible\" content=\"IE=edge\" />\n";
-        // BS4 required tag
-        $output .= "<meta name=\"viewport\" content=\"width=device-width, initial-scale=1, shrink-to-fit=no\" />\n";
```
mdsupport, line 91: BS takes care of responsive behavior so suggest not to interfere with their calculation. If you noticed any issues, there is a good chance that it was because script and/or library modified the same --or-- additional form-factor specific tags (sm,lg etc) were not included.

BS5 also changes this to 
`<meta name="viewport" content="width=device-width, initial-scale=1">`

juggernautsei:
I am going to back up and try a different approach with less changes to the code

src/Core/Header.php
sjpadgett: We have been removing any jquery-ui throughout core so not a good idea to introduce in portal.

juggernautsei: @sjpadgett now I remember that is why I stopped using jquery.

sjpadgett:
We'll probably never get rid of jQuery(very prevalent throughout core) but jQuery-ui is close to being gone.
If you can make it let's try to take a few minutes on call today to discuss this project.

PR #6468, opened 2023-05-23
Issue2101

### LINKED ISSUE #2101: Allow Whitelisting of IP's in Globals by SuperAdmin
This is super easy. A superadmin sets the IP addresses or subnet masks as the allowed range of  who can log on. These can be remote or RFC1918 addresses. 

`Index.php` quickly checks IP against database as the user comes to the application. If not in the whitelist then is redirected to blank page or perhaps `exit()` is called.

getusername67:
Fixes #2101

Adds the feature to whitelist IP addresses for administrators. Placed in admin->config->security. Supports IPv4 and CIDR notation.

I have also added a new textarea datatype that can be used in the configuration section.

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -547,7 +547,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                                                 if ($userMode) {
                                                     $globalTitle = $globalValue;
                                                 }
-                                                        echo "  <textarea name='form_$i' id='form_$i' rows='4' cols='50'>" . attr($fldvalue) . "</textarea>\n";
+                                                        echo "  <textarea name='form_$i' id=" . attr($fldid) . " rows='3' cols='20'>" . text($fldvalue) . "</textarea>\n";
                                             } elseif ($fldtype == GlobalSetting::DATA_TYPE_DEFAULT_RANDOM_UUID) {
                                                 if ($userMode) {
                                                     $globalTitle = $globalValue;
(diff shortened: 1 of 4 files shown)


mdsupport:
Why should a business application get bloated with low level access control when webservers (e.g. [apache](https://httpd.apache.org/docs/2.4/mod/mod_access_compat.html)) will do this very effectively.

If admins see any value, suggest the (optional) module route that uses lists rather than using global as catchall.

adunsulag:
I like @mdsupport's thoughts, is the need for this arising out of a shared hosting concern where you may not have access to ip blocking?

getusername67:
@mdsupport how does using lists differ from using the global?

mdsupport:
From an user perspective instead of having the whitelist limited to a single string, the lists provide multiple records with description and ability activate/deactivate at granular level.

robertdown:
I find a business value-add here as this is an extension of ACL management. While I agree that this could be accomplished in the webserver, holding admins to a higher level of security should be managed within the application, not at the web server level. Using a list would be an option, but I don't think adding the global is a bad approach.

interface/super/edit_globals.php
```diff
@@ -797,8 +802,92 @@ function checkBackgroundServices()
...
+
 $(function () {
     tabbify();
+
+    let elm = document.getElementById("white_list");
+    let form = document.getElementById('theform');
+    let is_valid = true;
+
+    // Create and style the helper text element
+    let helper_text = document.createElement("span");
```
robertdown, line 868: Check out https://getbootstrap.com/docs/4.6/components/forms/#validation as a way to better handle validation styling.

src/Common/Acl/AclMain.php
```diff
@@ -225,6 +225,32 @@ public static function aclCheckCore($section, $value, $user = '', $return_value
...
+            } else {
+                if ($remoteAddr == $ips[$i]) {
+                    $valid_IP = true;
+                }
+            }
+        }
+
+        // Allows access when white list of IPs is empty or if the user IP is in the white list
+        if (!empty($GLOBALS['white_list']) && !$valid_IP) {
+            exit();
```
robertdown, line 251: Instead of `exit()`'ing here you'd want to follow the convention of returning a `false` boolean to allow the upstream script to decide what to do next

robertdown (changes requested):

bradymiller:
To clarify. Is this a white list for administrators or all users? It looks like this would be for all users. If do something like this, would recommend instead doing this when auth on login (see the recently added IP blocking feature for timed out logins for likely better spot from this and could even build on the gui to support this feature: https://www.open-emr.org/wiki/index.php/Brute_Force_Login_Prevention )

src/Common/Acl/AclMain.php
```diff
@@ -225,6 +225,32 @@ public static function aclCheckCore($section, $value, $user = '', $return_value
             }
         }
 
+        $remoteAddr = $_SERVER['REMOTE_ADDR'] ?? null;
```
bradymiller, line 228: Note that $_SERVER['REMOTE_ADDR'] does not always get what you are looking for. This is why the log engine uses following function to collect info for this type of thing:
https://github.com/openemr/openemr/blob/61e4f5bc1d0d5c12661ad2d46b1d398426d138ca/library/sanitize.inc.php#L20
And then uses this info to place following into the log:
```
$ip = collectIpAddresses();
$ip['ip_string']
```

adunsulag:
Been a year since we've had action on this, closing down the PR.  Feel free to re-open it once the issues have been addressed.

PR #6474, opened 2023-05-26
When creating a lab order …

### LINKED ISSUE #6473: Nice to land on ICD10s
### Describe the problem
When creating a lab order and it is time to select the diagnosis. It would be nice if the popup opened to ICD10s.
In the past it use to do that. 
Let us bring that back.

Write a concise description of the problem 

juggernautsei:
the popup should land on ICD10

Fixes #6473


juggernautsei:
did this get fixed already?
Bump bump in the night

robertdown:
Shouldn't it respect `$GLOBALS['default_search_code_type']`?

juggernautsei:
> Shouldn't it respect `$GLOBALS['default_search_code_type']`?
When creating lab orders. The interface does not default to ICD10s, which are needed. I remembered that was the way it was in the past. I don't know what possessed it to be changed to a global, but that global is not applied to this location. 
A global here does not make sense because it should not be globally applied. The lab needs ICD10s, and the fee sheet needs CPTs. It is not wise to apply a global to individual use cases. IMO, That is why I was putting it back the way it was. 
The default code for procedure orders should be ICD10 and not changed by a global setting.
@robertdown

sjpadgett:
Yes you have to send a diagnosis code however that code is not always ICD10. Though whatever code system you wish to set as default makes sense to me! Many clinics overseas uses SNOMED

juggernautsei:
Understood.

PR #6475, opened 2023-05-27
Telehealth - fix thrown console errors (#6443)

### LINKED ISSUE #6442: Telehealth module javascript console errors.
When the video bridge shuts down it throws a javascript console error.  Sometimes when the patient waiting room shuts down, the intervals don't get cleared off quickly enough and it also throws javascript console errors on shut down.  Nothing is impacted from the UX side, but we want to make sure we clean up these errors.

adunsulag:
Fixes #6442

Updated the cvb library which no longer throws exceptions when the video bridge shuts down.  Fixed a couple of places where console errors were being thrown outside of the cvb library.

Bumped the script version to make sure people get the changes.


version.php
```diff
@@ -47,7 +47,7 @@
     $v_js_includes = md5(microtime());
 } else {
     // Change this number when bumping
-    $v_js_includes = 72;
+    $v_js_includes = 73;
```
bradymiller, line 50: I'm gonna do a quick bump of this counter in master so it is different from this branch

bradymiller: done :)

adunsulag: @bradymiller should I hold off from merging? This commit was already merged in master.

bradymiller (approved):

version.php
bradymiller: bring it in :)

PR #6486, opened 2023-05-31
Improved patient finder UI

robertdown:
Quick UI that moves the floating New Patient button into the navbar, adds a bit of padding, display the # of entries option on datatables, move `thead` to the light theme more in line with other areas.

This PR also creates the ability to extend the BaseMenuItem capabilities surrounding ACL checks. You can now pass in a 3rd element in when using `setAcl()` for the return value that AclCheckCore looks at, ensuring you can better describe the exact perms needed.


robertdown:
# Before 
image

# After
image

src/Events/UserInterface/PageHeadingRenderEvent.php
```diff
@@ -98,7 +99,22 @@ public function setPrimaryMenuItem(MenuItemInterface $item, ?int $position = nul
...
+
+        foreach ($items as $k => $i) {
+            // No check if ACL element is absent
+            if (count($i->getAcl()) == 0) {
+                continue;
+            }
+
+            // Unset the item if the user does not have access
+            $acl = $i->getAcl();
+            $rv = (count($acl) > 2) ? $acl[2] : [];
```
bradymiller, line 112: if acl[2] is not set, then would set it to empty string instead of an empty array

robertdown:
@bradymiller Ok, I made the code change based on your request. I also merged in upstream changes which included the recent patients functions.

# Updated After
image

stephenwaite:
nice @robertdown except for the maybe unnecessary borders around the Search box?
Screenshot from 2023-08-02 10-23-42

stephenwaite:
I guess I mean the long thin borders that are created by the `nav-tab` and then the line underneath the search box

PR #6515, opened 2023-06-04
Improved customizations on login page logo and portal navbar logo

robertdown:
Create new globals to set the default width of core login page logos. This mimics bootstrap built in variables for `w-25`, `w-50`, `w-75`, `w-100`. Also allows customization of logo placement against themselves by setting the flex-direction on the two logos. This allows the logos to be stacked on top of each other or side-by-side.

The connector between the two logos can be customized or removed all together. The default is a plus icon.

Additionally, the portal navbar is now outside of the standard container and full-width. The portal navbar logo height is customizable. The default is 30. This is based on an problem Jit raised.


interface/new/new_comprehensive.php
```diff
@@ -379,31 +379,41 @@ function srchDone(pid){
...
-                <div class="accordion" id="dem_according">
-                <form action='new_comprehensive_save.php' name='demographics_form' id='DEM'
-                      method='post'
-                      onsubmit='return submitme(<?php echo $GLOBALS['new_validate'] ? 1 : 0;?>,event,"DEM",constraints)'>
+        <div class="row flex-row">
+            <div class="">
+                <div class="list-group list-group-flush sticky-top" id="scrollspy-control">
+                    <?php $toc = $grparr; ?>
+                    <?php for ($i = 1; $i <= count($toc); $i++) : ?>
+                        <a href="#form_section_<?php echo $i; ?>" class="list-group-item list-group-item-action"><?php echo $toc[$i]['grp_title']; ?></a>
```
bradymiller, line 400: would escape `$toc[$i]['grp_title']` with text()

library/globals.inc.php
```diff
@@ -545,6 +545,40 @@ function gblTimeZones()
...
+                'w-50' => '50%',
+                'w-75' => '75%',
+                'w-100' => '100%'
+            ],
+            'w-50',
+            xl('Determine the width of the primary logo compared to the container'),
+        ],
+
+        'secondary_logo_width' => [
+            xl('Width of secondary compared to the container'),
```
bradymiller, line 561: is logo supposed to be in the title?
Width of secondary logo compared to the container

portal/home.php
```diff
@@ -302,7 +302,8 @@ function buildNav($newcnt, $pid, $result)
     'images_static_relative' => $GLOBALS['images_static_relative'],
     'youHave' => xl('You have'),
     'navMenu' => $navMenu,
-    'pagetitle' => xl('Home') . ' | ' . xl('OpenEMR Portal'),
+    'primaryMenuLogoHeight' => $GLOBALS['portal_primary_menu_logo_height'] ?? '30',
+    'pagetitle' => xl('Home') . ' | ' . xl($GLOBALS['openemr_name']) . ' ' . xl('Portal'),
```
bradymiller, line 306: would not translate `$GLOBALS['openemr_name']`

sjpadgett:
@robertdown 
Can you bring this in if you want these changes to follow my restructure.
My plan is to create a landing page dashboard to run portal apps and minimizing the use of top level menus as much as possible.

Another major reason is to create hooks for adding apps from modules to be called from dashboard. Also we are adding support for additional portal APIs and SMART app support to help support native app developers.

For branding I see an About module or link to clinic website. We'' need at least one sample module so this could be one or maybe you'd be interested in doing a module for patient document uploads though this may be better as attachments in secure messaging. Speaking of which it is planned to notify patient whenever a secure message is sent by clinicians.

So first go around is to remove custom theme and construct new landing page with our main portal apps front facing.

robertdown:
Hm, looks like I branched this PR off of a different PR I had not yet published instead of master. I need to pull back a couple of changes and will then bring in just the relevant stuff. Thanks for the review @bradymiller I will make the recommended changes.

portal/index.php
```diff
@@ -415,189 +415,203 @@ function enableVerifyBtn() {
...
+                </div>
+            </div>
+        <?php endif; ?>
+        <?php if ($GLOBALS['language_menu_login']) : ?>
+            <?php if (count($result3) != 1) : ?>
+                <div class="form-group row">
+                    <label class="col-form-label col-sm-12 col-md-3" for="selLanguage"><?php echo xlt('Language'); ?></label>
+                    <div class="col-sm-12 col-md-9">
                         <select class="form-control" id="selLanguage" name="languageChoice">
+                            <option value="<?php echo attr($defaultLangID); ?>" selected><?php echo xlt("Deault") . " - " . xl($defaultLangName); ?></option>
```
bradymiller, line 565: mispelling in `Deault`

bradymiller: xlt($defaultLangName)

PR #4502, opened 2021-07-04
Implemented duplicate patients management.

sunsetsystems:
Identification and correction of duplicate patients involves the following:

 1. Column "dupscore" in the patient_data table was created to contain a value representing
    the likelihood that the patient duplicates any of the patients created earlier (that is,
    with a lesser "pid" value). It may have any of the following values:

    o -9, initially assigned, indicating that it needs to be computed.
    o -1, indicating that the patient has been manually flagged as not duplicating a previous patient.
    o A score in the range 0-40 computed by the SQL expression found in library/dupscore.inc.php.
      A value of 12 or higher is considered to be very likely a duplicate.

 2. The script library/dupscore.inc.php which provides the SQL expression needed for scoring.

 3. The command-line tool contrib/util/dupscore.cli.php which should be run to initially
    assign all patient_data.dupscore values, and may be run from time to time to update
    or recompute them. See the source of that script for its usage syntax.

 4. The web tool Administration -> Patient -> Manage Duplicates.
    For each patient with a dupscore of 7 or higher, it shows below it all the corresponding
    matching patients. Thus you see a collection of patient groups.
    Within each group you can choose to mark the first patient as not a duplicate, or you can
    merge it with one of the other patients. You do that by choosing "Merge and Keep" or
    "Merge and Discard" for the other patient, which brings up the Merge Patients page to
    complete the merge. That page describes in more detail what it will do.

Fixes #

PR #6518, opened 2023-06-05
Improve dup patients

### LINKED ISSUE #6517: manage duplicate patients issues
### Describe the bug
In admin/patients/manage duplicates:
1. when a patient's demographics were changed so they then became a duplicate of another patient, they were not always in the Duplicate Management Patient table
2. a patient would be in the duplicate table but as the sole member of a group

1. When a patient's demographics are changed they should be checked as a duplicate again against all the patients, not just those with a lower pid
2. a patient would be on it's own in the displayed duplicates table - where it's dupscore > 7 but the other potential duplicate patients had a higher value pid

### To Reproduce

1. alter a patient's names, date of birth, ss number or  phone numbers to make it the duplicate of another patient whose pid was greater than the changed patient's

do 'admin/patients/manage duplicates' or if a duplicate table already displayed press 'renew'

the changed patient does not appear with the duplicate patient whose pid was greater

2. can be reproduced by changing patients demographics appropriatly: 
look at the pids in the following screen shot
dup patient bug

for patient named 'New Dupscore' with pid 17 and dup score 11: as dup score > 7 they are output to the table, but the subsequent search of patients with lower pids does not find their dups, so they are on a line alone. Later it's real dup which has pid 26 is found do have a score >7, its put in the table and now the code searches for patients with a lower pid and finds two others (correctly this time).

similarly the patient Monday Rungn is really a duplicate of Monday Runagain, but the latter's pid is 24

### Expected behavior
dup patient bug correct

### Client configuration

- Browser:  firefox
- OpenEMR version:  7.0.2-dev
- Operating system: OpenSuse Tumbleweed

ruthkonyn:
Fixes #6517

resolves incorrect reporting of potentially duplicate patients in admin/patients/manage duplicates, particularly when a patients demographics are changed
when a patient's demographics are changed generate it's dupscore by looking at all patients in the db, not just those with a smaller pid
when deciding which patients to display in the table do not display any who do not have a potential duplicate in the data base, i.e. would be in a group of 1, i.e. on their own

### DIFF
--- a/interface/patient_file/manage_dup_patients.php
+++ b/interface/patient_file/manage_dup_patients.php
@@ -262,7 +262,7 @@ if ($form_action == 'U') {
         array($_POST['form_toppid'])
     );
 } else if ($form_action == 'R') {
-    updateDupScore($_POST['form_toppid']);
+    updateDupScore($_POST['form_toppid'], false); /* this is not a new patient so check against all other patients */
 }
 
 $query = "SELECT * FROM patient_data WHERE dupscore > 7 " .
@@ -272,7 +272,10 @@ while ($row1 = sqlFetchArray($res1)) {
     displayRow($row1);
     $query = "SELECT p2.*, ($scorecalc) AS myscore " .
     "FROM patient_data AS p1, patient_data AS p2 WHERE " .
-    "p1.pid = ? AND p2.pid < p1.pid AND ($scorecalc) > 7 " .
+ /*   "p1.pid = ? AND p2.pid < p1.pid AND ($scorecalc) > 7 " .  */
+ /* check against all other patients, to fix bug where a lone patient appears in the table */
+    "p1.pid = ? AND p2.pid != p1.pid AND ($scorecalc) > 7 " .
+
     "ORDER BY myscore DESC, p2.pid DESC";
     $res2 = sqlStatement($query, array($row1['pid']));
     while ($row2 = sqlFetchArray($res2)) {
--- a/interface/patient_file/manage_dup_patients.php
+++ b/interface/patient_file/manage_dup_patients.php
@@ -262,24 +262,39 @@ if ($form_action == 'U') {
         array($_POST['form_toppid'])
     );
 } else if ($form_action == 'R') {
-    updateDupScore($_POST['form_toppid'], false); /* this is not a new patient so check against all other patients */
+    updateDupScore($_POST['form_toppid']); /* this is not a new patient so check against all other patients */
 }
 
+/* generate a new table of possible duplicate patients */
 $query = "SELECT * FROM patient_data WHERE dupscore > 7 " .
     "ORDER BY dupscore DESC, pid DESC LIMIT 100";
 $res1 = sqlStatement($query);
+$used = array(); /* record which patients we display in the table, so as not to use them more than once */
 while ($row1 = sqlFetchArray($res1)) {
-    displayRow($row1);
-    $query = "SELECT p2.*, ($scorecalc) AS myscore " .
-    "FROM patient_data AS p1, patient_data AS p2 WHERE " .
- /*   "p1.pid = ? AND p2.pid < p1.pid AND ($scorecalc) > 7 " .  */
- /* check against all other patients, to fix bug where a lone patient appears in the table */
-    "p1.pid = ? AND p2.pid != p1.pid AND ($scorecalc) > 7 " .
-
-    "ORDER BY myscore DESC, p2.pid DESC";
-    $res2 = sqlStatement($query, array($row1['pid']));
-    while ($row2 = sqlFetchArray($res2)) {
-        displayRow($row2, $row1['pid']);
+    global $used;
+    error_log ("p1 & score ".$row1['pid']." , ". $row1['dupscore']);
+    if (array_search($row1['pid'],$used) === false){ //haven't used this one yet
+    // build the sql query to compare to all other patients
+        $query = "SELECT p2.*, ($scorecalc) AS myscore " .
+        "FROM patient_data AS p1, patient_data AS p2 WHERE " .
+  /* check against all other patients - e.g. when demogrphics changed */
+        "p1.pid = ? AND p2.pid != p1.pid AND ($scorecalc) > 7 " .
+        "ORDER BY myscore DESC, p2.pid DESC";
+        $res2 = sqlStatement($query, array($row1['pid']));
+    /*only if more patients are potential duplicates display the first one -correct bug where only 1 patient in agroup displayed*/
+        $displayFirst = true;
+        while ($row2 = sqlFetchArray($res2)) {
+            error_log ("p2 & score ".$row2['pid']." , ".$row2['dupscore']);
+            if ($row2['dupscore'] != -1) { // don't use if marked as unique
+                if ($displayFirst )
+                {
+                    displayRow($row1);
+                    $displayFirst = false;
+                }
+                displayRow($row2, $row1['pid']);
+                $used[] = $row2['pid'];
+            }
+        }
     }
 }
 ?>
--- a/interface/patient_file/merge_patients.php
+++ b/interface/patient_file/merge_patients.php
@@ -567,8 +567,8 @@ if (!AclMain::aclCheckCore('admin', 'super')) {
                 resolveDuplicateEncounters($targets);
             }
 
-            // Recompute dupscore for target patient.
-            updateDupScore($target_pid);
+            // Recompute dupscore for target patient. check against all patients
+            updateDupScore($target_pid, false);
 
             echo "<br />" . xlt('Merge complete.') . "</div>";
 
--- a/library/patient.inc.php
+++ b/library/patient.inc.php
@@ -14,6 +14,8 @@
  * @copyright Copyright (c) 2018-2021 Stephen Waite <stephen.waite@cmsvt.com>
  * @copyright Copyright (c) 2021-2022 Rod Roark <rod@sunsetsystems.com>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ *
+ * update duplicate score modified so as to do this when a patients demographics are changed, as well as for new patients
  */
 
 use OpenEMR\Common\Uuid\UuidRegistry;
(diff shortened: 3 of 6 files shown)


interface/patient_file/merge_patients.php
```diff
@@ -567,8 +567,8 @@ function resolveDuplicateEncounters($targets): void
                 resolveDuplicateEncounters($targets);
             }
 
-            // Recompute dupscore for target patient.
-            updateDupScore($target_pid);
+            // Recompute dupscore for target patient. check against all patients
+            updateDupScore($target_pid, false);
```
adunsulag, line 571: It doesn't look like the method takes in a second parameter here.

library/patient.inc.php
```diff
@@ -1820,21 +1824,25 @@ function is_patient_deceased($pid, $date = '')
...
+// This computes, sets and returns the dup score for the given patient, compared to all patients earlier in the patient_data table
+// for updateing after patients demographics have been changed need to check against whole table, i.e. patients who come earlier and later in the table, i.e. where p2 != p1, this works for new patients as well
 //
 function updateDupScore($pid)
 {
+        // new patient or patient with editted demographics
     $row = sqlQuery(
         "SELECT MAX(" . getDupScoreSQL() . ") AS dupscore " .
         "FROM patient_data AS p1, patient_data AS p2 WHERE " .
-        "p1.pid = ? AND p2.pid < p1.pid",
```
adunsulag, line 1830: @bradymiller This makes sense to check the whole table.  Was there a reason you wrote the code to only check against earlier pids.

library/patient.inc.php
```diff
@@ -1820,21 +1824,25 @@ function is_patient_deceased($pid, $date = '')
...
+        "p1.pid = ? AND p2.pid != p1.pid",
         array($pid)
     );
+
     $dupscore = empty($row['dupscore']) ? 0 : $row['dupscore'];
     sqlStatement(
         "UPDATE patient_data SET dupscore = ? WHERE pid = ?",
         array($dupscore, $pid)
     );
+     error_log ("updateDupScore: pid =".$pid.", new is ".($new ? "true" : "false") . ", resultant dupscore is ".$dupscore);
```
adunsulag, line 1845: I would recommend doing the following as this looks like debug information.  I would do the same thing for the other error logs you have in this PR.
(new SystemLogger())->debug("updateDupScore: pid =".$pid.", new is ".($new ? "true" : "false") . ", resultant dupscore is ".$dupscore);

src/Services/PatientService.php
```diff
@@ -211,6 +211,7 @@ public function insert($data)
      */
     public function databaseUpdate($data)
     {
+        error_log("$$$ databaseupdate called $$$");
```
adunsulag, line 214: Either remove this line or turn it into a debug as I mentioned in my other comment.

interface/patient_file/manage_dup_patients.php
```diff
@@ -262,21 +262,39 @@ function selchange(sel, toppid, rowpid) {
...
 while ($row1 = sqlFetchArray($res1)) {
-    displayRow($row1);
-    $query = "SELECT p2.*, ($scorecalc) AS myscore " .
-    "FROM patient_data AS p1, patient_data AS p2 WHERE " .
-    "p1.pid = ? AND p2.pid < p1.pid AND ($scorecalc) > 7 " .
-    "ORDER BY myscore DESC, p2.pid DESC";
-    $res2 = sqlStatement($query, array($row1['pid']));
-    while ($row2 = sqlFetchArray($res2)) {
-        displayRow($row2, $row1['pid']);
+    global $used;
```
adunsulag, line 274: Shouldn't have to have global here as its in context on line 272.

interface/patient_file/manage_dup_patients.php
```diff
@@ -262,21 +262,39 @@ function selchange(sel, toppid, rowpid) {
...
+    error_log ("p1 & score ".$row1['pid']." , ". $row1['dupscore']);
```
adunsulag, line 275: Again make this a debug or remove the error log.

interface/patient_file/manage_dup_patients.php
```diff
@@ -262,21 +262,39 @@ function selchange(sel, toppid, rowpid) {
...
+        while ($row2 = sqlFetchArray($res2)) {
+            error_log ("p2 & score ".$row2['pid']." , ".$row2['dupscore']);
+            if ($row2['dupscore'] != -1) { // don't use if marked as unique
+                if ($displayFirst )
+                {
+                    displayRow($row1);
+                    $displayFirst = false;
+                }
+                displayRow($row2, $row1['pid']);
+                $used[] = $row2['pid'];
```
adunsulag, line 295: if you do the following $used['pid'] = $row2['pid'] you can remove the array search on line 276 and just change it to be the following:
```
if (!empty($used[$row2['pid']])) {
}
```
This will be more performant since you're going to be working with a bunch of patients.

interface/patient_file/manage_dup_patients.php
```diff
@@ -262,21 +262,39 @@ function selchange(sel, toppid, rowpid) {
...
+        $query = "SELECT p2.*, ($scorecalc) AS myscore " .
+        "FROM patient_data AS p1, patient_data AS p2 WHERE " .
+  /* check against all other patients - e.g. when demogrphics changed */
+        "p1.pid = ? AND p2.pid != p1.pid AND ($scorecalc) > 7 " .
+        "ORDER BY myscore DESC, p2.pid DESC";
+        $res2 = sqlStatement($query, array($row1['pid']));
+    /*only if more patients are potential duplicates display the first one -correct bug where only 1 patient in agroup displayed*/
+        $displayFirst = true;
+        while ($row2 = sqlFetchArray($res2)) {
+            error_log ("p2 & score ".$row2['pid']." , ".$row2['dupscore']);
```
adunsulag, line 287: make this a debug log or remove it.

adunsulag (changes requested):
Thanks for the PR.  Couple changes, but overall good work, I'll let @bradymiller chime in on the one line he wrote.

interface/patient_file/manage_dup_patients.php
ruthkonyn: thanks that's useful - yes we do have quite large data sets!

ruthkonyn:
@adunsulag hi stephen, please see my expanded comment on the changes that you requested, thanks.

library/patient.inc.php
bradymiller: I just migrated this whole script from patient.inc to patient.inc.php. Looks like @sunsetsystems created this function:
https://github.com/openemr/openemr/blame/rel-700/library/patient.inc

this PR looks good to me if testing well

sunsetsystems: Dupscore was defined/designed to be a comparison only with earlier pids. If you don't like that definition then you need to recompute scores for all pids on every demographics update, and that would be computationally prohibitive. More practical, I think, to run a nightly or weekly process to recompute all scores.

ruthkonyn: @sunsetsystems hi rod. when a patient's demographic are changed their duplicate score is recalculated looking at all the other patients, but this is done only for the patient whose demographics were changed. I don't see a need to do it for every patient in the system at this point.

would it help to have two buttons on editing a patient's demographics: one 'save' and one 'save and update duplicate score' (or something) ? (obviously the save wouldn't do any recalculations). 

if this is useful @adunsulag and @sunsetsystems  then let me know and i'll implement it.

sunsetsystems: Hi Ruth. As I noted you can't recompute everything on each demographics update. In a somewhat large database that can take hours. This is why I suggest a nightly or weekly process. Also for comparing every patient with every other patient it's sufficient for such a process to compare only with lesser pids, which cuts the time in half. You can do what you originally suggested but I think it's better to have a more complete solution where a full recomputation is done before a manual session of managing duplicates. I hope that makes sense.

ruthkonyn: hi rod, i think we are still talking at cross purposes. what we find works is that when a patient's info is changed to compare just that one patient with all the others, in order to recompute their dup score. comparing to only those with lower pids has proved not to work: If the change of a patient makes them a dup of a patient with a higher pid this won't show, however comparing that one patient to all in the data base works in our situation, with no need to recompare every patient with every other (which i think is your suggestion). This is why i suggested the buttons, because then one can use the technique that suites one's own configuration. 

interestingly, what i suggest above is no more compute heavy than when a new patient is added to the system - when an automatic comparison with the whole database is carried out in order to compute their initial dup score. 

perhaps we also need a 'compare all' button/function somewhere which does what you suggest, i.e. recompute the dup score for every patient in the system, which as you rightly say, could take a long time depending on the size of the db. If people are editing patient's data without recomputing ('save' button)  that patient's dup score, then this 'compare all' function would be needed - or some way to run it via cron, as an example.

sunsetsystems: Hi Ruth, the problem with your solution is that it will potentially created "orphaned" instances of high score values that the GUI tool will not show the matches for. It may help you in some cases but it's not robust.

See https://github.com/openemr/openemr/pull/4502 for the original PR and details of the design. The correct way to solve your problem is by periodically running dupscore.cli.php which is also noted in that PR. Yes it works with cron.

ruthkonyn: oh that's interesting, because one of the bugs we found was exactly that - orphaned instances of high score values that were turning up in the gui tool (manage duplicates). so far my new code has not thrown up any, but we are still testing. I'll take a look at issue #4502 - thanks for the lead.

sunsetsystems: As I recall when a patient is marked unique, that is with respect to previously created patients (lower pids). They are still fair game to be compared with more recently created patients.

But more importantly, testing during production is not enough to assure that something does or doesn't work. Furthermore you must run dupscore.cli.php initially, and periodically after that to ensure accuracy.

robertdown:
My only suggestion here is that if we are re-scoring upon a save (which I think is a decent idea) we should not have a separate button - it should be a configuration state for what the button does. Just 1 save button, which either rescores or does not rescore (based on the Admin config).

ruthkonyn:
thanks @robertdown - that seems a really sensible idea - i'm going to be busy with holidays and some more pressing development work for the next few weeks, but i'll implement a configuration option to recalculate on save when i get the time. (unless there are opinions against it?)

adunsulag:
@ruthkonyn what would you like to do here? I"m not sure we ever addressed the issues that @sunsetsystems brought up.  Do you want to keep working on this or should we close the PR?

kojiromike:
This has been addressed in #10644 with a cleaner implementation that also prevents orphan records in the display (addressing Rod's concern from the original review).

PR #6524, opened 2023-06-08
Added scss-migrator

stevenpjohnso:
This hopefully resolves the scss depreciation warning messages. In my testing, instead of many warning messages, `docker-compose up` now ends at couchdb cluster stable. 

Add scss-migrator to package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "moment": "2.29.4",
     "numeral": "2.0.6",
     "purecss": "3.0.0",
+    "sass-migrator": "1.7.3",
     "select2": "4.0.13",
     "sortablejs": "1.15.0",
     "underscore": "1.13.6",


adunsulag:
@stevenpjohnso I would think this should be a dev dependency and not a regular dependency.  We only want the scss to be compiled on development.

stephenwaite:
but I think we compile as well if someone choose to deploy in production with:
```
composer install --no-dev
npm install
npm run build
composer dump-autoload -o

stephenwaite:
the scss-migrator may be an option but it's failing in this PR

Screenshot from 2023-06-08 10-43-26

sjpadgett:
I couldn't get migrator to work for me. Upstream is killer.

stephenwaite:
thanks for bringing this up @stevenpjohnso , Jerry slayed the dragon in #6543 :)

PR #6540, opened 2023-06-11
Display correct portal login name

mdsupport:
Fixes portal login user name display.


sjpadgett:
If you want @mdsupport I can add to my portal PR bringing in later tonight. Would save you some headache fooling with tests.
Unless you have more!

mdsupport:
Nope. Just wanted staff to be able to help patients when they forget their username.

PR #6549, opened 2023-06-12
Master backports for patch 2

sjpadgett:
- this may get tricky!

Fixes #

Start of backporting for next patch.
I want to have questionnaire SMART and smoother forms workflow in next patch.

### DIFF
--- a/_rest_routes.inc.php
+++ b/_rest_routes.inc.php
@@ -18,7 +18,7 @@
  */
 
 /**
- *  @OA\Info(title="OpenEMR API", version="7.0.1")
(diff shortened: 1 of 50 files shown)


sjpadgett:
I'm not sure who worked on the portals Reports->Report Content but it looks very nice and seems to work much better so thanks, good job.
I think I need to rename the item though. Like My Report Generator or such.

sjpadgett:
I better bring this in as it is getting out of hand file wise plus the rate @stephenwaite is fixing bugs we'll soon run into merge conflicts.
I'll be testing more on this today and hope @juggernautsei will also.

I'm beginning to get excited about portal again!:)

PR #6560, opened 2023-06-15
More portal

sjpadgett:
- restyle main portal login to be responsive in portrait
- add select to auto save
- alert patient for unsaved edits if in edit and refresh page

Fixes #


sjpadgett:
will test on demo in a,m

bradymiller:
dance-bear

PR #6567, opened 2023-06-16
Fixes #6566 Lforms version update, copyright fix

### LINKED ISSUE #6566: Fix LForms Questionnaire Copyright Copy issue
Right now the lforms copyright property does not get populated in a questionnaire when the lform is converted to a questionnaire.  Any imported questionnaire using our loinc lookup tool ends up not saving the copyright for the questionnaire.

I've pushed a fix upstream in the lforms here: https://github.com/lhncbc/lforms/pull/130

In the meantime, we should update out lforms version and bring in the copyright fix into the OpenEMR installation.

adunsulag:
Updates the lforms version to the latest one and implements the copyright fix for it.

Fixes #6566


bradymiller (approved):

adunsulag:
There is a bug in the latest lforms that will create an infinite JS loop and freeze the browser on certain SDC conditions.  We are going to hold off on bringing this in until that bug as been fixed.

PR #6577, opened 2023-06-17
Openemr fix 6574 6575 6576 fhir updates

### LINKED ISSUE #6574: Fix QuestionnaireResponse search operation to deal with duplicate field ids in where clause
The questionnaire response service's search operation has problems when searching on fields due to duplicate field ids.

### LINKED ISSUE #6575: Make it possible to save OAuth2 client outside of AuthorizationController.php
We should refactor the code for inserting a OAuth2 client record into the codebase from AuthorizationController into the ClientRepository so OpenEMR consumers can register OAuth2 clients if needed.  We should have the AuthorizationController call this new class.

### LINKED ISSUE #6576: Update FHIR Capability statement to indicate we do not allow resources to be created with an update operation
We need to tell API consumer sin our capability statement for FHIR that we do not allow api consumers to provide us their own resource logical id to be used when creating a resource.  IE, if someone sends us a resource that contains an FHIRResource.id property as part of a create operation, we will ignore the passed in id and generate our own.

adunsulag:
Fixes #6574 
The questionnaire response search would fail when searching on things
such as id due to duplicate column names.  In order to make this service
work in a FHIR context we need to fix up the search method and handle
the uuid translations.

Fixes #6575 
Made it possible to save an oauth2 client using the client repository.
This makes it so OpenEMR callers don't need to include the RestConfig
class or deal with the AuthorizationController to generate a smart app
client record if needed.  Module writers can register their own smart
app as needed using this functionality.

Fixes #6576 
made it so the create update flag is set to false so API consumers know
they can't provider their own logical ids when creating a resource.

Also had some minor utility methods added in preparation for future FHIR work with Task/Questionnaire endpoints.

### DIFF
--- a/src/Common/Auth/OpenIDConnect/Repositories/ClientRepository.php
+++ b/src/Common/Auth/OpenIDConnect/Repositories/ClientRepository.php
@@ -16,6 +16,8 @@ use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
 use OpenEMR\Common\Auth\OpenIDConnect\Entities\ClientEntity;
 use OpenEMR\Common\Crypto\CryptoGen;
 use OpenEMR\Common\Logging\SystemLogger;
+use OpenEMR\Common\Utils\HttpUtils;
+use OpenEMR\Common\Utils\RandomGenUtils;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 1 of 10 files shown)


bradymiller (approved):

bradymiller:
tenor99

PR #6592, opened 2023-06-20
feat: Fixes #6591 Appointment Event Firing

### LINKED ISSUE #6591: feat: Add Delete and Render Event Hooks to appointment add screen
We fire off events when an appointment is set(created/inserted) but we do not fire off any events if an appointment is deleted.  This makes it challenging for module writers to cleanup data if an appointment is removed from the calendar (or via the API). We should fire an event when an appointment is deleted.

We should also fire an event that lets module writers add to the screen above the add/edit appointment dialog command bar (button bars at the bottom of the form).

adunsulag:
Added a ServiceDeleteEvent and refactored the add_edit_event script to fire off both the pre and post service delete events.

Added an event to add content at the bottom of the add_edit_event page.

Did some minor documentation cleanup on the service save event.
Fixes #6591


sjpadgett:
Seriously you do nice work

bradymiller (approved):

bradymiller:
2DV(1)

PR #6594, opened 2023-06-20
Add SMS event to Secure Messaging

### LINKED ISSUE #6593: Add event to send SMS from Secure Messaging
### Describe the problem
Have been wanting this and is main reason I did the SMS module in the first place.
So from portal dashboard Secure Message you get a button to popup dialog to send to currently selected patient or use the Patient dialog to select any patient.

image

sjpadgett:
- support sending SMS to patient from messaging.

Fixes #6593

### DIFF
--- a/src/Common/Session/SessionUtil.php
+++ b/src/Common/Session/SessionUtil.php
@@ -75,20 +75,22 @@ class SessionUtil
     {
         // Note there is no system logger here since that class does not
         //  yet exist in this context.
-        session_start([
-            'read_and_close' => $read_only,
-            'cookie_samesite' => self::$use_cookie_samesite,
-            'cookie_secure' => self::$use_cookie_secure,
-            'name' => 'OpenEMR',
-            'cookie_httponly' => false,
-            'cookie_path' => ((!empty($web_root)) ? $web_root . '/' : '/'),
-            'gc_maxlifetime' => self::$gc_maxlifetime,
-            'sid_bits_per_character' => self::$sid_bits_per_character,
-            'sid_length' => self::$sid_length,
-            'use_strict_mode' => self::$use_strict_mode,
-            'use_cookies' => self::$use_cookies,
-            'use_only_cookies' => self::$use_only_cookies,
-        ]);
+        if (session_status() === PHP_SESSION_NONE) {
+            session_start([
+                'read_and_close' => $read_only,
+                'cookie_samesite' => self::$use_cookie_samesite,
+                'cookie_secure' => self::$use_cookie_secure,
+                'name' => 'OpenEMR',
+                'cookie_httponly' => false,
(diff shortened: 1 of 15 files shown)


src/Common/Session/SessionUtil.php
```diff
@@ -75,20 +75,22 @@ public static function coreSessionStart($web_root, $read_only = true): void
...
-            'cookie_httponly' => false,
-            'cookie_path' => ((!empty($web_root)) ? $web_root . '/' : '/'),
-            'gc_maxlifetime' => self::$gc_maxlifetime,
-            'sid_bits_per_character' => self::$sid_bits_per_character,
-            'sid_length' => self::$sid_length,
-            'use_strict_mode' => self::$use_strict_mode,
-            'use_cookies' => self::$use_cookies,
-            'use_only_cookies' => self::$use_only_cookies,
-        ]);
+        if (session_status() === PHP_SESSION_NONE) {
```
bradymiller, line 78: This kind of thing makes me nervous (i think then any session that is not core could circumvent core). What is the use case for this?

sjpadgett: I suppose one could start a session and prevent starting a named session. Regardless if you can create a session, you can create one using the OpenEMR name. Still whichever session is active, all that will happen is a noisy PHP notice.
The issue stems from our setSession() method does a session start  to allow write mode since we default to read only. https://github.com/openemr/openemr/blob/61e4f5bc1d0d5c12661ad2d46b1d398426d138ca/src/Common/Session/SessionUtil.php#L94-L97 

Now we assume that since we set the read_and_close option that the session will be closed when attempting a session write and the session will be reopened in a write mode. Not always true plus if indeed the session is still open it may or may not mean the session file wasn't written from a previous write open with a session_write_close().
This may explain some of our random session timeout issues.

My point is we need to ensure our session state when possible and if a session is open when calling the core session start then we possibly missed a session save. This started with me getting rid of annoying php warning which in my case is caused by doing a session write from module though I still use the session util i.e
```
public function setSession($key, $value): static
    {
        // ensure write is allowed by using utility.
        SessionUtil::setSession($key, $value);
        return $this;
    }
```
So unknow why the wrong session state.

I think a better solution is
```
if (session_status() === PHP_SESSION_ACTIVE) {
            session_write_close();
        }
```

Furthermore it is time we implement our own session manager because we are beyond what the basic PHP handler is designed for!

bradymiller (approved):

sjpadgett:
Since that small change I am unable to re-create the session errors. I’d be willing to bet that it will also get rid of the random session timeout since session manager writes the timer value to sessions.
It be nice if somehow I could log when we hit an open session to keep an eye on it.
Thanks for the review Brady.
@adunsulag Do you need this merged for anything you are doing? If not I may leave open for today

PR #6602, opened 2023-06-22
Allow repeater services to be turned off

### LINKED ISSUE #6601: Add new environment variable for repeater services
### Describe the problem
Debugging app can be a pain when dated reminders and background service are running.
Will introduce new var OPENEMR__NO_BACKGROUND_TASKS to turn them off regardless if dev mode is set.
i.e. may just want one or both.

sjpadgett:
add new environment debug variable to turn off dated reminders and background  service from running.

Fixes #6601 

Decided to keep as separate PR


bradymiller:
2DV(1)

PR #6603, opened 2023-06-22
fix: Use bootstrap for list items on multi sorted list options

robertdown:
Use bootstrap's list-group on the sortable list.


stephenwaite:
southpark_cheering_edited

PR #6604, opened 2023-06-22
feat: Enhanced Default Open Tabs Options

robertdown:
Converts the current "Main top pane screen" and "Default second tab" globals into "Default Open Tabs" which is powered by the new default_open_tabs List. This allows users to set as many tabs to be open by default as they want and sort them as they see fit. Developers can simply add the url and title to teh Default Open Tabs list to allow their pages to be open by default.


robertdown:
Want to make this a bit more flexible, keeping as a draft for now

robertdown:
Ok, dropping the use of globals entirely here. Instead, the Default Open Tabs list is used, any active row will be open by default. Just need to touch up the upgrade SQL to migrate the current globals value to the list table active state.

robertdown:
Ready for a review, testing well on my end. Transition is seemless; the sql update script will create a new list, populate it with the current options from the globals, and then only activate the 2 tabs that could be set (Calendar and Messages by default).

To manage - go to Admin > Forms > Lists and select "Default Open Tabs." This list controls which tabs will be opened by default. For now it is an all or nothing deal, no logic is applied. Only **Active** rows of the list will be rendered. Modules can write to this list to allow their own pages to automatically be open.

**Future Enhancement Ideas**
1. Restore tabs based on what was open when a session is logged out
2. Open a set of tabs based on the ACL of the user
3. Allow more than 1 tab to be open by default

sql/7_0_1-to-7_0_2_upgrade.sql
```diff
@@ -216,3 +216,15 @@ ALTER TABLE `facility` ADD COLUMN `inactive` tinyint(1) NOT NULL DEFAULT '0';
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`, `toggle_setting_1`, `toggle_setting_2`, `activity`) VALUES ('Document_Template_Categories','notification_template','Notification Template',105,0,0,'','','',0,0,1);
 #EndIf
 
+#IfNotRow2D list_options list_id lists option_id default_open_tabs
+-- Create new list Default Open Tabs
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`) VALUES ('lists', 'default_open_tabs', 'Default Open Tabs');
```
bradymiller, line 221: would increment seq by 10 to provide some flexibility if add something in future

sql/7_0_1-to-7_0_2_upgrade.sql
```diff
@@ -216,3 +216,15 @@ ALTER TABLE `facility` ADD COLUMN `inactive` tinyint(1) NOT NULL DEFAULT '0';
...
+-- Populate list with sensible defaults based on previous optiosn in globals, assume no tabs are actually active
+INSERT INTO `list_options` (`list_id`, `notes`, `title`, `seq`, `option_id`, `activity`) VALUES ('default_open_tabs', 'main_info.php', 'Calendar', 1, 'cal', '0');
+INSERT INTO `list_options` (`list_id`, `notes`, `title`, `seq`, `option_id`, `activity`) VALUES ('default_open_tabs', '../new/new.php', 'Patient Search / Add', 2, 'pat', '0');
+INSERT INTO `list_options` (`list_id`, `notes`, `title`, `seq`, `option_id`, `activity`) VALUES ('default_open_tabs', '../../interface/main/finder/dynamic_finder.php', 'Patient Finder', 3, 'fin', '0');
```
bradymiller, line 225: this opens door to directory traversal vulnerability. not sure how to mitigate this. gotta think about it for a bit (likely involves ensuring the flattened path with realpath() function does not lie outside the `$GLOBALS['fileroot']` path).

bradymiller:
Looks like the functional testing is failing

sql/7_0_1-to-7_0_2_upgrade.sql
```diff
@@ -216,3 +216,15 @@ ALTER TABLE `facility` ADD COLUMN `inactive` tinyint(1) NOT NULL DEFAULT '0';
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`, `toggle_setting_1`, `toggle_setting_2`, `activity`) VALUES ('Document_Template_Categories','notification_template','Notification Template',105,0,0,'','','',0,0,1);
```
bradymiller, line 216: won't the need to be added to database.sql ? (with the 2 current active entries set)

sql/7_0_1-to-7_0_2_upgrade.sql
robertdown: Yep, good call. I'll work on solving this as well.

sql/7_0_1-to-7_0_2_upgrade.sql
robertdown: Not sure, I didn't touch anything with Facilities?

sql/7_0_1-to-7_0_2_upgrade.sql
stephenwaite: it's these that need to be added to `sql/database.sql`

sql/7_0_1-to-7_0_2_upgrade.sql
robertdown: Done

sql/7_0_1-to-7_0_2_upgrade.sql
robertdown: Added to database.sql as well. Address the vulnerability by concatenating the file root to the URL and then running it through `realpath()` - if that returns false that row in the array is removed and we move on; otherwise, the path does exist (within `$fileroot` so we can move forward.

I tested this by creating a new entry in the list called "Test" with a URL set to "/var/log/apache2/error.log" which doesn't survive the load.

For better control, when inserting the row in the list, the URL should assume it starts at the file path (no need to to `../../` for any reason).

robertdown:
1. Security risk mitigated
2. Ensured both upgrades and new installs get the proper list values
3. Functional testing was passing on my local docker

sql/database.sql
```diff
@@ -13314,3 +13314,10 @@ CREATE TABLE `patient_settings` (
      `setting_value` varchar(255) NOT NULL DEFAULT '',
      PRIMARY KEY (`setting_patient`, `setting_label`)
 ) ENGINE=InnoDB;
+
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`) VALUES ('lists', 'default_open_tabs', 'Default Open Tabs');
+INSERT INTO `list_options` (`list_id`, `notes`, `title`, `seq`, `option_id`, `activity`) VALUES ('default_open_tabs', 'interface/main_info.php', 'Calendar', 10, 'cal', '1');
```
bradymiller, line 13319: @robertdown , I think there is a flaw here if we assume that the target/option_id will be unique. If peruse the main menu json, there are lots of duplicates in target to ensure they populate the same frame (sometimes necessary to ensure never open up more than 1 patient at a time and other times necessary to avoid opening up a huge number of tabs).

bradymiller: actually @robertdown , i wouldn't worry about this issue now. It is something can address in future if ever run into this issue in the future.

interface/main/tabs/main.php
```diff
@@ -302,15 +303,25 @@ function assignPatientDocuments(patientId) {
...
-        <?php if (!empty($_SESSION['frame2url']) && !empty($_SESSION['frame2target'])) { ?>
-            // Use session variables and tabStatus object to set up initial/default second tab, if none is set in globals, this tab will not be displayed initially
-            app_view_model.application_data.tabs.tabsList.push(new tabStatus(<?php echo xlj("Loading"); ?> + "...", <?php echo json_encode("../" . $_SESSION['frame2url']); ?>, <?php echo json_encode($_SESSION['frame2target']); ?>, <?php echo xlj("Loading"); ?> + " " + <?php echo json_encode($_SESSION['frame2label']); ?>, true, false, false));
-        <?php } ?>
+        <?php
+        if ($_SESSION['default_open_tabs']) :
+            // For now, only the first tab is visible, this could be improved upon by further customizing the list options in a future feature request
+            $visible = "true";
+            foreach ($_SESSION['default_open_tabs'] as $i => $tab) :
+                $_unsafe_url = preg_replace('/(\?.*)/m', '', Path::canonicalize($fileroot . DIRECTORY_SEPARATOR . $tab['notes']));
```
bradymiller, line 311: Feel silly for asking this. But where is $fileroot defined (I am guessing it is a PHP thing maybe)?

interface/main/tabs/main.php
```diff
@@ -302,15 +303,25 @@ function assignPatientDocuments(patientId) {
...
+                if (realpath($_unsafe_url) === false) {
```
bradymiller, line 312: How is above line making sure that $_unsafe_url is within the `$GLOBALS['fileroot']` directory or subdirectory etc.?
(it looks like it is just making sure the file exists wherever it may be)

sql/database.sql
robertdown: A lot of tabs require a patient be selected, which wouldn't happen by default.

robertdown:
Between the two checks I feel confident in bringing this in. I tested by adding these two paths as URLs in the list`/var/log/apache2/error.log` as well as `/var/www/localhost/htdocs/index.html` and neither would fire the tab. Not even a 404, these paths were completely removed from the tab list

bradymiller:
2DV-2(6)

PR #6607, opened 2023-06-23
Bug fixes while testing

sjpadgett:
- add use include for rest_config class in api authorization controller
- remove key so it can regenerate if key is empty or corrupted.

Fixes #

Just need PR for various fixes as I come across them.
I'll create an issue if I come across something noteworthy

### DIFF
--- a/src/Common/Auth/OAuth2KeyConfig.php
+++ b/src/Common/Auth/OAuth2KeyConfig.php
@@ -85,8 +85,10 @@ class OAuth2KeyConfig
             // collect the encryption key from database
             $this->oaEncryptionKey = $this->cryptoGen->decryptStandard($eKey['value']);
             if (empty($this->oaEncryptionKey)) {
+                // delete corrupted key to regenerate on next attempt.
+                sqlStatementNoLog("DELETE FROM `keys` WHERE `name` = 'oauth2key'");
                 // if decrypted key is empty, then critical error and must exit
-                throw new OAuth2KeyException("oauth2 key was blank after it was decrypted");
+                throw new OAuth2KeyException("oauth2 key problem after decrypted. Key deleted, try again.");
             }
         } else {
             // create a encryption key and store it in database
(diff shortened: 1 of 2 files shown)


src/Common/Auth/OAuth2KeyConfig.php
```diff
@@ -85,8 +85,10 @@ public function configKeyPairs(): void
             // collect the encryption key from database
             $this->oaEncryptionKey = $this->cryptoGen->decryptStandard($eKey['value']);
             if (empty($this->oaEncryptionKey)) {
+                // delete corrupted key to regenerate on next attempt.
+                sqlStatementNoLog("DELETE FROM `keys` WHERE `name` = 'oauth2key'");
```
bradymiller, line 89: This will silently invalidate all prior tokens even if there is short term issue with reading of key from drive. A corrupt key seems like it should be rare and the actual key should never be blank. This likely stems from the key in drive not matching the keys in the database, so this would not fix that underlying problem (and would also have errors when trying to read other encrypted stuff from the database). There must be something in the testing environment that is doing this.
@adunsulag , I think I recall you noting a bug in the easy dev docker environment where the old keys in drive were not being removed when reset it. I wonder if that is what is causing @sjpadgett to see this issue?

sjpadgett: You make good points that I have thought about.
However if the key is broken APIs do not run so an invalidated token is not an issue.
On the other hand if the key is invalid because the file key has changed then a solution of using a backed up file key would fix the problem.

maybe a better solution is to emit a better error message.
In a dev environment I will go ahead and delete it because most likely this will happen during TV development.

in my case the file key is correct and all my other passwords etc. work so no clue.

thanks Brady

adunsulag: @bradymiller and @sjpadgett I'm all for better error messages.  

I wish we could figure out what is going on though on our dev environments w/ old keys, Jerry and I seem to hit this pretty frequently.  I know I hit it with the docker resets, but I don't know how much you do that @sjpadgett? 

 Deleting the key in a production instance I think would create quite a bit of havoc with anyone that has a refresh token.  I don't care so much about access tokens, but refresh tokens would be a major pain point.

sjpadgett: Yes but if the encryption key is invalid then nothing’s gonna get decrypted.
Am I missing something here

PR #6609, opened 2023-06-23
Changing the if conditions fixes the issue

### LINKED ISSUE #6608: Deceased or not??
### Describe the bug
There is nothing to do. The chart shows up as deceased today.

### To Reproduce
image

Open a chart this is automatic.

### Expected behavior
Not to show deceased unless a date is in the chart.

thumbnail_image

### Client configuration

- Browser: ALL
- OpenEMR version: 7.0.1
- Operating system:

juggernautsei:
When I dumped deceased, it showed an array.
array(2) { ["days_deceased"]=> NULL ["date_deceased"]=> string(0) "" }
And neither of these will match the if statement because NULL can't be greater than 0 and a string "0" can be compared with an integer. 
 

Fixes #6608

### DIFF
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -1013,8 +1013,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                 <div class="col-md-8">
                     <?php
 
-                    (int)$deceased_days = $deceased['date_deceased'];
-                    if ($deceased_days > 0) :
+                    if ($deceased['days_deceased'] !== null && $deceased['date_deceased'] !== "0") :
                         echo $twig->getTwig()->render('patient/partials/deceased.html.twig', [
                             'deceasedDays' => deceasedDays($deceased),
                         ]);
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -1013,7 +1013,8 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                 <div class="col-md-8">
                     <?php
 
-                    if ($deceased > 0) :
+                    (int)$deceased_days = $deceased['date_deceased'];
+                    if ($deceased_days > 0) :
                         echo $twig->getTwig()->render('patient/partials/deceased.html.twig', [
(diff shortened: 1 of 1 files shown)


interface/patient_file/summary/demographics.php
```diff
@@ -1013,7 +1013,7 @@ function setMyPatient() {
                 <div class="col-md-8">
                     <?php
 
-                    if ($deceased > 0) :
+                    if ($deceased['days_deceased'] !== null && $deceased['date_deceased'] !== "0") :
```
stephenwaite, line 1016: hi @juggernautsei , how about 
```
if (!empty($deceased)) :

juggernautsei: @stephenwaite, I tried it, and no. It still shows deceased on the chart.

bradymiller:
Would fix the is_patient_deceased() function would should return false when patient is alive (ie. $deceased should simply be false, so something going wrong in that is_patient_deceased() function which populates $deceased).

bradymiller:
For a patient that shows this error, what is the setting in database for the deceased_date field in patient_data?

juggernautsei:
This is the setting for the deceased date. 
deceased_reason | varchar(255) | utf8mb3_general_ci

stephenwaite:
> This is the setting for the deceased date. deceased_reason | varchar(255) | utf8mb3_general_ci

that's the reason the date should be
Screenshot from 2023-06-27 09-39-35

juggernautsei:
This system was an upgrade from v5 (5). So, that field was not updated.

stephenwaite:
huh, it's in the 4.0.0 to 4.0.1 upgrade sql

bradymiller:
That is deceased_reason, what is deceased_date? Also what is stored there for a patient that is showing the deceased message erroneously?

juggernautsei:
@bradymiller there is nothing stored in the database for the deceased date. There is nothing in the deceased reason either.

bradymiller:
What do you see in deceased_date for the bad patient when look in phpmyadmin (ie. is it NULL, blank, 0000-00-00, etc.) ?

juggernautsei:
Both fields are blank. 
image

PR #6611, opened 2023-06-23
feat: Fixes #6610 pdf refactor return document

### LINKED ISSUE #6610: feat: Patient portal pdf document return Document object instead of success message
We should refactor the bPatientPortalPDFDocumentCreator->createPdfDocument method to return an actual Document object instead of just a success message or empty string.  This allows consumers to get the id of the document record that was created and other information.

adunsulag:
Return the document object in the PatientPortalPDFDocumentCreator class.

Also type defined the document uuid method in the Document class.
Fixes #6610

### DIFF
--- a/library/classes/Document.class.php
+++ b/library/classes/Document.class.php
@@ -61,7 +61,7 @@ class Document extends ORDataObject
     public $id;
 
     /**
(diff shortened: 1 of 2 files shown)


bradymiller (approved):

bradymiller:
2DV-2(8)

Issue #6525, opened 2023-06-08
The client app should avoid re-authenticate during it’s launch within OpenEMR


YuvaReddy:
### Describe the bug
In OpenEMR, registered a client app and tried to launch it with in the hosted open EMR instance. Everytime during the launch it is redirecting to login page for user authentication instead of using the launch context of logged in openEMR user. It should request additional login and use SSO, similar to other EHR systems.

### To Reproduce
1. Register a client app (below is sample screenshot)
![image](https://github.com/openemr/openemr/assets/42370221/86f8af86-44eb-4632-98cf-97b87a2662ee)
2. Try to launch it from any patient demographic/dashboard (below is sample screenshot)
![image](https://github.com/openemr/openemr/assets/42370221/31af8113-335b-401d-86aa-6f0e09303f5f)

### Expected behavior
Current flow: Launch page → Login page (manually enter login details and Authorize) → Redirect page
Expected flow: Launch page → Redirect page
Can we bypass this authentication as the user have already authenticated into the system?

### Client configuration
Using Open EMR version 7.0.1 version Linux zip ([link 1](https://www.open-emr.org/wiki/index.php/OpenEMR_7.0.1_Linux_Installation))
Client App using FHIR-client JS library ([link 1](http://docs.smarthealthit.org/client-js/#smart-api))

adunsulag:
Are you expecting the login bypass if the offline_scope is requested?  OpenEMR doesn't currently support the online_access scope and we assume most applications will request the offline_access scope to avoid showing the scope authorization screen to users every time the app launches. 

As the offline_access scope is a very sensitive scope to grant (since it lasts for a year) we made the decision to force the login screen due to the higher security risk granting that scope gives. We debated this a lot when we built this a few years ago, we decided to force the login screen as we felt that most applications will require the offline_access scope.

I'm interested in your experience with other EMRs,  do they let you get the offline_access scope w/o requiring a login?  I was under the impression that Cerner required it but perhaps that has changed now.  I don't know if Epic bypasses it or not.

Or are you expecting the user to have to do the scope granting screen each time they launch the app since their authorization token only lasts an hour?

Please note that we would need to check with our certification lab to see if making a change like this requires re-certification.  If it does require re-certification on our g10 FHIR interface, then it will likely take some time to bring this change into the codebase.

YuvaReddy:
Hi @adunsulag 
[...]

PR #6628, opened 2023-07-01
feat: #6525 Use EHR authorization for SMART Apps

adunsulag:
First stab at #6525 by allowing the EHR to launch a smart app and reuse the authorization context that the currently logged in user has.  Keeping it as draft for now in order to discuss the security implications and for a followup commit to add the database tracking to prevent replay attacks.

I introduce a global that by default is off to allow OpenEMR administrators to enable smart apps to skip the authorization screen when the smart app is launched inside the EHR.

I refactored the launch sequence now to be an on-demand generation of the launch token using javascript instead of generating the token and stuffing it into the html.

I added a flag on the individual client app to turn on or off the authorization skip flag.  This gives admins control over allowing apps to launch immediately inside the EHR without requiring an authorization / authentication step.

I expanded the size of the SMART app window so that the app can take up more of the window and offer more functionality.  I'm wondering if we should offer different size options based upon the registered intent, or if there is some mechanism the app can communicate of how much real estate it wants to take up.

The app will receive any permissions it requests that was originally granted in its registration request.  The flow will continue to discard permissions that were not granted in the original registration request or were reduced in a subset of the refresh token.

Added to the launch token the currently authenticated user and the type of user.  Right now the ehr-launch is only inside the provider side of things, but I put in infrastructure so we can later extend it to launch inside the patient portal.

I originally relied on opening up the OpenEMR session cookie, but this doesn't work if the SMART app is hosted on a domain other than the domain OpenEMR is running under.  Browsers will drop the cookies on the browser redirect as its not considered a first-party origin so we have to rely on the launch token being the authority for the authentication / authorization.

I'm thinking that for more security in order to prevent replay attacks we need to store the generated launch token in the database, time expire it, and make it a one time use token.  If a repeat is detected we should invalidate all of the current user's access tokens and do some kind of notification to the administrator. I'm not sure yet if this is warranted, but it seems like we should treat this as serious as what we do with the refresh token.

### DIFF
--- a/sql/7_0_1-to-7_0_2_upgrade.sql
+++ b/sql/7_0_1-to-7_0_2_upgrade.sql
@@ -245,6 +245,9 @@ CREATE TABLE recent_patients (
     patients TEXT,
     PRIMARY KEY (user_id)
 ) ENGINE=InnoDB;
+
+#IfMissingColumn oauth_clients skip_ehr_launch_authorization_flow
+ALTER TABLE `oauth_clients` ADD COLUMN `skip_ehr_launch_authorization_flow` tinyint(1) NOT NULL DEFAULT '0';
 #EndIf
 
 #IfMissingColumn document_template_profiles notify_trigger
--- a/src/RestControllers/AuthorizationController.php
+++ b/src/RestControllers/AuthorizationController.php
@@ -64,8 +64,11 @@ use OpenEMR\Common\Utils\RandomGenUtils;
 use OpenEMR\Common\Uuid\UuidRegistry;
 use OpenEMR\FHIR\Config\ServerConfig;
 use OpenEMR\FHIR\SMART\SmartLaunchController;
+use OpenEMR\FHIR\SMART\SMARTLaunchToken;
 use OpenEMR\RestControllers\SMART\SMARTAuthorizationController;
+use OpenEMR\Services\BaseService;
 use OpenEMR\Services\TrustedUserService;
+use OpenEMR\Services\UserService;
 use OpenIDConnectServer\ClaimExtractor;
 use OpenIDConnectServer\Entities\ClaimSetEntity;
 use Psr\Http\Message\ResponseInterface;
@@ -480,6 +483,10 @@ class AuthorizationController
             $_SESSION['launch'] = $request->getQueryParams()['launch'] ?? null;
             $_SESSION['redirect_uri'] = $authRequest->getRedirectUri() ?? null;
             $this->logger->debug("AuthorizationController->oauthAuthorizationFlow() session updated", ['session' => $_SESSION]);
+            if (!empty($_SESSION['launch']) && $this->shouldSkipAuthorizationFlow($authRequest)) {
+                $this->processAuthorizeFlowForLaunch($authRequest, $request, $response);
+                exit;
+            }
             // If needed, serialize into a users session
             if ($this->providerForm) {
                 $this->serializeUserSession($authRequest, $request);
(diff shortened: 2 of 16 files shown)


adunsulag:
Couple of screenshots.
Smart apps now take up more real estate:
image

Added connector setting for allowing confidential apps to skip the authorization flow.  Default is to have this OFF.
 **OAuth2 EHR-Launch Authorization Flow Skip Enable App Setting**
image

Added a button to disable the EHR authorization flow for the app:
image

Current setting is shown on the left in the checkboxes area:
image

adunsulag:
So I took inspiration from our one time token implementation and was able to figure out a more secure approach than having the user id passed in the launch token.  Now if an in-ehr launch is requested, it will do a client side javascript post back to the server which allows the session cookies to be transmitted in a first-party domain handshake instead of treated as a third-party domain. 

By switching to a javascript submitted form on the authorize page to establish a first party context I was able to retrieve the session cookies and verify the user is logged in, even inside of a third party dialog.

If for some reason the third party cookie doesn't exist it brings up the standard login form.

The downside of this approach is a bit slower processing as an additional round trip to the client has to happen to submit the first party form.  Reviewers should pay special attention to the AuthorizationController::processAuthorizeFlowForLaunch() method where I am switching between the openemr session an oauth2 session.  This will have a slight performance hit as the switching of the session has to deal with brief session locking and filesystem access if that is how the php sessions are configured.  Redis session configurations will have less of a performance hit.

Overall though this eliminates the security loophole of someone grabbing the launch token via js or some other mechanism and logging in as a user just using the launch token.  Now we can tie it to the specific logged in session and close down that security issues that the prior approach was introducing.

adunsulag:
Looks like I need to figure out what is going on with the API tests.  My local environment is passing just fine on the api tests.  I'll test this out either tomorrow or Saturday on an external server or on a vanilla install. Odd.
```
openemr-cmd cst
```

sql/7_0_1-to-7_0_2_upgrade.sql
```diff
@@ -245,6 +245,9 @@ CREATE TABLE recent_patients (
     patients TEXT,
     PRIMARY KEY (user_id)
 ) ENGINE=InnoDB;
+
+#IfMissingColumn oauth_clients skip_ehr_launch_authorization_flow
+ALTER TABLE `oauth_clients` ADD COLUMN `skip_ehr_launch_authorization_flow` tinyint(1) NOT NULL DEFAULT '0';
```
bradymiller, line 250: needs to also be in database.sql

src/RestControllers/AuthorizationController.php
```diff
@@ -141,16 +145,28 @@ public function __construct($providerForm = true)
...
-        );
+
 
         $this->trustedUserService = new TrustedUserService();
     }
 
+    private function getSmartAuthController(): SMARTAuthorizationController
+    {
+        if (!isset($this->smartAuthController)) {
+            $twigContainer = new TwigContainer(__DIR__ . "/../../oauth2/", $GLOBALS['kenerl']);
```
bradymiller, line 156: looks like kernel is mispelled

bradymiller:
neat stuff!

tenor99

sql/7_0_1-to-7_0_2_upgrade.sql
adunsulag: This is what was causing the api test failures.  Wasn't revealed by doing tests against the demo population (openemr-cmd drid) but it is revealed with an openemr-cmd dri which does not populate the database and uses the database.sql file.

src/RestControllers/AuthorizationController.php
adunsulag: Fixed

sql/7_0_1-to-7_0_2_upgrade.sql
adunsulag: Fixed

adunsulag:
@bradymiller all the issues should be fixed, let me know if this looks good and I'll bring it in.

bradymiller (approved):

bradymiller:
@adunsulag , looks great!

PR #6630, opened 2023-07-01
feat: Fixes #6629 Encounter forms list render event

### LINKED ISSUE #6629: feat: Add Render event on forms.php page for encounter forms page.
Allow module writers to render content above or below the list of encounters in the forms.php page.  Eventually we'd like a render event to filter each individual encounter form, but this at least allows custom forms / content to be put on the page.

adunsulag:
Fixes #6629 add render events to top and bottom of encounter form.

Exposed the table name in the encounters service as well so we can actually use it instead of copying the name elsewhere.


bradymiller (approved):

bradymiller:
looks like couple php styling issues. otherwise looks great!

2DV(1)

PR #6633, opened 2023-07-01
feat: openemr fix #6631,#6632 twig datetime filters,multiple template path directories

### LINKED ISSUE #6631: feat: Add OpenEMR regular date time twig filter function
In our twig extensions we have a **shortDate** filter but we don't have any filter to display a full date.  We should add a filter that will render the date using the user's selected locale date fields.  The filter should take string, timestamps, and datetime objects.

### LINKED ISSUE #6632: feat: Allow multiple paths to be added to TwigContainer for template directories
Currently our TwigContainer class only allows a single path to be appended to the standard OpenEMR template directory search path.  We should allow multiple directories to be added.

adunsulag:
Fixes #6631 
Fixes #6632 

Add datetime filter for twig to respect OpenEMR global times.

Add multiple search paths to the twig template container.

### DIFF
--- a/src/Common/Twig/TwigContainer.php
+++ b/src/Common/Twig/TwigContainer.php
@@ -41,12 +41,17 @@ class TwigContainer
     {
         $this->paths[] = $GLOBALS['fileroot'] . '/templates';
         if (!empty($path)) {
-            $this->paths[] = $path;
+            $this->addPath($path);
         }
 
         if ($kernel) {
             $this->kernel = $kernel;
         }
+    }
+
+    public function addPath(string $path): void
+    {
+        $this->paths[] = $path;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


robertdown:
Since we have a default method for handling dates, I recommend we simply overwrite the built in `date` filter. That way we save a few extra keystrokes and don't go down the path of prepending redundant function names with `oe` (a practice I generally really dislike).

bradymiller (approved):

adunsulag:
@robertdown Figured out a way to configure the default date format for twig instead of override the date filter function.  It accomplishes the same purpose and allows us to drop the custom oeDate function I added.  I believe this addresses your concerns.

robertdown:
@adunsulag good opportunity to look at the use of namespaces for our templates. We are probably a small PR away from allowing modules to inject a custom path. So if you wanted to redirect the Medical Record Dashboard to your module's own custom solution, adding that path to the namespace would allow that.

PR #6636, opened 2023-07-03
feat: Fixes #6634 CORS content-encoding

### LINKED ISSUE #6634: feat: Allow CORS content-encoding header
We should allow the CORS header for content-encoding.  If we want to send compressed data (such as gzipped) data via javascript, we should allow the content-encoding CORS header to be passed in our HttpRestRouteHandler.php class.

adunsulag:
Adds content-encoding to allowed CORS headings
Fixes #6634


bradymiller:
2DV(1)

PR #6647, opened 2023-07-09
chore: Improved handling of inserting and updating encounters

robertdown:
Just improve the way we are handling some data, better code formatting in a couple of ares.

When building the long inert SQL statement, we now combine the column names and values into an associative array which can be manipulated to make the building process a bit easier.

This doesn't actually add any new functionality or features, it is just a value-add for the development process


stephenwaite:
looks good @robertdown, prior formatting was to avoid the 120 character warning
Screenshot from 2023-07-10 10-32-00

stephenwaite:
it's just a warning so doesn't fail tests

PR #6652, opened 2023-07-13
fix: Better UI of PAM UI

### LINKED ISSUE #6651: bug: Long Problem List leads to bad UI
In the event the Problem List grows to be many items long, the UI of the top row will fail.

robertdown:
Move from a flawed flex-based layout to a column based layout to ensure the UI remains usable. Also update the medication card to render the dosage instructions.

Fixes #6651 

Note I copied and pasted a few problem list many times to render a mock PL that is incredibly long as a test that this looks better.

# Before
image

# After
image


bradymiller:
hi @robertdown I think for readability that each medical problem should be on a separate row.

sjpadgett:
Card also needs to have a max height with vertical scroll turned on

robertdown:
Sounds good. I'll change that out.

PR #6659, opened 2023-07-16
Refactor the generate_select_list function and incorporate it to Twig

robertdown:
Ok, I spent some time refactoring the `generate_select_list()` function. This PR will require approval from all admins because this function is so widely used.

What has changed? Previous state was that the entire selectbox string (and children options) were generated inline throughout the function. The changes here make the function a bit more flexible. All attributes of the element are added to an array which gets imploded at the end. Similar changes occur to each option returned from the DB; they go into an array and eventually get imploded. Why? This makes it easier to manipulate the function without having the actual string output be impacted as an unintended side effect.

Additionally, I want to add support for Select2 placeholders, which require some odd manipulation. First, if the selectbox is a single select is must contain an empty `<option></option>` as the first element. However, if the select box is a multiple select, it _cannot_ have that tag. So, there's a regexx that searches for `select2` in the class parameter string, if found, it injects the empty  tag. Still need to tweak this because there are times where the user can pass an empty string with the desired effect of displaying a placeholder.

For Twig, I introduced support for this function with some the following usage guide:
* Use `selectList()` with the following parameters signature: `selectList($name, $list, $value, $title, $opts)`
* `$opts` is an associative array with each key representing the name of the parameter upstream that have defaults set in `options.inc.php`.
* To use this function in twig: `{{ selectList('name', 'someList', value, 'Select Box Title', {class: 'class-one class-two', 'data-custom-attribute': 'the key of this must have quotes because it has a dash in the name' }}

For select2 support, the rendering page will still have to associate the input element to the proper jQuery


robertdown:
This is testing surprisingly well. I still need to sort out handling `optgroup` but once I do that I think this will be ready for a review.

robertdown:
This is ready for a review @bradymiller @sunsetsystems @sjpadgett @stephenwaite @adunsulag

robertdown:
Nudging @sunsetsystems for approval. Is testing very well, was given some feedback about unassigned values not rendering which has since been resolved.

library/options.inc.php
```diff
@@ -134,162 +134,157 @@ function generate_select_list(
...
-            $fontText = xlt('Fix this');
-            $s .= " <span class='text-danger' title='$fontTitle'>$fontText!</span>";
+        $lrow = sqlQuery("SELECT title FROM list_options WHERE list_id = ? AND option_id = ?", [$list_id,$currvalue]);
+
+        $_options[] = [
+            'value' => attr($currvalue),
+            'selected' => true,
+            'label' => ($lrow > 0 && !empty($backup_list)) ? text(xl_list_label($lrow['title'])) : text($currvalue),
+        ];
+        if ($lrow == 0 && empty($backup_list)) {
```
bradymiller, line 283: above conditional does not look right. Shouldn't it be something like? : 
```
 if (empty($lrow) || empty($backup_list)) {
```

bradymiller:
code looks good :) (didn't test)

stephenwaite:
something odd going on with Male being added as the last option to various dropdowns

Screenshot from 2023-07-28 09-37-40

how much testing have you done?

robertdown:
Is the Male option being rendered into non-sex fields? That wasn't occurring in my testing. I'll review the latest change to see what's going on.

stephenwaite:
right, Male is inserted into just about any dropdown

created an [up for grabs demo](https://four.openemr.io/a/openemr/interface/login/login.php?site=default) for easier testing

robertdown:
Looks like the previous issue with Male showing up as an option all the time was an artifiact from me doing some testing. I left hard-coded variables in, which, you guessed it..... sets the current value to always be male.....

stephenwaite (approved):

bradymiller (approved):

sunsetsystems:
Haven't tested but it looks reasonable to me. Thanks Robert!

robertdown:
@openemr/admins Seems everyone has signed off, going to bring this in today

PR #6661, opened 2023-07-16
fix: Problem list UI improvements

robertdown:
Addressing @bradymiller's recs from #6651 to render Problem List as a list versus inline text. Adds a max-height to Problem List, Allergies, and Medication cards to ensure the UI doesn't get out of hand (overflow-y is set to scroll if needed).

Also, only non-resolved PAM will be displayed on the MRD now. I'm pretty sure this was the functionality prior to the twig conversion. No reason to clutter the MRD with data that is rarely used. This data still exists in the Issues page.

#6651


robertdown:
Assuming the tests pass, this is relatively low impact, going to bring it in.

It should be noted, given there is not clinical data potentially off screen (given our max-height of the PAM cards) a potentially serious allergic reaction could not be visible on initial glance. The clinician would **have** to scroll through the allergy list.

Possible solutions are:
1. Allow the allergy card to have a height as large as it needs (I dislike this option)
2. Migrate the allergy card from the current `list-group` display to a running text (similar to how I had the PL originally setup). Downside to this is it's no longer a list per se, but we could put more info on the card.
3. Split the allergy list out based on severity. For any severe, life threatening, or fatal allergy, render that as a badge at the top of the card, leaving all the other allergies in the current list style. Only the list style could be scrolled. This allows the best of both worlds, ensuring the critical allergies always remain visible, while allowing other allergies to be hidden based on max height. I like this solution the best, of course it requires more work 😄

robertdown:
Ok, I just went ahead and built a solution that supports allergies with a severity of severe, life threatening, or fatal to **always** be shown. Allergies with lesser severity or no severity recorded continue to be rendered in a list which has a max-height, thus giving us the best of both worlds. It's tough to tell in the screenshot, but the allergy list does in fact have more items

image

bradymiller:
just need to really make sure that it is very clear to the user that all the allergies are not being shown on the screen (some message at bottom of that card that shows need to scroll down to them all)

bradymiller:
and would also do same for the medical problems and medications

bradymiller:
was looking at the allergy card and it seems odd to have completely different formatting for allergies. Somebody may just look at the "list" if quickly perusing the allergies and even miss the cirtical items or vise versa. All the issues above (and potential for missing these critical items by tired provider eyes) are having me lean towards just putting these cards into the right column without then doing any size reductions of the panels (ie. they show all items (unless the card was minimized by the user)).

robertdown:
There's definitely a path to show a "Scroll to see more" if there's an overflow. What if we moved to the badge style for just the allergies? We could highlight just the high risk ones and leave everything else with a plain background. I find significant value for clinical staff to have allergies, problems, and medications front and center.

Let me evaluate a couple of different solutions to the problem. I wanted to defer the configurability of this page to a later date, but perhaps that is part of the solution. Additionally, I'll look at possible creating a collage style layout which may help.

I'm guessing the real solution is going to be adjusting the layout on a per-user or per-ACL basis. It's just a lot of work that I was hoping to avoid. But no time like the present I suppose.

robertdown:
Haven't forgotten about this. I have a couple of ideas I'm playing with.

PR #6663, opened 2023-07-16
fix: Alarm on the 3 most critical allergic reactions

### LINKED ISSUE #6662: bug: High severity allergies do not display with proper alarms
Currently, only allergies with a Severity of "severe" will render with the correct alarm classes (`bg-danger text-white, font-weight-bold`). There are more severe reactions that should also display an alarm (Life Threatening and Fatal).

A known issue is the severity is a List which means the user could manipulate the list in a way that no longer renders the alarm class as the check on the severity is hard coded. This could be addressed by adding metadata to the Severity List which could have the class to render tied to the actual list item. However, in practice, my recommendation would be for sys admins to **_not_** alter the Severity List as it's a standard list.

robertdown:
Resolves #6662

### DIFF
--- a/templates/patient/card/allergies.html.twig
+++ b/templates/patient/card/allergies.html.twig
@@ -14,8 +14,9 @@
     </div>
     {% endif %}
 {% else %}
+    {% set severityAlarms = ["severe", "life_threatening_severity", "fatal"] %}
     {% for l in list %}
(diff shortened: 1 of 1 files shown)


robertdown:
Low risk for unintended side effects, tests are passing - bringing this in.

bradymiller:
2DV(1)

PR #6668, opened 2023-07-16
fix: Hide the portal card if the portal is disabled and ALL APIs are disabled

robertdown:
Ensure the portal card is rendered if any of the API options or the portal is enabled, otherwise hide the portal card.

See #6653


bradymiller (approved):

bradymiller:
2DV-2(13)

PR #6671, opened 2023-07-17
Feat openemr fix #6669 date partial seconds

### LINKED ISSUE #6669: feat: FHIR Date Search support partial seconds component
Currently our FHIR date search fields only support searching down to the second.  Server support for partial seconds is optional in the FHIR specification but is a very nice feature when converting from javascript's ISO8601 (new Date().toISOString()) to searching a date field in FHIR.  

We should support partial seconds IE (HH::mm:ss.u) up to the milliseconds (4 digits after the period) which is the optional format specified via FHIR.

For example searching for a Questionnaire Response's authored date the following value is rejected:
lt2023-07-06T18:14:19.423Z

Whereas the following follow works fine:
lt2023-07-06T18:14:19Z

We will want to run the Inferno test suite to verify this doesn't break any of the current ONC test suites.

adunsulag:
Fixes #6669 changed out the regex for the date parsing to handle partial seconds up to milliseconds. Also documented the code a little bit better for the regex.


bradymiller (approved):

bradymiller:
neat!

It took me about 30.0145 seconds to complete this review
Then about 1.0453 seconds to click approve
Then about 20.4567 second to type this comment
:)

2DV(1)

PR #6673, opened 2023-07-17
Fixes #6672 source_url and order QRs by date

### LINKED ISSUE #6672: bug: Questionnaire source_url breaks if oauth2 fhir url does not end in a slash
Have an issue where I imported a couple of questionnaires and the source_url was invalid because my oauth fhir url wasn't setup with a slash at the end.  Need to just fix it in the questionnaire service so it adds a slash if one is missing, just so we have valid urls.

adunsulag:
Made it so the source_url for questionnaire imports is properly formatted.

Also fixed the questionnaire reponse to be ordered by date and exposed the questionnaire uuid in the sql query.
Fixes #6672


bradymiller (approved):

bradymiller:
tenor99

PR #6685, opened 2023-07-24
fix: Better Facilities UI

robertdown:
Resolves #6683 - does not change the voice or selection, simply moves the toggler to the toggler section


robertdown:
Closes, superseded by #6687

PR #6701, opened 2023-07-30
fix(stylelint): correct stylelint "extends" configuration

### LINKED ISSUE #6700: Duplicated "extends" key in stylelint configuration file
### Describe the bug

The key "extends" is duplicated in the stylelint configuration file:
image

In the [stylelint docs](https://stylelint.io/user-guide/configure/#extends), multiple configurations are included using array notation:
image

raskolnikov-rodion:
Fixes #6700

### DIFF
--- a/contrib/forms/example2/print.php
+++ b/contrib/forms/example2/print.php
@@ -68,9 +68,9 @@ Printed on <?php echo date("F d, Y", time()); ?>
 <span class="title"><?php echo xlt($form_name); ?></span><br />
 
 <!-- container for the main body of the form -->
-<div id="print_form_container">
+<div id="print-form-container">
 
-<div id="print_general">
+<div id="print-general">
 <table>
 <tr><td>
 Date:
@@ -96,7 +96,7 @@ Address: <input name="address" id="address" type="text" size="80" maxlength="250
 </table>
 </div>
 
-<div id="print_bottom">
+<div id="print-bottom">
 Use this space to express notes <br />
 <textarea name="notes" id="notes" cols="80" rows="4"><?php echo attr($record['notes']);?></textarea>
 <br /><br />
--- a/swagger/index.css
+++ b/swagger/index.css
@@ -1,16 +1,16 @@
 html {
-    box-sizing: border-box;
-    overflow: -moz-scrollbars-vertical;
-    overflow-y: scroll;
+  box-sizing: border-box;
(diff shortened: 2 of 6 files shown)


stephenwaite (approved):
Nice!

swagger/index.css
```diff
@@ -1,16 +1,16 @@
 html {
-    box-sizing: border-box;
-    overflow: -moz-scrollbars-vertical;
-    overflow-y: scroll;
+  box-sizing: border-box;
+  overflow: -moz-scrollbars-vertical;
+  overflow-y: scroll;
 }
```
bradymiller, line 5: note i wouldn't consider this change persistent (ie. still would need to ignore this folder) since this folder gets replaced from updated swagger packager every once in a while

contrib/forms/example2/print.php
```diff
@@ -68,9 +68,9 @@
 <span class="title"><?php echo xlt($form_name); ?></span><br />
 
 <!-- container for the main body of the form -->
-<div id="print_form_container">
+<div id="print-form-container">
```
bradymiller, line 71: I am guessing this is going to be a very prevalent issue in OpenEMR's codebase. Is there a chance of css breakage when using underscore instead of a dash?
(if not and this is, as i suspect, a very prevalent issue in OpenEMR's codebase, then would consider ignoring this rule since the extensive fixes required could open the door to lots of bugs)

swagger/index.css
raskolnikov-rodion: That's good to know, I'll include it in .stylelintingnore. Thanks!

contrib/forms/example2/print.php
raskolnikov-rodion: Indeed, it's a common issue. Currently, we have almost 700 occurrences of it. The goal of this change to conform to the default pattern of the `selector-id-pattern` rule, which is kebab-case. But according to the [MDN docs for the id attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id), there's no issue in using `_` for ids. Since both kebab-case and snake_case ids are very common patterns in this project, it makes sense to disable this rule. I'll update the config. Thanks!

bradymiller (approved):

bradymiller:
hi @raskolnikov-rodion , looks great and bringing in!

dance-bear

PR #6709, opened 2023-08-01
feat: Care Plan UI Wider Textbox

robertdown:
Put the metadata on it's own row, allowing for the description to be wider

image


stephenwaite:
smaller_excited_elephant

robertdown:
I think the name of @stephenwaite's gif really does say it all

image

PR #6716, opened 2023-08-06
feat: Hook into the CDA Parsing functions

### LINKED ISSUE #6715: feat: Add events to CDA Import
Adding events during the CDA import process would allow ill-formed CDA specs to be manipulated by a module instead of changing core functionality

robertdown:
Will resolve #6715. Add a new Event allowing hooks into `parseCDAEntryComponents()` for every component that gets parsed. The event dispatches the string of the component (encounter, allergy, medication, problems, etc), the OID, the component array, and the current state of `$templateData`.

`CdaTemplateParse->templateData` is always updated with the results of `$event->getTemplateData()` The event runs __after__ the core parsing function runs

### DIFF
--- a/src/Services/Cda/CdaTemplateParse.php
+++ b/src/Services/Cda/CdaTemplateParse.php
@@ -13,7 +13,8 @@
 namespace OpenEMR\Services\Cda;
 
 use DOMDocument;
-use OpenEMR\Events\CDA\CDAParseEvent;
+use OpenEMR\Events\CDA\CDAPreParseEvent;
+use OpenEMR\Events\CDA\CDAPostParseEvent;
 use OpenEMR\Services\CodeTypesService;
 
 class CdaTemplateParse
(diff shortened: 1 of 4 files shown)


src/Services/Cda/CdaTemplateParse.php
```diff
@@ -22,11 +23,18 @@ class CdaTemplateParse
...
     protected $is_qrda_import;
 
+    /**
+     * @var \Symfony\Contracts\EventDispatcher\EventDispatcherInterface
+     */
+    private $ed;
+
     public function __construct()
     {
+        global $GLOBALS;
```
bradymiller, line 33: this seems redundant since $GLOBALS is already a global

bradymiller (approved):

bradymiller:
seems good to me :) looks like styling issues to fix. would have @sjpadgett give blessing before bringing in.

bradymiller (approved):

bradymiller:
tenor99

robertdown:
@sjpadgett you ok with these changes? It's testing well, I'm deep into manipulating a CDA doc that is a bit weird all using the hooks and it's working as expected.

sjpadgett:
I haven't had time to look over but seems Brady and others are okay with so bring in and thanks. Kind of neat

PR #6719, opened 2023-08-09
Lint js updates

bradymiller:
fix: add vendor directory to .eslintignore

feat: add lint-js fix support for the devtools


bradymiller:
testing well

PR #6721, opened 2023-08-09
Fixes #6720 appointment dialog hooks

### LINKED ISSUE #6720: feat: Extend appointment dialog with hooks for modules and smart apps
We should allow a smart app to be launched with an appointment dialog intent so that it can appear inside the appointment dialog window with the currently selected appointment.  

We also should allow module writers to hook into the close dialog event on the appointment so we can do additional processing / launch additional screens on the appointment screen.

adunsulag:
Made it so we can launch a smart app with a specific appointment context as well as hook into the appointment close dialog.

Also added in an ehr-launch-client.php file so we can quickly do an ehr initiated app launch.

Fixes #6720

### DIFF
--- a/interface/smart/ehr-launch-client.php
+++ b/interface/smart/ehr-launch-client.php
@@ -1,5 +1,15 @@
 <?php
 
+/**
+ * ehr-launch-client.php  Main entry point for the OpenEMR OAUTH2 / SMART client in ehr launch
+ * Allows a smart app to launch into the OpenEMR EHR in a seamless interaction
+ * @package openemr
+ * @link      http://www.open-emr.org
+ * @author    Stephen Nielson <snielson@discoverandchange.com>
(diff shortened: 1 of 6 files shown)


interface/smart/ehr-launch-client.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
bradymiller, line 1: need header

src/Events/Appointments/AppointmentDialogCloseEvent.php
```diff
@@ -0,0 +1,36 @@
+<?php
```
bradymiller, line 1: need header

interface/smart/ehr-launch-client.php
adunsulag: Fixed

PR #6728, opened 2023-08-10
fix(eslint): fix no-mixed-spaces-and-tabs error

raskolnikov-rodion:


.eslintrc.json
```diff
@@ -19,8 +19,7 @@
         "no-undef": "warn",
         "no-unused-vars": "warn",
         "no-redeclare": "warn",
-        "no-useless-escape": "warn",
-        "no-mixed-spaces-and-tabs": "warn"
```
robertdown, line 23: This feels like a good rule, why remove it?

robertdown (commented):
Just 1 question on the .eslintrc.json change, everything else looks good

.eslintrc.json
raskolnikov-rodion: The rule will still be included as it's part of the `eslint:recommended` config. By removing it from here we're simply reverting to its original error level 🙂

robertdown (approved):

PR #6729, opened 2023-08-10
fix(stylelint): fix errors in library and public folders

raskolnikov-rodion:


public/fonts/lato/lato.css
```diff
@@ -1,191 +1,207 @@
...
   font-weight: bold;
   src: url('Lato-Bold.eot');
-  src: url('Lato-Bold.eot?#iefix') format('embedded-opentype'),
+  src:
+    url('Lato-Bold.eot?#iefix') format('embedded-opentype'),
     url('Lato-Bold.woff2') format('woff2'),
     url('Lato-Bold.woff') format('woff'),
     url('Lato-Bold.ttf') format('truetype');
-  text-rendering: optimizeLegibility;
+  text-rendering: optimizelegibility;
```
bradymiller, line 11: hi @raskolnikov-rodion , Are we sure that above change will still work (all the documentation I could find used the camelcase form of optimizeLegibility)?

raskolnikov-rodion: Good question! Here are the suggested values for this property in different browsers:

Google Chrome v.115.0.5790.171
image

Firefox v.116.0.1
image

Microsoft Edge v.115.0.1901.200
image

Brave v.1.56.20
image

Except for Firefox, they all mention both the lowercase and camelCase forms. If you search for optimizelegibility in [caniuse](https://caniuse.com/?search=optimizelegibility), you get:

> When used in HTML, the unofficial text-rendering: optimizeLegibility CSS property enables high-quality kerning and ligatures in certain browsers. **Newer browsers have this behavior enabled by default.**

I also tested locally and couldn't notice any detrimental differences. This all leads me to believe this is a safe change - the docs might not be totally in sync with browser implementations. But I encourage people to try the change themselves and see if they can notice any difference - extra pairs of 👀 won't hurt 🙂 And if you can think of exceptions and cases where this might break, let me know. I believe we can configure the rule to allow for this value in camelCase.

bradymiller:
@raskolnikov-rodion , Thanks for the analysis. Interesting that firefox does not have the camelcase form. Gonna bring this in. Thanks for another nice improvement!

bradymiller:
incorporated the new stylelint command into the devtools: https://github.com/openemr/openemr-devops/pull/366

PR #6733, opened 2023-08-10
Fixes #6732 fhir _search query params

### LINKED ISSUE #6732: bug: FHIR POST _search operation returns different results than GET search
Our FHIR POST _search operation is returning different results than doing a GET search.  It almost looks like the POST is ignoring all filtering.

This is happening on the latest development master branch using the ONC Inferno tool V.3.8.0 with the US Core 3.1.1 / USCDI v1, SMART App Launch 1.0.0, Bulk Data 1.0.1.

image

adunsulag:
Query params were not being set due to how the http rest request is not being populated.

Fixes #6732

### DIFF
--- a/src/Common/Http/HttpRestRequest.php
+++ b/src/Common/Http/HttpRestRequest.php
@@ -192,7 +192,7 @@ class HttpRestRequest implements ServerRequestInterface
 
     public function setQueryParams($queryParams)
     {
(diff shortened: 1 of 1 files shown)


adunsulag:
If this passes CI, I'm going to bring it in as the FHIR POST api endpoints are broken and the inferno test suite on the patient apis is now passing with this fix.

bradymiller:
2DV(5)

PR #6734, opened 2023-08-11
fix(eslint): fix no-useless-escape error

raskolnikov-rodion:
This rule flags escapes that can be safely removed without changing behaviour.

Docs: https://eslint.org/docs/latest/rules/no-useless-escape


bradymiller:
neat stuff! bringing in

PR #6735, opened 2023-08-11
feat(jest): coverage reports for js unit tests

raskolnikov-rodion:
Create script to generate test coverage reports for JS files.

Running `test:js-coverage` will generate the report inside the folder `coverage/js-unit`.

By opening the `index.html` file inside the folder `coverage/js-unit/lcov-report`, you can easily navigate through the project and check statistics for each file:

image

image

image

If there's interest, we could also include the `coverageThreshold` option to discourage people from adding more JS without including tests. The initial value would allow the project to pass as is. As coverage increases over time, we would increase the value of the threshold to prevent coverage from falling. This is a decision to be made by the team, so it's not included in this initial proposal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ ccdaservice/node_modules
 
 # testing
 .phpunit.result.cache
+coverage/js-unit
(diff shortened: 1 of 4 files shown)


bradymiller:
@raskolnikov-rodion , some more really cool stuff! Gonna bring this in. And I will incorporate this into the devtools and openemr-cmd command so easy to use this stuff in the docker dev environment.

dance-bear

bradymiller:
Incorporated this feature into devtools and openemr-cmd command: https://github.com/openemr/openemr-devops/pull/366

PR #6749, opened 2023-08-16
feat: Better UI and Code for encounter report

### LINKED ISSUE #6748: feat: Improve UI and code of newpatient report.php

robertdown:
Move the encounter report.php to twig, make some minor UI improvements

Will resolve #6748

### DIFF
--- a/interface/forms/newpatient/report.php
+++ b/interface/forms/newpatient/report.php
@@ -1,12 +1,14 @@
 <?php
-
 /**
  * Encounter form report function.
  *
  * @package   OpenEMR
  * @link      http://www.open-emr.org
  * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @author    Robert Down <robertdown@live.com
  * @copyright Copyright (c) 2019 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2023 Robert Down <robertdown@live.com
+ * @copyright Copyright (c) 2023 Providence Healthtech
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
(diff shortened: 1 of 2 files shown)


interface/forms/newpatient/report.php
```diff
@@ -15,25 +18,34 @@
...
-            print "<span>" . xlt('Provider') . ": </span><span class=text>" . text($provider['lname'] . ", " . $provider['fname']) . "</span><br />\n";
-            print "<span>" . xlt('Referring Provider') . ": </span><span class=text>" . text(($referringProvider['lname'] ?? '') . ", " . ($referringProvider['fname'] ?? '')) . "</span><br />\n";
-            print "<span>" . xlt('POS Code') . ": </span><span class=text>" . text(sprintf('%02d', trim($result['pos_code'] ?? ''))) . "</span><br />\n";
-        }
-    }
+        $reason = (!$hasAccess) ? false : nl2br(text($result['reason']));
+        $provider = (!$hasAccess) ? false : text($rawProvider['fname'] . " " . $rawProvider['lname']);
+        $referringProvider = (!$hasAccess) ? false : $rawRefProvider;
+        $posCode = (!$hasAccess) ? false : text(sprintf('%02d', trim($result['pos_code'] ?? false)));
+        $facility_name = (!$hasAccess) ? false : $result['facility_name'];
```
bradymiller, line 39: recommend instead getting in habit of doing the escaping as far downstream as possible, which in this case would be in the twig template (provides more flexibility of which escaping function will be used (for example, if a future dev decides to place the provider name in a html attribute in the twig template, then much easier to do it there than need to work out how to do it here, which would then essentially require 2 different variables with different escaping) and makes it much easier to audit code for security)

bradymiller: to this end, would encapsulate the nl2br function in a twig call

robertdown:
Thanks as always for the feedback @bradymiller - I made those changes and I think we are good to go. I think it would be good to expand the report a bit more, but this PR is really just to get it moved over to twig

interface/forms/newpatient/templates/report.html.twig
```diff
@@ -0,0 +1,13 @@
+<div class="w-100">
+{% for e in encounters %}
+    {% set _reason = (e.reason) ? " - " ~ e.reason|nl2br|xl : "" %}
+    <h3>{{ e.category|xlt }}{{ _reason }}</h3>
```
bradymiller, line 4:
```
<h3>{{ e.category|text }}{{ _reason }}</h3>
```

interface/forms/newpatient/templates/report.html.twig
```diff
@@ -0,0 +1,13 @@
+<div class="w-100">
+{% for e in encounters %}
+    {% set _reason = (e.reason) ? " - " ~ e.reason|nl2br|xl : "" %}
```
bradymiller, line 3:
```
{% set _reason = (e.reason) ? " - " ~ e.reason|text|nl2br : "" %}
```
(i am assuming the piped commands in twig go in order from left to right)

interface/forms/newpatient/templates/report.html.twig
```diff
@@ -0,0 +1,13 @@
...
+    <h3>{{ e.category|xlt }}{{ _reason }}</h3>
+    <h5 class="text-muted">{{ e.provider|text }} {{ "at"|xlt }} {{ e.facility|text }}</h5>
```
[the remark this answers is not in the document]
robertdown: Yes, I think it would be fine without the "at" constant. Just lost the facility in parentheses.

interface/forms/newpatient/templates/report.html.twig
bradymiller: Looks like this has not been addressed yet. Rec keeping it what it was in above code where it escapes and does not translate. If want to translate it, then would need to use the xl_appt_category() function which uses the global setting `$GLOBALS['translate_appt_categories']` to decide whether to translate (allows users to decide if translate these or if they simply do their own translations directly in the label). In that case would make my recommended change above (use text). and then wrap the label in report.php line 43 with the xl_appt_category() call

interface/forms/newpatient/templates/report.html.twig
```diff
@@ -0,0 +1,13 @@
+<div class="w-100">
+{% for e in encounters %}
+    {% set _reason = (e.reason) ? " - " ~ e.reason|nl2br|text : "" %}
```
bradymiller, line 3: Note that the escaping with `text` needs to happen before the `nl2br` call or else the newlines will turn into `<br>`.

interface/forms/newpatient/templates/report.html.twig
robertdown: Got it, I'll use `xl_appt_category()` in the controller to make it translatable.

robertdown:
Going to bring this in, I took out the `nl2br` call and am translating the appointment category from the controller, not the template.

interface/forms/newpatient/templates/report.html.twig
bradymiller: hi @robertdown , this was not addressed in this PR??

robertdown: It should have been? I called that function from the controller. I'll look at the code, but I made that change before committing. If it's not there I'll add it.

PR #6755, opened 2023-08-17
feat: Make component list in Nation Note more user friendly

robertdown:
Make the list of components to insert more manageable by not rendering the body of the component, instead render the short form.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -241,3 +241,6 @@ assets:
     checkpwd_validation:
         basePath: '%webroot%/interface/usergroup/'
         script: checkpwd_validation.js
+    ajax-functions-writer:
+        basePath: '%webroot%/library/js/'
+        script: ajax_functions_writer.js
--- a/interface/themes/core/navmenu.scss
+++ b/interface/themes/core/navmenu.scss
@@ -149,43 +149,15 @@
       margin: 0;
       padding: 0;
 
-      a:hover {
-        background: lighten($light, 20%);
-        text-decoration: underline;
-      }
-
       ul {
         display: block;
-      }
-
-      li {
-        span {
-          background: lighten($yellow, 60%);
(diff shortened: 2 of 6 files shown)


robertdown:
This turned out to be bigger than I had thought.

1. Adding new components or updating components now take place in a `dialog`
2. You can now supply a short name to the component, which will render in the list
3. Insert / Update / Delete is a bit more separated from the UI
4. Several jQuery ajax calls have been replaced with `fetch()` calls. 
5. In the event there is only 1 category in the given context, it will automatically load that categories components.
6. The short name was expanded in the DB to allow longer names
7. The actual component is now LONGTEXT instead of TEXT in the database

**Still Pending**
- [ ] SQL to copy current components into the short name and truncate (otherwise the list will break for old components)

adunsulag:
Its unfortunate we didn't review this earlier.  There's now a lot of conflicts.  @robertdown did you want to resolve these issues or should we close the PR?

kojiromike:
Thanks for the work on this, @robertdown. The UX improvements here (dialog-based editing, short names in the list) would be nice to have.

Unfortunately, after 2+ years there are some significant hurdles to merging this:

1. **CKEditor 5 migration** - The PR was written for CKEditor 4, but OpenEMR migrated to CKEditor 5 in #8058. The APIs are completely different, so the frontend JS would need to be rewritten rather than rebased.

2. **7 files with merge conflicts** - Including the main `custom_template.php` which has had several changes (rector refactoring, URLSearchParams, bug fixes).

3. **Obsolete upgrade SQL** - The PR targets `7_0_1-to-7_0_2_upgrade.sql`, but we're now on 8.0.1.

Given that @adunsulag asked about this back in June with no response, we're going to assume you're not planning to pick this back up and close it out. If that's wrong, just let us know and we can reopen.

PR #6767, opened 2023-08-17
feat: openemr 6752 6751 address book user rest api

### LINKED ISSUE #6751: feat: REST API add user endpoint
We have an endpoint for grabbing practitioners but if we want to grab a user, organization, external entity, etc that has been populated via the address book there is no current mechanism to do that.  We need to add a REST API endpoint to grab the more general purpose user.  We should also allow retrieving this user by their 'id' parameter since so many standard api results return the actual id (not the uuid) of the user.

### LINKED ISSUE #6752: feat: globals setting allow address book data type
Our globals setting should support entering / selecting an address book user id value from the settings.  In my use case I want to select a user to contact for a particular type of notifications.  The user is defined in the address book.  Therefore I want to be able to pop open the address book dialog and select a user that will get saved in the global settings.

adunsulag:
Fixes #6752 
Fixes #6751 

Allows an address book config data type to be added for the global config.

Adds an endpoint for retrieving users from the system.  If there are users we wish to retrieve that are not practitioners (such as external entities that are in the address book), we can't use the practitioner api.  Instead we expose the users endpoint.

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -31,6 +31,7 @@ use OpenEMR\Common\Logging\EventAuditLogger;
 use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Common\Logging\SystemLogger;
 use OpenEMR\Core\Header;
+use OpenEMR\FHIR\Config\ServerConfig;
 use OpenEMR\OeUI\OemrUI;
 use OpenEMR\Services\Globals\GlobalSetting;
 use Ramsey\Uuid\Uuid;
@@ -373,7 +374,7 @@ $arrOeUiSettings = array(
     'help_file_name' => ""
 );
 $oemr_ui = new OemrUI($arrOeUiSettings);
-$serverConfig = new \OpenEMR\FHIR\Config\ServerConfig();
+$serverConfig = new ServerConfig();
 $apiUrl = $serverConfig->getBaseApiUrl();
 ?>
 <script src="edit_globals.js" type="text/javascript"></script>
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -373,8 +373,13 @@ $arrOeUiSettings = array(
     'help_file_name' => ""
 );
 $oemr_ui = new OemrUI($arrOeUiSettings);
+$serverConfig = new \OpenEMR\FHIR\Config\ServerConfig();
+$apiUrl = $serverConfig->getBaseApiUrl();
 ?>
 <script src="edit_globals.js" type="text/javascript"></script>
+<script>
+    window.oeUI.api.setApiUrlAndCsrfToken(<?php echo js_escape($apiUrl); ?>, <?php echo js_escape(CsrfUtils::collectCsrfToken('api')); ?>);
+</script>
 </head>
 
 <body <?php if ($userMode) {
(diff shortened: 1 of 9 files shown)


interface/super/edit_globals.php
```diff
@@ -373,8 +373,13 @@ function checkBackgroundServices()
     'help_file_name' => ""
 );
 $oemr_ui = new OemrUI($arrOeUiSettings);
+$serverConfig = new \OpenEMR\FHIR\Config\ServerConfig();
```
bradymiller, line 376: would place this in use clause at top script

interface/super/templates/globals-address-book.php
```diff
@@ -0,0 +1,38 @@
...
+ * @copyright Copyright (c) 2022 Discover and Change <snielson@discoverandchange.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+use OpenEMR\Common\Csrf\CsrfUtils;
+
+$textLabel = "";
+$i = $i ?? 0;
+if (!empty($fldvalue)) {
+    $users = \OpenEMR\Common\Database\QueryUtils::fetchRecords("SELECT fname, lname FROM users WHERE id = ?", [$fldvalue]);
```
bradymiller, line 20: use clause

interface/super/templates/globals-address-book.php
```diff
@@ -0,0 +1,38 @@
...
+    $user = $users[0] ?? ['fname' => '', 'lname' => ''];
+    $textLabel = trim($user['fname'] . ' ' . $user['lname']);
+}
+$serverConfig = new \OpenEMR\FHIR\Config\ServerConfig();
```
bradymiller, line 24: use clause

interface/super/templates/globals-address-book.php
```diff
@@ -0,0 +1,38 @@
...
+$apiUrl = $serverConfig->getStandardApiUrl();
+?>
+<div class="gbl-field-address-book-widget">
+    <input type='hidden' class="address-book-widget-input" name='form_<?php echo attr($i); ?>' id='form_<?php echo attr($i); ?>' value='<?php echo attr($fldvalue); ?>' />
+    <div class="input-group">
+        <input type='text' class="address-book-widget-label form-control" readonly="readonly" id='form_label_<?php echo attr($i); ?>' value='<?php echo text($textLabel); ?>'/>
```
bradymiller, line 30: `attr($textLabel)` (not text())

interface/super/edit_globals.php
```diff
@@ -373,8 +373,13 @@ function checkBackgroundServices()
...
+$apiUrl = $serverConfig->getBaseApiUrl();
```
bradymiller, line 377: getting the full url including server may not always work if the global is not set correctly. Can't we just use the relative path here like we normally do when calling scripts (then don't need to worry if the global is set correctly)?

adunsulag: I think the issue here is that I'm hitting the API endpoints which all require the oauth global to be setup. All of our apis utilize that oauth global and if anyone has done any kind of proxy on the api and setup the url to be different than the local relative server url then we'd have problems hitting the api.  Since we have a global that could be different then the local setup, I need to make sure if I'm hitting the api that we are hitting the configured endpoint for the api.

I'd rather that we didn't have the global at all, but since we have the global for the api I think we need to use it when communicating with the api.

interface/super/edit_globals.php
adunsulag: Fixed

interface/super/templates/globals-address-book.php
adunsulag: Fixed.

interface/super/templates/globals-address-book.php
adunsulag: Fixed.

interface/super/edit_globals.php
adunsulag: Resolved by using local api.

adunsulag:
Bringing this in as all concerns have been addressed.

bradymiller:
tenor99

PR #6768, opened 2023-08-17
Add container, better rendering of header, space buttons

robertdown:
A bit better styling of the Encounter page. Still need to add a bunch of forms to test.

# Before
image

# After
image

### DIFF
--- a/interface/patient_file/encounter/forms.php
+++ b/interface/patient_file/encounter/forms.php
@@ -352,17 +352,6 @@ function refreshVisitDisplay() {
         padding: 8px;
     }
 
-    div.form_header {
-        float: left;
-        min-width: 400px;
-    }
-
-    div.form_header_controls {
-        float: left;
-        margin-bottom: 2px;
-        margin-left: 6px;
-    }
-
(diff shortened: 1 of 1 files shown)


interface/patient_file/encounter/forms.php
```diff
@@ -627,7 +616,7 @@ function findPosY(obj) {
 ]);
 ?>
 
-<div id="encounter_forms" class="mx-1">
+<div id="encounter_forms" class="container">
```
sjpadgett, line 619: Recommend class="container-xl" or container-lg because container doesn't fill screen on small devices.

stephenwaite:
this is a good PR for Brady's tip to put a `?w=1` at the end of the url to ignore whitespace changes

```
https://github.com/openemr/openemr/pull/6768/files?w=1

robertdown:
Good call. I had to wrap a bunch of code with a new div which indented a lot of lines. I think you can set a preference now in GitHub to always ignore whitespace changes.

robertdown:
image

PR #6779, opened 2023-08-21
added ability to see multiple eligibilities if

### LINKED ISSUE #6778: Not all eligibilities were displaying
### Describe the bug

sometimes the system would return more than one eligibility when checking a single payer. we were only displaying the first item in the array, but there could be more. In the case of checking CMS, a replacement plan may also be returned. 

### To Reproduce
run eligibility on a patient with a replacement plan

### Expected behavior

Should display all returned insurances

claimrevolution:
if more than one come back on a check.
Increased table column size too, truncation was happening

Fixes #6778

this will display the missing insurances that should of been presented.  Also the text data column was to small and data was being truncated, so more data could be missing.

Just looping the data and using the correct property. Also changing the column types on the table the module uses.

### DIFF
--- a/interface/modules/custom_modules/oe-module-claimrev-connect/table.sql
+++ b/interface/modules/custom_modules/oe-module-claimrev-connect/table.sql
@@ -4,16 +4,23 @@ CREATE TABLE IF NOT EXISTS `mod_claimrev_eligibility`(
     ,`pid` bigint(20)
     ,`payer_responsibility` varchar(2)
     ,`request_json` TEXT
-    ,`response_json` TEXT
-	,`eligibility_json` TEXT
-	,`individual_json` TEXT
+    ,`response_json` LONGTEXT
+	,`eligibility_json` LONGTEXT
+	,`individual_json` LONGTEXT
     ,`response_message` TEXT
     ,`status` varchar(25)
     ,`last_checked` datetime default NULL
     ,`create_date` datetime default NULL
-    ,`raw271` MEDIUMTEXT
-);
+    ,`raw271` LONGTEXT
+); 
 
+ALTER TABLE `mod_claimrev_eligibility` 
+CHANGE COLUMN `response_json` `response_json` LONGTEXT NULL DEFAULT NULL ,
+CHANGE COLUMN `eligibility_json` `eligibility_json` LONGTEXT NULL DEFAULT NULL ,
+CHANGE COLUMN `individual_json` `individual_json` LONGTEXT NULL DEFAULT NULL ,
+CHANGE COLUMN `raw271` `raw271` LONGTEXT NULL DEFAULT NULL ;
+  
+  
 -- Add the background service for sending claims
 #IfNotRow background_services name ClaimRev_Send
 INSERT INTO `background_services` (`name`, `title`, `active`, `running`, `next_run`, `execute_interval`, `function`, `require_once`, `sort_order`) VALUES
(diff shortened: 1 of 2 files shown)


interface/modules/custom_modules/oe-module-claimrev-connect/table.sql
```diff
@@ -4,16 +4,23 @@ CREATE TABLE IF NOT EXISTS `mod_claimrev_eligibility`(
...
-    ,`raw271` MEDIUMTEXT
-);
+    ,`raw271` LONGTEXT
+); 
 
+ALTER TABLE `mod_claimrev_eligibility` 
+CHANGE COLUMN `response_json` `response_json` LONGTEXT NULL DEFAULT NULL ,
+CHANGE COLUMN `eligibility_json` `eligibility_json` LONGTEXT NULL DEFAULT NULL ,
+CHANGE COLUMN `individual_json` `individual_json` LONGTEXT NULL DEFAULT NULL ,
+CHANGE COLUMN `raw271` `raw271` LONGTEXT NULL DEFAULT NULL ;
```
bradymiller, line 21: should be in a if (there is a if mechanism to avoid breaking things in database) and should should not mention defaults for longtext. here is what would change above to:
```
#IfNotColumnType mod_claimrev_eligibility response_json LONGTEXT
ALTER TABLE `mod_claimrev_eligibility` CHANGE `response_json` `response_json` LONGTEXT;
#EndIf

#IfNotColumnType mod_claimrev_eligibility eligibility_json LONGTEXT
ALTER TABLE `mod_claimrev_eligibility` CHANGE `eligibility_json` `eligibility_json` LONGTEXT;
#EndIf

#IfNotColumnType mod_claimrev_eligibility individual_json LONGTEXT
ALTER TABLE `mod_claimrev_eligibility` CHANGE `individual_json` `individual_json` LONGTEXT;
#EndIf

#IfNotColumnType mod_claimrev_eligibility raw271 LONGTEXT
ALTER TABLE `mod_claimrev_eligibility` CHANGE `raw271` `raw271` LONGTEXT;
#EndIf
```

interface/modules/custom_modules/oe-module-claimrev-connect/templates/eligibility.php
```diff
@@ -112,133 +112,139 @@
...
+                    if ($benefits != null) {
+                                include $path . '/subscriber_patient.php';
+                                include $path . '/benefit.php';
+                    }
+                    ?>
                                         </div>
                                     </div>
                                 </div>
-                                <div id="eligibility-validations" class="tab-pane">
+                                <div id="eligibility-validations-<?php echo($index); ?>" class="tab-pane">
```
bradymiller, line 233: to be safe would html escape all the $index with attr function:
```
<?php echo(attr($index)); ?>
```

bradymiller:
review complete

interface/modules/custom_modules/oe-module-claimrev-connect/table.sql
claimrevolution: made those changes. I was actually trying to figure out that if logic but couldn't.

interface/modules/custom_modules/oe-module-claimrev-connect/templates/eligibility.php
claimrevolution: done

bradymiller (approved):

bradymiller:
looks good and bringing in. will also bring into the rel-701 branch (so it will be in next 7.0.1 patch)

2DV(1)

PR #6780, opened 2023-08-21
chore(stylelint): specify interface folders with errors

raskolnikov-rodion:
Specifying interface folders with errors, so they can be separately fixed in future changes. This also reincorporates folders without errors in the stylelint analysis, so they can be monitored for errors from now on.


bradymiller (approved):

bradymiller:
hi @raskolnikov-rodion , looks good and bringing in. thanks!

PR #6784, opened 2023-08-22
Fixes #6783 missing user rest controller

### LINKED ISSUE #6783: bug: Missing user rest controller for API
Ouch... not sure how this slipped through but the user api fails because its missing the rest controller.

adunsulag:
Didn't see this until I tried an external server.  Ouch.

Fixes #6783

### DIFF
--- a/src/RestControllers/UserRestController.php
+++ b/src/RestControllers/UserRestController.php
@@ -1,0 +1,99 @@
+<?php
+
+/**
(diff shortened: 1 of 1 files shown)


adunsulag:
If everything passes I'm going to bring this in as the API is broken in master w/o it.

bradymiller:
2DV(1)

PR #6785, opened 2023-08-22
refactor(ccdaservice): extract and test cleanCode function

raskolnikov-rodion:
- extract cleanCode function to a separate file
- add tests
- create constant NULL_FLAVOR

The constant allows us to reuse the value without having to type it repeatedly, which is error-prone. It also simplifies the process of updating the value, if needed in the future.


bradymiller:
hi @raskolnikov-rodion , more neat stuff! bringing in. thanks!

Issue #6789, opened 2023-08-24
feat: Code Selections on Fee Sheet are difficult to read


robertdown:
When searching for additional codes on the fee sheet, the radio buttons to select the various code types are too close together. Using a button styling can improve readability on the fee sheet

PR #6790, opened 2023-08-24
feat: Better spacing of additional code buttons on fee sheet

robertdown:
Add button classes to the radio inline

# Before
image

# After
image


robertdown:
Resolves #6789

PR #6795, opened 2023-08-25
refactor(ccdaservice): extract and test date functions

raskolnikov-rodion:


ccdaservice/utils/date/date.js
```diff
@@ -0,0 +1,39 @@
...
+
+    if (input.length === 17) {
+        const sections = input.split(' ');
+        return `${sections[0].slice(0, 4)}-${sections[0].slice(4, 6)}-${sections[0].slice(6, 8)} ${sections[1]}`;
+    }
+
+    if (input.length === 19 && (input[14] === '-')) {
+        const sections = input.split('-');
+        const date = `${sections[0].slice(0, 4)}-${sections[0].slice(4, 6)}-${sections[0].slice(6, 8)}`;
+        const time = `${sections[0].slice(8, 10)}:${sections[0].slice(10, 12)}:${sections[0].slice(12, 14)}`;
```
bradymiller, line 27: The time code does not look right here. In original function it is slicing it from the original `str` (ie. `input` in this new function) instead of the `strZone` (ie. `sections` in the new function).

ccdaservice/utils/date/date.js
```diff
@@ -0,0 +1,39 @@
...
+        return `${date} ${time}-${sections[1]}`;
+    }
+
+    return input;
+}
+
+function templateDate(date, precision) {
+    return { date: fDate(date), precision };
```
bradymiller, line 35: Been really fun learning more about javascript in your PRs. I was unable to figure out the answer to this question:
Does the above line send the same json structure (I am assuming this is json) back as the original line `return {'date': fDate(date), 'precision': precision}` ?

ccdaservice/utils/date/date.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    });
+
+    afterAll(() => {
+        jest.useRealTimers();
+    });
+});
+
+describe('templateDate', () => {
+    it('should return an object containing the formatted date and precision', () => {
+        expect(templateDate('20230514', 'day')).toEqual({ date: '2023-05-14', precision: 'day' });
```
bradymiller, line 41: lol, looks like this answers my question above :)

ccdaservice/utils/date/date.js
raskolnikov-rodion: The original code has an assumption that makes it work: the first occurrence of `-` is at `input[14]`. Therefore, taking the characters from the original string and from `sections[0]` produces the same result, as both refer to the same characters at the beginning of the string.

If this assumption didn't hold, then doing `str1 = strDate + '  ' + strTime + '-' + strZone[1];` in the original code wouldn't be correct, as `strZone[1]` would be repeating characters that were already included in the formatted string.

ccdaservice/utils/date/date.js
raskolnikov-rodion: Yes, it's the property shorthand. Glad it's been helpful!

ccdaservice/utils/date/date.js
bradymiller: thanks for the explanation. sounds good.

bradymiller:
hi @raskolnikov-rodion , bringing this in. thanks for another very nice code improvement!

PR #6801, opened 2023-08-26
WIP Fixes #6800,#6799 fhir/std api sort/pagination

### LINKED ISSUE #6799: feat: API support for _sort parameter
Provide support in the API (standard / FHIR) for sorting the results.  We should follow the FHIR format which uses the _sort parameter in FHIR as defined here: https://www.hl7.org/fhir/search.html#_sort

For our initial round of this we should handle ascending and descending (noted by a dash '-') 

FHIR has a number of recommendations, I'll outline what I expect this feature to handle in those cases
- We do not currently support the _score parameter so that option will not be allowed in this round.
- Complex elements such as Patient.name will sort on parameters from left to right order as defined in the FhirSearchParameterDefinition::$mappedFields property
- Sorting on string search parameters will be done using the underlying table collation, most of these are set to case insensitive, but if the collation is case sensitive there is no data casting that will occur.
- We will not support in this initial go around the :text modifier for Token / Reference values
- We do not have to worry about period or other esoteric date types as all of our current API types are of date/datetime instances.

### LINKED ISSUE #6800: feat: Patient api endpoints support _lastUpdated search parameter
We should support searching (and when we support sorting on the API) on the _lastUpdated field in the patient api (both standard and FHIR).  Currently we don't expose this property which makes it difficult to do any kind of filtering on the patient list based on what was last changed.

API consumers would like to be able to retrieve a list of patients who have been updated within the last 24 hours and within the last 7 days.

adunsulag:
Fixes #6800 
Fixes #6799 

Add common _sort parameter that can be used in the _rest_routes

Add date,_offset,_limit query parameters to the /api/patient route

Added helper method QueryUtils::escapeLimit for escaping limit parameters in SQL queries.

Changed standard api to return the links parameters if there are paginated responses for the search (someone searched with the _limit/_offset parameters).

Added to FhirPatientService the _lastUpdated search parameter that can be used.

Changed FHIR api to support in the base service the ability to do a searchForOpenEMRRecordsWithConfig method that will allow more customized configuration options in the future in addition to the basic search parameters.  It currently supports pagination and sort order.  Right now only the patient service supports the options.

I modified the ResourceServiceSearchTrait to be able to handle the _sort,_limit,_count parameters and put them all in a _config property that is handled in the FhirServiceBase class to generate the search config and pass it down to services that implement the searchForOpenEMRRecordsWithConfig method.

This is a work in progress commit as the FHIR links in the bundles still need to be populated.  Also need to do more extensive testing with the FHIR api on the sort order and _offset/_limit parameters.

Not sure I like seperating out the clause builder for the where clause and having a separate functionality for the orderby / pagination pieces. It might make sense to have it all in one central function governed by the search parameters.  Not sure if I like the centralization, or if it violates the Separation of Concerns principle.  Still thinking it through but this is an initial stab at things.  I really wish we had done doctrine back when we started all of this as it feels very much like we're getting further and further into reproducing all of the goodies of the doctrine ORM.  That'd be a major overhaul that would require a good bit of funding unfortunately.

### DIFF
--- a/_rest_routes.inc.php
+++ b/_rest_routes.inc.php
@@ -169,6 +169,26 @@
  *      name="standard-patient",
  *      description="Standard Patient Portal OpenEMR API"
  *  )
+ *  @OA\Parameter(
+ *          name="_sort",
+ *          in="query",
+ *          parameter="_sort",
+ *          description="The sort criteria specified in comma separated order with Descending order being specified by a dash before the search parameter name. (Example: name,-category)",
+ *          required=false,
+ *          @OA\Schema(
+ *              type="string"
+ *          )
+ *  )
+ *  @OA\Parameter(
+ *          name="_lastUpdated",
+ *          in="query",
+ *          parameter="_lastUpdated",
+ *          description="The date the resource was last updated.",
+ *          required=false,
+ *          @OA\Schema(
+ *              type="string"
+ *          )
+ *  )
  *  @OA\Response(
  *      response="standard",
  *      description="Standard Response",
@@ -304,6 +324,7 @@ use OpenEMR\RestControllers\PrescriptionRestController;
 use OpenEMR\RestControllers\ProcedureRestController;
 use OpenEMR\RestControllers\TransactionRestController;
 use OpenEMR\RestControllers\UserRestController;
+use OpenEMR\Services\Search\SearchQueryConfig;
 
 // Note some Http clients may not send auth as json so a function
 // is implemented to determine and parse encoding on auth route's.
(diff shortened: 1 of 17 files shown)


adunsulag:
@sjpadgett and @bradymiller I'd be interested in hearing your thoughts on how FHIR wants us to return a totalCount when we do pagination.  When I do that we end up doing 3 queries on every patient search instead of our previous 2 queries per search.  One to grab the count, one to grab the unique uuids, and one to grab the individual patients.

Is it worth complying with that part of the spec?  In FHIR its a 'should' but we may want to skip it.  For example anywhere in the codebase we do a getOne() operation on retrieving the patient we end up doing 3 queries.  Not sure sure I like it, but with a hot db cache it may not be that big of a deal.

adunsulag:
Hah, well I ran the API tests but apparently skipping the full test suite is a bad idea lol.

bradymiller:
2DV-2(1)

bradymiller (approved):

PR #6802, opened 2023-08-27
feat: Improve clinical note report UI

robertdown:
Move clinical note report view to Twig. Improve the UI by dropping tables in favor of a more content-focused layout. Also introduce `nl2br` to the note content as status quo did not respect line breaks.

# Before
image

# After
image

### DIFF
--- a/interface/forms/clinical_notes/report.php
+++ b/interface/forms/clinical_notes/report.php
@@ -17,6 +17,7 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Services\ClinicalNotesService;
 
 require_once(__DIR__ . "/../../globals.php");
@@ -31,38 +32,13 @@ function clinical_notes_report($pid, $encounter, $cols, $id)
         return $val['activity'] == ClinicalNotesService::ACTIVITY_ACTIVE;
     });
 
-    if ($data) {
-        ?>
-        <table class="table w-100">
-            <thead>
-            <tr>
-                <th class="border p-1"><?php echo xlt('Date'); ?></th>
-                <th class="border p-1"><?php echo xlt('Note Type'); ?></th>
-                <th class="border p-1"><?php echo xlt('Narrative'); ?></th>
-                <th class="border p-1"><?php echo xlt('Author'); ?></th>
-                <th class="border p-1"><?php echo xlt('Note Category'); ?></th>
-                <th class="border p-1"><?php echo xlt('Code'); ?></th>
-            </tr>
-            </thead>
-            <tbody>
-            <?php
-            foreach ($data as $key => $value) {
-                ?>
-                <tr>
-                    <td class="border p-1"><span class='text'><?php echo text($value['date']); ?></span></td>
-                    <td class="border p-1"><span class='text text-wrap'><?php echo text($value['codetext']); ?></span></td>
-                    <td class="border p-1"><span class='text'><?php echo text($value['description']); ?></span></td>
-                    <td class="border p-1"><span class='text'><?php echo text($value['user']); ?></span></td>
-                    <td class="border p-1"><span class='text text-wrap'><?php echo text($value['category_title']); ?></span></td>
-                    <td class="border p-1"><span class='text'><?php echo text($value['code']); ?></span></td>
-                </tr>
-                <?php
-            }
-            ?>
-            </tbody>
-        </table>
-        <?php
-    }
+    $viewArgs = [
+        'notes' => $data
+    ];
+
+    $twig = new TwigContainer(__DIR__, $GLOBALS['kernel']);
+    $t = $twig->getTwig();
+    echo $t->render('templates/report.html.twig', $viewArgs);
 }
 
 ?>
--- a/interface/forms/clinical_notes/templates/report.html.twig
+++ b/interface/forms/clinical_notes/templates/report.html.twig
@@ -1,0 +1,31 @@
+<div class="w-100">
+    {% for n in notes %}
+        <div class="d-flex justify-content-between bg-light border border-dark p-1">
+            <div class="form-group mb-0">
+                <div class="font-weight-bold">{{ "Date"|xlt }}</div>
+                <div>{{ n.date|date() }}</div>
+            </div>
+            <div class="form-group mb-0">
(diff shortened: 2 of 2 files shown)


bradymiller:
hi @robertdown , Will this also look fine for when there is more than 1 clinical note in there?

interface/forms/clinical_notes/templates/report.html.twig
```diff
@@ -0,0 +1,41 @@
...
+@copyright Copyright (c) 2023 Robert Down <robertdown@live.com>
+@copyright Copyright (c) 2023 Providence Healthtech
+@license https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+#}
+<div class="w-100">
+    {% for n in notes %}
+        <div class="d-flex justify-content-between bg-light border border-dark p-1">
+            <div class="form-group mb-0">
+                <div class="font-weight-bold">{{ "Date"|xlt }}</div>
+                <div>{{ n.date|date() }}</div>
```
bradymiller, line 16: would also run above through text
```
n.date|date()|text
```

interface/forms/clinical_notes/templates/report.html.twig
```diff
@@ -0,0 +1,41 @@
...
+            </div>
+            <div class="form-group mb-0">
+                <div class="font-weight-bold">{{ "Type"|xlt }}</div>
+                <div>{{ n.codetext|default("Unspecified"|xl)|text }}</div>
+            </div>
+            <div class="form-group mb-0">
+                <div class="font-weight-bold">{{ "Category"|xlt }}</div>
+                <div>{{ n.clinical_notes_category|default("Unspecified"|xl)|text }}</div>
```
bradymiller, line 24: above line is not matching `category_title` that was used (`<?php echo text($value['category_title']); ?>`)

bradymiller:
review completed (didn't test)

robertdown:
> hi @robertdown , Will this also look fine for when there is more than 1 clinical note in there?

I'll post a screenshot but it should just showcase the next note directly below, with the titles still rendered. I might try adding a sticky position to the meta data of the note so that as you scroll, the metadata remains at the top.

raskolnikov-rodion:
hi @robertdown , I've seen some of your PRs moving views to Twig, nice move. Are there any plans to also incorporate web components or other JS-based view tools any time soon?

interface/forms/clinical_notes/report.php
```diff
@@ -31,38 +32,11 @@ function clinical_notes_report($pid, $encounter, $cols, $id)
...
-                <?php
-            }
-            ?>
-            </tbody>
-        </table>
-        <?php
-    }
-}
+    $viewArgs = [
+        'notes' => $data
```
raskolnikov-rodion, line 36: In the previous code there was an `if` guarding against falsy values of `$data`. What happens if a falsy value is passed here? Could there be any errors?

robertdown: Shouldn't be a problem, however, I can always define the data variable prior to the query and only append if there are results.

interface/forms/clinical_notes/templates/report.html.twig
```diff
@@ -0,0 +1,41 @@
...
+            {% if n.code %}
+            <div class="form-group mb-0">
+                <div class="font-weight-bold">{{ "Code"|xlt }}</div>
+                <div>{{ n.code|text }}</div>
+            </div>
+            {% endif %}
+        </div>
+        <div>
+            {{ n.description|text|nl2br }}
+        </div>
```
raskolnikov-rodion, line 39: Consider using other html elements such as headers, p and span to make the view code more descriptive. It helps with accessibility too 🙂

robertdown:
> hi @robertdown , I've seen some of your PRs moving views to Twig, nice move. Are there any plans to also incorporate web components or other JS-based view tools any time soon?

That is a touchy subject. Some parts of OpenEMR are certainly primed for JS-based view rendering but we must consider the trade off between a JS based approach and increased complexity for developers. Because so much of the UI has been baked directly into PHP via echo commands mixed in with business logic, the decision to move to Twig outweighed a JS-based solution as a means to drive the codebase forward while not causing an untenable amount of changes.

There is probably a path forward that includes a mix of both Twig and JS solutions. However, given the relatively "Wild West" mentality of the JavaScript world, there is always the concern that a JS based tool just dies one day and loses support. That is true of any tool (even Twig) but JS seems to be particularly susceptible to the "whatever is latest and greatest."

I've been looking at HTMX lately, which seems to be a middle ground between full server side rendering and JS based approaches.

Twig is in the codebase to stay for the foreseeable future, and I'm not sure at this point moving to a JS based approach would be too disruptive. I try to find a middle ground where I can.

interface/forms/clinical_notes/templates/report.html.twig
robertdown: Wrapping that in a p tag would result in the entire note being one paragraph - not an ideal situation. I agree with you in principal, but practically it won't work here, paragraphs would need to be managed upstream.

There is a balance between enough progress to move the codebase forward and 100% perfect semantics - at the moment I'm leaning a bit more on practical over perfect. It's a long held battle for me, because I have been one to let perfect get in the way of good for many years.

interface/forms/clinical_notes/report.php
robertdown: Actually, it looks like data is already returning an array regardless of whether there are notes.

interface/forms/clinical_notes/templates/report.html.twig
raskolnikov-rodion: I see your point. I was mostly refering to the divs that hold data inside the loop, but I guess that's on me for inserting the comment at this point of the template. I don't think it's a blocker, done and functional is better than perfect

raskolnikov-rodion:
Thanks for the info @robertdown !

robertdown:
> hi @robertdown , Will this also look fine for when there is more than 1 clinical note in there?

Yes, it just keeps looping down the page like before

interface/forms/clinical_notes/templates/report.html.twig
robertdown: Fixed

interface/forms/clinical_notes/templates/report.html.twig
robertdown: You nudged me a bit... swapping out the outermost `div` for an `article role="document"` and inner `divs` to `header` and `section`

PR #6811, opened 2023-08-30
fix: several clinical_notes form fixes

bradymiller:
1. Sql query fix (grabbing of the category stuff from list_options was broken)
2. Translation fixes
3. Escaping fix

### DIFF
--- a/interface/forms/clinical_notes/templates/report.html.twig
+++ b/interface/forms/clinical_notes/templates/report.html.twig
@@ -14,15 +14,15 @@ The Report template of the clinical note form.
             <header class="d-flex justify-content-between bg-light border border-dark p-1 sticky-top">
                 <div class="form-group mb-0">
                     <div class="font-weight-bold">{{ "Date"|xlt }}</div>
-                    <time datetime="{{ n.date|date()|text }}">{{ n.date|date()|text }}</time>
+                    <time datetime="{{ n.date|date()|attr }}">{{ n.date|date()|text }}</time>
                 </div>
                 <div class="form-group mb-0">
                     <div class="font-weight-bold">{{ "Type"|xlt }}</div>
-                    <div>{{ n.codetext|default("Unspecified"|xl)|xlt }}</div>
+                    <div>{{ n.codetext is defined ? n.codetext|xlListLabel|text : "Unspecified"|xlt }}</div>
                 </div>
                 <div class="form-group mb-0">
                     <div class="font-weight-bold">{{ "Category"|xlt }}</div>
-                    <div>{{ n.category_title|default("Unspecified"|xl)|text }}</div>
+                    <div>{{ n.category_title is defined ? n.category_title|xlListLabel|text : "Unspecified"|xlt }}</div>
                 </div>
                 <div class="form-group mb-0">
                     <div class="font-weight-bold">{{ "Author"|xlt }}</div>
--- a/interface/forms/clinical_notes/templates/report.html.twig
+++ b/interface/forms/clinical_notes/templates/report.html.twig
@@ -18,11 +18,11 @@ The Report template of the clinical note form.
                 </div>
                 <div class="form-group mb-0">
                     <div class="font-weight-bold">{{ "Type"|xlt }}</div>
-                    <div>{{ n.codetext is defined ? n.codetext|xlListLabel|text : "Unspecified"|xlt }}</div>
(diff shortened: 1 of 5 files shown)


interface/forms/clinical_notes/templates/report.html.twig
```diff
@@ -14,15 +14,15 @@ The Report template of the clinical note form.
             <header class="d-flex justify-content-between bg-light border border-dark p-1 sticky-top">
                 <div class="form-group mb-0">
                     <div class="font-weight-bold">{{ "Date"|xlt }}</div>
-                    <time datetime="{{ n.date|date()|text }}">{{ n.date|date()|text }}</time>
+                    <time datetime="{{ n.date|date()|attr }}">{{ n.date|date()|text }}</time>
                 </div>
                 <div class="form-group mb-0">
                     <div class="font-weight-bold">{{ "Type"|xlt }}</div>
-                    <div>{{ n.codetext|default("Unspecified"|xl)|xlt }}</div>
+                    <div>{{ n.codetext is defined ? n.codetext|xlListLabel|text : "Unspecified"|xlt }}</div>
```
robertdown, line 21: Generally I am a big fan of ternary operators, and while this code will get to the same end, the following would be a cleaner solution
```
{{ n.codetext|xlListLabel|default("Unspecified"|xlt) }}
``

It is more "Twig-ish" in its processing. The only caveat would be if `xlListLabel` interferes with `default()` - I'd have to test.

**Comment by sjpadgett** [comment]
umm, this form is a big part of CCDA certification and was thoroughly tested and worked well so beware.
I'm very happy to see work to better organize and improve the form so please don't get me wrong. I'm just worried

Still I can retest for cert compliance once done.

File: interface/forms/clinical_notes/templates/report.html.twig (hunk 1 above)
Comment (bradymiller) [reply interface/forms/clinical_notes/templates/report.html.twig:21]: The issue with that solution is if the translation ends up taking a string and making it empty, then it will erroneously set the title of that value to Unspecified rather than just making it blank (this type of thing should not happen, but if it does the ternary operator would avoid that weird scenario).

Comment (bradymiller) [reply interface/forms/clinical_notes/templates/report.html.twig:21]: i wonder if `is defined` will cover everything though ...? (ie. also need to check if empty etc. ?)

**Comment by bradymiller** [comment]
@sjpadgett , Note the sql query was broken (it was not returning the category or category code making the category blank in the report.php). I also placed the translated title type in a separate variable so it wouldn't mess around with the current logic in new/view.php (so the untranslated title is in same place for the current logic and stored in forms_clinical_notes entry (ie. the translated title is only there for vanity :) ).

File: interface/forms/clinical_notes/templates/report.html.twig [hunk 2]
```diff
@@ -35,7 +35,7 @@ The Report template of the clinical note form.
                 </div>
                 {% endif %}
             </header>
-            <section>{{ n.description|text|nl2br }}</section>
+            <section class="mb-3">{{ n.description|text|nl2br }}</section>
```
bradymiller, line 38: @robertdown , not sure if this is best place to increase the bottom margin of each separate note (without separation, it made it more difficult to read them since they were all squashed together)

interface/forms/clinical_notes/templates/report.html.twig
bradymiller: turns out need to use `is not empty`

src/Common/Twig/TwigExtension.php
```diff
@@ -308,6 +308,12 @@ function ($string) {
                     return oeFormatShortDate($string);
                 }
             ),
+            new TwigFilter(
+                'xlListLabel',
+                function ($string) {
+                    return xl_list_label($string);
+                }
+            ),
```
stephenwaite, line 316: smaller_excited_elephant

interface/patient_file/encounter/forms.php
```diff
@@ -867,7 +867,7 @@ function findPosY(obj) {
         }
 
         // $form_info = getFormInfoById($iter['id']);
-        $form_class_list = (strtolower(substr($ite['form_name'], 0, 5)) == 'camos') ? "" : "text onerow";
+        $form_class_list = (strtolower(substr($iter['form_name'], 0, 5)) == 'camos') ? "" : "text onerow";
```
stephenwaite, line 870: Cat_laugh

bradymiller:
testing well and bringing well

PR #6817, opened 2023-08-30
Add disable forced reset credential flag on portal login reset/create

### LINKED ISSUE #6813: Add ability to prevent forced portal credential change
### Describe the feature
Currently we force patient to change their credentials when portal credentials are created from MRD.
Some admins prefer to prevent this action.

Add and persist a checkbox flag to create_portallogin to set the patient_access_onsite  portal_pwd_status column to prevent.

A nice addition while here would be to create a onetime to auto login and take patient to say portal documents to create intake forms or to appointments or even the portals change password dialog.

sjpadgett:
Add user setting to persist option for the user setting it in MRD
add checkbox and persist js function to template.

Fixes #6813

### DIFF
--- a/interface/patient_file/summary/create_portallogin.php
+++ b/interface/patient_file/summary/create_portallogin.php
@@ -46,16 +46,28 @@ function displayLogin($patient_id, $message, $emailFlag)
 
 $patientAccessOnSiteService = new PatientAccessOnsiteService();
 $credentials = $patientAccessOnSiteService->getOnsiteCredentialsForPid($pid);
-$forced_reset_disable = PatientAccessOnsiteService::fetchUserSetting('portal_login.credential_reset_disable');
+
+$option = $GLOBALS['portal_force_credential_reset'] ?? '0';
+if ($option == '2') {
+    $forced_reset_disable = PatientAccessOnsiteService::fetchUserSetting('portal_login.credential_reset_disable');
+} elseif ($option == '0') {
+    $forced_reset_disable = 0; // sets database to force reset on login
+} else {
+    $forced_reset_disable = 1; // sets database to ignore force reset on login
+}
 
 $credMessage = '';
 if (isset($_POST['form_save']) && $_POST['form_save'] == 'submit') {
     if (!CsrfUtils::verifyCsrfToken($_POST["csrf_token_form"])) {
         CsrfUtils::csrfNotVerified();
     }
-
+    if ($option == '2') {
+        $forced_reset_disable = $_POST['forced_reset_disable'] ?? 0;
+    } else {
+        $forced_reset_disable = $option;
+    }
     // TODO: @adunsulag do we clear the pwd variables here?? Hard to break it out into separate functions when we do that...
-    $result = $patientAccessOnSiteService->saveCredentials($pid, $_POST['pwd'], $_POST['uname'], $_POST['login_uname'], $_POST['forced_reset_disable']);
+    $result = $patientAccessOnSiteService->saveCredentials($pid, $_POST['pwd'], $_POST['uname'], $_POST['login_uname'], $forced_reset_disable);
     if (!empty($result)) {
         $emailResult = $patientAccessOnSiteService->sendCredentialsEmail($pid, $result['pwd'], $result['uname'], $result['login_uname'], $result['email_direct']);
         if ($emailResult['success']) {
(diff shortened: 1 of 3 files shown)


sjpadgett:
Decided to put up separate PR to ease backport if wanted @adunsulag 
I also decided to make the flag to persist by user. I didn't want a global which is logical but would make it harder to backport.

adunsulag:
@sjpadgett By putting it on the patient dashboard won't users think the setting will only apply to that one patient instead of it applying globally to all patients?  If someone checks/unchecks the box for a single patient I'm not sure they'll realize they ended up changing the setting for all patients.  If I'm following the code correctly, I believe that's how things will work with this commit right?

bradymiller:
Would make this something that the admin gets to set (ie. global).

interface/patient_file/summary/create_portallogin.php
```diff
@@ -12,21 +12,20 @@
...
  */
 
 require_once("../../globals.php");
 require_once('../../../library/amc.php');
 
-use OpenEMR\Common\ {
-    Csrf\CsrfUtils,
-    Utils\RandomGenUtils,
-};
+use OpenEMR\Common\{Csrf\CsrfUtils,};
```
bradymiller, line 28: was this meant to remove Utils\RandomGenUtils ?

sjpadgett: not used so not needed

sjpadgett:
I don't want a global. I want to leave it up to the user and the ability to decide patient by patient if needed. It will persist for each user therefore once a user sets it, it will remain set and up to user to decide to unset.

Because one admin decides they don't like the forced reset that doesn't mean all users agree especially in practice groups.

It can be seen by many as a security hole.

interface/patient_file/summary/create_portallogin.php
bradymiller: sounds good, then would change to since no longer compounding:
```
```

bradymiller:
To clarify, trying to ensure no security issues (let me know if i am missing something).

My concern regarding security here is that if a practice wishes to enforce high security (make the user change the password), which has been the status quo, this will essentially remove that protection (since now any user that creates an account gets to decide this; ie. might as well not even use it then in that case). Makes more sense from a practice perspective for an admin/practice to decide what to do and then make it uniform. Having lay users decide likely mean they will just use whatever is set on toggle (and they will likely click is either way by mistake every once in awhile until the next mistaken click :) ). Also recommend keeping status quo on the global (ie. enforce the password change).

sjpadgett:
I disagree but don't care enough to argue the point. Closing for Stephen to pick up.

interface/patient_file/summary/create_portallogin.php
sjpadgett: btw already is

sjpadgett:
I hope this compromise satisfies everyone's concerns or wishes.

interface/patient_file/summary/create_portallogin.php
bradymiller: sounds like maybe issue with PR since I see `use OpenEMR\Common\{Csrf\CsrfUtils,};` instead of `use OpenEMR\Common\Csrf\CsrfUtils;` in line above

bradymiller:
Looks great! (noted you have the global in another PR)

bradymiller:
tenor99

sjpadgett:
One complaint I often get is from practices that have mental health and/or seniors or otherwise patients that are not computer savvy and want the sign in, register and general portal use easier. This is the main reason to allow force reset to be optional per patient on their credential create/reset.
Secondary we have to get away from using globals for every little option and should refrain when possible. In this case by checking the option in dialog it would be the same as setting in globals but provide the option to change as needed. This is UX 101 stuff.
My old man mental capacity may be affecting my coding, my creativity and ability to think features through is still as sharp as ever. My communication skills on the other hand still needs improvement!:)

bradymiller:
Two opposing thoughts here. With a provider hat on, agree with being a patient advocate. With a security hat on, sadly it is those patients (no computer savviness) that benefit most from the additional security (ie. they are most likely to not control access to their email etc. ) :)

sjpadgett:
I almost feel like I need to develop a single sign on option for the portal.
I don't think the one time token would be available option for extended login

PR #6821, opened 2023-08-31
Bug openemr fix #6820 fax api call

### LINKED ISSUE #6820: bug: Fix fax/sms module event usage from API dispatch.php call
Currently if you fire the SendNotificationEvent event inside an API call it will be silently ignored in the fax/sms module because the API request pathway does not populate the session_database_uuid value that the SessionTracker class populates.  We need to change the values used in the sms/fax module to determine if the session is populated to make sure the user is logged in.

adunsulag:
Fixes #6820 

Fixed it so that if you fire the SendNotificationEvent event inside an API
call it will no longer be silently ignored in the fax/sms module.

Changed the module to use the user_id or pid in the session as a
determinant of the session being populated instead of the
session_database_uuid.

Fixed spelling error in the email subject being sent.


sjpadgett (approved):
will merge once test complete. Thanks for the fix and sorry Stephen as I completely forgot this change.

PR #6823, opened 2023-09-01
refactor: Improve the Flowboard UI

robertdown:
No twig, just some UI improvements.

### Todo
- [ ] Toggle filter button chevron flip
- [ ] Settings popup

### Long Term Needs (Will not be addressed in this PR, but should be done)
- [ ] Refactor the data management to a controller which manages the returning data based only the filters provided and only return JSON.
- [ ] Twig the base
- [ ] Events
- [ ] Pull the MedEx stuff out and into a separate module (Dependent on the correct Event structure being setup)


usama-kabel:
Is it possible to convert the table to a data table for the many rows retrieved?

PR #6829, opened 2023-09-03
fix static domain included in URL

sjpadgett:
Forgot to refactor full URL's that includes static domain from when testing a session issue.


sjpadgett:
thanks Brady

PR #6831, opened 2023-09-04
Update view.php

ophthal:
Shorthand icon reposition

Fixes #


robertdown:
What icon does `fa-2` map to? Should it be a bootstrap margin helper of `m-2`?

ophthal:
I can’t say I understand any css since the ui changes.  In 7.0 the short-hand icon in every panel and theme is a couple px too high. Removing these two classes fixes it.  There are lots of other issues with the ui and the eye form most of which are not present in light and manila theme.  This seems to fix the shorthand icon position anyway.On Sep 4, 2023, at 12:49 PM, Robert Down RN ***@***.***> wrote:﻿
What icon does fa-2 map to? Should it be a bootstrap margin helper of m-2?

bradymiller:
maybe was meant to be fa-2x ?

bradymiller (approved):

stephenwaite:
jumping

PR #6835, opened 2023-09-09
styling issue

claimrevolution:
Fixes #

Robert made a small change for bootstrap sytling


bradymiller:
looks good and bringing in (will also put thus change into the rel-701 branch)

PR #6836, opened 2023-09-11
fix: fix summary screen cards expand and collapse persistence

bradymiller:
Mostly fixed. Still couple more things to work out.


bradymiller:
Going to bring this in now and work on other bugs as separate PR(s)

PR #6837, opened 2023-09-11
refactor(ccdaservice): test and refactor populateDemographics function

raskolnikov-rodion:


ccdaservice/utils/demographics/populate-demographics.js
```diff
@@ -0,0 +1,169 @@
...
+        {
+            relation: guardian.relation,
+            addresses: [
+                {
+                    street_lines: [guardian.address],
+                    city: guardian.city,
+                    state: guardian.state,
+                    zip: guardian.postalCode,
+                    country: guardian.country || 'US',
+                    use: 'primary home',
```
adunsulag, line 31: I've never really liked trailing commas due to switching between JSON vs non, but I'm guessing this is the style syntax we are following?

ccdaservice/utils/demographics/populate-demographics.js
```diff
@@ -0,0 +1,169 @@
...
+            ? 'sp-US'
+            : 'en-US';
+}
+
+function populateDemographics(patient, guardian, documentData) {
+    const oidFacility =
+        documentData.encounter_provider.facility_oid ||
+        '2.16.840.1.113883.19.5.99999.1';
+    const npiFacility =
+        documentData.encounter_provider.facility_npi || NOT_INFORMED;
```
adunsulag, line 64: I'm not sure this change is correct and would need to be tested / looked at.  You took the original line from 3228 in generateUnstructured which sets npiFacility to be: 
```
npiFacility = all.encounter_provider.facility_npi || "NI";
```
However in generateCcda on line 2754 on npiFacility does not have an option for "NI":
```
npiFacility = all.encounter_provider.facility_npi;  
```
@sjpadgett does the spec allow "NI" for a structured document?

ccdaservice/utils/demographics/populate-demographics.js
```diff
@@ -0,0 +1,169 @@
...
+                preferred: true,
+                mode: 'Expressed spoken',
+                proficiency: 'Good',
+            },
+        ],
+        attributed_provider: {
+            identity: [
+                {
+                    root: '2.16.840.1.113883.4.6',
+                    extension: npiFacility,
```
adunsulag, line 135: You've dropped the ```|| ""``` clause here.  @sjpadgett can this value be "NI" instead of an empty string?

ccdaservice/utils/demographics/populate-demographics.mock.js
```diff
@@ -0,0 +1,541 @@
...
+    birth_fname: 'Lilly',
+    dob: '1969-09-25',
+    gender: 'female',
+    uuid: '052698',
+    status: 'married',
+    phone_home: '111-1111',
+    phone_mobile: '222-2222',
+    phone_work: '333-3333',
+    phone_emergency: '444-4444',
+    email: 'cathy@zeta.jones',
```
adunsulag, line 30: I recommend either using ```@example.com``` or one of the possible sample emails that we know won't resolve instead of us accidentally sending email to someone in the future.

adunsulag (commented):
Overall looks fine but I have questions about the changes made to the npiFacility object and am concerned if it changes behavior.

ccdaservice/utils/demographics/populate-demographics.js
raskolnikov-rodion: I'm following this style but it's totally optional, the linter doesn't enforce it. Quoting MDN:

> Trailing commas (sometimes called "final commas") can be useful when adding new elements, parameters, or properties to JavaScript code. If you want to add a new property, you can add a new line without modifying the previously last line if that line already uses a trailing comma. This makes version-control diffs cleaner and editing code might be less troublesome.

ccdaservice/utils/demographics/populate-demographics.js
raskolnikov-rodion: Good point, I'm not sure if the abscence of a fallback value in line 2754 is intentional or not so it'll be good to clarify that. In any case, if it's necessary to prevent "NI" in structured documents, it should be easy to accommodate for that in the refactored version

ccdaservice/utils/demographics/populate-demographics.js
raskolnikov-rodion: Yes, the clause should not be necessary after initializing the value with the proper fallback

ccdaservice/utils/demographics/populate-demographics.mock.js
raskolnikov-rodion:
```suggestion
    email: 'cathy@example-openemr.com',
```

Got it, thanks for the suggestion!

ccdaservice/utils/demographics/populate-demographics.js
sjpadgett: Leave any attributes or replace values as they may have significance to builder. Unless a default is set.

ccdaservice/utils/demographics/populate-demographics.js
adunsulag: Looks like you resolved this.

ccdaservice/utils/demographics/populate-demographics.js
adunsulag: Latest changes keep the fallback and switch between unstructured and structured.  I consider this now resolved.

adunsulag (approved):
I don't see any additional problems here.  @sjpadgett did you want to hold off on bringing this in until you've had a chance to look at it?  I know you mentioned it'd likely be next week.  If you're good we can bring this in, otherwise I'm fine holding it.

sjpadgett:
bring it but it still needs to be tested against ONC test document generation.

ccdaservice/serveccda.js
```diff
@@ -24,8 +24,9 @@ const { fDate, templateDate } = require('./utils/date/date');
 const { countEntities } = require('./utils/count-entities/count-entities');
 const { populateTimezones } = require('./utils/timezones/timezones');
 const {
-    fetchPreviousAddresses,
-} = require('./utils/demographics/previous-addresses');
```
bradymiller, line 28: was the above line supposed to be removed?

raskolnikov-rodion: Yes, this function is not needed in this file anymore :)

bradymiller:
bringing in. thanks @raskolnikov-rodion !

PR #6842, opened 2023-09-12
Adding event to billing deleting of payment

### LINKED ISSUE #6841: Add an event to billing payment deletion
### Describe the problem

For the credit card processing, it would be nice to have an event fired off. T

Build event class.
Add event dispatch to the search payments where the patient payments are deleted

juggernautsei:
Fixes #6841

Add class for billing event when payments are deleted

Edits to search_payment.php
New class added to the Events folder.

### DIFF
--- a/interface/billing/search_payments.php
+++ b/interface/billing/search_payments.php
@@ -58,7 +58,7 @@ if (isset($_POST["mode"])) {
             }
         }
         //dispatch this payment is being deleted trigger refund process
-        $eventDispatcher->dispatch(new DeletePayment(add_escape_custom($DeletePaymentId)), DeletePayment::ACTION_DELETE_PAYMENT, 10);
+        $eventDispatcher->dispatch(new DeletePayment($DeletePaymentId), DeletePayment::ACTION_DELETE_PAYMENT, 10);
     //delete and log that action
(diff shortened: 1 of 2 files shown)


interface/billing/search_payments.php
```diff
@@ -51,7 +57,8 @@
                 sqlStatement("update form_encounter set last_level_closed=last_level_closed - 1 where pid =? and encounter=?", [$PId, $Encounter]);
             }
         }
-
+        //dispatch this payment is being deleted trigger refund process
+        $eventDispatcher->dispatch(new DeletePayment(add_escape_custom($DeletePaymentId)), DeletePayment::ACTION_DELETE_PAYMENT, 10);
```
adunsulag, line 61: I wouldn't escape the payment id here.  Downstream consumers will then have to remove the escaping it if they use it for something different than database insertion.

adunsulag (changes requested):

interface/billing/search_payments.php
juggernautsei: I understand what you are saying, and in my use case, that'd be great. I only use the ID to retrieve information once I determine that the credit card processor makes the entry in the database.

adunsulag (approved):
This looks good to me.

bradymiller:
dance-bear

PR #6847, opened 2023-09-12
feat: Fixes #6839 show mfa shared secret

### LINKED ISSUE #6839: feat: MFA TOTP should display a seed code to enter into an application that does not use QR Codes
If an application that generates TOTP does not have a QR scanner a seed code should be displayed on the TOTP user screen in the MFA Management screen.

Most TOTP systems will display both the QR code and the seed value.  I ran into this problem when trying to test the login screen with MFA.

adunsulag:
In order to use a non-qr type TOTP code we need to show the shared secret when going to add the code.

Fixes #6839

### DIFF
--- a/interface/usergroup/mfa_totp.php
+++ b/interface/usergroup/mfa_totp.php
@@ -185,6 +185,8 @@ $user_full_name = $user_name['fname'] . " " . $user_name['lname'];
                                             <br />
                                             <img src="<?php echo attr($qr); ?>" class="img-responsive center-block" style="height:200px !Important"/>
                                             <br />
+                                            <p><?php echo xlt("Or paste in the following code into your authenticator app"); ?></p>
+                                            <p><?php echo $mfaAuth->getSecret(); ?></p>
                                             <p><?php echo xlt('Example authenticator apps include'); ?></p>:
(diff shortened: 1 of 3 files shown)


interface/usergroup/mfa_totp.php
```diff
@@ -185,6 +185,8 @@ function docancel() {
                                             <br />
                                             <img src="<?php echo attr($qr); ?>" class="img-responsive center-block" style="height:200px !Important"/>
                                             <br />
+                                            <p><?php echo xlt("Or paste in the following code into your authenticator app"); ?></p>
+                                            <p><?php echo $mfaAuth->getSecret(); ?></p>
```
bradymiller, line 189: also should do at https://github.com/openemr/openemr/blob/master/setup.php#L1377-L1404
(this allows setting up mfa on the user during openemr installation)

adunsulag: @bradymiller Thanks for catching that.  Added to the setup.

sjpadgett: Yep I call him ole Eagle Eye! 😉

bradymiller (approved):

bradymiller:
2DV-2(2)

PR #6855, opened 2023-09-14
Openemr rel701 backport 6497 6499 6500 6503 6505 6509 6504 6507 6506 6566

### LINKED ISSUE #6497: bug - FHIR Confidential Client fails CORS OPTIONS check in browser window when requesting /token endpoint
The CORS OPTIONS check that gets made when using the Authorization: Basic header to receive a token from the /token endpoint fails with a 400 status error.  The code attempts to treat the OPTIONS check as the regular POST request and looks for the token_type.  This throws an 400 status error.  

We didn't detect this as our API checks grab the headers which have the CORS authorization and moves on.  The browser environments such as chrome correctly checks the status code and fails the CORS check if the status is not 200.

We need to just return a valid 200 when we do an OPTIONS request and stop the request at that point.

### LINKED ISSUE #6499: bug - Telehealth Patient Updated event signature type broken with patient update by uuid
If the patient is updated by the uuid (typically occurs in the Patient FHIR PUT) it fires off the updated event.  Currently the event signature on that is very different than the one in the patient save operation.  It sends a ProcessingResult object as the original patient data.  That object has a patient record array in it with a string converted uuid instead of the binary uuid in the insert event.

We need these interfaces to be the same so module writers can rely on the consistency.  Right now if the Telehealth module is enabled and someone updates a patient record via the API it throws an exception in the module due to the incongruence of the event interfaces.

### LINKED ISSUE #6500: bug - Telehealth Transfer Appointment fails to launch appointment on latest 7.0.1 patch
If a Telehealth enabled user attempts to launch a Telehealth session that belongs to another provider it is supposed to transfer the appointment to the currently logged in user and launch the session.  The system then creates a duplicate appointment and marks it as a transferred appointment for its status.  

However, in the latest patch the Telehealth appointment is reassigned to the current user but the process fails when it attempts to create the duplicate transferred appointment.  This causes the whole launch sequence to fail and the user sees an error message that their session failed to launch.

The actual error that gets saved into the logs is the following:
```
SQL Error with statement:insert failed:  INSERT INTO openemr_postcalendar_events SET     uuid=?,     pc_pid=?,     pc_catid=?,     pc_title=?,     pc_duration=?,     pc_hometext=?,     pc_eventDate=?,     pc_apptstatus=?,     pc_startTime=?,     pc_endTime=?,     pc_facility=?,     pc_billing_location=?,     pc_informant=1,     pc_eventstatus=1,     pc_sharing=1,     pc_aid=?--Column &#039;pc_duration&#039; cannot be null
```

It looks like the pc_duration is not being populated and that is why the error triggers.  Earlier versions of OpenEMR treated the pc_duration null value as a 0 and saved appropriately.

### LINKED ISSUE #6503: Change HttpRestRequest class to implement PSR7 ServerRequestInterface
Our HttpRestRequest that is used in our APIs to get information about the current request interface should be PSR7 interoperable so we can leverage other libraries and frameworks methods that consume a PSR7 ReqestInterface or PSR7 ServerRequestInterface.  This will be a huge boon to module writers as well as to core API functionality.

### LINKED ISSUE #6504: Add RestAPISecurityCheckEvent to allow module writers to extend or change the API security checks.
We should allow module writers to change up the security checks on an individual API endpoint request.  Functionality could be to add additional security checks, or to open up apis for experimental usage that are currently locked down in the OpenEMR installation.

The particular use case I need to address is allowing Questionnaire and QuestionnaireResponse apis to be writeable in the patient context.  Our current security checks prevent ANY patient context writes while using FHIR to the patient record which makes the ability for patients to take Questionnaires and create QuestionnaireResponse records unworkable.  Rather than changing core, a module writer becomes responsible for handling these security checks if they decide to open up specific endpoints for specific use cases.

### LINKED ISSUE #6505: Add to UtilsService methods for dealing with FHIR Canonical URLs, OperationOutcomes, and Reference resources
FHIR resources have to deal with OperationOutcomes when creating resources and dealing with error conditions.  Some FHIR resource types need to work with canonical urls as well as reference resources and we should have more helper methods to parse and create these resources.

### LINKED ISSUE #6506: Add core event hooks to QuestionnaireResponse save/insert records.
We should allow module writers to hook into the QuestionnaireResponse save/insert action to be able to do additional processing when a QuestionnaireResponse is saved (such as updating tasks, sending emails, creating additional FHIR resources, etc).

### LINKED ISSUE #6507: Questionnaire URL needs to be a cannonical FQDN url
The FHIR specification requires that the Questionnaire FHIR resource's URL be a fully qualified domain name url with https being the preferred schema used.  SMART on FHIR clients that consume Questionnaire resources from the FHIR server need to have an actual canonical URL to go and grab the resource from so we need to update this in our system.

I do see a couple of challenges with this approach.  If someone changes the FHIR oauth2 base server url location, then these URLs will no longer be resolvable.  If someone is doing multi-site, the questionnaire will be specific to the site that they used which I think is ok.  If someone copies a database over from one location to another, these URLs will be out of sync, but that is something the migration will have to deal with.

Also, I'm not sure of a good mechanism to deal with existing data as any created questionnaires before this fix will need to have the data migrated over.

### LINKED ISSUE #6509: Extend PatientService to have methods for retrieving a list of the PCP for a given list of patients
We should add helper methods to the PatientService class to retrieve a list of the assigned primary care provider for a given list of patients.

### LINKED ISSUE #6566: Fix LForms Questionnaire Copyright Copy issue

adunsulag:
This is an empty PR because the files were copied over directly into rel-701 via this commit: 36f24522bcbc2cc4b3bcab7cdeaa06ca3d64c91a

We still want to bring in the changelogs for it though as the commit the files were copied into do not have the changelog history.

Fixes #6497 
Fixes #6499 
Fixes #6500
Fixes #6503
Fixes #6505
Fixes #6509
Fixes #6504
Fixes #6507
Fixes #6506
Fixes #6566


adunsulag:
We decided not to do a patch release and are doing a version release. So closing this PR.

PR #6860, opened 2023-09-17
fix: fix none mechanism in issues gui

bradymiller:
fix: fix none mechanism in issues gui


bradymiller:
testing well so bringing in

PR #6862, opened 2023-09-18
fix: patient summary screen fix for receptionist

bradymiller:
fix: patient summary screen fix for receptionist


bradymiller:
testing well and bringing in

PR #6867, opened 2023-09-20
chore: packages update and min php version 8.1

bradymiller:
chore: packages update and min php version 8.1


composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "openemr/openemr",
-    "version": "7.0.0",
+    "version": "7.0.2",
```
stephenwaite, line 3: cool

interface/procedure_tools/gen_universal_hl7/gen_hl7_order.inc.php
```diff
@@ -532,7 +532,7 @@ function send_hl7_order($ppid, $out)
         }
 
         // Connect to the server and write the file.
-        $sftp = new \phpseclib\Net\SFTP($remote_host);
+        $sftp = new \phpseclib3\Net\SFTP($remote_host);
```
stephenwaite, line 535: we could add the `php-ssh2` extension and remove this package someday?

stephenwaite: actually the php-ssh2 extension doesn't have the handy classes that the package has

bradymiller: yep, there are a couple pain points so far with the package updates.
The google/apiclient updated package is requiring going to phpseclib3 which will involve refactoring code in at least 1 place (sha384 public key verify stuff) and also guessing here.
Even worse it the most updated lcobucci/jwt package has decided to remove the `forUnsecuredSigner` stuff which is causing the normal bloating, nausea, diaphoresis :)
And I am guessing there will be more :)

stephenwaite:
nice

bradymiller:
Testing well and bringing in.

"Let the Games Begin" :) : https://www.open-emr.org/wiki/index.php/QA_and_Release_Process#Version_7.0.2


dance-bear

sjpadgett:
You are a pro Brady no doubt

PR #6880, opened 2023-09-23
fix: Clinical report not working when searching for patients with a specific medical problem

### LINKED ISSUE #6879: Clinical reports are not filtering by diagnosis
### Describe the bug

If we want to generate a clinical report (Reports > Clients > Clinical) and we set a problem DX from any external code (ICD10 for example) and you submit the form, you will always have a result showing all patients (as if you have no selected any Problem DX)

### To Reproduce

image

Select any asthma code (J45.20 in my example)
image

image

1. Add a Problem DX, by selecting a code from the list (ICD10) then in search for you will enter some problem (asthma, for example)
2. Click on code (J45.20)
3. Scroll down to submit
4. Click on Submit
5. See error (all patients will be listed)

### Expected behavior

I was expecting to see just asthma patients and its diagnosis.

### Client configuration

- Browser:Chrome
- OpenEMR version:7.0.1
- Operating system:Ubuntu 22.04

macomeza:
Fixes #6879

You can´t get patients that have had any medical problem (have associated external code problems), that filter is not working and every patient is been shown.

Modify clinical_reports.php to set sql statement that will filter by diagnosis.

### DIFF
--- a/interface/patient_file/download_template.php
+++ b/interface/patient_file/download_template.php
@@ -65,7 +65,6 @@ function dataFixup($data, $title = '')
         $data = str_replace('&', '[and]', $data);
         $data = str_replace('<', '[less]', $data);
         $data = str_replace('>', '[greater]', $data);
-        $data = str_replace("\r\n", '<text:line-break />', $data);
         // If in a group, include labels and separators.
         if ($groupLevel) {
             if ($title !== '') {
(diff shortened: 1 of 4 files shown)


macomeza:
@bradymiller could you please authorize this fix?

bradymiller:
hi @macomeza , Code looks good and bringing in. Thanks for the fix!

PR #6884, opened 2023-09-24
Dashboard card repositioning

### LINKED ISSUE #6883: Restyle front dashboard cards
There is a hundred ways this can be done however I think the general consensus is, three top cards with prescriptions on it own row. 
Really only minor adjustments to code as most everything is there, thanks Robert!

I decided to reduce Allergies and Issues to col-3 and medications to col-6 for readability.

image
image

Of course there may be other opinions so please let me know soon as I'm moving onward:)
btw: Thanks for the Allergy/issues card populate fixes @bradymiller !

sjpadgett:
Restyle cards
Fixes #6883 

- Change column sizes

### DIFF
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -1034,6 +1034,12 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                     $pl = (AclMain::aclCheckIssue('medical_problem')) ? 1 : 0;
                     $meds = (AclMain::aclCheckIssue('medication')) ? 1 : 0;
                     $rx = (!$GLOBALS['disable_prescriptions'] && AclMain::aclCheckCore('patients', 'rx')) ? 1 : 0;
+                    $cards = $allergy + $pl + $meds;
+                    $col = "p-1 ";
+
+                    $colInt = 12 / $cards;
+                    $col = "col-" . $colInt;
+
                     /**
                      * Helper function to return only issues with an outcome not equal to resolved
                      *
@@ -1065,7 +1071,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'btnLabel' => 'Edit',
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=allergy')"
                         ];
-                        echo "<div class=\"col-md-3\">";
+                        echo "<div class=\"$col\">";
                         echo $t->render('patient/card/allergies.html.twig', $viewArgs);
                         echo "</div>";
                     }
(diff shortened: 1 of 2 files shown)


interface/patient_file/summary/demographics.php
```diff
@@ -1034,14 +1034,6 @@ function setMyPatient() {
                     $pl = (AclMain::aclCheckIssue('medical_problem')) ? 1 : 0;
                     $meds = (AclMain::aclCheckIssue('medication')) ? 1 : 0;
                     $rx = (!$GLOBALS['disable_prescriptions'] && AclMain::aclCheckCore('patients', 'rx')) ? 1 : 0;
-                    $cards = $allergy + $pl + $meds + $rx;
-                    $col = "p-1 ";
-
-                    if ($cards > 0) {
-                        $colInt = 12 / $cards;
-                        $col = "col-" . $colInt;
-                    }
```
bradymiller, line 1043: Above was accounting for when user did not have access to all the issue categories
What happens if the user has access to medical_problem issues but not medication issues on the user interface?
(if looks a bit weird ok, just making sure no breakage)

bradymiller:
2DV-2(13)

interface/patient_file/summary/demographics.php
sjpadgett: ahhh wondered what that was about i'll put back thanks

sjpadgett: I'm not going to try to persist top expand. Really should make it responsive to screen sizes with something like container-xl which will grow page when needed.
I never understood why you need to indicate with change expand icon whether an item is expanded or not! Umm just look at the page.

interface/patient_file/summary/demographics.php
```diff
@@ -1034,13 +1034,11 @@ function setMyPatient() {
                     $pl = (AclMain::aclCheckIssue('medical_problem')) ? 1 : 0;
                     $meds = (AclMain::aclCheckIssue('medication')) ? 1 : 0;
                     $rx = (!$GLOBALS['disable_prescriptions'] && AclMain::aclCheckCore('patients', 'rx')) ? 1 : 0;
-                    $cards = $allergy + $pl + $meds + $rx;
+                    $cards = $allergy + $pl + $meds;
                     $col = "p-1 ";
 
-                    if ($cards > 0) {
```
bradymiller, line 1040: Guessing removing above comparison will  break for folks that do not have acl access to any of the above 3 cards (ie. receptionist) since I recently just added the above comparison to fix that bug :)

bradymiller: btw, the break was a dividing by zero error, which is why ensured $cards was  > 0

bradymiller: also, don't worry about fixing/testing this minor issue. plan to test and fix it, if needed, tonight :)

sjpadgett:
ihanks

PR #6892, opened 2023-09-29
Add combined Nationality and Country list.

### LINKED ISSUE #6891: Add a combined nationality with country list to Demographics Stats group
### Describe the problem

Title says it all. List has 249 items with search.
I don't remember if I put the list with search datatype in or not whatever the case very nice addition. Needed to change inline style from align center to normal. Even for display: Grid it is better than trying to center on list that are not multi select or grids.

sjpadgett:
update database and upgrade for new list and layout option in Stat.

Fixes #6891 

update database and upgrade for new list and layout option in Stat.


sjpadgett:
Found new issue with List select dialog and wouldn't be notice unless view in console or user debug is turned on!

image

stephenwaite:
related to #5667

stephenwaite:
I think this package, https://github.com/mledoze/countries, is a better solution than adding this custom csv list.

sjpadgett:
Same list and who is going to import the list after install? There is only a certain number of countries and one list is as good as another. What about maintaining legacy is an admin going to try and merge changes to existing lists.
The only diff in the project you like is more geo info etc is available.

sjpadgett:
@stephenwaite I'll look at how much csv changes I need to make toi import to list if you feel more comfortable with this project generated list.

stephenwaite:
ok, thanks @sjpadgett , was wondering if we don't have to save the csv list to the db but that probably presents other issues for the list stuff.

sjpadgett:
I've already justified adding list and for many non profit refugee centers and international alone is justification to have list permanent. I have tried to justify list in DB otherwise if I didn't feel comfortable with my integration I would not have taken on the project.
However if this is still uncomfortable or of concern to others I will pull the PR.

stephenwaite:
was just thinking the dropdown would hit that value from the package but this is probably better, thanks

sjpadgett:
I'd be worried about legacy i.e. if field option id values change etc. I get your worry about DB size though Stephen.

stephenwaite:
not worried about DB size just trying to think of an easy way to update, also to combine country and nationality seems odd but fits Ruth's requirement and if I remember Brady was fine with it.

stephenwaite:
great job getting all these pieces working, just tested it and it's working well!

sjpadgett:
My original plan was two lists but after testing the combined list it seemed easier for user.
Thanks for testing for me!

bradymiller:
dance-bear

PR #6897, opened 2023-09-30
feat: Support icons in main menu

### LINKED ISSUE #6896: feat: Menu Icons
### Describe the problem

The menu is hard to read being just words and so compact. it would be nice if it were broken up by icons that could help the visual

Add icons after the words or before the words to give a visual representation of the menu item for quick reference.
With our eyes, we can identify pictures faster than words.

juggernautsei:
Fixes #6896

Improves the menu UI to help visually see faster

Add icons to the top-level menu. 

image

### DIFF
--- a/interface/main/tabs/templates/menu_template.php
+++ b/interface/main/tabs/templates/menu_template.php
@@ -16,12 +16,12 @@
 
 <script type="text/html" id="menu-action">
     <i data-bind="css: icon,text:helperText" class="fa closeButton"></i>
-    <div class='menuLabel px-1' data-bind="text:label,click: menuActionClick,css: {menuDisabled: ! enabled()}"></div>
+    <div class='menuLabel pr-5' data-bind="text:label,click: menuActionClick,css: {menuDisabled: ! enabled()}"></div>
 </script>
 <script type="text/html" id="menu-header">
     <i data-bind="css: icon" class="fa closeButton"></i>
     <div class="menuSection dropdown">
-        <div class='menuLabel px-1 dropdown-toggle oe-dropdown-toggle' data-bind="text:label" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"></div>
+        <div class='menuLabel pr-5 dropdown-toggle oe-dropdown-toggle' data-bind="text:label" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"></div>
         <ul class="menuEntries dropdown-menu rounded-0 border-0" name="menuEntries" data-bind="foreach: children">
             <li data-bind="template: {name:header ? 'menu-header' : 'menu-action', data: $data }"></li>
         </ul>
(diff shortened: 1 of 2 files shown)


interface/main/tabs/templates/menu_template.php
```diff
@@ -16,12 +16,12 @@
 
 <script type="text/html" id="menu-action">
     <i data-bind="css: icon,text:helperText" class="fa closeButton"></i>
-    <div class='menuLabel px-1' data-bind="text:label,click: menuActionClick,css: {menuDisabled: ! enabled()}"></div>
+    <div class='menuLabel pr-md-3 pr-sm-1 pr-xl-4' data-bind="text:label,click: menuActionClick,css: {menuDisabled: ! enabled()}"></div>
```
bradymiller, line 19: How do the changes in this script look without the icons? (just thinking may be good to at least have code in here to make the icons looks good even if we do not include them in the official json menus (note it is very easy to create custom menus and assign them to users)

sjpadgett: We discussed on call that we have global to turn on icons only or default labels as currently styled.

juggernautsei: @bradymiller I will see what the menu looks like without the icons added. I will post the screenshot here.

juggernautsei:
This project is going in the right direction. I will create two more standard JSON files. One with icons only, and keep the original. Add a global to select which one to load.

bradymiller:
hi @juggernautsei , Would be much better to go with 1 json file (that had the icons) and then the global controls whether to show the icons in it (then we won't be managing 2 separate official menu jsons that are only different in icons since would be high risk of divergence over time).

juggernautsei:
> hi @juggernautsei , Would be much better to go with 1 json file (that had the icons) and then the global controls whether to show the icons in it (then we won't be managing 2 separate official menu jsons that are only different in icons since would be high risk of divergence over time).

Noted. I will work on that

juggernautsei:
Before changing line 19 and 24 in the menu_template.php
image

After:
image

Looks virtually the same to me.

robertdown:
Love this PR. But I think the spacing still needs some work and please find something else for the Miscellaneous item, it doesn't make sense to use Stackoverflow's logo. Also, the icon color will have to match the color of the link text.

Lastly, just need to make sure this works if the user is manipulating the menu items via code (Using the Menu Service)

robertdown:
Also, I think the icons should be left of related text? I Maybe that's what feels just a bit off

juggernautsei:
I am going to close this PR. Thanks everyone for their comments and suggestions 
@mdsupport @robertdown @sjpadgett @bradymiller

robertdown:
I think there minimum viable solution is not too far off. Mind if I take over the issue?

sjpadgett:
Aria

juggernautsei:
> I think there minimum viable solution is not too far off. Mind if I take over the issue?

No, I don't mind. Do I need to reopen the issue?

Issue #6908, opened 2023-10-03
bug: Clinical Note Type does not respect Active flag


robertdown:
Same as #6907 but for the Clinical Note Type field

![image](https://github.com/openemr/openemr/assets/1381170/7a9fe2be-58c8-4dc0-8743-8ac15aa7b207)

Issue #6907, opened 2023-10-03
bug: Clinical Note Category does not respect Active setting


robertdown:
### Describe the bug

The Clinical Note Category dropdown pulls the entire list, ignoring the Active flag

![image](https://github.com/openemr/openemr/assets/1381170/6dc88e9f-9dc3-4413-8ddd-88a3b94065d7)

### To Reproduce

1. Disable several Clinical Note Categories by unchecking the active flag in the List Manager
2. Open an encounter, create a Clinical Note
3. Category dropdown still renders entire list

### Expected behavior

Should only render active Categories

PR #6910, opened 2023-10-04
fix: Return only active clinical note types and categories by default

robertdown:
Resolves both #6907 and #6908 

This changes the default behavior of the two relevant functions in the ClinicalNotesService by assuming the request is for active list options only. Can pass `true` to the function (Setting `$includeInactive`) which will return all options

### DIFF
--- a/src/Services/ClinicalNotesService.php
+++ b/src/Services/ClinicalNotesService.php
@@ -344,17 +344,19 @@ class ClinicalNotesService extends BaseService
         return !empty($options);
     }
 
-    public function getClinicalNoteTypes()
+    public function getClinicalNoteTypes($includeInactive = false)
     {
         $listService = new ListService();
-        $options = $listService->getOptionsByListName('Clinical_Note_Type');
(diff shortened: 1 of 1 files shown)


adunsulag (approved):
Looked into this and everything looks fine with this change.

src/Services/ClinicalNotesService.php
```diff
@@ -344,17 +344,19 @@ public function isValidClinicalNoteCode($code)
         return !empty($options);
     }
 
-    public function getClinicalNoteTypes()
+    public function getClinicalNoteTypes($includeInactive = false)
```
sjpadgett, line 347: I guess this is on me however I fully expected the service to return only active items from the list. Really, why would you need inactive items ever!
Nice save robert, thanks.

sjpadgett (approved):
Can this come in @robertdown?

robertdown:
I'd love for it to come in. Just wanted a second set of eyes to look at it before I pushed the big green button

bradymiller:
2DV-2(5)

PR #6912, opened 2023-10-04
fix parsing of response from NewCrop

kchapple:
Fixes #

Something changed in PHP, or a package dependency that broke the parsing of allergy and medication responses from NewCrop.

This fixes that, and also prevents sending diags other than ICD10 to NewCrop because NewCrop only supports ICD10.

@bradymiller If possible, this should also be merged into rel-701 and added to next patch.

### DIFF
--- a/interface/eRxSOAP.php
+++ b/interface/eRxSOAP.php
@@ -31,8 +31,11 @@ class eRxSOAP
     private $soapSettings = array();
     private $siteId;
 
-    protected static function fixHtmlEntities($array, $xmltoarray) {
-        if (!is_array($array)) return;
+    protected static function fixHtmlEntities($array, $xmltoarray)
+    {
+        if (!is_array($array)) {
+            return;
+        }
 
         foreach ($array as $key => $value) {
             if (is_array($value)) {
--- a/interface/eRxSOAP.php
+++ b/interface/eRxSOAP.php
@@ -31,19 +31,11 @@ class eRxSOAP
     private $soapSettings = array();
     private $siteId;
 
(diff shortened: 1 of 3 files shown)


interface/eRxSOAP.php
```diff
@@ -31,6 +31,21 @@ class eRxSOAP
     private $soapSettings = array();
     private $siteId;
 
+    protected static function fixHtmlEntities($array, $xmltoarray)
```
bradymiller, line 34: Won't you need to pass $array by reference here?

bradymiller: also, fyi, shooting for 7.0.2 release in several weeks (rather than another 7.0.1 patch)

kchapple: Yes, thanks!

kchapple: Testing a JSON solution...

interface/eRxSOAP.php
```diff
@@ -31,19 +31,11 @@ class eRxSOAP
...
-        foreach ($array as $key => $value) {
-            if (is_array($value)) {
-                self::fixHtmlEntities($value, $xmltoarray);
-            } else {
-                $array[$key] = $xmltoarray->fix_html_entities($value);      //returns proper html values
-            }
-        }
+        $encoded = json_encode($array);
+        $fixed = $xmltoarray->fix_html_entities($encoded);
+        return json_decode($fixed, true);
```
stephenwaite, line 38: smaller_excited_elephant

bradymiller (approved):

bradymiller:
2DV-2(13)

PR #6916, opened 2023-10-08
Update mfa_totp.php

DrBassman:
On the original version, if you typed your password and pressed <ENTER>, a blank screen would be displayed.  This is because the <Form onsubmit=""> tag had one action, and the submit <Button> had a different onclick="" action.  This change makes it so pressing <ENTER> or clicking the Submit <Button> execute the same code...

Fixes #

On the create MFA screen, it would prompt you for your passworrd.   If you hit <ENTER> after typing your password, the screen would be empty.  If you clicked the "Submit" button with your mouse, the QR code would be displayed, which is the expected result.  This pull request FIXES that issue.

### DIFF
--- a/interface/usergroup/mfa_totp.php
+++ b/interface/usergroup/mfa_totp.php
@@ -96,7 +96,7 @@ $user_full_name = $user_name['fname'] . " " . $user_name['lname'];
         <?php
     } ?>    <div class="row">
                 <div class="col-sm-12">
-                    <form method='post' class="form-horizontal" action='mfa_totp.php' onsubmit='return top.restoreSession()'>
+                    <form method='post' class="form-horizontal" action='mfa_totp.php' onsubmit="doregister('reg2')">
                         <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken()); ?>" />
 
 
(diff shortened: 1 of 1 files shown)


interface/usergroup/mfa_totp.php
```diff
@@ -126,7 +126,7 @@ function docancel() {
                                 </fieldset>
                                 <div class="form-group clearfix">
                                 <div class="col-sm-12 text-left position-override">
-                                        <button type="button" class="btn btn-secondary btn-save" value="<?php echo xla('Submit'); ?>" onclick="doregister('reg2')"><?php echo xlt('Submit'); ?></button>
+                                        <button type="submit" class="btn btn-secondary btn-save" value="<?php echo xla('Submit'); ?>"><?php echo xlt('Submit'); ?></button>
```
mdsupport, line 129: `<?php echo xla('Submit'); ?>`

Unrelated to this PR - wonder why original code translates submit value? It's like a variable content to be used internally. It can also be set to 'reg2' eliminating need to hardcode function parameter.

DrBassman: I was trying to remove the redundant call to `doregister('reg2')`.  Any future update could change it in the `<form>` tag and not in this `<button>` tag.

interface/usergroup/mfa_totp.php
```diff
@@ -96,7 +96,7 @@ function docancel() {
         <?php
     } ?>    <div class="row">
                 <div class="col-sm-12">
-                    <form method='post' class="form-horizontal" action='mfa_totp.php' onsubmit='return top.restoreSession()'>
+                    <form method='post' class="form-horizontal" action='mfa_totp.php' onsubmit="doregister('reg2')">
```
stephenwaite, line 99: jumping

adunsulag:
Looks good to me.  Was concerned that we removed the top.restoreSession() but doregister() already makes the call, so no concern there. Bringing it in.

interface/usergroup/mfa_totp.php
sjpadgett: Sorry late. @mdsupport You're correct concerning translating post variables as confusing because if you're acting on the variable, you must use the translated value. This pattern occurs throughout legacy code where I think we just got overzealous at the time we introduced translations.

bradymiller:
dance-bear

PR #6922, opened 2023-10-10
feat: Twigified SMART Style url for module writers - Fixes #6919

### LINKED ISSUE #6919: Feat: Allow module writers to override / hook into the SMART style_url
SMART apps have the option of mimicking some of the basic styling elements of the OpenEMR configuration.  Currently we only support a fixed static url with hard coded styles from the light theme for the SMART style url.  We should allow module writers to override this value for module themes / etc.

adunsulag:
Fixes #6919
Moves location of the SMART URL to be part of the OAuth2 url scheme which lets us customize it and use it in twig.

Fires off a TemplateRender event now that allows module writers to override the template if they want to.

Changed up the php documentation in the Capability statement with some minor url fixes.

Removed unused code where we had leftover SMART style code.


adunsulag:
I'll need to do a round of Inferno testing here.  Shouldn't change anything as the interface stays the same but we'll want to do that prior to the release going out anyways.

bradymiller (approved):

bradymiller:
tenor99

PR #6938, opened 2023-10-16
bug fix

### LINKED ISSUE #6937: Bug fix
### To Reproduce

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

juggernautsei:
Fixes #


interface/orders/orders_results.php
```diff
@@ -142,7 +142,7 @@ function QuotedOrNull($fld)
                 "`range` = '" . oresData("form_result_range", $lino) . "', " .
                 "units = '" . oresData("form_result_units", $lino) . "', " .
                 "facility = '" . oresData("form_facility", $lino) . "', " .
-                "comments = '" . $form_comments . "', " .
+                "comments = '" . htmlspecialchars($form_comments) . "', " .
```
bradymiller, line 145: hi @juggernautsei ,
Shouldn't store html escaped elements in database and should instead escape it when it is outputted to html/xml. Otherwise risk cases of not escaping, incorrect escaping context, and overescaping.

PR #6941, opened 2023-10-18
Quest Lab Hub module support

### LINKED ISSUE #6940: It would be nice to have a module to connect Quest
### Describe the problem

Quest no longer supports SMTP communication. They have switched to an API system to transmit and receive lab orders.

Having a module that manages the communication between Quest Lab Hub and OpenEMR would be nice.
The module would connect with the lab hub, send lab orders, and automatically create requisition forms. The module would set a background service to pull in lab results once a day.  The module will be able to print labels also.

juggernautsei:
Fixes #6940

### DIFF
--- a/interface/forms/procedure_order/common.php
+++ b/interface/forms/procedure_order/common.php
@@ -26,10 +26,16 @@ require_once(__DIR__ . "/../../../custom/code_types.inc.php");
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Forms\ReasonStatusCodes;
 use OpenEMR\Core\Header;
+use OpenEMR\Events\Services\LabTransmitEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
 
 if (!$encounter) { // comes from globals.php
     die("Internal error: we do not seem to be in an encounter!");
 }
+/**
+ * @var EventDispatcher
+ */
+$ed = $GLOBALS['kernel']->getEventDispatcher();
 
 // Defaults for new orders.
 $provider_id = getProviderIdOfEncounter($encounter);
@@ -408,6 +414,9 @@ if (($_POST['bn_save'] ?? null) || !empty($_POST['bn_xmit']) || !empty($_POST['b
                 require_once(__DIR__ . "/../../procedure_tools/labcorp/ereq_form.php");
                 require_once(__DIR__ . "/../../procedure_tools/labcorp/gen_hl7_order.inc.php");
                 $alertmsg = gen_hl7_order($formid, $hl7, $reqStr);
+            } elseif ($gbl_lab === 'quest') {
+                require_once(__DIR__ . "/../../procedure_tools/quest/gen_hl7_order.inc.php");
+                $alertmsg = gen_hl7_order($formid, $hl7, $reqStr);
             } else { // Default lab. Add more labs here.
                 require_once(__DIR__ . "/../../orders/gen_hl7_order.inc.php");
                 $alertmsg = gen_hl7_order($formid, $hl7);
@@ -427,6 +436,12 @@ if (($_POST['bn_save'] ?? null) || !empty($_POST['bn_xmit']) || !empty($_POST['b
                         xlt("Order Successfully Sent") . "...\n" .
                         xlt("Order HL7 Content") .
                         ":\n" . $hl7 . "\n";
+                    if ($gbl_lab === 'quest') {
+                        $order_log .= xlt("Transmitting order to Quest");
+                        $ed->dispatch(new LabTransmitEvent($hl7), LabTransmitEvent::EVENT_LAB_TRANSMIT, 10);
+                        $ed->dispatch(new LabTransmitEvent($pid), LabTransmitEvent::EVENT_LAB_POST_ORDER_LOAD, 10);
+                    }
+
                     if ($_POST['form_order_psc']) {
                         if ($gbl_lab === 'labcorp') {
                             $order_log .= "\n" . date('Y-m-d H:i') . " " .
@@ -456,6 +471,7 @@ if (($_POST['bn_save'] ?? null) || !empty($_POST['bn_xmit']) || !empty($_POST['b
                 file_put_contents($log_file, $order_log);
             }
         }
+
         unset($_POST['bn_xmit']);
     }
     unset($_POST['bn_save']);
@@ -967,11 +983,13 @@ if (!empty($row['lab_id'])) {
             let acctid = <?php echo js_escape($gbl_client_acct); ?>;
             let order = f.id.value;
             let patient = <?php echo js_escape($patient['lname'] . ', ' . $patient['fname'] . ' ' . $patient['mname']); ?>;
+            let dob = <?php echo js_escape($patient['DOB']); ?>;
             let pid = <?php echo js_escape($patient['pid']);  ?>;
             let url = top.webroot_url + "/interface/procedure_tools/libs/labs_ajax.php";
             // this escapes above
             let uri = "?action=print_labels&count=" + encodeURIComponent(count) + "&order=" + encodeURIComponent(order) + "&pid=" + encodeURIComponent(pid) +
                 "&acctid=" + encodeURIComponent(acctid) + "&patient=" + encodeURIComponent(patient) + "&specimen=" + encodeURIComponent(tarray) +
+                "&dob=" + encodeURIComponent(dob) +
                 "&csrf_token_form=" + <?php echo js_url(CsrfUtils::collectCsrfToken()); ?>;
 
             // retrieve the labels
--- a/interface/forms/procedure_order/common.php
+++ b/interface/forms/procedure_order/common.php
@@ -26,7 +26,7 @@ require_once(__DIR__ . "/../../../custom/code_types.inc.php");
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Forms\ReasonStatusCodes;
 use OpenEMR\Core\Header;
-use OpenEMR\Events\Services\LabTransmitEvent;
+use OpenEMR\Events\Services\QuestLabTransmitEvent;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 
 if (!$encounter) { // comes from globals.php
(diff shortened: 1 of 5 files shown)


interface/forms/procedure_order/common.php
```diff
@@ -26,10 +26,16 @@
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Forms\ReasonStatusCodes;
 use OpenEMR\Core\Header;
+use OpenEMR\Events\Services\LabTransmitEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
```
bradymiller, line 30: don't see why line 30 above is needed. Am I missing something?

interface/procedure_tools/quest/gen_hl7_order.inc.php
```diff
@@ -0,0 +1,590 @@
...
+        exit;
+    } else if ($protocol == 'SFTP') {
+        // Compute the target path/file name.
+        $filename = $msgid . '.txt';
+        if ($pprow['orders_path']) {
+            $filename = $pprow['orders_path'] . '/' . $filename;
+        }
+
+        // Connect to the server and write the file.
+        $sftp = new \phpseclib\Net\SFTP($remote_host);
```
bradymiller, line 555: note most recent codebase is now using phpseclib3 instead of phpseclib (just need to add the 3 above and it should work fine)

interface/forms/procedure_order/common.php
adunsulag: Because its referenced in line 36 in the php annotation this line should remain.  The dispatch will still work but php intellisense will not if the line is missing.  Another alternative is to put the FQDN on line 36, but I'd rather it be here on line 30.

interface/forms/procedure_order/common.php
```diff
@@ -427,6 +436,12 @@ function getListOptions($list_id, $fieldnames = array('option_id', 'title', 'seq
                         xlt("Order Successfully Sent") . "...\n" .
                         xlt("Order HL7 Content") .
                         ":\n" . $hl7 . "\n";
+                    if ($gbl_lab === 'quest') {
+                        $order_log .= xlt("Transmitting order to Quest");
+                        $ed->dispatch(new LabTransmitEvent($hl7), LabTransmitEvent::EVENT_LAB_TRANSMIT, 10);
```
adunsulag, line 441: Where is LabTransmitEvent defined? I don't see it in this commit and I can't find it in the code system.

interface/procedure_tools/quest/gen_hl7_order.inc.php
```diff
@@ -0,0 +1,590 @@
...
+* ORDER BY i.name, i.id;
+*
+* Then export as a CSV file and read it into your favorite spreadsheet app.
+*/
+
+require_once("$webserver_root/custom/code_types.inc.php");
+
+use OpenEMR\Common\Logging\EventAuditLogger;
+
+function hl7Text($s)
```
adunsulag, line 42: Looks like a lot of these functions are just copied across.  I don't think its required you do this for your PR but it'd be great if we could pull these common functions shared across all the gen_hl7_order* files into a shared library.  Maybe open an issue on github and mark it as a newcomer issue for someone to do.

adunsulag (commented):
@juggernautsei Only major problem I see is that LabTransmitEvent appears to have not been checked in.  Everything else looks ok. Is there a test environment that people can use to test this code out? I looked through the gen_hl7_order_inc.php but I don't see anyway to test if its working correctly w/o a test bed environment so I have to assume its just working correctly.

Once the LabTransmitEvent is checked in and reviewed we can probably go ahead and bring this in.  We'll want to merge it into the release branch once this is in.

interface/forms/procedure_order/common.php
juggernautsei: @adunsulag It is in the preocedure_order/common.php line 441 and 442

interface/procedure_tools/quest/gen_hl7_order.inc.php
juggernautsei: @adunsulag Actually what I did was copy the file from labcorp and modified it to work.

src/Events/Services/LabTransmitEvent.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/*
+ *   package   OpenEMR
+ *   link      http://www.open-emr.org
+ *  author    Sherwin Gaddis <sherwingaddis@gmail.com>
+ *  Copyright (c)
+ *  All rights reserved
```
adunsulag, line 8: Should be GPL license.

src/Events/Services/LabTransmitEvent.php
```diff
@@ -0,0 +1,41 @@
...
+class LabTransmitEvent extends Event
+{
+    /**
+     * This event is triggered when a lab is to be transmitted
+     */
+    const EVENT_LAB_TRANSMIT = 'lab.transmit';
+
+    /**
+     * This event is triggered when a lab requisition form is returned from Quest
+     * Requisition form has to be enabled in the globals
```
adunsulag, line 25: If we think this event will only apply to Quest, we should rename this class to be QuestLabTransmitEvent.

adunsulag: If its a generic event, probably need to change up the comment and just mention quest as being an example of where this is triggered.

src/Events/Services/LabTransmitEvent.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    const EVENT_LAB_POST_ORDER_LOAD = 'lab.post_order_load';
+
+    private string $order;
+    public function __construct($hl7)
```
adunsulag, line 30: If this is going to be a generic event do you think that every lab transmission will use an HL7 string as the order transmission?  Just wondering if we need to consider options where more data needs to be sent as part of the lab transmit event.  If it will always be a string then this is fine.

src/Events/Services/LabTransmitEvent.php
sjpadgett: Quest provides a requisition form via API call whereas Labcorp used to have to be built by EHR. We'd want this event to be generic and able to be triggered for a Labcorp eReq(or any other) so please change now so we won't have to move it later.
How are you handling user request for results. Is there an event already in place for electronic report?

interface/procedure_tools/quest/gen_hl7_order.inc.php
```diff
@@ -0,0 +1,590 @@
+<?php
+
+/**
+* Functions to support HL7 order generation.
+*
```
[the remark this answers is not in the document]
juggernautsei: @sjpadgett I have no way to test lab Corp or quest. @adunsulag the sandbox for Quest is only for developers. They don't have a test mode. So the test mode I have in there will not work without a developer account.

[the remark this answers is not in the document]
sjpadgett: > or quest

If you haven't tested with Quest with this hl7 orders script then why are you putting in here?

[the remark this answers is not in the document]
juggernautsei: @sjpadgett you misunderstood. The code has been tested.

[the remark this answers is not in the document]
adunsulag: k @juggernautsei if we can address the LabTransmitEvent review items, I'm fine with bringing this in.  It looks like there isn't a sandbox mode unless we have a developer account with Quest.  Since you've verified this is working on your developer account I'm fine with bringing it in once you address the LabTransmitEvent.

juggernautsei:
@adunsulag https://github.com/openemr/openemr/pull/6941/commits/dc35d47e55d6caae5ac90551b01f867a2086f53a
LabTransmitEvent was uploaded yesterday. Dr. Alex Berlin is doing production testing now. Once we have a satisfactory run. Quest will approve Go Live.

adunsulag:
@juggernautsei I noted two items that need to be addressed in the LabTransmitEvent.  One for sure is the code licensing in the documentation, the other is clarification on the input parameter.  That is what I meant by taking care of those review items.

interface/procedure_tools/quest/gen_hl7_order.inc.php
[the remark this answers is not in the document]
sjpadgett: If this stays as a singular use event and while I'd prefer this to be a generic event that fires when any  lab order is sent, at least rename the class to reflect it's specific to quest. Otherwise any dev coming behind this PR will think behavior is for any lab send! The naming is deceptive.

src/Events/Services/LabTransmitEvent.php
juggernautsei: @adunsulag all of my dealings with labs. They all have used HL7 string. I believe until a new standard is developed on high. It will always be a string. IMO

src/Events/Services/LabTransmitEvent.php
juggernautsei: When I was building this feature, all of you are playing like a record in my head to make this flexible and not to hardcode things into place. Use the dispatch system to couple the system loosely. That is what I was trying to do. I discovered that Labcorp requisition is built into the system. It is buried in there. I found it during my build process. So, I used the dispatch to send the lab order over to the module to send out of the API connection to Quest. 
All the module does is handle the communication between OpenEMR and Quest. All other processes are the same. I did not add anything to the codebase. I did not have to. I knew from all my years of working with the codebase what it was capable of doing. I did not reinvent the wheel. I was standing on the shoulders of Jerry and Rob. The only thing I needed to do was to get the order out to Quest and retrieve results via an API call.

juggernautsei:
@adunsulag I have addressed the input parameters above. Here is the license change.

src/Events/Services/LabTransmitEvent.php
sjpadgett: And I can appreciate your efforts to follow those guidelines Sherwin, I do! But this is a singular event and you missed my whole naming convention concern. This event should have been put on the button click event and qualified in the handler by passing the lab id! Then you are following the flavor of useful events that the maintainers have to be mindful of

sjpadgett: @adunsulag took charge of this review so I leave it to him. Be mindful though Sherwin there is a good possibility this event will get modified in the future!

interface/procedure_tools/quest/gen_hl7_order.inc.php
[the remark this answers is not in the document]
adunsulag: New class name I believe resolves the issue.

src/Events/Services/LabTransmitEvent.php
adunsulag: Looks like the button click event and generic lab id thing could be handled in another PR.  With the  new class name I believe this resolves the issue.

src/Events/Services/LabTransmitEvent.php
adunsulag: Sounds good.

src/Events/Services/LabTransmitEvent.php
adunsulag: Looks like this is now fixed.

adunsulag:
Once CI engine finishes, should be able to bring this in.  Thanks @juggernautsei.

interface/procedure_tools/quest/gen_hl7_order.inc.php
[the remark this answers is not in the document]
juggernautsei: I don't understand your thinking that someone would think that all labs used this LabTransmitEvent. Yes, if you just looked only at the name and did nothing else. It could imply that all labs are transported into the ether through this magical conduit. But you can say that about a lot of the naming conventions in the system. That is why a developer needs to investigate. But as you said, some skills are hard to teach. This is not the battle of Little Bighorn. So, I changed the name.

sjpadgett:
Thank you Sherwin for listening to my concerns.

bradymiller:

PR #6953, opened 2023-10-24
Bug Fix - default pos code

juggernautsei:
Fixes 6952

Default POS code not being saved to the encounter.

### DIFF
--- a/interface/forms/newpatient/save.php
+++ b/interface/forms/newpatient/save.php
@@ -40,7 +40,7 @@ if ($_POST['mode'] == 'new' && ($GLOBALS['enc_service_date'] == 'hide_both' || $
 } else {
     $date = isset($_POST['form_date']) ? DateTimeToYYYYMMDDHHMMSS($_POST['form_date']) : null;
 }
-$defaultPosCode = $encounterService->fetchPosCode($_POST['facility_id']);
+$defaultPosCode = $encounterService->getPosCode($_POST['facility_id']);
 $onset_date = isset($_POST['form_onset_date']) ? DateTimeToYYYYMMDDHHMMSS($_POST['form_onset_date']) : null;
 $sensitivity = $_POST['form_sensitivity'] ?? null;
 $pc_catid = $_POST['pc_catid'] ?? null;
--- a/interface/forms/newpatient/save.php
+++ b/interface/forms/newpatient/save.php
@@ -40,7 +40,7 @@ if ($_POST['mode'] == 'new' && ($GLOBALS['enc_service_date'] == 'hide_both' || $
 } else {
     $date = isset($_POST['form_date']) ? DateTimeToYYYYMMDDHHMMSS($_POST['form_date']) : null;
 }
-$defaultPosCode = fetchPosCode($_POST['facility_id']);
+$defaultPosCode = $encounterService->fetchPosCode($_POST['facility_id']);
 $onset_date = isset($_POST['form_onset_date']) ? DateTimeToYYYYMMDDHHMMSS($_POST['form_onset_date']) : null;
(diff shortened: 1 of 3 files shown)


interface/forms/newpatient/save.php
```diff
@@ -50,7 +50,7 @@
 $mode = $_POST['mode'] ?? null;
 $referral_source = $_POST['form_referral_source'] ?? null;
 $class_code = $_POST['class_code'] ?? '';
-$pos_code = $_POST['pos_code'] ?? null;
+$pos_code = (empty($_POST['pos_code'])) ? $defaultPosCode : $_POST['pos_code'];
```
stephenwaite, line 53: smaller_excited_elephant

library/encounter.inc.php
```diff
@@ -32,3 +32,10 @@ function fetchDateService($encounter)
     $result = explode(" ", $result['date']);
     return $result[0];
 }
+
+function fetchPosCode($facility_id)
+{
```
stephenwaite, line 37: maybe putting this function in the encounter service will be better long term since we've been moving away from the old libraries

juggernautsei: When does the engineering OpenEMR meeting occur? I would like to sit in on them when I can.

src/Services/EncounterService.php
```diff
@@ -725,4 +725,17 @@ public function getEncountersByDateRange($startDate, $endDate)
...
     }
+
+    /**
+     * Returns the default POS code that is set in the facility table.
+     *
+     * @param $facility_id
+     * @return mixed
+     */
+    public function fetchPosCode($facility_id)
+    {
```
stephenwaite, line 736: how about `getPosCode($facility_id)` for consistency :) ?
soccer

library/encounter.inc.php
stephenwaite: sat afternoons 2 pm eastern :laughing:

src/Services/EncounterService.php
juggernautsei: No problem

stephenwaite: southpark_cheering

stephenwaite (approved):

sjpadgett:
thinking about this, this solution is the wrong place to set this because the user doesn't have a say in the value. What if they don't want a POS? Should be set in user interface should it not? Is POS required in all claims? If default is set does that mean it should appear in all claims? Enquiring minds want to know!:)

juggernautsei:
@sjpadgett The behavior that this is solving is when the POS is not set in the encounter (hidden). Nothing is being saved. Thus, on the billing, it will have 00 as the place of service and not the default that was set for the facility. Users are expecting the default POS that is in the facilities setting to be used. But when the POS is hidden in the encounter. There is no way to set it. 
Then, there is a lot of work correcting the POS.

sjpadgett:
hmm why does the POS get hidden. Is it because the global isn't turned on?
If I remember correctly POS is tricky when it comes to claim and specific billing codes.
I don't oppose the PR but I had an inkling there is something special concerning POS and billing.
I think it came up concerning home health visits.

juggernautsei:
Yes, there is a setting in the config to set POS in the encounter. If this is unchecked, the POS option is removed from the encounter.  This should be checked if the POS needs to be set per appointment/encounter type. But when not needed. This data should come from the facilities table.

sjpadgett:
Good enough for me Sherwin, thanks!

sjpadgett (approved):
thanks for taking the time to fix this bug. I wish more vendors followed your example Sherwin!

bradymiller (approved):

bradymiller:
2DV(5)

PR #6985, opened 2023-11-06
EHI Export Bug Fixes, track anything implementation.

adunsulag:
Realized I had forgotten the track anything implementation and needed to add a custom table definition for that.

Testing the track anything custom form revealed a bug in the clinical notes custom table definition which is now fixed.


bradymiller:
tenor99

PR #6992, opened 2023-11-07
Layout editor list dialog error fix

### LINKED ISSUE #6984: error message on changing table in demographic forms
### Describe the bug
admin/forms/layouts - demographics. Change a list. An error message is displayed

concise description of what the bug is:
if one clicks on the list field, to change the list to be used, an error message is displayed in a pop up window in the 'select list' window. It varies from system to system, examples are 

 `DataTables warning: table id=my_data_table - Requested unknown parameter '2' for row 0, column 2. For more information about this error, please see http://datatables.net/tn/4`
and
`undefined
URL: undefined
Line: undefined Column: undefined
Error object: undefined`

If one clicks 'ok' to each error message then one is finally able to change the table

### To Reproduce

image

1. choose menu admin/forms/layouts 
2.  use drop down to choose 'demographics'. 
3. click on a 'list' field to change the list
5. click 'ok' until errors stop displaying
6. procede to change the list as desired

### Expected behavior

one should be able to change the list without error messages displaying first

### Client configuration

- Browser: firefox
- OpenEMR version: 7.0.1(1) and  7.0.2-dev
- Operating system: linux 

possibly other OS and browser, i can find out if necessary

sjpadgett:
- add modifier to lists search returned array

Fixes #6984 

- see issue
Apparently datatables project is now enforcing all columns from table build has a data object.
In the past we default to the code type header for lists (three columns) however failed to populate the modifier column on search return results.

### DIFF
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -96,9 +96,9 @@ if (count($emr_app)) {
 
         $div_app = sprintf(
             '
-            <div id="divApp" class="form-group">
-                <label for="appChoice" class="text-right">%s:</label>
-                <div>
+            <div id="divApp" class="form-group row">
+                <label for="appChoice" class="col-form-label col-sm-4">%s:</label>
+                <div class="col">
                     <select class="form-control" id="selApp" name="appChoice" size="1">%s</select>
                 </div>
             </div>',
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -83,11 +83,13 @@ if (count($emr_app)) {
     if (isset($_REQUEST['app']) && $emr_app[$_REQUEST['app']]) {
         $div_app = sprintf('<input type="hidden" name="appChoice" value="%s">', attr($_REQUEST['app']));
     } else {
+        $opt_htm = '';
+        // Is $opt_default suppose to be defined somewhere?
         foreach ($emr_app as $opt_disp => $opt_value) {
             $opt_htm .= sprintf(
                 '<option value="%s" %s>%s</option>\n',
                 attr($opt_disp),
-                ($opt_disp == $opt_default ? 'selected="selected"' : ''),
+                ($opt_disp == ($opt_default ?? '') ? 'selected="selected"' : ''),
                 text(xl_list_label($opt_disp))
             );
         }
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -55,6 +55,7 @@ $layout = $GLOBALS['login_page_layout'];
 // If this script is called with app parameter, validate it without showing other apps.
 //
 // Build a list of valid entries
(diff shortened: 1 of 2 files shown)


sjpadgett:
This one had me scratching my head for a minute. The reporting error is somewhat cryptic!

interface/login/login.php
```diff
@@ -83,11 +83,13 @@
...
         $div_app = sprintf('<input type="hidden" name="appChoice" value="%s">', attr($_REQUEST['app']));
     } else {
+        $opt_htm = '';
+        // Is $opt_default suppose to be defined somewhere?
         foreach ($emr_app as $opt_disp => $opt_value) {
             $opt_htm .= sprintf(
                 '<option value="%s" %s>%s</option>\n',
                 attr($opt_disp),
-                ($opt_disp == $opt_default ? 'selected="selected"' : ''),
+                ($opt_disp == ($opt_default ?? '') ? 'selected="selected"' : ''),
```
bradymiller, line 92: `$opt_default` should be `$emr_app_def`
(per above code)

interface/login/login.php
```diff
@@ -83,11 +83,13 @@
...
                 text(xl_list_label($opt_disp))
```
bradymiller, line 93: also above line 91 should use `$opt_value` and not `$opt_disp`
(per code above)

bradymiller: clarify (for above line):
```
text(xl_list_label($opt_value))
```

sjpadgett:
Thanks @bradymiller you made take a closer look at this! lol well maybe thanks as this feature wasn't included to be rendered in the login twig template!
For those curious this is a feature that @mdsupport put in way back that allows a menu to support redirecting login to a specific page as an different application login.

sjpadgett:
I cleaned this up to now work if somebody wants to invest the time to add to twig. This hasn't worked since v6.1.0 with nobody complaining it is missing! I think since we have modules now this feature is better served using an event so IMO this feature can go.

bradymiller:

mdsupport:
> Thanks @bradymiller you made take a closer look at this! lol well maybe thanks as this feature wasn't included to be rendered in the login twig template! For those curious this is a feature that @mdsupport put in way back that allows a menu to support redirecting login to a specific page as an different application login.



> I cleaned this up to now work if somebody wants to invest the time to add to twig. This hasn't worked since v6.1.0 with nobody complaining it is missing! I think since we have modules now this feature is better served using an event so IMO this feature can go.

@sjpadgett, in case you want to complete the cleanup you can leverage PR #4821 we submitted sometime ago. Apart from the module approach, our [mdpub repo](https://github.com/mdsupport/mdpub) provides less intrusive approach.

PR #6996, opened 2023-11-08
fix bug in pnVarCleanFromInput call

ishme-al:
small bug in call to pnVarCleanFromInput

fix the error


kojiromike:
This fix was already applied to master in #8240. Thanks for catching it, @ishmeals\!

PR #7004, opened 2023-11-09
Fixes #7001 Bulk FHIR 500 error patient ids

### LINKED ISSUE #7001: bug: Bulk FHIR fails w/ 500 error in patient/system export due to missing patient ids exception
In the patient and system $export operations there are no patient ids.  A refactor of the Provenance endpoint caused the operation to fail when it requests the list of patients to filter provenance records on.  This should only be done in a group export setting.  Our automated testing didn't check this as the default setting is to hit the group $export.

adunsulag:
Provenance was relying on patient ids for patient/$export and system/$export operation.  This should resolve that issue.

Fixes #7001


bradymiller (approved):

bradymiller:

PR #7005, opened 2023-11-09
Fixes #7002 FHIR api warnings missing keys

### LINKED ISSUE #7002: bug: FHIR api warnings for missing array keys in AllergyIntolerance, Vitals Observations, Social History Observations, Procedures
Going through the testing noticed there are a number of php warnings in the logs due to missing array keys in AllergyIntolerance, Vitals Observations, Social History Observations, Procedures.

With AllergyIntolerance if someone chooses the unassigned property there are no associated codes and that triggers a warning.

Vitals observation has a warning on the temp_method if it is empty.

Social History observations doesn't actually have an encounter reference, nor a user_uuid, so those trigger warnings.

Procedures was not properly populating the result document_id if one exists, same with a provider's name.

adunsulag:
Allergy intollerance was failing when the unassigned reaction was given since there are no codes and that threw a warning.

social history was relying on columns that didn't exist that also threw warnings.

Vitals services temp_method was not being set properly and threw a warning.

Procedure service had a number of columns it attempted to populate but the actual data values did not match with the array keys.

Fixes #7002.

### DIFF
--- a/src/Services/FHIR/FhirAllergyIntoleranceService.php
+++ b/src/Services/FHIR/FhirAllergyIntoleranceService.php
@@ -171,7 +171,8 @@ class FhirAllergyIntoleranceService extends FhirServiceBase implements IResource
 
         // cardinality is 0..*
         // however in OpenEMR we currently only track a single reaction, we will populate it if we have it.
-        if (!empty($dataRecord['reaction'])) {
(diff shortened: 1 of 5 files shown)


bradymiller (approved):

bradymiller:
dance-bear

bradymiller:

PR #7017, opened 2023-11-13
Persist summary screen container state

### LINKED ISSUE #7018: Summary screen user setting not presisting
### Describe the bug
Simply when user selects expand or compress container width setting is not persisted.

### To Reproduce

1. Add a '....'

### Client configuration

- Browser: all
- OpenEMR version: 6.10 on
- Operating system: windows or linux

sjpadgett:
- Just needed a value for utility

Fixes #7018


interface/patient_file/summary/demographics.php
```diff
@@ -335,7 +335,7 @@ function image_widget($doc_id, $doc_catg)
     'heading_title' => xl('Medical Record Dashboard'),
     'include_patient_name' => true,
     'expandable' => true,
-    'expandable_files' => array(), //all file names need suffix _xpd
+    'expandable_files' => array('demographics_xpd'), //all file names need suffix _xpd
```
stephenwaite, line 338: PHGrLWau

sjpadgett:
My goodness! If you turn features on we have warnings galore. I could spend a week and never catch up. Since we know what causes a majority of these warnings I'd bet we could automate a core run through

sjpadgett:
Testing well. Merging so others that may be testing won't look at what this PR resolves.

I'm really happy I spent the time on this! Summary is much nicer which I really like the layout. I think it might catch our legacy users a little by surprise but I think it will be accepted as a good change mostly. Nice work Robert and after going through new twig templates I have a much better understanding of where it is heading.

bradymiller:
Going to bring this into rel-702 to expedite

dance-bear

sjpadgett:
sorry forgot!

PR #7020, opened 2023-11-13
Fixes #7019 operation name export operation

### LINKED ISSUE #7019: bug: Bulk FHIR Export Capability Statement Operation Name incorrect
Inferno is giving a warning that the operation name should be the OperationDefinition code property rather than the OperationDefinition's name property in the HL7 specification.  Investigated this on the chat.fhir.org server where they say this is the specification, can't find it documented anywhere but need to change this since ONC Inferno is now testing against it.

adunsulag:
Fixes the export operation name to be 'export' which is the expected name coming from the OperationDefinition.code property for the $export operation for Patient/Group/System export.

Also added a few more debug logs to the json key stuff to help with debugging.
Fixes #7019.

### DIFF
--- a/src/Common/Auth/OpenIDConnect/JWT/JsonWebKeySet.php
+++ b/src/Common/Auth/OpenIDConnect/JWT/JsonWebKeySet.php
@@ -80,7 +80,7 @@ class JsonWebKeySet implements Key
      */
     public function getJSONWebKey($kid, $alg)
     {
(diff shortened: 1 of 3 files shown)


bradymiller (approved):

bradymiller:
Looks good and bringing in to master and rel-702 (expediting since possibly may be getting release out this week)

bradymiller:
2DV

PR #7023, opened 2023-11-14
fix: login language selector settings for allowed languages and hiding dummy language

bradymiller:
fix: login language selector settings for allowed languages and hiding dummy language

### DIFF
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -142,7 +142,17 @@ function getLanguagesList(): array
     $langList = [];
 
     while ($row = sqlFetchArray($res)) {
-        $langList[] = $row;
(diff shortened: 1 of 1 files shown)


interface/login/login.php
```diff
@@ -142,7 +142,17 @@ function getLanguagesList(): array
     $langList = [];
 
     while ($row = sqlFetchArray($res)) {
-        $langList[] = $row;
+        if (!$GLOBALS['allow_debug_language'] && $row['lang_description'] == 'dummy') {
+            continue; // skip the dummy language
+        }
```
stephenwaite, line 147: PHGrLWau

bradymiller:
testing well and bringing in

PR #7039, opened 2023-11-20
Fatal Error Unsupported operand types: string - string

juggernautsei:
Fixes 7038

Math error caught by the log file.


juggernautsei:
After time passes and revisit see this is the value that is in the array and not the array itself.

PR #7066, opened 2023-12-01
FHIR Appointment/Patient/Encounter/ValueSet

### LINKED ISSUE #7061: bug: FHIR Appointment fields wrong or missing
FHIR Appointment uses current time for meta.lastUpdated instead of an actual modified time.  The "Location" participant actor is missing. The appointmentType is missing.

### LINKED ISSUE #7062: feat: FHIR ValueSet is missing
The FHIR ValueSet endpoint is missing. If implemented, it could be used to implement the "coding" for appointmentType in #7061. It could also implement the options for all other pick lists.

### LINKED ISSUE #7063: feat: FHIR Patient missing deceasedBoolean
The FHIR Patient response is missing the deceased flag.

### LINKED ISSUE #7064: bug: FHIR Encounter lastUpdated field is wrong and missing as search param
The FHIR Encounter response returns lastUpdated as "now". It should have a real last modified date and be searchable.

robertjustjones:
Fixes #7061 
Fixes #7062 
Fixes #7063 
Fixes #7064 

This fixes and add fields in the listed FHIR resources. For Appointment.appointmentType, which is an instance specific and expandable set of categories, ValueSet was implemented to expose that list of codes in a FHIR way.

The ValueSet resource is added for the above. It encodes the appointment-type from openemr_postcalendar_categories.  While I was at it, I implemented all of the other list_id's from option_lists.


robertjustjones:
Dropped the client_credentials commit that failed testing.

swagger/openemr-api.yaml
```diff
@@ -6586,6 +6586,41 @@ paths:
...
+                  type: collection
+                  total: 0
+                  link: [{ relation: self, url: 'https://localhost:9300/apis/default/fhir/ValueSet' }]
+        '400':
+          $ref: '#/components/responses/badrequest'
+        '401':
+          $ref: '#/components/responses/unauthorized'
+      security:
+        -
+          openemr_auth: []
```
bradymiller, line 6623: looks like this is missing the /fhir/ValueSet/{uuid} endpoint
Note this script is built (we do not manually edit it) via a script. Let me know if need guidance on doing that.

bradymiller:
hi @robertjustjones , Noted some minor issues above. Otherwise code looks good to me. Will request OpenEMR's fhir guru @adunsulag to review this.

sjpadgett:
I updated the description to put issue linkage on separate lines. I seem to remember issues won't be closed if links are in series.

I'm impressed @robertjustjones you caught on so well to our API structure and are a welcomed addition to advancing the FHIR project. Thank you!
While I don't have any issues with the P.R., I do have a question/observation: How is the last_updated to be updated from somewhere other that the API? The encounter form in itself, once created, is generally not updated/edited and even if it was the new column is not touched from what I see. However any form attached to the encounter that is created or updated would not update the encounters last_updated but would be an update to the encounter on whole.
So is the plan to search all forms attached to the encounter from the API to resolve the last updated request?
I'm horrible at reviews and may have misunderstood the changes and if so, my apologies.

sql/database.sql
```diff
@@ -1944,6 +1944,7 @@ CREATE TABLE `form_encounter` (
   `referring_provider_id` INT(11) DEFAULT '0' COMMENT 'referring provider, if any, for this visit',
   `date_end` DATETIME DEFAULT NULL,
   `in_collection` tinyint(1) default NULL,
+  `last_update` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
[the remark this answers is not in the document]
robertjustjones: Thanks for that pointer, file and syntax.

swagger/openemr-api.yaml
robertjustjones: The syntax for examples didn't like array [], replaced w {} and it worked. Thanks!

sjpadgett:
> Now, whenever a changes is made from any source, the last_updated will be updated.

Yes that is clever and used throughout code however what I was trying to communicate is that the form_encounter is referenced by other encounter forms such as clinical notes, care plans or procedures etc. where if updated or added to encounter the form_encounter is never touched and won't update the last_update.
Let's say an encounter is created on 12-1-2023 but a procedure is ordered referenced to encounter on 12-2-2023 then the last_update would still reflect the original encounter date. So while the new procedure updated the encounter the last_update would not reflect. I'd imagine this would defeat the intention of adding the column.

robertjustjones:
> the form_encounter is referenced by other encounter forms such as clinical notes, care plans or procedures etc.

Ah, I see what you meant. Can you list all/more of "etc."? And the corresponding tables if they're top of mind. I can pull those together in "better" last_updated date.  Preference would be to let this PR through and do that improvement in a separate issue/pr it that's ok.

stephenwaite:
> Dropped the client_credentials commit that failed testing.

Can you update the PR title please @robertjustjones?

Thanks for the improvements and thank you to your team for becoming official sponsors!
smaller_excited_elephant

sjpadgett:
> let this PR through and do that improvement in a separate issue/pr it that's ok.

Sure that's fine. I just didn't want you to make a wrong assumption on the workflow.
Concerning possible forms attached to an encounter, that is difficult to answer. One reason is LBF form created by a clinic, a custom encounter form, contributed form and so forth. Checking forms admin(registry table) for enabled forms could be a path then query all forms using the encounters id as foreign id and eval the dates

adunsulag:
@sjpadgett and @robertjustjones The lastUpdated value is when the specific resource was last updated. https://build.fhir.org/resource-definitions.html#Meta.lastUpdated

I think its a business decision that needs to be made as sub-resources that link to the encounter being updated doesn't mean that the encounter resource itself was changed.  We can make a determination of that, but the cascading effects for that logic can be large as there can be dozens and dozens of sub-resources linked to an encounter based on how many forms an individual has.  I think we leave it that only if the core form_encounter table changes then the encounter 'Resource' changed as that is the 'specific resource' that was last updated.  It makes the coding logic a lot easier.

We used the most current date on a lot of these resources when we first built FHIR because we were not at the time tracking updates very well in much of the resources.  

From a backwards compatibility standpoint it will be a bit problematic and I'm not sure how to address it.  Anyone consuming the API will now have different lastUpdated dates after the update than before.  Overall the system will behave better, but it may lead to some API consumers running into problems if they are doing any kind of fancy logic on the encounter lastUpdated date.

robertjustjones:
I agree that in-depth modified dates needs discussion, and I think handling it separately is appropriate. 

I thought about them backward compatible issue, too. Anyone who implemented code based on that realized what was happening. Every api call shows a new timestamp of now. I think because of that, this change will be transparent to code function, unless for some strange reason someone is counting on lastUpdated to be now, which is clearly wrong.

sjpadgett:
This is something we can hash out offline but in the meantime I recommend if it's important to know the last activity on an encounter you look at the attached forms(may be different than the encounter form update). Other ancillary data that reference the encounter(allergies/problems/issues/procedures) generally have an updated date and specific FHIR resource.  As to what is considered an updated encounter for FHIR spec I think is debatable and not for this encounter!:)
@adunsulag probably won't get to this to review 'till later tonight when he'll probably merge or I will.

sjpadgett:
btw: I wouldn't worry about legacy in this case because any dates would be in the future.

src/Services/ListService.php
```diff
@@ -65,6 +65,17 @@ public function getListOptionsForLists($lists)
         return $records;
     }
 
+    public function getListIds()
+    {
+        $sql = "SELECT DISTINCT list_id FROM list_options ORDER BY list_id";
+        $records = QueryUtils::fetchRecords($sql, $lists, false);
```
adunsulag, line 71: QueryUtils::fetchTableColumn($sql, 'list_id', []); will accomplish this for you as well, but you don't need to change this if you don't want to.

src/Services/FHIR/FhirValueSetService.php
```diff
@@ -0,0 +1,184 @@
...
+        $this->listOptionService = new ListService();
+    }
+
+    /**
+     * Returns an array mapping FHIR Resource search parameters to OpenEMR search parameters
+     */
+    protected function loadSearchParameters()
+    {
+        return [
+        '_id' => new FhirSearchParameterDefinition('_id', SearchFieldType::TOKEN, [new ServiceField('uuid', ServiceField::TYPE_UUID)]),
```
adunsulag, line 106: Is this actually working for you when you do a call to a specific value set?  All of your id values that are returned appear to be string values and are not uuids from what I see in the code.  

Seems like you will need to use this line here:
```
'_id' => new FhirSearchParameterDefinition('_id', SearchFieldType::TOKEN, [new ServiceField('id', ServiceField::TYPE_STRING)]),

File: src/Services/FHIR/FhirValueSetService.php [hunk 5]
```diff
@@ -0,0 +1,184 @@
...
+     * Retrieves all of the fhir observation resources mapped to the underlying openemr data elements.
+     * @param $fhirSearchParameters The FHIR resource search parameters
+     * @return processing result
+     */
+    public function getAll($fhirSearchParameters, $puuidBind = null): ProcessingResult
+    {
+        $fhirSearchResult = new ProcessingResult();
+        try {
+              // Start with postcalendar_categories as appointment-type
+              $calendarCategories = $this->appointmentService->getCalendarCategories($fhirSearchParameters);
```
adunsulag, line 120: Did something get skipped here?  AppointmentService::getCalendarCategories does not take any parameters. I thought perhaps this is where you were filtering on the ValueSet _id but it doesn't appear so as the method has not been modified.

src/Services/FHIR/FhirValueSetService.php
```diff
@@ -0,0 +1,184 @@
...
+        '_id' => new FhirSearchParameterDefinition('_id', SearchFieldType::TOKEN, [new ServiceField('uuid', ServiceField::TYPE_UUID)]),
+        ];
+    }
+
+    /**
+     * Retrieves all of the fhir observation resources mapped to the underlying openemr data elements.
+     * @param $fhirSearchParameters The FHIR resource search parameters
+     * @return processing result
+     */
+    public function getAll($fhirSearchParameters, $puuidBind = null): ProcessingResult
```
adunsulag, line 115: You need to handle the $fhirSearchParameters here in the valueset for the ```_id``` search parameter.  Otherwise the FhirServiceBase::getOne() call which is being called in your FhirValueSetRestController() is just going to return the entire list of values here with no filtering. on the search parameters. 

Since you are handling this in memory you'll just have to key off the calendar category or the list_id that matches with the _id search parameter.  Remember in FHIR that you can specify multiple ```_id``` values in a search query so the search field will give you an array of ids (usually this will just be one, but again it can be multiples).

Let me know if you need more clarity on this.

adunsulag (changes requested):
Overall looks good.  Just two issues I see with the getOne operation and _id search parameter filtering.

src/Services/FHIR/FhirValueSetService.php
adunsulag: I can see that you aren't filtering on this value, so this is likely why it didn't blow up on you earlier.  Still needs to be fixed.

robertjustjones: yep, just forgot to go back and do that.

src/Services/FHIR/FhirValueSetService.php
```diff
@@ -116,30 +114,35 @@ public function getAll($fhirSearchParameters, $puuidBind = null): ProcessingResu
...
-              // Start with postcalendar_categories as appointment-type
-              $calendarCategories = $this->appointmentService->getCalendarCategories($fhirSearchParameters);
-              $valueSet = new FHIRValueSet();
-              $valueSet->setId('appointment-type');
-              $compose = new FHIRValueSetCompose();
-              $include = new FHIRValueSetInclude();
-            foreach ($calendarCategories as $category) {
-                if ($category["pc_cattype"] != 0) {
-                    continue; // only cat_type==0
+            if (!isset($fhirSearchParameters[ '_id' ]) || $fhirSearchParameters[ '_id' ] == 'appointment-type') {
```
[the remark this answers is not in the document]
sjpadgett: While there maybe put a comment here for us so others know.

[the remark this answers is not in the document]
robertjustjones: Actually, array is AND and that will never work for multiple values, but the spec also includes comma delimited OR so that's now implemented.  I wasn't grokking the method from the trait so I coded this explicitly with comments.

src/Services/FHIR/FhirValueSetService.php
```diff
@@ -0,0 +1,207 @@
...
+              // Now the same for list_options selected in $listNames
+              $list_ids = $this->listOptionService->getListIds();
+            foreach ($list_ids as $listName) {
+                if (
+                    isset($fhirSearchParameters[ '_id' ])
+                    // could be array (AND) or comma-delimited string value (OR)
+                    // check array first but should only be len 1 ("AND", becuase cannot be 2 simultaneous)
+                    && ( ( is_array($fhirSearchParameters[ '_id' ])
+                         && count($fhirSearchParameters[ '_id' ]) == 1
+                         && $fhirSearchParameters[ '_id' ][ 0 ] == $listName )
```
adunsulag, line 159: Shouldn't this be a ```$fhirSearchParameters[ '_id' ][ 0 ] != $listname``` in order to skip it if there is only one _id clause and it doesn't match the current list item? Or am I misreading this expression?

adunsulag (changes requested):

src/Services/FHIR/FhirValueSetService.php
robertjustjones: ha ha yes

adunsulag:
@bradymiller did something change on the CI engine or a dependency further upstream?  This PR doesn't contain any changes to the composer.json or package.json so we shouldn't be seeing the failure errors that are happening across all php versions. Any idea what could be going on here?

robertjustjones:
@adunsulag The last one passed. Maybe it was something spurious. Can you kick off those checks again?

sjpadgett:
This sometimes happens. A rerun usually will fix.

stephenwaite:
keeps failing
angry-mad

sjpadgett:
huh! beats me need to look at test

adunsulag:
Brought this in.  Thank you @robertjustjones for the fixes and contributions!

jack5github:
Screenshot

Just a heads up that this commit is causing an SQL failed preparation message to be prepended to all encounters because of the missing column.

bradymiller:
hi @jack5github , what is the missing column?

stephenwaite:
hi @jack5github , if you merge master into a branch when there were upstream changes in the `sql` folder you have to run `sql_upgrade.php` accordingly.

PR #7114, opened 2023-12-15
Implement email_direct in the FHIR Patient api.

### LINKED ISSUE #7113: Trusted Email is missing from FHIR Patient
The Trusted Email (email_direct) field is missing from the FHIR Patient api. It should be in the output response and be included in the search params "email" and "telecom". It should be available for read and write.

robertjustjones:
Adds email_direct to the FHIR Patient Service.  

This commit maps the "Trusted Email" to (system)email, (use)mobile, both for output and input. It avoids the pre-existing mapping conflict/collision for home/work email -> email.  Justification of the mapping comes from the FHIR ContactPointUse spec for "mobile" (for urgent matters, not routine ones):

> A telecommunication device that moves and stays with its owner.  May have characteristics of all other use codes, suitable for urgent matters, not the first choice for routine business."

Fixes #7113.

### DIFF
--- a/src/Services/FHIR/FhirPatientService.php
+++ b/src/Services/FHIR/FhirPatientService.php
@@ -138,11 +138,11 @@ class FhirPatientService extends FhirServiceBase implements IFhirExportableResou
                 ['state', 'contact_address_state']
             ),
 
-            'email' => new FhirSearchParameterDefinition('email', SearchFieldType::TOKEN, ['email']),
+            'email' => new FhirSearchParameterDefinition('email', SearchFieldType::TOKEN, ['email','email_direct']),
             'family' => new FhirSearchParameterDefinition('family', SearchFieldType::STRING, ['lname']),
             'given' => new FhirSearchParameterDefinition('given', SearchFieldType::STRING, ['fname', 'mname']),
             'phone' => new FhirSearchParameterDefinition('phone', SearchFieldType::TOKEN, ['phone_home', 'phone_biz', 'phone_cell']),
-            'telecom' => new FhirSearchParameterDefinition('telecom', SearchFieldType::TOKEN, ['email', 'phone_home', 'phone_biz', 'phone_cell']),
+            'telecom' => new FhirSearchParameterDefinition('telecom', SearchFieldType::TOKEN, ['email','email_direct', 'phone_home', 'phone_biz', 'phone_cell']),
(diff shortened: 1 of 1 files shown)


robertjustjones:
CC'ing @bradymiller @mi3guyc

robertjustjones:
Looks like the CI issue persists.

sjpadgett:
I'm rerunning failed test. may have to rerun them all though.

src/Services/FHIR/FhirPatientService.php
```diff
@@ -308,6 +308,13 @@ private function parseOpenEMRPatientTelecom(FHIRPatient $patientResource, $dataR
         if (!empty($dataRecord['email'])) {
             $patientResource->addTelecom($this->createContactPoint('email', $dataRecord['email'], 'home'));
         }
+        if (!empty($dataRecord['email_direct'])) {
+            $patientResource->addTelecom($this->createContactPoint('email', $dataRecord['email_direct'], 'mobile'));
```
adunsulag, line 312: This seems odd to me.  The email_direct is for email that goes over the secure Direct Network Exchange (a good reading is here: https://www.emrdirect.com/about-directed-exchange-and-secure-direct-messaging.html).  It seems like these communications would be for routine business that needs to be used for transmitting secure health information and not for urgent communications.  I would think we would use 'work' here instead of 'mobile'.

robertjustjones: Thanks for the link. I had missed that connotation. It should be explained as such in the UI. If it is, I missed that. Without going deeper, seems like there's a registration step implied/involved beyond just an email address. I'll look for that.

With all of that, I think "work" email is the opposite.  For most, that's definitely not a more private/protected channel like "direct" is. I certainly assume my "work" email is scanned, and it should be anywhere, and I work for a terrific guy (hi @mi3guyc). 

"mobile" on the other hand is (assumed) private, and that choice is rarely used with "email" in my experience with other FHIR implementations.  Since the list of choices is short, I still think mobile is logical.

I alluded to collisions on email. The db doesn't support work/home email so the code is/was written to push the last one received into the email field.  That's still an issue not addressed here, but I think it should be someday like other EMR's.

robertjustjones: I think this is resolved.

PR #7130, opened 2023-12-30
Template share selection bug fix

### LINKED ISSUE #7129: bug: Unable to share templates in nations note
### Describe the bug

In the nation's note, there used to be the ability to share templates with other users. That feature is broken,
In the first step of sharing the templates, the desired template would be selected and moved from left to right on the screen.
The screen shows a failure in this step.

### To Reproduce

image

### Expected behavior

The selected list is to move left to right. 

### Client configuration

- Browser: All
- OpenEMR version: 7.0.2
- Operating system:

juggernautsei:
Fixes #7129

fixes the failed error received when trying to share a template

### DIFF
--- a/library/ajax/template_context_search.php
+++ b/library/ajax/template_context_search.php
@@ -37,8 +37,7 @@ Where
     cl2.cl_deleted = 0 And
     cl3.cl_list_type = 3 And
     cl3.cl_deleted = 0 And
-    cl2.cl_list_item_long Like ? And
-    cl3.cl_creator = ?
+    cl2.cl_list_item_long Like ?
 Group By
     cl2.cl_list_item_long
 createQuery;
@@ -46,7 +45,7 @@ createQuery;
 $search = $_GET['search'];
 $eSearch = "%" . $search . "%";
 $results = [];
-$r = sqlStatementNoLog($cq, array($eSearch, (int)$_SESSION['authUserID']));
+$r = sqlStatementNoLog($cq, array($eSearch));
 
 while ($result = sqlFetchArray($r)) {
     $results[] = array_map('text', $result);
--- a/library/custom_template/ajax_code.php
+++ b/library/custom_template/ajax_code.php
@@ -80,14 +80,15 @@ if ($Source == "add_template") {
     $newid = sqlInsert("INSERT INTO customlists (cl_list_id,cl_list_type,cl_list_item_long,cl_order,cl_creator) VALUES (?,?,?,?,?)", array($templateid, 4, $content, $order, $_SESSION['authUserID']));
     sqlStatement("UPDATE template_users SET tu_template_id=? WHERE tu_template_id=? AND tu_user_id=?", array($newid, $item, $_SESSION['authUserID']));
 } elseif ($Source == 'item_show') {
-    $sql = "SELECT * FROM customlists WHERE cl_list_id=? AND cl_list_type=4 AND cl_deleted=0";
+    $list_id = substr($_REQUEST['list_id'], -1);
+    $sql = "SELECT * FROM customlists WHERE cl_list_id=? AND cl_list_type=3 AND cl_deleted=0";
     $res = sqlStatement($sql, array($list_id));
     $selcat = sqlQuery("SELECT * FROM customlists WHERE cl_list_slno=? AND cl_list_type=3 AND cl_deleted=0", array($list_id));
     $selcont = sqlQuery("SELECT * FROM customlists WHERE cl_list_slno=? AND cl_list_type=2 AND cl_deleted=0", array($selcat['cl_list_id']));
     $cnt = sqlNumRows($res);
     if ($cnt) {
         echo "<table width='100%'>";
-        echo "<tr class='text'><th colspan=2  style='background-color:var(--white)'>" . htmlspecialchars(xl('Preview of'), " " . $selcat['cl_list_item_long'] . "(" . $selcont['cl_list_item_long'] . ")", ENT_QUOTES) . "</th></tr>";
+        echo "<tr class='text'><th colspan=2  style='background-color:var(--white)'>" . htmlspecialchars(xl('Preview of') . " " . $selcat['cl_list_item_long'] . "(" . $selcont['cl_list_item_long'] . ")", ENT_QUOTES) . "</th></tr>";
         $i = 0;
         while ($row = sqlFetchArray($res)) {
             $i++;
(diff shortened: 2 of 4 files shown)


juggernautsei:
I have corrected most of the issues with sharing templates. The template is being saved to the new users profile. Now, it needs to be called back when the user needs it. (Note to self)

juggernautsei:
Hi @mdsupport ,
I see you have a note in the custom_template.php file on line 37. What I am failing to see right now is why the system can't retrieve the shared template. 
image

 Even on reloading the entire modal, the frame is empty.
image

The template_context_search.php does not reference the template_users table at all. 

    let allowTemplateWarning = <?php echo $allowTemplateWarning; ?>;
    <?php if (!$isNN) { ?>
        $(function () {
            $('#contextSearch').select2({
                placeholder: <?php echo xlj('Select Template Context'); ?>,
                width: 'resolve',
                theme: 'bootstrap4',
                ajax: {
                    url: top.webroot_url + '/library/ajax/template_context_search.php',
                    data: function (params) {
                        let query = {
                            search: params.term,
                            csrf_token_form: <?php echo js_escape(CsrfUtils::collectCsrfToken()); ?>
                        };
                        return query;
                    },
                    dataType: 'json',
                },
                <?php require($GLOBALS['srcdir'] . '/js/xl/select2.js.php'); ?>
            });

This populates the Context select dropdown.  Template users is not used to select user. If the user is not the one that created the entries. This is where I am stuck. I don't understand the way the template users table is supposed to relate to the

juggernautsei:
When I remove the criteria to fetch the data by user id. The data for the person that the code is shared to appears in the context window.

bradymiller:
hi @juggernautsei , Look like you've sorted out this issue per the comment in your last commit. If so, then code looks good. Just let us know it is testing well for you and then will bring this in.

juggernautsei:
@bradymiller I have tested on three installs including v7.0.1 and it works well on all three.

library/ajax/template_context_search.php
```diff
@@ -37,16 +37,15 @@
...
+    cl2.cl_list_item_long Like ?
 Group By
     cl2.cl_list_item_long
 createQuery;
 
 $search = $_GET['search'];
 $eSearch = "%" . $search . "%";
 $results = [];
-$r = sqlStatementNoLog($cq, array($eSearch, (int)$_SESSION['authUserID']));
+$r = sqlStatementNoLog($cq, array($eSearch));
```
sjpadgett, line 48: There's a reason the query includes the creator id, do you know what it is before removing to make this work?
Are you sure that now all user templates aren't exposed?
Does this affect test templates. I doubt it but worth testing to be sure.

I don't want to have to did into this code as last time I worked in it it was working and my clinic uses both text and NN extensively so I don't want to see any more issues I have to get calls on:)
So far they have not had any issues but only had running 702 for 5 business days.

juggernautsei: @sjpadgett I don't have enough testing to answer your questions. I searched the code trying to find why the authUserID was there. I went all the way back to v5. The auth user ID is in v5 but I am too ignorant to see the purpose. In my use case, there is only one person creating the templates. I can tell you that when removed the parameter. The shared template showed to the person that wanted to use it. They don't show up to everyone and only the ones that are selected show to the one that requested the share. I would have to do more digging and trapping to figure out more than that.

mdsupport:
> Hi @mdsupport ,
> I see you have a note in the custom_template.php file on line 37.

The function on line 38 relates purely to emitting html. Intent must have been to make the original code bit modular and extensible.

mdsupport:
> I don't understand the way the template users table is supposed to relate to the

Dropping existing selection filters will fix your use case but also change basic design. Have you tried to look at commit history to find what broke the original functionality?

sjpadgett:
Copy templates aren't being inserted so retrieving by user id won't show is why it looks like it will work.
Also changing the list type may not be appropriate and should be 4....

sjpadgett:
Several of the queries in personalize.php are broken. Select context is one around L-395

juggernautsei:
@mdsupport  "Have you tried to look at commit history to find what broke the original functionality?"
No, there were a lot of changes made from 5 to 6 since 6. It has not had as many changes in the personalize file. 
 
@sjpadgett  "list type may not be appropriate and should be 4"
In our system the list type is not four that is why the fail comes up when trying to select a context to copy. This is where the whole thing started.

It was sure that it would fix my use case but break past. Someone change the list type from 4 to 3 because that is what saves currently.

sjpadgett:
@juggernautsei Are you sure you are using correctly? or am I?
you should be able to select a user and context. move template(4) cat over to right and stick.  The 4 in query is used  to lookup and show the category details in a bottom panel when you select from left panel.
First thing is to fix the query errors. One query is being passed a filter however it is not bound in statement about L-385.

BTW list type is set in stone(hard coded) and 4 is the component details. 3 is category, 2 is context no clue of 1 and 6

Fixing errors is step one

sjpadgett:
@juggernautsei 
I found fix for the original issue. Looks like the list event is also submitting the form so on refresh the list is emptied because not saved.

Add a event.preventDefault() or in tag event in the `function jsub_selected(form, selectFrom, selectedList)` function. Around L-206 of personalize.php.

Remove change to lookup using 3 and completely remove change to my select lookup script i.e put back authUserID.

sjpadgett:
Hi Sherwin, Where are you at with this? I hope you are also testing flow of selecting a context and user when assigning shared templates. There is still a bug with sql queries. The one error from selecting a context is kind of weird and may be a design issue.
Report back if you're seeing error also.

sjpadgett:
`
[04-Jan-2024 12:22:09 America/New_York] SQL Error with statement:query failed: SELECT * FROM customlists WHERE cl_list_type=3 AND cl_deleted=0 AND cl_list_slno NOT IN (SELECT DISTINCT tu_template_id FROM template_users)  AND cl_list_id=? ORDER BY cl_list_id,cl_list_item_long--You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near &#039;? ORDER BY cl_list_id,cl_list_item_long&#039; at line 1==&gt;C:\xampp\htdocs\openemr\library\custom_template\personalize.php at 398:sqlStatement
`

sjpadgett:
@juggernautsei Could I go ahead and take this issue over. There are other bugs that need fixed. Also warnings that I'm sure will turn into errors in PHP 8.4! I'll try to have a PR up today and bring in tomorrow.

juggernautsei:
@juggernautsei not feeling well right now. I will take this up as soon as I am in better condition. Thanks for the insight.

sjpadgett:
No problem for me to go ahead and fix this since I have already fix a bunch of warnings and sql errors. It's not a slight against you but would make my fixes easier to bring in so I can move on to other things.

juggernautsei:
That is a good idea. Carry on.

sjpadgett:
Thanks buddy, hope you feel better. I'll close this and pick up your issue in my PR.

PR #7140, opened 2024-01-04
feat: Include other "users" in FhirPractitioner.

### LINKED ISSUE #7139: FHIR Practitioners should include all users
FHIR Practitioners includes only the records returned by the `/api/practitioner` where `/api/user` is available to retrieve "other" users.  FHIR [defines Practitioner](http://hl7.org/fhir/R4/practitioner.html) to include those "others".  The FHIR Practitioners resource should return all of the users.

Screenshot 2024-01-04 at 1 43 35 PM

robertjustjones:
Modify FHIR Practitioners api to return all users.

Fixes #7139 

@adunsulag @sjpadgett @mi3guyc

### DIFF
--- a/src/Common/Auth/OpenIDConnect/Entities/UserEntity.php
+++ b/src/Common/Auth/OpenIDConnect/Entities/UserEntity.php
@@ -36,13 +36,7 @@ class UserEntity implements ClaimSetInterface, UserEntityInterface
             $userRole = $uuidToUser->getUserRole();
             $fhirUserResource = "Person";
             if ($userRole == UuidUserAccount::USER_ROLE_USERS) {
-                // need to find out if its a practitioner or not
-                $practitionerService = new PractitionerService();
-                // ONC validation does not accept Person as a valid test case so we have to differentiate practitioners
-                // from the more generic Person resource.
-                if ($practitionerService->isValidPractitionerUuid($this->identifier)) {
-                    $fhirUserResource = "Practitioner";
-                }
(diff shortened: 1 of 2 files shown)


robertjustjones:
@adunsulag I missed running the tests and seeing the post/patch errors. 

So now a decision on how to go forward... how/where to implement adding FHIR Practitioners without NPI numbers?  

We could implement the missing methods in UserService, with a UserValidator that omits the required NPI, or relax the corresponding checks in Practitioner.  Preference?

robertjustjones:
I compromised here to avoid the decision above. I changed my PR to reference all users for the GET, but not for PATCH/POST. So non-NPI "users" cannot be added or updated (still).

adunsulag:
@robertjustjones We'll need to make sure this passes ONC's inferno test suite.  At one point the test tools would fail if Practitioner's did not have NPI numbers.  I've been doing some digging and it looks like they may have fixed that test case.  We will still need to test that assumption.

robertjustjones:
@adunsulag I ran the ONC tests and found a fail for Admin using fhirUser Person, which I realized I'd meant to change as well.  All user role Users can be Practitioners in the OIDC getClaims code.  That commit is coming.  With it, here's the ONC EHR Practitioner App test using user Admin (without NPI).

Screenshot 2024-01-05 at 9 22 20 AM

stephenwaite:
sometimes they randomly fail, no biggie

PR #7144, opened 2024-01-05
fix: Fix Nation Notes unable to share templates

### LINKED ISSUE #7129: bug: Unable to share templates in nations note

sjpadgett:
- add preventdefault to keep form from submitting on button
- fix numerous php warnings
- fix sql query error for orphan query
- some restyle for colors

This picks up from PR #7130

Fixes #7129

### DIFF
--- a/library/custom_template/ajax_code.php
+++ b/library/custom_template/ajax_code.php
@@ -31,12 +31,12 @@ require_once("../../interface/globals.php");
 
 use OpenEMR\Common\Acl\AclMain;
 
-$templateid = $_REQUEST['templateid'];
-$Source = $_REQUEST['source'];
-$list_id = $_REQUEST['list_id'];
-$item = $_REQUEST['item'];
-$multi = $_REQUEST['multi'];
-$content = $_REQUEST['content'];
+$templateid = $_REQUEST['templateid'] ?? '';
+$Source = $_REQUEST['source'] ?? '';
+$list_id = $_REQUEST['list_id'] ?? '';
+$item = $_REQUEST['item'] ?? '';
+$multi = $_REQUEST['multi'] ?? '';
+$content = $_REQUEST['content'] ?? '';
 
 if ($Source == "add_template") {
     $arr = explode("|", $multi);
@@ -86,18 +86,17 @@ if ($Source == "add_template") {
     $selcont = sqlQuery("SELECT * FROM customlists WHERE cl_list_slno=? AND cl_list_type=2 AND cl_deleted=0", array($selcat['cl_list_id']));
     $cnt = sqlNumRows($res);
     if ($cnt) {
-        echo "<table width='100%'>";
-        echo "<tr class='text'><th colspan=2  style='background-color:var(--white)'>" . htmlspecialchars(xl('Preview of'), " " . $selcat['cl_list_item_long'] . "(" . $selcont['cl_list_item_long'] . ")", ENT_QUOTES) . "</th></tr>";
+        echo "<table class='table table-dark table-striped table-sm'>";
+        echo "<tr class='text bg-dark text-light'><th colspan=2 class='text bg-dark text-light'>" . text(xl('Preview of') . " " . $selcat['cl_list_item_long'] . "(" . $selcont['cl_list_item_long'] . ")") . "</th></tr>";
         $i = 0;
         while ($row = sqlFetchArray($res)) {
             $i++;
-            $class = ($class == 'reportTableOddRow') ? 'reportTableEvenRow' : 'reportTableOddRow';
-            echo "<tr class='text'><td style='background-color:var(--white)'>" . $i . "</td><td style='background-color:#ffffff'>" . htmlspecialchars($row['cl_list_item_long'], ENT_QUOTES) . "</td></tr>";
+            echo "<tr class='text'><td class='bg-dark text-light'>" . text($i) . "</td><td>" . htmlspecialchars($row['cl_list_item_long'], ENT_QUOTES) . "</td></tr>";
         }
         echo "</table>";
     } else {
         echo "<table width='100%'>";
-        echo "<tr class='text'><th colspan=2  style='background-color:var(--white)'>" . htmlspecialchars(xl('No items under selected category'), ENT_QUOTES) . "</th></tr>";
+        echo "<tr class='text bg-dark text-light'><th colspan=2>" . htmlspecialchars(xl('No items under selected category'), ENT_QUOTES) . "</th></tr>";
         echo "</table>";
     }
     $Source = "add_template";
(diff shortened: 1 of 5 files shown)


sjpadgett:
I still need to fix text template to allow template created in nation notes and vica versa.

sjpadgett:
@juggernautsei I thought you were testing NN and not Text Notes.
Concerning authUserID in search for context in TN I prevent using component created in NN in TN because I thought because of text limitation in TN there might be issues with stylings. Turns out thats not true so I'm going to allow for this PR.
I wish you'd have mention using TN in your PR....

sjpadgett:
@juggernautsei This is testing very well for me. I'm glad you brought up sharing templates between NN and TN which I now allow. Would have done sooner if it was brought up.
Sharing between users is working well also where you can now select context, users and left panel now sticks. I have hundreds of templates on my dev server so it is a pretty good test but let me know if your client finds issues.

stephenwaite:
excited_elephant

sjpadgett:
fyi just cherry picked to v7.0.2 for next patch

library/custom_template/ajax_code.php
```diff
@@ -171,9 +170,9 @@
...
+        echo "<li class='bg-dark text-light' style='cursor:pointer'><span class='bg-dark text-light' onclick='add_item()'>" . htmlspecialchars(xl('Click to add new components'), ENT_QUOTES);
         echo "</span><div id='new_item' style='display:none' class='w-100'>";
-        echo "<textarea name='item' id='item' class='w-100'></textarea><br />";
-        echo "<input type='button' name='save' value='" . htmlspecialchars(xl('Save'), ENT_QUOTES) . "' onclick='save_item()'><input type='button' name='cancel' value='" . htmlspecialchars(xl('Cancel'), ENT_QUOTES) . "' onclick=cancel_item('" . htmlspecialchars($row['cl_list_slno'], ENT_QUOTES) . "')></div></li>";
+        echo "<textarea name='item' id='item' class='w-100 bg-dark text-light'></textarea><br />";
+        echo "<input type='button' name='save' value='" . htmlspecialchars(xl('Save'), ENT_QUOTES) . "' onclick='save_item()'><input type='button' name='cancel' value='" . htmlspecialchars(xl('Cancel'), ENT_QUOTES) . "' onclick=cancel_item('" . attr_js($row['cl_list_slno'] ?? '') . "')></div></li>";
```
bradymiller, line 176: should drop the outside `'` around the attr_js call
(nice to get to provide some escaping input every once i awhile to feel useful :) )

library/custom_template/personalize.php
```diff
@@ -288,161 +288,155 @@ function personalize_save() {
...
+                    if (AclMain::aclCheckCore('nationnotes', 'nn_configure')) {
+                        ?>
+                        <a href="delete_category.php" id="share_link" class="iframe_medium btn btn-primary" onclick="top.restoreSession();"><?php echo text(xl('Delete Category')); ?></a>
+                        <?php
+                    }
+                    ?>
+                    <?php
+                    if (AclMain::aclCheckCore('nationnotes', 'nn_configure')) {
+                        ?>
+                        <a href="add_template.php?list_id=<?php echo attr($_REQUEST['list_id']); ?>" onclick="top.restoreSession();" class="iframe_small btn btn-primary" title="<?php echo text(xl('Add Category')); ?>"><?php echo text(xl('Add Category')); ?></a>
```
bradymiller, line 337: attr instead of text for title attribute

library/custom_template/personalize.php
```diff
@@ -288,161 +288,155 @@ function personalize_save() {
...
+                        <?php
+                    }
+                    ?>
+                    <?php
+                    if (AclMain::aclCheckCore('nationnotes', 'nn_configure')) {
+                        ?>
+                        <a href="add_context.php" class="iframe_medium btn btn-primary" onclick="top.restoreSession();" title="<?php echo text(xl('Add Context')); ?>"><?php echo text(xl('Add Context')); ?></a>
```
bradymiller, line 344: attr instead of text for title attribute

library/custom_template/personalize.php
```diff
@@ -288,161 +288,155 @@ function personalize_save() {
...
+                    <button name="remove" class="btn btn-secondary" onclick="jsub_selected(document.myform,'personalized','topersonalized')">&raquo;</button>
+                    <br />
+                    <button name="remove" class="btn btn-secondary" onclick="check_user_category(document.myform,'topersonalized','personalized')">&laquo;</button>
+                </div>
+                <div class="col-sm-5">
+                    <select multiple class="form-control" name="personalized[]" id="personalized" size="6">
+                        <?php
+                        $where = '';
+                        if ($filter_context ?? null) {
+                            $where .= " AND cl_list_id='" . $filter_context . "'";
```
bradymiller, line 419: bookmarking this for issue

bradymiller:
tenor99

library/custom_template/ajax_code.php
sjpadgett: it's fine Brady catch as catch can!:)
Maybe @stephenwaite can add this to his PR though I disagree it is needed. Again, inspect it in browser.

library/custom_template/personalize.php
sjpadgett: Really should be bound but this is why I hate touching these old scripts, everything needs to be done to them....
Again i'd appreciate if @stephenwaite would fix these or it will have to wait for my next PR

library/custom_template/ajax_code.php
bradymiller: @sjpadgett , Current codebase doesn't have the `'` in all the calls and have not seen issue with this yet. Is there a use case where this breaks?

sjpadgett: Last issue I had was in one of my modules. I believe it was ringcentral but no matter. In this use case I suppose it doesn't matter but in the case where passing a string or an int was my issue I believe. Still either way it will parse and render the same up to the browser. I don't have an explanation for that!
I'll use it the way you intended though.

bradymiller: Sounds good. Note in the special case of an int, attr_js will not add an `'` (json_encode does not add surrounding quote to int), which would be a reason not to also hard-code quotes around this function (since would then convert a int to a string).

PR #7150, opened 2024-01-07
Modules Manager action hooks for module cleanup

### LINKED ISSUE #7147: Add module perform after actions functionality to Module Manager
### Describe the problem
Need a way for a module to performs actions on a module disable, enable, unregister

Will call a defined static class function from a module root file to perform the various actions performed by the laminas InstallerController class.

File name will be ModuleManagerAfterActions.php with a static class of same. Unless laminas gives me trouble!
Hopefully we can use for upgrades too however manager has a path for upgrade based on module sql version I'm still figuring out.

sjpadgett:
- new class to include in modules directory root called by Laminas Module Mananger

Fixes #7147


sjpadgett:
@adunsulag If you get a chance before you disappear tomorrow could you take a look at this concept.
I'd like to get merged early Monday.

sjpadgett:
btw: not finished with cleanup

interface/modules/zend_modules/module/Installer/src/Installer/Controller/InstallerController.php
```diff
@@ -135,6 +135,10 @@ public function manageAction()
         $request = $this->getRequest();
         $status = $this->listenerObject->z_xlt("Failure");
         if ($request->isPost()) {
+            $modId = $request->getPost('modId') ?? null;
+            $registryEntry = $this->getInstallerTable()->getRegistryEntry($modId, "mod_directory");
+            $dirModule = $registryEntry->modDirectory;
+            $modType = $registryEntry->type;
             if ($request->getPost('modAction') == "enable") {
                 $status = $this->EnableModule($request->getPost('modId'));
```
sjpadgett, line 143: What the heck! This is backwards. enable actually disables the module and disable enables in the controller functions.
Appears interface is reporting the actions button name and not what it is supposed to do!
I guess while here I'll fix this but gesh all I wanted was to add the hooks and not go traipsing all over Laminas!!!

interface/modules/custom_modules/oe-module-faxsms/ModuleManagerActionListener.php
```diff
@@ -0,0 +1,106 @@
...
+ * Currently register isn't supported and support should be a part of install.
+ * If an error needs to be reported to user, return description of error.
+ * However, whatever action trapped here has already occurred in Manager.
+ * Catch any exceptions because chances are they will be overlooked in Laminas module.
+ * Report them in return value.
+*/
+
+class ModuleManagerActionListener
+{
+    // Prevent instantiation
```
sjpadgett, line 24: @bradymiller Do we have somewhere to put this file perhaps we have example module somewhere.

sjpadgett:
It is always an adventure when I look into the Laminas module code!:)

sjpadgett:
This button does the same as Default Install. I'm missing something. Looking for upgrade path maybe from disabled.
Looks like if have a 'sql' directory, either button looks for table/install.php there only. Had to fix path for Install SQL to account for custom_module.
I though billing takes you on a journey but Laminas is a marathon.

image

zerai:
What do you think about using events? 
A "fire and forget" approach can be more flexible and remove the risk of coupling with modules.

sjpadgett:
Hi @zerai Thanks for the comment!
I actually considered the event route because on the face it, it seems the logical choice. My concern was bridging the two event systems and concerns for maintaining the two event dependencies where say a new symfony version may break the laminas version vica versa.
The current method is working well where if the class is in module root it is auto enabled and any unused methods are ignored if just the skeleton class is there.

Still I'm not married to current and may change my mind upon further research. I'm happy to take input on either.

sjpadgett:
Restyled Module Manager to follow theme background and text colors. Also added hover making select correct buttons easier. Below is dark theme.
image
image

sjpadgett:
merging soon. tired of this!:)

sjpadgett:
Anyone have suggestions on workflow to do an auto upgrade on module bootstrap from ModuleLoader.
Or maybe develop a new function in sql_upgrade.php that scans modules looking for upgrade.

sjpadgett:
Does it ever end!!! I still want to go ahead and style the other care modules to at least follow our themes bg and text colors.
install, upgrade and default sql actions work except upgrade needs to be fixed to update version in DB and refresh view after action.

It would have been nice to have some companionship for this project because it is more enjoyable doing these tedious projects where ideas can be bounced around and at minimum vane some interest.....

I don't know if I will do more or not. I hope this process is not too hard for people to catch onto but I may put a period on it for now and merge and test on dev demo later..

bradymiller:
tenor99

PR #7172, opened 2024-01-18
Fix #7171 section event append

### LINKED ISSUE #7171: bug: SectionEvent->addCard append places second function call in first index position
### Describe the bug
SectionEvent->addCard implementation has a bug where the second function call to append a card will actually PREPEND the card to the card arrays.  

This occurs because of the use of array_splice being given a -1 offset which means it goes back from the array length - 1.   For an array of only 1 items this will make the offset to be 0 which ends up prepending the item.  You can see this with array_splice by doing the following: 
```
$arr = [5]; array_splice($arr, -1, 0, [3]); var_dump($arr); // output will be [3,5]
```

### To Reproduce
Call the following arguments
```
$event = new SectionEvent();
$event->addCard(new DemographicsViewCard());
$event->addCard(new BillingViewCard());

var_dump($event->getCards());
// output will have billing view card followed by demographics view card.
```
You'll find that BillingViewCard is the first card in the list.  

### Expected behavior
Cards should be appended in their proper order on the demographics screen page.

### Client configuration

- Browser:
- OpenEMR version: 7.0.1+

adunsulag:
Fixes #7171 so that the section event order of operations is done correctly.

### DIFF
--- a/src/Events/Patient/Summary/Card/SectionEvent.php
+++ b/src/Events/Patient/Summary/Card/SectionEvent.php
@@ -81,8 +81,11 @@ class SectionEvent extends Event
         // if (!is_int($position) || !is_null($position)) {
         //     throw new LogicException('Position parameter must be either null or an interger');
         // }
-
-        array_splice($this->cards, $position ?? -1, 0, array($card));
+        if ($position == null || !is_int($position)) {
+            $this->cards[] = $card;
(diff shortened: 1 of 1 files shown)


robertdown (approved):

bradymiller:
tenor99

PR #7177, opened 2024-01-19
Fix for Nation Notes

### LINKED ISSUE #7176: Nation Notes components render broken.
### Describe the bug
Recent refactor of ajax_code.php caused an issue with building the component list.
image
Should be:
image

I haven't been able to put my finger on issue yet but will get it eventually!:)

### Client configuration

- Browser: all
- OpenEMR version: 7.0.3-dev
- Operating system:

sjpadgett:
Render build for component list needed refactor.

Fixes #7176


stephenwaite:
coffee_dance

sjpadgett:
This is the popup to replace tags in a component. For some reason it is a window which is not modal and should be a dialog so work action stays with the parent dialog. Was this always this way? rhetorical mostly!

image

sjpadgett:
The original dialog was misused so a window replaced instead of fixing how dialog was called. I wonder if I developed dialog where folks don't know how to use! Regardless I fixed.

bradymiller:
2DV-2

PR #7192, opened 2024-01-25
Hide dashboard cards option

### LINKED ISSUE #7191: feat: Hide cards in the patient dashboard
### Describe the problem

Every practice that use OpenEMR is not a primary care physician. It would be nice to be able to turn off the cards that are not needed. 

The ability to turn off these cards
Allergies,
Amendments,
Disclosures,
Labs,
Medical Problems,
Prescriptions,
Vitals

This would clean up the dashboard of cards that are not being used by clinics.

juggernautsei:
Fixes #7191

Creates the ability to turn off cards that are not being used by the clinic

Create a list of cards and use it in an array to to display or hide the card. This is what the dashboard looks like with the cards turned off.

image

below added by @bradymiller :
UpForGrabs demo for testing this PR is at: https://www.open-emr.org/wiki/index.php/Development_Demo#Alpha_-_Up_For_Grabs_Demo

### DIFF
--- a/Documentation/help_files/medical_dashboard_help.php
+++ b/Documentation/help_files/medical_dashboard_help.php
@@ -163,7 +163,7 @@ require_once("../../interface/globals.php");
                 <div class="col-sm-12">
                     <h4 class="oe-help-heading"><?php echo xlt("Data Section"); ?><a href="#"><i class="fa fa-arrow-circle-up oe-pull-away oe-help-redirect" aria-hidden="true"></i></a></h4>
 
-                    <p><?php echo xlt("The some of the cards on the dashboard can be turned off if not being utilized by the clinic");?></p>
+                    <p><?php echo xlt("Some of the cards on the dashboard can be turned off if not being utilized by the clinic");?></p>
 
                     <p><?php echo xlt("To see which cards can be hidden, go to Admin, Forms, Lists, then search for Hide Dashboard Cards");?>.</p>
 
--- a/Documentation/help_files/medical_dashboard_help.php
+++ b/Documentation/help_files/medical_dashboard_help.php
@@ -162,6 +162,11 @@ require_once("../../interface/globals.php");
             <div class= "row" id="section3">
                 <div class="col-sm-12">
                     <h4 class="oe-help-heading"><?php echo xlt("Data Section"); ?><a href="#"><i class="fa fa-arrow-circle-up oe-pull-away oe-help-redirect" aria-hidden="true"></i></a></h4>
+
+                    <p><?php echo xlt("The some of the cards on the dashboard can be turned off if not being utilized by the clinic");?></p>
+
+                    <p><?php echo xlt("To see which cards can be hidden, go to Admin, Forms, Lists, then search for Hide Dashboard Cards");?>.</p>
+
                     <p><?php echo xlt("The data section of the dashboard page lists all pertinent items related to a patient");?>.</p>
 
                     <p><?php echo xlt("These items can be edited if the user has sufficient privilege");?>.</p>
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -74,7 +74,7 @@ if (isset($_GET['set_pid'])) {
 // want smart support in their system.
 $smartLaunchController = new SMARTLaunchController($GLOBALS["kernel"]->getEventDispatcher());
 $smartLaunchController->registerContextEvents();
-$activeCards = getActiveDashboardCards();
+$hiddenCards = getHiddenDashboardCards();
 
 /**
  * @var EventDispatcher
@@ -1080,7 +1080,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=allergy')"
                         ];
                         echo "<div class=\"$col\">";
-                        if (in_array('card_allergies', $activeCards)) {
+                        if (in_array('card_allergies', $hiddenCards)) {
                             echo $t->render('patient/card/allergies.html.twig', $viewArgs);
                         }
                         echo "</div>";
@@ -1106,7 +1106,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=medical_problem')"
                         ];
                         echo "<div class=\"$col\">";
-                        if (in_array('card_medicalproblems', $activeCards)) {
+                        if (in_array('card_medicalproblems', $hiddenCards)) {
                             echo $t->render('patient/card/medical_problems.html.twig', $viewArgs);
                         }
                         echo "</div>";
@@ -1130,7 +1130,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=medication')"
                         ];
                         echo "<div class=\"$col\">";
-                        if (in_array('card_medication', $activeCards)) {
+                        if (in_array('card_medication', $hiddenCards)) {
                             echo $t->render('patient/card/medication.html.twig', $viewArgs);
                         }
                         echo "</div>";
@@ -1461,7 +1461,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        if (in_array('card_patientreminders', $activeCards)) {
+                        if (in_array('card_patientreminders', $hiddenCards)) {
                             echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
                         }
                     endif; //end if prw is activated
@@ -1484,7 +1484,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        if (in_array('card_disclosure', $activeCards)) {
+                        if (in_array('card_disclosure', $hiddenCards)) {
                             echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
                         }
                     endif; // end if disclosures authorized
@@ -1514,7 +1514,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        if (in_array('card_amendments', $activeCards)) {
+                        if (in_array('card_amendments', $hiddenCards)) {
                             echo $twig->getTwig()->render('patient/card/amendments.html.twig', $viewArgs);
                         }
                     endif; // end amendments authorized
@@ -1544,7 +1544,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        if (in_array('card_lab', $activeCards)) {
+                        if (in_array('card_lab', $hiddenCards)) {
                             echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
                         }
                     endif; // end labs authorized
@@ -1569,7 +1569,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        if (in_array('card_vitals', $activeCards)) {
+                        if (in_array('card_vitals', $hiddenCards)) {
                              echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
                         }
                     endif; // end vitals
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -74,6 +74,7 @@ if (isset($_GET['set_pid'])) {
 // want smart support in their system.
 $smartLaunchController = new SMARTLaunchController($GLOBALS["kernel"]->getEventDispatcher());
 $smartLaunchController->registerContextEvents();
+$activeCards = getActiveDashboardCards();
 
 /**
  * @var EventDispatcher
@@ -1079,7 +1080,9 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=allergy')"
                         ];
                         echo "<div class=\"$col\">";
-                        echo $t->render('patient/card/allergies.html.twig', $viewArgs);
+                        if (in_array('card_allergies', $activeCards)) {
+                            echo $t->render('patient/card/allergies.html.twig', $viewArgs);
+                        }
                         echo "</div>";
                     }
 
@@ -1103,7 +1106,9 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=medical_problem')"
                         ];
                         echo "<div class=\"$col\">";
+                    if (in_array('card_medicalproblems', $activeCards)) {
                         echo $t->render('patient/card/medical_problems.html.twig', $viewArgs);
+                    }
                         echo "</div>";
                     }
 
@@ -1125,7 +1130,9 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=medication')"
                         ];
                         echo "<div class=\"$col\">";
+                    if (in_array('card_medication', $activeCards)) {
                         echo $t->render('patient/card/medication.html.twig', $viewArgs);
+                    }
                         echo "</div>";
                     }
 
@@ -1152,8 +1159,7 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                                 'auth' => false,
                                 'rxList' => $rxArr,
                             ];
-
-                            echo $t->render('patient/card/erx.html.twig', $viewArgs);
+                                echo $t->render('patient/card/erx.html.twig', $viewArgs);
                         }
 
                         $id = "prescriptions_ps_expand";
@@ -1455,7 +1461,10 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_patientreminders', $activeCards)) {
+                            echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        }
+
                     endif; //end if prw is activated
 
                     if (AclMain::aclCheckCore('patients', 'disclosure')) :
@@ -1476,7 +1485,9 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_disclosure', $activeCards)) {
+                            echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        }
                     endif; // end if disclosures authorized
 
                     if ($GLOBALS['amendments'] && AclMain::aclCheckCore('patients', 'amendment')) :
@@ -1504,7 +1515,10 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/amendments.html.twig', $viewArgs);
+                        if (in_array('card_amendments', $activeCards)) {
+                            echo $twig->getTwig()->render('patient/card/amendments.html.twig', $viewArgs);
+                        }
+
                     endif; // end amendments authorized
 
                     if (AclMain::aclCheckCore('patients', 'lab')) :
@@ -1532,7 +1546,9 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_lab', $activeCards)) {
+                            echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        }
                     endif; // end labs authorized
 
                     if ($vitals_is_registered && AclMain::aclCheckCore('patients', 'med')) :
@@ -1555,7 +1571,9 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_vitals', $activeCards)) {
+                             echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        }
                     endif; // end vitals
 
                     // if anyone wants to render anything after the patient demographic list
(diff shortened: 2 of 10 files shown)


sql/7_0_1-to-7_0_2_upgrade.sql
```diff
@@ -614,3 +614,16 @@ CREATE TABLE `fee_schedule` (
     UNIQUE KEY `ins_plan_code_mod_type_date` (`insurance_company_id`, `plan`, `code`, `modifier`, `type`, `effective_date`)
 ) ENGINE=InnoDb AUTO_INCREMENT=1;
 #EndIf
+
```
stephenwaite, line 617: hi @juggernautsei, no need for this file

sql/7_0_2-to-7_0_3_upgrade.sql
```diff
@@ -113,3 +113,15 @@
 ALTER TABLE `form_encounter` ADD `last_update` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP;
 #EndIf
 
+#IfNotRow2D list_options list_id lists option_id default_open_tabs
+-- Create new list Default Open Tabs
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`) VALUES ('lists', 'Hide_Dashboard_Cards', 'Hide Dashboard Cards');
```
stephenwaite, line 118: above should be looking for option id `Hide_Dashboard_Cards`

sql/7_0_1-to-7_0_2_upgrade.sql
juggernautsei: That is not a part of my PR.

stephenwaite:
Should this take into account the ACL of the user so that way any medical info could not be put out of reach of a physician or clinician?

bradymiller:
Pretty neat way to turn certain things off for a practice.
- Are all the cards given this functionality or just some of them? If not, would support for all the current cards
- In the list editor, would give this list a little customization and would change the Table Header (in list editor gui) for `Active` to `Hide` and then have it hidden when active setting is 1.

juggernautsei:
This only hide certain cards. I don't understand the second part of your question but the list can be added to. I did the ones that I am most often asked to remove. The core dashboard cards I left. These are the ones that everyone uses. 
I think I know what you are saying for the second. Change the list editor to display Hide instead of Active for this list to match what is being done. I will dive into that and see if I can figure that one out.

juggernautsei:
> Should this take into account the ACL of the user so that way any medical info could not be put out of reach of a physician or clinician?

@stephenwaite I understand your concern but the clinic that does this is usually a none primary care. If it is a primary care clinic or a clinic that needs those cards. Why would they turn them off? I deal with behavioral health a lot and most of them do not need or use those cards. But the psychiatrist that prescribes. We leave the prescription card on for them and the medication card says on. The ACL component can be added so that no matter way a provider would still see the cards. But kind of defeats the purpose. The person that turns these cards off would need admin privileges. It is hard to prevent malicious or ignorant actors in a clinic.

juggernautsei:
Bump. Can we get this into the next patch?

interface/patient_file/summary/demographics.php
```diff
@@ -74,6 +74,7 @@
 // want smart support in their system.
 $smartLaunchController = new SMARTLaunchController($GLOBALS["kernel"]->getEventDispatcher());
 $smartLaunchController->registerContextEvents();
+$activeCards = getActiveDashboardCards();
```
bradymiller, line 77:
```
$hiddenCards = getHiddenDashboardCards();
```

library/lists.inc.php
```diff
@@ -142,3 +142,13 @@ function setListTouch($patient_id, $type)
         sqlStatement("INSERT INTO `lists_touch` ( `pid`,`type`,`date` ) VALUES ( ?, ?, NOW() )", array($patient_id,$type));
     }
 }
+
+function getActiveDashboardCards()
```
bradymiller, line 146:
```
function getHiddenDashboardCards()
```

sql/7_0_1-to-7_0_2_upgrade.sql
bradymiller: drop the changes in this script

sql/7_0_2-to-7_0_3_upgrade.sql
```diff
@@ -113,3 +113,15 @@
 ALTER TABLE `form_encounter` ADD `last_update` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP;
 #EndIf
 
+#IfNotRow2D list_options list_id lists option_id Hide_Dashboard_Cards
+-- Create new list Hide Dashboard Cards
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`) VALUES ('lists', 'Hide_Dashboard_Cards', 'Hide Dashboard Cards');
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`,`activity`) VALUES ('Hide_Dashboard_Cards', 'card_allergies', 'Allergies', 10, 1);
```
bradymiller, line 119: set all the active settings from 1 to 0

sql/database.sql
```diff
@@ -13638,3 +13638,13 @@ INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'YE', 'Yemeni', '2480', '0', '0', '', 'Yemen', 'YEM:887');
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'ZM', 'Zambian', '2490', '0', '0', '', 'Zambia', 'ZMB:894');
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'ZW', 'Zimbabwean', '2500', '0', '0', '', 'Zimbabwe', 'ZWE:716');
+
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`) VALUES ('lists', 'Hide_Dashboard_Cards', 'Hide Dashboard Cards');
```
bradymiller, line 13642: set all the active settings from 1 to 0

bradymiller:
- Since the point is to Hide cards, see my above recommendations (i think this is much more clear since the list is for hiding cards and the list does not include every single possible card).
- Then need to modify the column name for the hide list in the list editor from Active to Hide (can see other examples in that list editor script where column names are changed dependent on the list). This then makes the user turn the toggle on to hide it (note the database active will then be set to 1 when the user wants to hide the card).

Documentation/help_files/medical_dashboard_help.php
```diff
@@ -162,6 +162,11 @@
             <div class= "row" id="section3">
                 <div class="col-sm-12">
                     <h4 class="oe-help-heading"><?php echo xlt("Data Section"); ?><a href="#"><i class="fa fa-arrow-circle-up oe-pull-away oe-help-redirect" aria-hidden="true"></i></a></h4>
+
+                    <p><?php echo xlt("The some of the cards on the dashboard can be turned off if not being utilized by the clinic");?></p>
```
bradymiller, line 166: Change the 'The some' to 'Some'

bradymiller:
Note the above proposed strategy will also then make it easier to support adding tthe other cards to this mechanism without hiding them by mistake. For example, can place the conditionals around the other cards, and they will still show by default even if they are not yet listed in the Hide list.

bradymiller:
review complete :)

stephenwaite:
hi @juggernautsei , check out https://github.com/openemr/openemr/blob/master/interface/super/edit_list.php#L1251 for where you can change active to hide

sql/7_0_1-to-7_0_2_upgrade.sql
stephenwaite: to remove this from the PR just copy the [raw file](https://raw.githubusercontent.com/openemr/openemr/master/sql/7_0_1-to-7_0_2_upgrade.sql) into the file in this branch

interface/patient_file/summary/demographics.php
```diff
@@ -1079,7 +1080,9 @@ function filterActiveIssues(array $i): array
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=allergy')"
                         ];
                         echo "<div class=\"$col\">";
-                        echo $t->render('patient/card/allergies.html.twig', $viewArgs);
+                        if (in_array('card_allergies', $hiddenCards)) {
```
stephenwaite, line 1083: these should be `if (!in_array`

sql/7_0_2-to-7_0_3_upgrade.sql
juggernautsei: Is this what it should be
#IfNotRow lists option_id Hide_Dashboard_Cards

sjpadgett:
> A front desk member would not need to see medications while an MA would. It may be worth pursuing. Good thought.

The card themselves have acl's so should not need in list options.
Also now'd be a good time to add lists by user id! Maybe a new table that references the list by list id and user id!
You'd be doing the project a huge favor if you have time or inclination Sherwin. Then could be used in start up tabs and several future features that you may even run across. I'd help when I have a minute.

surfacehazard:
The ability to hide cards would provide significant utility to my clinics. User-specific lists, while a worthy enhancement, represents a large expansion of the bike shed. Suggest to merge #7192 first..

interface/patient_file/summary/demographics.php
```diff
@@ -1125,7 +1130,9 @@ function filterActiveIssues(array $i): array
                             'btnLink' => "return load_location('{$GLOBALS['webroot']}/interface/patient_file/summary/stats_full.php?active=all&category=medication')"
                         ];
                         echo "<div class=\"$col\">";
-                        echo $t->render('patient/card/medication.html.twig', $viewArgs);
+                        if (in_array('card_medication', $hiddenCards)) {
```
stephenwaite, line 1133: also needs `!in_array`

interface/patient_file/summary/demographics.php
```diff
@@ -1455,7 +1461,9 @@ function filterActiveIssues(array $i): array
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_patientreminders', $hiddenCards)) {
```
stephenwaite, line 1464: also needs `!in_array`

interface/patient_file/summary/demographics.php
```diff
@@ -1476,7 +1484,9 @@ function filterActiveIssues(array $i): array
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_disclosure', $hiddenCards)) {
```
stephenwaite, line 1487: also needs `!in_array`

interface/patient_file/summary/demographics.php
```diff
@@ -1504,7 +1514,9 @@ function filterActiveIssues(array $i): array
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/amendments.html.twig', $viewArgs);
+                        if (in_array('card_amendments', $hiddenCards)) {
```
stephenwaite, line 1517: also needs `!in_array`

interface/patient_file/summary/demographics.php
```diff
@@ -1532,7 +1544,9 @@ function filterActiveIssues(array $i): array
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_lab', $hiddenCards)) {
```
stephenwaite, line 1547: also needs `!in_array`

interface/patient_file/summary/demographics.php
```diff
@@ -1555,7 +1569,9 @@ function filterActiveIssues(array $i): array
                             'prependedInjection' => $dispatchResult->getPrependedInjection(),
                             'appendedInjection' => $dispatchResult->getAppendedInjection(),
                         ];
-                        echo $twig->getTwig()->render('patient/card/loader.html.twig', $viewArgs);
+                        if (in_array('card_vitals', $hiddenCards)) {
```
stephenwaite, line 1572: also needs `!in_array`

sjpadgett:
> The ability to hide cards would provide significant utility to my clinics

You will be in a minority as most practice disciplines require these cards. Forcing an administrator of a practice group to decide whether everyone or noone gets the feature will just make this feature only useable to a minority of users as most admins will opt to leave on IMO. This defeats the whole purpose of having user settings. 

If I were reviewing, I wouldn't allow in because for one It's another list that is user corruptible and if it was it would have a significant impact on users. I'd insist on a hard coded multiselect list of cards in a global where it could be then used in user settings.
I don't care if Sherwin makes this adaptable or not but as I say, I'm opposed to the feature as is.

juggernautsei:
> You will be in a minority as most practice disciplines require these cards.

@sjpadgett in my experience, this program has grown well beyond the use of primary care physicians. This past week, we picked up a project from an injection company. We  hide cards a lot and just thought it was time to stop carrying this by ourselves. We have billers that use the program. No need for the primary care view. @sjpadgett this baby has matured.

sjpadgett:
Bullcrap! Primary Care, Occupational Health, Outpatient, Home Health, LPN and RNs reviewing chart for before visits. For people using only for billing, get a billing program!
We are growing to the enterprise and don't want to specialized to the single practitioner as in the past.
As far as I'm concerned this should be a hard coded multiselect and be useful to the entire community, period!

bradymiller:
hi all, just my 2 cents.
- To point out, this PR was changed to have no impact if it is corrupted (part of my review recommendations were to flip flop things so when set to active it hides the card; this means if the list does not exist or is corrupted it will show all the cards anyways). This will open door to a clinic turning off any card they wish for the entire practice while not requiring tracking a master card list, which i think would be useful (maybe things like Disclosures, Amendments, Recall for some clinics and some clinics may just need bare minimum for example a health spa like clinic).
- Per user settings are already mostly supported via the acls which would use (and expand on if needed) rather than this mechanism for the per user or usergroup settings (I think @robertdown was planning something like this down the road to better customize the summary screen for different groups of users; and if he did decide to leverage the mechanism in this PR, then he could migrate it from the list_options to another more specific table along with a gui interface to manage it). Would be interested in hearing from @robertdown if this kind of mechanism would jibe with what he was thinking of doing in the future.

sjpadgett:
I don't see how you'd do this with ACL!
Different users in same practice can have different chart requirements. Same goes for default tabs on logon.
This just isn't that hard to do then it's finished once and for all. This PR could only require touching 2 files, demographics and globals. So much easier patching and upgrading.

bradymiller:
I was also actually thinking about the default tabs on logon since is also
kind of related to this :)
I think there is a difference however:
1. It is ok for users to set the defaults tabs on logon, although the admin
will get to choose which tabs they can choose from (this is a pending bug
fix)
2. Users should not be allowed to choose what they see in the patient
summary screen since it allows a user to break ONC 2015 and also mistakenly
turn off critical things like allergies. User-specific or group-specific
settings should be set by the admin. Another way to think about this is how
the custom menus are managed. The admin can create custom menus and
assign them to users, but the user has no say in this (again, for the same
reason why wouldn't let them touch what they see in the patient summary
screen).

Regarding ACL, it is already doing group-specific settings to a certain
degree on the patient summary cards (ie. receptionist is blocked from
certain cards etc).

sjpadgett:
Well if turning off certain cards breaks certification then letting anyone, even an admin the capability, then it would still break certification. Besides if the user doesn't have proper permissions then the ACL wouldn't allow.
This is just circle logic and my bottom line has been this method of implementing is unnecessary plus adding a corruptible list regardless if the default is on, still could breaks the feature.
A multiselect list is just better engineering in my opinion plus easier to clean out if we get around to implementing context and scopes to app startup config.
Not my review so do as you feel best.

juggernautsei:
@sjpadgett I don't understand your veracity about a corruptible list. I model what I did after the list was put in for the default tabs. If you can have a list for the default tabs, why not for the cards on the dashboard. Your stance is a little ingeminating but not clear as to why your are making this stance when there are some many other lists in the code base. We make lists all over the place. None of those lists have be come corrupt. Why would this list all of a sudden on its own be the one to be corrupt?

sjpadgett:
Most all lists can be corrupted simply by renaming ids, codes or titles etc. This includes tabs where a list wouldn't have been my choice because a multiselect in globals would allow user settings easier.

I've found that lists can get you in trouble if where used requires a list item to be an expected value. In openemr it seems this is an acceptable risk. What we need for lists is a way to lock developer required fields from edits if we are going to use list as a app setting especially hard coded needs! Probably a feature I'll have to do but I believe we may already have examples of this in code now. Unsure!

I could counter argue this for days but the best course for not using developer only lists to enable settings are multiselect! Any additions to these list will be done by developer through new or upgrades so shouldn't be exposed to misuse.

If you don't want to take my advice fine but adding this pattern just helps give precedence for IMO, continued misuse of lists.

juggernautsei:
I understand the misuse of lists because I have had to deal with people changing lists that have affected the operation of the program. But then, IMO the config is not a good place to put the controls because it is not as clean as the list and would require more engineering to make it work. You suggested a multi select dropdown as a place to start. Let me see what I can do. What section would that go under? Appearance or Features or Patient

sjpadgett:
> Appearance or Features or Patient

I'd think Appearance. I said I'd help but you should be able to reuse/model lang multi select or select with multiple attr. LBF options has a multi thay may help for ideas.
I just got a weno api for credential validation I need to get finished but hope I can look into this late with you tomorrow...

juggernautsei:
@sjpadgett thanks for offering to help but I am not lost in the wood. I got this.

sjpadgett:
sorry didn't mean to offend.

bradymiller:
Added UpForGrabs demo for testing this PR at: https://www.open-emr.org/wiki/index.php/Development_Demo#Alpha_-_Up_For_Grabs_Demo

juggernautsei:
It is testing well for me in the up for grabs

interface/super/edit_globals.php
```diff
@@ -645,6 +645,22 @@ function checkBackgroundServices()
                                                     echo "</option>\n";
                                                 }
                                                 echo "  </select>\n";
+                                            } elseif ($fldtype == GlobalSetting::DATA_TYPE_MULTI_DASHBOARD_CARDS) {
+                                                $res = sqlStatement("SELECT * FROM dashboard_card_visibility");
+                                                echo "  <select multiple class='form-control' name='form_{$i}[]' id='form_{$i}[]' size='3'>\n";
+                                                while ($row = sqlFetchArray($res)) {
+                                                    echo "   <option value='" . attr($row['card_abrev']) . "'";
```
sjpadgett, line 652: We need a default none or empty so user can select to hide 0 cards. Currently once you select cards you will always have to select at least one card.
`<option value="">attr("Hide None")</option>`
Or something like after select echo.

sjpadgett: actually I suppose a ctrl click will work so maybe okay the way you have it.
Thanks for the effort Sherwin.

bradymiller:
hi @juggernautsei , This is definitely headed in the right direction with the multiselect global.
1. Why not just now have a simple list in the list_options called 'dashboard_cards' that is used to populate the global selector (rather than creating a new table just to store and id and title, which list_options already does nicely and then have a ui where can add new ones when needed). Creating new tables for simple lists will get tedious and note list_options is made for this use (can add new elements via ui, have control of the titles, and in database so can be different is using multisite function).
2. Can populate your multi global like do for 'language_menu_other' at https://github.com/openemr/openemr/blob/master/interface/globals.php#L382. This will then allow to create a simple isHiddenDashboardCard('cardname') in the 'if' calls to see if it is hidden.

sjpadgett:
Just hardcode the multilist. It will not need to be dynamic

bradymiller:
hi @juggernautsei and @sjpadgett ,
Agree with hardcoding the list as a php array instead of in ist_options.

juggernautsei:
Hi @bradymiller can this be resolved and brought into the code base

stephenwaite:
hi @juggernautsei , looks like you have that conflict with the upgrade script?

juggernautsei:
Ok I will pull it back and take a look at it. I was being lazy. Thx @stephenwaite

Documentation/help_files/medical_dashboard_help.php
```diff
@@ -162,6 +162,11 @@
             <div class= "row" id="section3">
                 <div class="col-sm-12">
                     <h4 class="oe-help-heading"><?php echo xlt("Data Section"); ?><a href="#"><i class="fa fa-arrow-circle-up oe-pull-away oe-help-redirect" aria-hidden="true"></i></a></h4>
+
+                    <p><?php echo xlt("Some of the cards on the dashboard can be turned off if not being utilized by the clinic");?></p>
+
+                    <p><?php echo xlt("To see which cards can be hidden, go to Admin, Forms, Lists, then search for Hide Dashboard Cards");?>.</p>
```
adunsulag, line 168: This description is no longer accurate as it looks like settings are now done in the globals configuration.

sql/7_0_1-to-7_0_2_upgrade.sql
adunsulag: This still needs to happen.

sql/7_0_2-to-7_0_3_upgrade.sql
```diff
@@ -112,4 +112,8 @@
 #IfMissingColumn form_encounter last_update
 ALTER TABLE `form_encounter` ADD `last_update` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP;
 #EndIf
+#IfNotRow4D supported_external_dataloads load_type CQM_VALUESET load_source NIH_VSAC load_release_date 2023-05-04 load_filename ec_only_cms_20230504.xml.zip
+INSERT INTO `supported_external_dataloads` (`load_type`, `load_source`, `load_release_date`, `load_filename`, `load_checksum`) VALUES
```
adunsulag, line 116: This shouldn't be here, not sure how it even made it into this PR.

sql/database.sql
adunsulag: Revert this file as its no longer being used due to the globals setting.

adunsulag (changes requested):
Looks like there's some cleanup that needs to occur with all of the changes.  Unless that happens today, we'll need to bump this to the next patch.

sql/7_0_1-to-7_0_2_upgrade.sql
juggernautsei: @adunsulag 
@stephenwaite removed this. It was not apart of my PR.

sql/7_0_2-to-7_0_3_upgrade.sql
juggernautsei: This is one of those GitHub things. Not is my PR.

sql/database.sql
juggernautsei: This line is gone. I can't find it in the PR. I just looked.

juggernautsei:
Everything that @adunsulag addresses has been corrected and moved. There is nothing for me to do here. @sjpadgett

sjpadgett:
Okay thanks @juggernautsei . Let me test and review again and i'll let you know.

interface/super/edit_globals.php
```diff
@@ -645,6 +645,22 @@ function checkBackgroundServices()
                                                     echo "</option>\n";
                                                 }
                                                 echo "  </select>\n";
+                                            } elseif ($fldtype == GlobalSetting::DATA_TYPE_MULTI_DASHBOARD_CARDS) {
+                                                $res = sqlStatement("SELECT * FROM dashboard_card_visibility");
```
sjpadgett, line 649: You were supposed to get rid of table and hard code these options!

Documentation/help_files/medical_dashboard_help.php
sjpadgett: Address this!

sql/7_0_1-to-7_0_2_upgrade.sql
sjpadgett: Still here. You can't simply remove, you have to commit the correct version.
You really need to pull and merge master in your local then push.

You know, better yet @juggernautsei, close this and push new PR with correct files for the feature option. I'm not sure what happened to this PR but it is a mess. For sure get rid of new table and list options so really no need to push any database/upgrade changes.
Since now in globals with hard coded select options I foresee only two or three files for feature.
This IMO is easiest way to get this in....

interface/super/edit_globals.php
juggernautsei: @sjpadgett eye eye captain

sql/7_0_1-to-7_0_2_upgrade.sql
sjpadgett: @juggernautsei Did you see this above?

interface/super/edit_globals.php
```diff
@@ -645,22 +645,6 @@ function checkBackgroundServices()
...
-                                                        if ($glrow['gl_value'] == $row['card_abrev']) {
-                                                            echo " selected";
-                                                            break;
-                                                        }
-                                                    }
-                                                    echo ">";
-                                                    echo xlt($row['card_name']);
-                                                    echo "</option>\n";
-                                                }
-                                                echo "  </select>\n";
```
sjpadgett: @juggernautsei I don't understand what is going here. This is the preferred way you strove for! Just needed to hard code the options that current are in the new table then  revert table.
Did you accidently push an older version?

juggernautsei: Accident.

juggernautsei: I got the code and will start a new PR for clarity sake.

juggernautsei:
Going to start a new PR.

PR #7195, opened 2024-01-26
Global user settings not set

### LINKED ISSUE #7194: User settings not set if not in globals.
### Describe the problem
Currently user settings must match a setting value set in globals to be included in the $GLOBAL super. If a module wants to set a needed setting that is user specific and does not want a default from global setting then the setting is never set in globals.php. E.g. Module wants have a global admin value for one reason but doesn't want that value to be valid for a specific user.
I propose to remove the limitation

sjpadgett:
add routine to set uninitialized user settings not in globals.

Fixes #7194

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -434,6 +434,13 @@ if (!empty($glrow)) {
             $GLOBALS[$gl_name] = $gl_value;
         }
     }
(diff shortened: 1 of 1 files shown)


sjpadgett:
I'm going to merge quickly as it is needed for module development currently under way!

PR #7198, opened 2024-01-30
Weno bug fixes

sjpadgett:
- rename class file because of case difference. windows doesn't care but linux does

weird how I had to rename for git prob a windows thing.

### DIFF
--- a/interface/modules/custom_modules/oe-module-weno/templates/facilities.php
+++ b/interface/modules/custom_modules/oe-module-weno/templates/facilities.php
@@ -73,10 +73,13 @@ $pharm_log = $logService->getLastPharmacyDownloadStatus();
         }
         
         function downloadPharmacies(){
+            if (!window.confirm(jsText("This download may take several minutes. Do you want to continue?"))) {
+                return false;
+            }
             $('#notch-pharm').removeClass("hide");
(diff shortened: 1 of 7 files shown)


sjpadgett:
@dataengr my repo

sjpadgett:
Runs on docker easy but this module just seems to need improvements in UX. Merging

interface/modules/custom_modules/oe-module-weno/templates/facilities.php
```diff
@@ -73,10 +73,13 @@ function activateFacility(){
         }
         
         function downloadPharmacies(){
+            if (!window.confirm(jsText("This download may take several minutes. Do you want to continue?"))) {
```
bradymiller, line 76: no need to escape literal that goes to dialog in js context. Would translate it via the javascript xl() function though.

bradymiller:
2DV

interface/modules/custom_modules/oe-module-weno/templates/facilities.php
sjpadgett: oops you're right and I've forgotten about the neat translate js function. We could start using it more in twig I'd think!

PR #7204, opened 2024-02-02
Weno fixes and updates

sjpadgett:
- Improve pharmacy import by using transactional queries. From 4 minutes to 40 seconds.
- improve validation and error reporting.
- some code cleanup for warnings and in general
- fix frame screen output to use the current theme so user always see content.

Fixes #

### DIFF
--- a/interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
+++ b/interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
@@ -13,9 +13,9 @@ use OpenEMR\Modules\WenoModule\Services\WenoLogService;
 use OpenEMR\Modules\WenoModule\Services\PharmacyService;
 
 $cryptoGen = new CryptoGen();
-$weno_username = $GLOBALS['weno_admin_username'];
-$weno_password = $cryptoGen->decryptStandard($GLOBALS['weno_admin_password']);
-$encryption_key = $cryptoGen->decryptStandard($GLOBALS['weno_encryption_key']);
+$weno_username = $GLOBALS['weno_admin_username'] ?? '';
+$weno_password = $cryptoGen->decryptStandard($GLOBALS['weno_admin_password'] ?? '');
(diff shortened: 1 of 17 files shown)


interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
```diff
@@ -105,87 +105,98 @@ function download_zipfile($fileUrl, $zipped_file)
...
-
-            ++$l;
         }
+    } catch (Exception $e) {
+        throw new RuntimeException($e->getMessage(), $e->getCode(), $e);
     }
     $wenolog->insertWenoLog("pharmacy", "Success");
     fclose($records);
+    sqlStatementNoLog('COMMIT');
+    sqlStatementNoLog('SET autocommit=1');
```
sjpadgett, line 198: I wonder if I should be concerned with default settings for PHP and engine resources with approx 100, 000 inserts?
Perhaps safer to batch maybe every 10,000 or so!

sjpadgett:
That's it for this go around. Thanks SW for warning fix

stephenwaite:
weno bueno :)
arlo-hello

interface/modules/custom_modules/oe-module-weno/src/Services/WenoPharmaciesImport.php
```diff
@@ -27,94 +27,99 @@ public function importPharmacy($csvFile, $files)
...
-                    ++$l;
+                        ++$l;
+                    }
                 }
+            } catch (Exception $e) {
+                return $e->getMessage();
             }
             fclose($records);
+            sqlStatementNoLog('COMMIT');
+            sqlStatementNoLog('SET autocommit=1');
```
bradymiller, line 122: Doesn't the following need to be placed above:
```
            sqlStatementNoLog('SET autocommit=0');
            sqlStatementNoLog('START TRANSACTION');
```

sjpadgett: Nice catch @bradymiller and thank you!

PR #7208, opened 2024-02-04
Weno module changes

sjpadgett:
Fixes #

Weno module fixup.

- fix transaction for pharmacy download
- refactor user default fac to include select from user facility when user is not logged into facility.
- convert error to interface style helper to static
- create static styler for echo to interface.
- refactor all show errors to include a style compatible with themes.


bradymiller:
dance-bear

PR #7215, opened 2024-02-11
Fixes #7214 pass action status to listener

### LINKED ISSUE #7214: bug: InstallerController needs to pass current status to ModuleManagerAfterActionListener action methods
If the module action fails the ModuleManagerAfterActionListener needs to be aware that the system is already in a failure state.  We need to pass the current action status into InstallerController.  It looks like this was the original intent but was likely overlooked.

This is in InstallerController.php#L216

adunsulag:
Action status was not being passed in the listener. Fixes #7214

### DIFF
--- a/interface/modules/zend_modules/module/Installer/src/Installer/Controller/InstallerController.php
+++ b/interface/modules/zend_modules/module/Installer/src/Installer/Controller/InstallerController.php
@@ -213,7 +213,7 @@ class InstallerController extends AbstractActionController
         $instance = $className::initListenerSelf();
         if (class_exists($instance::class)) {
             if (method_exists($instance, 'moduleManagerAction')) {
-                return call_user_func([$instance, 'moduleManagerAction'], $methodName, $modId);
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/module/Installer/src/Installer/Controller/InstallerController.php
```diff
@@ -213,7 +213,7 @@ private function callModuleAfterAction($action, $modId, $dirModule, $currentStat
         $instance = $className::initListenerSelf();
         if (class_exists($instance::class)) {
             if (method_exists($instance, 'moduleManagerAction')) {
-                return call_user_func([$instance, 'moduleManagerAction'], $methodName, $modId);
+                return call_user_func([$instance, 'moduleManagerAction'], $methodName, $modId, $currentStatus);
```
sjpadgett, line 216: oops! good catch.

sjpadgett (approved):

bradymiller:
2DV

Issue #7217, opened 2024-02-11
feat: Insurance Company Service allow setting id and make phone settings optional


adunsulag:
When importing / transferring data into OpenEMR it is helpful to be able to set the specific id of the insurance company in the service class.  Also, if there are no phone settings in the data we should skip creating phone information as part of the service.

Lastly, if the service fails to update / insert we should have it throw an exception instead of silently fail.

PR #7225, opened 2024-02-13
Fixes #7224, #7217 insurance company changes

### LINKED ISSUE #7224: bug: Practice Settings Insurance Company list extremely slow to load with hundreds of insurance companies
The Practice Settings Insurance Company list takes an extremely long time to load if there are hundreds of records in the data table list.  In the system I was testing it had almost 900 insurance companies in the list.  This took almost a full minute to load up the list.  We need to change up the way this page is built in order for it to load faster.  Since the js datatables loads all of the data up front we either need to rewrite the front end or optimize how the data is loaded on the backend to make this faster.

adunsulag:
Fixed the list loading taking forever on the insurance company list. This fixes #7224.

Had a bug in the fix for #7217 so this fixes that as well. Wasn't actually replacing the insurance id.

### DIFF
--- a/controllers/C_InsuranceCompany.class.php
+++ b/controllers/C_InsuranceCompany.class.php
@@ -1,5 +1,8 @@
 <?php
 
+use OpenEMR\Common\Twig\TwigContainer;
+use OpenEMR\Services\InsuranceCompanyService;
+
 class C_InsuranceCompany extends Controller
 {
     var $template_mod;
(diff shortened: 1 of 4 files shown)


library/classes/InsuranceCompany.class.php
```diff
@@ -330,19 +334,29 @@ public function persist()
...
-        //print_r($results);
-        while ($row = sqlFetchArray($results)) {
-                $icompanies[] = new InsuranceCompany($row['id']);
+
+        $listAll = $insuranceCompanyService->search([]);
+        if ($listAll->hasData()) {
+            $data = $listAll->getData();
+            foreach ($data as $record) {
+                // we pass in the service array so we don't recreate it each time
+                $company = new InsuranceCompany("", "", $insuranceCompanyService);
```
stephenwaite, line 345: seal-dunk

stephenwaite (approved):

controllers/C_InsuranceCompany.class.php
```diff
@@ -42,10 +45,39 @@ public function edit_action($id = "", $patient_id = "", $p_obj = null)
...
+                return strcasecmp($a['name'], $b['name']);
+            });
+        }
+        $templateVars = [
+            'CURRENT_ACTION' => $GLOBALS['webroot'] . "/controller.php?" . "practice_settings&insurance_company&"
+            ,'icompanies' => $iCompanies
+        ];
 
-        return $this->fetch($GLOBALS['template_dir'] . "insurance_companies/" . $this->template_mod . "_list.html");
+        return $twig->getTwig()->render('insurance_companies/general_list.html.twig', $templateVars);
```
bradymiller, line 80: kung-fu-funny

templates/insurance_companies/general_list.html.twig
```diff
@@ -0,0 +1,62 @@
+<html>
+    <head>
+        {{ setupHeader(['common','datatables','datatables-colreorder','datatables-dt','datatables-bs']) }}
+        <title>{{ "Insurance Companies"|xlt }} {{ applicationTitle|text }}</title>
+    </head>
+<body>
+<a href="{{webroot|attr_url}}/controller.php?practice_settings&insurance_company&action=edit" onclick="top.restoreSession()" class="btn btn-secondary btn-add">{{'Add a Company'|xlt}}</a>
```
bradymiller, line 7: only need to do attr on the webroot. the url may cause weirdness (not really sure, but not something have done in past, since generally just use url to encode the parameters to allow special things like & and ? without breaking the parameters)

bradymiller (approved):

PR #7227, opened 2024-02-13
Fixes #7226 Insurance company address query

### LINKED ISSUE #7226: bug: Insurance Company Service does not load records that have no address entry
If an insurance company is created / imported that has no city or state then a corresponding address record is not created.  However, when we retrieve the record there is a JOIN clause on the address table which will exclude insurance company records inadvertently. 

We need to change this so that records are still returned even if there are no created address records.  This functions similar to the work and fax phone numbers.

adunsulag:
Make it so that we still return records even if there is no corresponding address record.  This happens because city and street if they are empty forces an address record to be skipped in creation.

Fixes #7226


stephenwaite (approved):
southpark_cheering

bradymiller (approved):

PR #7253, opened 2024-02-29
Dispatching an event here allows custom modules to modify the C-CDA 

### LINKED ISSUE #7252: Event for C-CDA generration
### Describe the problem

There is a need for organizations to extend the C-CDA files that are generated by the ccdaservice.  This could be to add information from custom forms or to restructure the document for compliance. 

A code change as recommended by @adunsulag [in the community forums](https://community.open-emr.org/t/seeking-advice-on-plugin-to-modify-ccda/21923/2?u=dennisonwilliams1312) is forthcoming.

DennisonWilliams:
Dispatching an event here allows custom modules to modify the C-CDA before it is saved

Fixes #7252 

This change allows authors of custom modules to change a C-CDA before it is saved to file. 

Dispatch a service event before the C-CDA is saved by the documents subsystem

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
@@ -3391,13 +3391,10 @@ class EncounterccdadispatchTable extends AbstractTableGateway
 
         $mimeType = "text/xml";
 
-	// Let listeners know we are about to save a C-CDA
-	$send = array("CCDA" => $content, 'pid' => $pid);
-	$event = new ServiceSaveEvent($patientService, $send);
-	$updatedPreSaveEvent = $this->getEventDispatcher()->dispatch(
-          $event,
-          ServiceSaveEvent::EVENT_PRE_SAVE
-	);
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -3378,6 +3391,19 @@ public function logCCDA($pid, $encounter, $content, $time, $status, $user_id, $d
 
         $mimeType = "text/xml";
 
+        // Let listeners know we are about to save a C-CDA
+        $send = array("CCDA" => $content, 'pid' => $pid);
+        $event = new ServiceSaveEvent($patientService, $send);
```
adunsulag, line 3396: I recommend using the CDADocumentService in your ServiceSaveEvent.  The PatientService has a number of listeners in it and those listeners will have a problem if it receives a non patient record.

If you use the CDADocumentService and your $send represents a ccda table record entry then just changing the CCDA column to be **ccda_content** should make it so everything plays nice in the event system.

adunsulag (changes requested):
Looked it over, everything looks fine from what I can see.  I haven't pulled the code down to test it yet.

PR #7302, opened 2024-03-28
Resolve #7300: Resolve bug accidentally introduced in API

### LINKED ISSUE #7300: Adding calendar event (appointment) via API causes endTime to be set to 0:00:00
### Describe the bug

When you add a calendar event via `/api/patient/{pid}/appointment` regardless of which duration you give the endTime will always be set to midnight or 0:00:00.

### To Reproduce
Do a post request to said API endpoint

1. Send POST request
2. Check the newly made appointment and see the endTime is set to 0:00:00

### Expected behavior

The endTime is set properly

### Client configuration

- Browser: N.A.
- OpenEMR version: 7.0.2
- Operating system: N.A.

It is caused by this line https://github.com/openemr/openemr/blob/master/src/Services/AppointmentService.php#L293 which even has the added comment of how does it even work. Which is kinda ironic since the person that added that TODO broke the actual code.

AngelsDustz:
Fixes #7300 

A bug that was accidentally introduced.

Fix bug

### DIFF
--- a/src/Services/AppointmentService.php
+++ b/src/Services/AppointmentService.php
@@ -289,8 +289,8 @@ class AppointmentService extends BaseService
     public function insert($pid, $data)
     {
         $startTime = date("H:i:s", strtotime($data['pc_startTime']));
-        // TODO: Why are we adding strings with numbers?  How is this even working
-        $endTime = $startTime . $data['pc_duration'];
+        // We need to keep this an addition rather than concatination.
+        $endTime = $startTime + $data['pc_duration'];
         $uuid = (new UuidRegistry())->createUuid();
 
         $sql  = " INSERT INTO openemr_postcalendar_events SET";
--- a/src/Services/AppointmentService.php
+++ b/src/Services/AppointmentService.php
@@ -288,9 +288,14 @@ class AppointmentService extends BaseService
 
     public function insert($pid, $data)
     {
-        $startTime = date("H:i:s", strtotime($data['pc_startTime']));
-        // We need to keep this an addition rather than concatination.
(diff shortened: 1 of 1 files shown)


adunsulag:
@AngelsDustz Have you tested your change on PHP 8.0+? I believe this was throwing a typecast error which is what led to the change. PHP 8.0+ throws typecast warnings when string and numeric (float/int) are added together. Right now it's a warning but future versions of PHP could at some point fatally fail.

I'm surprised its even working as a start date of "10:00:00" + a duration of 3600 (one hour) would result in a typecast value of 3610 inserted as the endTime from what my testing is showing.

I'll admit owning the mistake as I should have just fixed the code logic bug instead of fixing the typecast error. Reverting back to the original code however I don't see how that can be working for you.

AngelsDustz:
Hey, yes it was tested on php 8.2, which with a concatenation results in "10:00:00600" with a duration of 600. Maybe the better solution is to use https://www.php.net/manual/en/datetime.add.php.

Tho now that you mention what you did, I wonder why my version worked for us too. Does the datetime add function sound like a better way for you?

adunsulag:
I'd prefer the DateTime add function as a better way to do it, you could use the strtotime conversion already and add the integer value of duration but that doesn't take into account leap seconds/hours, etc.

AngelsDustz:
Updated it, I'd love to hear your feedback 💪 thanks for being so fast btw

adunsulag:
@AngelsDustz Looks good.  Probably could have optimized it a bit further by using DateTime to begin with instead of the strtotime but I don't see the need to speed it up further.  Once tests have passed we can merge it in.

PR #7303, opened 2024-03-28
Resolve #7301: Add title to appointment

### LINKED ISSUE #7301: feat: Title is not returned when getting list of Appointments via API.
### Describe the bug

When you get a list of all Appointments it does not return the title for said appointment making it impossible to create a proper list. The hometext however is returned.

### To Reproduce

1. Do a GET request to `/api/appointment`
2. See no title field

### Expected behavior

The title is returned for the list operation.

### Client configuration

- Browser: N.A.
- OpenEMR version: 7.0.2
- Operating system: N.A.

---
It seems it was intentionally not added? Which is kinda weird for now you need to do a request PER appointment to get the title.

AngelsDustz:
Fixes #7301 

Add title to select query

### DIFF
--- a/src/Services/AppointmentService.php
+++ b/src/Services/AppointmentService.php
@@ -263,6 +263,7 @@ class AppointmentService extends BaseService
                        pce.pc_room,
                        pce.pc_pid,
                        pce.pc_hometext,
+                       pce.pc_title,
                        f1.name as facility_name,
                        f1_map.uuid as facility_uuid,
                        f2.name as billing_location_name,


adunsulag:
@AngelsDustz Have you tested your change on PHP 8.0+?  I believe this was throwing a typecast error which is what led to the change. PHP 8.0+ throws typecast warnings when string and numeric (float/int) are added together.  Right now it's a warning but future versions of PHP could at some point fatally fail.  

I'm surprised its even working as a start date of "10:00:00" + a duration of 3600 (one hour) would result in a typecast value of 3610 inserted as the endTime from what my testing is showing.

I'll admit owning the mistake as I should have just fixed the code logic bug instead of fixing the typecast error.  Reverting back to the original code however I don't see how that can be working for you.

adunsulag (approved):
Looks good. Disregard my earlier comment as that was meant for your other PR.

PR #7306, opened 2024-03-30
Some finishing touches

sjpadgett:
Keeping my up for grab branch to save time and edits
previous review fixes.
- don't expose sensitive data
- change datepicker option. though it would work in US not so much Europe's day/m/Y


bradymiller:
2DV-2

sjpadgett:
oh I setup for dev accounts for testing

sjpadgett:
no clue what's up with tests! Will re-run tomorrow....

sjpadgett:
@juggernautsei What do ya think?

image

interface/modules/custom_modules/oe-module-weno/templates/download_log_viewer.php
```diff
@@ -120,6 +120,19 @@ function downloadPresLog(){
...
     <div class="container mt-3" id="pharmacy">
+        <?php
+        $backGroundTask = sqlStatement("SELECT `name`, `title`, `next_run` FROM `background_services` WHERE `name` LIKE ? ORDER BY `next_run` DESC", ['%weno%']);
+        // first show some download info. Why not!
+        if ($backGroundTask ?? false) {
+            echo '<h6 class="mb-2">';
+            while ($task = sqlFetchArray($backGroundTask)) {
+                $title = $task['title'];
+                $nextRun = $task['next_run'];
+                echo '<span class="mr-5 text-success">' . $title . '  ' . xlt("next run") . ': <span class="text-dark">' . $nextRun . '</span></span>';
```
bradymiller, line 131: text() $title and $nextRun

juggernautsei:
That is good placement. Thanks!

interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
```diff
@@ -8,14 +8,13 @@
...
 use OpenEMR\Modules\WenoModule\Services\PharmacyService;
-use OpenEMR\Modules\WenoModule\Services\TransmitProperties;
 use OpenEMR\Modules\WenoModule\Services\WenoLogService;
 
 $cryptoGen = new CryptoGen();
 $weno_username = $GLOBALS['weno_admin_username'] ?? '';
 $weno_password = $cryptoGen->decryptStandard($GLOBALS['weno_admin_password'] ?? '');
 $encryption_key = $cryptoGen->decryptStandard($GLOBALS['weno_encryption_key'] ?? '');
-$baseurl = "https://online.wenoexchange.com/en/EPCS/DownloadPharmacyDirectory";
+$baseurl = "https://dev.wenoexchange.com/en/EPCS/DownloadPharmacyDirectory";
```
adunsulag, line 17: Do you mean to leave the URL as a dev location?  Or will this change in the future?  Might want to have an option to swap the system from dev to production mode.

interface/modules/custom_modules/oe-module-weno/templates/synch.php
```diff
@@ -24,8 +24,11 @@
     exit;
 }
 
-$logproperties = new LogProperties();
-$result = $logproperties->logSync();
+$logProperties = new LogProperties();
+try {
+    $result = $logProperties->logSync();
+} catch (Exception $e) {
```
adunsulag, line 30: Do we need to log this exception or is it a misbehaving class?

interface/modules/custom_modules/oe-module-weno/scripts/weno_log_sync.php
```diff
@@ -13,55 +13,109 @@
...
+ * @param string $errorMessage
+ */
+function handleDownloadError(string $errorMessage)
+{
+    EventAuditLogger::instance()->newEvent(
+        "pharmacy_background",
+        $_SESSION['authUser'],
+        $_SESSION['authProvider'],
+        1,
+        text($errorMessage)
```
bradymiller, line 104: shouldn't html escape stuff going into database (sorry, just noted this)

interface/modules/custom_modules/oe-module-weno/scripts/weno_log_sync.php
```diff
@@ -13,55 +13,109 @@
...
+    );
+
+    error_log($errorMessage);
```
bradymiller, line 107: Would html escape via attr() or use the wrapper errorLogEscape() on this. I agree the road is windy to have a vulnerability from this,  however, there is no downside to doing this and we have been doing this for awhile as standard process. At the least, it will avoid dealing with reported vulnerabilities regarding this issue and will make the folks that deal with security reports lives easier.

src/Common/Crypto/CryptoGen.php
```diff
@@ -267,16 +268,48 @@ private function coreDecrypt($sValue, $customPassword = null, $keySource = 'driv
...
+                // The goal here is so the call stack is exposed to track back to where the call originated.
+                $stackTrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+                $formattedStackTrace = $this->formatExceptionMessage($stackTrace);
+                error_log(text($e->getMessage()) . "\n" . text($formattedStackTrace));
+                return false;
+            }
         }
     }
 
+    private function formatExceptionMessage($stackTrace): string
```
sjpadgett, line 288: All you guys may want to look at this as it is an essential core function.

src/Common/Crypto/CryptoGen.php
```diff
@@ -28,9 +28,9 @@
 
 namespace OpenEMR\Common\Crypto;
 
+use Exception;
 use OpenEMR\Common\Utils\RandomGenUtils;
 
-#[\AllowDynamicProperties]
```
bradymiller, line 33: Did you mean to remove this?

sjpadgett: No but this is strange as I saw this before I pushed and copied corrected version to local and pushed. I then verified but both versions were present in master so beats me and will correct.
Dynamic properties were deprecated and this tag is for php8.2 but errors silently on lower version.
I'll try to track down and resolve the properties so we don't have to worry in future.

src/Common/Crypto/CryptoGen.php
```diff
@@ -373,6 +407,7 @@ public function aes256DecryptOne($sValue, $customPassword = null)
 
     // Function to decrypt a given string
     // This specific function is only used for backward compatibility
+    // TODO: Should be removed in the future or now if not needed. Brady should I refactor this?
```
bradymiller, line 410: this function still used for the first generation of legacy/encrypted log entries to decrypt them (always makes me chuckle a bit when i look at this function since the key is right there (and taken right from an example in the php documentation))

src/Common/Crypto/CryptoGen.php
```diff
@@ -405,16 +440,17 @@ public function aes256Decrypt_mycrypt($sValue)
...
+     * @param string $sub         This is the sublabel of the key
+     * @param string $keySource   This is the source of the standard keys. Options are 'drive' and 'database'
      *                            The 'drive' keys are stored at sites/<site-dir>/documents/logs_and_misc/methods
      *                            The 'database' keys are stored in the 'keys' sql table
      * @return string             Returns the key in raw form.
      */
     private function collectCryptoKey($version = "one", $sub = "", $keySource = 'drive')
     {
         // Check if key is in the cache first (and return it if it is)
+        // TODO: I believe method used here is deprecated in PHP 8.2 and will throw error in PHP 9.0
```
bradymiller, line 453: yep, you are correct in why i added the AllowDynamicProperties attribute :)
I'll look into a better long term solution.

bradymiller: This will fix it:
https://github.com/openemr/openemr/pull/7342

bradymiller: btw, I'll bring in my PR after you get your PR in (to avoid conflcts in your PR)

interface/modules/custom_modules/oe-module-weno/src/Services/TransmitProperties.php
```diff
@@ -379,10 +383,7 @@ private function getPatientInfo(): mixed
      */
     public static function styleErrors($error): string
     {
-        $log = "<div><p style='font-size: 1.0rem; color: red;'>" .
-            $error . "<br />" . xlt('Please address errors and try again!') .
-            "</p></div>";
-        return $log;
+        return "<div><p style='font-size: 1.0rem; color: red;'>" . $error . "</p></div>";
```
bradymiller, line 386: does $error above need to be html escaped?

sjpadgett: yep having trouble multitasking these days so you may catch me a lot!

src/Common/Crypto/CryptoGen.php
```diff
@@ -267,16 +268,49 @@ private function coreDecrypt($sValue, $customPassword = null, $keySource = 'driv
...
+            } catch (Exception $e) {
+                // log the exception message and call stack then return legacy null as false for
+                // those evaluating the return value as $return == false which with legacy will eval as false.
+                // I've seen this in the codebase, and it's a bit of a hack, but it's a way to return false instead of null.
+                // Dev's should use empty() instead of == false to check return from this function.
+                // The goal here is so the call stack is exposed to track back to where the call originated.
+                $stackTrace = debug_backtrace();
+                $formattedStackTrace = $this->formatExceptionMessage($stackTrace);
+                error_log(errorLogEscape($e->getMessage()) . "\n" . text($formattedStackTrace));
+                return false;
```
bradymiller, line 283: Changing the behavior of this function to return false instead of null (ie. no explicit return) here will possibly break things and would likely require scanning all code calling the decryptStandard() function. Is this an important change?

sjpadgett: yes and I insist. It is already broken in API's because this the legacy return null value is being evaluated as returning false.
== null vs == false eval differently. I know to check around but either way it's already broken. I comment in code about the reasons for this.

sjpadgett: Okay, Thanks sir and I'm still testing so we'll see.

sjpadgett:
@bradymiller I'll send you credential to setup and run module if you get time.
I'd like to see your impression of install and activating module w/o instructions from me and see how it goes.
I'm too close to it.
It won't take much time once I get on Gamma up for grab in a few minutes. It might even be fun! Boy oh boy, I'm really selling it huh?

interface/modules/custom_modules/oe-module-weno/templates/weno_setup.php
```diff
@@ -164,8 +184,9 @@ function togglePasswordVisibility(inputField) {
         <form id="set_form" name="set_form" class="form" role="form" method="post" action="">
             <div id="set-weno">
                 <input type="hidden" name="csrf_token_form" id="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken()); ?>" />
+                <button type="submit" id="form_save_top" name="form_save_top" class="d-none" value="Save"></button>
```
bradymiller, line 187: translate "Save"

src/Common/Crypto/CryptoGen.php
```diff
@@ -208,13 +226,13 @@ private function coreEncrypt($sValue, $customPassword = null, $keySource = 'driv
...
-     * @param  string  $keyNumber       This is the key number/version.
-     * @return string or false          returns the decrypted data or false if failed.
+     * @param string      $sValue         Encrypted data that will be decrypted.
+     * @param string|null $customPassword If null, then use standard keys. If provide a password, then will derive key from this.
+     * @param string      $keySource      This is the source of the keys. Options are 'drive' and 'database'
+     * @param string|null $keyNumber      This is the key number/version.
+     * @return false|string or false      returns the decrypted data or false if failed.
      */
-    private function coreDecrypt($sValue, $customPassword = null, $keySource = 'drive', $keyNumber = null)
+    private function coreDecrypt(string $sValue, string $customPassword = null, string $keySource = 'drive', string $keyNumber = null): false|string
```
bradymiller, line 235: change $customPassword and $keyNumber to `?string`
may make sense to change $value to that also since guessing null values may get in there sometimes which would return empty string i think

src/Common/Crypto/CryptoGen.php
```diff
@@ -75,14 +93,14 @@ public function encryptStandard($value, $customPassword = null, $keySource = 'dr
...
-     *                                   from outside OpenEMR to ensure bad actor is not trying to use an older version).
+     * @param string      $value          This is the data to decrypt.
+     * @param string|null $customPassword If provide a password, then will derive keys from this.(and will not use the standard keys)
+     * @param string      $keySource      This is the source of the standard keys. Options are 'drive' and 'database'
+     * @param int|null    $minimumVersion This is the minimum encryption version supported (useful if accepting encrypted data
+     *                                    from outside OpenEMR to ensure bad actor is not trying to use an older version).
      *
      */
-    public function decryptStandard($value, $customPassword = null, $keySource = 'drive', $minimumVersion = null)
+    public function decryptStandard(string $value, string $customPassword = null, string $keySource = 'drive', mixed $minimumVersion = null): false|string
```
bradymiller, line 103: change $customPassword to `?string`
change $minimumVersion to `?int`
may make sense to change $value to `?string` also since guessing null values may get in there sometimes which would return empty string i think

interface/modules/custom_modules/oe-module-weno/templates/weno_setup.php
```diff
@@ -89,23 +91,41 @@
...
+                    method: form.method,
+                    body: formData
+                }).then(response => response.json()).then(data => {
+                    window.scrollTo(0, scrollPosition);
+                }).catch(error => {
+                    window.scrollTo(0, scrollPosition);
+                });
+
+                if (isPersistEvent) {
+                    const successMsg = "<?php echo xlt('Auto Saved!'); ?>";
```
bradymiller, line 118:
```
const successMsg = <?php echo xlj('Auto Saved!'); ?>;
```

stephenwaite:
seal-dunk

PR #7309, opened 2024-04-02
Change search behaviour of select codes

### LINKED ISSUE #7276: bug: SNOMED lookup is slow in some places
OpenEMR 7.0.1 (1) on Linux.

"Adding hardware won't fix."

"You can do this:

This seems quick:

	- Type in a search term
	- Search
	
Where as this is slow:

Open a patient record
Open an encounter
Edit
Add Issue
Coding: Add
Select SNOMED Diagnosis
Search for: ...

Open a patient record
Medical Problems: + Add
Coding: Add
Select SNOMED Diagnosis
Search for: ...

sjpadgett:
- change by adding a search and reset button implemented by datatables search function.

Removes the type ahead feature. Even on a very fast machine search is slow on the large code datasets. 
With type ahead an event fired a search request every character where events would stack in DOM. IMO a no no

Fixes #7276 

image

### DIFF
--- a/interface/patient_file/encounter/select_codes.php
+++ b/interface/patient_file/encounter/select_codes.php
@@ -20,7 +20,7 @@ use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Core\Header;
 
 $codetype = empty($_GET['codetype']) ? '' : $_GET['codetype'];
-if (! empty($codetype)) {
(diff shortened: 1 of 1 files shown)


bradymiller:
2DV

sjpadgett:
Will merge and cherry pick directly from master. Not worth a pr unless someone wants me to pick some others.

PR #732, opened 2017-05-10
Remove undoMagicQuotes function

robertdown:
Our PHP min is 5.4, magic quotes was deprecated in PHP 5.3 and removed
in 5.4 so we no longer need to worry about this

PR #7313, opened 2024-04-03
Additional module visibility

juggernautsei:
Fixes #732

### DIFF
--- a/interface/modules/zend_modules/module/Installer/view/installer/installer/index.phtml
+++ b/interface/modules/zend_modules/module/Installer/view/installer/installer/index.phtml
@@ -373,6 +373,9 @@ $help_items = [
             <th class="bg-light text-dark">
                 <span style="font-size: 1.25rem"><?php echo $listener->z_xlt('Manage Laminas Modules'); ?></span>
             </th>
+            <th style="float:right" class="bg-light text-dark">
+                <a  class="ml-3 btn btn-secondary" href="https://www.open-emr.org/wiki/index.php/OpenEMR_Modules" target="_blank"><?php echo xlt('Additional Modules'); ?></a>
+            </th>
         <tr>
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/module/Installer/view/installer/installer/index.phtml
```diff
@@ -373,6 +373,9 @@ $help_items = [
             <th class="bg-light text-dark">
                 <span style="font-size: 1.25rem"><?php echo $listener->z_xlt('Manage Laminas Modules'); ?></span>
             </th>
+            <th style="float:right" class="bg-light text-dark">
+                <a  class="ml-3 btn btn-secondary" href="https://www.open-emr.org/wiki/index.php/OpenEMR_Modules" target="_blank"><?php echo xlt('Additional Modules'); ?></a>
+            </th>
```
sjpadgett, line 378: Tab is turned off so I don't understand how you are testing. I think I will delete the old tab code as looks like it causes some confusion. I left it in in case we wanted to use a tab for something else.

I like this idea and if you want I will add to my open PR and you can close this. I'll also refactor out old code.

juggernautsei: I can't test this because the changes that where made are not visible anywhere. I was testing in the current release and not on the bleeding edge. The bleeding edge is different as you pointed out. I see you got rid of where I thought it would go. And I just wanted to get the idea out there. Could you move it to where it would be visible? I can close this PR and you can pick it up elsewhere or take over this PR?

sjpadgett: Close this. I think I'll try it out either in new module help or separate button. I'll ping you and you'll be able to see it on my Gamma up for grab later.

PR #7321, opened 2024-04-05
Fixes #7319 ccda extra drug_unit list options

### LINKED ISSUE #7319: bug: CCDA import creates extra entries in list_options for drug_units list_id
When running the CCDA import process with a CCDA for a patient the CdaTemplateImportDispose->InsertImmunization() function is creating empty list_option entries for the drug_units list_id.  

You can reproduce this easily by running the ``openemr-cmd irp 20`` utility command or by passing a CCDA directly with the following command (run in the easy-dev docker environment)
```
php console openemr:ccda-newpatient-import --document=/root/synthea/output/ccda/Alysia661_Kristan816_Turcotte120_0f3907cd-027f-2ccf-cea2-6e81989c2704.xml 
```
This was discovered in the latest OpenEMR master release and likely exists in OpenEMR 7.0.2.

adunsulag:
Fixes #7319 
This fixes the error in the ccda that was creating tons of entries in the drug_entry list options list.  The CCDA was not checking against empty values and would continue to create new list_options for each immunization / prescription drug_unit when the unit was an empty string

I also added some debugging instruments on the prescription and immunization process which makes it easy to compare for a ccda what is being parsed and what is being stored in the database.

Added a debug flag to all of the ccda commands so that these options are available for people needing to do any debugging of the process.

### DIFF
--- a/src/Services/Cda/CdaTemplateImportDispose.php
+++ b/src/Services/Cda/CdaTemplateImportDispose.php
@@ -17,6 +17,7 @@ namespace OpenEMR\Services\Cda;
 use Application\Model\ApplicationTable;
 use Carecoordination\Model\CarecoordinationTable;
 use Document;
+use OpenEMR\Common\Command\Trait\CommandLineDebugStylerTrait;
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Logging\SystemLogger;
 use OpenEMR\Services\CodeTypesService;
@@ -28,6 +29,8 @@ require_once __DIR__ . '/../../../library/forms.inc.php';
 
 class CdaTemplateImportDispose
 {
+    use CommandLineDebugStylerTrait;
+
     protected $codeService;
     protected $userauthorized;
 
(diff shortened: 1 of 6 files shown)


src/Services/Cda/CdaTemplateImportDispose.php
```diff
@@ -822,31 +825,35 @@ public function InsertImmunization($imm_array, $pid, CarecoordinationTable $care
...
-            $q1_unit = "SELECT * FROM list_options WHERE list_id='drug_units' AND title=?";
-            $res_q1_unit = $appTable->zQuery($q1_unit, array($value['amount_administered_unit']));
-            foreach ($res_q1_unit as $val) {
-                $oid_unit = $val['option_id'];
-            }
-            if ($res_q1_unit->count() == 0) {
-                $lres = $appTable->zQuery("SELECT IFNULL(MAX(CONVERT(SUBSTRING_INDEX(option_id,'-',-1),UNSIGNED INTEGER))+1,1) AS option_id FROM list_options WHERE list_id = ?", array('drug_units'));
-                foreach ($lres as $lrow) {
-                    $oid_unit = $lrow['option_id'];
+            if (trim($value['amount_administered_unit'] ?? '') != '') {
```
stephenwaite, line 828: shouldn't have to compare against `''`? at least haven't need that for all of the php warnings fixed in the past...

adunsulag: Don't know when it was introduced, but trim throws an error if its null.  I believe that's what you're talking about?  As far as the != '', I decided to do that instead of empty because if the value is 0 empty returns true.

stephenwaite: okay so we don't want `amount_administered_unit` values of `0` to be recorded, thanks

adunsulag: ok, thought it was safer to err on the side of storing data, but if we want to skip saving a unit of 0, I'll switch it to use empty.

adunsulag:
Updated the condition check, supposedly the style fixes are all cleared... we'll see, still don't understand why I sometimes get a discrepancy on the stylechecker vs my local machine but c'est la vie.

bradymiller (approved):

bradymiller:
dance-bear

PR #7324, opened 2024-04-05
Fixes #7323 Billing Code Widget,deprecate isOption

### LINKED ISSUE #7323: feat: Extract billing code selector logic from options.inc.php into its own class
I need to be able to use the billing code selector logic in a module without having to grab everything needed to load the options.inc.php or LBF forms.  As a user I want to just include the widget that is needed for the billing code selector and not have to deal with everything else.

adunsulag:
Fixes #7323 

This is a billing code selector widget that was used in options.inc.php. I extracted it into its own class file so I can use it in a module as well as use it in options.inc.php.

I also moved the options.inc.php isOption command to the LayoutUtils so it can also be used externally.  I marked the isOption function as deprecated.


bradymiller (approved):

bradymiller:
2DV-2

PR #7325, opened 2024-04-05
Add SHA-512 hashing to session token in URL

surya-prakash-susarla:
This change adds SHA-512 hashing to the session token in the URL parameter to thwart visibility of session token which is currently being exposed through plain-text URL encoding.

### DIFF
--- a/interface/main/main_screen.php
+++ b/interface/main/main_screen.php
@@ -470,6 +470,6 @@ if ((isset($_POST['appChoice'])) && ($_POST['appChoice'] !== '*OpenEMR')) {
 
 // Pass a unique token, so main.php script can not be run on its own
 $_SESSION['token_main_php'] = RandomGenUtils::createUniqueToken();
-header('Location: ' . $web_root . "/interface/main/tabs/main.php?token_main=" . urlencode($_SESSION['token_main_php']));
+header('Location: ' . $web_root . "/interface/main/tabs/main.php?token_main=" . urlencode(hash('sha512',$_SESSION['token_main_php'])));
 exit();
 ?>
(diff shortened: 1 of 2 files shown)


bradymiller:
hi @surya-prakash-susarla ,
Awesome to see security focused work. In this case, note that this is not a session id. Exposing it is not an issue since it is only use one time when loading the script and is then removed. There is an option that allows not removing it (so users can reset the screen without it breaking), which is off by default. But even in that case, an exposed hash can be used. The critical point i am making here though is that this is not a session id, so a bad actor can do nothing with this token value.

Let me know if i am missing something.

surya-prakash-susarla:
Hello @bradymiller , thank you very much for taking a look at the change. I agree, the session token by itself seemed pretty harmless since it is randomly generated and has low potential for exploitation. I am unaware of the option to hide the token but considering that this is a randomly generated value which can be optionally used for extra security and is written to the logs, my idea was to increase the safety through hashing since it is minimally invasive and adds protection. 

I also noticed in this section that there is a direct comparison of this value and felt that hashing the value coming into this comparison from the external source would be better:
[interface/main/tabs/main.php](https://github.com/openemr/openemr/blob/b681b42fddc20494441115d71177f19cb72f94a0/interface/main/tabs/main.php#L35C1-L50C2)

Please let me know what you think. Thanks!

kojiromike:
Thanks for the security-conscious contribution, @surya-prakash-susarla!

After reviewing this with fresh eyes, I agree with Brady's original assessment — this token isn't sensitive since it's single-use and requires the session cookie to be useful. Hashing it doesn't change the security properties: the hash itself becomes the token that would need to be captured.

Since the PR has been stale for a while and doesn't provide a clear security benefit, I'm going to close it to keep the backlog tidy. If you'd like to revisit this with a specific threat model in mind, feel free to reopen or start a new discussion.

Thanks again for thinking about security — we always appreciate contributors looking out for potential improvements!

PR #7333, opened 2024-04-10
Add PATCH and PUT functions for Appointments

BerwoutK:
This patch will add two new api end-points, PUT and PATCH /api/appointment/{eid} these can be used to update the entity or part of the entity.

In order to make a more feature complete API, and to support the use case of wanting to e.g. edit the status of an appointment
the ability to update and alter the appointment has been made available.

+ Add PUT request for Appointments
+ Add PATCH request for Appointments

I copy-pasted a lot of code or used it as an example but there are quite a few bits I am uncertain about if it is correct, please advise where necessary.

### DIFF
--- a/_rest_routes.inc.php
+++ b/_rest_routes.inc.php
@@ -5288,6 +5288,82 @@ RestConfig::$ROUTE_MAP = array(
     },
 
     /**
+     *  @OA\Patch(
+     *      path="/api/appointment/{eid}",
+     *      description="Edits an appointment",
+     *      tags={"standard"},
+     *      @OA\Parameter(
+     *          name="eid",
+     *          in="path",
+     *          description="The eid for the appointment.",
+     *          required=true,
+     *          @OA\Schema(
+     *              type="string"
+     *          )
+     *      ),
+     *      @OA\Response(
+     *          response="200",
+     *          ref="#/components/responses/standard"
+     *      ),
+     *      @OA\Response(
+     *          response="400",
+     *          ref="#/components/responses/badrequest"
+     *      ),
+     *      @OA\Response(
+     *          response="401",
+     *          ref="#/components/responses/unauthorized"
+     *      ),
+     *      security={{"openemr_auth":{}}}
+     *  )
+     */
+    "PATCH /api/appointment/:eid" => function ($eid) {
+        RestConfig::authorization_check("patients", "appt");
+        $data = (array) (json_decode(file_get_contents("php://input")));
+        $return = (new AppointmentRestController())->patch($eid, $data);
+
+        RestConfig::apiLog($return);
+        return $return;
+    },
+
+    /**
+     *  @OA\Put(
+     *      path="/api/appointment/{eid}",
+     *      description="Updates an appointment",
+     *      tags={"standard"},
+     *      @OA\Parameter(
+     *          name="eid",
+     *          in="path",
+     *          description="The eid for the appointment.",
+     *          required=true,
+     *          @OA\Schema(
+     *              type="string"
+     *          )
+     *      ),
+     *      @OA\Response(
+     *          response="200",
+     *          ref="#/components/responses/standard"
+     *      ),
+     *      @OA\Response(
+     *          response="400",
+     *          ref="#/components/responses/badrequest"
+     *      ),
+     *      @OA\Response(
+     *          response="401",
+     *          ref="#/components/responses/unauthorized"
+     *      ),
+     *      security={{"openemr_auth":{}}}
+     *  )
+     */
+    "PUT /api/appointment/:eid" => function ($eid) {
+        RestConfig::authorization_check("patients", "appt");
+        $data = (array) (json_decode(file_get_contents("php://input")));
+        $return = (new AppointmentRestController())->put($eid, $data);
+
+        RestConfig::apiLog($return);
+        return $return;
+    },
+
+    /**
      *  @OA\Delete(
      *      path="/api/patient/{pid}/appointment/{eid}",
      *      description="Delete a appointment",
--- a/src/RestControllers/AppointmentRestController.php
+++ b/src/RestControllers/AppointmentRestController.php
@@ -90,4 +90,48 @@ class AppointmentRestController
         }
         return RestControllerHelper::responseHandler($serviceResult, null, 200);
     }
+
+    /**
+     * Processes an HTTP PATCH request used to update an existing Appointment.
+     *
+     * @param string $uuid The identifier of the Appointment.
+     * @param array $data The new data for the Appointment.
+     *
+     * @return void
+     */
+    public function patch(string $uuid, array $data): void
+    {
+        try {
+            $result =  $this->appointmentService->edit($uuid, $data);
+        } catch (\Exception $exception) {
+            (new SystemLogger())->errorLogCaller($exception->getMessage(), ['trace' => $exception->getTraceAsString(), 'uuid' => $uuid, 'data' => $data]);
+            RestControllerHelper::responseHandler(['message' => 'Failed to update appointment'], null, 500);
+
+            return;
+        }
+
+        RestControllerHelper::responseHandler($result, null, 200);
+    }
+
+    /**
+     * Processes an HTTP PUT request used to update an existing Appointment.
+     *
+     * @param string $uuid The identifier of the Appointment.
+     * @param array $data The new data for the Appointment.
+     *
+     * @return void
+     */
+    public function put(string $uuid, array $data): void
+    {
+        try {
+            $result =  $this->appointmentService->update($uuid, $data);
+        } catch (\Exception $exception) {
+            (new SystemLogger())->errorLogCaller($exception->getMessage(), ['trace' => $exception->getTraceAsString(), 'uuid' => $uuid, 'data' => $data]);
+            RestControllerHelper::responseHandler(['message' => 'Failed to update appointment'], null, 500);
+
+            return;
+        }
+
+        RestControllerHelper::responseHandler($result, null, 200);
+    }
 }
--- a/src/Services/AppointmentService.php
+++ b/src/Services/AppointmentService.php
@@ -32,6 +32,7 @@ class AppointmentService extends BaseService
     const PATIENT_TABLE = "patient_data";
     const PRACTITIONER_TABLE = "users";
     const FACILITY_TABLE = "facility";
+    const APPOINTMENT_TABLE = self::TABLE_NAME;
 
     /**
      * @var EncounterService
@@ -94,7 +95,7 @@ class AppointmentService extends BaseService
         $validator->required('pc_hometext')->string();
         $validator->required('pc_apptstatus')->string();
         $validator->required('pc_eventDate')->datetime('Y-m-d');
-        $validator->required('pc_startTime')->length(5); // HH:MM is 5 chars
+        $validator->required('pc_startTime')->lengthBetween(5, 8); // HH:MM is 5 chars, HH:MM:SS is 8 as returned by OpenEMR.
         $validator->required('pc_facility')->numeric();
         $validator->required('pc_billing_location')->numeric();
         $validator->optional('pc_aid')->numeric()
@@ -189,7 +190,7 @@ class AppointmentService extends BaseService
         return $processingResult;
     }
 
-    public function getAppointmentsForPatient($pid)
+    public function getAppointmentsForPatient(?string $pid)
     {
         $sqlBindArray = array();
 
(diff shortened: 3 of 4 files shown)


sjpadgett:
@adunsulag Was there a reason we haven't reviewed this? I don't have my API dev setup for testing this.

bradymiller:
hi @BerwoutK , Sorry for the very long delay on this. We don't use `patch` endpoints (basically just use `put` to do the same thing as patch would do).

adunsulag:
@BerwoutK I'm open to considering supporting patching on our standard APIs, but it's not how we've typically done things in the rest of the APIs.  I'd prefer consistency here, but I'm open to hearing an argument on why you think we should start using PATCH. Its a lot more involved if we wanted to support patching in our FHIR apis.

BerwoutK:
There is no real reason to differentiate between PUT/PATCH other than it is what the standard says. And I like to play by the book, I can remove the PATCH and keep it as PUT doing PATCH.

adunsulag:
@BerwoutK I'm ok with partial updates, again its more of a consistency thing.  

If people are willing to help contribute and add more PATCH endpoints with partial updates I'm fine with that.  The key point is partial updates have to make sure the rest of the data record has internal consistency, as long as we have people like yourself or others driving PATCH endpoints I'm fine with breaking from prior patterns if we will actually see more usage of it.

sjpadgett:
I think I'll stay out of this!:)

bradymiller:
There will be package/docker infrastructure changes that will need to be made if support PATCH (simple, but still a bit of a pain :) ):
https://github.com/openemr/openemr-devops/blob/master/docker/openemr/7.0.2/openemr.conf#L27

adunsulag:
@bradymiller let's target patch support for 7.0.3.1 in the dockers. I'd like to bring this in the next patch.

bradymiller:
@adunsulag , ask and you shall receive :)
https://github.com/openemr/openemr-devops/pull/409

(also rebuilding all the pertinent dockers now)

_rest_routes.inc.php
```diff
@@ -5287,6 +5287,82 @@
...
+     *      @OA\Response(
+     *          response="401",
+     *          ref="#/components/responses/unauthorized"
+     *      ),
+     *      security={{"openemr_auth":{}}}
+     *  )
+     */
+    "PATCH /api/appointment/:eid" => function ($eid) {
+        RestConfig::authorization_check("patients", "appt");
+        $data = (array) (json_decode(file_get_contents("php://input")));
```
kojiromike, line 5321: `(array) json_decode(...)` only flattens the **top-level** stdClass into an associative array — any nested object in the payload stays as `stdClass`. Downstream, `$data` flows into `validate()` (which treats it as an array) and `buildUpdateColumns()` (which iterates keys), so a nested object value silently produces garbage in the UPDATE.

Pass `true` as the second argument so the whole tree is associative:

```suggestion
        $data = (array) json_decode(file_get_contents("php://input"), true);
```

Same fix needed on the PUT route below.

_rest_routes.inc.php
```diff
@@ -5287,6 +5287,82 @@
...
+        $return = (new AppointmentRestController())->patch($eid, $data);
```
kojiromike, line 5322: The result of `->patch()` is a `void` method (see `AppointmentRestController::patch`/`put`). Assigning it to `$return` gives `null`, which is then passed to `RestConfig::apiLog($return)` and `return`ed from the route closure. The controller already calls `RestControllerHelper::responseHandler(...)` itself, so the route never gets a response object to log/return.

Compare with the sibling `POST /api/patient/:pid/appointment` route, where the controller returns the responder output and the route forwards it. Either have `patch()`/`put()` return the responder result (and stop double-handling), or restructure to mirror `post()`.

_rest_routes.inc.php
```diff
@@ -5287,6 +5287,82 @@
...
+     *      @OA\Response(
+     *          response="401",
+     *          ref="#/components/responses/unauthorized"
+     *      ),
+     *      security={{"openemr_auth":{}}}
+     *  )
+     */
+    "PUT /api/appointment/:eid" => function ($eid) {
+        RestConfig::authorization_check("patients", "appt");
+        $data = (array) (json_decode(file_get_contents("php://input")));
```
kojiromike, line 5359: Same `json_decode` flattening issue as the PATCH route — pass `true` as the second argument so nested objects become arrays.

_rest_routes.inc.php
```diff
@@ -5287,6 +5287,82 @@
...
+        $return = (new AppointmentRestController())->put($eid, $data);
```
kojiromike, line 5360: Same controller-returns-void issue as the PATCH route — `$return` is `null` here.

src/RestControllers/AppointmentRestController.php
```diff
@@ -90,4 +90,48 @@ public function delete($eid)
...
+
+    /**
+     * Processes an HTTP PATCH request used to update an existing Appointment.
+     *
+     * @param string $uuid The identifier of the Appointment.
+     * @param array $data The new data for the Appointment.
+     *
+     * @return void
+     */
+    public function patch(string $uuid, array $data): void
```
kojiromike, line 102: Two issues with this signature:

1. `@return void` plus calling `responseHandler(...)` without returning it means the route closure can never forward a response. The existing `delete()` method (which does return the `responseHandler` result) is the correct pattern to follow.
2. The parameter is named `$uuid` and documented as a UUID, but the route binds the URL `:eid` (an integer `pc_eid`, not a UUID — `getAppointment($id)` searches `WHERE pc_eid = ?`). Rename to `$eid` and adjust the PHPDoc to match.

Same issues apply to `put()` below.

src/RestControllers/AppointmentRestController.php
```diff
@@ -90,4 +90,48 @@ public function delete($eid)
...
+    {
+        try {
+            $result =  $this->appointmentService->edit($uuid, $data);
+        } catch (\Exception $exception) {
```
kojiromike, line 106: Catching bare `\Exception` here masks programmer errors (TypeErrors, etc.) as "Failed to update appointment" 500s. Prefer catching the specific exceptions the service can actually throw, or let them propagate so the framework's error handler logs them — `errorLogCaller` + a generic 500 hides root causes from operators.

src/RestControllers/AppointmentRestController.php
```diff
@@ -90,4 +90,48 @@ public function delete($eid)
...
+            (new SystemLogger())->errorLogCaller($exception->getMessage(), ['trace' => $exception->getTraceAsString(), 'uuid' => $uuid, 'data' => $data]);
+            RestControllerHelper::responseHandler(['message' => 'Failed to update appointment'], null, 500);
```
kojiromike, line 108: Validation failures from the service surface as a `ProcessingResult` with `validationMessages` set — and this controller hands it straight to a 200 response. The caller cannot tell a successful write from a 400-shaped failure.

Look at how `post()` (line 70) uses `RestControllerHelper::validationHandler($validationResult)` — that helper returns a 400 array when validation fails. The same gate is needed here before the success path.

Same issue in `put()`.

src/Services/AppointmentService.php
```diff
@@ -32,6 +32,7 @@ class AppointmentService extends BaseService
     const PATIENT_TABLE = "patient_data";
     const PRACTITIONER_TABLE = "users";
     const FACILITY_TABLE = "facility";
+    const APPOINTMENT_TABLE = self::TABLE_NAME;
```
kojiromike, line 35: This constant is just an alias for `self::TABLE_NAME` defined two lines up — every existing query in this file uses `self::TABLE_NAME`. Adding a parallel name for the same value creates a place where future edits could drift (one constant changes, the other doesn't). The new `UPDATE` statements below can use `self::TABLE_NAME` directly:

```suggestion
```
(remove the line)

src/Services/AppointmentService.php
```diff
@@ -189,7 +190,7 @@ public function search($search, $isAndCondition = true)
         return $processingResult;
     }
 
-    public function getAppointmentsForPatient($pid)
+    public function getAppointmentsForPatient(?string $pid)
```
kojiromike, line 193: `pc_eid` in `openemr_postcalendar_events` is an integer (`INT(11) AUTO_INCREMENT`). The existing call sites that reach this method pass `pid` values that come from `patient_data.pid` (also an integer) or `null`. Widening the parameter to `?string` here doesn't reflect the real domain; the comparison `WHERE pid = ?` will still work via MySQL coercion but the type signature is now misleading. Prefer `?int`, or leave the original loose typing if there's a backwards-compat concern with other callers.

Is there a specific caller that needs `string`? If not, this looks like an accidental change carried in from somewhere else.

src/Services/AppointmentService.php
```diff
@@ -647,4 +648,96 @@ public function getOneCalendarCategory($cat_id)
         $sql = "SELECT * FROM openemr_postcalendar_categories WHERE pc_catid = ?";
         return QueryUtils::fetchRecords($sql, [$cat_id]);
     }
+
+    public function edit(string $id, array $data): ProcessingResult
+    {
+        if (empty($data)) {
```
kojiromike, line 654: `empty()` is unreliable here — `empty(['0' => 'something'])` is false but `empty([0 => 0])` is true; an array with a single `0`/`''`/`false` value is treated as empty. For PATCH the right check is "no fields supplied":

```suggestion
        if (count($data) === 0) {
```

Same nit in `update()` below.

src/Services/AppointmentService.php
```diff
@@ -647,4 +648,96 @@ public function getOneCalendarCategory($cat_id)
...
+            $result = new ProcessingResult();
+            $result->setValidationMessages('Invalid Data');
+
+            return $result;
+        }
+
+        $data['pc_eid'] = $id;
+        $appointment = $this->getAppointment($id)[0] ?? null; // We really need to fix the getAppointment function...
```
kojiromike, line 662: `getAppointment($id)` returns a JOIN against `users`, `patient_data`, `facility`, and the calendar categories table — the returned row contains keys like `ufname`, `fname`, `lname`, plus duplicated/aliased column names — not the raw `openemr_postcalendar_events` row. Merging those keys into `$data` and then handing the result to `validate()` (which expects `pc_catid`, `pc_title`, etc.) and `buildUpdateColumns()` means:

1. Required validator keys (`pc_catid`, `pc_facility`, `pc_billing_location`) may not be present under those names in the joined row, so PATCH against an existing record can still fail validation.
2. JOIN-only keys get filtered out by `buildUpdateColumns()` (it checks against `$this->fields`), which is the only thing saving you from writing nonsense — but it's coincidental, not designed.

For a real PATCH, fetch the raw row from `openemr_postcalendar_events` (e.g. via `QueryUtils::fetchRecords('SELECT * FROM ' . self::TABLE_NAME . ' WHERE pc_eid = ?', [$id])`) and merge from that, not from the read-side joined view.

src/Services/AppointmentService.php
```diff
@@ -647,4 +648,96 @@ public function getOneCalendarCategory($cat_id)
...
+            return $result;
+        }
+
+        foreach ($appointment as $key => $value) {
+            if (!array_key_exists($key, $data)) {
+                $data[$key] = $value;
+            }
+        }
+
+        $validation = $this->validate($data);
```
kojiromike, line 677: The `validate()` method declares **every** non-optional field as `required`. That's correct for POST/PUT (full-resource replace) but wrong for PATCH (partial update) — a client sending only `{"pc_apptstatus": "~"}` will be rejected because they didn't resend `pc_title`, `pc_duration`, etc. Combined with the joined-row issue above, this means PATCH may be effectively unusable for the partial-update case it's meant to support.

Consider either (a) a separate validator that marks fields as `optional` for the partial-update path, or (b) merging from the real row first and validating the merged result (only viable once the row-source bug above is fixed).

src/Services/AppointmentService.php
```diff
@@ -647,4 +648,96 @@ public function getOneCalendarCategory($cat_id)
...
+            return $result;
+        }
+
+        $query = $this->buildUpdateColumns($data);
+        $sql = " UPDATE " . self::APPOINTMENT_TABLE . " SET ";
+        $sql .= $query['set'];
+        $sql .= " WHERE `pc_eid` = ?";
+
+        array_push($query['bind'], $id);
+        $sqlResult = sqlStatement($sql, $query['bind']);
```
kojiromike, line 692: `pc_eid` is integer (auto-increment column). `$id` arrives here as a string from the URL. Binding works via MySQL coercion, but explicit `(int) $id` would match the column's actual type and lines up with how PHPStan level max wants to see this. Same in `update()` below.

Also: `$data['pc_eid'] = $id` a few lines up means `pc_eid` is in `$data`. `buildUpdateColumns()` filters auto-increment columns out, so this won't end up in the `SET` clause, but it's confusing — consider removing the assignment.

src/Services/AppointmentService.php
```diff
@@ -647,4 +648,96 @@ public function getOneCalendarCategory($cat_id)
...
+            $result = new ProcessingResult();
+            $result->addErrorMessage('error processing SQL Update');
+
+            return $result;
+        }
+
+        return $this->search(['pc_eid' => $id]);
+    }
+
+    public function update(string $id, array $data): ProcessingResult
```
kojiromike, line 705: `update()` (PUT) doesn't fetch the existing record at all — it relies on the client sending every required field, and any column the client omits (and which is not nullable) gets nulled out via `buildUpdateColumns()` (well, set to the null_value option, which defaults to `''`). That's a footgun even for a strict PUT semantics, because the table has columns like `pc_apptstatus`, `pc_facility`, etc. that would be silently blanked.

A safer PUT either:
- fetches the row first and merges before update (effectively the same flow as the merge step in `edit()`, just without making missing fields optional), or
- restricts the update to only the columns explicitly present in `$data` (let unprovided columns keep their current value).

This also means the audit-trail story is unclear — the read-back via `$this->search([...])` will reflect the blanked columns, masking the data loss in the response.

kojiromike (changes requested):
Is there any interest in continuing this? More API verbs would be welcome. Review below, as-is.

---

Two HTTP semantics issues and several correctness/typing issues. The most important: validation failures are returned as HTTP 200 (callers cannot distinguish a successful write from a validator rejection), and `(array) json_decode(...)` only converts the outermost layer — nested JSON objects remain `stdClass` and reach `validate()` / `buildUpdateColumns()` with unexpected shapes.

Discussion above settled on supporting both PUT and PATCH; the comments below assume that decision stands.

Mechanical state: this PR has drifted significantly from `master` (over 18 months since last update) and will need a rebase plus re-running the test matrix before merge.

PR #7346, opened 2024-04-13
Fix: delete entry in immunization

### LINKED ISSUE #7345: Cannot delete entry from immunization list
### Describe the bug

The delete option in the immunization list of a patient is not working when clicked. It just reloads the tab.

### To Reproduce

1. Login to Open EMR as admin.
2. Navigate to the dashboard of a patient.
3. Click on the edit option in the immunization box on the bottom right corner.
4. Add an immunization entry by filling the necessary fields.
5. If you scroll to the bottom and then scroll to the right you can see the delete button.
6. Click it but it does not work.
image

### Expected behavior
The immunization entry should be deleted when the user clicks on the delete button.

### Client configuration

- Browser: Chrome v123
- OpenEMR version: 7.0.2
- Operating system: windows

sathiya06:
Added the delete class to the button which is used by the java script function to select the delete button.

Fixes #7345

It fixes the issue with the delete button in the immunization list of a patient. 

Adding delete to the class attribute of the button element solves this issue because this triggers the javascript onclick function responsible for deleting the immunization entry.

### DIFF
--- a/interface/patient_file/summary/immunizations.php
+++ b/interface/patient_file/summary/immunizations.php
@@ -822,7 +822,7 @@ tr.selected {
 
                                 echo "<td><input type='checkbox' class='error' id='" . attr($row["id"]) . "' value='" . xlt('Error') . "' " . $checkbox . " /></td>";
 
-                                echo "<td><button type='button' class='btn btn-danger btn-delete' id='" . attr($row["id"]) . "' value='" . xlt('Delete') . "'>" . xlt('Delete') . "</button></td>";
+                                echo "<td><button type='button' class='delete btn btn-danger btn-delete' id='" . attr($row["id"]) . "' value='" . xlt('Delete') . "'>" . xlt('Delete') . "</button></td>";
                                 echo "</tr>";
                         }
                         ?>


interface/patient_file/summary/immunizations.php
```diff
@@ -822,7 +822,7 @@ function saveImmunizationObservationResults($id, $immunizationdata)
 
                                 echo "<td><input type='checkbox' class='error' id='" . attr($row["id"]) . "' value='" . xlt('Error') . "' " . $checkbox . " /></td>";
 
-                                echo "<td><button type='button' class='btn btn-danger btn-delete' id='" . attr($row["id"]) . "' value='" . xlt('Delete') . "'>" . xlt('Delete') . "</button></td>";
+                                echo "<td><button type='button' class='delete btn btn-danger btn-delete' id='" . attr($row["id"]) . "' value='" . xlt('Delete') . "'>" . xlt('Delete') . "</button></td>";
```
sjpadgett, line 825: This trigger class must have been removed during one of the many restyling forays. This is a bug fix so looks good to me.

sjpadgett (approved):
Will try for patch 1.

Issue #7355, opened 2024-04-16
 Bug: OpenEMR allows invalid input for Google signin email


mabeshark:
### Describe the bug

CWE-20: Improper Input Validation

ASVS V5.1 Input Validation  5.1.3 Verify that all input (HTML form fields, REST requests, URL parameters, HTTP headers, cookies, batch files, RSS feeds, etc) is validated using positive validation

OpenEMR does not perform input validation/sanitation on the Google signin email field. This allows a user to save text containing JavaScript and other potentially malicious data to OpenEMR , which could result in a persisted XSS or other attacks/vulnerabilities.

### To Reproduce

1. Login to OpenEMR as an administrator or user with permissions to manage users
2. On the top navigation bar, hover over **Admin** and select **Users**
3. Click on **New user** or select to **Edit an existing user**
4. For the Google Email for Login, type in any invalid input, such as `<script><javascript>` or any other invalid and potentially malicious data.
5. Click on **Save** and notice that the user details are saved and the invalid Google email is **incorrectly** allowed.

![image](https://github.com/openemr/openemr/assets/38800918/59e8ac34-865b-41b9-8f05-463882fc791e)

### Expected behavior

OpenEMR front-end UI should not allow an invalid and potentially malicious email address, it should sanitize the input using a industry standard Regular Expression (RegEx). Additionally, the back-end should also have protections in place in case the front-end is circumvented.

### Client configuration

- Browser: Firefox 124.0.2 (64-bit)
- OpenEMR version: v7.0.2
- Operating system: Windows 10

PR #7356, opened 2024-04-16
Fix: Input sanitation on google signin email

mabeshark:
Fixes https://github.com/openemr/openemr/issues/7355

This PR adds validation on the front-end UI and the backend to ensure that input is validated and sanitized for the google signin email input field. Without the PR, invalid or potentially malicious data can be saved in the field, such as JavaScript or scripts. (ASVS 5.1.3 and CWE-20)

After update:
image

1. Use an industry standard Regular Expression (RegEx) from the Open Worldwide Application Security Project (OWASP) on the front-end https://owasp.org/www-community/OWASP_Validation_Regex_Repository
2. Use the built in PHP function filter_var to validate the email on the backend https://www.php.net/manual/en/function.filter-var.php or https://www.w3schools.com/php/filter_validate_email.asp

### DIFF
--- a/interface/usergroup/user_admin.php
+++ b/interface/usergroup/user_admin.php
@@ -129,6 +129,18 @@ function submitform() {
         }
 
     }//If pwd null ends here
+
+    // Valiate Google email (if provided)
+    if(document.forms[0].google_signin_email.value != ""){
+        // https://owasp.org/www-community/OWASP_Validation_Regex_Repository
+        var mailformat = /^[a-zA-Z0-9_+&*-]+(?:\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/;
+        if(!document.forms[0].google_signin_email.value.match(mailformat)) {
+            flag=1;
+            alert(<?php echo xlj('Google email provided is invalid/not properly formatted (e.g. first.last@gmail.com)') ?>);
+            return false;
+        }
+    }
+
 <?php } ?>
   if (document.forms[0].access_group_id) {
     var sel = getSelected(document.forms[0].access_group_id.options);
--- a/interface/usergroup/user_admin.php
+++ b/interface/usergroup/user_admin.php
@@ -131,7 +131,7 @@ function submitform() {
     }//If pwd null ends here
 
     // Valiate Google email address (if provided)
-    if(document.forms[0].google_signin_email.value != "" && !isValidEmail(document.forms[0].google_signin_email.value)){
+    if(document.forms[0].google_signin_email.value != "" && !isValidEmail(document.forms[0].google_signin_email.value)) {
         flag=1;
         alert(<?php echo xlj('Google email provided is invalid/not properly formatted (e.g. first.last@gmail.com)') ?>);
         return false;
--- a/interface/usergroup/user_admin.php
+++ b/interface/usergroup/user_admin.php
@@ -130,15 +130,11 @@ function submitform() {
 
     }//If pwd null ends here
 
-    // Valiate Google email (if provided)
-    if(document.forms[0].google_signin_email.value != ""){
-        // https://owasp.org/www-community/OWASP_Validation_Regex_Repository
-        var mailformat = /^[a-zA-Z0-9_+&*-]+(?:\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/;
-        if(!document.forms[0].google_signin_email.value.match(mailformat)) {
-            flag=1;
-            alert(<?php echo xlj('Google email provided is invalid/not properly formatted (e.g. first.last@gmail.com)') ?>);
-            return false;
-        }
+    // Valiate Google email address (if provided)
+    if(document.forms[0].google_signin_email.value != "" && !isValidEmail(document.forms[0].google_signin_email.value)){
+        flag=1;
+        alert(<?php echo xlj('Google email provided is invalid/not properly formatted (e.g. first.last@gmail.com)') ?>);
+        return false;
     }
 
 <?php } ?>
--- a/interface/usergroup/usergroup_admin.php
+++ b/interface/usergroup/usergroup_admin.php
@@ -303,7 +303,7 @@ if (isset($_POST["privatemode"]) && $_POST["privatemode"] == "user_admin") {
         }
         if (isset($_POST["google_signin_email"])) {
             // Save email as null if input was empty or invalid email address format (prevent attacks such as stored xss)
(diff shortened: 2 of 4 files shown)


mabeshark:
Hi @stephenwaite and @bradymiller -- thank you in advance for your time reviewing this pull request, I really appreciate it.

mabeshark:
Hi @adunsulag any chance you would be able to review this? I appreciate your time and feedback. Thank you

interface/usergroup/usergroup_admin_add.php
```diff
@@ -117,6 +117,16 @@ function submitform() {
         }
     } //secure_pwd if ends here
 
+    // Valiate Google email (if provided)
+    if(document.new_user.google_signin_email.value != ""){
+        // https://owasp.org/www-community/OWASP_Validation_Regex_Repository
+        var mailformat = /^[a-zA-Z0-9_+&*-]+(?:\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/;
+        if(!document.new_user.google_signin_email.value.match(mailformat)) {
+            alert(<?php echo xlj('Google email provided is invalid/not properly formatted (e.g. first.last@gmail.com)') ?>);
```
adunsulag, line 125: I've never liked this approach of how these alerts are done, but I see you've followed the same pattern as the erx_enable code so we'll keep it. 

I've never been a big believer in regex validators for email, the OWASP isn't technically valid for all RFC 5322 email addresses since it tends to blow up when using local email addresses, however, since this is specific to google email addresses, I'm fine with you using this.

For any other admin's I did confirm this is copied directly from the OWASP example regex validator.

adunsulag (changes requested):
One minor change on the php side of things and then this looks fine.

interface/usergroup/usergroup_admin.php
```diff
@@ -301,7 +301,9 @@
             sqlStatement("update users set supervisor_id = ? where id = ? ", array((int)$_POST["supervisor_id"], $_POST["id"]));
         }
         if (isset($_POST["google_signin_email"])) {
-            if (empty($_POST["google_signin_email"])) {
+            // Save email as null if input was empty or invalid email address format (prevent attacks such as stored xss)
+            // https://www.php.net/manual/en/filter.filters.validate.php
+            if (empty($_POST["google_signin_email"]) || !filter_var($_POST["google_signin_email"], FILTER_VALIDATE_EMAIL))  {
```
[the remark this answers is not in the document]
mabeshark: Thanks @adunsulag ! I've updated my code/PR. Does it look good to approve or is anything else needed?

interface/usergroup/user_admin.php
```diff
@@ -129,6 +129,18 @@ function submitform() {
...
+    // Valiate Google email (if provided)
+    if(document.forms[0].google_signin_email.value != ""){
+        // https://owasp.org/www-community/OWASP_Validation_Regex_Repository
+        var mailformat = /^[a-zA-Z0-9_+&*-]+(?:\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/;
+        if(!document.forms[0].google_signin_email.value.match(mailformat)) {
+            flag=1;
+            alert(<?php echo xlj('Google email provided is invalid/not properly formatted (e.g. first.last@gmail.com)') ?>);
+            return false;
+        }
+    }
```
bradymiller, line 142: Instead of duplicating code, rec placing a js function for checking email addresses in the library/js/utility.js script. This script is automatically brought into these php scripts, so you will be able to call the js function from them.

Also, are email addressing strictly ascii characters or can they include accented characters and chinese characters etc? (I don't know the answer to this)

interface/usergroup/usergroup_admin.php
```diff
@@ -301,7 +302,8 @@
             sqlStatement("update users set supervisor_id = ? where id = ? ", array((int)$_POST["supervisor_id"], $_POST["id"]));
         }
         if (isset($_POST["google_signin_email"])) {
-            if (empty($_POST["google_signin_email"])) {
+            // Save email as null if input was empty or invalid email address format (prevent attacks such as stored xss)
+            if (empty($_POST["google_signin_email"]) || !ValidationUtils::isValidEmail($_POST["google_signin_email"]))  {
```
bradymiller, line 306: We have stayed away from this type of thing (ie. silently fail importing of data if not valid since validation can be tough especially for an application that fully supports internationalization). We generally are very flexible with what can get into the database. For example, the demographics form is filled with places where can insert a stored xss vector, however, we are vigilant on the html/js output side of things to mitigate this with escaping.

interface/usergroup/user_admin.php
bradymiller: btw, the nice thing about centralizing this kind of thing is ability to adjust the function as needed in the future.

mabeshark: Great tip! Thanks I have refactored the code and pushed an updated commit.

interface/usergroup/usergroup_admin.php
mabeshark: Google email addresses do not allow Chinese characters for example (screenshot below), this PR adds protections to the UI and this code is just in case an attacker circumvents the UI and sends the request directly (e.g. Postman). Are you ok with that and approving, or would like me to modify some functionality?

image

mabeshark:
Hi @adunsulag sorry to bug you, are there any other changes needed? Thanks for your help!

interface/usergroup/usergroup_admin.php
bradymiller: Recommend removing the validation check on the server side of things. The protection it provides is not worth the potential bugs that can arise (for example, if the validator doesn't like a valid email address because of a bug or more flexibility in email addresses in the future, then it just simply will not save it without user knowing it; also would silently go away when editing a prior current user for other reasons). Just not worth it. This is why OpenEMR's strategy is to generally escape the output rather than try to validate the input on the server side. Ok to do on the client side since bugs will be obvious and then easily addressed with the central email js function.

mabeshark: Sounds good @bradymiller  I have removed the validation check on the server side, it only validates on the client side now. Does this look ok for approval and merging? Thanks!

mabeshark: Looks like all automated tests/checks have passed too if you're ok with the changes.

mabeshark: Hi @bradymiller and @adunsulag  hope you had a nice weekend. Anything else you need me to change on this PR for approval? Thank you.

adunsulag (approved):
Pulled this down locally and tested.  Everything is working well from my tests.

mabeshark:
Thanks again both of you!

PR #7358, opened 2024-04-16
Fix: Update minimum password requirement to at least 12 characters

### LINKED ISSUE #7357: Bug: Minimum password requirement is less than 12 characters
### Describe the bug

ASVS 2.1.1 (Verify that user set passwords are at least 12 characters in length (after multiple spaces are combined).

CWE-521 (Weak Password Requirements)

OpenEMR accepts passwords 9 characters or greater, creating a conflict with the ASVS requirement above.

### To Reproduce

**When creating a new user:**
1. Open OpenEMR and login using username: admin and password: pass.
2. Hover over Admin in the top menu bar.
3. While hovering over Admin, click on Users.
4. Select “Add User”
5. The following are the requirements for the field inputs:
6. Username: test1
7. Password: choose an 11-character string such as: “Testing123!”
8. Your Password: pass
9. First Name: Test
10. Last Name: User
11. Access Control: Clinicians
12. Click Save.

**When changing password:**
1. Open OpenEMR and log in.
2. Click on the profile icon at top-right and choose “Change Password.”
3. Change password to an 11-character string such as: “Testing123!"
4. Click “Save Changes.”

### Expected behavior

Passwords set when creating a user or changing a password that are less than 12-characters should not be accepted.

### Client configuration

Browser: Firefox 124.0.2 (64-bit)
OpenEMR version: v7.0.2
Operating system: Windows 11

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

dbaidya9006:
Fixes #7357

OpenEMR allows passwords at least 9 characters in length. To meet the ASVS 2.1.1 requirement, passwords under 12 characters should not be allowed. Additionally, OpenEMR does not have any constraints for the ‘password’ attribute of the ‘users_secure’ table to prevent bypassing the front-end validation and storing a password that does not meet minimum character requirements.

**After update:**
Screenshot of creating a new user with an 11-character password with a 12-character minimum password requirement enforced:
image

Screenshot of changing a password to a 11-character password with a 12-character minimum password requirement enforced:
image

Screenshot of adding a new row to the users_secure table with an 11-character password to the running database with a constraint enforcing a 12-character minimum length for the password attribute:
image

1. In globals.inc.php, update the default value for the global variable ‘gbl_minimum_password_length’ from ‘9’ to ‘12’. This variable is used in ‘AuthUtils.php’ to enforce a minimum password length requirement on the front-end.
2. In database.sql, add a constraint to the ‘password’ attribute of the ‘users_secure’ table to complete a check of the length of the password entered and confirm it is greater than or equal to 12. OpenEMR hashes the password entered in the application before storing the value in the database, so passwords should be much greater than 12 characters when saved from the front-end; however, if a user identifies a way to bypass the front-end validation to create/update a user (such as using a REST API endpoint), then the constraint on the attribute within MySQL will still preserve the minimum password requirement. This ensures OpenEMR does not solely rely on the front-end validation to protect the minimum password length requirement.

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -2117,7 +2117,7 @@ $GLOBALS_METADATA = array(
                 '19' => '19',
                 '20' => '20',
             ),
-            '9',                              // default
+            '12',                              // default
             xl('Minimum length of password.')
         ),
 
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -8925,7 +8925,8 @@ CREATE TABLE `users_secure` (
   `last_login_fail` datetime DEFAULT NULL,
   `auto_block_emailed` tinyint DEFAULT 0,
   PRIMARY KEY (`id`),
-  UNIQUE KEY `USERNAME_ID` (`id`,`username`)
+  UNIQUE KEY `USERNAME_ID` (`id`,`username`),
+  CONSTRAINT `min_password_length` CHECK ((LENGTH(`password`) >= 12))
 ) ENGINE=InnoDb;
 
 -- --------------------------------------------------------


bradymiller:
hi @dbaidya9006 ,
1. Doesn't NIST still only recommend 8 or greater for password size? Not against going to 12, but would check other standards to get an idea of where they all lie before making a decision on this.
2. Also no reason for the database constraint since it stores a hash that is always > 12. And a user would not be able to log in if there was not a hash stored there.

adunsulag:
Closing due to lack of response from original author

Issue #7275, opened 2024-03-18
Visit History's "Results per page" applies only to encounters, not documents


pete-boyd:
In OpenEMR 7.0.1 (1), in Visit History, if you set "Results per page", it only affects encounters. Any documents uploaded to the patient record are not included in the total figure defined by "Results per page", and appear at the end of the list of encounters.

E.g., set "Results per page: 5" and 5 encounters are listed, but then 8 of "Document:...".

sjpadgett:
@pete-boyd Yeah I can see how this would be annoying especially for paperless clinics that upload most everything dealing with a patient.
Maybe another range for documents included like the last 5 or latest. Or only docs attached to an encounter. I don't know how much attaching documents to encounters is used in the wild and would need feedback for that.
What would be your recommendation Pete?

pete-boyd:
Here's a description of how Visit History is being used:

"Some of the GP's that run our clinic like to see the full history from the EMR for each client (they do not have access to the EMR but our RGN running the clinic does). For this reason we routinely print the list of encounters for each client and that is given to the GP when the client arrives for appointment. So a clinic might have 10 appointments we print 10 pages, the client name and DOB and list of encounters."

In this scenario, I'm imagining that any info on documents is superfluous.

I've made an inquiry with regard to your questions though, and will get back to you.

ruthkonyn:
I think there are two issues:
1. the way the results per page count works is buggy - it reports say N results but displays N+3 entries, where there are 3 documents and N encounters. In some circumstances, such as the 'ALL' setting for results per page the only problem is the reported results count (documents are listed in date order amongst the encounters) , but for others, e.g. '5' or '20' there may be extra documents listed on the last page as well as those same documents included with the list of encounters, i.e. in date order. I'm looking at how to fix this bug so that documents are included in the results counts, and are listed in date order in the history, not at the end (or is the intended behaviour only to list non encounter attached documents at the end, not also in date order in the history?)
2. the other is whether a user would want documents listed at all, in this case enhance with another set of document options: 'do not include any documents' 'include only documents attached to an encounter' 'include all documents'

pete-boyd:
I received this in relation to documents appearing in Visit History:
"Actually useful for us to see, in visit history, if documents have been attached. For example when an existing/previously seen client presents to triage we check the visit history in the EMR to see if there is a Document xxxxx.pdf (patient id card) and use this as a check in our process."

PR #7359, opened 2024-04-16
 add print button and put patient's details at top of visit history page #7270

### LINKED ISSUE #7270: bug: In Visit History, print, or save as PDF, a multi-page list of encounters doesn't print or save all results.
In OpenEMR 7.0.1 (1), in Visit History, for a patient where there are numerous pages of results, Ctrl+P and then either printing, or saving as PDF, in any web browser on Windows or Linux, gives inconsistent results, and never successfully prints or saves the full list of encounters for the patient.

### To Reproduce

Use either "Visit History" icon in the toolbar / banner area or the "Visit History" tab open when a user logs in.

Results per page: ALL

Ctrl+P
Pages: All

A multi-page web page on other web sites works as expected with Ctrl+P.  So it appears to us to be an OpenEMR issue.

ruthkonyn:
Fixes #7270

to print or make a pdf of a visit history report one had to use the browsers print function, with ctrl+p or file/print etc. This only prints what had been displayed on the screen, so for a long history that takes several screens only the first would be printed.

add a 'print' button that prints the complete page of the history being displayed. for page length 'ALL' this is the complete history which will produce a multi page document where necessary
display the patient's details at the top of the history window and on the first page of the print or pdf document.

### DIFF
--- a/interface/patient_file/history/encounters.php
+++ b/interface/patient_file/history/encounters.php
@@ -310,12 +310,13 @@ window.onload = function() {
     <br />
     <span class="heading" >
     <?php
+    if ($attendant_type == 'pid') {
   // RM put patienes name, id and dob at top of the history -->
         $name =  getPatientNameFirstLast($pid);
         $dob =  text(oeFormatShortDate(getPatientData($pid, "DOB")['DOB']));
          $external_id = getPatientData($pid, "pubpid")['pubpid'];
         echo $name . " (" . $external_id . ")" .  "&nbsp;  &nbsp; DOB: " . $dob ;
-
+    }
     ?>
     </span>
 
--- a/interface/patient_file/history/encounters.php
+++ b/interface/patient_file/history/encounters.php
@@ -316,7 +316,7 @@ window.onload = function() {
          $external_id = getPatientData($pid, "pubpid")['pubpid'];
         echo $name . " (" . $external_id . ")" .  "&nbsp;  &nbsp; DOB: " . $dob ;
 
-        ?>
+    ?>
     </span>
 
     <div class="table-responsive">
@@ -390,7 +390,6 @@ window.onload = function() {
             }
 
             $numRes = 0;
-(new SystemLogger())->debug("size, start, numRes", array( $pagesize, $pagestart, $numRes));
 
             $sqlBindArray = array();
             if ($attendant_type == 'pid') {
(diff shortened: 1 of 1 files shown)


interface/patient_file/history/encounters.php
```diff
@@ -401,7 +422,7 @@ function changePageSize() {
 
             $countRes = sqlStatement($countQuery, $sqlBindArray);
             $count = sqlFetchArray($countRes);
-            $numRes = $count['c'];
+            $numRes += $count['c'];
```
adunsulag, line 425: It looks like this didn't actually change the code logic, was thinking it did, but on further examination it keeps the code the same.

interface/patient_file/history/encounters.php
```diff
@@ -297,6 +308,16 @@ function changePageSize() {
     </span>
 
     <br />
+    <span class="heading" >
+    <?php
+  // RM put patienes name, id and dob at top of the history -->
```
adunsulag, line 313: This breaks if the attendant_type is not a patient (IE therapy group encounters are being viewed.  We need to surround all of this:
```
 if ($attendant_type == 'pid') {
```

interface/patient_file/history/encounters.php
ruthkonyn: oh yes, apologies: - i've been working on this code to sort out the problems where the documents are displayed on every page, issue #7275, and this new line has crept in from that work. It's not needed simply to add the print functionality. I'll fix and resubmit with a new PR

ruthkonyn:
thank you so much stephen - i was just fixing these for a new PR when i noticed you'd done it already :-)

interface/patient_file/history/encounters.php
```diff
@@ -297,6 +307,17 @@ function changePageSize() {
...
 
     <br />
+    <span class="heading" >
+    <?php
+    if ($attendant_type == 'pid') {
+  // RM put patienes name, id and dob at top of the history -->
+        $name =  getPatientNameFirstLast($pid);
+        $dob =  text(oeFormatShortDate(getPatientData($pid, "DOB")['DOB']));
+         $external_id = getPatientData($pid, "pubpid")['pubpid'];
+        echo $name . " (" . $external_id . ")" .  "&nbsp;  &nbsp; DOB: " . $dob ;
```
bradymiller, line 317: need escaping:
```
echo text($name ). " (" . text($external_id) . ")" .  "&nbsp;  &nbsp; DOB: " . $dob ;
```

PR #7367, opened 2024-04-17
Adding Password Strength Meter Fixes #7366

### LINKED ISSUE #7366: Password strength meter missing
### Describe the problem

According to ASVS 2.1.8 there should be a password strength meter provided to help users set a stronger password which is missing in the application. This may be a security threat.

According to the ASVS, when a user is setting their password, a password strength meter should be provided.

Kashika08:

According to ASVS 2.1.8 there should be a password strength meter provided to help users set a stronger password which is missing in the application. The code changes provided here add a password strength meter which tells the users the strength of their password. Fixes the Issue https://github.com/openemr/openemr/issues/7366

The code makes changes to usergroup_admin_add.php file. It gives the javascript code for Password Strength Meter which is easily reusable. It adds it on the Add User page itself. 

#### Working Screenshots

image

image

image


interface/usergroup/usergroup_admin_add.php
```diff
@@ -645,6 +655,52 @@ function authorized_clicked() {
 
 });
 </script>
+
+<!-- Password Strength Meter JavaScript -->
+<script>
+function checkPasswordStrength() {
+    console.log("checkPasswordStrength called");
```
adunsulag, line 662: this shouldn't be here.

interface/usergroup/usergroup_admin_add.php
```diff
@@ -645,6 +655,52 @@ function authorized_clicked() {
...
+    var number = /([0-9])/;
+    var alphabets = /([a-zA-Z])/;
```
adunsulag, line 664: Your password strength algorithm is not multi-lingual.  Your regex assumes that everyone using the system will create passwords with only ASCII characters which is not correct as I'm pretty sure OpenEMR passwords can support unicode characters.  That's correct isn't it @bradymiller?

interface/usergroup/usergroup_admin_add.php
```diff
@@ -645,6 +655,52 @@ function authorized_clicked() {
...
+        } else if (pwd.match(number) && pwd.match(alphabets)) {
+            strength += 2;
+        } else if (pwd.match(alphabets)) {
+            strength += 1;
+        }
+
+        switch (strength) {
+            case 1:
+                document.getElementById('password_strength_meter').style.backgroundColor = "#ffcc00";
+                document.getElementById('password_strength_text').innerHTML = "<?php echo xlt('Weak'); ?>";
```
adunsulag, line 685: Please use innerText calls instead of innerHTML.

adunsulag (changes requested):

interface/usergroup/usergroup_admin_add.php
Kashika08: Resolved in my latest commit.

interface/usergroup/usergroup_admin_add.php
Kashika08: Resolved in my latest commit.

interface/usergroup/usergroup_admin_add.php
Kashika08: Resolved in my latest commit.

interface/usergroup/usergroup_admin_add.php
```diff
@@ -645,6 +655,51 @@ function authorized_clicked() {
...
+    var number = /(\p{N})/u;
+    var alphabets = /(\p{L})/u;
+    var special_characters = /([~,!,@,#,$,%,^,&,*,-,_,+,=,?,>,<])/;
+
+    var pwd = document.getElementById("stiltskin").value;
+    var strength = 0;
+
+    if (pwd.length < 6) {
+        document.getElementById('password_strength_meter').style.backgroundColor = "#ff6666";
+        document.getElementById('password_strength_text').innerHTML = "<?php echo xlt('Very Weak'); ?>";
```
bradymiller, line 671: innerText

also use xlj instead (and note dropping of the outside quotes (also do this for the 3 lines below):
```
document.getElementById('password_strength_text').innerHTML = <?php echo xlj('Very Weak'); ?>;
```

Kashika08: Fixed the issue

interface/usergroup/usergroup_admin_add.php
```diff
@@ -237,7 +237,17 @@ function authorized_clicked() {
 <tr>
```
bradymiller, line 237: also should do in the password change script at:
interface/usergroup/user_info.php

And would do in way to not duplicate code. Could add checkPasswordStrength function to the library/js/utility.js script. And can convert those lines with php (since will be in a javascript library file) to instead use the openemr javascript translation function, for example:
```
 document.getElementById('password_strength_text').innerText = xl('Very Weak');
```

bradymiller:
@adunsulag , yes, in theory any characters including emojis can be used in the password :)

bradymiller:
@adunsulag ,
Confirmed that following password with emojis works :)
🚀🚁🚂🚃🚄heythere8HEYTHERE!

interface/usergroup/usergroup_admin_add.php
```diff
@@ -645,6 +654,51 @@ function authorized_clicked() {
 
 });
 </script>
+
+<!-- Password Strength Meter JavaScript -->
+<script>
+function checkPasswordStrength() {
+    var number = /(\p{N})/u;
+    var alphabets = /(\p{L})/u;
+    var special_characters = /([~,!,@,#,$,%,^,&,*,-,_,+,=,?,>,<])/;
```
bradymiller, line 663: would it simply make more sense and broaden special characters (emojis etc) to have it just be:
special character = Not a Number or a Letter

Also, does this capture numbers/letters in other languages? (just curious)

Kashika08: Made the above change and yet it does capture numbers and letters in other languages.

interface/usergroup/usergroup_admin_add.php
Kashika08: Added the file to utility.js in the latest commit. 
It is working in both places now. 

image

interface/usergroup/user_info.php
```diff
@@ -127,6 +128,9 @@ function(data)
                         <label class='control-label col-sm-2'><?php echo xlt('New Password') . ":"; ?></label>
                         <div class='col-sm-3'>
                             <input type='password' class='form-control' name='newPass'  value="" autocomplete='off'>
+                            <!-- Password Strength Meter -->
+                            <div id="password_strength_meter" style="width:100%; height:15px; border:1px solid #ccc; margin-top: 5px;"></div>
```
bradymiller, line 132: don't use embedded styling. instead use bootstrap classes for styling (see this scripts and other scripts for examples)

interface/usergroup/usergroup_admin_add.php
```diff
@@ -237,7 +238,16 @@ function authorized_clicked() {
...
 <?php if (empty($GLOBALS['gbl_ldap_enabled']) || empty($GLOBALS['gbl_ldap_exclusions'])) { ?>
-<td style="width:150px;"><span class="text"><?php echo xlt('Password'); ?>: </span></td><td style="width:250px;"><input type="password" style="width:120px;" name="stiltskin" class="form-control"><span class="mandatory"></span></td>
+<td style="width:150px;">
+    <span class="text"><?php echo xlt('Password'); ?>:</span>
+</td>
+<td style="width:150px;">
+    <input type="password" style="width:120px;" name="stiltskin" id="stiltskin" class="form-control" onkeyup="checkPasswordStrength(this);">
+    <span class="mandatory"></span>
+    <!-- Password Strength Meter -->
+    <div id="password_strength_meter" style="width:120px; height:20px; border:1px solid #ccc; margin-top: 5px;"></div>
```
bradymiller, line 248: try not to use use embedded styling (looks likes will need the width:120px since used throughout in this script). instead use bootstrap classes for styling (see this scripts and other scripts for examples)

Kashika08:
> hi @Kashika08 , Getting close. Noted couple more issues (avoiding the embedded styling is not mandatory, but would be very nice since much easier to make future changes in script). After addressed, plan to then bring this in if it tests well.

Hello!
Thank you for your comments and helping me make the code better. I have addressed the comments and tested them on my end. I would be glad if the feature can be included in the application!

interface/usergroup/usergroup_admin_add.php
adunsulag: I think he's trying to stay consistent with the styling from line 238.  Probably why they copied it in the divs up higher in the script.

Kashika08:
Hi @bradymiller @adunsulag, 
I've resolved all the comments so far. Can you please check the code now and let me know soon if any other changes are required?

interface/usergroup/user_info.php
adunsulag: Looks like this is resolved.

interface/usergroup/usergroup_admin_add.php
```diff
@@ -42,6 +42,7 @@
 <?php Header::setupHeader(['common','opener', 'erx']); ?>
 
 <script src="checkpwd_validation.js"></script>
+<script src="../../library/js/utility.js"></script>
```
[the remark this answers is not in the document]
adunsulag: Looks like this is resolved.

adunsulag (approved):

adunsulag:
@bradymiller This looks good to me at this point.  If you're fine as well, we can go ahead and merge this assuming everything passes on the CI engine.

bradymiller (approved):

bradymiller:
@Kashika08 , Just brought this into codebase. Congrats and thanks for the contribution!

Issue #7378, opened 2024-04-20
Allows user to select Date of Birth in the future


riyagori1203:
### Describe the bug

On creating a new patient and adding the insurance information, we notice that openemr allows the user to enter a future date of birth in the insurance fields.

### To Reproduce

1. Create a new Patient.
2. Add necessary details
3. Scroll down to Insurance information for the new patient.
4. Try adding date of birth as a future date. Eg: try putting it as 06/06/2065
5. CLick on Save new patient

### Expected behavior
OpenEmr must have basic validation check in order to prevent the user from entering a date of birth in the future

### Client configuration

- Browser: Google Chrome/FireFox
- Operating system: MacOs/Windows/Linux

PR #7379, opened 2024-04-20
Fix: to not allow user to pick future date for DOB

riyagori1203:
Fixes https://github.com/openemr/openemr/issues/7378

It would not allow the user to enter a future date of birth in the insurance field which is a basic validation check.

Just the addition of a word -past that disables any future dates for the user. So that user can only select dates previous to the current date


sjpadgett:
tested on master dev okay although I don't understand why this fix also wasn't added to new Patient validation for DOB
For those that don't know the past datepicker will grey out future dates in UI picker.

PR #7386, opened 2024-04-22
Visit history - correct count for displaying ALL results on a page

### LINKED ISSUE #7385: visit history: count for display of ALL results incorrect, if shorter pages displayed previously
### Describe the bug
in visit history
sometimes when displaying 'ALL' the results, lets say n results, the heading which should say  "1-n of n" is incorrect

### To Reproduce
1. in visit history
2. choose 'results per page' to be '20' (or a number less than the total number of available results)
3. click 'next'
4. choose 'results per page' to be 'ALL'
the heading is incorrect, in this example in the screen shot it ought to read '1-27 of 27'

Screenshot from 2024-04-22 16-41-13
 

- Browser: firefox
- OpenEMR version: 7.0.3-dev
- Operating system: ubuntu linux

ruthkonyn:
Fixes #7385

when displaying all the results on a page, having previously displayed a shorter page, give the correct count of results being displayed


sjpadgett:
@ruthkonyn unfortunately this PR has conflicts with encounter. I hoped to get in patch. Maybe you can get resolved early Monday morning before I build the patch!

sjpadgett:
Doesn't make sense for Styling to fail after I made and merged a fixed file so either git is trying to merge an older file(a no no) or actions is testing the older version.
So I'm going to test the theory so we'll know and merge then pull the file. My changes should still be there.

sjpadgett:
Okay looks like actions is based on the last master merge into branch otherwise it should have picked up my merged changes.
My first question is why we don't just test against file in PR i.e. Stylings only on current PR. Some test may need to run against other files but I'm hoping there's a better way of what files we test.

ruthkonyn:
thank you @sjpadgett  - so there is nothing i need to do now :-) . Thanks for accepting into the patch

PR #7341, opened 2024-04-11
Fix: OpenEMR logs sensitive information such as payment details

jubittajohn:
Fixes #7340 

#### Short description of what this resolves:
This PR adds a log scrubbing template to mask sensitive details in the log(which is currently exposed in logs). Currently, this template only masks the field "reference" in the table "ar_session" to conceal sensitive payment details in the log. However, if approved, this template can be used to extend the masking of other sensitive details to other tables and fields too. Masking of sensitive information in logs is essential to prevent the stealing of information. (ASVS 7.1.1).

After update: 
log-sensitive-payment-detail-after-fix

#### Changes proposed in this pull request:
1. In "EventAuditLogger.php", add a class constant to keep track of the tables and the fields in each table that contains sensitive information.
2. In "EventAuditLogger.php", for any insert operation, if the current table contains any field containing sensitive information, the value for that field is masked to match its length.

PR #7396, opened 2024-04-23
Fix: Revert "Fix: OpenEMR logs sensitive information such as payment details (#7341)"

bradymiller:
Turns out probably shouldn't mess with log information to ensure remain compatible with ONC certification.

### DIFF
--- a/src/Common/Logging/EventAuditLogger.php
+++ b/src/Common/Logging/EventAuditLogger.php
@@ -144,14 +144,6 @@ MSG;
 </ParticipantObjectIdentification>
 MSG;
 
-
-    /**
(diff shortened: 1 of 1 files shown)


bradymiller:
@adunsulag , @stephenwaite , 
I was not able to find obvious language in the ONC certification that stated unable to modify things that are logged, however there was language about not modifying things that have been logged. Since not clear, seems reasonable to just not mess with things that are being logged. Note when we are inserting sensitive things like credentials or checking credentials, we use a sql function that skips logging (so we have a mechanism to deal with when running queries that contain information that should not be logged).

adunsulag (approved):

bradymiller:
also will revert this in rel-702

PR #7408, opened 2024-04-25
Weno better error handling from fetches

sjpadgett:
- this needs a rewrite but maybe later

Making some errors easier to understand.

### DIFF
--- a/interface/modules/custom_modules/oe-module-weno/src/Services/LogProperties.php
+++ b/interface/modules/custom_modules/oe-module-weno/src/Services/LogProperties.php
@@ -139,7 +139,7 @@ class LogProperties
     /**
      * @throws Exception
      */
-    public function logSync()
+    public function logSync($tasked = 'background')
     {
         $wenoLog = new WenoLogService();
         $provider_info['email'] = $this->weno_admin_email;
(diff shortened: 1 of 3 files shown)


interface/modules/custom_modules/oe-module-weno/src/Services/LogProperties.php
```diff
@@ -167,9 +167,20 @@ public function logSync()
...
                 EventAuditLogger::instance()->newEvent("prescriptions_log", $_SESSION['authUser'], $_SESSION['authProvider'], 0, ($error));
-                die(js_escape($error));
+                // if background task then return false
+                if ($tasked == 'background') {
+                    $wenoLog->insertWenoLog("prescription", "Invalid Prescriber Credentials Background Service.");
+                    return false;
+                }
+                if ($tasked == 'downloadLog') {
+                    $wenoLog->insertWenoLog("prescription", "Invalid Prescriber Credentials User Download.");
+                    echo(js_escape($error));
```
bradymiller, line 178: would use text() instead of js_escape()

interface/modules/custom_modules/oe-module-weno/src/Services/LogProperties.php
```diff
@@ -167,9 +167,20 @@ public function logSync()
...
+                    exit;
+                }
+                $wenoLog->insertWenoLog("prescription", "Invalid Prescriber Credentials Hint:see previous errors.");
+                echo(js_escape($error));
```
bradymiller, line 182: would use text() instead of js_escape()

interface/modules/custom_modules/oe-module-weno/src/Services/LogProperties.php
sjpadgett: Needs to be js escaped as it is a return to ajax downloadPresLog().
It gets escaped with 
```
let errorMsgSpan = document.getElementById('error-msg');
errorMsgSpan.textContent = jsText(data);
```

bradymiller: Ajax server scripts still need to do html escaping (in case a bad actor calls them directly from browser) with text/attr. js_escape won't prevent xss on these if bad actor calls the ajax server script directly from browser. Then will also need to re-escape it when run it on the ajax client script when bring it into html element (like you are doing). Escaping is fun stuff :)

sjpadgett: not going to do!

PR #7418, opened 2024-05-01
Bring into patch 1 2420E #7405

sjpadgett:
See PR #7405

### DIFF
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -208,8 +208,6 @@ INSERT INTO `background_services` (`name`, `title`, `active`, `running`, `next_r
 INSERT INTO `background_services` (`name`, `title`, `active`, `running`, `next_run`, `execute_interval`, `function`, `require_once`, `sort_order`) VALUES
 ('X12_SFTP', 'SFTP Claims to X12 Partner Service', 0, 0, '2021-01-18 11:25:10', 1, 'start_X12_SFTP', '/library/billing_sftp_service.php', 100);
 INSERT INTO `background_services` (`name`, `title`, `active`, `running`, `next_run`, `execute_interval`, `function`, `require_once`, `sort_order`) VALUES
-('WenoExchange', 'Weno Log Sync', 0, 0, '2021-01-18 11:25:10', 0, 'start_weno', '/library/weno_log_sync.php', 100);
-INSERT INTO `background_services` (`name`, `title`, `active`, `running`, `next_run`, `execute_interval`, `function`, `require_once`, `sort_order`) VALUES
 ('UUID_Service', 'Automated UUID Creation Service', 1, 0, '2021-01-18 11:25:10', 240, 'autoPopulateAllMissingUuids', '/library/uuid.php', 100);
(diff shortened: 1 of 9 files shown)


sql/patch.sql
```diff
@@ -55,3 +55,7 @@ UPDATE list_options SET option_id='DOB' WHERE list_id='recent_patient_columns' A
 #IfMissingColumn form_encounter last_update
 ALTER TABLE `form_encounter` ADD `last_update` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP;
 #EndIf
+
+#IfMissingColumn form_encounter ordering_provider_id
+ALTER TABLE `form_encounter` ADD `ordering_provider_id` INT(11) DEFAULT '0' COMMENT 'ordering provider, if any, for this visit';
+#EndIf
```
stephenwaite, line 61: also need this in `database.sql`?

sjpadgett: We don't modify production database so that installing DB below the patch level could break.
DB changes should always be at patch level version via patch.sql.

stephenwaite: But this is rel-702 as well so need it.

sjpadgett:
Can't quite figure out why your mod's cause EncounterService to fail!

stephenwaite:
Maybe a database problem 😀 ?

sjpadgett:
I think we need the database that has patched changes for tests. 
I don't like the idea that dependencies don't follow patch level but I'm not going to rewrite the test.
Sorry @stephenwaite

sjpadgett:
Are we okay with this now?

sjpadgett:
I forgot. while looking over encounterservice I found a bug in that `$result->getData() in array_pop($result->getData());`  is passed as reference which must be an array and not a returned array from call.
It's used in `public function getMostRecentEncounterForPatient($pid)` which currently shows no usages.
It prob just creates a warning for now.

sql/patch.sql
bradymiller: hi , reason we also do database.sql is to support new installs. Need for dockers (since we update docker with most recent patch) and when we need to rebuild the packages from a patch.

stephenwaite:
smaller

sql/database.sql
```diff
@@ -8053,7 +8014,7 @@ DROP TABLE IF EXISTS `rule_filter`;
 CREATE TABLE `rule_filter` (
   `id` varchar(31) NOT NULL DEFAULT '' COMMENT 'Maps to the id column in the clinical_rules table',
   `include_flag` tinyint(1) NOT NULL default 0 COMMENT '0 is exclude and 1 is include',
-  `required_flag` tinyint(1) NOT NULL default 0 COMMENT '0 is required and 1 is optional',
+  `required_flag` tinyint(1) NOT NULL default 0 COMMENT '0 is optional and 1 is required',
```
bradymiller, line 8017: I think this is reverting one of Hank's fixes. Not a biggie since just a comment and most important that remains as is in master.

PR #7423, opened 2024-05-05
Hide dashboard card 2

juggernautsei:
Fixes #7192 

Hide unutilized dashboard cards

### DIFF
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -55,9 +55,6 @@ use OpenEMR\Services\ImmunizationService;
 use OpenEMR\Services\PatientIssuesService;
 use OpenEMR\Services\PatientService;
 use Symfony\Component\EventDispatcher\EventDispatcher;
-use OpenEMR\Patient\Cards\InsuranceViewCard;
-use OpenEMR\Patient\Cards\BillingViewCard;
-use OpenEMR\Patient\Cards\DemographicsViewCard;
 
 $twig = new TwigContainer(null, $GLOBALS['kernel']);
 
@@ -79,6 +76,7 @@ if (isset($_GET['set_pid'])) {
 // want smart support in their system.
 $smartLaunchController = new SMARTLaunchController($GLOBALS["kernel"]->getEventDispatcher());
 $smartLaunchController->registerContextEvents();
+$hiddenCards = getHiddenDashboardCards();
 
 /**
  * @var EventDispatcher
@@ -119,6 +117,17 @@ if ($GLOBALS['insurance_only_one']) {
     $insurance_array = array('primary');
 } else {
     $insurance_array = array('primary', 'secondary', 'tertiary');
+}
+
+function getHiddenDashboardCards(): array
+{
+    $hiddenList = [];
+    $ret = sqlStatement("SELECT gl_value FROM `globals` WHERE `gl_name` = 'hide_dashboard_cards'");
+    while ($row = sqlFetchArray($ret)) {
+        $hiddenList[] = $row['gl_value'];
+    }
+
+    return $hiddenList;
 }
 
 function print_as_money($money)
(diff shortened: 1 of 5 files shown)


sjpadgett:
Thanks @juggernautsei I assume you had an opportunity to test too?
Looks okay to me but could use another set of eyes in case I've missed something. @bradymiller if you get a chance maybe check this out for us. thanks

juggernautsei:
@sjpadgett I did test. Flawless

interface/patient_file/summary/demographics.php
```diff
@@ -1038,9 +1047,9 @@ function setMyPatient() {
                     <?php
                     $t = $twig->getTwig();
 
-                    $allergy = (AclMain::aclCheckIssue('allergy')) ? 1 : 0;
-                    $pl = (AclMain::aclCheckIssue('medical_problem')) ? 1 : 0;
-                    $meds = (AclMain::aclCheckIssue('medication')) ? 1 : 0;
+                    $allergy = (AclMain::aclCheckIssue('allergy') ? 1 : 0) && !in_array('card_allergies', $hiddenCards) ? 1 : 0;
```
bradymiller, line 1050: won't this result in  boolean instead of 1 or 0. Wouldn't this make more sense:
```
$allergy = (AclMain::aclCheckIssue('allergy') && !in_array('card_allergies', $hiddenCards)) ? 1 : 0;
```

interface/super/edit_globals.php
```diff
@@ -645,6 +645,38 @@ function checkBackgroundServices()
...
                                                 echo "  </select>\n";
+                                            } elseif ($fldtype == GlobalSetting::DATA_TYPE_MULTI_DASHBOARD_CARDS) {
+                                                $hiddenList = [];
+                                                $ret = sqlStatement("SELECT gl_value FROM `globals` WHERE `gl_name` = 'hide_dashboard_cards'");
+                                                while ($row = sqlFetchArray($ret)) {
+                                                    $hiddenList[] = $row['gl_value'];
+                                                }
+                                                // The list of cards to hide. For now add to array new cards.
+                                                $res = array(
+                                                    ['card_abrev' => '', 'card_name' => xlt('None or Reset')],
```
bradymiller, line 656: how does this work for a multiselect? Couldn't they select this and still select other items?

interface/super/edit_globals.php
```diff
@@ -645,6 +645,38 @@ function checkBackgroundServices()
...
+                                                    ['card_abrev' => attr('card_allergies'), 'card_name' => xlt('Allergies')],
+                                                    ['card_abrev' => attr('card_amendments'), 'card_name' => xlt('Amendments')],
+                                                    ['card_abrev' => attr('card_disclosure'), 'card_name' => xlt('Disclosures')],
+                                                    ['card_abrev' => attr('card_lab'), 'card_name' => xlt('Labs')],
+                                                    ['card_abrev' => attr('card_medicalproblems'), 'card_name' => xlt('Medical Problems')],
+                                                    ['card_abrev' => attr('card_medication'), 'card_name' => xlt('Medications')],
+                                                    //['card_abrev' => 'card_rx', 'card_name' => 'Prescriptions'], // For now don't hide because can be disabled as feature.
```
bradymiller, line 663: up to you, but i think ok if you want to also include prescriptions in this feature

juggernautsei: @bradymiller prescriptions already have its own disable feature. No need to include it here.

sjpadgett:
thank you @bradymiller as always very insightful review. I just hope my memory serves correct for truthy:) will check anyway cause I have a pr for some bug fixes going up.

interface/super/edit_globals.php
juggernautsei: But I think I see @bradymiller point is to move the old turn-off feature into here so that people don't have to go to two spots to turn off a card.

sjpadgett: I agree but I left to others to decide and I asked admins on signal is why Brady chimed in. but i'll add in my PR. thanks

PR #7429, opened 2024-05-07
Module fixes and updates

### LINKED ISSUE #7430: etherFax fails render on large download log.
### Describe the bug
Happens on a users site with the only error trapped is an out of memory during table render from log of 165 items.

I don't think this is actual cause but I'm still investigating.

### LINKED ISSUE #7436: Need tiff image support and image to PDF
### Describe the bug
Mostly for fax however new classes will support any image type to PDF
Plan to look for installed support dependencies allowing three options for all except tiff or all including tiff
- GD 2
- ImageMagick
- Javascript browser support.

All support PDF generation with new general use Mpdf class exending base Mpdf class. There's advantages here.

sjpadgett:
- PHP warnings

Fixes #7430 
Fixes #7436


sjpadgett:
Not ready for review yet.

sjpadgett:
This is done but I'm still testing on various environment:)
Hopefully peer reviews goes well as I'm greatly tired of looking at modules.
I may still add a fix for @claimrevolution  adding listener to turn off background tasks on module disable. Just had a user try the module but then disabled however error log keeps getting flooded with failed background tasks. Also tasks set to 1 minute is really 2 minutes which is what background task dispatcher is called so, there is that. Brad please get back to me concerning this because it become a real bad issue for some.

PR #7431, opened 2024-05-09
Fix for non defaut database port usage

wjassell:
We utilize a non standard database port and this file was not using the globals port in all areas

Fixes #

### DIFF
--- a/interface/modules/zend_modules/config/autoload/global.php
+++ b/interface/modules/zend_modules/config/autoload/global.php
@@ -17,96 +17,80 @@
 
 use OpenEMR\Common\Crypto\CryptoGen;
 
-// If to use utf-8 or not in my sql query
+// Set SQL mode and character set based on utf-8 settings
 if (!$GLOBALS['disable_utf8_flag']) {
-    if (!empty($GLOBALS["db_encoding"]) && ($GLOBALS["db_encoding"] == "utf8mb4")) {
-        $tmp = "SET NAMES 'UTF8MB4', sql_mode = ''";
-    } else {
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/config/autoload/global.php
```diff
@@ -17,112 +17,91 @@
...
 
-// This settings can be change in the global settings under security tab
-$adapters = array();
+// Set up additional adapters if multiple databases are supported
+$adapters = [];
 if (!empty($GLOBALS['allow_multiple_databases'])) {
-    // Open pdo connection
-    $dbh = new PDO('mysql:dbname=' . $GLOBALS['dbase'] . ';host=' . $GLOBALS['host'], $GLOBALS['login'], $GLOBALS['pass']);
+    // Open PDO connection
+    $dbh = new PDO("mysql:dbname={$GLOBALS['dbase']};host={$GLOBALS['host']};port={$GLOBALS['port']}", $GLOBALS['login'], $GLOBALS['pass']);
```
kojiromike: Good fix adding the port to the DSN here! This is the key change that makes non-standard ports work.

However, since main branch now uses `$sqlConf` instead of individual globals, this would need to be updated to use `$sqlConf['port']` after rebasing.

kojiromike:
Closing in favor of #10636, which is a fresh implementation on current master.

This PR became too stale to merge cleanly (21 months of upstream changes). The new PR applies the same fix: including the database port in PDO DSN strings.

Thank you @wjassell for identifying this issue!

PR #7438, opened 2024-05-17
Previous names in New/Search doesn't save.

### LINKED ISSUE #7437: Add previous name doesn't save New/Search
### Describe the bug
Because the Previous Name plug in will save name when added and not on form save, the new name gets saved to current patient opened pid because a new pid doesn't get generated until saved.
There are a couple hacky ways to get around this limitation, I chose to turn it off.

### To Reproduce

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

sjpadgett:
- Added an alert disabling the capability in create patient.

Fixes #7437

Because the Previous Name plug in will save name when added and not on form save, the new name gets saved to current patient opened pid because a new pid doesn't get generated until saved.
There are a couple hacky ways to get around this limitation, I chose to turn it off.
forum issue.

### DIFF
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -26,6 +26,7 @@
  */
 
 require_once("../../globals.php");
+
 require_once("$srcdir/lists.inc.php");
 require_once("$srcdir/patient.inc.php");
 require_once("$srcdir/options.inc.php");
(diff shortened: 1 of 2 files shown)


sjpadgett:
image

bradymiller:
nice!

dance-bear

PR #7442, opened 2024-05-20
Module uninstall

### LINKED ISSUE #7441: bug: Claimrev leaves background service on when module is uninstalled
Need to cleanup the removal of the background service if the module is uninstalled.

Added the module uninstall feature to the claim rev module to clean up after removal of module from service.
Minor UI work was added.

juggernautsei:
Fixes #7441


sjpadgett:
@claimrevolution Need you to look this over today. Want for patch!

interface/modules/custom_modules/oe-module-claimrev-connect/ModuleManagerListener.php
```diff
@@ -0,0 +1,232 @@
...
+     * @param $modId
+     * @param $currentActionStatus
+     * @return mixed
+     */
+//    private function disable($modId, $currentActionStatus): mixed
+//    {
+//        // allow config button to show before enable.
+//        ModuleServices::setModuleState($modId, '0', '1');
+//        return $currentActionStatus;
+//    }
```
sjpadgett, line 152: uncomment method and add action to disable tasks because module is now disabled. You can make this cleaner or use below. Just use same for enable action by setting active to 1. Then unregister action will delete tasks so default  table.sql will reinstall.
```
UPDATE `background_services` SET `active` = '0' WHERE `background_services`.`name` = 'ClaimRev_Elig_Send_Receive'; 
UPDATE `background_services` SET `active` = '0' WHERE `background_services`.`name` = 'ClaimRev_Receive';
UPDATE `background_services` SET `active` = '0' WHERE `background_services`.`name` = 'ClaimRev_Send';
```

interface/modules/custom_modules/oe-module-claimrev-connect/ModuleManagerListener.php
```diff
@@ -0,0 +1,232 @@
...
+        sqlQuery($sql, array('ClaimRev_Send', 'ClaimRev_Receive', 'ClaimRev_Elig_Send_Receive'));
+        return $currentActionStatus;
+    }
+
+    /**
+     * @param $modId
+     * @param $currentActionStatus
+     * @return mixed
+     */
+//    private function reset_module($modId, $currentActionStatus): mixed
```
sjpadgett, line 171: If needed or wanted, this gets called also when unregistered action is called.

sjpadgett (changes requested):
Just need to finish the task handling for enable and disable actions. Thanks for taking care of the bootstrap includes. I'm surprised we let that slide when bringing module into core.
I'm not testing and leaving that to you.

interface/modules/custom_modules/oe-module-claimrev-connect/ModuleManagerListener.php
```diff
@@ -0,0 +1,232 @@
...
+//    {
+//        $modService = new ModuleServices();
+//        /* setting the active ui flag here will allow the config button to show
+//         * before enable. This is a good thing because it allows the user to
+//         * configure the module before enabling it. However, if the module is disabled
+//         * this flag is reset by MM.
+//        */
+//        $modService::setModuleState($modId, '0', '1');
+//        return $currentActionStatus;
+//    }
```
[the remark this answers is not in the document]
juggernautsei: I was unsure of the role of the comment methods. I commented them out to make sure they were not causing the issue that I was facing at the time. I will uncomment. I think that will be best for future utilization

interface/modules/custom_modules/oe-module-claimrev-connect/ModuleManagerListener.php
juggernautsei: When I enabled all the methods that were commented out. The module will not install. It can enable but it won't install. I have uploaded what I have. The code does not produce any errors in the errors in the logs.

juggernautsei: I will try to figure this out tomorrow. I got some other stuff to work on right now.

sjpadgett: @juggernautsei Just delete the unused methods(commented). The controller will simply ignore them. You need unregister, enable and disable for task maintenance.

juggernautsei: I figured out what you were saying and just return current status. I have installed and removed. It removes all the globals and tables.

sjpadgett: I wanted a cleanup action that I chose to name reset i.e. reset module to a state before original install. All these still need to be documented but besides our admins, you and Brad, I've seen no third party user make any attempt to give feedback or contrib.
So, you're in the know now so to say:)

interface/modules/custom_modules/oe-module-claimrev-connect/ModuleManagerListener.php
sjpadgett: @juggernautsei You still need the enable and disable tasks i.e active flag. If user disables then background tasks don't have access to module so need to make tasks inactive. Same with reverse and enable action. You're only half way there.
See the sql example I gave above for disable.

interface/modules/custom_modules/oe-module-claimrev-connect/ModuleManagerListener.php
juggernautsei: @sjpadgett I move the getModuleState method into the listener. The reset_module method is not being called. I checked the error log and it does not show there. The background services are being removed but not the table or globals as you would know.

sjpadgett: Forget reset_module just delete it you won't be using.
You don't need getModuleState. I wrote those for how I use weno module actions.
All you need is enable method with the sql to set the 3 active flag in background services, disable to unset active flag in background service table and what you already have to delete the tasks.

juggernautsei: got it and here are the updates.

sjpadgett:
Thanks for doing Brads job here @juggernautsei I appreciate it as well as all those users that enable the module then decide to disable.
I've seen on a user site this issue causing one to three fatal errors every time background service ran every one to two minutes.

I'm looking into making this class mandatory or module will not get initialized from boot.
Anyway good job my friend and I'll be merging

sjpadgett:
woops! forgot to squash!

juggernautsei:
It is always a learning experience with you. I seek after as much. I will go back to some of my modules and revamp them to include this basic feature request starting with the Quest module. Always improving.

PR #7452, opened 2024-05-22
Backport etherfax #7429 #7447

sjpadgett:
- etherfax changes


sjpadgett:
I think and hope I am done with etherfax. I'll need to add instructions for user of module that they need to run npm i to install tiff library if they want viewers to work with tiff fax. Or they can install imagick ext and I automatically use it instead of missing tiff library.

PR #7463, opened 2024-06-02
Add new modules loaded event

### LINKED ISSUE #7462: Need module bootstrap status events
### Describe the problem
In preparation for a background system module monitor development I am gradually adding useful events to eval module behavior.

sjpadgett:
- Notify any listeners that modules are loaded.
- Include modules that failed and reason in modules event return array.

Fixes #7462

Refactor our src/Core/ModulesApplication.php bootstrapCustomModules method to use a new method isFileReadableWithRetry() to retry if readable fails. Default is three tries. I fear these type of problem occur more often than we notice:

- File System Latency
- File locks
- Disk I/O Issues

I'm still looking into if or when, file read/write errors may occur. I don't know of openemr using file locks anywhere and if anything maybe file contention or timeouts is more likely. However you look at it, it is an interesting issue.
I will still, for now, keep the disable module if this check fails. I'd like to add a new runtime status to the modules entry in modules table but I'm still thinking over a possible module rehab/monitor backend task application.


sjpadgett:
Decided to remove module disable. Will merge after a few more tests.

sjpadgett:
Installed on an effected site so will wait until I see what happens. Strange for the file read to fail and must have something to do with environment or our app holding it's tongue just right.:)
I sure wish I could recreate locally or anywhere else.

sjpadgett:
Working towards international support for the weno module openemr version. I've been able to cut full pharmacy of around 92,300 records download and import to table to around 25seconds. A long way from original time of 1.5+ minutes!

Greatly improved search engine and getting closer to class cleanup to remove unused or spaghetti code and give better clarity for those that follow me.

I'll still have more code reduction coming in next day or two.

If anyone wants to play with then it's on GAMMA dev demo. You have credentials if not let me know and I'll send.
A lot of etherfax here too. It's working very nicely. Again let me know for credentials. All these accounts are setup for the project to use going forward for lifecycle updates.

stephenwaite:
smaller_excited_elephant

sjpadgett:
image

PR #7488, opened 2024-06-16
picks for patch2

sjpadgett:
Fixes #


interface/modules/custom_modules/oe-module-weno/scripts/weno_pharmacy_search.php
```diff
@@ -91,22 +92,28 @@
...
-        );
+        if (strlen($row['ncpdp_safe']) < 8) {
+            $return_arr[] = array(
+                "name" => $row['Business_Name'] . " " . $row['address_line_1'] . " " . $row['city'] . ", " . $row['state'],
+                "ncpdp" => $row['ncpdp_safe']
+            );
+        }
     }
-    echo text(json_encode($return_arr));
+    echo json_encode($return_arr);
```
bradymiller, line 112: Was text() call breaking something here? Hopefully not since we do this frequently (html escape a json_encoding value when echo'ing into html context)

interface/modules/custom_modules/oe-module-weno/scripts/weno_pharmacy_search.php
```diff
@@ -143,18 +156,21 @@
...
             $return_arr[] = array(
-                "name" => ucwords(strtolower($row['Business_Name'] . " " . $row['address_line_1'] . " " . $row['city'])),
-                "ncpdp" => $row['ncpdp']
+                "name" => $row['Business_Name'] . " " . $row['address_line_1'] . " " . $row['city'] . ", " . $row['state'],
+                "ncpdp" => $row['ncpdp_safe']
             );
         }
     }
-    echo text(json_encode($return_arr));
+    echo json_encode($return_arr);
```
bradymiller, line 175: same question here on above regarding the text() call

interface/modules/custom_modules/oe-module-weno/sql/1_0_0-to-2_0_0_upgrade.sql
```diff
@@ -0,0 +1,92 @@
...
+     `Country_Code` varchar(64) DEFAULT NULL,
+     `International` varchar(5) DEFAULT NULL,
+     `Latitude` varchar(255) DEFAULT NULL,
+     `Longitude` varchar(255) DEFAULT NULL,
+     `Pharmacy_Phone_safe` varchar(24) DEFAULT NULL,
+     `Test_Pharmacy` varchar(15) DEFAULT NULL,
+     `State_Wide_Mail_Order` varchar(15) NOT NULL,
+     `Mail_Order_US_State_Serviced` varchar(255) DEFAULT NULL,
+     `Mail_Order_ US_Territories_Serviced` varchar(255) DEFAULT NULL,
+     `On_WENO` tinytext DEFAULT NULL,
```
bradymiller, line 30: best to have no modifiers for 'text' fields

interface/modules/custom_modules/oe-module-weno/src/Services/DownloadWenoPharmacies.php
```diff
@@ -12,11 +12,149 @@
...
+    {
+        $wenoLog = new WenoLogService();
+
+        if (date("l") == "Monday" && $isInsertOnly) {
+            $sql = "TRUNCATE TABLE weno_pharmacy";
+            sqlStatement($sql);
+        }
+
+        // Database configuration. Create connection.
+        $connect = new mysqli($GLOBALS["host"], $GLOBALS["login"], $GLOBALS["pass"], $GLOBALS["dbase"]);
```
bradymiller, line 47: Just checking why you needed to do use direct mysqli calls.
Couple issues that will come up (can be dealt with with example in how done in setup.php script):
1. Setting of encoding
2. Breaking if mysql is set up to enforce encryption over wire

bradymiller: Actually, code is in Installer.class.php and not setup.php

interface/modules/custom_modules/oe-module-weno/table.sql
```diff
@@ -4,46 +4,40 @@
...
+     `Country_Code` varchar(64) DEFAULT NULL,
+     `International` varchar(5) DEFAULT NULL,
+     `Latitude` varchar(255) DEFAULT NULL,
+     `Longitude` varchar(255) DEFAULT NULL,
+     `Pharmacy_Phone_safe` varchar(24) DEFAULT NULL,
+     `Test_Pharmacy` varchar(15) DEFAULT NULL,
+     `State_Wide_Mail_Order` varchar(15) NOT NULL,
+     `Mail_Order_US_State_Serviced` varchar(255) DEFAULT NULL,
+     `Mail_Order_ US_Territories_Serviced` varchar(255) DEFAULT NULL,
+     `On_WENO` tinytext DEFAULT NULL,
```
bradymiller, line 36: best for text fields to have no modifiers

interface/modules/custom_modules/oe-module-weno/src/Services/DownloadWenoPharmacies.php
sjpadgett: yeah I simply forgot to go back to setting names. I was going to used same as we used in portal and got side tracked but you're correct this can cause issues that could be very hard to track down.

The reasons for direct management are of concern for me and something we need to address. 1. Our wrappers are slow. 2. If I was just importing from UI I wouldn't care too much however I've found out how easy running a database intensive script in background services can cause problem for front facing foreground runtime. It can hang foreground enough to be noticed by user where they can start clicking randomly while trying to figure out why app is hung.

The performance difference is quit noticeable by a factor of 50% which I benched marked.

I'm going to once again warn everyone that we need to take a close look at what we run in these background tasks.
My experience looking at a couple production sites that have several modules turn on plus our house keeping tasks like the UUID update, direct email polling, X12 severely affect openemr runtime from UI POV. 

What I've done here for this query intensive transaction needs to be an example of doing the same for some of our other queries that affect performance.

Having our current wrappers are/were a nice convenience for the type of workloads expected 10 years ago but as openemr grows with far greater data processing requirements and moving more to the enterprise, we have to/must put an effort into profiling and tuning openemr if we are going to compete for large enterprise requirements.
I haven't even brough into the affect API's will have as well.

I just don't want to be the only admin that's investigating this stuff or not worried about. We have to get used to having much larger data sets for testing with more realistic options turned on.

Anyway thanks for the catch Brady. I knew about but simply forgot!

interface/modules/custom_modules/oe-module-weno/scripts/weno_pharmacy_search.php
sjpadgett: Yes there can be issues escaping with text here.  Main issue is this data comes from database and has special characters like '&' that can be part of a pharmacy name and thus you wind up with mangled(noticeable) escaped characters in select.
I let it slide but got called out about it twice but Weno testers.
I've had other times where wrapping with text can be an issue but not enought to be concerned and just have to be aware where used.

interface/modules/custom_modules/oe-module-weno/table.sql
sjpadgett: is supposed to be tinyint:)

interface/modules/custom_modules/oe-module-weno/src/Services/DownloadWenoPharmacies.php
sjpadgett: @bradymiller Can you help me out with this? Not sure best way to open connection and my brain is so tired...

sjpadgett: never mind @bradymiller I found a way that's an easy compromise. Going to use the existing connection from globals

PR #7492, opened 2024-06-18
Weno Search Tuning

sjpadgett:
- trigger persisted filters on form entry
- add refresh button to widget


interface/modules/custom_modules/oe-module-weno/templates/indexrx.php
```diff
@@ -188,16 +219,32 @@
             </div>
         </div>
         <div class="container-xl mt-3">
-            <iframe id="wenoIfram"
-                title="Weno IFRAME"
-                width="100%"
-                height="900"
-                src="<?php echo $urlOut; ?>">
-            </iframe>
+            <iframe id="wenoIfram" title="Weno IFRAME" width="100%" height="900" src="<?php echo $urlOut; ?>"></iframe>
```
bradymiller, line 222: Pretty sure $urlOut will need to be escaped with attr() to prevent cross scripting attack via the $provider_info['email'] variable.

PR #7527, opened 2024-06-26
Fixes #7526 labcore fix date collected issues

### LINKED ISSUE #7526: bug: Lab Core Procedure Manual eReq breaks due to PHP8 date issue if date_collected is not filled in
If you attempt to use the manual eReq button when doing a labcorp procedure order the pdf fails to generate if you don't put in a date_collected.  Since the date_collected are an empty string the php date functions blow up.

image

adunsulag:
Made it so the date collected issue works when doing the manual ereq for labcore.

Fixes #7526

### DIFF
--- a/interface/procedure_tools/labcorp/ereq_form.php
+++ b/interface/procedure_tools/labcorp/ereq_form.php
@@ -70,8 +70,13 @@ STYLES;
         $pdfContent .= '<td style="width:55%;">';
         $pdfContent .= '<table>';
 
-        $collection_date = date("m/d/Y", strtotime($procedure['date_collected']));
-        $collection_time = date("H:i", strtotime($procedure['date_collected']));
+        if (!empty($procedure['date_collected'])) {
+            $collection_date = date("m/d/Y", strtotime($procedure['date_collected']));
+            $collection_time = date("H:i", strtotime($procedure['date_collected']));
+        } else {
+            $collection_date = '';
+            $collection_time = '';
+        }
         $pdfContent .= '<tr>';
         $pdfContent .= '<td style="text-align:right;width:35%;">Collection Date:</td>';
         $pdfContent .= '<td style="width:65%;">' . text($collection_date) . '</td>';
@@ -161,6 +166,7 @@ STYLES;
         $procedure_left = $proc_order - $procedure_right;
         $all_procedures = array();
         $all_diagnoses = array();
+
         if (!empty($procedure['order_diagnosis'])) {
             $all_diagnoses[] = $procedure['order_diagnosis'];
         }
@@ -181,7 +187,9 @@ STYLES;
             $pdfContent .= '<tr>';
             $pdfContent .= '<td class="width50">' . text($all_procedures[$i]['procedure_code']) . '</td>';
             $temp_diag = explode(";", $all_procedures[$i]['diagnoses']);
-            $all_diagnoses[] = $temp_diag;
+            foreach ($temp_diag as $sub_diag) {
+                $all_diagnoses[] = $sub_diag;
+            }
             $pdfContent .= '</tr>';
         }
         $pdfContent .= '</table>';
@@ -201,7 +209,9 @@ STYLES;
             $pdfContent .= '<tr>';
             $pdfContent .= '<td class="width50">' . text($all_procedures[$i]['procedure_code']) . '</td>';
             $temp_diag = explode(";", $all_procedures[$i]['diagnoses']);
-            $all_diagnoses[] = $temp_diag;
+            foreach ($temp_diag as $sub_diag) {
+                $all_diagnoses[] = $sub_diag;
+            }
             $pdfContent .= '</tr>';
         }
         $pdfContent .= '</table>';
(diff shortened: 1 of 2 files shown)


interface/procedure_tools/labcorp/ereq_form.php
```diff
@@ -70,8 +70,13 @@ function ereqForm($pid, $encounter, $form_id, $reqStr = null, $doDoc = true)
...
 
-        $collection_date = date("m/d/Y", strtotime($procedure['date_collected']));
-        $collection_time = date("H:i", strtotime($procedure['date_collected']));
+        if (!empty($procedure['date_collected'])) {
+            $collection_date = date("m/d/Y", strtotime($procedure['date_collected']));
+            $collection_time = date("H:i", strtotime($procedure['date_collected']));
+        } else {
+            $collection_date = '';
+            $collection_time = '';
+        }
```
sjpadgett, line 79: Good call!

interface/procedure_tools/labcorp/ereq_form.php
```diff
@@ -365,25 +375,27 @@ function ereqForm($pid, $encounter, $form_id, $reqStr = null, $doDoc = true)
         $pdfContent .= '<td style="padding-left:8px;" colspan="8"><b>Diagnosis Codes:</b>List all applicable Diagnosis codes. Must be at Highest Level Specificity.</td>';
         $pdfContent .= '</tr>';
         $pdfContent .= '<tr>';
-        $pdfContent .= '<td style="width:12.5%;" >' . text(str_replace('ICD10:', '', $all_diagnoses[0])) . '</td>';
-        $pdfContent .= '<td style="width:12.5%;" >' . text(str_replace('ICD10:', '', $all_diagnoses[1])) . '</td>';
```
sjpadgett, line 369: Yes this was expecting ICD10 which are required in good old USA. I think code type was removed for space on the requisition form.
BTW: I didn't design this tedious form but inherited it from I think was Nilesh for Jit whom donated to openemr.
I just made it work.

interface/procedure_tools/labcorp/gen_hl7_order.inc.php
```diff
@@ -355,7 +354,11 @@ function gen_hl7_order($orderid, &$out, &$reqStr)
     $P[70] = $vitals['height'];
     $P[88] = $vitals['bps'] . '^' . $vitals['bpd'];
     $P[89] = $vitals['waist_circ'];
-    $C[17] = hl7Date(date("Ymd", strtotime($porow['date_collected'])));
+    if (!empty($porow['date_collected'])) {
+        $C[17] = hl7Date(date("Ymd", strtotime($porow['date_collected'])));
+    } else {
+        $C[17] = null;
```
sjpadgett, line 359: Not sure hl7 allows null if this is making it to the segment. Would probably blow up parser. Recommend empty as a place holder but perhaps a quick check of the spec is needed to verify.
Wait, I see! data() may generate the good old 1969 date if not null but I forget if it does or doesn't on an empty string so be sure to check the downloaded hl7 file if provider is set to download or or the file in directory which you fix later on.
Sorry for rambling. I may be over explaining like I'm doing now for this sorry!:)

interface/procedure_tools/labcorp/gen_hl7_order.inc.php
```diff
@@ -817,6 +820,10 @@ function send_hl7_order($ppid, $out)
         // Compute the target path/file name.
         $filename = $msgid . '.hl7';
         if ($pprow['orders_path']) {
+            if (!file_exists($pprow['orders_path'])) {
+                // attempt to make the directory
+                mkdir($pprow['orders_path']);
```
sjpadgett, line 824: If order_path contains file name this will create wrong path. But I think you've tested already.

sjpadgett (approved):
Thanks Stephen. Thoughtful changes. We need to add some of the not vendor specific changes to the other generators we have which I will take on when I copy ereq to be available for all lab providers. I will make the ereq generic removing LabCore refs.

interface/procedure_tools/labcorp/gen_hl7_order.inc.php
adunsulag: @sjpadgett I can set this to be an empty string.  Previously strtotime would return 0 which would set the date_collected to be the current date which seems like a big error. Will an empty string work or does it need to be an HL7 null value code?

interface/procedure_tools/labcorp/gen_hl7_order.inc.php
adunsulag: orders_path is just the directory, the filename gets added after.

interface/procedure_tools/labcorp/gen_hl7_order.inc.php
sjpadgett: Yep empty string. Null flavor is a v3 concept where in v2.x empty string is common. However for LabCorp's as with other receivers they may require a special code which I don't think applies here and I no longer have labcorps spec.
Bottom line, set to empty string.:)

adunsulag: Changed it to be empty string, once CI finishes, I'll bring it in if there aren't any problems.

PR #7530, opened 2024-06-27
Adding double quotes to route

### LINKED ISSUE #7529: FHIR Route issue
### Describe the bug
Found this in error log

      OpenEMR.DEBUG: Found operation that is not supported in system {"resource":"insurance","operation":"$swap-insurance","items":["GET ","api","patient",":puuid","insurance","$swap-insurance"]} []

### To Reproduce

Attempt to register an app on the register-app.php page.

1. fill in all the information
2. Click on submit
3. Check log

### Expected behavior

Register an app

### Client configuration

- Browser: Firefox
- OpenEMR version: 7.0.2 (1) with addendum 

juggernautsei:
Fixes #7529

Error message that is being shown in the error logs

Adding double quotes - simple stuff

### DIFF
--- a/_rest_routes.inc.php
+++ b/_rest_routes.inc.php
@@ -6323,7 +6323,7 @@ RestConfig::$ROUTE_MAP = array(
      *      security={{"openemr_auth":{}}}
      *  )
      */
-    'GET /api/patient/:puuid/insurance/$swap-insurance' => function ($puuid, HttpRestRequest $request) {
+    "GET /api/patient/:puuid/insurance/$swap-insurance" => function ($puuid, HttpRestRequest $request) {
         if ($request->isPatientRequest()) {
             $puuid = $request->getPatientUUIDString();
         }


adunsulag:
This makes no sense @juggernautsei, double quotes leads to a variable evaluation on the $swap variable which wouldn't exist... so I don't know how this solves anything, if anything it should break the functionality.

adunsulag:
@juggernautsei Look at the bulkdata-status and $export operations, you'll see they are all single quotes.  The issue you're trying to resolve has to be somewhere else, as switching this to double quotes does not solve the problem.  If you are dealing with an access issue you may want to look at the ScopeRepository.

juggernautsei:
I am learning the $swap-insurance is a cascade from somewhere else. I am trying to track it down. Removing the single quotes did remove the error message from the log. I will put them back and close this.

adunsulag:
@juggernautsei I was trying to track it down until my van broke down lol. I'm pretty sure its because we didn't add the operation to the ScopeRepository, but I'll need to dig into it further and see what is going on because the swagger isn't even requesting the scope so it shouldn't have mattered.

sjpadgett:
@juggernautsei 
I recommend improving your debugging skills and can learn along the way by trying to help out and track this down in code if you believe it to be a bug.
Even if you don't want to do a PR you'd at least save a whole lot by pointing the actual bug out.

juggernautsei:
"I recommend improving your debugging skills"
@sjpadgett there are not a lot of debugging tools to use. Not knowing is my biggest hurdle. Cachegrind don't help me a lot because I have no way of know what I am reading. It is just a pretty picture that tells me nothing. Static code analysis is not a big help either for debugging. I turn on the debugging in the system. So, now I am back to ignorance.

PR #7532, opened 2024-06-28
CCDA Service fixes

### LINKED ISSUE #7528: CCDA Service compile is broken node version > v18.x
### Describe the bug
Node server app serveccda.js fails to start
### To Reproduce
npm install with version greater than 18.x or current recommended 20.10.0 won't work because evidentially I didn't update the lock file for new versions.
I'm testing now if the compiled version on 20.10.0 is backwards compatible with lower versions or minor releases in v20.x.0 of node.

Works on v18.20.1 So use this node version until I get the solution into the patch pipeline. @juggernautsei please note. Or delete package lock file and compile will work.

RE: I'm sorry if reading this before this edit but in order to use serccda.js before a new patch you must:
- delete the following directories: `ccdaservice/oe-blue-button-generate/node_modules`, `ccdaservice/node_modules`
- then delete `package-lock.json`
- cd `oe-blue-button-generate and npm install` // not sure this is needed any longer and will investigate.
- cd ..
- then `npm install`

That should do it. again @juggernautsei work around for your current use case.

I will use node v20.11.1 for my next package release of CCDA because this is what docker currently uses but say v20.15.0 will also work on 20.11.1 compiled version etc.

### Steps to reproduce the behavior:

compile with node version 20.x

### Client configuration

- Browser:
- OpenEMR version:
- Operating system: all three we support

sjpadgett:
- refactor formatDate() to include an ISO 8106 reformat causing a moment() warning.
- turn debug off
- bump moment version
- remove unneeded node module directory

Fixes #7528

### DIFF
--- a/ccdaservice/oe-blue-button-generate/index.js
+++ b/ccdaservice/oe-blue-button-generate/index.js
@@ -5,16 +5,16 @@ This script converts CCDA data in JSON format (originally generated from a Conti
 standard XML/CCDA format) back to XML/CCDA format.
 */
 
-var bbu = require("../oe-blue-button-util");
+const bbu = require("../oe-blue-button-util");
 
-var engine = require('./lib/engine');
-var documentLevel = require('./lib/documentLevel');
+const engine = require('./lib/engine');
+const documentLevel = require('./lib/documentLevel');
 
-var bbuo = bbu.object;
+const bbuo = bbu.object;
 
-var html_renderer = require('./lib/htmlHeaders');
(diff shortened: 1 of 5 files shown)


sjpadgett:
I'm getting really tired of being the nice guy to a community I see as ungrateful!
Merging Can go into patch2 if anyone wants it...

sjpadgett:
keep forgetting to check merge type

bradymiller:
2DV

PR #7545, opened 2024-07-02
update ci and dev environments to use mariadb 11.4

bradymiller:
update ci and dev environments to use mariadb 11.4


stephenwaite:
creepin

PR #7554, opened 2024-07-09
Questionnaire - configure where LOINC notice should go

### LINKED ISSUE #7550: Questionnaire - configure where LOINC terms note is displayed
The text
`Important to Note: LOINC form definitions are subject to the LOINC [terms of use.](http://loinc.org/terms-of-use)`
is displayed at the top of a questionnaire, when it is in edit mode.

It would be better to have this displayed at the foot or head of the form, based on a configuration item.

ruthkonyn:
Fixes #7550

adds a configuration/appearance item to specify where the LOINC copywrite statement should be displayed when editing a questionnaire - at the top, at the bottom, neither, both 

option 'Display LOINC note on questionnaires' - in configuration/appearance. Possible values top, bottom, both, don't display - added to globals.inc.php
Modify questionnaire_assessments.php to take account of the configured option when displaying the note

### DIFF
--- a/interface/forms/questionnaire_assessments/questionnaire_assessments.php
+++ b/interface/forms/questionnaire_assessments/questionnaire_assessments.php
@@ -114,13 +114,13 @@ $top_note = $bottom_note = false; // value of '3' says don't display
 
 switch ($GLOBALS['questionnaire_display_LOINCnote']) {
     case '0':
-            $top_note = true;
-            break;
+        $top_note = true;
+        break;
     case '1':
-            $bottom_note = true;
-            break;
+        $bottom_note = true;
+        break;
     case '2':
-            $top_note = $bottom_note = true;
+        $top_note = $bottom_note = true;
 }
 
 ?>
--- a/interface/forms/questionnaire_assessments/questionnaire_assessments.php
+++ b/interface/forms/questionnaire_assessments/questionnaire_assessments.php
@@ -109,8 +109,10 @@ try {
 } catch (Exception $e) {
     die(xlt("Can not continue with reason.") . '<br />' . text($e->getMessage()));
 }
-/* where to put the LOINC statement */
+/* where to put the LOINC statement , and the statement itself */
 $top_note = $bottom_note = false; // value of '3' says don't display
+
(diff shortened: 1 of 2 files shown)


ruthkonyn:
i don't understand the final php style error - it looks to me as though my style is correct?

interface/forms/questionnaire_assessments/questionnaire_assessments.php
```diff
@@ -112,11 +112,16 @@
...
+switch ($GLOBALS['questionnaire_display_LOINCnote']) {
+    case '0':
+            $top_note = true;
+             break;
+    case '1':
+            $bottom_note = true;
+             break;
+    case '2':
+            $top_note = $bottom_note = true;
+}
```
sjpadgett: indent 4 spaces for each level so case '?': level is correct however last level(setting var) is 8 spaces instead of 4 i.e. $top_notes '$' should be under the space after case.
```
case '0':
    $top_note` = true;
    break;
```

sjpadgett (commented):
I'll have to look at this later tonight or morrow!

interface/forms/questionnaire_assessments/questionnaire_assessments.php
ruthkonyn: ah thanks jerry, i'll amend and resubmit - it's not urgent btw

ruthkonyn: i'm just improving the code - will do another PR shortly...

adunsulag:
@ruthkonyn just pulled this down and tested it.  If there is no globals value set we should set the LOINC term to display either at the top or bottom of the page.  Everything else is working properly.  Just when I initially pulled the PR down and looked at it, the LOINC term doesn't show.  Default should be to show it for existing installations.  Once that's fixed we can bring this in.

ruthkonyn:
Thanks @adunsulag I think this corrects the problem - i've made the default to show at the top of the page, as this is how it was originally. 
Test / PHP 8.1 - Apache - MariaDB 11.4 (pull_request)  - failing again?

ruthkonyn:
hi @adunsulag - is there anything else i should be doing on this? - so it can be brought into the core code?

adunsulag:
@ruthkonyn Not that I can see, was waiting for that 8.1 test to pass as I've run it a couple of times.  I'll pull it down locally and run the tests, if it passes on my machine I'm assuming it will be part of that CI engine bug Brady fixed this week.

adunsulag:
@ruthkonyn Tests look fine.  Bringing this in.

PR #7572, opened 2024-07-17
Fixes #7570 list service search by codes

### LINKED ISSUE #7570: feat: Add to ListService method option to search options by their code column
We should add a method to the ListService class to be able to search list_options table by the code column.

adunsulag:
Fixes #7570 search for an option by the codes column

### DIFF
--- a/src/Services/ListService.php
+++ b/src/Services/ListService.php
@@ -114,6 +114,21 @@ class ListService
         return null;
     }
 
+    /**
+     * Retrieves a list option for a given list by the code value
+     * @param string $list_id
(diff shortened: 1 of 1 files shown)


bradymiller (approved):

src/Services/ListService.php
```diff
@@ -114,6 +114,21 @@ public function getListOption($list_id, $option_id)
         return null;
     }
 
+    /**
+     * Retrieves a list option for a given list by the code value
+     * @param string $list_id
+     * @param string $code The exact code(s) to match.  This must match the string precisely and fuzzy matching is not currently supported.
```
bradymiller, line 120: note that can place multiple codes in the codes column (in the lists gui), so may need "fuzzy" at some point

PR #7645, opened 2024-08-08
Fixes #7644 message casing

### LINKED ISSUE #7644: feat: change clinical messages casing in messages.php
The "To" field contain in the messages screen has odd spelling, as though this is an SQL statement: "SELECT Users FROM The Dropdown LIST."

If you select a user, then select Clear, the "To" field instead becomes "Select Users From The Dropdown List", which is an improvement but doesn't need uppercase words.

image

adunsulag:
I fixed the message casing on the messages so it doesn't look like a SQL statement.  Also saw we were using xls which is deprecated so I cleaned it up everywhere it was being used in the messages.php file.

Fixes #7644


interface/main/messages/messages.php
```diff
@@ -716,11 +716,11 @@ function confirmDeleteSelected() {
                                         }
                                     }
                                     if (int_checked > 0){
-                                        if (confirm("<?php echo xls('Do you really want to delete the selection?'); ?>")) {
+                                        if (confirm(<?php echo xlj('Do you really want to delete the selection?'); ?>)) {
```
bradymiller, line 719: 2DV

bradymiller (approved):

PR #7649, opened 2024-08-08
Fixes #7648 secure messaging

### LINKED ISSUE #7648: feat: Improve patient and clinician message user interface
On the secure messaging for portal & clinician we should make the compose message button more obvious (give it a color and relabel it). We should also adjust the menu drop down label items to be more clear for the archive and the compose message items.

In addition on the secure messaging we should make some of the text look like a link button.  Even though the row is clickable, there should be a visable indicator that the user should click something.  Testing with patients in the field who are not technically savy they don't realize they can click the entire row and having a link text makes this more obvious.

On the provider messaging window we should make the entire row clickable like we have on the portal secure messaging.

adunsulag:
Made the main clinician messages screen clickable for the entire row. Made it so the entire row shows a cursor icon that its clickable when you hover over it

Portal secure messaging - made the compose message a primary button. Moved the refresh button to the very end of the portal message bar and changed its title to be more descriptive.

Added text to the compose message button to make it more clear what should be done here for better discoverability.

Made the sender/receiver and subject name clickable.

Fixes #7648


bradymiller (approved):

stephenwaite:
smaller_excited_elephant

PR #7653, opened 2024-08-08
Fixes #7652 portal twigify appointments and no appointments font update

### LINKED ISSUE #7652: feat: Twigify patient portal appointments card list and show clearer message when no appointments are found.
If a module writer wants to override just the appointment card list they have to override the entire portal home.html.twig file.  We should break appointments out into its own twig file.

Also we should display a nicely formatted message when no appointments can be found that is mobile friendly.  Currently on a mobile device the appointments header and no appointment message are the same size which some users have reported as jarring to them.

adunsulag:
Changed up the appointments text when there are no appointments.

Also twigify the appointments card list.
Fixes #7652

### DIFF
--- a/templates/portal/home.html.twig
+++ b/templates/portal/home.html.twig
@@ -314,53 +314,6 @@
             }
         }
 
-        function editAppointment(mode, deid) {
-            let mdata = {};
-            let title = '';
-
-            if (mode === 'add') {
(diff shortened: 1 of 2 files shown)


bradymiller (approved):

bradymiller:
looks like a merge conflict

PR #7671, opened 2024-08-13
display correct name when 'choices' group name is changed

### LINKED ISSUE #7670: patient portal card on dashboard mentions 'demographics choices' even if the group has been renamed
### Describe the bug
patient portal card on dashboard mentions 'demographics choices' even if the group has been renamed to something other than choices

### To Reproduce
use admin -> forms -> layouts -> demographics to rename the 'choices' group to something else, e.g. 'Options' 
got to a patient's dashboard
make sure that ' allow patient portal' is not set to YES
look at Patient Portal card, it will appear like this, instead of 'Demographics Options'

Screenshot from 2024-08-13 11-35-05
 
### Expected behaviour

the card should say 'Allow Patient Portal in Demographics Options

### Client configuration

- Browser: firefox
- OpenEMR version: 7.0.3-dev
- Operating system: ubuntu

ruthkonyn:
Fixes #7670

patient portal card on dashboard mentions 'demographics choices' even if the group has been renamed to something other than choices 

PostalCard.php now looks for the group's name (group 4) in the db and uses the new name - or reverts to 'Choices' if the group name is not returned

### DIFF
--- a/src/Patient/Cards/PortalCard.php
+++ b/src/Patient/Cards/PortalCard.php
@@ -64,16 +64,16 @@ class PortalCard extends CardModel
         $this->opts['templateVariables']['appendedInjection'] = $dispatchResult->getAppendedInjection();
     }
 
-    // used in portal.html.twig and card_base.html.twig
+    // used in twigs
     private function setOpts()
     {
         global $GLOBALS;
         global $pid;
         // RM get name for 'choices' group, i.e. group 4 in 'layout_gropu_properties' table in db
         $sql = "SELECT grp_title FROM layout_group_properties WHERE grp_group_id = 4 AND grp_form_id = 'DEM'";
-        $res = sqlStatement ($sql);
+        $res = sqlStatement($sql);
         $nrow = sqlFetchArray($res);
-        $groupName = ($nrow['grp_title']  !=='' ? $nrow['grp_title'] : 'Choices');
+        $groupName = ($nrow['grp_title']  !== '' ? $nrow['grp_title'] : 'Choices');
 
         $this->opts = [
             'acl' => ['patients', 'demo'],
--- a/src/Patient/Cards/PortalCard.php
+++ b/src/Patient/Cards/PortalCard.php
@@ -85,7 +85,7 @@ class PortalCard extends CardModel
             'identifier' => self::CARD_ID,
             'title' => xl('Patient Portal') . ' / ' . xl('API Access'),
             'templateVariables' => [
-                'allowpp' => (xl('Allow Patient Portal in Demographics') . ' ' . xl_layout_label($groupName)),
+               'allowpp' => $groupName,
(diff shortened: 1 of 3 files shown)


src/Patient/Cards/PortalCard.php
```diff
@@ -78,6 +85,7 @@ private function setOpts()
             'identifier' => self::CARD_ID,
             'title' => xl('Patient Portal') . ' / ' . xl('API Access'),
             'templateVariables' => [
+                'allowpp' => xl('Allow Patient Portal in Demographics ' . $groupName),
```
bradymiller, line 88: can't include leading/trailing spacing in translations and need to translate the group title with a special translation function:
```
'allowpp' => xl('Allow Patient Portal in Demographics' . ' ' . xl_layout_label($groupName)),
```

bradymiller: actually, would just pass the $groupName into the twig template:
```
'allowpp' => $groupName,
```

templates/patient/partials/portal.html.twig
```diff
@@ -33,7 +33,7 @@ The Patient Portal card for the Medical Record Dashboard
     {% if isPortalAllowed == false %}
         <div class="alert alert-warning" role="alert">
             <p class="font-weight-bold">{{  "Portal Access"|xlt }}</p>
-            <p> {{ "Allow Patient Portal in Demographics Choices."|xlt }} </p>
+            <p> {{ allowpp }} </p>
```
bradymiller, line 36:
```
<p> {{ "Allow Patient Portal in Demographic"|xlt }} {{ allowpp|xlLayoutLabel|text }} </p>
```

Note need to add a new xlLayoutLabel function in src/Common/Twig/TwigExtension.php (would put it above the xlListLabel function):
```
            new TwigFilter(
                'xlLayoutLabel',
                function ($string) {
                    return xl_layout_label($string);
                }
            ),
```

src/Patient/Cards/PortalCard.php
ruthkonyn: thanks @bradymiller - i'm doing your first suggestion as it needs the complete string  'Allow Patient Portal in Demographics <groupName>' - rather than simply '<groupName>'

templates/patient/partials/portal.html.twig
ruthkonyn: many thanks @bradymiller - have implemented it the way you suggest in the twig template.

ruthkonyn:
in the past i've ignored the failing Nginx - MariaDB test?

bradymiller (approved):

bradymiller:
hi @ruthkonyn , Code looks great and bringing in. Thanks for the fix!

bradymiller:
btw, ok to fail up to a couple of the ci Test/PHP* runs since these runs just break every once in a while

PR #7679, opened 2024-08-25
fix: update couple packages in ci

bradymiller:
update couple packages in ci


bradymiller:
Very odd where e2e is now failing on php 8.3, but not php 8.1, 8.2, or 8.4-dev (noted this on the last commit that just came into codebase even though not noted on the same code in PR; not sure what has changes but means likely not related to the code and more the environment; also confirmed now happening in the easy dev environment, so not sure what is happening; maybe a bad chromium-chromedriver package???)

bradymiller:
my guess is that the e2e/php8.3 issue will sort itself over time, so will skip the Check Add User test for now :)

PR #7702, opened 2024-09-09
Fixes #7701 register app error

### LINKED ISSUE #7701: bug: js_xl not found in register-app.php scope
I'm scratching my head on what changed, but the register-app.php file throws an error now saying js_xl is not found in the scope.  Not sure why that would be but I changed it up to use window.top.xl instead of js_xl.

adunsulag:
Fixes the translation error when an app was registered.

Fixes #7701


interface/smart/register-app.php
```diff
@@ -160,7 +160,7 @@ function registerApp() {
                     document.querySelector(".errorResponse").classList.add("hidden");
                     document.querySelector("#clientID").value = resultJSON.client_id;
                     document.querySelector("#clientSecretID").value = resultJSON.client_secret;
-                    msgText = js_xl('Your app has been registered!');
+                    msgText = window.top.xl('Your app has been registered!');
```
bradymiller, line 163: looks good, but looks like some other custom stuff got in here.

adunsulag: lol, great catch, I'll back out the file.

PR #7711, opened 2024-09-11
add ability to configure two logos on portal login page

### LINKED ISSUE #7710: configure two logos on the portal login page - as one can on user login page
it would be good to be able to add an institutions logo to the portal login page, as one can for the user login page and have this configurable. also be able to configure which order they are displayed in

ruthkonyn:
Fixes #7710

enable one to have two logos on the portal login page, configure which to display and in which order

add 3 configuration items to  admin -> config -> portal 
Screenshot from 2024-09-11 16-42-37
and then display zero, one or two logos and if two in either order

add the directory openemr/public/images/logos/portal/login/secondary - in which to place the secondary image

### DIFF
--- a/portal/index.php
+++ b/portal/index.php
@@ -51,6 +51,7 @@ use OpenEMR\Services\LogoService;
 $landingpage = "index.php?site=" . urlencode($_SESSION['site_id']);
 $logoService = new LogoService();
 $logoSrc = $logoService->getLogo("portal/login/primary");
+$logo2ndSrc = $logoService->getLogo("portal/login/secondary","logoHSE.png"); /*rm - add secondary logo */
 
 // allow both get and post redirect params here... everything will be sanitized in get_patient_info.php before we
 // actually do anything with the redirect
@@ -550,7 +551,24 @@ if (!(isset($_SESSION['password_update']) || (!empty($GLOBALS['portal_two_pass_r
         <?php } else {
             ?> <!-- Main logon -->
         <div class="container-xl p-1">
-            <div class="img-fluid text-center"><img class="login-logo" src='<?php echo $logoSrc; ?>'></div>
+        <!-- Optionally show two logos, and in either order -->
+        <?php if($GLOBALS['secondary_portal_logo_position'] == 'second'){ ?>
+              <?php if ($GLOBALS['show_portal_primary_logo']) { ?>
+                <div class="img-fluid text-center"><img class="login-logo" src='<?php echo $logoSrc; ?>'></div>
+              <?php } ?>
+              <?php if ($GLOBALS['extra_portal_logo_login']) { ?>
+                <div class="img-fluid text-center"><img class="login-logo" src='<?php echo $logo2ndSrc; ?>'></div>
+              <?php } ?>
+        <?php } else {
+               if($GLOBALS['secondary_portal_logo_position'] == 'first'){ ?>
+                  <?php if ($GLOBALS['extra_portal_logo_login']) { ?>
+                        <div class="img-fluid text-center"><img class="login-logo" src='<?php echo $logo2ndSrc; ?>'></div>
+                  <?php } ?>
+                   <?php if ($GLOBALS['show_portal_primary_logo']) { ?>
+                       <div class="img-fluid text-center"><img class="login-logo" src='<?php echo $logoSrc; ?>'></div>
+                   <?php } ?>
+               <?php  } ?>
+          <?php } ?>
             <legend class="text-center bg-light text-dark pt-2 py-1"><h2><?php echo $GLOBALS['openemr_name'] . ' ' . xlt('Portal Login'); ?></h2></legend>
             <form class="mx-1" action="get_patient_info.php" method="POST" onsubmit="return process()">
                 <?php if (isset($redirectUrl)) { ?>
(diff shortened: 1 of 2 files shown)


portal/index.php
```diff
@@ -51,7 +51,7 @@
 $landingpage = "index.php?site=" . urlencode($_SESSION['site_id']);
 $logoService = new LogoService();
 $logoSrc = $logoService->getLogo("portal/login/primary");
-$logo2ndSrc = $logoService->getLogo("portal/login/secondary","logoHSE.png"); /*rm - add secondary logo */
+$logo2ndSrc = $logoService->getLogo("portal/login/secondary", "logoHSE.png"); /*rm - add secondary logo */
```
sjpadgett, line 54: Where did "logoHSE.png" come from? This needs to be configurable.

portal/index.php
```diff
@@ -552,23 +552,23 @@ function enableVerifyBtn() {
...
-        <?php if($GLOBALS['secondary_portal_logo_position'] == 'second'){ ?>
-              <?php if ($GLOBALS['show_portal_primary_logo']) { ?>
-                <div class="img-fluid text-center"><img class="login-logo" src='<?php echo $logoSrc; ?>'></div>
-              <?php } ?>
-              <?php if ($GLOBALS['extra_portal_logo_login']) { ?>
-                <div class="img-fluid text-center"><img class="login-logo" src='<?php echo $logo2ndSrc; ?>'></div>
-              <?php } ?>
-        <?php } else {
-               if($GLOBALS['secondary_portal_logo_position'] == 'first'){ ?>
+            <?php if ($GLOBALS['secondary_portal_logo_position'] == 'second') { ?>
```
sjpadgett, line 555: You need to null coalesce most, especially new, variables to prevent warnings and eventually errors.
` if($GLOBALS['secondary_portal_logo_position'] ?? null == 'second'){`
always try to follow type such as if int `$x ?? 0` etc.

sjpadgett (changes requested):
I guess I don't have a problem with this but these branding changes tend to cause issues in the future with styling updates.

portal/index.php
ruthkonyn: oh thanks for pointing out the HSE in the variable name - i shouldn't need it - i'll take it out...

portal/index.php
ruthkonyn: thanks will make the change, that is good advice!

portal/index.php
adunsulag: Hey @ruthkonyn did you forget to push this change?  Saw you marked it as resolved, but looks like the issue is still there.

adunsulag:
@ruthkonyn once you've made the changes Jerry suggested and fixed the styling error, I think this is fine to bring in.  Let us know if you need help fixing the styling issue as that can be a challenge sometimes.

portal/index.php
ruthkonyn: hi @adunsulag - no, it was around 8pm when i saw jerry's message and i'd just come out of a meeting. i felt too tired. if i marked it as resolved that was by accident, apologies (still exploring the way github works :-) ) . I'm planning to resolve both issues this morning.

ruthkonyn:
hi @adunsulag - i've fixed jerry's requested changes and the styling errors - so i think/hope it's good to go now.

pete has asked "Please can you ..[get] it to [stephen] for the Demo EMR and our module and patch, ..he keeps track of changes that he makes for us, and will deliver that as a patch or a module that we deploy on e.g. 7.0.2 Patch 1 systems that are in production"

or if you prefer i can give him the two files and ask him to put them on the relevant servers (this is how we've worked in the past with changes i've made to the code)

adunsulag:
We'll make a patch from the 7.0.2 release, you bringing this in as a PR into OpenEMR master is the right approach as I'll make sure it gets merged into the 7.0.2 release that the patch will be made from.

portal/index.php
```diff
@@ -51,7 +51,7 @@
 $landingpage = "index.php?site=" . urlencode($_SESSION['site_id']);
 $logoService = new LogoService();
 $logoSrc = $logoService->getLogo("portal/login/primary");
-$logo2ndSrc = $logoService->getLogo("portal/login/secondary", "logoHSE.png"); /*rm - add secondary logo */
+$logo2ndSrc = $logoService->getLogo("portal/login/secondary"); /*rm - add secondary logo */
```
adunsulag, line 54: So if the secondary logo is turned on it will show a broken image link (src="").  I think that behavior is ok, people can just go and uncheck the config option if the logo is missing.

adunsulag:
@sjpadgett I believe this looks fine at this point, you ok if I bring this in?

sjpadgett:
OK with me

portal/index.php
ruthkonyn: thanks @adunsulag for taking it into the patch

and it works ok if secondary is turned on but either the .png file or indeed the directory 'secondary' are missing - no error reports and there is simply no secondary logo displayed on the login screen.

ruthkonyn: if i understand what you are saying: i tested with secondary logo turned on then deleted the logo file, then the secondary folder. both times this caused no problem or error message.

PR #7723, opened 2024-09-15
Update API.php

ophthal:
This <ul> only displays MedEx Recall messaging progress.  It does need to be displayed otherwise.

Fixes #


bradymiller (approved):

sjpadgett:
@ophthal Is this needed for patch 2? we're getting the patch ready now and I'll do one more backport PR before creating patch zip

ophthal:
It is a cosmetic fix only. That row documents recall message progress via color for MedEx.  Makes no sense without MedEx.On Sep 27, 2024, at 10:33 AM, Jerry Padgett ***@***.***> wrote:﻿
@ophthal Is this needed for patch 2? we're getting the patch ready now and I'll do one more backport PR before creating patch zip

sjpadgett:
Okay thanks. I went ahead and added to patch so it's there for those that need it.

PR #7757, opened 2024-10-06
feat: incorporate traits into e2e ci testing

bradymiller:
incorporate traits into e2e ci testing


bradymiller:
testing well and bringing in

PR #7767, opened 2024-10-14
Feat openemr fix 7764 last updated

### LINKED ISSUE #7764: feat: Update EHI documentation to include FHIR _lastUpdated columns
We added a number of created and last_updated fields in order to deal with the FHIR _lastUpdated columns.  We need to update all of the EHI documentation to support this.

adunsulag:
Fixes #7764 

Update EHI documentation and fix generator file to work with latest docker images for development easy environment.


adunsulag:
@bradymiller I don't like how big this makes the git history with all of these images but not sure if we do this any other way.

bradymiller:
@adunsulag . Agreed. In the future when we have more time, can work out a way to put this into a separate repo and place on website (since just html).

PR #7779, opened 2024-10-23
fix: e2e fix user add

bradymiller:
Still working on the user add test, which still fails every once in awhile (sometimes it simply does not add the user to database and sometimes is simply doesn't see it on the user listing)


bradymiller:
Looks like have stamped out most of the random fails related to creating new user and new patient. But time will tell :)

stephenwaite:
squirrel_dance

PR #7743, opened 2024-09-30
Feat openemr b11 7742 decision support interventions

adunsulag:
Lot of things happening in this PR so I'll explain what I'm doing here and will add screenshots tomorrow.  Going to put this as a draft PR for now until I finish one last piece with the existing CDR rules.

I moved all of the current CDR classes into their own namespace so we could re-use them in the src/ folder.  Also made them PSR-4 compliant so they can be loaded without requires and reused.  Which is why there's the large file changes and so I could make some test classes for what I was doing.

I added more source attributes into the existing clinical_rules table.

I created two new lists to hold the evidence based and predictive dsi source attribute configurations for the smart apps.  This can eventually be leveraged by the CDS Hooks specification if we decide to go that route in the future.  This enables installations to configure the source attributes per spec requirements.

I added a section for the DSI services in the Client API management screen. 

I changed a number of the Client API management screens over to twig.  Plan on converting the token tools over to twig and that will make the ClientAdminController consistent / cohesive.

I updated the smart app launch card in the demographics screen to show the DSI information.  People can click on it and pop up a dialog that shows the source attribute information.

I fixed up the smart app registration screen to have more helpful tips for people running into problems with their app registration.  Added a radio select option to choose if the app has DSI information and give app writers the option to fill that information out.

Changed up the AuthorizationController to handle the new registration pieces and return the DSI information on client api introspection (once the pathing issue is fixed in a separate PR).

Issue #7742, opened 2024-09-30
feat: Implement US ONC (b)(11) Decision Support Interventions functionality


adunsulag:
We need to implement the requirements of ONC (b)(11) Decision Support Interventions as outlined by ONC here:
https://www.healthit.gov/test-method/decision-support-interventions

As a high level checklist the following are required:
- [x] Decision Support Configuration - allow limited sets of users to configure source attributes
- [x] Enable interventions when a patient's medications, allergies and intolerance, and problems are incorporated from a transition of care or referral summary received and pursuant to paragraph
- - Note this is unchanged from the (a)(9) regulation so existing functionality is unchanged
- [x] Enable a user to provide electronic feedback data for evidence-based decision support interventions
- - Note user feedback is still being assessed on what this means as the specification states feedback must include the intervention, action taken, user feedback provided (if applicable), user, date, and location
- [x] Decision support intervention selection - allow limited sets of users to activate evidence based and predictive based DSIs
- - Note this will be done using the current CDR rules engine for evidence based DSI, AND enabling SMART apps to specify if they use evidence based or predictive DSI
- [x] Source attribute and modification -- allow limited sets of users to update source attribute information for evidence & predictive DSI
- - Evidence based DSI for CDR rules will be done in the existing system built for the old (a)(9) regulation
- - SMART APP source attributes will be editable from within the Client API Management utilities
- [x] Add source attributes not defined in (b)(11)(iv)(B)
- - This will be done via the Admin -> Forms -> Lists management screen by editing either of the following two lists
- - - Evidence Based Decision Support Interventions Source Attributes
- - - Predictive Decision Support Interventions Source Attributes
- [x] Intervention risk management. 
- - - We do not bundle any predictive DSI in OpenEMR currently so we do not need to deal with this requirement.
- [x] Specify source attributes are unknown if 3rd party DSI provider has  not provided the information or if historical data on DSI bundled with certified software is not available on source attributes.
- [x] Safety Enhanced Design
- - - Both the admin and clinician facing portions of this DSI that were changed from (a)(9) will need to undergo a Safety Enhanced Design.  Plan is for this to be completed during the month of October and possibly first week of November.

The actual ONC text (pasting here because ONC updates their URLs from time to time)
![image](https://github.com/user-attachments/assets/d75f1b28-0fb6-4485-b542-ca8319cbf462)
![image](https://github.com/user-attachments/assets/ed181e4f-4213-4335-b755-8aa2b790e89a)
[...]

PR #7786, opened 2024-10-27
Feat openemr b11 #7742 decision support interventions (#7743)

adunsulag:
Backports #7743 into rel-702 for preparation for patch 3.

* CDR rules refactor, PSR-4, testability

Refactored the CDR rules editing engine to be in the source tree so we can have PSR-4 autoloading of the files.

Refactored the CDR controllers to be able to use Request/Response objects for PSR-7 compatability and to make it possible to unit test without having to do guzzle or curl requests.

This is prep work to be able to edit some of the interface to support our onc b11 requirements.

* Moved template files fixed template bugs

Moved the template files to be in the templates directory to better align with OpenEMR coding standards.  Eventually it would be nice to move these to all be twig files.  Leaving them alone for now.

Removed the undecorated template file as its not used anywhere.

Removed the edit/helper/common.php file and moved its elements into the RuleTemplateExtension where the hope is to eventually have these be twig functions.

* Refactored AJAX code for rules to centralize code

Made the AJAX code for the CDR rules to be centralized using the same controller logic that the rest of the rule engine uses for consistency and access control.

* Cleaned up Frontend Controller includes

Removed the common, header, and ui files that were being included in the index.php and moved all of the logic into the src folder tree.

* DSI additional source attributes

Added the additional ONC required dsi source attributes for our existing CDR engine.  This includes the display and editing for those attributes.

Following attributes are added
- patient race usage
- patient ethnicity usage
- patient language usage
- patient sexual orientation usage
- patient sex usage
- patient dob usage
- patient health status assessments usage
- patient social determinants of health usage

* Initial prototype of b11 View rules

Put in the foundation for the b11 viewing of predictive dsi attributes.

Right now it shows as a questionnaire, but may switch this to a regular form.  I'll have to play around and see what I like better here.

* WIP b11 Source Attributes Predictive DSI

Added predictive source attributes into the list options to give admin's the ability to add custom attributes.

Added new table dsi_source_attributes for tracking the source attribute information on an oauth_clients entity.

Added a field to oauth_clients table to track if the client is a predictive or evidence based api client.  At some point we want a many to one relationship here, but for now we treat api clients as a 1:1 relationship for evidence based / predictive ai clients.

Refactored a great deal of the api client admin interface to use twig to modernize the codebase and add twig files for the external-cdr pieces.

This is a foundational work to support in the future the CDS Hooks standard.

Also did code style fixes on the class

* Style fixes for clinical decision rules refactor

* Fix db/dsi type mismatch, helper links on client

Fixed a mismatch between the database comments for the dsi type and what the code constants in client entity were.

Fixed the helper links on the client.

* Implement source attribute dialog display

Implemented the display of the source attributes from the smart launch screen for displaying Decession Support Interventions Source Information.

* Add CSRF token to attribute saves.

* b11 DSI api client registration

Added the registration pieces to the smart app client registration page.

Added the dsi_type and dsi_source_attributes metadata attributes to the registration endpoint.  Only existing attributes configured in the system can be configured and supported.

* Remove hyphens from dsi attribute definitions

* Add dsi information to introspection endpoint

Added the dsi information to the api endpoint that is retrieved from the introspection.  Note there is another outstanding PR that this work depends on in order for the endpoint to execute due to some pathing issues.  So this commit will not be useuable without bringing in PR
 #7741.

* Implement evidence based rules for external cdr.

* Fix no-dsi api clients not loading

* Display source attribute info on cdr rule info

Added a dialog info box that shows the cdr rule information when you click on the help icon for a cdr evidence based rule.

Added a new review controller to handle this.

Moved ACL checks into the router controller.  Opened up the review controller to regular permissions.  I don't see any security check that needs to happen on read only basis here.

* Cleaned up cdr demographics tool tip

Now that source attributes are showing up in a dialog, we don't need the tooltip to show a bunch of stuff on the rule tooltip.

Fixed width constraints on the source attribute editing as 50 characters was not correct when it should be 255 characters.

Fixed the web reference not showing up in the edit/details view of the CDR rules.

* Inserting feedback on evidence based CDR

Made it so that feedback gets inserted into the alert logs.  Users can provide feedback on the action taken in the CDR for user development.

* Implement feedback in cdr_log

Made it so the cdr log will show the feedback and allow you to export the CDR evidence based feedback in a CSV download.

Refactored the CDR_Log process to leverage the same controller infrastructure as the CDR admin controllers and dialog viewer.  Trying to centralize the code more in order to improve cohesion and maintenance.

I kept the existing view infrastructure as it'd be too major of a refactor right now to switch to twig.

Fixed some style issues as well.

* Added facility for location tracking

* Change column size, sync database.sql

Made the database.sql have the changes that were put in the sql patch.

Changed the column size to be text instead of varchar

* Fix escaping issue.

* Change sequence on list_options for more flexibility

* Fix escaping

* Fix style changes

* Fix dsi_type if its not provided

Needed to fix the dsi type if the client does not provide one in the registration.  Fixed some other errors showing up on the logs if data was not provided.

* Fix registration with predictive dsi

Prior commit was breaking the registration ouch.

* Fix broken unit tests

Routing controllers needed more DI service injectors to handle the unit testing changes.  Removed the list test and added a save test on the external cdr.

Fixed some other broken unit tests.

* Fix thrown warnings with vital forms test

* Hide cdr link if no services enabled.

* Default unknown text if source attribute missing

Made it so an unknown text is showing on details view of CDR rules and DSI smart app source attributes per ONC requirements.

* Fix style issues.

* See if we can fix the styles.

Not sure why the auto-corrector isn't catching this, but trying this again.

* Fix sql text datatype, improve source descriptions.

Made the source descriptions clearer.

Changed the SQL text datatype to remove not null attribute, fixed php code to support nullable attributes.

* Direct edit links to simplify UX/testing

Added a direct edit button if you are an admin to the CDR rules and to the smart 3rd party DSI attributes.  This simplifies the UX testing and allows users to go right from the dialog popups to editing the attributes.  I don't have deep back links supported which would be a bigger change but this is good enough to get from viewing to editing.

I'm using a one time message passing via postMessage to communicate between the dialogs and the pages.  At some point it'd be nice to generalize this, but that's an exercise for another day.

Fixes #


adunsulag:
@bradymiller I think we want to run the Usability testing machines off this PR, I can merge into rel-702 if we want but not sure if we want to do that while patch 2 is still out there.

Also, @sjpadgett I saw we deleted the sql upgrade script from 7.0.2 to 7.0.3 and put it in patch.sql.  I followed the same procedure of putting everything into patch.sql but not sure how the dockers handle patch.sql on something like openemr-cmd drid.

PR #7821, opened 2024-11-19
Implement new commands, refactory layout editor

BerwoutK:
Fixes #

This PR will make it more easy to integrate automated deployment of OpenEMR by enabling features
via commands. In this case, the setup command and the ability to add a field to a form layout.

- We need to unify the way the code works, right now it is all clumped together in 1 or 2 files.
E.g. the edit_layout file has the logic, design and more of a single action we should split it up and clean it up.

Nope

There was a lot of logic/code I didn't understand, so please help me fix the code so that it can work within OpenEMR

### DIFF
--- a/interface/super/edit_layout.php
+++ b/interface/super/edit_layout.php
@@ -189,7 +189,7 @@ function addOrDeleteColumn($layout_id, $field_id, $add = true)
 {
     $table_name = getTableNameFromLayoutId($layout_id);
     if ($add === true) {
-        addColumn($layout_id, $field_id);
+        addColumn($table_name, $field_id);
         EventAuditLogger::instance()->newEvent(
             "alter_table",
             $_SESSION['authUser'],
@@ -201,7 +201,7 @@ function addOrDeleteColumn($layout_id, $field_id, $add = true)
         return;
     }
 
-    deleteColumn($layout_id, $field_id);
+    deleteColumn($table_name, $field_id);
     EventAuditLogger::instance()->newEvent(
         "alter_table",
         $_SESSION['authUser'],
--- a/interface/super/edit_layout.php
+++ b/interface/super/edit_layout.php
@@ -26,13 +26,6 @@ use OpenEMR\Core\Header;
 // Indicates if deactivated layouts are included in the dropdown.
 $form_inactive = !empty($_REQUEST['form_inactive']);
 
-function setLayoutTimestamp($layout_id)
-{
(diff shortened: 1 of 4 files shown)


adunsulag:
@BerwoutK I'd like to review this, but probably wont' get to it until early next week.  I did a brief glance and from what I saw it looks like something we want to bring in.  Will need to pull it down and test to make sure nothing is broken in current workflow use cases.

BerwoutK:
I need to resolve the code styling (white space issues), and need a proper check on the logic as I _think_ it has what it needs but also contains some hardcoded stuff that might be very wrong. Appreciate it.

adunsulag:
@BerwoutK I pulled down your PR and attempted to run the SiteSetup piece and I'm wondering if you ever were able to run that command?  When you look at the command runner it actually includes the globals.php file which in turn tries to create a database connection.  You can see the entry point here: https://github.com/openemr/openemr/blob/075f30052b67d364a7c85c9e3f905ccd80358cd4/bin/console#L30

I'm not sure how you'd get around that as a lot of the commands need access to the globals (which are stored in the database).  Also the command runner fires off an event to modules to be able to bring in more commands.  We'd have to look at architecting some kind of 2 pass system, first pass for all system loaded modules and pieces that don't require DB access, 2nd pass for everything post database access.  Is this something you want to take a stab at?

I'll look at the layout command next.

BerwoutK:
> @BerwoutK I pulled down your PR and attempted to run the SiteSetup piece and I'm wondering if you ever were able to run that command? When you look at the command runner it actually includes the globals.php file which in turn tries to create a database connection. You can see the entry point here:
> 
> https://github.com/openemr/openemr/blob/075f30052b67d364a7c85c9e3f905ccd80358cd4/bin/console#L30
> 
> I'm not sure how you'd get around that as a lot of the commands need access to the globals (which are stored in the database). Also the command runner fires off an event to modules to be able to bring in more commands. We'd have to look at architecting some kind of 2 pass system, first pass for all system loaded modules and pieces that don't require DB access, 2nd pass for everything post database access. Is this something you want to take a stab at?
> 
> I'll look at the layout command next.

Hey, yeah we are able to just call `public/bin/console <command>` from the docker image. We simply run:
```
php public/bin/console form:layout:field:add --layout_id=DEM --newseq=260 --newid=openemr_test --newtitle="OpenEMR Test" --newuor=1 --newlengthWidth=1 --newlengthHeight=1 --newmaxSize=0 --newtitlecols=1 --newdatacols=1
```

Though I just now realized we dont use the data one, and dataType is currently wrong too.

PR #7823, opened 2024-11-19
Fix SMS and email appointment reminder notifications

### LINKED ISSUE #7822: feat: Appointment reminder notifications Clickatell Integration
### Appointment reminder SMS and email notifications don't work

Appointment reminder SMS messages (via Clickatell) and emails (through Google Workspaces account) do not work.  Clickatell no longer supports new customers using the API used for the Clickatell integration, and email reminders use a different email sending configuration that did not work with GMail/Google Workspaces.

### To Reproduce
Configure SMS notification settings and email sending.  Uncomment the exit statements in cron_sms_notificaiton.php and cron_email_notification.php and invoke the scripts.  Configure a patient with contact details, HIPPA opt ins, and an applicable upcoming appointment.  Invoke the notification scripts.

No email or SMS is received.

### Expected behavior
Email and SMS is received after scripts are invoked.

### Client configuration
N/A

ChrisOstler:
Fixes #7822

This fixes (Clickatell) SMS notifications and (Google Workspaces) email notifications.

**Only set SERVER_NAME if it was not already set** - the scripts included logic to set a SERVER_NAME and REQUEST_URI to hardcoded values.  This would support invoking the scripts directly.  However, it will overwrite legitimate values if the scripts were instead invoked through the web server.  This caused later integration with MyMailer to fail due to the hostname being rejected by the SMTP server.  Instead, only set these if they were not already set.

**Store formatted messages in local variable** - when sending messages, the message format was updated in place.  If multiple notifications were sent, this would result in subsequent messages being sent with the formatted content of the first.  Instead, store the formatted message in a local variable.

**Don't echo potential PII in notification sending output** - the scripts included echo statements for debugging.  Since the scripts are served without a login requirement through the web server, this could expose PII - names, email addresses, phone numbers.  Comment out these statements; the actions will still be logged.

**Use MyMailer for sending appointment email reminders** - other places use the `MyMailer` class for sending emails.  The notification scripts used a different mechanism, expecting different configuration values.  Change the script to use `MyMailer`, instead.

**Use passed in values for user, password, API key** - the username, password, and API key were already being passed in to the Clickatell SMS class, but not used.  Use those values.

**Update Clickatell integration to use new API** - The API used in the Clickatell integration is no longer supported for new customers.  This updates it to use the [current API](https://docs.clickatell.com/channels/sms-channels/sms-api-reference/#tag/SMS-API/operation/sendMessageHTTP).  The new API requires a sending phone number, so it adds that field to the configuration page (the `email_sender` value was already being passed through).

All of these changes are being used and working on a live instance.

### DIFF
--- a/interface/batchcom/smsnotification.php
+++ b/interface/batchcom/smsnotification.php
@@ -59,7 +59,7 @@ if (!empty($_POST['form_action']) && ($_POST['form_action'] == 'save')) {
         $sql_value = " (?, ?, ?, ?, ?, ?, ?) ";
         $values = array($_POST['notification_id'], $_POST['sms_gateway_type'],
                         $_POST['provider_name'], $_POST['message'],
-                        '', '', 'SMS');
+                        $_POST['sender'], '', 'SMS');
         $query = "REPLACE INTO `automatic_notification` $sql_text VALUES $sql_value";
         //echo $query;
         $id = sqlInsert($query, $values);
@@ -79,6 +79,7 @@ if ($result) {
     $notification_id = $result['notification_id'];
     $sms_gateway_type = $result['sms_gateway_type'];
     $provider_name = $result['provider_name'];
+    $sender = $result['email_sender'];
     $message = $result['message'];
 } else {
     $sql_msg = xl('Missing SMS config record');
@@ -118,7 +119,7 @@ $sms_gateway = array ('CLICKATELL','TMB4');
             <input type="hidden" name="type" value="SMS">
             <input type="hidden" name="notification_id" value="<?php echo attr($notification_id); ?>">
             <div class="row">
-                <div class="col-md-6 form-group">
+                <div class="col-md-4 form-group">
                     <label for="sms_gateway_type"><?php echo xlt('SMS Gateway') ?>:</label>
                     <select name="sms_gateway_type" class="form-control">
                         <option value=""><?php echo xlt('Select SMS Gateway'); ?></option>
@@ -134,7 +135,11 @@ $sms_gateway = array ('CLICKATELL','TMB4');
                         <?php }?>
                     </select>
                 </div>
-                <div class="col-md-6 form-group">
+                <div class="col-md-4 form-group">
+                    <label for="sender"><?php echo xlt('Sending Number'); ?>:</label>
+                    <input class="form-control" type="text" name="sender" size="20" value="<?php echo attr($sender); ?>" placeholder="<?php xla('sender number'); ?>">
+                </div>
+                <div class="col-md-4 form-group">
                     <label for="provider_name"><?php echo xlt('Name of Provider'); ?>:</label>
                     <input class="form-control" type="text" name="provider_name" size="40" value="<?php echo attr($provider_name); ?>" placeholder="<?php xla('provider name'); ?>">
                 </div>
--- a/modules/sms_email_reminder/cron_email_notification.php
+++ b/modules/sms_email_reminder/cron_email_notification.php
@@ -15,8 +15,10 @@
 exit;
 
 // larry :: hack add for command line version
-$_SERVER['REQUEST_URI'] = $_SERVER['PHP_SELF'];
-$_SERVER['SERVER_NAME'] = 'localhost';
+if (!array_key_exists('REQUEST_URI', $_SERVER)) {
+    $_SERVER['REQUEST_URI'] = $_SERVER['PHP_SELF'];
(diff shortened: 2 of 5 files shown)


sjpadgett:
We have a module for notifications. 

image

image

image

adunsulag:
Our desire @ChrisOstler is to move all of the SMS / appointment reminder code into the Fax/SMS module.  Super happy to see the contribution and it looks like we have another PR that adds in AWS Pinpoint in some of the same areas of code as this PR. 

Are you interested in working with @sjpadgett to incorporate your use cases and code you've done here into the Fax/SMS notifications module?  

You mentioned you are using this in a live system already so happy to see it working for you, just for the long term maintenance of the project we'd like to centralize the notification code system into a single area of responsibility.

sjpadgett:
I think you're working from an old module. FaxSMS is in core source: `interface/modules/custom_modules/oe-module-faxsms` and `interface/modules/custom_modules/oe-module-faxsms/library/rc_sms_notification.php` does now include email(new patch 2).

I have a credential table for keeping vendor credential format uniform throughout with setup scripts avaliable depending on the service type(sms,fax or email) and vendor(RC,Twilio or etherFax currently add clickatel). You don't have to use if you wanted to use the Config globals via translate or add method to emulate getCredentials().

Otherwise you are pretty spot on with architecture. If you wish I can get my phone # to you if it will help. I've developed a problem typing because of losing feeling in left hand so you'll find my posts kind of short while trying to get to the point.:)

I'm excited to get new options in the module for our users. Although, is Clickatell worth the effort? Too me it seems expensive compared to Twilio or RC, or am I wrong?

Also I haven't been able to test SMS, any vendor, for populating UI tables since I reintroduced RingCentral(removed in past due to hard vendor setup for our users) and had started creating separate UI scripts per vendor instead of trying to use just one universally.

Anyway I will open a new forum developer topic for discussions and too allow other dev's to have input.
I have to work on getting our next patch out today so not really available. But let me know if you want a zoom or phone call so we can schedule.

sjpadgett:
This may help although it's kind of old.
https://community.open-emr.org/t/openemr-fax-and-or-sms-module-guide/20273?u=sjpadgett

ChrisOstler:
> FaxSMS is in core source

Google led me wrong!  Thanks for the pointer (and also the link to the thread)!

> is Clickatell worth the effort? Too me it seems expensive compared to Twilio or RC, or am I wrong?

We had set up Clickatell originally (months ago) because it had been listed at the time as supported.  That led to fixing the Clickatell integration, thinking it would be better from a maintenance standpoint to be on a supported path.  Integrating Clickatell into the new module would give consistency to patients on the source of reminders.  We may well opt to change SMS services, but I'm trying to understand the options.

In summary - the old `sms_email_reminder` module is deprecated and unsupported, so that there's not interest in merging this patch (totally understood!).  My immediate options are to: 1) keep using this patch locally, 2) migrate to a new SMS provider and use the new module, or 3) add Clickatell support to the new module (if that's something that would be supported by the core team).

(RE price - a couple cents difference per message will be less than a couple dollars per month unless you send lots of reminder messages.  We're not yet to that point, but would love to get there!)

sjpadgett:
@ChrisOstler Thanks. I didn't realize you are using this in production or currently and was just contributing to community. Whatever the case the existing BatchCom or Clickatell has not been officially deprecated. It simply hasn't been maintained and has fallen behind with many requests to support.

I'm also not opposed to having existing brought up to date and working again, especially since it's not often we have a generous capable developer offer their services. I just thought if effort was going into any SMS it'd be best to keep in module and eventually remove BatchCom. 
So having said that perhaps you should go ahead and finish this integration and we'll revisit as I get time. Or I might go ahead later and create a new Clickatell client in module which wouldn't be to much work for me because of being familiar with process. Or we may even be able through events allow BatchCom, acting as a client, to use a module vendor.

Up to you.

sjpadgett:
Well if Clickatell is comparable to other vendors in price then would love to see yet another client in module.
So let me know how I can help. I could get the module ready for you to do the client after I help get patch out. This would give you a leg up.
I'd imagine two clients. One for Clickatell and one for Pinpoint or just start with Pinpoint. Your choice.

ChrisOstler:
Here's a fast past at re-working into the FaxSMS module: https://github.com/openemr/openemr/pull/7838 (Draft PR).  I'd love to get your feedback there as to whether that's a reasonable approach.

ChrisOstler:
Closing in favor #7838.

PR #7827, opened 2024-11-19
WIP: Implements AWS Pinpoint support for SMS

Jmevorach:
Still need to test. Wanted to get everyone's thoughts on the proposed edits and how I implemented this in the codebase.

This would bring back the "SMS Gateway Type" (SMS_GATEWAY_TYPE) multi-choice drop down in the Notifications section of the Config UI. I also updated the allowed SMS gateway types to include "AWSPINPOINT".

I didn't take "TMB4" out of the options smsnotficiation.php because I figure it's best to leave that in there in case something else I haven't seen expects it to be there. I did however not inlcude "TMB4" in globals.inc.php because as I learned from their website (https://www.tm4b.com/) they seem to be in the process of closing down.

I use the existing values of SMS_GATEWAY_USENAME, SMS_GATEWAY_PASSWORD, and SMS_GATEWAY_APIKEY. The idea is that SMS_GATEWAY_USENAME would be your AWS access key ID, SMS_GATEWAY_PASSWORD would be your AWS secret access key and SMS_GATEWAY_APIKEY would be a string of the format $aws_pinpoint_application_id:$aws_region_application_is_in (i.e. 6e0b7591a90841d2b5d93fa11143e5a7:us-east-1).

Extended the if statement in cron_sms_notification.php to look at whether or not AWSPINPOINT was used.

I didn't use the "$sender" variable because while there is a concept of a message originator in Pinpoint according to documentation "There's no additional charge for using sender IDs. However, support and requirements for sender ID authentication varies by country. Several major markets (including Canada, China, and the United States) don't support using sender IDs. Some areas require that companies who send SMS messages to individual customers must use a sender ID that's pre-registered with a regulatory agency or industry group." (https://docs.aws.amazon.com/sns/latest/dg/channels-sms-originating-identities.html) but we could add support for it if that's something that the group prefers.

The long term plan with this is that I will automate the setup of AWS Pinpoint credentials and the encoding of them as environment variables into containers in a future version of https://github.com/openemr/host-openemr-on-aws-fargate. This would provide an out of the box SMS setup that should work pretty much anywhere.

Hope this message finds everyone well!

Fixes #

### DIFF
--- a/.github/workflows/docker-test-flex-324.yml
+++ b/.github/workflows/docker-test-flex-324.yml
@@ -1,0 +1,39 @@
+# This will run the OpenEMR flex dockers with Alpine 3.24 version. It will test the flex docker images in both
+#  development (ie. dev) mode (codebase used from a shared volume) and production (ie. prod) mode (codebase collected via git).
+#  - is_production_docker is required and needs to be false.
+#  - alpine_version is set to 3.24
+#  - php_versions are set to 8.3, 8.4 and 8.5
+#  - flex_timing is set to true to emit TIMING markers in openemr.sh (openemr-devops#797).
+#  - note not using the coverage settings (is_flex_coverage_docker, flex_coverage_php_version) in this workflow. see
+#    test-flex-edge.yml for use of these settings.
+
+name: OpenEMR Flex 3.24 Docker Test
+
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - '.github/workflows/docker-test-flex-324.yml'
+    - 'docker/flex/**'
+    - '.github/actions/test-actions-core/action.yml'
+    - '.github/workflows/docker-test-core.yml'
+    - '.github/docker/compose.yml'
+  pull_request:
+    branches:
+    - master
+    paths:
+    - '.github/workflows/docker-test-flex-324.yml'
+    - 'docker/flex/**'
+    - '.github/actions/test-actions-core/action.yml'
+    - '.github/workflows/docker-test-core.yml'
+    - '.github/docker/compose.yml'
+
+jobs:
+  build:
+    uses: ./.github/workflows/docker-test-core.yml
+    with:
+      is_production_docker: false
+      alpine_version: "3.24"
+      php_versions: '["8.3", "8.4", "8.5"]'
+      flex_timing: true


adunsulag:
Hey @Jmevorach love to see the improvement in the notification system and to make it possible to use AWS Pinpoint.  We actually have a SMS/Fax module that @sjpadgett put together and our plan is to try and move all forms of SMS/Fax notifications into that area of the code to keep things more maintainable. 

Would you be interested in working with @sjpadgett to incorporate what you've done here into the Fax/SMS notifications module?  Part of this work would likely be moving the batchcom code to leverage the SMS/Fax module code (which is being used in other areas of the codebase).  I think this would accomplish your objective of making sure all SMS mechanisms can be delivered via AWS PinPoint.

Jmevorach:
Greetings @adunsulag and @sjpadgett! Hope you're both having good days! Appreciate you both taking a look at this.

Sure sounds good. I'm happy to put this anywhere. Also please feel free to close the PR if you think it would be easier to just start from scratch. 

What I'm really after with this is not so much implementing Pinpoint support `per se`. What I'm more interested in is providing support for a set of SMS credentials that I can automate the creation of. The [latest PR for the architecture we're working on](https://github.com/openemr/host-openemr-on-aws-fargate/pull/29) has support for automating the provisioning of SES credentials and then signs those into functioning SMTP credentials during build and encodes them as global environment variables so that when the, off the shelf, v7.0.2 containers pulled straight from DockerHub on boot the user is more or less ready to get going sending emails without really having to set up SMTP with a third party service. This way it's just included as part of the architecture and if you change "configure_ses" in the cdk.json from "false" to "true" you get all that set up automatically.

As long as we have a way to send messages with Pinpoint and I can pass credentials to it on boot as environment variables then I'll be able to do the same thing for us for SMS. I see SMS as just the lowest hanging fruit here. The other two things I think I'll probably look into for the projects we have going on will be biometrics (with [Rekognition](https://aws.amazon.com/rekognition/) and telehealth scheduling (with [Chime SDK](https://aws.amazon.com/blogs/business-productivity/embed-healthcare-appointment-scheduling-widget-with-the-amazon-chime-sdk/).

Also for the work I'm planning to do on OpenEMR I'm planning to be a soldier in the core team's army! I'll take your lead here so if there's a place you think this should go that would be better than here I'll gladly put it there. Full disclosure though I'm planning to take it **_pretty easy_** these next few 6 to 8 weeks as far as any coding on Github is concerned. In early January I'll be back at it but I won't take any offense in the meantime if someone wants to take this and run with it.

PR #7838, opened 2024-11-22
Add Clickatell integration to FaxSMS module.

### LINKED ISSUE #7822: feat: Appointment reminder notifications Clickatell Integration

ChrisOstler:
Fixes #7822

This adds support for Clickatell as an SMS provider in the FaxSMS module.  It has been tested to work both for sending using API actions (through the Secure Messaging Send SMS) and notifications (rc_sms_notification.php).  The data tables in the Services/SMS page display, including entries for sent SMS reminder notifications.


interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
```diff
@@ -49,6 +50,7 @@ public function __construct($type = null)
         if (empty(self::$_apiModule)) {
             self::$_apiModule = $_REQUEST['type'] ?? $_SESSION["oefax_current_module_type"] ?? null;
         }
+        $this->crypto = new CryptoGen();
```
ChrisOstler, line 53: Was there a reason instantiating this was deferred to the sub classes?  It seemed odd not to just do it here, since all sub classes would need it for `saveSetup()`.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
```diff
@@ -619,4 +625,12 @@ public function verifyAcl($sect = 'patients', $v = 'docs', $u = ''): bool
     {
         return AclMain::aclCheckCore($sect, $v, $u);
     }
+
+    /**
+     * @return null
+     */
+    protected function index()
```
ChrisOstler, line 632: This was invoked as part of the `dispatchActions()`.  I didn't take the time to figure it out, but implementations in other sub classes didn't seem to do anything, so I thought it would simplify new providers if there was a worked default to use.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return $credentials;
+    }
+
+    /**
+     * @return string
+     */
+    public function sendSMS(string $toPhone = '', string $subject = '', string $message = '', string $from = ''): string
+    {
+        // $subject and $from do not have valid/meaningful values passed in
```
ChrisOstler, line 44: I was curious why there were parameters for subject and from, when there were not values passed in.  The notification script generates the `$db_sms_msg`, but doesn't populate those fields and [falls back to empty strings](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/library/rc_sms_notification.php#L179-L183), instead.

sjpadgett: message comes from dialogs or passed in. $db_sms_msg is a canned template in setups.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
sjpadgett: I believe index was meant for the API use but unsure.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
ChrisOstler: In the old scripts, `$db_sms_msg` was queried from the database.  In the new script, it instead is [generated in place in the script](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/library/rc_sms_notification.php#L116-L118).  Since we never set values for `email_subject` or `email_sender`, when we try to pull out values for parameters to `sendSMS()` we'll always [fall back to the empty strings](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/library/rc_sms_notification.php#L179-L183).

(Since an SMS doesn't have a separate subject I wouldn't expect to even have it as a parameter, unless it was to unify an interface with email. For emails, though, there aren't parameters for sender or subject, and instead those are [hardcoded or global variables](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/src/Controller/EmailClient.php#L144-L148).  Not anything that needs to be covered for this PR, but something I noticed on the way as a fresh set of eyes.)

ChrisOstler:
Thanks for looking at this, @sjpadgett .  Is there anything that you'd want to point in a different direction, or should I go ahead and publish this (and close #7823)?

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
sjpadgett: If you don't need then don't use. I'm unsure why they're there and may go back to five years ago. Message/comment comes from send SMS dialog events. Some internal send events allow for sending to both email and SMS especially sending onetime tokens. Some dialogs were shared at one time to conserve on firing events.
Look at NotificationEventListener.php and interface/modules/custom_modules/oe-module-faxsms/openemr.bootstrap.php

sjpadgett:
Close other and we'll get this in after I look over and do some testing.
Are you going to be on call tomorrow?

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
```diff
@@ -0,0 +1,117 @@
...
+
+        /* Reformat $toPhone number */
+        $cleanup_chr = array ("+", " ", "(", ")", "\r", "\n", "\r\n");
+        $toPhone = str_replace($cleanup_chr, "", $toPhone);
+        if (!str_starts_with($toPhone, "1")) {
```
sjpadgett, line 49: Because we don't know if this method is being called from API or a send dialog need to test for and resolve:

      $authErrorMsg = $this->authenticate(); // for API calls at minimum
        if ($authErrorMsg !== 1) {
            return text(js_escape($authErrorMsg));
            // goes to alert
        }
       
       $toPhone = $toPhone ?: $this->getRequest('phone');
        $from = $from ?: $this->getRequest('from');
        $message = $message ?: $this->getRequest('comments');

        $toPhone = $this->formatPhoneForSave($toPhone);

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
```diff
@@ -0,0 +1,117 @@
...
+    private $phone;
+
+    public function __construct()
+    {
+        if (empty($GLOBALS['oefax_enable_sms'] ?? null)) {
+            throw new \RuntimeException(xlt("Access denied! Module not enabled"));
+        }
+        parent::__construct();
+    }
+
```
sjpadgett, line 25: Check that user has permission via ACL in dispatch. Required by all dialogs or MessageUI

/**
     * @param $acl
     * @return int
     */
```
    public function authenticate($acl = ['patient', 'doc']): int
    {
          list($s, $v) = $acl;
          return $this->verifyAcl($s, $v);
    }
```

sjpadgett (changes requested):
I'm impressed, good job!
This will help others to add clients I hope.
There are some consolidations I see for later.
Once you're done I do a quick test then we can merge.
After doing these modifications the Send SMS button in Secure Messaging will work.

If planning on doing Pinpoint should be it's own PR which I'm very much looking forward.

Thanks for shifting gears and taking up the module. Doing so helps keep a single focus for all our notifications.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
ChrisOstler: I added this to AppDispatch, assuming that other vendor modules will need the same logic.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
```diff
@@ -49,6 +50,8 @@ public function __construct($type = null)
         if (empty(self::$_apiModule)) {
             self::$_apiModule = $_REQUEST['type'] ?? $_SESSION["oefax_current_module_type"] ?? null;
         }
+        $this->crypto = new CryptoGen();
+        $this->getCredentials();
```
ChrisOstler, line 54: Does this make sense?  It seems like for API handling we were having to sprinkle calls to `getCredentials()` around to make sure that the config had been loaded (and triggered the corresponding side effects).  This front loads it so that it's ready before calling `dispatchAction()`.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
```diff
@@ -545,7 +554,7 @@ public static function getLoggedInUser(): array
     public function getPatientDetails(): bool|string
     {
         $id = $this->getRequest('pid');
-        $query = "SELECT fname, lname, phone_cell, email FROM Patient_data WHERE pid = ?";
+        $query = "SELECT fname, lname, phone_cell, email FROM patient_data WHERE pid = ?";
```
ChrisOstler, line 557: Turns out that on Linux, table names are case sensitive...

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
sjpadgett: What side effects? I prefer to call credentials from client on demand. Please leave this way.

sjpadgett (approved):
Once minor changes done we can merge if you want.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
ChrisOstler: I had originally seen/modeled on the Twilio integration, which [sets fields](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/src/Controller/TwilioSMSClient.php#L76-L82) as a side effect of loading the credentials.  The [email](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/src/Controller/EmailClient.php#L51-L56) and [EtherFAX](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/src/Controller/EtherFaxActions.php#L65-L69) integrations do the same thing, though the RingCentral one [does not](https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-faxsms/src/Controller/RCFaxClient.php#L236).

Looking more closely, many of those seem like they were just copy/paste and not needed.  Also, all but email *also* call `getCredentials()` in the subclass constructor, and then store that, as well.

Given all of that, I'd lean towards adding a protected `credentials` field to `AppDispatch`, and populating it by calling `getCredentials()` in `AppDispatch::_construct()`.  That may mean cleaning up all the others that have already declared `credentials` fields - I'm not a PHP dev so not sure how the language handles field hiding.  (Then the side effects of setting fields in `getCredentials()` could also be removed.)

How would you like me to handle this?

interface/modules/custom_modules/oe-module-faxsms/openemr.bootstrap.php
```diff
@@ -62,13 +62,26 @@ function oe_module_faxsms_add_menu_item(MenuEvent $event): MenuEvent
...
+    };
+
     $menu = $event->getMenu();
     // Our SMS menu
     $menuItem = new stdClass();
     $menuItem->requirement = 0;
     $menuItem->target = 'sms';
     $menuItem->menu_id = 'mod0';
-    $menuItem->label = xlt("SMS");
+    $menuItem->label = xlt($sms_label);
```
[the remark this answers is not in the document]
ChrisOstler: Yeah.  Good catch.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
sjpadgett: Again like I said I had this in the beginning and had to change because of an issue I can't remember. Could have been contention or scope was lost.
You are stating the more logical way but I'm sure I did this by design. Why are you having a problem calling from your client?
Also RC used to be 2FA with caching tokens so was handled differently until I recently went JWT.

I tell you what. I have some refactors to do for our next release so I'll clean this issue up then when I can do a full test suite.
I'm just fearful of any patches until next release having this change done.

ChrisOstler: > Why are you having a problem calling from your client?

I could certainly make a late call to `getCredentials()` in `sendSMS()`.  I've been trying to take an eye to "how can the next integration be made simpler" as I've run into things.  I'm quite happy to whichever approach you'd prefer; these are just my takes on potential simplifications.

This particular one (credentials) is interesting because of the timing of `dispatchAction()` - it's invoked at the end of `AppDispatch::__construct()`.  That means that e.g. `sendSms()` will be called _before_ `parent::__construct()` returns.  I believe that this is why `crypto` was being populated in the sub classes `__construct()` - they also had to call `getCredentials()` (which depends on `crypto`) *before* invoking `parent::__construct()` to have them before `sendSMS()` was called, and so had to make sure that `crypto` was populated.

One alternative is to JIT load credentials where/when they are needed.  This has the downside of having potentially multiple locations in an integration that do so.  Another alternative is to populate a `credentials` field once in `AppDispatch::__construct()` so that it will be available always.

(I will note that the current approach of copying data to properties from credentials has the benefit of localizing the extraction/mapping of credential fields to named fields to a single place, rather than potentially multiple places that credentials are used.  That can't be done without using side effects due to the timing of the `parent::__construct()` invocation.)

I understand your concern about keeping changes minimal for a patch.  What's your preferred approach:
1. Keep current implementation of calling `getCredentials()` in `AppDispatch::__construct()` and use side effects to capture what's needed (this means that `getCredentials()` will be called twice for most current integrations)
2. Instantiate `crypto` in `ClickatellSMSClient::__construct()` and call `getCredentials()` directly (like other integrations)
3. Wait and call `getCredentials()` in `sendSMS()`, since for this integration that's the only place they will be needed

I'll happily take whichever approach you'd like.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
sjpadgett: This method in client was originally meant for authenticating vendor API account, Twilio I believe.
I keep in client for security of bad actors and a must have for dialog use.

What needs to happen is implementing an alias for sendSMS/Fax regarding how called, API or internal. I think I did this somewhere but again old age memory fails me.

On another matter I'm thinking of adding the capability/option to use my Document Templates for notification messages instead of the canned message in setup. This would allow for better formatting and addition data points.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/ClickatellSMSClient.php
```diff
@@ -61,9 +45,9 @@ public function sendSMS($toPhone = '', string $subject = '', string $message = '
 
         $url = sprintf(
             "https://platform.clickatell.com/messages/http/send?apiKey=%s&to=%s&from=%s&content=%s",
-            $this->appKey,
+            $this->credentials['appKey'],
```
sjpadgett, line 48: Just a thought. You may want to validate credentials first. Could return empty.

ChrisOstler: Good thought.  I'd lean towards keeping that check as close as possible to the user entering it.  Right now, the input fields on the service setup form (in `setup.php`) have a `required` attribute, so that validation will happen when the form is submitted.  In the case that the credentials are edited some other way, the API call will fail with an error message that will get exposed to the user.

As for checking that `$this->credentials` is actually set, since we're setting that in the constructor, the only way it could be null is if we made some logic error, which a user couldn't do anything about.  The send will fail one way or another.

My two cents would be to not add a check here since null `credentials` isn't an expected/possible condition.

sjpadgett: That's fine. It'd be very rare for me to try to force any change on you. So in future you don't have to justification just say you disagree or prefer to do as you have it.
Too be honest, I'm horrible at reviews and avoid them when I can so take that into account:)
If I feel strongly about something, I'll let you know. You're doing a great job and we're lucky to have you.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
sjpadgett: Okay as long as we don't have static properties for credentials which I don't think we have then should be okay.
Still for some reason I remember an issue concerning background task requests and credentials so maybe I'll take the time to research just to satisfy myself. Keep as you have it. I may go ahead and refactor other clients and test.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
bradymiller: yep, OpenEMR learned this lesson the hard way a long back on the form_CAMOS_* tables (when install the CAMOS form). If install on windows (not case sensitive for table names), then it is made to be lower case, and if then migrate that install to linux, it would then break (since linux is case sensitive). We had to create a special function mechanism so it wouldn't break the CAMOS forms: 
https://github.com/openemr/openemr/blob/3298e343649730c5fd353ff93592db980fbd3dd6/library/formdata.inc.php#L182-L192

adunsulag:
@sjpadgett is this good to bring this in at this point?  Looks like you approved it so I think we can bring this in.

PR #7873, opened 2024-12-16
Quick escape issues fix

### LINKED ISSUE #7876: Relax permissions for faxsms notifications
### Describe the problem
Currently only admins may send notifications from U.I.
Change to any with patients appointment permissions.

sjpadgett:
status needs to preserve html for display
missed endings for refactor escaping and translations.

Fixes #7876


bradymiller:
2DV

PR #7881, opened 2024-12-21
Correctly replace invalid code unit sequences

tsimonq2:
When trying to generate a report for an LBF form with a long text field containing invalid characters, an empty string is returned and the data is missing from the report.

This simply adds `ENT_SUBSTITUTE` which will do substitution instead of returning the empty string.

### DIFF
--- a/library/options.inc.php
+++ b/library/options.inc.php
@@ -2440,9 +2440,9 @@ function generate_display_field($frow, $currvalue)
             }
         }
     } elseif ($data_type == 2) { // simple text field
-        $s = nl2br(htmlspecialchars($currvalue, ENT_NOQUOTES));
+        $s = nl2br(htmlspecialchars($currvalue, ENT_NOQUOTES | ENT_SUBSTITUTE));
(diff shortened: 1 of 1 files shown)


bradymiller:
hi @tsimonq2 ,
If this is happening here, then likely in lots of other places will experience issues since htmlspecialchars is used everywhere in the codebase to prevent XSS vulnerabilities. May be an opportunity to fix this on a systemic issue since most of our calls to this function are centralized via the text()/attr() functions. Couple quick questions:
1. What encoding are you using? (this is generally UTF-8 which htmlspecialchars uses by default but looks like will experience your issue if using another encoding)?
2. What are the illegal characters?

kojiromike:
Thanks for this fix, @tsimonq2 — you identified a real problem with `htmlspecialchars()` silently dropping data on invalid byte sequences.

Since there hasn't been activity here in over a year, I'm going to close this and track the fix in #10554. If you'd like to pick it up there, you're welcome to.

PR #7889, opened 2024-12-23
Fix and refactor contrib form scanned_notes.

### LINKED ISSUE #7888: Contrib form scanned_notes need to be fixed
### Describe the problem
ImageMagick command line syntax changed and form save and delete misbehave.
Also need to update to bootstrap.

sjpadgett:
mostly refactor form and add correct form actions. remove empty check in sanitize check_file_dir_name() method. Caused string '0' label to fail also empty string which is acceptable.

Fixes #7888

### DIFF
--- a/library/js/utility.js
+++ b/library/js/utility.js
@@ -53,6 +53,12 @@ if (typeof htmlEscapesText === 'undefined') {
             return htmlEscapesAttr[match];
         });
     };
+    jsXla = function (string) {
+        return jsAttr(xl(string));
+    };
+    jsXlt = function (string) {
+        return jsText(xl(string));
+    };
 }
 
 // another useful function
--- a/contrib/forms/scanned_notes/new.php
+++ b/contrib/forms/scanned_notes/new.php
@@ -2,15 +2,18 @@
 
 /**
  * Encounter form for entering clinical data as a scanned document.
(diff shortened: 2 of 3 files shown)


sjpadgett:
This has turned out very nice if I say so. It could be worthwhile adding to default form in form admin.

image

library/js/utility.js
```diff
@@ -53,6 +53,12 @@ if (typeof htmlEscapesText === 'undefined') {
             return htmlEscapesAttr[match];
         });
     };
+    jsXla = function (string) {
+        return jsAttr(xl(string));
+    };
+    jsXlt = function (string) {
+        return jsText(xl(string));
+    };
```
bradymiller, line 61: 2DV

bradymiller:
Nice work.
2 things that would work on if goal is to get this into main forms directory:
1. Encrypt files saved in the documents directory (note there is a way to then only decrypt files that have been encrypted to not break backwards compatibility)
2. Utilize OpenEMR's documents class for importing documents rather than directly copying them into the documents directory (this would then also fix 1 above automatically)

sjpadgett:
Thanks Brady! I'll consider bringing into core and definitely will use document class and encrypt. Just now I'm bored with it.:)

PR #7891, opened 2024-12-25
chore: update packages in package.json and ccdaservice/package.json

bradymiller:
update packages in package.json and ccdaservice/package.json


ccdaservice/serveccda.js
```diff
@@ -2164,7 +2164,7 @@ function populateParticipant(participant) {
         "name": {
             "prefix": participant.prefix || "",
             "suffix": participant.suffix || "",
-            "middle": [participant.mname] || "",
+            "middle": [participant.mname ?? ""],
```
bradymiller, line 2167: hi @sjpadgett ,
The new version js linter was throwing "Unexpected constant truthiness on the left-hand side of a || expression". Turns out an array is always truthy whether empty or not, so I asked chatgpt how to fix it :). This is testing well (ccda creating works both when there is and is not a middle name). Just making sure you are ok with this (also did it one more time below).

bradymiller:
testing ok, so bringing in.

PR #7895, opened 2024-12-28
The toencounter and toinsurance issue.

### LINKED ISSUE #7874: bug: toencounter billing manager bug

sjpadgett:
Fixes #7874 

- create async/await functionality to make bit more synchronous
- new async loadtab function.
- added javascript function in top to fetch session key/value pairs.

### DIFF
--- a/library/ajax/set_pt.php
+++ b/library/ajax/set_pt.php
@@ -3,12 +3,14 @@
 /**
  * sets pid
  *
- * @package OpenEMR
- * @link    https://www.open-emr.org
- * @author  Brady Miller <brady.g.miller@gmail.com>
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @author    Jerry Padgett <sjpadgett@gmail.com>
+ * @copyright Copyright (c) 2024 Jerry Padgett <sjpadgett@gmail.com>
  * @copyright Copyright (c) 2017 Brady Miller <brady.g.miller@gmail.com>
- * @license https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
-*/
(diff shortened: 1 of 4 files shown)


sjpadgett:
@stephenwaite I still need to test this version on Linux.

stephenwaite:
southpark_cheering

library/ajax/set_pt.php
```diff
@@ -22,3 +24,10 @@
 if ($_GET["set_pid"] && $_GET["set_pid"] != $_SESSION["pid"]) {
     setpid($_GET["set_pid"]);
 }
+
+// For future use, if needed
```
bradymiller, line 28: There is a security issue here allowing access to all session values. For example, if bad actor has access to 'csrf_private_key' session value (which can be done ), then they can broadly bypass all the csrf protections.

We should whitelist which session values that a user can get from this script.

sjpadgett: Agreed, thanks Brady. Though I decided I didn't need the feature I think I'll move ajax to utility.js and limit key access.

stephenwaite: removed it in my pending PR, https://github.com/openemr/openemr/pull/7896/commits/dcb644aa3822745a88840cac0a2a98887a535e02

adunsulag: I saw @stephenwaite removed some of the functionality used in main.php, but doesn't this break the getSessionValue functionality in billing.php?  Since the set_pt.php no longer returns a pid value it seems like all of that needs to be changed.  Or am I missing something?

see here https://github.com/openemr/openemr/blob/master/interface/billing/billing_report.php#L355C13-L355C59

Do we need to cleanup the main.php method where getSessionValue is still stored?

PR #7907, opened 2025-01-07
load form from module

### LINKED ISSUE #7905: Adding a form to the Encounter form from a module
### Describe the problem

When creating a form that is in a module to appear in the encounter. There needs to be a hook in the load_form that can bring in the form from the module into the encounter. 

In the load_form logic add the hook that would look for the module and point to the module directory to load the form.

juggernautsei:
Fixes #7905

Adds a pointer to the load_form logic to point to the modules folder and use the directory name that is passed 

No

### DIFF
--- a/interface/patient_file/encounter/load_form.php
+++ b/interface/patient_file/encounter/load_form.php
@@ -26,7 +26,7 @@ if (substr($_GET["formname"], 0, 3) === 'LBF') {
     //Change the directory to meet module requirements
     $dir = $incdir . "/modules/custom_modules/$_GET[formname]/public/index.php";
     include_once($dir);
-}  else {
+} else {
     if ((!empty($_GET['pid'])) && ($_GET['pid'] > 0)) {
         $pid = $_GET['pid'];
         $encounter = $_GET['encounter'];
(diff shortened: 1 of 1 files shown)


adunsulag:
@juggernautsei there seems to be a problem with the Recall board in the e2e tests, that seems pretty odd as I can't see how your PR would mess with that.  We'll have to see if its something with the CI engine.

interface/patient_file/encounter/load_form.php
```diff
@@ -21,6 +21,11 @@
 if (substr($_GET["formname"], 0, 3) === 'LBF') {
```
adunsulag, line 21: @juggernautsei do you have thoughts on making this an event instead of hard-coding the directory?  This is short and simple but not as flexible, just wondering your thoughts on this?  I need to actually deal with loading forms in the modules as well and my preference is using events but I'd like to know your thoughts on this one over an event piece?

juggernautsei: Stephen, I appreciate you asking me about this. I did this as a precursor to making it what you are suggesting an event. This is working and I was trying to be thoughtful of using the standard nomenclature the modules were started with even though you can call a module anything and it will work now. But it would not take to much to add and event here and anyone listening can act. I can work on adding that event. I have some work around this that will not take me far out of the way. Should I leave this open or close it?

adunsulag: So you bypass the form registry all together?

juggernautsei: yes

adunsulag: @juggernautsei How are you handling the report.php and the view.php pieces that come into play with forms.php and view_form.php in the encounter display?  I'm moving a custom form for a customer out of core into their module and it looks like those areas have to be touched as well.

PR #7926, opened 2025-01-16
Create API-Endpoints for Therapy Groups

BerwoutK:
Fixes #

This pull request introduces functionality to manage therapy groups within OpenEMR. It includes the creation and updating of API components to allow CRUD operations for therapy groups, as well as the ability to manage patients within those groups.

TherapyGroupsRestController and TherapyGroupsService:

    Added TherapyGroupsRestController with methods to handle:
        post: Create a new therapy group.
        getAll: Retrieve all therapy groups.
        addPatient: Add a patient to a therapy group.
        removePatient: Remove a patient from a therapy group.
        getAllForPatient: Retrieve all therapy groups associated with a specific patient.
    Added TherapyGroupsService with the following methods:
        insert: Insert a new therapy group into the database.
        getAll: Fetch all therapy groups.
        getAllForPatient: Fetch therapy groups for a specific patient.
        search: Search for therapy groups.
        addPatient: Add a patient to a therapy group.
        removePatient: Remove a patient from a therapy group.

API Updates:

    Updated _rest_routes.inc.php to include new routes for therapy group operations.
    Updated PatientRestController and PatientService:
        Added getAllForTherapyGroup: Retrieves all patients in a specific therapy group.
        Added removePatient: Removes a patient from a therapy group.
    Enhanced scope handling in docker/library/api-scope-listing and ScopeRepository to include:
        user/therapy_groups.read
        user/therapy_groups.write

Validation:

    Added TherapyGroupsValidator to ensure required fields for therapy groups are validated before processing.

No


adunsulag:
Tested all endpoints everything working as expected.  Will bring in once CI engine finished.  Will go into next patch as we've frozen all release features now for 7.0.3

_rest_routes.inc.php
```diff
@@ -157,7 +159,8 @@
  *              "api:port": "Standard Patient Portal OpenEMR API",
  *              "patient/encounter.read": "Read encounters the patient has access to (api:port)",
  *              "patient/patient.read": "Write encounters the patient has access to (api:port)",
- *              "patient/appointment.read": "Read appointments the patient has access to (api:port)"
+ *              "patient/appointment.read": "Read appointments the patient has access to (api:port)",
+ *          }   "user/participants.read": "Read participants the user has access to (api:oemr)"
```
kojiromike, line 163: This line is malformed: there's a stray closing `}` mid-comment and a `"user/participants.read"` entry sitting outside the JSON shape. The next line is the actual closing `}` of the scopes block. This breaks the @OA OAuthFlow doc parse. Drop this line entirely (and see the separate comment about `user/participants.read` being unused).

_rest_routes.inc.php
```diff
@@ -535,6 +539,53 @@
...
+    /**
+     *  @OA\Post(
+     *      path="/api/therapy_groups",
+     *      description="Creates a new therapy group.",
+     *      tags={"standard"},
+     *      security={{"openemr_auth":{}}}
+     * )
+     */
+    "POST /api/therapy_groups" => function () {
+        RestConfig::authorization_check("admin", "users");
```
kojiromike, line 551: `authorization_check("admin", "users")` requires admin-level user-management ACL to create a therapy group. That's not the right gate — a clinician needs to create therapy groups without being a system admin. Should this be `authorization_check("groups", "glog")` (or whatever therapy_groups ACL exists), matching what the per-patient route uses (`"therapy_groups", "med"`)?

_rest_routes.inc.php
```diff
@@ -535,6 +539,53 @@
...
+    /**
+     *  @OA\Get(
+         path="/api/therapy_groups",
+         description="Retrieves a list of all therapy groups.",
+         tags={"standard"},
+         security={{"openemr_auth":{}}}
+     )
+     */
+    "GET /api/therapy_groups" => function () {
+        RestConfig::authorization_check("admin", "users");
```
kojiromike, line 567: Same authorization concern as the POST route — listing therapy groups shouldn't require admin/users.

_rest_routes.inc.php
```diff
@@ -535,6 +539,53 @@
...
+    /**
+     *  @OA\Get(
+     *    path="/api/therapy_groups/:tgid/participants",
+     *    description="Retrieves a list of all participants in a therapy group.",
+     *    tags={"standard"},
+     *    security={{"openemr_auth":{}}}
+     *  )
+     */
+    "GET /api/therapy_groups/:tgid/participants" => function ($tgid) {
+        RestConfig::authorization_check("admin", "users");
```
kojiromike, line 582: Three concerns on this route:

1. `authorization_check("admin", "users")` is the wrong ACL — listing the patients in a therapy group is patient/clinical data, not admin/users.
2. `$tgid` comes straight from the URL as an arbitrary string and is forwarded into a SQL query inside `PatientService::getAllForTherapyGroup()`. The query is parameterized so it isn't injectable, but there's no validation that `$tgid` is a positive integer; a malformed value will silently return an empty result with no 400.
3. This route enumerates patient PHI by group_id with no further authorization check against the requesting user's relationship to the group. Even with the right ACL, consider whether group membership/ownership should also be checked.

_rest_routes.inc.php
```diff
@@ -1727,6 +1803,73 @@
...
+     *              type="string"
+     *          )
+     *      ),
+     *     security={{"openemr_auth":{}}},
+     *   )
+     */
+    "PUT /api/patient/:puuid/therapy_groups/:pgid" => function ($puuid, $pgid) {
+        RestConfig::authorization_check("patients", "med");
+        $data = (array)(json_decode(file_get_contents("php://input")));
+        $return = (new TherapyGroupsRestController())->addPatient($puuid, $pgid, $data);
```
kojiromike, line 1835: `$pgid` is typed `int $gpid` on the controller method but flows through here as the raw string from the URL. Either cast/validate it here, or change the controller signature to `string` and validate inside the service. The current state will trip strict_types if the caller ever passes a non-numeric segment.

_rest_routes.inc.php
```diff
@@ -1727,6 +1803,73 @@
...
+        RestConfig::apiLog($return, $data);
+        return $return;
+    },
+
+    /**
+     * @OA\Schema(
```
kojiromike, line 1841: `@OA\Schema` is the wrong annotation for a DELETE endpoint — should be `@OA\Delete`. As written, swagger-php will treat this as a component schema definition with a (now invalid) `path` property, not as a documented DELETE operation.

_rest_routes.inc.php
```diff
@@ -1727,6 +1803,73 @@
...
+     *              type="string"
+     *          )
+     *      ),
+     *     security={{"openemr_auth":{}}},
+     )
+     */
+    "DELETE /api/patient/:puuid/therapy_groups/:pgid" => function ($puuid, $pgid) {
+        RestConfig::authorization_check("patients", "med");
+        $return = (new TherapyGroupsRestController())->removePatient($puuid, $pgid);
+        RestConfig::apiLog($return, $data);
```
kojiromike, line 1869: `$data` is undefined in this closure — there's no `json_decode(file_get_contents(...))` above. With `declare(strict_types)` and PHPStan level 10 this is a clear error; otherwise it logs a warning and passes `null` to `apiLog`. Drop the second argument (DELETE has no body), or read and pass the body if you actually want it logged.

src/Common/Auth/OpenIDConnect/Repositories/ScopeRepository.php
```diff
@@ -602,11 +602,14 @@ public function apiScopes(): array
...
             "user/surgery.read",
             "user/surgery.write",
+            "user/therapy_groups.read",
+            "user/therapy_groups.write",
             "user/transaction.read",
             "user/transaction.write",
             "user/user.read",
             "user/vital.read",
             "user/vital.write",
+            "user/participants.read",
```
kojiromike, line 612: `user/participants.read` is added to the scope list but no route, controller, or check references it. Either remove it, or wire it up to the `/api/therapy_groups/:tgid/participants` route (which currently uses an admin ACL instead of any scope check). The name is also ambiguous — `participants` of what? `therapy_group_participants.read` would be self-describing and consistent with `therapy_groups.read`.

src/RestControllers/TherapyGroupsRestController.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
kojiromike, line 1: Missing `declare(strict_types=1);` — required for new files per the repo's CLAUDE.md ("Every new PHP file starts with `declare(strict_types=1)`"). Same applies to `TherapyGroupsService.php` and `TherapyGroupsValidator.php`.

src/RestControllers/TherapyGroupsRestController.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function getAll($search = []): array
+    {
+        $processingResult = $this->therapyGroupService->getAll($search);
+
+        return RestControllerHelper::handleProcessingResult($processingResult, 200, true);
+    }
+
+    public function addPatient(string $puuid, int $gpid, array $data): array
```
kojiromike, line 30: Parameter `int $gpid` here, but the service method declares `string $id` and the route closure passes the URL segment through as a string. Pick one type and apply it consistently end-to-end — `int` with validation at the boundary is cleaner.

src/Services/TherapyGroupsService.php
```diff
@@ -0,0 +1,175 @@
...
+            'group_name' => $data['group_name'],
+            'group_start_date' => $data['group_start_date'] ?? date('Y-m-d H:i:s'),
+            'group_type' => $data['group_type'],
+            'group_participation' => $data['group_participation'],
+            'group_status' => $data['group_status'],
+        ];
+
+        $query = $this->buildInsertColumns($data);
+
+        $results = sqlInsert("INSERT INTO " . $this::TABLE_NAME . " SET " . $query['set'], $query['bind']);
```
kojiromike, line 47: `sqlInsert("INSERT INTO " . $this::TABLE_NAME . ...)` — `self::TABLE_NAME` is conventional here (and `$this::` invites confusion). More importantly, please switch the new service code to `QueryUtils` (used in `search()` already) rather than the deprecated legacy `sqlInsert`/`sqlStatement`/`sqlFetchArray` globals — the repo's CLAUDE.md calls these out as legacy patterns for new code in `/src/`.

src/Services/TherapyGroupsService.php
```diff
@@ -0,0 +1,175 @@
...
+
+        if (!$results) {
```
kojiromike, line 49: If the insert fails, an internal error is added to `$processingResult` but the code then unconditionally calls `$processingResult->addData(['group_id' => $results])` with the falsy value. The caller gets both an error *and* a misleading `group_id: false` (or `null`). Wrap the `addData` in an `else` so the failure path returns errors only.

src/Services/TherapyGroupsService.php
```diff
@@ -0,0 +1,175 @@
...
+
+        $sql = <<<SQL
+SELECT tg.*
+FROM therapy_groups tg
+INNER JOIN therapy_groups_participants tgp ON tg.group_id = tgp.group_id
+INNER JOIN patient_data pd on pd.pid = tgp.pid
+WHERE pd.uuid = ?
+SQL;
+
+        $results = sqlStatement($sql,[$puuidBytes]);
```
kojiromike, line 89: Missing space after the comma: `sqlStatement($sql, [$puuidBytes])`. Also consider `QueryUtils::fetchRecords()` instead of the legacy `sqlStatement` + `sqlFetchArray` loop — cleaner and matches the `search()` style above.

src/Services/TherapyGroupsService.php
```diff
@@ -0,0 +1,175 @@
...
+    public function addPatient(string $puuid, string $id, array $data): ProcessingResult
+    {
+        $processingResult = new ProcessingResult();
+
+        $data = [
+            'pid' => $this->getIdByUuid(UuidRegistry::uuidToBytes($puuid), 'patient_data', "pid"),
+            'group_id' => $id,
+        ];
+
+        $query = $this->buildInsertColumns($data);
```
kojiromike, line 123: `$query = $this->buildInsertColumns($data)` is dead — `$query` is never used; the SQL below is hand-written instead. Either delete this line, or actually use `$query['set']`/`$query['bind']` (which would also let `group_patient_status` flow through naturally instead of being hardcoded below). Same dead `buildInsertColumns` call in `removePatient()`.

src/Services/TherapyGroupsService.php
```diff
@@ -0,0 +1,175 @@
...
+
+        $sql = "INSERT INTO therapy_groups_participants SET ";
+        $sql .= "pid = ?, group_id = ?, group_patient_status = ?";
+
+        $results = sqlInsert($sql, [
+            $data['pid'],
+            $data['group_id'],
+            $data['group_patient_status'] ?? 10,
```
kojiromike, line 131: Hardcoded magic number `10` for the default `group_patient_status` with no validation that callers can't override it to an arbitrary int. Worth either an enum/constant referencing the meaning, or a validator context that whitelists allowed values.

src/Services/TherapyGroupsService.php
```diff
@@ -0,0 +1,175 @@
...
+        ]);
+
+        if ($results) {
+            $processingResult->addData([
+                'id' => $results,
+            ]);
+        } else {
```
kojiromike, line 138: `else` after a block that returns early is fine here, but the broader pattern (set internal error, then return $processingResult) reads more naturally without an `else` — early-return the error path and keep the success path at the outer scope.

src/Services/TherapyGroupsService.php
```diff
@@ -0,0 +1,175 @@
...
+
+        $sql = "DELETE FROM therapy_groups_participants WHERE ";
+        $sql .= "pid = ? AND group_id = ?";
+
+        $results = sqlStatement($sql, [
+            $data['pid'],
+            $data['group_id'],
+        ]);
+
+        if ($results) {
```
kojiromike, line 165: `$results` here is the statement handle returned by `sqlStatement()`, which is truthy for any successful query — including a DELETE that matched zero rows. So `removePatient` reports success even when the patient was never in the group (or the group doesn't exist), and the API returns 200 with `{id: <pid>}` for a no-op. Check affected rows instead — e.g., via `generic_sql_affected_rows()` after `sqlStatement`, or by using `QueryUtils` and checking its result. A failed precondition should be a 404.

src/Services/PatientService.php
```diff
@@ -372,6 +372,29 @@ public function getAll($search = array(), $isAndCondition = true, $puuidBind = n
         return $this->search($querySearch, $isAndCondition, $config);
     }
 
+    public function getAllForTherapyGroup(string $tgid)
```
kojiromike, line 375: Missing return type. PHPStan level 10 will flag this. Should be `: ProcessingResult` to match the other service methods in this file.

src/Services/PatientService.php
```diff
@@ -372,6 +372,29 @@ public function getAll($search = array(), $isAndCondition = true, $puuidBind = n
...
+        $processingResult = new ProcessingResult();
+
+        $sql = <<<SQL
+SELECT p.*
+FROM patient_data p
+INNER JOIN therapy_groups_participants tgp on tgp.pid = p.id
+WHERE tgp.group_id = ?
+SQL;
+
+        $results = sqlStatement($sql, [$tgid]);
```
kojiromike, line 388: `$tgid` is typed `string` but is being used as a group_id (integer FK). Either narrow to int at the boundary, or validate that it's numeric before passing to SQL. Also — should this method enforce any authorization check, or is it purely the caller's responsibility? Today the only caller is the `:tgid/participants` route which uses an admin ACL; if a different caller is added later it would bypass any group-level access control.

src/Validators/TherapyGroupsValidator.php
```diff
@@ -0,0 +1,23 @@
...
+class TherapyGroupsValidator extends BaseValidator
+{
+    protected function configureValidator()
+    {
+        parent::configureValidator();
+
+        $this->validator->context(
+            self::DATABASE_INSERT_CONTEXT,
+            function (Validator $context) {
+                $context->required('group_name')->lengthBetween(2, 255);
```
kojiromike, line 16: The validator requires `group_type`, `group_participation`, and `group_status` but doesn't constrain their values at all. Looking at the `therapy_groups` schema, these are enum-like columns (`group_type` is `support`/`closeended`/`openended`; `group_status` is `active`/`inactive`). Without `inArray()` constraints the API will accept any string and the database will either silently truncate or reject with a 500. Worth adding enum-style validators here.

Also: `group_start_date` has a fallback in the service but no format validation on the validator side — a malformed date will fail at the DB layer rather than as a clean 400.

kojiromike (changes requested):
Is there any interest in continuing this work? Adding API endpoints would be very helpful to OpenEMR. I went ahead and reviewed it as-is.

---

A few correctness and security issues need addressing before this can land. The biggest concerns are:

1. **Authorization is too broad / inconsistent.** The new `POST /api/therapy_groups`, `GET /api/therapy_groups`, and `GET /api/therapy_groups/:tgid/participants` routes gate on `authorization_check("admin", "users")` — i.e., admin-level user management — rather than a therapy-groups ACL. The patient-scoped routes use `authorization_check("patients", "med")` and the per-patient list uses `"therapy_groups", "med"`. These should be consistent and should not require admin to read or write a therapy group.

2. **The OpenAPI/PHPDoc block in `_rest_routes.inc.php` is malformed.** A literal `}` and a stray `"user/participants.read"` scope landed inside the OAuth-flow scopes JSON. This will break Swagger generation for the entire scope list.

3. **A new `user/participants.read` scope is being introduced but never used anywhere.** It's also misnamed (participants of what?). If dead, drop it; if needed, gate one of the routes on it.

4. **`DELETE /api/patient/:puuid/therapy_groups/:pgid` references an undefined `$data`** in the `apiLog($return, $data)` call. With strict mode this is an error; otherwise a warning. Either drop the second arg or capture the body.

5. **`removePatient` reports success based on the wrong signal.** `sqlStatement()` returns a statement handle that is truthy regardless of how many rows were deleted. A DELETE that matched zero rows still resolves to "success", silently masking "patient not in this group" or "group does not exist". Use affected-rows, not the handle.

6. **No tests.** The repo's CLAUDE.md requires PHPUnit coverage for new code; the entire CRUD surface here is untested. Service-level tests (against the DB suite) are the natural home — see existing `tests/Tests/Services/`.

7. **No `declare(strict_types=1)`** on any of the new files (CLAUDE.md requires it).

Inline comments cover the rest. The PR also currently shows merge conflicts against master — needs a rebase before re-review.

PR #7927, opened 2025-01-16
Bug Fixes, Log Improvements

adunsulag:
Log http requests and additional data.

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -2870,7 +2870,7 @@ $GLOBALS_METADATA = array(
         'audit_events_http-request' => array(
             xl('Audit Logging user page history.'),
             'bool',                           // data type
-            '0',                              // default = false for backwards compatibility
+            '1',                              // turn it on by default for better traceability in the logs
             xl('Enable logging of HTTP page requests as the user utilizes the web interface.') . ' (' . xl('Note that Audit Logging needs to be enabled above') . ')'
         ),
         'gbl_force_log_breakglass' => array(
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -2827,7 +2827,12 @@ $GLOBALS_METADATA = array(
             '1',                              // default
             xl('Enable logging of ordering activities.') . ' (' . xl('Note that Audit Logging needs to be enabled above') . ')'
         ),
-
+        'audit_events_lab-results' => array(
+            xl('Audit Logging Lab Results'),
+            'bool',                           // data type
+            '1',                              // default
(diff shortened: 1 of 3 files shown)


src/Common/Logging/EventAuditLogger.php
```diff
@@ -844,6 +844,55 @@ public function recordLogItem($success, $event, $user, $group, $comments, $patie
...
+        $this->newEvent(
+            "http-request-$event",  // event
+            $_SESSION['authUser'] ?? null, // user
+            $_SESSION['authProvider'] ?? null, // groupname
+            1, // success
+            $comment, // comments
+            $_SESSION['pid'] ?? null, // patient_id
+            'http-request', // log_from
+            null, // menu_item
+            0 // ccda_doc_id
```
bradymiller, line 892: rec dropping the last 3 parameters since these are just used by the portal logging.

library/globals.inc.php
```diff
@@ -2862,7 +2867,12 @@ function gblTimeZones()
             '0',                              // default
             xl('Enable logging of CDR Engine Queries.') . ' (' . xl('Note that Audit Logging needs to be enabled above') . ')'
         ),
-
+        'audit_events_http-request' => array(
+            xl('Audit Logging user page history.'),
+            'bool',                           // data type
+            '0',                              // default = false for backwards compatibility
```
bradymiller, line 2873: Any reason not to make this on by default. Otherwise it will never be used :)

bradymiller:
@adunsulag , Looks like a nice logging/auditing upgrade. Have couple notes above.

library/globals.inc.php
adunsulag: Wasn't sure if we wanted to add a ton of logs for older installations.  But I can make this default.  Because of the way the background services works it will add a lot of logs for those requests but I think that's preferable to not logging at all.  I actually had the thought to add in a unique request id also that would allow grouping entire page request histories into a single log trail.

bradymiller: That would be neat to group them by session. Check out SessionTracker where a session marker is stored in database and as a session variable (added the mechanism for a more robust way to track session expirations).

bradymiller: Realized that mine above would be a a login session and looks like yours would be by request grouping which would also be neat. They both could be helpful when auditing.

src/Common/Logging/EventAuditLogger.php
adunsulag: Fixed

library/globals.inc.php
adunsulag: Made default, will push request id / session id logging to a future update.

bradymiller (approved):

PR #7944, opened 2025-01-24
Tab Identifier Service

### LINKED ISSUE #7943: Encapsulating browser session
### Describe the problem

In a very busy office, staff will lose track of what browser tab the program is opened in and then open another. This will cause the new window to pull data from the previously opened patient into the new patient chart that has just been opened. Or conversely, users will try to have two patient's charts open in two different tabs in the same browser. This will cause cross-pollination of records. 

Update the system to be able to have multiple charts open  at the same time.

ChatGPT wrote:

To improve OpenEMR's session management so that multiple patient records can be opened simultaneously in different browser tabs without data cross-contamination, the following strategies can be implemented:

---

### 1. **Remove Patient Data from Session Variables**
Currently, patient-specific data stored in session variables is causing cross-contamination when switching between tabs. To resolve this:

- **Use a Tab-Specific State Storage Approach**: Instead of storing patient-specific data in PHP sessions, use client-side state management (e.g., `localStorage`, `sessionStorage`, or IndexedDB) to track the currently opened patient in each tab. This way, the patient data becomes tab-specific rather than session-specific.

---

### 2. **Implement a Token-Based Tab Identifier**
- Assign a unique token or identifier to each browser tab when a user opens a patient record.
- Include this token in every request (e.g., as a query parameter or part of a hidden form field).
- On the server side, associate the token with the corresponding patient data, ensuring that data fetched or modified is scoped to the tab's context.

---

### 3. **Switch to a Stateless Session Design**
A stateless design avoids storing patient-related data in the session. Here's how to do it:
- Store patient-specific data in the database or server cache instead of the session.
- Retrieve the patient context using a combination of tab identifiers and secure tokens passed with each request.
- This reduces the dependency on sessions for maintaining state, making the application more robust and scalable.

---

### 4. **Introduce a Scoped Session Context**
If session management cannot be entirely removed, enhance it to support scoped contexts:
- Use a hierarchical session structure where each tab has its own isolated session context under the main session.
- For example, the session could store multiple contexts as an array or object keyed by tab IDs:
  ```php
  $_SESSION['tabContexts'][$tabId] = [
      'currentPatientId' => 123,
      'otherContextData' => '...'
  ];
  ```

---

### 5. **Use Modern Frameworks for Client-Side State Management**
If OpenEMR is being modernized or extended, frameworks like React or Vue can help manage client-side state more effectively:
- Patient data and UI states can be tied to components or context managers, avoiding the need for server-side session storage for patient data.

---

### 6. **Secure URL-Based State Management**
For simpler implementations, include the patient ID as a query parameter in the URL:
- Example: `https://yourdomain.com/patient-record?patientId=123`
- The backend fetches data based on the `patientId` query parameter rather than relying on session data.

---

### 7. **Lock Mechanisms and Warnings for Edits**
To prevent accidental overwrites:
- Implement a locking mechanism to ensure multiple tabs don't overwrite patient data inadvertently.
- Warn users if two tabs try to edit the same patient record simultaneously.

---

### 8. **Testing and Feedback**
After implementing any of the above solutions:
- Test thoroughly with various workflows to ensure the session handling logic correctly isolates patient data per tab.
- Gather user feedback to refine the solution and ensure usability.

---

### Practical Next Steps:
- Start by identifying the specific patient-related session variables and migrating them to a more appropriate storage mechanism.
- Create a token-based tab identifier and update server-side logic to work with tokens.
- Incrementally test and implement these changes to avoid disruptions.

Since the code base is implementing Symfony components. Here is a detailed plan to move forward.

Here’s a **detailed implementation guide** to improve session management for OpenEMR using **Symfony components** to enable multiple patient records to be opened simultaneously in different tabs without data contamination.

---

juggernautsei:
Fixes #7943

Nothing. This is a foundation to the bigger task of having multiple browser tabs open and encapsulation patient data with in the browser session.

Adding the Symfony Session Interface to the Kernel

Yes

### DIFF
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -62,6 +62,19 @@ use OpenEMR\Services\PatientIssuesService;
 use OpenEMR\Services\PatientService;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 
+/** @var TabIdentifierService $tabIdentifierService */
+$tabIdentifierService = $GLOBALS['kernel']->getContainer();
+// Retrieve the session service
+/** @var SessionInterface $session */
+$browserTabSession = $tabIdentifierService->get(SessionInterface::class);
+
+// Retrieve the TabIdentifierService
+/** @var TabIdentifierService $tabIdentifierService */
+$tabIdentifierService = $tabIdentifierService->get(TabIdentifierService::class);
+
+// Use the services
+$tabToken = $tabIdentifierService->generateTabIdentifier();
+
 // Reset the previous name flag to allow normal operation.
 // This is set in new.php so we can prevent new previous name from being added i.e no pid available.
 OpenEMR\Common\Session\SessionUtil::setSession('disablePreviousNameAdds', 0);
--- a/src/Core/Kernel.php
+++ b/src/Core/Kernel.php
@@ -50,6 +50,13 @@ class Kernel
             $definition = new Definition(EventDispatcher::class, [new Reference('service_container')]);
             $definition->setPublic(true);
             $builder->setDefinition('event_dispatcher', $definition);
+            // Register the Session service
+            $builder->register(SessionInterface::class, Session::class)
+                ->setPublic(true);
+            // Register other services (e.g., TabIdentifierService)
+            $builder->register(\OpenEMR\Services\Tabs\TabIdentifierService::class)
+                ->addArgument(new Reference(SessionInterface::class))
+                ->setPublic(true);
             $builder->compile();
             $this->container = $builder;
         }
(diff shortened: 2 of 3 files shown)


juggernautsei:
@mdsupport 
@adunsulag 
@sjpadgett - engineer extraordinaire. 
I am trying to lay a foundation to build browser tab session where we can encapsulate the patient data and make it specific to a browser tab. This will stop cross-browser migration of patient data into a tab it does not below. That is the end goal of this start. I always admit. I don't know what I am doing. I enlisted the help of ChatGPT. It doesn't know much but it can provide some best practices. I want you all thoughts on the direction I am heading. The demographics.php is where this is going since the patient data starts there.

juggernautsei:
https://community.open-emr.org/t/multiple-windows-a-solution/5570/2
I found this article. Even though, I still have people reporting issues. I will see if I can get them to share with me which pages are having the issue.

bradymiller:
Hi @juggernautsei , There is a already mechanism to prevent session leaking when opening multiple tabs, but definitely not a long term solution (since setting httponly to be false makes it more difficult to prevent XSS attacks). If there is a bug where leaking is happening, basically means missing a restoreSession call somewhere. Here's a issue thread where the this is discussed in more detail:
https://github.com/openemr/openemr/issues/2556

juggernautsei:
That is what I shared with the people that reported the issue to me is that they need to share with me the page they were accessing so that I can add the restoreSession
Thanks.

PR #7979, opened 2025-02-05
Bug openemr fix #7276 snomed code lookup, fix #7976 snomed file format

### LINKED ISSUE #7276: bug: SNOMED lookup is slow in some places

### LINKED ISSUE #7976: feat: Support SNOMED-CT US Edition file format changed
### Describe the problem

Tried importing the US edition of snomed-ct and the file format appears to be changed.  We need to update the code import to support us edition.

The file I downloaded looks like this.  I double checked but can't see any other version to download to verify I wasn't accidentally downloading the wrong file. 
SnomedCT_ManagedServiceUS_PRODUCTION_US1000124_20240901T120000Z.zip

Looks like we need the following regex to support the import:
```
} elseif (preg_match("/SnomedCT_ManagedServiceUS_PRODUCTION_US[0-9]{7}_([0-9a-zA-Z]{8})T[0-9Z]{7}.zip/", $file, $matches)) {
```
In the interface/code_systems/list_staged.php file.

adunsulag:
Fixes #7976 
Fixes #7276 

Changed up the search mechanism for code types with external tables that
define a fulltext search on the search term.  Use the database search
features for improved lookup performance. Made this generic so it uses
all the same code lookup functionalities instead of specific to snomed.

Also made it so tables can define skipping the table count properties in
the find code files (find_code_dynamic and select_codes).

Made it so that ICD10 and other tables show a list of
codes when the finder first pops up, whereas snomed when it first shows
will show an empty table with a message that you need to put in a search term

I brought back the auto-complete typing though I bumped it up to 350ms
delay instead of 200ms delay.  This fixes the enter key closing the
dialog bug as well as makes the finder more usable.  With the improved
search performance the autocomplete now can function well.  Will have to
test this when there's a lot of concurrency.

This also fixes an issue with the current snomed us edition file name being changed.

**Original Issue**
> This changes up the snomed-ct code lookup to use its own specialized functions for searching that better leverage the snomed table indexes. 

#### Does your code include anything generated by an AI Engine? No (originally it did, now it doesn't)


adunsulag:
@bradymiller or @stephenwaite thoughts on this making it into 7.0.3? Or is it too big of a change?

bradymiller:
hi @adunsulag ,
This looks like it breaks searches across different code types:
From Admin->Coding->Codes
(select both ICD10 and SNOMED and then search for something)

Is there a way to optimize it with keeping of the prior framework (without requiring a new function)? Prior framework ensured generic enough to allow searching across different code types.

adunsulag:
I wasn't aware of the use case for searching across multiple code sets.  I'll see if we can make it more generic.

bradymiller:
@adunsulag ,
I also realized there is functionality built into these code type functions that allows edit/modification/assigning things like billing of the codes (the original code sets are not touched, but the modified items are stored in the codes table). So when searching for codes, it also searches for these in the codes table. I hope this makes sense.

adunsulag:
@bradymiller I've gone ahead and did a more generic solution.  Let me know what you think.

adunsulag:
Closing this PR, no idea what happened  and why I'm getting the 114 file changes.  What a mess.

PR #7991, opened 2025-02-07
Fixes #7907 load forms from a module

adunsulag:
This implements the loading of a form from a module using events.  This allows the module writer to customize the directory of an existing form or to implement their own forms.

It assumes that forms will be added to the registry via the module's table.sql.  It does NOT modify the forms administration piece.  That is still a work in progress.

I implemented also a FormLocator class for finding the paths for a form and caching the file.  Caching is based upon the formdir, page request (visit.php, report,php, new.php), and the current report page.

The report renderer got added everywhere that the LBF / standard forms are rendered.

Also added a safety mechanism to the ModulesApplication to make sure a module can only include files that are inside the module.

Added a new event LoadEncounterFormFilterEvent that gets fired whenever an encounter form is going to be loaded that module writers can hook into.

Fixes #7907

### DIFF
--- a/interface/patient_file/encounter/forms.php
+++ b/interface/patient_file/encounter/forms.php
@@ -33,8 +33,8 @@ use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Core\Header;
-use OpenEMR\Events\Encounter\EncounterMenuEvent;
 use OpenEMR\Events\Encounter\EncounterFormsListRenderEvent;
+use OpenEMR\Events\Encounter\EncounterMenuEvent;
 use OpenEMR\Common\Forms\FormLocator;
 use OpenEMR\Common\Forms\FormReportRenderer;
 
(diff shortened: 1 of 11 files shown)


interface/patient_file/encounter/forms.php
```diff
@@ -26,10 +34,13 @@
 use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Core\Header;
 use OpenEMR\Events\Encounter\EncounterMenuEvent;
+use OpenEMR\Events\Encounter\EncounterFormsListRenderEvent;
+use OpenEMR\Common\Forms\FormLocator;
+use OpenEMR\Common\Forms\FormReportRenderer;
```
bradymiller, line 39: keep use alphabetized here and in other scripts

bradymiller:
Didn't test, but code looks good!

interface/patient_file/encounter/forms.php
adunsulag: Fixed.

adunsulag:
Bringing in.  I think there's still an unrelated concern on the CI engine with php 8.1 but I can't reproduce it in the flex-3.17 docker on php 8.1 so going to bring this in and we'll look at the CI engine in a different issue.

PR #7992, opened 2025-02-07
WIP #7907 for form admin registration

adunsulag:
Reworked the form admin page to twigify it and to make it so we can have modules work inside the admin registration piece.  I'm not happy with how this has turned out though, going to leave it as a WIP though and may either close it or come back to it.

Its part of the work for #7907 but not sure it needs to be pulled in.  Just seems like its incomplete if a user can't change the ACO for a form that's registered by a module, or to rename it, or anything else.  Going to leave it here and mull on it, but if anyone wants to look it over and suggest improvements its here.


adunsulag:
This PR does depend on core template file that I will bring in on a different PR as I think that's an important piece to start consolidating our templates off a core base file and provide a page heading.

PR #8002, opened 2025-02-10
rel-703 Fixes #7907 load forms from a module (#7991)

### LINKED ISSUE #7905: Adding a form to the Encounter form from a module

### LINKED ISSUE #8007: bug: multiple instances of the same form trigger duplicate function fatal error
### Describe the bug

If you have multiple instances of the same form (Care Plan, Clinical Notes, etc), a fatal error is triggered due to the duplicate php functions being redefined.  This comes down to a use of include instead of include_once.

### To Reproduce

1. Add a CarePlan for an encounter
2. Add another CarePlan for an encounter
3. View the encounter summary display screen
4. See error in php logs and end of output on the screen.

### Expected behavior
All the forms should render.

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

adunsulag:
* Fixes #7905 load forms from a module

This implements the loading of a form from a module using events.  This allows the module writer to customize the directory of an existing form or to implement their own forms.

It assumes that forms will be added to the registry via the module's table.sql.  It does NOT modify the forms administration piece.  That is still a work in progress.

I implemented also a FormLocator class for finding the paths for a form and caching the file.  Caching is based upon the formdir, page request (visit.php, report,php, new.php), and the current report page.

The report renderer got added everywhere that the LBF / standard forms are rendered.

Also added a safety mechanism to the ModulesApplication to make sure a module can only include files that are inside the module.

Added a new event LoadEncounterFormFilterEvent that gets fired whenever an encounter form is going to be loaded that module writers can hook into.

Fixes #7905

* Reorder imports, fix view_form.php

view_form.php was not using the file locator, and was not consistent with the load_form and forms.php piece.

Fixed the import orders and tested to make sure everything loads properly again.

Fixes #8007 with duplicate function definitions.

### DIFF
--- a/interface/patient_file/encounter/forms.php
+++ b/interface/patient_file/encounter/forms.php
@@ -13,6 +13,14 @@
  */
 
 require_once(__DIR__ . "/../../globals.php");
+/**
+ * @global $srcdir
+ * @global $attendant_type
+ * @global $therapy_group
+ * @global $pid
+ * @global $userauthorized
+ * @global $rootdir
+ */
 require_once("$srcdir/encounter.inc.php");
 require_once("$srcdir/group.inc.php");
 require_once("$srcdir/patient.inc.php");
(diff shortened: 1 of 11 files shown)


adunsulag:
Brought in the duplicate fatal error fix #8007 with this PR as well, so probably good that I hadn't merged this in yet lol.

PR #8008, opened 2025-02-10
Fixes #8007 duplicate fatal error

### LINKED ISSUE #8007: bug: multiple instances of the same form trigger duplicate function fatal error

adunsulag:
Fixes #8007 by using include_once instead of just include.


adunsulag:
Not going to wait for CI as its broken in master.

PR #8009, opened 2025-02-11
Utility import_ccda Fixes

sjpadgett:
fix duplicate check to continue logic.
fix parse for current legal name if more than one. i.e. previous name. 
relax address to one line in duplicate check
show a dot per import for a run status

Fixes #

### DIFF
--- a/contrib/util/ccda_import/import_ccda.php
+++ b/contrib/util/ccda_import/import_ccda.php
@@ -86,6 +86,7 @@ function showHelp(): void
 
 function outputMessage($message): void
 {
+    echo("\n");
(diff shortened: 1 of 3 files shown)


contrib/util/ccda_import/import_ccda.php
```diff
@@ -187,6 +194,7 @@ function outputMessage($message): void
     } catch (Exception $e) {
         outputMessage("Error moving file: " . $e->getMessage() . "\n");
     }
+    echo('.');
```
bradymiller, line 197: 2DV

PR #8011, opened 2025-02-11
appointments report csv file header row does not match report headers

### LINKED ISSUE #8010: appointments report csv file header row does not match report headers
### Describe the bug
for an appointments report the csv file header row does not match report headers, instead it uses db fields
 

Image

Image

Image

### To Reproduce
reports/visits/appointments
select appropriate dates etc
click 'Submit' to generate a report
click 'Export to csv' to generate a csv file. 
the header row in the csv file uses the db field names instead of the header strings in the report itself. 

db field names would not mean much to many users
 

### Client configuration

- Browser: firefox
- OpenEMR version: 7.0.3-dev
- Operating system: ubuntu

ruthkonyn:
Fixes #8010

row 1 of the csv file, which contains column headings, now matches the headings used in the report itself

a new array is generated with the csv data and with row 1, the headers, matching the report (rather than db field names)

no

### DIFF
--- a/interface/reports/appointments_report.php
+++ b/interface/reports/appointments_report.php
@@ -464,8 +464,19 @@ if (!empty($_POST['form_refresh']) || !empty($_POST['form_orderby'])) {
     $appointments = sortAppointments($appointments, $form_orderby);
     if (!empty($_POST['form_csvexport'])) {
         // include provider as well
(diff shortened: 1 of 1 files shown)


bradymiller (approved):
Code looks great (didn't test it).

bradymiller:

PR #8033, opened 2025-02-14
Bug openemr #7276 snomed lookup fixes

### LINKED ISSUE #7276: bug: SNOMED lookup is slow in some places

adunsulag:
Fixes #7276 

Changed up the search mechanism for code types with external tables that
define a fulltext search on the search term.  Use the database search
features for improved lookup performance. Made this generic so it uses
all the same code lookup functionalities instead of specific to snomed.

Also made it so tables can define skipping the table count properties in
the find code files (find_code_dynamic and select_codes).

Made it so that ICD10 and other tables show a list of
codes when the finder first pops up, whereas snomed when it first shows
will show an empty table with a message that you need to put in a search term

I brought back the auto-complete typing though I bumped it up to 350ms
delay instead of 200ms delay.  This fixes the enter key closing the
dialog bug as well as makes the finder more usable.  With the improved
search performance the autocomplete now can function well.  Will have to
test this when there's a lot of concurrency.


adunsulag:
@bradymiller this should address the generic nature of the lookup.  It handles multiple code sets and is very fast on text searching now as well for code lookups.

custom/code_types.inc.php
```diff
@@ -122,8 +129,17 @@ function define_external_table(&$results, $index, $table_name, $col_code, $col_d
...
                            EXT_JOINS => $joins,
                            EXT_VERSION_ORDER => $version_order,
-                           DISPLAY_DESCRIPTION => $display_desc
+                           DISPLAY_DESCRIPTION => $display_desc,
+                           CODE_COLUMN_TYPE => CODE_COLUMN_TYPE_STRING,
+                           // tables with hundreds of thousands or millions of rows have abysmal performance
+                            // when doing a SELECT count(*) of tablename
+                           // because innodb has to do a full table scan to get the count due to row level transaction isolation
+                           // so we give the option to skip the total table count for these tables
+                           SKIP_TOTAL_TABLE_COUNT => false
```
stephenwaite, line 138: bigfoot

interface/patient_file/encounter/find_code_dynamic.php
```diff
@@ -87,6 +88,16 @@
...
                 <?php } ?>
             },
+            fnDrawCallback: function(obj) {
+                const infoBox = $(".dataTables_wrapper p.moreResults");
+                if (infoBox.length) {
+                    infoBox.parent().remove();
+                }
+                if (obj.json && obj.json.iTotalHasMoreRecords) {
+                    $(".dataTables_wrapper .dataTables_info").parent().parent()
+                        .prepend("<div class='col-sm-12'><p class='alert alert-warning moreResults'>" + <?php echo xlj("Maximum displayable results reached.  Narrow your search"); ?> + "</p></div>");
```
bradymiller, line 98: Avoid double spacing in translation constants.

interface/patient_file/encounter/find_code_dynamic.php
```diff
@@ -107,6 +118,39 @@
...
+                }
+            },
+            fnDrawCallback: function(obj) {
+                const infoBox = $(".dataTables_wrapper p.moreResults");
+                if (infoBox.length) {
+                    infoBox.parent().remove();
+                }
+                if (obj.json && obj.json.iTotalHasMoreRecords) {
+                    $(".dataTables_wrapper .dataTables_info").parent().parent()
+                        .prepend("<div class='col-sm-12'><p class='alert alert-warning moreResults'>" + <?php echo xlj("Maximum displayable results reached.  Narrow your search"); ?> + "</p></div>");
```
bradymiller, line 138: Avoid double spacing in translation constants.

interface/patient_file/encounter/find_code_dynamic_ajax.php
```diff
@@ -294,8 +296,21 @@ function genFieldIdString($row)
...
 } elseif ($what == 'codes') {
+    // for an external table with tons of codes for performance reasons we need to force them to search for something
+    $externalTableId = $code_types[$codetype]['external'] ?? 0;
+    /**
+     * @global $code_external_tables
+     */
+    $stopEmptySearch = $externalTableId && $code_external_tables[$externalTableId][SKIP_TOTAL_TABLE_COUNT] ?? false;
+    if (empty(trim($searchTerm)) && $stopEmptySearch) {
+        $out['iSearchEmptyError'] = xl('Search term is required for this code type.');
+        echo json_encode($out);
```
bradymiller, line 307: if this is going to html (pretty sure it is), then need to surround the json_encode() with a text()

interface/patient_file/encounter/select_codes.php
```diff
@@ -82,9 +83,28 @@
...
+                    }
+                },
+                fnDrawCallback: function(obj) {
+                    const infoBox = $(".dataTables_wrapper p.moreResults");
+                    if (infoBox.length) {
+                        infoBox.parent().remove();
+                    }
+                    if (obj.json && obj.json.iTotalHasMoreRecords) {
+                        $(".dataTables_wrapper .dataTables_info").parent().parent()
+                            .prepend("<div class='col-sm-12'><p class='alert alert-warning moreResults'>" + <?php echo xlj("Maximum displayable results reached.  Narrow your search"); ?> + "</p></div>");
```
bradymiller, line 102: Avoid double spacing in translation constants.

bradymiller:
Code looks great (didn't test)! Just noted some minor things (think ok to bring in after address these minor things).

bradymiller:
dance-bear

interface/patient_file/encounter/find_code_dynamic_ajax.php
adunsulag: It is pulled via AJAX by the DataTables  jquery library so it doesn't get directly injected into the page. Look at select_codes.php and you can see that its using xlj for all the translation screens.  If there's an escaping problem fundamentally in the DataTables library let me know and we can address it.

interface/patient_file/encounter/find_code_dynamic.php
adunsulag: Fixed.

interface/patient_file/encounter/find_code_dynamic.php
adunsulag: Fixed.

interface/patient_file/encounter/select_codes.php
adunsulag: Fixed.

PR #8052, opened 2025-02-23
Refactor Vitals form save

### LINKED ISSUE #7857: bug: Vitals Form weight / height when validating data blanks screen after confirming validation prompt
### Description
On the vitals form if you attempt to put in an invalid weight or height value it will display an error alert about the invalid data.  When you click Ok, it submits the page and shows a blank screen.

### Screenshots
Enter in an invalid Weight 
image

See the validation error
image

Clicking OK still submits the form!!
image

### Steps to reproduce the behavior:
1. Open an encounter
2. Click on Clinical -> Add Vitals Form (or edit an existing Vitals form)
3. Weight input field
4. Enter **asdfasfas** into the field
5. Scroll down and hit the save button
6. Click OK on the validation form
7. See blank screen

### Expected behavior
The form submission should be cancelled.  Preferrably it would be good to scroll the viewport window to the source of the validation error, but not required.

### Client configuration

- Browser: Chrome Version 126.0.6478.126 (Official Build) (64-bit)
- OpenEMR version: 7.0.2 patch 3 (latest master release)
- Operating system: Easy Dev docker environment on Ubuntu 20.04

### LINKED ISSUE #8051: Vitals save new form
### Describe the bug
After batch ccda imports creating a new vital for fails with:
```
[21-Feb-2025 18:20:51 America/New_York] PHP Fatal error:  Uncaught OpenEMR\Common\Database\SqlQueryException: Failed to execute statement. Error: Duplicate entry '1055' for key 'PRIMARY' Statement: INSERT INTO form_vitals SET `id` = ?,`uuid` = ?,`date` = ?,`pid` = ?,`user` = ?,`groupname` = ?,`activity` = ?,`bps` = ?,`bpd` = ?,`height` = ?,`weight` = ?,`temperature` = ?,`temp_method` = ?,`pulse` = ?,`respiration` = ?,`note` = ?,`BMI` = ?,`BMI_status` = ?,`waist_circ` = ?,`head_circ` = ?,`oxygen_saturation` = ?,`oxygen_flow_rate` = ?,`ped_weight_height` = ?,`ped_bmi` = ?,`ped_head_circ` = ?,`inhaled_oxygen_concentration` = ? in C:\xampp\htdocs\openemr\library\sql.inc.php:164
```
Initial debugging shows that the generate_id() for the forms form_id is not returning the max id value in form_vital table.

The save method in this form is contrary to normal encounter form save i.e. created the new vital form reference in forms and then save to the form_vital table. Should be opposite in case the form_vital save fails. Plus the form_save insert would return the max Pimary id for he forms insert..

### To Reproduce

Difficult because requires large import of ccdas.
### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

sjpadgett:
refactor saveVitalsArray to create new form_vitals first to get new id then create the forms record. Removes the reliance on sequence value. fix doc block entry for event trait.

Fixes #8051 
Fixes #7857 
On new vitals form save and for the sake of pure logic create the form_vital form first to get the record id from table to use in the new encounter forms record.

refactor saveVitalsArray to create new form_vitals first to get new id then create the forms record. Removes the reliance on sequence value. fix doc block entry for event trait.

### DIFF
--- a/interface/forms/vitals/templates/vitals/vitals.html.twig
+++ b/interface/forms/vitals/templates/vitals/vitals.html.twig
@@ -17,7 +17,7 @@
         'weight_input_usa': window.xl("Weight") + ' (' + window.xl("lbs") + ')'
         ,'weight_input_metric': window.xl("Weight") + ' (' + window.xl("kg") + ')'
         ,'height_input_usa': window.xl("Height/Length") + ' (' + window.xl("in") + ')'
-        ,'height_input_usa_metric': window.xl("Height/Length") + ' (' + window.xl("cm") + ')'
+        ,'height_input_metric': window.xl("Height/Length") + ' (' + window.xl("cm") + ')'
         ,'bps_input': window.xl("BP Systolic")
         ,'bpd_input': window.xl("BP Diastolic")
         ,'validateFailed': window.xl("Please correct the value(s) before proceeding!")
--- a/interface/forms/vitals/templates/vitals/vitals.html.twig
+++ b/interface/forms/vitals/templates/vitals/vitals.html.twig
@@ -14,10 +14,10 @@
 <script src="{{ FORM_ACTION|attr }}/interface/forms/vitals/vitals.js?v={{ assetVersion|attr_url }}" type="text/javascript"></script>
 <script>
     let vitalsTranslations = {
-        'weight_input': window.xl("Weight") + ' (' + window.xl("lbs") + ')'
+        'weight_input_usa': window.xl("Weight") + ' (' + window.xl("lbs") + ')'
         ,'weight_input_metric': window.xl("Weight") + ' (' + window.xl("kg") + ')'
-        ,'height_input': window.xl("Height/Length") + ' (' + window.xl("in") + ')'
-        ,'height_input_metric': window.xl("Height/Length") + ' (' + window.xl("cm") + ')'
+        ,'height_input_usa': window.xl("Height/Length") + ' (' + window.xl("in") + ')'
+        ,'height_input_usa_metric': window.xl("Height/Length") + ' (' + window.xl("cm") + ')'
         ,'bps_input': window.xl("BP Systolic")
         ,'bpd_input': window.xl("BP Diastolic")
         ,'validateFailed': window.xl("Please correct the value(s) before proceeding!")
@@ -192,7 +192,7 @@ $(function () {
 
 function ShowGrowthchart(doPDF) {
     // get values from the current form elements
-    vitals[0] = formdate+'-'+$("#height_input").val()+'-'+$("#weight_input").val()+'-'+$("#head_circ_input").val();
+    vitals[0] = formdate+'-'+$("#height_input_usa").val()+'-'+$("#weight_input_usa").val()+'-'+$("#head_circ_input_usa").val();
     // build the data string
     var datastring = "";
     for(let i=0; i<vitals.length; i++) {
--- a/src/Services/VitalsService.php
+++ b/src/Services/VitalsService.php
@@ -356,7 +356,6 @@ class VitalsService extends BaseService
             }
             $isInsert = true;
             $sqlOperation = "INSERT INTO ";
-            // use our generate_id function here for us to create a new id
             $vitalsData['id'] = null;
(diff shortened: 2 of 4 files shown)


src/Services/VitalsService.php
```diff
@@ -345,51 +347,70 @@ public function saveVitalsArray(array $vitalsData)
...
 
         // now go through and update all of our vital details
-
         $updatedDetails = [];
         foreach ($details as $column => $detail) {
             $detail['form_id'] = $vitalsData['id'];
-            $updatedDetails[$column] = $this->saveVitalDetails($detail);
+            // $updatedDetails[$column] = $this->saveVitalDetails($detail); // TODO: sjp IMO this is a bug, didn't make sense to me
+            $this->saveVitalDetails($detail); // save the detail record
+            $updatedDetails[$column] = $detail; // update the details array with the saved record
```
sjpadgett, line 409: @adunsulag can you take a look at my todo comment for the related code. saveVitalDetails returns empty record set.

adunsulag: I'm surprised we haven't had issues in the past with this.  Since we are returning the updated details array we probably should change the saveVitalsDetails function instead of ignoring its return.  What you've done doesn't handle the insert case where the details array will be missing the id field with the newly generated form_vital_details.id column.

Here's the updated function I'd recommend doing.  Note the changes are the last 4 lines where we populate the $vitalDetails['id'];
```
    private function saveVitalDetails(array $vitalDetails)
    {
        $id = $vitalDetails['id'] ?? null;
        $sqlOperation = "UPDATE ";
        if (empty($id)) {
            $sqlOperation = "INSERT INTO ";
        }

        unset($vitalDetails['id']);
        $keys = array_keys($vitalDetails);
        $values = array_values($vitalDetails);
        $fields = array_map(function ($val) {
            return '`' . $val . '` = ?';
        }, $keys);
        $sqlSet = implode(",", $fields);

        $sql = $sqlOperation . FormVitalDetails::TABLE_NAME . " SET " . $sqlSet;
        if (!empty($id)) {
            $sql .= " WHERE `id` = ? ";
            $values[] = $id;
            QueryUtils::sqlStatementThrowException($sql, $values);
        } else {
            $vitalDetails['id'] = QueryUtils::sqlInsert($sql, $values);
        }
        return $vitalDetails;
    }
```

sjpadgett: testing well thanks @adunsulag wondered how this wasn't causing problems but could be most users don't use details.

adunsulag (approved):
Looks good Jerry. There's the minor issue that @ocv-robert mentioned on the comment, but in terms of code everything looks fine.  We probably should have bundled everything inside a transaction, but we'd have to look at everywhere the service is used to see if it'd create a nested transaction so I don't think we address that in this PR.

sjpadgett:
Yep decided against transaction because saves will never happen if insert fails. I fixed validation on save issue and am working on fixing invalid id names in form calculations. Will put up those changes later today.

interface/forms/vitals/templates/vitals/vitals.html.twig
```diff
@@ -14,10 +14,10 @@
 <script src="{{ FORM_ACTION|attr }}/interface/forms/vitals/vitals.js?v={{ assetVersion|attr_url }}" type="text/javascript"></script>
 <script>
     let vitalsTranslations = {
-        'weight_input': window.xl("Weight") + ' (' + window.xl("lbs") + ')'
+        'weight_input_usa': window.xl("Weight") + ' (' + window.xl("lbs") + ')'
         ,'weight_input_metric': window.xl("Weight") + ' (' + window.xl("kg") + ')'
-        ,'height_input': window.xl("Height/Length") + ' (' + window.xl("in") + ')'
-        ,'height_input_metric': window.xl("Height/Length") + ' (' + window.xl("cm") + ')'
+        ,'height_input_usa': window.xl("Height/Length") + ' (' + window.xl("in") + ')'
+        ,'height_input_usa_metric': window.xl("Height/Length") + ' (' + window.xl("cm") + ')'
```
bradymiller, line 20: This has both usa and metric in the name?

sjpadgett: thanks

bradymiller:
2DV (1)

PR #8062, opened 2025-02-26
Improve appointment report

### LINKED ISSUE #8061: appointments report: display patients address; select multiple providers
### Describe the problem

It would be useful to be able to display the address of the patient. 
It would be useful to be able to select multiple providers rather than only one or all

have an option to display patients' address, and if ticked display them in the report

in the provider drop down menu be able to select more than one individual provider, i.e. html 'multiple', and have all the selected providers' appointments displayed in the report

ruthkonyn:
Fixes #8061

In the report->visits->appointments

optionally display the patients' address in the report

select more than one provider, or all and display all their appointments
have a checkbox which when ticked displays patients' address in the report

in the provider drop down menu, be able to select more than one provider, or all providers, and display all their appointments
no

### DIFF
--- a/library/appointments.inc.php
+++ b/library/appointments.inc.php
@@ -137,16 +137,16 @@ function fetchEvents($from_date, $to_date, $where_param = null, $orderby_param =
         }
 
         // RM add values before more custom conditions are added to the search string
-     if ($bind_param) {
-      $sqlBindArray = array_merge($sqlBindArray, $bind_param);
-   }
+        if ($bind_param) {
+            $sqlBindArray = array_merge($sqlBindArray, $bind_param);
+        }
 
         // Filter out appointments based on a custom module filter
         $apptFilterEvent = new AppointmentsFilterEvent(new BoundFilter());
         $apptFilterEvent = $GLOBALS["kernel"]->getEventDispatcher()->dispatch($apptFilterEvent, AppointmentsFilterEvent::EVENT_HANDLE, 10);
         $boundFilter = $apptFilterEvent->getBoundFilter();
-       $sqlBindArray = array_merge($sqlBindArray, $boundFilter->getBoundValues());
-       $where .= " AND " . $boundFilter->getFilterClause();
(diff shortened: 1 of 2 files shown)


library/appointments.inc.php
```diff
@@ -371,9 +376,17 @@ function fetchAppointments($from_date, $to_date, $patient_id = null, $provider_i
 
     $where = "";
 
+ //RM multiple providers
     if ($provider_id) {
-        $where .= " AND e.pc_aid = ?";
-        array_push($sqlBindArray, $provider_id);
+        $quantity = sizeof($provider_id) ;
```
adunsulag, line 381: Going to step through the code but this breaks the patient flow board report when searching on a specific provider.  Not sure why but will investigate.

adunsulag: Looks like it fails when $provider_id is a string when the function expects an array. 
Following code fixes it:
```
    if (!empty($provider_id)) {
        // provider_id can be a string or an array
        if (is_array($provider_id)) {
            $quantity = sizeof($provider_id);
            $where .= " AND ( e.pc_aid = ?";
            for ($i = 1; $i < $quantity; $i++) {
                $where .= " OR e.pc_aid = ? ";
            }
            $where .= ")";
            foreach ($provider_id as $x) {
                array_push($sqlBindArray, $x);
            }
        } else {
            $where .= " AND e.pc_aid = ?";
            array_push($sqlBindArray, $provider_id);
        }
    }
```

ruthkonyn: oh dear, thanks, i thought i'd tested that !!

adunsulag: You gave me access to push so I went ahead and fixed it.

adunsulag: Everything else checks out good.  Once the CI engine passes I can bring it in, unless you want to review my changes.

ruthkonyn: many thanks, no i trust you completely!! please do bring it in. I'm just cooking but will read the solution more carefully later.

bradymiller:

PR #8066, opened 2025-02-26
Fix Dicom Viewer 404 image load

### LINKED ISSUE #8065: Dicom view image load fails
### Describe the bug
Path build is missing web root when server domain isn't aliased where web_root being empty would work.
We see this often. Problem is $web_root wasn't being passed in to viewer twig template

### To Reproduce

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

sjpadgett:
twig template missing web_root var init

Fixes #8065 

image loading

### DIFF
--- a/library/dicom_frame.php
+++ b/library/dicom_frame.php
@@ -44,6 +44,7 @@ if ($web_path) {
 $twig = (new TwigContainer(null, $GLOBALS['kernel']))->getTwig();
 echo $twig->render("dicom/dicom-viewer.html.twig", [
     'assets_static_relative' => $GLOBALS['assets_static_relative']
+    ,'web_root' => $web_root
     ,'web_path' => $web_path
     ,'state_url' => $state_url ?? null
     ,'docid' => $docid ?? null


library/dicom_frame.php
```diff
@@ -44,6 +44,7 @@
 $twig = (new TwigContainer(null, $GLOBALS['kernel']))->getTwig();
 echo $twig->render("dicom/dicom-viewer.html.twig", [
     'assets_static_relative' => $GLOBALS['assets_static_relative']
+    ,'web_root' => $web_root
```
adunsulag, line 47: I was scratching my head thinking what broke, but apparently the bug was that the twig file expects **web_root** when the global twig value defined here is **webroot** https://github.com/openemr/openemr/blob/3a1343e3861caf760d199e8385f17f26419d5d13/src/Common/Twig/TwigExtension.php#L72

Ouch, my bad.

adunsulag: So this solves the problem but I should have used webroot when I did the refactor.  Keep forgetting to check the webroot extension when doing twig refactors.

sjpadgett: If this is all from a refactor to twig i'd say you got off pretty good! :)

HARI-14052002:
Thanks for the fix brother.It worked perfectly.

PR #8070, opened 2025-03-02
FHIR and API's auth keys verification

### LINKED ISSUE #8069: oauth2 keys generation fix
### Describe the problem
Currently the general oauth2 keys don't get created until the first client registration or API access.
Now if the user has an issue an the creation of any one of the keys or oaut2passprase get corrupted or deleted then any other client trying to auth a session will fail because the oauth keys are global to all clients.
Also there is a much better chance keys create will fail if user is on windows rather than linux.

Have openemr check and auto create keys on a system login. If necessary do a clean up and regen or simply create if missing.
In this way we have control and saves users having to figure out how to fix and clean up a failed create.

sjpadgett:
- Add methods to support keys verification and creations

Fixes #8069

validate api key creation

- Add methods to support keys verification and creations

### DIFF
--- a/src/Common/Auth/OAuth2KeyConfig.php
+++ b/src/Common/Auth/OAuth2KeyConfig.php
@@ -2,12 +2,13 @@
 
 /**
  * Oauth2KeyConfig is responsible for configuring, generating, and returning oauth2 keys that are used by the OpenEMR system.
- * @package openemr
+ *
+ * @package   openemr
  * @link      http://www.open-emr.org
  * @author    Jerry Padgett <sjpadgett@gmail.com>
  * @author    Brady Miller <brady.g.miller@gmail.com>
  * @author    Stephen Nielson <stephen@nielson.org>
- * @copyright Copyright (c) 2020 Jerry Padgett <sjpadgett@gmail.com>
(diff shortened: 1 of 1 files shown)


src/Common/Auth/OAuth2KeyConfig.php
```diff
@@ -124,8 +137,14 @@ public function configKeyPairs(): void
                 "encrypt_key" => true,
                 "encrypt_key_cipher" => OPENSSL_CIPHER_AES_256_CBC
             ];
-            $keys = \openssl_pkey_new($keysConfig);
+            $msg = getenv('OPENSSL_CONF');
```
bradymiller, line 140: @sjpadgett , Just checking on this line above since don't see where this variable is used (except for being written over below).
Also thinking of doing a bit of work on the configKeysPairs function (remove the pieces that create things things are doing this nicely in the new function you created; then can avoid having rather complicated code in 2 different places). Just making sure you are ok with it. Will promise not to break anything, just going to remove unused code :)

sjpadgett:
That’s fine, Brady. I was going to spend some more time in this code later, but make whatever changes you need.

PR #8071, opened 2025-03-03
fix: broken real world testing report 2025 if certain module not installed

bradymiller:
fix broken real world testing report 2025 if certain module not installed


bradymiller:
will also bring into rel-703

PR #8076, opened 2025-03-03
Fix relative paths in forms (default and contrib)

### LINKED ISSUE #8075: bug: reports with relative globals.php include do not render in encounter summary screen.
Looks like the refactor of the encounter summary screen breaks any form that has a relative include of the globals.php ie:
```
```

If the form file is included inside another php class file like the FormReportRenderer it breaks if the require_once is relative.  We need to fix all the relative link forms so that they render in the encounter summary page.

adunsulag:
Changed up the relative path setting in the forms in order to fix the FormReportRenderer so modules can load forms.

Also fixed the form generator scripts (both xml and perl).

Discovered as part of this that because we are loading the form file in the context of a function it does not have access to global values such as $srcdir, $webroot, and $rootdir.  Added those variables into the FormReportRenderer function so that they exist in the context of the form.

If the form is relying on any other global php variables they could have issues in this new format, from my testing it appears the forms are working fine but we'll need to keep an eye on it.

Fixes #8075


bradymiller (approved):

bradymiller:
dance-bear

bradymiller:
hi @adunsulag , I am staging this for 7.0.3.4 patch. This looks like was a potentially critical bug fix. I am not seeing it yet in rel-703, so getting into patch. Just making sure this was not already addressed in rel-703 via another mechanism/patch.

PR #8078, opened 2025-03-05
feat: update Oauth2KeyConfig for more logging and removing duplicated code

bradymiller:
Update Oauth2KeyConfig for more logging and removing duplicated code

Prelim testing ok. Still more testing to do.

Here is what the log shows with some of the testing (so can track what is happening to the oauth2 stuff; ideally should only be created once and then set, but if need to autocreate-them for some reason then can follow what is going on in the log to problem solve the issue (since recreating oauth2 stuff will break all api tokens, will be good to know can problem solve from log when this is happening)).
image

### DIFF
--- a/src/Common/Auth/OAuth2KeyConfig.php
+++ b/src/Common/Auth/OAuth2KeyConfig.php
@@ -41,7 +41,10 @@ class OAuth2KeyConfig
      * @var string File location of the oauth2 public key
      */
     private $publicKey;
+
     private CryptoGen $cryptoGen;
+
+    private OAuth2KeyMissing $oauth2KeyMissing;
 
     public function __construct($siteDir = null)
     {
@@ -57,10 +60,10 @@ class OAuth2KeyConfig
         $this->publicKey = $siteDir . '/documents/certificates/oapublic.key';
 
         // Create a Oauth2KeyMissing object that will track missing elements and used in below verifyKeys method and createOrRecreateKeys method
-        $oauth2KeyMissing = new OAuth2KeyMissing();
-        if ($this->verifyKeys($oauth2KeyMissing) === false) {
+        $this->oauth2KeyMissing = new OAuth2KeyMissing();
+        if ($this->verifyKeys() === false) {
             try {
-                $this->createOrRecreateKeys($oauth2KeyMissing);
+                $this->createOrRecreateKeys();
             } catch (OAuth2KeyException $e) {
                 // if unable to recreate keys, then force exit
                 throw new OAuth2KeyException("Unable to create/recreate oauth2 keys" . $e->getMessage());
@@ -103,13 +106,7 @@ class OAuth2KeyConfig
             $this->oaEncryptionKey = $this->cryptoGen->decryptStandard($eKey['value']);
             if (empty($this->oaEncryptionKey)) {
                 // if decrypted key is empty, then critical error and must log and exit
-                if (($_ENV['OPENEMR__ENVIRONMENT'] ?? '') === 'dev') {
-                    // Special case for developers. Delete all the oauth2 keys, which will force the keys to be recreated at next attempt.
-                    // TODO: see if this mechanism is still required
-                    $devModeLog = " (in development mode, so will delete all oauth2 keys, and should all be recreated at next attempt)";
-                    $this->deleteKeys();
-                }
-                EventAuditLogger::instance()->newEvent("oauth2", ($_SESSION['authUser'] ?? ''), ($_SESSION['authProvider'] ?? ''), 0, "oauth2 ConfigKeyPairs: oauth2 encryption key was blank after it was decrypted" . ($devModeLog ?? ''));
+                EventAuditLogger::instance()->newEvent("oauth2", ($_SESSION['authUser'] ?? ''), ($_SESSION['authProvider'] ?? ''), 0, "oauth2 ConfigKeyPairs: oauth2 encryption key was blank after it was decrypted");
                 throw new OAuth2KeyException("oauth2 encryption key was blank after it was decrypted");
             }
         } else {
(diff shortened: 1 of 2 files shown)


src/Common/Auth/OAuth2KeyConfig.php
```diff
@@ -86,116 +90,82 @@ public function getPrivateKeyLocation()
...
-            // collect the encryption key from database
+        if (!empty($eKey['name']) && ($eKey['name'] == 'oauth2key')) {
             $this->oaEncryptionKey = $this->cryptoGen->decryptStandard($eKey['value']);
             if (empty($this->oaEncryptionKey)) {
+                // if decrypted key is empty, then critical error and must log and exit
                 if (($_ENV['OPENEMR__ENVIRONMENT'] ?? '') === 'dev') {
-                    // delete corrupted key if doesn't exist to regenerate on next attempt.
-                    sqlStatementNoLog("DELETE FROM `keys` WHERE `name` = 'oauth2key'");
+                    // Special case for developers. Delete all the oauth2 keys, which will force the keys to be recreated at next attempt.
+                    // TODO: see if this mechanism is still required
```
bradymiller, line 109: @sjpadgett , Just checking is still need this mechanism. No prob with keeping it and tested that it works, but figured might as well remove it if no longer need it.

bradymiller:
kind of neat. I asked co-pilot to review this and got following reply:

### PR Review Summary

#### Files Changed
1. `src/Common/Auth/OAuth2KeyConfig.php`
   - Added new namespace import: `use OpenEMR\Common\Logging\EventAuditLogger;`
   - Added new class `OAuth2KeyMissing` to track missing OAuth2 key elements.
   - Modified methods to use `OAuth2KeyMissing` for verifying and creating/recreating keys.
   - Added logging using `EventAuditLogger` for various key-related activities.

2. `src/Common/Auth/OAuth2KeyMissing.php`
   - New file created.
   - Class `OAuth2KeyMissing` added to track missing OAuth2 keys (`encryptionKey`, `passphrase`, `publicKey`, `privateKey`).

#### Changes Summary
- **Enhancements:**
  - Improved logging and error handling with `EventAuditLogger`.
  - Introduced `OAuth2KeyMissing` to handle missing key pieces more granularly.
  - Enhanced key verification and creation logic by consolidating responsibilities.

- **Code Quality:**
  - Adheres to coding standards and best practices.
  - Improved readability and maintainability by introducing `OAuth2KeyMissing`.

- **Potential Issues:**
  - Ensure that `EventAuditLogger` and `OAuth2KeyMissing` are thoroughly tested.
  - Verify that the new logging and key handling mechanisms do not introduce performance overhead.

- **Test Coverage:**
  - Ensure test cases cover the new logging and key handling logic.
  - Validate scenarios where keys are missing and need to be recreated.

#### Recommendations
- **Approval:** The changes look good, improve the robustness of the key handling, and add necessary logging. Ensure thorough testing before merging.
- **Request Changes:** None required at this moment based on the provided diff.

src/Common/Auth/OAuth2KeyConfig.php
sjpadgett: Yep you can remove. In most cases this was effective but for some reason I had issue with passphrase or just had states of keys that weren't resolved. May be a window issue but I decided to just trim and test from instantiation.
I think from time this method was first create we added to it as we found issues. The entire handling of keys works very well. I just got aggravated trying to register a new client via endpoint and kept running into issues with keys. Some came from window environment with the biggest issue that befuddled me was the oauthpassphrase was either not created or was deleted. I never did resolve why as the code looked good.

I was going to add a new class called oauth2KeyManagement but decided that it is overkill for basically a class that has a bunch of conditionals that easily can be handled in one or two new methods in original class. The class concern is the same. For logging the user only needs to know that the keys were regenerated in case they get questions from users if suddenly their client keys were failing. The only thing that the user may have to deal with is a re-login and reauthenticate. Otherwise correctly done any affected keys should be regenerated and authorized. This is one issue I wanted to test but I thought time was short so I just implemented a simple solution for key management.

You are not off base with what you're doing but I have found in many instances in programming scripting languages the simpler the better! Originally I was going to make this change in keys management available to any class wanting to verify keys e.g. openemr login in main or in api auth. In this case, a new class is appropriate since we are no longer a single concern.
I don't understand, besides cleaning up deprecated code, what is wrong with what I was doing. Simple and straight forward. However, in the end Brady, I'm fine with whatever you want to do and I understand you enjoy and keep an eye on any security concerns. Maybe if going the route of a new class you can make keys verify and recreate available to be instantiated and called anywhere.

I'm still testing patient/standard/portal API and fear we have some gotchas that bite occasionally.

bradymiller: hi @sjpadgett ,
Main use case that drove my updates (except for the removal of unused code) was if there was a consistent fail (for example, the authpassphrase not being created) or if lost temporary access to the drive (then it would look like the keys do not exist on drive) or if have permission issues of drive after upgrade or transitioning to another server, then the api tokens would break (since the oauth keys/certs would change). So I figured I'd keep a log trail for troubleshooting on this (like is done for the login/logout stuff) to be able to discern these issues. Creating the log trail then led to the new class (to pass the tracking of missing stuff along with some related worker function). I'm not a real fan of the class pass by reference that i am doing with the new class and will likely instead use that new class as a property in the main class.

sjpadgett: @bradymiller I don't have any problems with you improving the concept. I probably should have spent the time to implement my original thought but feared I was out of time.
So again, I appreciated you spending the time on this.

sjpadgett (approved):
Hi Brady. I won't be able to test until later tonight however; if you want to merge tonight I have a PR I'm working on if I find issue in my windows testing. Thanks for cleaning this up.

src/Common/Auth/OAuth2KeyConfig.php
sjpadgett: started testing...

bradymiller:
@sjpadgett , did more extensive testing on this and testing well. will bring it in :)

PR #8114, opened 2025-03-11
Fix dsi window launch w/ unescaped action #8103

### LINKED ISSUE #8103: System API Client issues
### Describe the bug
- Under API Client clicking External CDR returns an internal 404 status. Cause for this item is escaping issue for
 `let url = webroot + '/interface/smart/admin-client.php?action=external-cdr/cdr-info&serviceId=' + encodeURIComponent(dsi) + "&csrf_token=" + encodeURIComponent(csrfToken);` at the query location `?action=external-cdr/cdr-info` with treating query param as a path. Forward slash in query on windows is a no no.

So far haven't been able to escape!

### To Reproduce

1. Add a '....'

### Expected behavior
return url w/o 404
### Client configuration

- Browser: Edge, FF etc.
- OpenEMR version:7.0.3 - 7.0.4 dev
- Operating system: windows

adunsulag:
This fixes #8103 on windows and other browsers that don't like a forward slash in a query parameter. Was missing the encoding of the action.

Fixes #8103

### DIFF
--- a/library/js/utility.js
+++ b/library/js/utility.js
@@ -577,7 +577,8 @@ if (typeof top.userDebug !== 'undefined' && (top.userDebug === '1' || top.userDe
                 };
                 window.top.addEventListener('message', windowMessageHandler);
 
-                let url = webroot + '/interface/smart/admin-client.php?action=external-cdr/cdr-info&serviceId=' + encodeURIComponent(dsi)
+                let url = webroot + '/interface/smart/admin-client.php?action=' + encodeURIComponent("external-cdr/cdr-info")
+                    + '&serviceId=' + encodeURIComponent(dsi)
                     + "&csrf_token=" + encodeURIComponent(csrfToken);
                 let title = node.dataset.smartName || JSON.stringify(xl("Smart App"));
                 // we allow external dialog's  here because that is what a SMART app is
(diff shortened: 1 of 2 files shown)


adunsulag:
@sjpadgett I believe this will fix your issue.  Tested locally appears to work, can you give it a try and let me know?

library/js/utility.js
```diff
@@ -577,7 +577,8 @@ if (typeof top.userDebug !== 'undefined' && (top.userDebug === '1' || top.userDe
                 };
                 window.top.addEventListener('message', windowMessageHandler);
 
-                let url = webroot + '/interface/smart/admin-client.php?action=external-cdr/cdr-info&serviceId=' + encodeURIComponent(dsi)
+                let url = webroot + '/interface/smart/admin-client.php?action=' + encodeURIComponent("external-cdr/cdr-info")
+                    + '&serviceId=' + encodeURIComponent(dsi)
                     + "&csrf_token=" + encodeURIComponent(csrfToken);
                 let title = node.dataset.smartName || JSON.stringify(xl("Smart App"));
                 // we allow external dialog's  here because that is what a SMART app is
```
sjpadgett, line 584: This isn't where the anchor for button is located. This is url for a dialog.
I believe the anchor for button is incomplete. In ClientAdminController.php L-651

```
<h2>
<?php echo text($title); ?>
  <a class="btn btn-secondary btn-sm float-right" href="<?php echo attr($this->getActionUrl([self::TOKEN_TOOLS_ACTION])); ?>" onclick="top.restoreSession()"><?php echo xlt("Token Tools"); ?></a>
  <a class="btn btn-secondary btn-sm float-right mr-2" href="<?php echo $GLOBALS['webroot']; ?>/interface/smart/register-app.php" onclick="top.restoreSession()"><?php echo xlt("Register New App"); ?></a>
  <a class="btn btn-secondary btn-sm float-right mr-2" href="<?php echo attr($this->getActionUrl([RouteController::EXTERNAL_CDR_ACTION])); ?>" onclick="top.restoreSession()"><?php echo xlt("External CDR"); ?></a>
</h2>
```

adunsulag: Looks like the button was a leftover from my first initial implementation of the external cdr as it used to go to a list of all of the cdr client apps.  However, I pushed the CDR attribute onto the list page for all of the api clients so this button doesn't do anything.  So I removed the button.

PR #8118, opened 2025-03-11
Fixes #8057 : Refactor $group_disabled logic by separating PHP and JavaScript

### LINKED ISSUE #8057: Refactor add_edit_event.php to Separate JavaScript from PHP for Better Readability
In interface/main/calendar/add_edit_event.php, JavaScript is currently embedded within PHP, reducing readability and making debugging more difficult. Separating JavaScript into an external file or placing it inside a <script> block in the HTML section would improve maintainability and adhere to best practices for separation of concerns.

**Steps to Reproduce**
Open add_edit_event.php.
Observe inline JavaScript within PHP code (e.g., line 156).
Notice that JavaScript logic is mixed with PHP, which makes it harder to maintain

**Expected Behavior**
JavaScript should be moved to a separate .js file or placed inside a <script> tag in the HTML section.
Improves maintainability and debugging.

**Suggested Approach**
Option 1: Move JavaScript to an External JS File and include it properly
Option 2: Place JavaScript inside a <script> tag in the HTML section

**Additional Context**
This aligns with best practices for clean code and separation of concerns.
Would improve maintainability in future development.

hannah-kim117:
Fixes #8057 

Refactors the $group_disabled logic by moving JavaScript functionality out of PHP and into an external JS file (add_edit_event.js).
The goal is to confirm if this is the right approach before refactoring additional logic in the file.

**Changes**:
- Replaced inline JS in PHP (lines 154-157) with phpData object containing groupDisabled and validationRules
- Moved logic to add_edit_event.js for better separation of concerns
- phpData is globally available so external JS can access it

**Before (Inline JS in PHP):**
`$group_disabled = ($_GET['group'] && !$g_edit && $have_group_global_enabled ) ? ' disabled=true; ' : '';
if ($group_disabled) {
    echo '<script>$(function () {$("input").prop("disabled", true);$("select").prop("disabled", true);}) </script>';
}`

**After (PHP and JS separated):**
`$group_disabled = ($_GET['group'] && !$g_edit && $have_group_global_enabled);`
`<script type="text/javascript">
    var phpData = {
        groupDisabled: <?php echo $group_disabled ? 'true' : 'false'; ?>,
        validationRules: <?php echo json_encode($collectthis); ?>
    };
</script>`

add_edit_event.js
`document.addEventListener('DOMContentLoaded', function () {
    if (phpData.groupDisabled === 'true') {
        document.querySelectorAll("input, select").forEach(function (element) {
            element.disabled = true;
        });
    }
});
`


interface/main/calendar/add_edit_event.php
```diff
@@ -151,10 +151,17 @@
...
 
-$group_disabled = ($_GET['group'] && !$g_edit && $have_group_global_enabled ) ? ' disabled=true; ' : '';
-if ($group_disabled) {
-    echo '<script>$(function () {$("input").prop("disabled", true);$("select").prop("disabled", true);}) </script>';
-}
+
+$group_disabled = ($_GET['group'] && !$g_edit && $have_group_global_enabled);
+?>
+<script type="text/javascript">
+    var phpData = {
```
adunsulag, line 158: This is ok, but it'd be nicer if you called a function defined in add_edit_event.js so there's an explicit dependency without people having to hunt for it.  It follows the principle of element of least surprise.

adunsulag: ie something like
```
if (window.AddEditEventController.init) {
AddEditEventController.init({
        groupDisabled: <?php echo $group_disabled ? 'true' : 'false'; ?>,
        validationRules: <?php echo json_encode($collectthis); ?>
});
``` 

Then you could define AddEditEventController in your add_edit_event.js
```
window.AddEditEventController = {
  init: function(phpData) { /* do code here */ }
};
```

Or you could do it as IIFE (Immediately Invoked Function Expression) to prevent data leakage
```
(function(window) { 
function init(phpData) { /* do code here */ }

window.AddEditEventController = init;
})(window);
```

kojiromike:
Closing this as stale — no activity since March 2025 and the review feedback from @adunsulag was never addressed.

For anyone picking this up in the future: the approach should follow the controller pattern adunsulag suggested (similar to what was done in #8000 for the newpatient form), rather than using a global `phpData` variable. See the review comments on this PR for details.

The underlying issue #8057 remains valid as a refactoring task.

PR #8129, opened 2025-03-13
Fixes api sorting, pagination problems #8127,#8128

### LINKED ISSUE #8127: bug: api _sort always defaults to descending
### Describe the bug

When using the api _sort parameter it always defaults the sort to descending order due to an order of operations evaluation bug.

### To Reproduce
Use an api query like the following /apis/default/api/patient?_sort=date with multiple patients created and see the sort order is descending instead of ascending like the query expects.

### Expected behavior
Query results return in sort order specified.

### Client configuration
- Client: Insomnia
- OpenEMR version: 7.0.2+

### LINKED ISSUE #8128: bug: pagination _limit does not actually limit the results so results come back with +1 records
### Describe the bug

We handle records using a query limit of _limit + 1 in order to check if the query has more results.  However, the pagination in ProcessingResult has a bug where it allows unlimited number of records to be added.  We should fix this bug.  Also since FHIR now has a Trial Use _maxresults property instead of using _limit like some experimental FHIR clients do, we should support that property as well.

### To Reproduce

1. Create multiple patients
2. Use curl, insomnia, postman, etc to make http request to /apis/default/fhir/Patient?_sort=-date&_limit=1
3. See 2 patients instead of 1

### Expected behavior
Only see 1 patient.

### Client configuration

- OpenEMR version: OpenEMR 7.0.2+

adunsulag:
Removed _REWRITE_COMMAND from pagination links as its extraneous information that's not needed.

Fixed bug where FHIR resources were not properly handling the search config parameters.

Fixed a bug where pagination was not actually limiting the results to the specified _limit parameter.  Also added support for FHIR TU _maxresults parameter in addition to supporting experimental clients that use _limit.  Fixes #8128.

Fixed bug where sort fields were always set to be descending (ie search field date treated as search field -date).  Fixes #8127.

Many of these bugs were introduced in the original paging/sorting PR in  #6801

### DIFF
--- a/src/RestControllers/RestControllerHelper.php
+++ b/src/RestControllers/RestControllerHelper.php
@@ -151,6 +151,10 @@ class RestControllerHelper
                 // cleanup _limit and _offset
                 unset($getParams['_limit']);
                 unset($getParams['_offset']);
+                // cleanup a mod_rewrite piece so the URL is nicer.
+                if (isset($getParams['_REWRITE_COMMAND'])) {
+                    unset($getParams['_REWRITE_COMMAND']);
+                }
                 $queryParams = http_build_query($getParams);
 
                 $pagination->setSearchUri($bundleUrl . '?' . $queryParams);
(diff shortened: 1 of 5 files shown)


adunsulag:
If CI passes going to bring this in so I can do a round of inferno testing on rel-703.  Want to make sure we are fully passed there.

Issue #1986, opened 2018-11-20
CSP Policy and Config


danehrlich1:
Content Security Policy is something the web server adds to HTTP responses and requests. It's like an embedded firewall in every packet sent to and from your server. Already have this working in the `openemr-devops` repo but need to do further testing.

- Google recommendations for CSP: https://developers.google.com/web/fundamentals/security/csp/
- Build CSP programatically instead of in `httpd.conf`: https://github.com/paragonie/csp-builder/tree/v1.1.0
- Automated CSP builder: https://report-uri.com/home/generate

danehrlich1:
Currently this is what I've got after playing around with: https://report-uri.com/home/generate

`Content-Security-Policy: default-src 'self' data: 'unsafe-inline' ; media-src 'none' ; object-src 'none' ; worker-src 'none' ; upgrade-insecure-requests; block-all-mixed-content; reflected-xss filter; referrer no-referrer-when-downgrade;`

Will need to do testing next.

odra:
I am looking for an issue to contribute to - is someone working on this?

bradymiller:
hi @odra , This looks like a really neat project and anything that boosts security is a good thing (it's not something I knew much about until I just read this: https://www.projectseven.net/php-content-security-policy.php ). Let me know if your interested, and then I'll assign it to you.

odra:
Cool, I would be happy to take a look at it :-)

wcushen:
Hi @odra, happy to take on if not being worked on presently.

odra:
Sure, @stephenwaite could you please assign the issue to @wcushen?

bradymiller:
hi @wcushen , Just assigned the issue to you. thanks!

wcushen:
Thanks Brady and Leo!

Will Cushen

CONSULTANT, CUSTOMER SOLUTIONS GROUP

Red Hat Asia-Pacific Southern Region <https://www.redhat.com/>

wcushen@redhat.com    M: +61400691166 <+61-400-691-166>

[image: AnsibleFest 2020 is now a virtual experience]

PR #8164, opened 2025-03-20
feat: adds standard http security headers

msummers42:
Fixes Issue #1986

Adds additional content security policy header directives and additional security headers

Adds HTTP headers

No.

### DIFF
--- a/portal/index.php
+++ b/portal/index.php
@@ -17,8 +17,11 @@
  */
 
 // prevent UI redressing
-Header("X-Frame-Options: DENY");
-Header("Content-Security-Policy: frame-ancestors 'none'");
+Header("X-Frame-Options: DENY"); // Or SAMEORIGIN
+Header("Content-Security-Policy: frame-ancestors 'none'; default-src 'self'; img-src 'self' data:; script-src 'self'; style-src 'self'");
+Header("X-XSS-Protection: 1; mode=block");
+Header("X-Content-Type-Options: nosniff");
+Header("Referrer-Policy: strict-origin-when-cross-origin");
 
 //setting the session & other config options
 


portal/index.php
```diff
@@ -17,8 +17,11 @@
  */
 
 // prevent UI redressing
-Header("X-Frame-Options: DENY");
-Header("Content-Security-Policy: frame-ancestors 'none'");
+Header("X-Frame-Options: DENY"); // Or SAMEORIGIN
+Header("Content-Security-Policy: frame-ancestors 'none'; default-src 'self'; img-src 'self' data:; script-src 'self'; style-src 'self'");
+Header("X-XSS-Protection: 1; mode=block");
+Header("X-Content-Type-Options: nosniff");
+Header("Referrer-Policy: strict-origin-when-cross-origin");
```
bradymiller, line 24: neat stuff. looks like line 22 is deprecated and covered by line 21.

portal/index.php
```diff
@@ -17,8 +17,11 @@
  */
 
 // prevent UI redressing
-Header("X-Frame-Options: DENY");
-Header("Content-Security-Policy: frame-ancestors 'none'");
+Header("X-Frame-Options: DENY"); // Or SAMEORIGIN
+Header("Content-Security-Policy: frame-ancestors 'none'; default-src 'self'; img-src 'self' data:; script-src 'self'; style-src 'self'");
```
CloCkWeRX, line 21: This one is probably very correct, but is it worth allowing an ENV variable or similar to configure some of these options?
Possibly also a [Content-Security-Policy-Report-Only](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy-Report-Only) ?


That would let people opt in gradually if they are doing something non standard with their deploy.

kojiromike:
Is this also handled by the nascent front controller? @Firehed

PR #8171, opened 2025-03-24
adding minimum height for all appts under 15 minutes for all categories

### LINKED ISSUE #8146: bug: Calendar: user created appointment is invisible due to duration of zero minutes
### Describe the bug

A user created an appointment.  The appointment is not visible to the user in the calendar, thus not clickable, due to a duration of zero minutes.  (The duration of zero minutes happened because of a custom category, but was not the user's intention.)

### To Reproduce

1. Create a new appointment.
2. Set the duration to zero minutes.
3. Save the appointment.
4. Observe that the appointment is not visible.

To clarify, using my account, I see a height of 2px, likely due to a border.  (I have not managed to click the 2px-high appointment.)  The user who encountered the bug has customized the appearance of the calendar and is not seeing anything, even the 2px border.

### Expected behavior

That the appointment would have some minimum height, thus be clickable.

### Client configuration

- Browser: Firefox 136
- OpenEMR version: 7.0.2 patch 3
- Client operating system: Windows 11
- Server operating system: Debian GNU/Linux 12.9

I see this code in ajax_template.php:
https://github.com/openemr/openemr/blob/ce76aef96a474699c315c4fb60ff5dce80bac180/interface/main/calendar/modules/PostCalendar/pntemplates/default/views/day/ajax_template.html#L661-L665

I wonder if we can remove the condition, so that every category of event has a minimum height.

TennisGod:
https://github.com/openemr/openemr/issues/8146

Fixes #8146 

Appointments with a duration of 0 minute used to be unclickable, now they have a minimum height and are clickable. 
Appointments that are under 15 minutes will now have a minimum height. 
No

### DIFF
--- a/interface/main/calendar/modules/PostCalendar/pntemplates/default/views/day/ajax_template.html
+++ b/interface/main/calendar/modules/PostCalendar/pntemplates/default/views/day/ajax_template.html
@@ -658,10 +658,10 @@ foreach ($providers as $provider) {
             $eEndInterval = $eMinDiff / $interval;
             // times the interval height
             $eHeight = $eEndInterval * $timeslotHeightVal;
-            if($event['catid']==3)
+            // For all categories: ensure minimum height for zero-duration events
+            if($eEndInterval <= 15 / $interval)
             {
-                // An "OUT" that is "zero duration" still needs height so we can click it.
-                $eHeight = $eEndInterval==0 ? $timeslotHeightVal : $eHeight ;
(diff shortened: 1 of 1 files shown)


adunsulag:
Brought this in locally and tested it.  Fixes the click issue.   Long term we'd want validation to prevent the 0 duration on events, but this is a good first step.

stephenwaite:
seal-dunk

PR #8174, opened 2025-03-24
Fixes #8173 date format encounter bug

### LINKED ISSUE #8173: bug: Encounter Date Of Service saves as empty date when server 'Date Display Format' is not YYYY-MM-DD
### Describe the bug

If your 'Date Display Format' is not YYYY-MM-DD the encounter date of service gets saved as an empty date '0000-00-0000' because the server can't parse the date format.

### To Reproduce
1. Go to Admin -> Config -> Locale and set the **Date Display Format** to MM/DD/YYYY.
2. Create or edit an existing encounter
3. Set the date of service field value.  Note the date is displayed in **YYYY-MM-DD** format for the field and not the MM/DD/YYYY format as expected.
4. Hit save

See the date saved as an invalid empty date.

Image

### Client configuration
Date displays and saves in the correct format.

- Browser: Chrome
- OpenEMR version: 7.0.3
- Operating system: Ubuntu 22.04 LTS

adunsulag:
Fixes #8173.

Formatting library was missing the time_display_format in the jsGlobals so time formatting was messed up.

Also had a wierd bug where sub-frames that loaded the i18formatting library would end up stomping on the global values of i18formatting but when the script was unloaded the browser appears to go and remove the references to those values so the functions would end up undefined.

I ended up defining the formatting library in the main entry points (main.php and in the portal/home.php) so the functions always exist.  I then removed the library loads from any of the sub-frames to avoid the behavior.  billing_tracker.php, insurance_edit, and the encounter stuff appears to be the only things impacted.

@stephenwaite I don't have any data setup for the x12 auto sftp claim tracking pieces that would impact billing_tracker.php.  Any chance when you have some free time to pull this down and verify I haven't broken anything.  Best way to check is to turn the date local format to 'dd/mm/yyyy' and just make sure the date selector formats properly.

### DIFF
--- a/interface/billing/billing_tracker.php
+++ b/interface/billing/billing_tracker.php
@@ -36,7 +36,7 @@ if (!AclMain::aclCheckCore('acct', 'eob', '', 'write') && !AclMain::aclCheckCore
 ?>
 <html>
 <head>
-    <?php Header::setupHeader(['datatables', 'datatables-colreorder', 'datatables-dt', 'datatables-bs', 'i18formatting']); ?>
+    <?php Header::setupHeader(['datatables', 'datatables-colreorder', 'datatables-dt', 'datatables-bs']); ?>
     <title><?php echo xlt("Claim File Tracker"); ?></title>
     <style>
         table.dataTable td.details-control:before {
@@ -104,8 +104,8 @@ if (!AclMain::aclCheckCore('acct', 'eob', '', 'write') && !AclMain::aclCheckCore
                             return data;
                         }
                     },
-                    { 
-                        "data": "created_at", 
+                    {
+                        "data": "created_at",
                         "render": function(data, type, row, meta) {
                             // Build the URL so the user can download the claim batch file
                             if (type === 'display') {
@@ -115,7 +115,7 @@ if (!AclMain::aclCheckCore('acct', 'eob', '', 'write') && !AclMain::aclCheckCore
                             return data;
                         }
                     },
-                    { 
+                    {
                         "data": "updated_at",
                         "render": function(data, type, row, meta) {
                             // Build the URL so the user can download the claim batch file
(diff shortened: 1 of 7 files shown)


interface/billing/billing_tracker.php
```diff
@@ -104,8 +104,8 @@
                             return data;
                         }
                     },
-                    { 
-                        "data": "created_at", 
+                    {
```
adunsulag, line 107: php cbf flagged these lines, I didn't really modify them but it auto-formatted them.

library/js/xl/formatting.js
```diff
@@ -76,10 +76,11 @@
         const format = typeof date_display_format !== 'undefined' ? date_display_format : 0;
 
         let formatted;
+        // uses php-date-formatter library which in turn uses the php datetime format @see https://www.php.net/manual/en/datetime.format.php
         if (format === 1) {
-            formatted = seconds ? "h:mm:ss a" : "h:mm a";
-        } else { // (format === 0)
-            formatted = seconds ? "HH:mm:ss" : "HH:mm";
```
sjpadgett, line 82: I wonder if I did this? I had to fix jsGlobal out of scope I think for portal appointments using globals date format

adunsulag: I didn't check the git blame on it as it looked like something I did lol

sjpadgett: ah ha! It was you:) kidding I don't know...

adunsulag: yea, git blame says it was me on the format.

sjpadgett: my goodness, I truly thought it was me! doesn't matter and I'll check portal for change tomorrow unless need tonight

stephenwaite (approved):
clapping_ape

bradymiller:
dance-bear

bradymiller (approved):

PR #778, opened 2017-05-21
Allow custom fonts and font sizes

robertdown:
This allows users to select their font-family and font-size in the preferences. Couple of things left to do. First, we use a LOT of inline CSS that have to come out for this to be completely effective. Ironically, the globals page suffers from this and it won't change (Well, the tabs will, content won't).

I know we've had CSS values in the globals before and I understand _why_ we took them out, but these two values (at minimum) should definitely be a per-user option (I think @sjpadgett mentioned it on a separate thread recently).

I don't have everything worked out yet (Though you can totally test this and it'll work). For instance, right now there's `themeBuilder.php` and `themeBuilder.css`. The CSS acts as a template for the php file which loops over possible custom options and does a `str_replace`, and then inside of (eventually core.css) style_light.css it `@imports url(themeBuilder.php);` which sets the content-type to text/css. There's probably a more elegant solution, but wanted to get this on the radar.

You'll see a font called "Lato" which is an open source font that I'm going to embed so we've got access to it (It's a great font and will be the Light themes new default here soon.

PR #8176, opened 2025-03-26
update dependencies

bradymiller:
update dependencies


bradymiller:
Does anybody know if the following is being used anywhere in OpenEMR:
https://github.com/openemr/openemr/blob/master/public/fonts/lato/lato.css
https://github.com/openemr/openemr/tree/master/public/fonts/lato

It is causing a lint error in ci and the easiest thing to do would be to remove it if it is not even being used :)

bradymiller:
disregard my above lato.css comment. I found where it is being used :)

bradymiller:
actually, i still don't get it. Although am seeing Lato font-family including on listings, I am not seeing anywhere where lato.css is not brought in. So it appears it is not being used?????

bradymiller:
Think i have it figured out. It was added with old themebuilder stuff here: https://github.com/openemr/openemr/pull/778

Then themebuilder stuff was removed and no longer used the lato font at that point.

So can remove this since it is not being used.

bradymiller:
testing well and bringing in

PR #8183, opened 2025-03-26
Disabling duration for in office events

### LINKED ISSUE #8182: feat: Duration Field for In-Office Events Obsolete
### Describe the bug

The duration field for in-office events are useless since they get completely ignored. 

1. Click the '+' sign on the calendar
2. Click on the 'Provider' tab
3. Change the duration
4. Notice the duration does not get reflected as In Office events will always last for the whole day

### Expected behavior
Disabling the duration field for In-Office events should be sufficient. 

### Client configuration

- Browser: any
- OpenEMR version: any
- Operating system: any

bryanopenemr:
https://github.com/openemr/openemr/issues/8182

Fixes #8182

The duration field for in-office events are useless since they get completely ignored.

The duration field is now disabled when we're adding/editing an In-Office event. 

No


interface/main/calendar/add_edit_event.php
```diff
@@ -1091,27 +1096,37 @@ function set_category() {
...
+            }
+        } else {
+            durationDisabled = false;
+        }
     }
     document.getElementById('tdallday1').style.color = color1;
     document.getElementById('tdallday2').style.color = color2;
     //document.getElementById('tdallday3').style.color = color2;
-    document.getElementById('tdallday4').style.color = color2;
+    // document.getElementById('tdallday4').style.color = color2;
```
adunsulag, line 1122: If the code is not used, remove it instead of commenting it out.

interface/main/calendar/add_edit_event.php
```diff
@@ -1091,27 +1096,37 @@ function set_category() {
 // radio buttons are clicked.
 function set_allday() {
     var f = document.forms[0];
+    var s = f.form_category;
     var color1 = 'var(--gray)';
     var color2 = 'var(--gray)';
-    var disabled2 = true;
+    var timeDisabled = true;
```
adunsulag, line 1102: using **var** is consistent with the older code, in the future I'd recommend changing all of these local variables to use **let** or **const** instead of **var** if you're going to heavily modify a function. I had to pull down the code and make sure there wasn't a ```var disabled``` already in the file since the variable was called **disabled2**.  Using **let** would have avoided that problem as it would have kept the variable local to the function.

interface/main/calendar/add_edit_event.php
```diff
@@ -1091,27 +1096,37 @@ function set_category() {
...
     if (document.getElementById('rballday1').checked) {
         color1 = '';
     }
     if (document.getElementById('rballday2').checked) {
         color2 = '';
-        disabled2 = false;
+        timeDisabled = false;
+        if (s.selectedIndex >= 0) {
+            var catid = s.options[s.selectedIndex].value;
+            if (catid != '2') {
```
adunsulag, line 1112: Let's pull out the magic constant of '2' into a constant variable.  I don't like magic numbers and I know its scattered throughout the code but for new code we should try to mitigate it.  Either define '2' as a constant inside the function, or even better define it in php near the top of the file and use that value.

interface/main/calendar/add_edit_event.php
```diff
@@ -1832,6 +1847,8 @@ function isRegularRepeat($repeat)
     $("#form_apptstatus").addClass('form-control-sm');
     $("#form_room").addClass('form-control-sm');
     $(".current a").addClass('active');
+
+    set_display();
```
adunsulag, line 1851: set_display is called in line 1792 but is conditioned if there is an existing encounter.  Does this line here break if there is no_encounter?  Does it make sense to move the functionality in lines 1792-1796 into this area to avoid redundant code?

adunsulag (changes requested):
Tested the code, it looks fine.  Couple minor items to look at just for general code hygiene and future maintenance.

interface/main/calendar/add_edit_event.php
bryanopenemr: Yep this makes sense, moved the lines from 1792-1796 to replace set_display()

bryanopenemr:
I have created a new PR with the changes in this PR and https://github.com/openemr/openemr/pull/8186 to avoid merge conflicts, please see below:
https://github.com/openemr/openemr/pull/8263

adunsulag:
Closing since new PR was opened with these changes.

PR #8186, opened 2025-03-27
adding validation to time and duration for add edit event

### LINKED ISSUE #8185: bug: No Validation for Hour, Minute, and Duration Fields in Add Event
### Describe the bug
While there is validation for some other fields, 

1. Click the + on the calendar
2. Put in random stuff for time and duration (even strings too)
3. Click save and notice there's validation on patient name
4. Put in a patient and click save
5. Notice there's no validation on time and duration
### Expected behavior
There should be the same validation on time and duration as patient. 

### Client configuration

- Browser: any
- OpenEMR version: any
- Operating system: any

bryanopenemr:
https://github.com/openemr/openemr/issues/8185

Fixes #8185

No validation for time and duration fields on add edit event

Added validation to time and duration fields on add edit event

No

### DIFF
--- a/interface/main/calendar/add_edit_event.php
+++ b/interface/main/calendar/add_edit_event.php
@@ -1065,11 +1065,16 @@ function set_display() {
             style_prefcat.display = '';
             f.form_apptstatus.style.display = 'none';
             f.form_prefcat.style.display = '';
+            f.form_duration.disabled = true;
+            f.form_duration.value = '';
+            document.getElementById('tdallday4').style.color = 'var(--gray)';
         } else {
             style_prefcat.display = 'none';
             style_apptstatus.display = '';
             f.form_prefcat.style.display = 'none';
             f.form_apptstatus.style.display = '';
+            f.form_duration.disabled = false;
+            document.getElementById('tdallday4').style.color = '';
         }
     }
 }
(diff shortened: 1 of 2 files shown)


interface/main/calendar/add_edit_event.php
```diff
@@ -1851,6 +1868,44 @@ function are_days_checked(){
 * */
 var collectvalidation = <?php echo $collectthis; ?>;
 function validateform(event,valu){
+    collectvalidation.form_hour = {
+        numericality: {
+            onlyInteger: true,
+            greaterThanOrEqualTo: 0,
+            lessThanOrEqualTo: 23,
+            message: "must be a valid hour (0-23)"
```
bradymiller, line 1876: hi @bryanopenemr ,
Just checking how these messages are used to figure out best way to translate/escape them. Are they literals that end up in a alert popup or something else?

bryanopenemr: Hi @bradymiller ,
The current behavior is the text box gets outlined red and the message will show up next to the box. (Though I think it should be changed in the future because it's ugly)

(collectvalidation.custom_messages = true;) - Adding this line will make the custom message show up in place of the default message if you'd like to try

interface/main/calendar/add_edit_event.php
```diff
@@ -1091,27 +1096,37 @@ function set_category() {
...
-    var disabled2 = true;
+    var timeDisabled = true;
+    var durationDisabled = true;
     if (document.getElementById('rballday1').checked) {
         color1 = '';
     }
     if (document.getElementById('rballday2').checked) {
         color2 = '';
-        disabled2 = false;
+        timeDisabled = false;
```
adunsulag, line 1109: These changes are going to conflict with your other PR #8183 .  We'll need to bring in one or the other and you'll have to merge the two.  Do you have a preference of which one you want brought in?

interface/main/calendar/add_edit_event.php
adunsulag: @bradymiller Line 1962 uses the submitme function defined in validation_script.js.php @see https://github.com/openemr/openemr/blob/789a132bde64ffd60ddd1e966051640fc0b8d6c2/library/validation/validation_script.js.php#L48

Which then calls the appendError method on L109, which eventually calls the jQuery text method on line 169.  I believe we are safe here since jQuery's text method uses the textContents dom node internally to sanitize the text for the DOM node its injecting into.  

Specific language translations are defined in lines 227-256. 

@bryanopenemr You can see on this line: https://github.com/openemr/openemr/blob/789a132bde64ffd60ddd1e966051640fc0b8d6c2/library/validation/validation_script.js.php#L164

You'll want to use the window.top.xl() method everywhere you have a message here, to make sure you are getting language translations. Either that or add your default translations to the getErrorMessage constants in lines 227-256 that I mentioned above.

adunsulag (changes requested):
Need to address the conflicts and language translations.  @bradymiller may have additional requirements on the message sanitation, but I think that's addressed.

adunsulag:
Just tested the code, not sure we can take it as is, without changing up how the error messages get displayed.  This a problem of both the validation library and the form. Invalid errors make the screen unusable, unless you manually drag / expand the event box.
image

interface/main/calendar/add_edit_event.php
bryanopenemr: I think we should go with the changes in PR #8183. Is there anything I need to change in this PR in regards to this?

interface/main/calendar/add_edit_event.php
bryanopenemr: Fixed by adding default translations to getErrorMessage.

bryanopenemr:
> Just tested the code, not sure we can take it as is, without changing up how the error messages get displayed. This a problem of both the validation library and the form. Invalid errors make the screen unusable, unless you manually drag / expand the event box. image

I changed validation so that it now shows the error message when user hovers over the text fields. This applies to all forms that use this validation and I tested the create patient form and encounter questionnaire form, both work well.

bryanopenemr:
I have created a new PR with the changes in this PR and https://github.com/openemr/openemr/pull/8183 to avoid merge conflicts, please see below:
https://github.com/openemr/openemr/pull/8263

adunsulag:
Closing since new PR was opened for this work.

PR #8189, opened 2025-03-27
Mws/upstr iss 8135 - patient scopes employerservice endpoint

### LINKED ISSUE #8135: bug: api missing employer.read scope for /api/patient/{puuid}/employer endpoint
### Describe the bug

So the /api/patient/{puuid}/employer endpoint was added but no associating scope for {context}/employer.read has been added.  So patients and user contexts cannot make a call to the endpoint.  I'm guessing the internal api must not check the scope so this got overlooked when we refactored the insurance data stuff.  

We need to either merge employer data into the patient endpoints (since it seems a basic demographic property) or add this scope.

### To Reproduce
Attempt to request the user/employer.read scope in a registration request and get an invalid request response.

- OpenEMR version: OpenEMR 7.0.2+

msummers42:
Fixes #8135 

The relevant scopes for the EmployerService API endpoint at route ` /api/patient/{puuid}/employer` were missing `{patient,user,system}/employer.read` scopes. 

Minor changes adding the relevant scopes to the `apiScopes()` array in the ScopeRepository and an update to the API_README. I would like to note that the API_README did not include patient or system contexts in the list, though I see them in the ScopeRepository. Perhaps this is for the sake of bevity.

No


msummers42:
@adunsulag let's chat about this a bit when you have a moment. I spent some time walking through the various rest services. I have not executed tests for this yet, but mainly wanted to get an idea around how this sort of thing is managed. I tend to agree that this could be moved under the Patient object, though keeping it separate yields some flexibility.

I am interested to know how access is managed for a patient, specifically.

adunsulag:
Will circle back with you when I've addressed a client fire right now.  Database architecture has user accounts siloed from patient accounts.  Patients have no ACLs only users so access is managed via scopes and very tight binding of the patient pid/uuid on the apis.  

Key point is to know that patient accounts cannot make calls into anything in /apis/default/api/ only into FHIR endpoints (/apis/default/fhir/) and portal endpoints (/apis/default/portal/).

There was a back and forth discussion amongst admins at time I built FHIR endpoints to consolidate the access apis for our 'standard' api but decision was made out of a security concern to keep a very clear separation from patient apis and regular user apis.

_rest_routes.inc.php
```diff
@@ -6222,15 +6224,29 @@
      *          response="401",
      *          ref="#/components/responses/unauthorized"
      *      ),
-     *      security={{"openemr_auth":{}}}
+     *      security={{"openemr_auth":{"user/employer.read", "patient/employer.read"}}}
      *  )
      */
     "GET /api/patient/:puuid/employer" => function ($puuid, HttpRestRequest $request) {
+        if (!UuidRegistry::isValidStringUUID($puuid)) {
```
adunsulag, line 6231: Only thing I could think of would be an architectural improvement to build a Validator class for GET requests.  We use the Validator on insert/update requests (@see PatientValidator).  FHIR will catch if the uuid is invalid, but these standard apis probably throw an exception if the UUID is invalid which is what you're checking here so I think this is fine.

msummers42: This basic pattern would be great to apply for all GETs, though I think that it may be tricky to do in a general way given the variance in possible inputs. Some are UUID, some are other string IDs, etc. Having some common methods, perhaps using an Attribute pattern could be useful.

PR #8194, opened 2025-03-29
Usage Tracking

### LINKED ISSUE #8188: feat: Collect usage data from production installs.
### Describe the feature

Track non PHI telemetry for usage and locale data

Under review. @adunsulag suggested using audit log perhaps and my original thought was tracking top level click events. Maybe combine the two.
@bradymiller and @stephenwaite suggest opt-out on the user level. (I hope I understood correctly. If not correct me)

I developed a click event javascript a couple years ago that worked well however it was never stress tested.
I'm looking at the audit log now so comments welcomed.

sjpadgett:
Fixes #8188 

Addition of usage data tracking.

created new javascript method to save click events 
hooked into existing menuActionClick method to call new event save method.
create back-end script to handle ajax calls for tasks and database actions


sjpadgett:
I think this is pretty slim but I could use a module. Of course still need the user opt-in/out piece.
Still need cleaning up but I wanted to give everyone a look at my thinking and get enough in place that @jesdynf can have something to work with and others to comment on.


Here is a endpoint payload json example
[payload.zip](https://github.com/user-attachments/files/19515088/payload.zip)

jesdynf:
I can get this into DDB, no problem. Which sounds nicer?

1. SiteID, IngestionTimestamp[^1], [...LocaleData fields], EventJSON (unprocessed JSON string)
2. SiteID, IngestionTimestamp, [...LocaleData fields], Events ([List of Maps](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html))

Door #&#8203;2 would use a List of Maps, translating the JSON into the native form DDB wants to work with. Which one you want depends on how you're thinking of writing your app -- if you're fine with eating a month's worth of sites all at once and then adding them up, you might not want to bother with dealing with Amazon's typing and structure. If you want to be able to do more subtle searches and query the database for events using Dynamo's search structure, then we need to marshal the events into something Dynamo recognizes.

[^1]: It would be useful if this was actually just a single YYYYMM block, with one row per site per month, instead of (or alternately, accompanied by) an actual ingestion timestamp. Having OpenEMR submit on March 3rd all the site data for Feb 2025 would let us unambigiously mark the record as the 202502 submission for that site ID, which would be nice.

sjpadgett:
I got to thinking that Asher is right that most likely IP will be the local intranet IP. Anybody have a hint where I can find a more acceptable location data point. Also Asher maybe we'll rename location_ip to location. Also can you send me on Signal(hopefully we don't have any reporters in our channel!) the endpoint working for you when you get there. Also your final DB table structure.

sjpadgett:
@jesdynf Here's my table structure 
```
CREATE TABLE IF NOT EXISTS `track_events`
(
    `id`           BIGINT(12) UNSIGNED NOT NULL AUTO_INCREMENT,
    `event_type`   VARCHAR(50)  NOT NULL,
    `event_label`  VARCHAR(255) NOT NULL,
    `event_url`    VARCHAR(255) DEFAULT NULL,
    `event_target` VARCHAR(50)  DEFAULT NULL,
    `first_event`   DATETIME   NOT NULL,
    `last_event`   DATETIME     NOT NULL,
    `count`        INT UNSIGNED NOT NULL DEFAULT 1,
    PRIMARY KEY (`id`),
    UNIQUE KEY `unique_event_label` (`event_label`)
) ENGINE = InnoDB;
```

library/ajax/track_events.php
```diff
@@ -0,0 +1,136 @@
...
+    http_response_code(400);
+    echo json_encode(array("error" => "Invalid action"));
+}
+
+function reportClicks()
+{
+    global $data;
+    $eventType = $data['eventType'] ?? '';
+    $eventLabel = $data['eventLabel'] ?? '';
+    $eventUrl = $data['eventUrl'] ?? '';
```
adunsulag, line 36: One thing that comes to my mind is if any of the menuURLs would have PHI appended to the query string.  I don't think we do this with any of our menu URLs but I could see someone modifying the standard.json or a module writer adding a url with a specific query string embdded in the click event. 

I wonder if eventUrl should be passed through a URL() object and grab everything but the query params.

There's also the possibility of someone doing a mod_rewrite and putting PHI directly as a directory separator but I think that's more unlikely.  However, if that could be a valid concern perhaps we need to look at having a way to blacklist certain URLs or URL patterns from the background tracking.

interface/main/tabs/js/tabs_view_model.js
```diff
@@ -372,6 +373,29 @@ function menuActionClick(data,evt)
 
 }
 
+function reportMenuClickData(data, evt = 'menuClick') {
+    top.restoreSession();
+    const clickEventData = {
```
adunsulag, line 378: I think this looks pretty good.  We can build on this to grab the AuditEvents where type is equal to 'http-request' and post those to the tracker at a later date as well.

sjpadgett:
Have dialog moved. 
image

interface/main/tabs/js/tabs_view_model.js
```diff
@@ -372,6 +373,29 @@ function menuActionClick(data,evt)
...
+        action: 'reportMenuClickData',
+        eventType: evt,
+        eventLabel: data.label(),
+        eventUrl: data.url(),
+        eventTarget: data.target,
+        csrf_token_form: top.csrf_token_js,
+    };
+
+    let url = top.webroot_url + '/library/ajax/track_events.php';
+    fetch(url, {
```
adunsulag, line 388: I verified this is all async non-blocking so it won't cause any performance penalties on the client side.

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -120,3 +120,19 @@ ALTER TABLE `onetime_auth` ADD `profile` tinytext COMMENT 'profile of scope for
...
+
+#ifNotTable track_events
+CREATE TABLE `track_events`
+(
+    `id`           BIGINT(12) UNSIGNED NOT NULL AUTO_INCREMENT,
+    `event_type`   VARCHAR(50)  NOT NULL,
+    `event_label`  VARCHAR(255) NOT NULL,
+    `event_url`    VARCHAR(255) DEFAULT NULL,
+    `event_target` VARCHAR(50)  DEFAULT NULL,
+    `first_event`   DATETIME   NOT NULL,
```
adunsulag, line 132: Are first_event,last_event and count a measure to save space?

From an architecture view will we want to track larger amounts of data that won't fit within a 255 varchar?

I'm thinking if we want to expand events at some point in the futureto include instance installation details such as (language, region, timezone, portal_enabled) etc?

adunsulag (commented):
Just some thoughts on URLs, db architecture, and future usage.

library/ajax/track_events.php
sjpadgett: good point will sanitize

sql/7_0_3-to-7_0_4_upgrade.sql
sjpadgett: For click events I only care that it was clicked. Didn't want to recorded new row for every click with turns into duplicated event records. The eventType would give use event context like portal or module etc.
The AWS DynamoDB stores the events array as json object so structure is free to change.


> I'm thinking if we want to expand events at some point in the futureto include instance installation details such as (language, region, timezone, portal_enabled) etc?

These are normalized and indexed. We can add new columns dynamically as NoSQL db.

jesdynf:
A quick reference, if anyone finds it useful:

```payload = json.loads(event['body'])
        dynamoRequest = { 'TableName': '#redacted#' }
        
        item = {}
        for key, value in payload['localeData'].items():
            item[key] = { 'S' : value }
        item['usageRecords'] = { 'S' : json.dumps(payload['usageRecords']) }
        item['SiteID'] = item['site_uuid'];
        # temporary fix for missing sort key
        item['ReportingInterval'] = item['reporting_interval'] if 'reporting_interval' in item else { 'S': '202503' }
        del item['site_uuid']
        item.pop('reporting_interval', None)

        dynamoRequest['Item'] = item
```

So basically the object is "every key in localeData, with two of them renamed" and then usageData which is just the usageRecords JSON, unpacked and repacked but otherwise unprocessed.

sql/7_0_3-to-7_0_4_upgrade.sql
bradymiller: Since the json structure may change in the future, does it makes sense to version it (ie. dynamodb column with json_version (or some other name) that stores the version of the json (so can then make changes and be backward compatible to still do analytics on prior version json structures)?

jesdynf:
It makes sense, but that's not the same thing as saying I'd bother. Just be
generous with what the report compiler accepts and you'll probably be fine?

library/ajax/track_events.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+    // Collect instance data.
+    $localeData = array(
+        'site_uuid' => $site_uuid,
+        'reporting_interval' => $interval,
+        'location' => '',
+        'time_zone' => $time_zone ?? $GLOBALS['gbl_time_zone'] ?? '',
+        'locale' => locale_get_default(),
+        'version' => $versionService->asString(),
+        'distribution' => getenv('OPENEMR_DOCKER_ENV_TAG', true) ?? $versionService->asString(),
```
bradymiller, line 108: prob just set above 'distribution' to blank if it doesn't exist

Also, would it make sense to record OS info (maybe in os column or something like that) from php_uname() command?

library/ajax/track_events.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    $versionService = new VersionService();
+    $endpoint = "https://reg.open-emr.org/api/usage?SiteID=" . urlencode($site_uuid);
+    $interval = date("Ym", strtotime("-30 Days"));
+    $time_zone = sqlQuery("SELECT `gl_value` as zone FROM `globals` WHERE `gl_name` = 'gbl_time_zone' LIMIT 1")['zone'] ?? null;
+    // Query the track_events table for usage data.
+    $sql = "SELECT event_type, event_label, event_url, event_target, first_event, last_event, count FROM track_events";
+    $result = sqlStatement($sql);
+    $usageRecords = array();
```
bradymiller, line 96: Is the plan to truncate/clear the data in the table after collecting it (or sending it) so don't keep sending same data on each transmit?

templates/login/partials/js/product_reg.js.twig
```diff
@@ -1,12 +1,26 @@
...
+
+    {% set regConstants = {
+        'webroot': webroot
+    } %}
 
-    document.addEventListener('DOMContentLoaded', () => {
-        document.getElementById('authUser').focus();
-    });
+    const registrationTranslations = {{ regTranslations|json_encode|raw }};
+    const registrationConstants = {{ regConstants|json_encode|raw }};
```
bradymiller, line 23: no need for the raw in above 2 lines. we do not use twig auto escaping, so no need to state that (ie. output from twig is always raw since we use our own custom escaping functions).

sjpadgett:
no clue why test fail.

sjpadgett:
example of my location from new geo class: Populate in the location: element sent to server.

{"country":"United States","region":"Florida","city":"Cape Coral","latitude":"26.6786","longitude":"-82.0263"}

city is the location of my web providers server location.

sjpadgett:
TODO trap and handle offline internet for all transactions.

tests/bootstrap.php
```diff
@@ -7,6 +7,10 @@
     exit;
 }
 
+if (!defined('APP_UNDER_TEST')) {
+    define('APP_UNDER_TEST', true);
+}
+
```
sjpadgett, line 13: Hey guys I don't have my windows environment set up to run these test so I'm guessing at a fix for the registration modal blocking test menu clicks.
If my last attempt doesn't work, you guys may need to jump in and help me out. thanks!

adunsulag:
I pulled down your PR and tried it locally.  Also experiencing the registration modal preventing the login click from working.  Wasn't able to get a fix working this morning, got a number of family events happening today but may have time this evening to try this out again otherwise perhaps Brady can take a look tomorrow.  Should be on the Sat call if want to discuss it.

src/Telemetry/BackgroundTaskManager.php
```diff
@@ -0,0 +1,68 @@
...
+ * Manages background task operations for telemetry.
+ */
+class BackgroundTaskManager
+{
+    /**
+     * Creates or updates the background telemetry task.
+     *
+     * @param string $period Number of days (as a string) for the task interval.
+     */
+    public static function modifyTelemetryTask(string $period = '33'): void
```
bradymiller, line 24: rec just hard-coding the 33 to simplify

interface/main/tabs/main.php
```diff
@@ -59,342 +77,343 @@
...
+    jsGlobals.time_display_format = <?php echo js_escape($GLOBALS['time_display_format']); ?>;
+    jsGlobals.timezone = <?php echo js_escape($GLOBALS['gbl_time_zone'] ?? ''); ?>;
+    jsGlobals.assetVersion = <?php echo js_escape($GLOBALS['v_js_includes']); ?>;
+    var WindowTitleAddPatient = <?php echo($GLOBALS['window_title_add_patient_name'] ? 'true' : 'false'); ?>;
+    var WindowTitleBase = <?php echo js_escape($openemr_name); ?>;
+    const isSms = "<?php echo !empty($GLOBALS['oefax_enable_sms'] ?? null); ?>";
+    const isFax = "<?php echo !empty($GLOBALS['oefax_enable_fax']) ?? null?>";
+    const isServicesOther = (isSms || isFax);
+    const allowTelemetry = <?php echo $allowTelemetry; ?>;
+    const registerOptOut = <?php echo $registerOptOut; ?>;
```
bradymiller, line 128: would js_escape() both variables above

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -120,3 +120,19 @@ ALTER TABLE `onetime_auth` ADD `profile` tinytext COMMENT 'profile of scope for
...
+    `last_event`   DATETIME     NOT NULL,
+    `count`        INT UNSIGNED NOT NULL DEFAULT 1,
+    PRIMARY KEY (`id`),
+    UNIQUE KEY `unique_event_label` (`event_label`)
+) ENGINE = InnoDB;
+#endIf
```
bradymiller, line 138: also need database changes in database.sql

bradymiller:
Getting a javascript log error on the login screen (not breaking anything but looks like something related to registration is still trying to be used on the login screen):
```
Uncaught TypeError: top.restoreSession is not a function
    ProductRegistrationService http://localhost:8300/interface/product_registration/product_registration_service.js?v=92c58df35f1c73380f3f116fedd7f22c:15
    ProductRegistrationController http://localhost:8300/interface/product_registration/product_registration_controller.js?v=92c58df35f1c73380f3f116fedd7f22c:69
    init http://localhost:8300/interface/login/login.php?site=default:34
    <anonymous> http://localhost:8300/interface/login/login.php?site=default:31
```

bradymiller:
Does a user opt-out by Submit click with a blank email and not selecting the toggle?

sjpadgett:
> Does a user opt-out by Submit click with a blank email and not selecting the toggle?
It opts out of registration and opts to whatever is selected for the checkbox

sjpadgett:
> Getting a javascript log error on the login screen (not breaking anything but looks like something related to registration is still trying to be used on the login screen):
> 
> ```
> Uncaught TypeError: top.restoreSession is not a function
>     ProductRegistrationService http://localhost:8300/interface/product_registration/product_registration_service.js?v=92c58df35f1c73380f3f116fedd7f22c:15
>     ProductRegistrationController http://localhost:8300/interface/product_registration/product_registration_controller.js?v=92c58df35f1c73380f3f116fedd7f22c:69
>     init http://localhost:8300/interface/login/login.php?site=default:34
>     <anonymous> http://localhost:8300/interface/login/login.php?site=default:31
> ```

Where or how are you running? Is test or normal log in?

sjpadgett:
no clue what is wrong for mysql from my addition to DB!

sql/database.sql
```diff
@@ -13718,3 +13727,22 @@ INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`) VALUES ('dsi
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`) VALUES ('dsi_evidence_source_attributes', 'evidence_based_dob', 'Intervention use of date of birth as expressed in the standards in US § 170.213', 110);
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`) VALUES ('dsi_evidence_source_attributes', 'evidence_based_sdoh', 'Intervention use of social determinants of health data as expressed in the standards in US § 170.213', 120);
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`) VALUES ('dsi_evidence_source_attributes', 'evidence_based_health_status', 'Intervention use of health status assessments data as expressed in the standards in US § 170.213', 130);
+
+-- -----------------------------------------------------
+--
+-- Table structure for table 'product_registration'
```
bradymiller, line 13733: wrong table name in comment

bradymiller:
@sjpadgett ,
Just running normal mode from easy dev environment. Reason v is not an integer if have a .dev file (see https://github.com/openemr/openemr/blob/master/.env.example) with `OPENEMR__ENVIRONMENT=dev`, which is a neat thing @robertdown built to have cachebuster always change rather than be 1 integer when developing:
https://github.com/openemr/openemr/blob/master/version.php#L46

bradymiller:
Haven't done extensive testing, but looks like this will always show both the email registration and the telemetry opt in. Is there a way to support use cases where only one is missing. For example, if the email registration is already set, then only show the opt in telemetry. And if already have opt in telemetry set but no email registration, then only show the email registration.

sjpadgett:
> gives an error popup that the email is invalid, so does not allow a user to leave it blank.

@bradymiller Works for me. Was an issue but I fixed so maybe I didn't put in commit.


> will always show both the email registration

I set up flags in main to do that but got distracted by a pretty girl and completely forgot!

bradymiller:
Goodness @sjpadgett , this was a lot of work! Looking forward to testing it more soon

2DV

sjpadgett:
I'm testing on dev easy now. been really busy with a couple user requests one of which is fax module. We got a great review by a user for etherfax on the forum for etherfax and I hope RC can be made as easy and feature rich.
BTW: I still marvel at your review skills!

bradymiller:

PR #8206, opened 2025-04-01
Show e-sign amendment. Fixes #8205

### LINKED ISSUE #8205: bug: e-sign does not show amendment information
### Describe the bug

If you create an e-sign with an amendment text, the amendment text does not show in the encounters summary screen.

Looks like the bug got introduced in this commit: 30d94bb6ca84c8c8b8372dbdf7f433679e0b322b when the suffix and valedictory were added to the e-sign log.

### To Reproduce

1. Open an unsigned encounter
2. Click the eSign button
3. Put in your account password in the dialog box that opens
4. Type in an amendment
5. Click Sign

Image

6. See on the Encounter Screen just the log and the datetimestamp with no amendment text.

Image

### Expected behavior
You see the amendment text.

Image

### Client configuration

- Browser: Google Chrome
- OpenEMR version: OpenEMR 7.0.2 (3)+ (Did not check earlier versions

adunsulag:
Fixes #8205 show the e-sign amendment.

The constructor order for the signature is off due to the addition of the suffix/valedictory tags which throws off the values of everything else.

### DIFF
--- a/library/ESign/Signature.php
+++ b/library/ESign/Signature.php
@@ -53,6 +53,8 @@ class Signature implements SignatureIF
         $uid,
         $firstName,
         $lastName,
+        $suffix,
+        $valedictory,
         $datetime,
         $hash,
         $amendment = null,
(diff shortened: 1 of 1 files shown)


bradymiller:
will bring into master and rel-703
(ok, got the right PR :) )

PR #8208, opened 2025-04-01
Fixes #8207 vital sign api duplicates

### LINKED ISSUE #8207: bug: VitalsService returns duplicate records if other encounter forms have same form_id
### Describe the bug

Testing out an API endpoint for a client with live patient data I discovered that the VitalsService is not filtering on **forms** table records that are specific to just the vitals form.  This results in the vital records being duplicated multiple times due to multiple entry ids on the cross join for the service.

### To Reproduce

1. Add vital records and clinical_note form records on either a fresh database or on an existing database with enough entries until there is at least 2 records in the **forms** that are identical for both  form records where formdir = 'vitals' and form records where formdir = 'clinical_notes'
2. Make an API call to the /fhir/Observation?category=vital-signs endpoint or to the /api/patient/{pid}/encounter/{eid}/vital endpoint 
3. See duplicate data entries

### Expected behavior
There are only a set of single unique values for the vitals.

### Client configuration

- Browser: google chrome
- OpenEMR version: 7.0.2+

adunsulag:
Fixes the duplicate entry problem in the vital sign services.

Fixes #8207

### DIFF
--- a/src/Services/VitalsService.php
+++ b/src/Services/VitalsService.php
@@ -154,7 +154,7 @@ class VitalsService extends BaseService
                         ,`date` AS date_created
                     FROM
                         forms
-                ) forms ON vitals.id = forms.form_id
+                ) forms ON vitals.id = forms.form_id AND forms.formdir='vitals'
                 LEFT JOIN (
                     select
                         encounter AS eid


bradymiller (approved):

bradymiller:
will bring into master and rel-703

bradymiller:
oh wait, disregard my note above, i meant for another PR (the esign one) :)

PR #8212, opened 2025-04-02
fixing sql mapping so validation works on the provider tab in add edi…

### LINKED ISSUE #7539: Bug: Unable to set "In Office" hours for providers
### Describe the bug

When attempting to set provider `In Office` hours using the `All day Event` and `Days Of Week` options, I see that an event doesn't get added.

### To Reproduce

> I can reproduce this on [the Demo environment](https://demo.openemr.io/) using `physician`/`physician`

<a href="url"></a>

<a href="url"></a>

1. Create a new provider event using the `In Office` category
   * Only things that I changed:
     * Set `All day event` for the date
     * Set `Days of the Week` to the current day (in my case Sunday)
2. Save the event 

### Expected behavior

When saving the event, it should appear and/or an error should be shown if there is an issue. (I checked my browser console logs as well and did not get any errors)

### Client configuration

- Browser: Brave `Version 1.66.113 Chromium: 125.0.6422.76 (Official Build)  (64-bit)` (Shields down)
- OpenEMR version: `7.0.2 (1)`
- Operating system: Pop!_OS `22.04 LTS`

bryanopenemr:
…t event

 https://github.com/openemr/openemr/issues/7539

Fixes #7539

Validation used to not work at all in the provider tab in add edit events.

There is now validation in the provider tab in add edit events. *Try saving the form with 'Repeats' or 'Days Of Week' checked but with no 'until date' set, and notice how there is now a validation message that prevents you from saving without an until date. 

No


sql/database.sql
```diff
@@ -10617,7 +10617,7 @@ INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `notes`, `ac
...
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `notes`, `activity`) VALUES ('page_validation', 'common#new-encounter-form', '/interface/forms/newGroupEncounter/common.php', 160, '{"pc_catid":{"exclusion": ["_blank"]}}', 1);
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `notes`, `activity`) VALUES ('page_validation', 'add_edit_event#theform_groups','/interface/main/calendar/add_edit_event.php?group=true',150, '{"form_group":{"presence": true}}', 1);
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `notes`, `activity`) VALUES ('page_validation', 'add_edit_event#theform_prov', '/interface/main/calendar/add_edit_event.php?prov=true', 170, '{}',  1);
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `notes`, `activity`) VALUES ('page_validation', 'add_edit_event#theform_prov', '/interface/main/calendar/add_edit_event.php?prov=true', 170, '{"form_title":{"presence": {"message": "Title Required"}}}',  1);
```
adunsulag, line 10620: Any changes you make need to be brought over into the sql/7_0_3-to-7_0_4_upgrade.sql file.  Since its a new release it would be the first entry.

Let me know if you run into problems with the database upgrade DSL, you'll likely have to use something like #IfRow3D or #IfRowNot3D for this.  Where the string constant has a '#' in it could break things but don't know for sure since the '#' is treated as a command character.

You also need to bump the /version.php $v_database value as part of this.

interface/main/calendar/add_edit_event.php
```diff
@@ -1904,7 +1912,7 @@ function validateform(event,valu){
     }
     ?>
 
-    var submit = submitme(1, event, 'theform', collectvalidation);
+    var submit = submitme(1, event, '<?php echo $form_id; ?>', collectvalidation);
```
adunsulag, line 1915: Drop the quotes and use <?php echo js_escape($form_id); ?> here (it will auto add the quotes).  Since this file is so huge even though you've defined constants 500 lines above its safer to escape it here in case anyone modifies it between the locality you defined this variable, and the locality it is used.

interface/main/calendar/add_edit_event.php
```diff
@@ -1925,11 +1933,11 @@ function validateform(event,valu){
 
 // disable all the form elements outside the recurr_popup
 function DisableForm() {
-    $("#theform").children().attr("disabled", "true");
+    $("#<?php echo $form_id; ?>").children().attr("disabled", "true");
```
adunsulag, line 1936: Same escaping thing as mentioned above, ```$("#" + <?php echo js_escape($form_id); ?>)).children()```

interface/main/calendar/add_edit_event.php
```diff
@@ -1925,11 +1933,11 @@ function validateform(event,valu){
...
 }
 
 function EnableForm() {
-    $("#theform").children().removeAttr("disabled");
+    $("#<?php echo $form_id; ?>").children().removeAttr("disabled");
```
adunsulag, line 1940: Same escaping thing as mentioned above, ```$("#" + <?php echo js_escape($form_id); ?>)).children()```

interface/main/calendar/add_edit_event.php
```diff
@@ -1336,7 +1336,15 @@ function find_available(extra) {
...
-<form role="form" method='post' name='theform' id='theform' action='add_edit_event.php?eid=<?php echo attr($eid) ?>'>
+<?php
+$form_id = 'theform';
+if (!empty($_GET['prov']) && ($_GET['prov'] == true)) {
+    $form_id = 'theform_prov';
+} elseif ($_GET['group'] == true) {
+    $form_id = 'theform_groups';
+}
+?>
+<form role="form" method='post' name='<?php echo $form_id; ?>' id='<?php echo $form_id; ?>' action='add_edit_event.php?eid=<?php echo attr($eid) ?>'>
```
adunsulag, line 1347: Change the attr($eid) to be attr_url($eid) to make sure $eid is urlencoded.  I saw that the existing code didn't have it, which looks like an oversight to me if $eid ever changes from a numeric value it will break url encoding.

Where the form_id is defined to close in locality you don't need attrs on those vars but for consistency would be good to have them.

adunsulag (changes requested):

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -120,3 +120,14 @@ ALTER TABLE `onetime_auth` ADD `profile` tinytext COMMENT 'profile of scope for
 #IfMissingColumn onetime_auth onetime_actions
 ALTER TABLE `onetime_auth` ADD `onetime_actions` text COMMENT 'JSON array of actions that can be performed with this token';
 #EndIf
+
```
bradymiller, line 123: This is a bit of a special situation where don't need to use the special # syntax. Would just include your last query with no #commands above or below it and add the following to the WHERE clause `AND \`note\` = ''`

This will effectively only modify the note element if it is blank, which it should be. Can safely assume that the entry is present and that note is blank (if not, this means the user has done things in the list editor that would not be consistent with a stable OpenEMR instance (ie. likely would be many more problems than the add edit form validation :) ) or if they know what they are doing have already added their own validation to that entry.

interface/main/calendar/add_edit_event.php
```diff
@@ -1336,7 +1336,15 @@ function find_available(extra) {
...
-<form role="form" method='post' name='theform' id='theform' action='add_edit_event.php?eid=<?php echo attr($eid) ?>'>
+<?php
+$form_id = 'theform';
+if (!empty($_GET['prov']) && ($_GET['prov'] == true)) {
+    $form_id = 'theform_prov';
+} elseif ($_GET['group'] == true) {
+    $form_id = 'theform_groups';
+}
+?>
+<form role="form" method='post' name='<?php echo $form_id; ?>' id='<?php echo $form_id; ?>' action='add_edit_event.php?eid=<?php echo attr_url($eid) ?>'>
```
bradymiller, line 1347: To be super safe would wrap both $form_id echos above with attr() function to htmlescape. May be seen as overkill since the variable is hard-coded above, however is somebody in the future decides to modify it and incorporate a user defined variable in it, then the attr() will stop a xss vulnerability in that case. For this reason, we generally just escape everything getting echos without putting much thought into where it comes from (except in the special case where it will include html in it, then need to escape further upstream in the code). Hope this makes sense.

sql/7_0_3-to-7_0_4_upgrade.sql
bryanopenemr: Just made another commit, let me know if it looks right :)

adunsulag:
@bradymiller pulling this down to verify it works, but from what I can see all your concerns are addressed.  Ok to bring this PR in assuming my testing works?

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -120,3 +120,12 @@ ALTER TABLE `onetime_auth` ADD `profile` tinytext COMMENT 'profile of scope for
...
+#IfNotRow3D list_options list_id page_validation option_id add_edit_event#theform_prov title /interface/main/calendar/add_edit_event.php?prov=true
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `notes`, `activity`)
+VALUES ('page_validation', 'add_edit_event#theform_prov', '/interface/main/calendar/add_edit_event.php?prov=true', 170, '{"form_title":{"presence": {"message": "Title Required"}}}', 1);
+#EndIf
```
bradymiller, line 127: would drop above block. can safely assume this is there (unless a user has purposefully removed it for some reason)

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -120,3 +120,12 @@ ALTER TABLE `onetime_auth` ADD `profile` tinytext COMMENT 'profile of scope for
...
+
+UPDATE `list_options`
+SET `notes` = '{"form_title":{"presence": {"message": "Title Required"}}}'
+WHERE `list_id` = 'page_validation' AND `option_id` = 'add_edit_event#theform_prov' AND `title` = '/interface/main/calendar/add_edit_event.php?prov=true' AND `note` = '';
```
bradymiller, line 131: `note` should be `notes` at end of line 131

sql/7_0_3-to-7_0_4_upgrade.sql
bryanopenemr: fixed in new commit :)

sql/7_0_3-to-7_0_4_upgrade.sql
bryanopenemr: fixed in new commit :)

sjpadgett (approved):
I'm about to merge a PR which will cause a conflict in this PR namely the database changes.
So if this is ready to merge, I'll merge and deal with conflict in my PR. @adunsulag are you okay with changes you've requested?

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -120,3 +120,7 @@ ALTER TABLE `onetime_auth` ADD `profile` tinytext COMMENT 'profile of scope for
 #IfMissingColumn onetime_auth onetime_actions
 ALTER TABLE `onetime_auth` ADD `onetime_actions` text COMMENT 'JSON array of actions that can be performed with this token';
 #EndIf
+
+UPDATE `list_options`
+SET `notes` = '{"form_title":{"presence": {"message": "Title Required"}}}'
+WHERE `list_id` = 'page_validation' AND `option_id` = 'add_edit_event#theform_prov' AND `title` = '/interface/main/calendar/add_edit_event.php?prov=true' AND `notes` = '';
```
bradymiller, line 126: upgrade not working on testing. Rec changing FROM `notes` ='' TO `notes` = '{}'
(this worked for me on testing, so after this change, can then bring in this PR)

bryanopenemr: Change added!

sjpadgett:
I tested the upgrade script okay so I'm merging this so I can fix conflicts with my PR and merge my tracking.
Thank you @bryanopenemr for this contribution, well done.

bradymiller:

PR #8215, opened 2025-04-02
adding minimum height for all appts under 15 minutes for all categories (#8171)

bradymiller:
rel-703
adding minimum height for all appts under 15 minutes for all categories (#8171)


stephenwaite:
homer_disappearing

PR #8220, opened 2025-04-03
feat(flow-board): show lock icon for signed/locked encounters

### LINKED ISSUE #8219: feat: Show signed documents on the flow board
### Describe the problem

Would like a visual on the flow board that shows which encounter and or encounter documents have been signed

Place a lock next to the encounter number on the flow board

juggernautsei:
Fixes #8219.

Adds a lock icon next to the encounter number on the patient flow board when that encounter has been signed and locked via ESign. Lets staff see at a glance which encounters are still pending sign-off before end of day.

## Behavior

Gated on two existing globals:

- `ptkr_show_encounter` — the flow-board encounter column must be enabled.
- `lock_esign_all` — the encounter-level esign lock feature must be enabled.

When both are on, each row with a non-zero encounter id renders `<i class='fa fa-lock'>` with a translated `Signed` title if that encounter has a locking signature in `esign_signatures`.

## Implementation notes

- Locked-encounter ids are batch-loaded once per page render (single `SELECT DISTINCT tid FROM esign_signatures ... tid IN (...)`) into a hash, and each row does an `isset()` lookup. Avoids an N+1 `SELECT` per appointment on busy boards.
- Encounter-id collection is folded into the existing `$chk_prov` loop, so the fix adds no extra pass over `$appointments`.
- Uses `QueryUtils::fetchRecords()` and `ESign\SignatureIF::ESIGN_LOCK` rather than raw `sqlStatement` + magic numbers.
- Uses Bootstrap `ml-1` for icon spacing (matches the pattern in `oe-module-dashboard-context`), with `aria-hidden='true'` on the decorative `<i>` so screen readers announce the title once.
- Attribute escaping uses `xla('Signed')` (correct helper for translated attribute values); `xlt` wraps `text()` which uses `ENT_NOQUOTES` and is not attribute-safe on its own.

## Takeover

Original work by @juggernautsei. Taken over via maintainer-edit to resolve a long-standing merge conflict and address review feedback; `billing_report.php` and `.editorconfig` touches from earlier revisions were dropped via merge with master so the PR is scoped to just the flow-board feature.


interface/billing/billing_report.php
juggernautsei, line 1: I know how this happened. OOPS

interface/patient_tracker/patient_tracker.php
```diff
@@ -137,6 +137,22 @@
...
+*/
+function is_signed(mixed$appt_enc){
+    $signed = sqlQuery(
+        "SELECT is_lock FROM esign_signatures
+               WHERE tid IN
+                     (SELECT id FROM `forms` WHERE `encounter` = ?) OR
+               tid IN (SELECT id  FROM `form_encounter` WHERE `encounter` = ?)",
+         array($appt_enc, $appt_enc)
+         );
+    return $signed['is_lock'] == 1 ? true : false;
```
adunsulag, line 153: $signed could come back as false, which you're not handling here.

interface/patient_tracker/patient_tracker.php
```diff
@@ -629,9 +645,14 @@ class="text-small"
                         <?php if ($GLOBALS['ptkr_show_encounter']) { ?>
                             <td class="detail text-center" name="kiosk_hide">
                                 <?php
+                                $signed = false;
                                 if ($appt_enc != 0) {
+                                    $signed = is_signed($appt_enc);
                                     echo text($appt_enc);
                                 }
+                                if ($signed) {
+                                    echo "<span class='text-success' title=''>&nbsp;&nbsp;<i class='fa fa-lock'></i></span>";
```
adunsulag, line 654: For accessibility, I'd recommend adding a title here.

adunsulag (changes requested):

interface/patient_tracker/patient_tracker.php
juggernautsei: @adunsulag the showing signer was not supposed to be part of this commit. That was a client request. I didn't even read what was bundled into this commit by mistake. I can clean it up and get it in. The client wanted to be able to see from the flow board which encounters had been signed or not. That way he can see what he needs to go and sign before end of his day. It was a feature in eCW.

juggernautsei: I have taken some time to think over what I was doing. I have made some code improvements. I moved the function out into the include file where is should be. Added better error handling and tightly coupled the return type.

interface/patient_tracker/patient_tracker.php
```diff
@@ -629,9 +629,15 @@ class="text-small"
                         <?php if ($GLOBALS['ptkr_show_encounter']) { ?>
                             <td class="detail text-center" name="kiosk_hide">
                                 <?php
+                                $signed = false;
                                 if ($appt_enc != 0) {
+                                    $signed = is_signed($appt_enc);
```
stephenwaite, line 634: why not use the `isSigned()` method of the the esign api like is done in the [custom report](https://github.com/openemr/openemr/blob/master/interface/patient_file/report/custom_report.php#L805)?

juggernautsei: On the patient tracker, the information that is available is the encounter number only. No form information is give in the current data set of the patient tracker. The API requires a for name that is not available at run time.

juggernautsei: Hi @stephenwaite, AI let me open the black-box and see what was inside. It found for me what I had forgotten was in the black-box ESign

juggernautsei: I tested the change in production, and it is producing the right results.

juggernautsei:
This would be a nice to have in the next version @stephenwaite @adunsulag @kojiromike

interface/patient_tracker/patient_tracker.php
```diff
@@ -629,9 +629,18 @@ class="text-small"
                         <?php if ($GLOBALS['ptkr_show_encounter']) { ?>
                             <td class="detail text-center" name="kiosk_hide">
                                 <?php
+                                $signed = false;
```
kojiromike: Where is `$signed` used?

interface/patient_tracker/patient_tracker.php
```diff
@@ -629,9 +629,18 @@ class="text-small"
...
                                 if ($appt_enc != 0) {
+                                    $signable = new ESign\Encounter_Signable($appt_enc);
+                                    // Check lock status - returns boolean
+                                    $isLocked = $signable->isLocked();
                                     echo text($appt_enc);
                                 }
+                                if ($isLocked) {
+                                    echo "<span class='text-success' title='" . xlt('Signed')
+                                    . "'>&nbsp;&nbsp;<i class='fa fa-lock'></i></span>";
+                                }
```
kojiromike, line 680:
```suggestion
                                if ($appt_enc != 0) {
                                    echo text($appt_enc);
                                    $signable = new ESign\Encounter_Signable($appt_enc);
                                    if ($signable->isLocked()) {
                                        echo "<span class='text-success' title='" . xlt('Signed')
                                        . "'>&nbsp;&nbsp;<i class='fa fa-lock'></i></span>";
                                    }
                                }
```

interface/patient_tracker/patient_tracker.php
```diff
@@ -629,9 +629,18 @@ class="text-small"
                         <?php if ($GLOBALS['ptkr_show_encounter']) { ?>
                             <td class="detail text-center" name="kiosk_hide">
                                 <?php
+                                $signed = false;
+                                $isLocked = false;
```
kojiromike:
```suggestion
```

kojiromike:
@adunsulag I think this one is just blocked on your review.

@juggernautsei it's not the end of the world, but it'd be nice if you could remove that whitespace-only change. I think that was unintentional, right?

.editorconfig
```diff
@@ -5,7 +5,6 @@ end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 4
-trim_trailing_whitespace = true
```
kojiromike, line 8: This didn't cause the problem. This could only cause whitespace to be removed from the end of lines. Your change to billing_report.php changes whitespace from the start of a line.

juggernautsei:
I was implementing this feature into another client when the agent suggested some improvements to enhance performance and security. 

Here is what it did.
Summary

I've successfully added the signed encounter lock feature to the flow board in this project with the following improvements:

Key Changes Made:

1. Import Statement (Line 30)
◦  Added use ESign\Encounter_Signable; for proper namespace support
2. Performance Optimization (Lines 363-386)
◦  Batch loading of all signed encounter statuses in a single database query
◦  Loads all locked encounters at once instead of checking each row individually
◦  Uses IN clause with parameterized query for efficiency and security
◦  Only runs if $GLOBALS['lock_esign_all'] is enabled
3. Table Header Update (Lines 446-450)
◦  Changed column header from "Copay" to "Encounter" for clarity
◦  Conditionally displayed based on $GLOBALS['ptkr_show_encounter']
4. Lock Icon Display (Lines 754-766)
◦  Uncommented the encounter column that was previously disabled
◦  Displays encounter number
◦  Shows green lock icon (fa-lock) for signed encounters
◦  Includes proper accessibility attributes (aria-label, aria-hidden)
◦  Uses translated tooltip text via xlt('Signed')
◦  Proper XSS protection with attr() function

Security & Best Practices:
✅ Parameterized queries prevent SQL injection  
✅ XSS protection with attr() and text() functions  
✅ Internationalization support with xlt()  
✅ Accessibility with ARIA attributes  
✅ Performance optimized with batch queries  
✅ Conditional check for $GLOBALS['lock_esign_all'] setting

kojiromike:
Taking this over — the branch had drifted 313 commits with a merge conflict, and @juggernautsei noted a few months back that the original goal was just to land the flow-board lock icon. Pushed three commits via maintainer-edit:

- **Merge of `upstream/master`** — resolved the `billing_report.php` conflict by taking master's version, which drops the unrelated whitespace churn (the substantive feature was only ever in `patient_tracker.php`).
- **`style(flow-board): use Bootstrap ml-1 for lock icon spacing`** — replaced `&nbsp;&nbsp;` with the `ml-1` utility class (matches the pattern in `oe-module-dashboard-context`), moved the `<i>` inside the `<span>` that already has the title, and added `aria-hidden="true"` on the decorative icon.

All pre-commit hooks green locally (phpcs, phpstan level 10, rector, codespell, require-checker). Will squash on merge so the intermediate history doesn't land.

@adunsulag @stephenwaite — re-requesting review; the only behavioral addition vs. the prior state is the lock-icon block at `interface/patient_tracker/patient_tracker.php` lines 639–648.

interface/billing/billing_report.php
kojiromike: Resolved by the takeover merge — `billing_report.php` was restored to master's version, so the unintentional whitespace changes are no longer in the PR.

interface/patient_tracker/patient_tracker.php
```diff
@@ -137,6 +137,22 @@
     }
 }
 
+/**
+ * Checks if an encounter of form is signed
+* @param mixed $appt_enc
+* @return bool
+*/
+function is_signed(mixed$appt_enc){
```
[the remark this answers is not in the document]
kojiromike: Superseded by the takeover — the helper function with the `mixed` return type no longer exists. The current implementation calls `$signable->isLocked()` inline at `interface/patient_tracker/patient_tracker.php` line 643, which returns `bool` via `ESign\Encounter_Signable`.

interface/patient_tracker/patient_tracker.php
kojiromike: Superseded by the takeover — there is no `$signed` variable anymore. The current code calls `$signable->isLocked()` directly, which returns `bool`, so there is no `false` vs. "signed" confusion to handle.

interface/patient_tracker/patient_tracker.php
kojiromike: Addressed — the lock icon now has a translated `title='Signed'` on the wrapping span and `aria-hidden='true'` on the decorative `<i>` (lines 644–646), so screen readers announce the label once and skip the icon glyph.

interface/patient_tracker/patient_tracker.php
kojiromike: The current code uses `ESign\Encounter_Signable::isLocked()`, which is the right fit here:

- The `ESign::isSigned()` method in `custom_report.php` lives on the assembled `ESign` object, which is built via `Form_Factory` and requires a `formdir` / `formid`. The flow board only has `$appt_enc` (the encounter id) at that point, so constructing a full `ESign` object isn't straightforward.
- `Encounter_Signable::isLocked()` takes only the encounter id and gates on the `lock_esign_all` global, delegating to the parent `DbRow_Signable::isLocked()` which queries `esign_signatures` for `is_lock = 1`.
- Semantically, "is this encounter locked?" matches the UI intent (show a lock icon for finalized encounters) better than "does it have any signature at all?" — an unlocked signature shouldn't display a padlock.

@juggernautsei tested this against production data and confirmed correct results.

interface/patient_tracker/patient_tracker.php
kojiromike: Superseded — pushed a better version in the takeover (commit `4ebd397`): replaced `&nbsp;&nbsp;` with the Bootstrap `ml-1` utility class (matches the pattern in `oe-module-dashboard-context`), moved the `<i>` inside the titled span, wrapped the title in `attr(xlt(...))`, and added `aria-hidden='true'` on the icon.

.editorconfig
kojiromike: Moot — `.editorconfig` is no longer touched by this PR. The takeover merge restored it to master along with `billing_report.php`.

kojiromike:
@aisle-analyzer

interface/patient_tracker/patient_tracker.php
```diff
@@ -339,10 +343,38 @@
...
+    // Batch-load locked encounter ids so the per-row lock-icon check doesn't
+    // issue a query per appointment.
+    $lockedEncounters = [];
+    if ($collectLockedEncounters && count($encounterIds) > 0) {
+        $ids = array_keys($encounterIds);
+        $placeholders = implode(',', array_fill(0, count($ids), '?'));
+        $lockQuery = "SELECT DISTINCT tid FROM esign_signatures "
+            . "WHERE `table` = 'form_encounter' AND is_lock = ? AND tid IN ($placeholders)";
+        $lockParams = array_merge([SignatureIF::ESIGN_LOCK], $ids);
+        foreach (QueryUtils::fetchRecords($lockQuery, $lockParams) as $lockRow) {
```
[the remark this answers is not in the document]
juggernautsei: @kojiromike do you want me to address the issue?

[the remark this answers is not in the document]
kojiromike: @juggernautsei sure, if you like. Filed #11824

PR #8222, opened 2025-04-03
Update vitals form to display date vs last_updated

### LINKED ISSUE #8221: Update vitals form to display date vs last_updated
### Describe the problem

When creating patient report (Patient --> Report), and selecting vitals, the data for time is pulled from `last_updated` column which may be clinically not relevant to the encounter if it was edited or entered at later times.

When pulling data from the vital signs table I would like to skip `last_updated` and keep `date`. Also I would like ot formate the date.

moussa854:
Fixes #8221 

When creating patient report (Patient --> Report), and selecting vitals, the data for time is pulled from `last_updated` column which may be clinically not relevant to the encounter if it was edited or entered at later times. Also, unlike `date`, `last_updated` can not be adjusted from the web interface. By replacing  `last_updated` with `date`:
- the `date` will the same as `last_updated` if not changed
- it can be adjusted with the date time picker
- It is more clinically relevant to the time vital sign actually recorded

image
In the above example teh encounter is in 2014 but becuase vital signs were updated recently it says 2025

with the proposed changes the date can be set using the time / date picker to the more clinically relevant time and date 
image
 
This PR edits one file only `interface/forms/vitals/report.php`

When pulling data from the vital signs table I would like to skip last_updated and keep date. Also I would like ot formate the date.

### DIFF
--- a/interface/forms/vitals/report.php
+++ b/interface/forms/vitals/report.php
@@ -46,13 +46,23 @@ function vitals_report($pid, $encounter, $cols, $id, $print = true)
                 $key == "id" || $key == "pid" ||
                 $key == "user" || $key == "groupname" ||
                 $key == "authorized" || $key == "activity" ||
-                $key == "date" || $value == "" ||
+                $key == "last_updated" || $value == "" ||
                 $value == "0000-00-00 00:00:00" || $value == "0.0"
             ) {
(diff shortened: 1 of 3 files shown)


moussa854:
opened new PR with only one file affected

PR #8223, opened 2025-04-03
fixing issue of allow_custom_report being treated as a string instead…

### LINKED ISSUE #8178: bug: Patient Portal allow_custom_report global setting is not used in portal dashboard page
### Describe the bug

Discovered that the allow_custom_report global option is not being used in the patient portal.  There's a bug where the twig value is being used as a string instead of an actual variable.  This is on lines 606, 711, and 786 in templates/portal/home.html.twig (can search for **'allow_custom_report'** to get all instances if line numbers have changed).

- OpenEMR version: 7.0.2 (3) +

bryanopenemr:
… of global

 https://github.com/openemr/openemr/issues/8178

Fixes #8178

The allow_custom_report global had quotations around them, which caused it to be treated as a string instead of a global. Which causes the check to show medical reports to be invalid - the report will be shown regardless of the value of the global. 

The allow_custom_report global now has no quotations around them, meaning they should now be treated as a global. 

No

### DIFF
--- a/templates/portal/home.html.twig
+++ b/templates/portal/home.html.twig
@@ -604,7 +604,7 @@
                                     } %}
                                 {% endif %}
                                 <!-- Reports -->
-                                {% if ccdaOk or 'allow_custom_report' or portal_onsite_document_download %}
+                                {% if ccdaOk or allow_custom_report or portal_onsite_document_download %}
                                     {% include "portal/partial/_nav_icon.html.twig"
                                         with {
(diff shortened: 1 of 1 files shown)


bradymiller:
hi @bryanopenemr , Code looks good (didn't test since straightforward fix). Thanks for the fix!

bradymiller:
Will bring this into master and rel-703

PR #8224, opened 2025-04-03
CCDA Fixes

sjpadgett:
remove all forms of br in XML

Fixes #

`<br />` can mess up parsing of narratives and lists so I remove. In this case I missed the invalid XHTML `<br></br>`
I also validate XML using DOMDocument() and convert to UTF-8 if needed.
After all the different versions of XML from other vendors, I've found not validating XML caused headaches that were difficult to track down.

Remove all parts of the element

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/CarecoordinationTable.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/CarecoordinationTable.php
@@ -33,6 +33,7 @@ use OpenEMR\Services\Cda\CdaValidateDocuments;
 use OpenEMR\Services\Cda\XmlExtended;
 use OpenEMR\Services\CodeTypesService;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use function PHPUnit\Framework\throwException;
 
 class CarecoordinationTable extends AbstractTableGateway
 {
@@ -228,7 +229,7 @@ class CarecoordinationTable extends AbstractTableGateway
             $xml_to_array = new XmlExtended();
             $xml = $xml_to_array->fromString($this->conditionedXmlContent);
         } catch (Exception $e) {
-            die($e->getMessage());
+            throw new Exception($e->getMessage());
         }
         // Document various sectional components
         $components = $xml['component']['structuredBody']['component'];
@@ -2322,13 +2323,13 @@ class CarecoordinationTable extends AbstractTableGateway
      * @return string Cleaned XML content.
      * @throws Exception If the input XML is invalid or cannot be parsed.
      */
-    function cleanCcdaXmlContent(string $xmlContent, bool $removeBr = false): string
+    function cleanCcdaXmlContent(string $xmlContent, bool $replaceBr = false): string
     {
         // Handle <br/> tags if required
-        if ($removeBr) {
-            $xmlContent = preg_replace('/<br\s*\/?>/', '', $xmlContent); // Remove <br/>
+        if ($replaceBr) {
+            $xmlContent = preg_replace('/<\/?br\s*\/?>/i', '', $xmlContent);
         } else {
-            $xmlContent = preg_replace('/<br\s*\/?>/', "\n", $xmlContent); // Replace <br/> with newline
+           $xmlContent = preg_replace('/<\/?br\s*\/?>/i', '\n', $xmlContent); // Replace <br/> with newline
         }
         $xmlContent = preg_replace('/\xC2\xA0/', '', $xmlContent);
         $xmlContent = str_replace('Â', '', $xmlContent);


sjpadgett:
@juggernautsei Looks like your sample imports well as it is a simple XML.
If you're moving a population of patients use our newly updated command line tool 
`contrib/util/ccda_import/import_ccda.php -- help`
ex: 
`php import_ccda.php --sourcePath=/var/www/localhost/htdocs/openemr/contrib/import_ccdas/available_ccdas --site=default --openemrPath=/var/www/localhost/htdocs/openemr --dedup=true`

Works very fast and well. Kofi just did a move for eMD of I think 15,000 patients.

juggernautsei:
Not working

      sherwin@sherwin-ThinkStation-P330-Tiny:/var/www/html/openemr703$ sudo contrib/util/ccda_import/import_ccda.php -- help
       [sudo] password for sherwin: 
       sudo: contrib/util/ccda_import/import_ccda.php: command not found

image

juggernautsei:
I have tried with --dedup=true and --dedup=false and now I realize what that command really is and it does nothing for running in development mode.

sjpadgett:
I've tested in both environments windows and easy-dev. De-dup will move duplicates to new directory if a duplicate is encountered. I forget how relaxed weighting is. I don't use dev since I refactored it.
I helped eMD via Mario/Kofi add fixes to imports which covered a lot of chart. Kofi, once he caught up to my updates, was importing 15K documents. He'd start with empty database or a backup run the tool, verify and keep or redo if import didn't work properly.

sjpadgett:
@juggernautsei Did you remove the exit() from startup? L-41

// comment this out when using this script (and then uncomment it again when done using script)
exit;

sjpadgett:
I think i'll create a webpage helper and move into the modules directory.
A start
image

juggernautsei:
> @juggernautsei Did you remove the exit() from startup? L-41
> 
> // comment this out when using this script (and then uncomment it again when done using script) exit;

Yes, I commented out that line. I will be testing again today.

sjpadgett:
Try out my module I sent you

sjpadgett:
Do you think it'd be worthwhile to add to core project? Seems to make it easier

juggernautsei:
Yes, it will make life easier for DIYers. It helps techs also. I was
planning a deep dive into the CCDA code to learn how it works and fix the
issues I have experienced for years. Thanks to you, I don't have to. I can
focus on marketing.

Sherwin
757-328-2736
Juggernaut Systems Express

       The most powerful EHR on the planet
<https://www.affordablecustomehr.com>

OpenEMR Simplicity at its best in cloud-based EMRs
“For those who believe, no proof is necessary. For those who don't believe,
no proof is possible <http://beingsaved.org/innercoc.va>.”

juggernautsei:
Everything else looks as a good place to cross over and start working in
the program.
Sherwin
757-328-2736
Juggernaut Systems Express

       The most powerful EHR on the planet
<https://www.affordablecustomehr.com>

OpenEMR Simplicity at its best in cloud-based EMRs
“For those who believe, no proof is necessary. For those who don't believe,
no proof is possible <http://beingsaved.org/innercoc.va>.”

sjpadgett:
> One of the bad things this does is the calendar categories. 2k calendar categories is outrageous. I generally have to go in and programmatically turn them off because the import makes them active. If the import is a must. They should be inactive by default.

Not sure about this Sherwin. Calendar cats are tied to the procedure codes for the encounter. 2K sounds like a lot though so I will need to look at.


> file:///var/www/html/openemr703/sites/default/documents/00/1004/9e9ad483-18c8-40f2-afd1-4480c8ae85a9 How do we move the documents to the right patie

This is what the isDev=false is for. If true we don't copy documents to patient as we do in production import. The reason for this is because we need a logged in user to check if the category for document is allowed to be copied.
This occurs using Document class can_access(). 

public function can_access($username = null)
    {
        $categories = $this->get_categories();

        // no categories to prevent access
        if (empty($categories)) {
            return true;
        }

        // verify that we can access every single category this document is tied to
        foreach ($categories as $category) {
            if (AclMain::aclCheckAcoSpec($category['aco_spec'], $username) === false) {
                return false;
            }
        }
        return true;
    }
I thought I had a work around for this but don't remember. May have been I bypassed somewhere.
I need to do a refactor

sjpadgett:
Looks like there's a bug that new document foreign_id(pid) is pid+1 when isDev=false!
W/O session all bets are off and will need different approach. Personally I think saving the ccda after import is a resource hog but I get why it is wanted.

sjpadgett:
@juggernautsei 
I found the issue with running in audit mode(isDev=false). Looks like the insert_patient is running twice. Something to do with task schedule.
I'll post up once I clean up.

sjpadgett:
@juggernautsei 
I think I have our tool cleaned up pretty good. In audit/production mode ccdas are properly referenced in the patient document. The bug I found is somewhat confusing  because I could not find any reason the command class was running the import and inserts twice.
The guys that created the command runner don't usually make these kind of mistakes. lol I keep saying to myself gosh, there must be a reason!

I like the move option so that ccdas already processed are moved out of the working directory makin start and stop operation easier to track.
Duplicate also get moved to a duplicate subdirectory for easier tracking.

I could use several example files of various lengths so I can test.
I just did a 300 patient run of a clients test set and will look at calendar categories.
Also note that if you set `--isDev-false` you must include the `--authName={valid $_SESSION['userAuth']}` or will throw exception from documents controller because the ACL test for cat and file fails.

sjpad@sjpadgett MINGW64 /c/xampp/htdocs/openemr
$ `php "C:\xampp\htdocs\openemr\modules\utility\ccda_import/import_ccda.php" --authName="Administrator" --sourcePath="C:/xampp/htdocs/openemr/sites/default/documents/import_ccdas/available_ccdas" --site="default" --openemrPath="C:/xampp/htdocs/openemr" --isDev="false" --enableMoves="false" --dedup="false"`

```
openemr path: C:/xampp/htdocs/openemr
ccda imports directory: --authName=Administrator
site: default
development mode is off
Starting patients import
..........
10 patients imported (55 total seconds) (5.5 average seconds per patient for last 10 patients)
..........
20 patients imported (127 total seconds) (7.2 average seconds per patient for last 10 patients)
..........
30 patients imported (153 total seconds) (2.6 average seconds per patient for last 10 patients)
..........
40 patients imported (181 total seconds) (2.8 average seconds per patient for last 10 patients)
..........
50 patients imported (311 total seconds) (13 average seconds per patient for last 10 patients)
..........
60 patients imported (409 total seconds) (9.8 average seconds per patient for last 10 patients)
..........
70 patients imported (432 total seconds) (2.3 average seconds per patient for last 10 patients)
..........
80 patients imported (481 total seconds) (4.9 average seconds per patient for last 10 patients)
..........
90 patients imported (524 total seconds) (4.3 average seconds per patient for last 10 patients)
..........
100 patients imported (566 total seconds) (4.2 average seconds per patient for last 10 patients)
..........
110 patients imported (595 total seconds) (2.9 average seconds per patient for last 10 patients)
..........
120 patients imported (618 total seconds) (2.3 average seconds per patient for last 10 patients)
..........
130 patients imported (639 total seconds) (2.1 average seconds per patient for last 10 patients)
..........
140 patients imported (665 total seconds) (2.6 average seconds per patient for last 10 patients)
..........
150 patients imported (689 total seconds) (2.4 average seconds per patient for last 10 patients)
..........
160 patients imported (713 total seconds) (2.4 average seconds per patient for last 10 patients)
..........
170 patients imported (806 total seconds) (9.3 average seconds per patient for last 10 patients)
..........
180 patients imported (864 total seconds) (5.8 average seconds per patient for last 10 patients)
..........
190 patients imported (903 total seconds) (3.9 average seconds per patient for last 10 patients)
..........
200 patients imported (1003 total seconds) (10 average seconds per patient for last 10 patients)
..........
210 patients imported (1159 total seconds) (15.6 average seconds per patient for last 10 patients)
..........
220 patients imported (1296 total seconds) (13.7 average seconds per patient for last 10 patients)
..........
230 patients imported (1365 total seconds) (6.9 average seconds per patient for last 10 patients)
..........
240 patients imported (1403 total seconds) (3.8 average seconds per patient for last 10 patients)
..........
250 patients imported (1430 total seconds) (2.7 average seconds per patient for last 10 patients)
..........
260 patients imported (1467 total seconds) (3.7 average seconds per patient for last 10 patients)
..........
270 patients imported (1567 total seconds) (10 average seconds per patient for last 10 patients)
..........
280 patients imported (1607 total seconds) (4 average seconds per patient for last 10 patients)
..........
290 patients imported (1637 total seconds) (3 average seconds per patient for last 10 patients)
..........
300 patients imported (1659 total seconds) (2.2 average seconds per patient for last 10 patients)

Completed patients import (300 patients) (1659 total seconds) (5.53 average seconds per patient)
Started uuid creation
Completed uuid creation (1723 total seconds; 0.47861111111111 total hours)

Finished patients import 300
```

src/Common/Command/CcdaImport.php
```diff
@@ -44,8 +44,12 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $output->writeln('document_id parameter is missing (required), so exiting');
             return 2;
         }
+        if (empty($input->getOption('auth_name'))) {
+            $output->writeln('auth_name parameter is missing (required), so exiting');
+            return 2;
+        }
+        $_SESSION['authUser'] = $input->getOption('auth_name');
 
-        $GLOBALS['modules_application']->getServiceManager()->build(CarecoordinationTable::class)->import($input->getOption('document_id'));
```
sjpadgett, line 48: @bradymiller Couldn't understand why this was used twice. Hope I didn't miss why and guessing was just overooked...

contrib/util/ccda_import/batchImport.php
```diff
@@ -0,0 +1,188 @@
...
+ * @link      https://www.open-emr.org
+ * @author    Jerry Padgett <sjpadgett@gmail.com>
+ * @copyright Copyright (c) 2025 Jerry Padgett <sjpadgett@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once("../../../interface/globals.php");
+
+use OpenEMR\Core\Header;
+
```
bradymiller, line 16: Would protect this with admin super acl

contrib/util/ccda_import/batchImport.php
```diff
@@ -0,0 +1,188 @@
...
+                echo "  --enableMoves    Optional. Set to 'true' to move processed files, 'false' to disable. Default: false.<br />";
+                echo "  --dedup          Optional. Set to 'true' to enable duplicate checking, 'false' to disable. Default: false.<br />";
+                echo "  --help           Show this help message.<br />";
+                ?>
+            </p><br />
+        </div>
+        <div class="col-12">
+            <form id="importForm" method="post" onload="scrollToBottom()">
+                <div class="form-group">
+                    <label for="sourcePath">Source Path:</label>
```
bradymiller, line 75: Translate the labels in this script (ensure do not include the colons on the labels)

contrib/util/ccda_import/batchImport.php
```diff
@@ -0,0 +1,188 @@
...
+            $sourcePath = escapeshellarg($sourcePath_val);
+            $processedPath = text(rtrim($sourcePath_val, '/') . "/processed");
+            $site = escapeshellarg($site_val);
+            $openemrPath = escapeshellarg($openemrPath_val);
+            $isDev = escapeshellarg($isDev_val);
+            $enableMoves = escapeshellarg($enableMoves_val);
+            $dedup = escapeshellarg($dedup_val);
+            $authName = escapeshellarg($_SESSION['authUser'] ?? 'admin');
+            $scriptPath = __DIR__ . '/import_ccda.php';
+            $cmd = "php " . escapeshellarg($scriptPath) .
```
bradymiller, line 137: Would put escapeshellcmd() around $scriptPath rather than escapeshellarg()

contrib/util/ccda_import/import_ccda.php
```diff
@@ -75,6 +76,7 @@ function showHelp(): void
     echo "\n";
     echo "Example:\n";
     echo "  php import_ccda.php --sourcePath=/path/to/import/documents \\\n";
+    echo "                      --authName=admin \\\n";
     echo "                      --site=default \\\n";
     echo "                      --openemrPath=/var/www/openemr \\\n";
     echo "                      --isDev=true \\\n";
```
bradymiller, line 82: would add a flag that this is running via the web page (then can ensure it will work in web page mode without breaking the original command line mode)

contrib/util/ccda_import/import_ccda.php
```diff
@@ -88,7 +90,8 @@ function outputMessage($message): void
 {
     echo("\n");
     echo $message;
-    file_put_contents("log.txt", $message, FILE_APPEND);
+    $file = $GLOBALS['OE_SITE_DIR'] . "/documents/temp/log.txt";
+    file_put_contents($file, $message, FILE_APPEND);
```
bradymiller, line 94: here's a good place where can use the above flag to save the log in the correct place (ie. in web page mode to /documents/temp/log.txt and in original command line mode in the directory that the script is run in)

contrib/util/ccda_import/import_ccda.php
```diff
@@ -123,32 +128,40 @@ function outputMessage($message): void
 use OpenEMR\Common\Uuid\UuidRegistry;
 use OpenEMR\Services\Cda\CdaComponentParseHelpers;
 
+
+// clear the log so we only keep last session. Saves cleanup time.
+unlink($GLOBALS['OE_SITE_DIR'] . "/documents/temp/log.txt");
```
bradymiller, line 133: here's a good place where can use the above flag to only do above in web page mode

contrib/util/ccda_import/import_ccda.php
```diff
@@ -123,32 +128,40 @@ function outputMessage($message): void
...
+
 // show parameters (need to do after globals)
-outputMessage("ccda directory: " . $argv[1]);
-outputMessage("site: " . $_SESSION['site_id']);
-outputMessage("openemr path: " . $openemrPath);
+outputMessage("OpenEMR path: " . $openemrPath);
+outputMessage("CCDA Imports Location: " . $argv[2]);
+outputMessage("Authority: " . $argv[1]);
```
bradymiller, line 138: Is this backwards compatible with this script used in other places (ie. docker environments). If not, is there a way to do that without needing to update all the dev dockers (for example, putting the new argument at the end or getting it in way that does not rely on ordering of parameters)?

contrib/util/ccda_import/import_ccda.php
```diff
@@ -181,22 +195,25 @@ function outputMessage($message): void
         $document->createDocument('00', 13, basename($file), 'text/xml', $fileContents);
         $documentId = $document->get_id();
         //  2. import to ccda table
-        exec("php " . $openemrPath . "/bin/console openemr:ccda-import --site=" . $_SESSION['site_id'] . " --document_id=" . $documentId);
```
bradymiller, line 184: Is there a way to have $authname not required in dev mode?

sjpadgett:
> * Also would be more secure to avoid command line stuff if possible (probably a project for another day). Ie. encapsulate functionality/duplicated code in class that is then used in both the command line script and  php web script.

This was the plan but was in hurry to get something out for a couple vendors going through moves to openemr. I now have been made aware of three different e??(eMd...) moves and it's strange that each use a different ccda format!  
I'm going to revert the webpage and revert import_ccda.php to previous with the development false fix.
Besides can't break out of run from webpage.
I'll see if I can't get new webpage version up on this PR but super busy.

Thanks for the valid review and I'll keep in mind for new webpage version and apply the ones applicable to command line version.

bradymiller:
Also regarding getting out of a run from the web page. Check out the mechanism the CDR reporting uses (it throws out processes that detach from the browser which can then be monitored via another reporting script). So could track its progress from that page without freezing. Not sure if the effort would be worth it though since can be run via commandline.

contrib/util/ccda_import/import_ccda.php
sjpadgett: this is required when not in dev mode. without it documents can't be saved to patient documents which should be the normal import mode so the original ccda can be available to user as a reference.
Sine there isn't a required order for args on command line unclear how it breaks anything.
In the end Brady authName can't go away otherwise import breaks in normal mode.

sjpadgett:
going to work encounter import to address authors being correctly identified and saved as a user.
plan username as firstlast

sjpadgett:
I could spend much more time on this subject to improve the feature. This go around has revealed more time is needed dealing with all the various text narratives because it seems that every time I get new CCDA's to look at their focus is to human readable structures and following structured template spec secondary. e.g. components that have text in text section but failed to give an id in component so we can know what text belongs to what component making predicting based on some logic(like order first to last in list) difficult to say the least.

With the CCDA's @juggernautsei provided me from eCW(?) I found they are the closest I've seen to well done. However, there were some unexpected surprises such as instead of using a LOINC or CPT4 code to describe visit type they opted to create their own code set and code types. Still having done that the document communicated it correctly as translated codes. It's these edge cases that makes importing difficult to be consistent.
However for each of these edge cases I can usually develop handlers w/o effecting normal specs. So there is that!

Anyway this PR gets us closer and next up will be another look at care/treatment plans and assessments.

Sherwin you might want to pull this PR down and quick test to check ma that I didn't forget to commit a change or script which I've be known to do when tired.

Will leave unmerged through tomorrow and I'm going back to tracking.

juggernautsei:
I did cherry-pick this before the merger. Thanks @sjpadgett.

contrib/util/ccda_import/import_ccda.php
bradymiller: As long as ordering doesn't matter and authName is not required when isDev=true, then i am good :)

sjpadgett: I've tested both options and all works well. When isDev=true I'll still pass authUser as session with the default but regardless, the $_SESSION['authUser'] is never looked at.

I'm happy I learned this feature and will now recommend to all seeking moving to OpenEMR.
Thanks for creating!

sjpadgett:
Okay that's it. Will merge after test.

bradymiller:

PR #8225, opened 2025-04-03
changing prefcat title to Exclusive Category and add description to a…

### LINKED ISSUE #7689: Appointment find available broken when preferred category is set.
### Describe the bug
If a provider has a Preferred Category default set in their schedule e.g. Office Visit then find available popup will only show time slots for that category when scheduling a new event!
### Expected behavior

New event find available should be time slot aware only. If this is a supposed feature then pipe up or I may change this behavior if I get to it! 

### Client configuration

- Browser:
- OpenEMR version: 7.0.2
- Operating system: All

bryanopenemr:
…void misleading

 https://github.com/openemr/openemr/issues/7689

Fixes #7689

While having this pref cat feature can definitely be useful in the real world, with providers that might only want to accept a certain event category for certain time frames, I think the title is quite misleading and can cause real problems. In the future I think it will be useful to have a project to make this drop down multi-select, but this should be a good temporary fix to avoid unintended filtering of availability. 

Changed the title **"Pref Cat"** to **"Exclusive Category"** and added a description next to it. 

No

### DIFF
--- a/interface/main/calendar/add_edit_event.php
+++ b/interface/main/calendar/add_edit_event.php
@@ -1706,7 +1706,10 @@ function isRegularRepeat($repeat)
 <div class="form-row mx-2">
     <div class="col-sm form-group">
         <label id='title_apptstatus'><?php echo xlt('Status'); ?>:</label>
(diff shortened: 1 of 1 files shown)


adunsulag:
I believe this is fine.  It changes up the language translations that people will have to address but the meaning is more clear on the intent.

PR #8240, opened 2025-04-07
fixing keyword processing issue in calendar search

bryanopenemr:
https://github.com/openemr/openemr/issues/8239

Fixes # 8239

#### Short description of what this resolves:
Calendar search Keywords field not working

#### Changes proposed in this pull request:
The Keywords field currently gets cleaned in the **pnVarCleanFromInput** function and since the Keywords field is just a string, it does not get processed properly and returns null. 

I added a new function, **pnStringCleanFromInput**, which properly cleans string inputs. 

#### Does your code include anything generated by an AI Engine? Yes / No
No

PR #8243, opened 2025-04-08
Improving calendar search UI

### LINKED ISSUE #8241: Calendar Search is very confusing
### Describe the bug
The and/or condition on the form is very confusing. I had no idea it is for the case where there are more than one keywords until I looked at the code. Also, the form can just use a lot of usability enhancements and features in general. 

### Expected behavior
And/Or condition should be more intuitive

### Client configuration

- Browser: All
- OpenEMR version: All
- Operating system: All

bryanopenemr:
https://github.com/openemr/openemr/issues/8241

Fixes #8241

The calendar search form is generally quite confusing and not intuitive. Could definitely use some work on usability. 

1. Added help text for the Keywords and AND/OR fields
2. Moved the AND/OR field next to Keywords field so it's clearer what it's for
3. Disabled the AND/OR field until user inputs two or more words in the Keywords field
4. Added 'Enter' to search feature
5. Added Reset/Clear button feature
6. Added hover effect to search results to show that it's clickable

No

### DIFF
--- a/interface/main/calendar/modules/PostCalendar/pntemplates/default/user/ajax_search.html
+++ b/interface/main/calendar/modules/PostCalendar/pntemplates/default/user/ajax_search.html
@@ -27,10 +27,15 @@
                 <div class="form-group">
                     <label for="pc_keywords">[-xlt t='Keywords'-]:</label>
                     <div class="form-row">
-                        <div class="col">
-                            <select class="form-control" name="pc_keywords_andor">
-                                <option value="AND">[-xlt t='AND'-]</option>
-                                <option value="OR">[-xlt t='OR'-]</option>
+                        <div class="col-md-8">
+                            <input type="text" class="form-control" name="pc_keywords" id="pc_keywords"
+                                placeholder="Search patients, providers, appointments, or comments"
+                                value="[-php-]echo attr($_POST['pc_keywords'] ?? '');[-/php-]" />
+                        </div>
+                        <div class="col-md-4">
+                            <select class="form-control" name="pc_keywords_andor" id="pc_keywords_andor" disabled>
+                                <option value="AND" [-php-] echo ($_POST['pc_keywords_andor'] == 'AND' || empty($_POST['pc_keywords_andor'])) ? 'selected' : ''; [-/php-]>[-xlt t='AND'-]</option>
+                                <option value="OR" [-php-] echo ($_POST['pc_keywords_andor'] == 'OR') ? 'selected' : ''; [-/php-]>[-xlt t='OR'-]</option>
                             </select>
                         </div>
                         <div class="col-1">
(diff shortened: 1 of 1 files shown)


interface/main/calendar/modules/PostCalendar/pntemplates/default/user/ajax_search.html
```diff
@@ -12,83 +12,89 @@
...
+                    <label for="pc_keywords">[-xlt t='Keywords'-]:</label>
                     <div class="form-row">
-                        <div class="col">
-                            <select class="form-control" name="pc_keywords_andor">
-                                <option value="AND">[-xlt t='AND'-]</option>
-                                <option value="OR">[-xlt t='OR'-]</option>
-                            </select>
+                        <div class="col-md-8">
+                            <input type="text" class="form-control" name="pc_keywords" id="pc_keywords"
+                                placeholder="Search patients, providers, appointments, or comments"
```
bradymiller, line 32: would translate the placeholder:
```
placeholder=[-xla t='Search patients, providers, appointments, or comments'-]
```

bradymiller:
Very nice improvements!
Noticed following on testing:
- The Clear doesn't reset everything back to the default (ie. empty search field and selectors back to all). Seems like it sometimes resets, but not to the original default state.
- The keywords are not making it into the search algorithm. This is not your fault and is a bug which can be fixed by changing line 125 in interface/main/calendar/modules/PostCalendar/pnuser.php to `$k = pnVarCleanFromInput('pc_keywords') ?? '';`

After address above issues, then this is ready for the codebase.

interface/main/calendar/modules/PostCalendar/pntemplates/default/user/ajax_search.html
bryanopenemr: I changed the placeholder to what you have here, but will need some help with the translations.

bryanopenemr:
> Very nice improvements! Noticed following on testing:
> 
>     * The Clear doesn't reset everything back to the default (ie. empty search field and selectors back to all). Seems like it sometimes resets, but not to the original default state.
>       ~- The keywords are not making it into the search algorithm. This is not your fault and is a bug which can be fixed by changing line 125 in interface/main/calendar/modules/PostCalendar/pnuser.php to)
>       `$k = pnVarCleanFromInput('pc_keywords') ?? '';`~ (already fixed in [fixing keyword processing issue in calendar search #8240](https://github.com/openemr/openemr/pull/8240))
> 
> 
> After address above issues, then this is ready for the codebase.

I did some more testing today, and it turns out the clear button stops working after having submitted once. Which is because the "default" values change to the submitted values after submitting, so clicking clear will only revert the values to the last submitted values. 

I tried to override this behavior by manually setting default values when the form first generates and stuff like that but couldn't get it to work. So I decided it might be better to hold off on the clear button functionality for now and get back to this in the future if necessary.

interface/main/calendar/modules/PostCalendar/pntemplates/default/user/ajax_search.html
bradymiller: Dont' need to worry about the actual translation.
fyi, we have a translation pipeline (some perl scripts) that we ideally run occassionally (we haven't run it for a couple years since it takes several hours and can be tedious): https://www.open-emr.org/wiki/index.php/OpenEMR_Internationalization_Translator_Guide#Overview

bradymiller:
hi @bryanopenemr , Sounds good and bringing this into codebase and will be in next 7.0.3 patch. thanks!

bradymiller:

PR #8246, opened 2025-04-09
Openemr feat 8162 8163 clinical notes forms

### LINKED ISSUE #8162: feat: Enable module writers to overwrite forms/clinical_notes new/edit template
### Describe the problem

As a module writer I would like to be able to override the layout as well as template variables of the forms_clinical_notes form. 

Twigify the forms/clinical_notes/new.php file.  Utilize the TemplatePageEvent and fire off an event for the twig file with the twig variables and then use the event in the forms/clinical_notes/new.php.

I also noticed there are a couple of bugs with the category selector not being reset when the twig template is cloned so will need to fix this as part of that work.

### LINKED ISSUE #8163: feat: improve clinical_notes form to use the lists Default property for note type & category.  Should hide dropdowns if lists are disabled.
### Describe the problem

It takes up time to have to set the note type and category on each note when its almost always the same for some clients.  However, since each clinic is different this should be configurable.  Also some clinics use just the note, others just the category, some don't want either option at all and just want a date and a textbox to take notes in.

As a note writer I'd like to be able to set a default note type and default category that is used when I first create a clinical notes form.  Ideally it would be best to use the Default property on the lists **Clinical Note Type** and the list **Clinical Category Type**.

I'd also like to make individual list items inactive and if the entire list is inactive then the dropdown for type or category is hidden.

### LINKED ISSUE #8245: feat: Change Clinical Notes Report Form to use list title instead of codetext for note type display
### Describe the problem

There is an inconsistency between the displaying of the note type title and what is shown when you edit a clinical note form.  Right now the form report uses the codetext value for what is displayed on the note type title which is passed through the list translation service, sometimes this value is empty if there is no corresponding code text, or the value doesn't match what is in the list options table.

The display on the report should match what is shown in the select box dropdown on the edit/create screen. I suggest using the list_options.title value for the clinical notes type list.

adunsulag:
Implemented twigify version of clinical notes.  Moved js logic to clinical-notes.js file.

Also made it so default type/category are selected in notes if there is a default put in the list for **Clinical Note Category** and **Clinical Note Type**
Option to hide type/category if list options are made inactive.

Also added the code column to always display, if the value is empty it just shows Unspecified.

Fixes #8162
Fixes #8163
Fixes #8245

### DIFF
--- a/interface/forms/clinical_notes/new.php
+++ b/interface/forms/clinical_notes/new.php
@@ -51,7 +51,7 @@ $getDefaultValue = function ($items) {
         return $val['selected'];
     });
     if (empty($selectedItem)) {
-        return ($items[0] ?? [])['value'] ?? '';
+        return ''; // default to an empty value if there is no default option
     } else {
         return array_pop($selectedItem)['value'] ?? '';
     }
--- a/interface/forms/clinical_notes/new.php
+++ b/interface/forms/clinical_notes/new.php
@@ -26,21 +26,21 @@ require_once($GLOBALS['srcdir'] . '/csv_like_join.php');
 
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Uuid\UuidRegistry;
+use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Core\Header;
+use OpenEMR\Events\Core\TemplatePageEvent;
 use OpenEMR\Services\ClinicalNotesService;
 
 $returnurl = 'encounter_top.php';
 $formid = (int) ($_GET['id'] ?? 0);
 
 $clinicalNotesService = new ClinicalNotesService();
-
+$alertMessage = '';
 if (empty($formid)) {
     $sql = "SELECT form_id, encounter FROM `forms` WHERE formdir = 'clinical_notes' AND pid = ? AND encounter = ? AND deleted = 0 LIMIT 1";
     $formid = sqlQuery($sql, array($_SESSION["pid"], $_SESSION["encounter"]))['form_id'] ?? 0;
     if (!empty($formid)) {
-        echo "<script>var message=" .
-            js_escape(xl("Already a Clinical Notes form for this encounter. Using existing Clinical Notes form.")) .
-        "</script>";
+        $alertMessage = xl("Already a Clinical Notes form for this encounter. Using existing Clinical Notes form.");
     }
 }
 if ($formid) {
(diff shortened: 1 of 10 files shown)


bradymiller (approved):

sjpadgett:
@adunsulag I'll pull down later today and test especially against already create/existing forms.
Unless you need to merge before then.

adunsulag:
@sjpadgett if you can test with existing forms that'd be great.  Thank you!

interface/forms/clinical_notes/new.php
```diff
@@ -53,218 +68,61 @@
             $record['uuid'] = UuidRegistry::uuidToString($record['uuid']);
```
sjpadgett, line 68:
```
            $record['full_name'] = sqlQuery("SELECT CONCAT(fname, ' ', lname) AS full_name FROM users WHERE username = ?", [$record['user']]) ['full_name'] ?? '';
```

sjpadgett (approved):
Testing good. Checked setting defaults in lists. Existing notes are rendering fine so thanks.
Any chance you can add the note author and the date last updated. Eventually we need to track edit author history but as of now the note author gets updated to logged in user if note is edited.(My original schema).
Of course if you don't want to fool with now, no biggie. I'll save my mods.

interface/forms/clinical_notes/templates/partials/new/_clinical-notes-row.html.twig
```diff
@@ -0,0 +1,68 @@
...
+                                {% endif %}
+                                {% if selected %}
+                                    {% set context = type.title %}
+                                {% endif %}
+                                <option value="{{ type.value|attr }}" {{ selected ? "selected='selected'" : "" }}>
+                                    {{ type.xlTitle|text }}</option>
+                            {% endfor %}
+                        </select>
+                    </div>
+                    {% endif %}
```
sjpadgett, line 53: I thought now'd be a good time to add note author and last edited date

```

                    <!-- Note Author -->
                    <div class="col-12 my-2">
                        <label class="h5">{{ 'Author'|xlt }}:</label>
                        <span>{{ obj.full_name }}</span>
                    </div>
                    <!-- Last edited date -->
                    <div class="col-12 mb-2">
                        <label class="h5">{{ 'Last Updated'|xlt }}:</label>
                        <span>{{ obj.last_updated }}</span>
                    </div>
```

adunsulag: @sjpadgett will get those added as part of this work.  Opened issue #8282 for that.  Also opened #8283 that I may roll into this, but not sure yet.

adunsulag:
Converted this to a draft so I can address the issues in #8282

sjpadgett:
@adunsulag Can I merge the PR that addressed your original issue? The Author and Last update wasn't important although I appreciate you taking it on.
I want to improve the rich text editor by expanding options on toolbar. I think I limited the toolbar a little to strictly.
I can't work on notes while this is pending. I don't plan my changes for this patch 4 but need to get my work out of the way before I forget.lol.

sjpadgett:
thanks for the fix @adunsulag

PR #8263, opened 2025-04-11
Adding validation and disabling duration for in office events

### LINKED ISSUE #8182: feat: Duration Field for In-Office Events Obsolete

bryanopenemr:
https://github.com/https://github.com/openemr/openemr/issues/8182
https://github.com/https://github.com/openemr/openemr/issues/8185

**This pull request combines pull requests #8183 and #8186 to fix the conflicts**

Fixes #8182 and #8185 

See above

See above

No

### DIFF
--- a/interface/main/calendar/add_edit_event.php
+++ b/interface/main/calendar/add_edit_event.php
@@ -1870,6 +1870,44 @@ function are_days_checked(){
 * */
 var collectvalidation = <?php echo $collectthis; ?>;
 function validateform(event,valu){
+    collectvalidation.form_hour = {
+        numericality: {
+            onlyInteger: true,
+            greaterThanOrEqualTo: 0,
+            lessThanOrEqualTo: 23,
+            message: "must have a valid hour (0-23)"
+        },
+        presence: {
+            allowEmpty: false,
+            message: "Hour is required"
+        }
+    };
+
+    collectvalidation.form_minute = {
+        numericality: {
+            onlyInteger: true,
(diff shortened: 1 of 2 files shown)


adunsulag:
So the hover looks a lot better, only downside is if we do have anyone on tablet it won't work, but I think since this is an enhancement so I'm ok with the behavior, just might be better for future work to have the errors actually inserted in the dom / removed from the dom when the validation fails.  Looks like there's a bug though when you click inside the invalid element.  Here is the hover on the initial validation fail:
image

Here is what happens when I click on the invalid element.  Note the small red circle next to the All Day Event verbiage.
image

bryanopenemr:
> So the hover looks a lot better, only downside is if we do have anyone on tablet it won't work, but I think since this is an enhancement so I'm ok with the behavior, just might be better for future work to have the errors actually inserted in the dom / removed from the dom when the validation fails. Looks like there's a bug though when you click inside the invalid element. Here is the hover on the initial validation fail: image
> 
> Here is what happens when I click on the invalid element. Note the small red circle next to the All Day Event verbiage. image

Fixed in new commit!

PR #8265, opened 2025-04-13
updated password info is now logged

nazmul-me:
I changed the updatePassword function of the AuthUtils.php file.

### DIFF
--- a/src/Common/Auth/AuthUtils.php
+++ b/src/Common/Auth/AuthUtils.php
@@ -517,7 +517,7 @@ class AuthUtils
             $this->errorMessage = xl("Password update error!");
             $this->clearFromMemory($currentPwd);
             $this->clearFromMemory($newPwd);
-            EventAuditLogger::instance()->newEvent('update events', $username, $authGroup, 1, "update password error");
+            EventAuditLogger::instance()->newEvent('password', $username, $authGroup, 0, "update password error");
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/Common/Auth/AuthUtils.php
```diff
@@ -517,6 +517,7 @@ public function updatePassword($activeUser, $targetUser, &$currentPwd, &$newPwd,
             $this->errorMessage = xl("Password update error!");
             $this->clearFromMemory($currentPwd);
             $this->clearFromMemory($newPwd);
+            EventAuditLogger::instance()->newEvent('update events', $username, $authGroup, 1, "update password error");
```
bradymiller, line 520: would use 'password' instead of 'update events' and then be more descriptive in the description. For example, here would do:
'Unsuccessful password change: update password error'

Also, for the unsuccessful logs, the 4th parameter should be a 0 instead of a 1.

bradymiller:
hi @nazmul-me , Looks good. I am going to bring in a commit into this PR to update things a bit (more granular logging of all the fails and more descriptions to make it easier to follow password administration in the logs).

bradymiller:
weird, I am unable to push to the branch in this PR even though is set that 'Maintainers are allowed to edit this pull request.' I am betting this is because this PR is from the master branch of the author. I am going to make another PR so can add my commit to this.

bradymiller:
New PR at https://github.com/openemr/openemr/pull/8279

PR #8293, opened 2025-04-21
fix null encounter class code (#8259)

### LINKED ISSUE #8259: bug: Cannot create encounter if "Class" field is hidden
### Describe the bug

If the global setting "Show Encounter Class option on Encounters" is set to "Hide on New and Edit Form", it becomes impossible to create encounters.

The message `Error creating encounter: array ( 'class_code' => array ( 'NotEmpty::EMPTY_VALUE' => 'class code must not be empty', ), )` is shown in the browser after submitting the New Encounter Form.

This is most likely the same issue as https://github.com/openemr/openemr/issues/6230

---

OpenEMR version: 7.0.3 (3)

ruthkonyn:
Fixes #8259

encounter class code can't be null. Set to default value if not shown in new encounter form
set the encounter class code to it's default value (from list _ActEncounterCode) or if no default set use the first entry
No


ruthkonyn:
i've been ignoring MariaDB test failures

interface/forms/newpatient/save.php
```diff
@@ -105,6 +105,22 @@
         $encounter_type_description = $option['title'];
     }
 }
+//RM - class_code can't be empty - use default class or if default not set take first in the list
+if (empty($class_code)) {
+    // use defualt from Value Set ActEncounterCode list
```
kojiromike, line 110:
```suggestion
    // use default from Value Set ActEncounterCode list
```

interface/forms/newpatient/save.php
```diff
@@ -105,6 +105,22 @@
...
+    $listService = new ListService();
+    $option = $listService->getOptionsByListName('_ActEncounterCode'); //get all classes
+    // find the default
+    foreach ($option as $code) {
+        if ($code['is_default']) {
+            $class_code = $code['option_id'];
```
kojiromike, line 116:
```suggestion
            $class_code = $code['option_id'];
            break;
```

kojiromike (commented):
Couple of suggestions…

interface/forms/newpatient/save.php
ruthkonyn: thanks. yes, a break is correct here - don't know why i didn't add it in!! - will do and resubmit

adunsulag (approved):

bradymiller:

PR #8295, opened 2025-04-22
Modernizing encounter report

### LINKED ISSUE #8286: Modernizing the Encounter Report
As a developer, I redesign and modernize the encounter report, I can gain more engineering experience. 
In working on the report's modernization, it came to my attention that those checkboxes can be removed if a signed column is added to the report. I would keep the default summary report. 

Image

Check the box for details, looks like this now. The original purpose for the check boxes would be encompassed in the Signed column which I have been asked to add to the legacy report a few times. 
The other thing that was less useful was if the encounter was closed on open. It is more important to know if the encounter is billed or not. The coding column should be changed to Billing. 

Image

@stephenwaite @adunsulag comments. 

All I have left to do at this point is work out the filters and the jump button

juggernautsei:
Fixes #8286

Updating the fundamental code of the report.


adunsulag:
@juggernautsei you may need to rebase this, it doesn't look like your branch is based on the most recent master version as you have a ton of extra commits here.

juggernautsei:
I understand what you are saying. From what I can tell on my side, when I ran the git command pull upstream. I got the master and the rel at the same time. Now my local branch is hosed. I am going to close this branch and make a new one. I just tested this report on a system with over 1M patients and 600 providers. It loaded 6 months of data in 15 sec.

PR #8303, opened 2025-04-24
modern take act 2

### LINKED ISSUE #8286: Modernizing the Encounter Report

juggernautsei:
Fixes #8286

Moving the report on to a framework of twig increasing the performance

### DIFF
--- a/interface/reports/encounters_report.php
+++ b/interface/reports/encounters_report.php
@@ -1,544 +1,107 @@
 <?php
 
 /**
- *  Encounters report.
- *
- *  This report shows past encounters with filtering and sorting,
- *  Added filtering to show encounters not e-signed, encounters e-signed and forms e-signed.
- *
- * @package   OpenEMR
- * @link      http://www.open-emr.org
- * @author    Rod Roark <rod@sunsetsystems.com>
- * @author    Terry Hill <terry@lilysystems.com>
- * @author    Brady Miller <brady.g.miller@gmail.com>
- * @copyright Copyright (c) 2007-2016 Rod Roark <rod@sunsetsystems.com>
- * @copyright Copyright (c) 2015 Terry Hill <terry@lillysystems.com>
- * @copyright Copyright (c) 2017-2018 Brady Miller <brady.g.miller@gmail.com>
- * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ * package   OpenEMR
+ * link      http://www.open-emr.org
+ * author    Sherwin Gaddis <sherwingaddis@gmail.com>
+ * author-AI Gemini, Cascade, and ChatGPT
+ * All rights reserved
+ * Copyright (c) 2025.
  */
 
-require_once("../globals.php");
-require_once("$srcdir/forms.inc.php");
-require_once("$srcdir/patient.inc.php");
-require_once "$srcdir/options.inc.php";
+require_once dirname(__DIR__) . '/globals.php'; // Include globals.php
 
 use OpenEMR\Common\Acl\AclMain;
-use OpenEMR\Billing\BillingUtilities;
 use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Reports\Encounter\EncounterReportData;
+use OpenEMR\Reports\Encounter\EncounterReportFormatter;
+use OpenEMR\Reports\Encounter\EncounterReportFormHandler;
 use OpenEMR\Common\Twig\TwigContainer;
-use OpenEMR\Core\Header;
+use Twig\Error\LoaderError;
+use Twig\Error\RuntimeError;
+use Twig\Error\SyntaxError;
+
 
 if (!AclMain::aclCheckCore('encounters', 'coding_a')) {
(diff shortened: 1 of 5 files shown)


juggernautsei:
This is the new standard for reports in the system. I am going to dedicate time to get through all of the base reports this year, tackling the  Aging report by the end of the year.

juggernautsei:
@kojiromike @adunsulag now ready for inclusion in version 7.0.4 setting the ground work for the revision of all the reports to this style if approved

WARP.md
```diff
@@ -0,0 +1,490 @@
...
+npm run dev-build
+
+# Watch for changes
+npm run gulp-watch
+```
+
+### Theme Compilation
+```bash
+# Build themes in Docker
+docker exec -i $(docker ps | grep _openemr | cut -f 1 -d " ") sh -c '/root/devtools build-themes'
```
kojiromike, line 64: You can give it the actual `docker compose` commands. (See the -devops readme.)

juggernautsei: I don't know why I would want to do docker commands. The warp terminal uses the docker hub in some way. 
The docker hub has to be installed to configure the MCP. I am still trying to find time to learn how to use MCP.

image

WARP.md
kojiromike, line 1: Aha! Please put fixes #9309 on the PR description

interface/reports/encounters_report.php
```diff
@@ -1,548 +1,205 @@
...
+use OpenEMR\Reports\Encounter\EncounterReportFormHandler;
 use OpenEMR\Common\Twig\TwigContainer;
-use OpenEMR\Core\Header;
+use Twig\Error\LoaderError;
+use Twig\Error\RuntimeError;
+use Twig\Error\SyntaxError;
+
 
 if (!AclMain::aclCheckCore('encounters', 'coding_a')) {
     echo (new TwigContainer(null, $GLOBALS['kernel']))->getTwig()->render('core/unauthorized.html.twig', ['pageTitle' => xl("Encounters Report")]);
```
kojiromike: One way to get a GlobalsBag instance is to do `$globalsBag = require_once dirname(__DIR__) . '/globals.php';` above.

juggernautsei: @kojiromike, what is the point in doing that?

juggernautsei:
When I asked copilot to explain the error. Below is the explanation for the error.  This seems to be out of scope for a report. Why should I have to fix a Docker issue that has nothing to do with my commit?

- name: Install required runtime deps
  run: |
    set -euo pipefail
    sudo apt-get update -y

    # Basic utilities
    sudo apt-get install -y jq curl unzip apt-transport-https ca-certificates gnupg lsb-release

    # Ensure docker CLI exists (GitHub runners normally have docker; keep as a check)
    if ! command -v docker >/dev/null 2>&1; then
      echo "docker not found — installing docker CLI"
      sudo apt-get install -y docker.io
    fi

    # Ensure docker compose is available (either `docker compose` plugin or docker-compose binary)
    if ! docker compose version >/dev/null 2>&1 && ! command -v docker-compose >/dev/null 2>&1; then
      echo "Installing docker-compose binary"
      curl -sSL "https://github.com/docker/compose/releases/latest/download/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
      sudo chmod +x /usr/local/bin/docker-compose
    fi

    # Optional extra deps seen commonly required by CI scripts
    sudo apt-get install -y git make

    # confirm installed versions (debug-friendly output)
    echo "jq: $(jq --version 2>/dev/null || true)"
    echo "docker: $(docker --version 2>/dev/null || true)"
    if command -v docker-compose >/dev/null 2>&1; then
      echo "docker-compose: $(docker-compose --version)"
    else
      echo "docker compose plugin: $(docker compose version || true)"
    fi

interface/reports/encounters_report.php
kojiromike: If you don't have a globalsBag, you'll continue to get phpstan errors about using $GLOBALS directly instead of globalsBag.

juggernautsei: So, there has been a rule implemented in the code base that no global is to be accessed directly? All Globals must be in the globalsBag variable? Is this a step in the direction of moving the system to a DI container for the configurations?

juggernautsei: @kojiromike don't answer. I answered my own question by finding the .phpstan folder.

kojiromike:
@juggernautsei - I noticed you also have #9370 "Refactoring & Modernizing Encounter report" which was opened in November. Is this PR superseded by that one, or are they tackling different aspects of the encounter report?

If #9370 replaces this work, we should close this one to keep the PR list tidy.

PR #8306, opened 2025-04-24
correct checking of memory_limit in setup.php

kojiromike:

Corrects setup.php's comparison of `memory_limit` so it doesn't flag inappropriate values.

- Update `memory_limit` from 256M to 512M to be consistent with `Documentation/INSTALL`
- Convert the ini value to M before the comparison, to avoid erroneously flagging 1G as too little, or 1024K as enough.

No

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -1492,7 +1492,26 @@ STP4TOP;
                     $post_max_size = ini_get('post_max_size');
                     $post_max_size_style = $post_max_size < 30 ? 'text-danger' : '';
                     $memory_limit = ini_get('memory_limit');
-                    $memory_limit_style = $memory_limit < 256 ? 'text-danger' : '';
+                    // Convert from shorthand to M
+                    if (preg_match('/^(\d+)([KMG]?)$/', $memory_limit, $matches)) {
+                        switch ($matches[2]) {
+                            case 'K':
+                                $memory_limit_mb = intdiv($matches[1], 1024);
+                                break;
+                            case 'M':
+                                $memory_limit_mb = (int)$matches[1];
+                                break;
+                            case 'G':
+                                $memory_limit_mb = $matches[1] * 1024;
+                                break;
+                            case '':
+                                $memory_limit_mb = intdiv($matches[1], 1024 * 1024);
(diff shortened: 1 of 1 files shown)


bradymiller:
Not sure why, but we are using memory_limit of 256M on the dockers.

@adunsulag and @stephenwaite , I faintly remember a long time that we either bumped up memory_limit to 512M or that we bumped it back down to 256M (ie. I don't recall which one is right, the docker settings or the settings recommended on wiki and in INSTALL documentation). My memory no good :)
If we can't recall, then prob makes sense to bump up the docker memory.

setup.php
```diff
@@ -1492,7 +1492,26 @@ function cloneClicked() {
...
+                                $memory_limit_mb = $matches[1] * 1024;
+                                break;
+                            case '':
+                                $memory_limit_mb = intdiv($matches[1], 1024 * 1024);
+                                break;
+                        }
+                    } else {
+                        $memory_limit_mb = intdiv($memory_limit, 1024 * 1024);
+                    }
+                    $memory_limit_style = $memory_limit_mb < 512 ? 'text-danger' : '';
```
bradymiller, line 1514: 2DV

bradymiller:
fyi @kojiromike , 1-2 failed 'Test' ci runs are generally false positives

adunsulag:
@bradymiller we bumped it up to 512M from 256M.  We ran into a memory limit use case I believe with CCDA generation and bundling of documents in order to make it through certification.  For almost all use cases we are under the 256M but CCDA just wouldn't work at that threshold.  Maybe w/ our CCDA changes this year we'll be able to bump it back down.  Will re-run the CI.

adunsulag:
@kojiromike we're just missing a github issue to tag this PR to.  Format should be "bug: [issue-name]" for the title of the issue.  Once that's in, edit the description of the PR here to include the issue "fixes [pound-with-issue-number-with]" and we can then bring this in.

kojiromike:
@adunsulag done, thanks for the guidance.

bradymiller:

PR #8316, opened 2025-04-28
Fixing UI For Demographics Form On Big Screens

### LINKED ISSUE #8315: Edit Current Patient label and buttons go to the side when on big display
### Describe the bug
If you open the demographics form in a display with a larger than 2560x1440 resolution, the form will look like this:

Image

The "Edit Current Patient" label is supposed to go above the form, and the buttons are supposed to be either above or under the form depending on settings. 

What makes this worse is that the label and buttons now vertically align with the middle of the form, so when you switch tabs on the form, the label and buttons move accordingly:

Image

### To Reproduce

1. Go to Patient -> New/Search on top nav bar and open a patient's dashboard
2. Click on the pencil icon in Demographics to access the demographics form
3. Do 'ctrl' + '-' or 'cmd' + '-' if on mac a couple times, and you should see the label and buttons move to the left
4. Switch tabs on the form to see the label and buttons move

### Client configuration

- Browser: All
- OpenEMR version: All
- Operating system: All

bryanopenemr:
https://github.com/openemr/openemr/issues/8315

Fixes #8315

The label and buttons will now stay at where they should be regardless of display size:

Screenshot 2025-04-28 at 4 48 21 PM

Changed container-xl to container-fluid so the label, buttons, and demographics form each take up the entire width of the viewport

No

### DIFF
--- a/interface/patient_file/summary/demographics_full.php
+++ b/interface/patient_file/summary/demographics_full.php
@@ -417,48 +417,41 @@ $constraints = LBF_Validation::generate_validate_constraints("DEM");
 <input type='hidden' name='db_id' value="<?php echo attr($result['id']); ?>" />
 <input type="hidden" name="isSwapClicked" value="" />
 
-    <div class="container-fluid">
-        <div class="row">
-            <?php if (FormActionBarSettings::shouldDisplayTopActionBar()) { ?>
-            <div class="col-12">
-                <h2><?php echo xlt('Edit Current Patient');?></h2>
+<div class="container-xl">
+    <div class="row">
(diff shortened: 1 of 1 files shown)


adunsulag:
@bryanopenemr so this fixes the broken behavior, but I think the intended desire was to constrain the container so its not stretched all the way across.  W/o pulling this down how easy is it to navigate through the fields at this point?  I would think it'd be better to fix the centering of the buttons  & form rather than making it stretch which is harder from a usability standpoint.  (Right now its not usable at all so this is at least an improvement).

PR #8327, opened 2025-05-01
change information pick location

### LINKED ISSUE #8321: Patient Statement Email
### Describe the bug
When sending email statement to patient.
All of the statements go to all of the patients. 

### To Reproduce

Image

[Gmail - ROSE ANESTHESIA PLLC Patient Statement Bill.pdf](https://github.com/user-attachments/files/19963720/Gmail.-.ROSE.ANESTHESIA.PLLC.Patient.Statement.Bill.pdf)

### Expected behavior
Expecting each of the email to go to the individual patients.

### Client configuration

- Browser: Firefox

Image

juggernautsei:
Fixes #8321

A summary pdf of all patients is sent to every patient which is a breach. 

Send individual patient statements

### DIFF
--- a/interface/billing/sl_eob_search.php
+++ b/interface/billing/sl_eob_search.php
@@ -1,4 +1,4 @@
-<?php
+<?php global $STMT_PRINT_CMD;
 
 /**
  * This the first of two pages to support posting of EOBs.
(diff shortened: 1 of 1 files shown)


interface/billing/sl_eob_search.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php global $STMT_PRINT_CMD;
```
kojiromike: This `global $STMT_PRINT_CMD` on line 1 looks like an accidental addition — it's at file scope outside any function, so it has no effect here. Can you revert this line back to just `<?php`?

kojiromike (commented):
The fix logic looks right — moving email dispatch inside the per-patient loop is the correct approach for #8321.

PR #8328, opened 2025-05-02
Telemetry portal tracking events

### LINKED ISSUE #8332: Telemetry disable logic and add API track events.
### Describe the bug
Fix telemetry disable to allow JS reporting to be enabled correctly.
Adding new telemetry data collection for API requests. ['API', action, normalized path, scopes]
### Client configuration

- Browser: all
- OpenEMR version: 7.0.4 dev
- Operating system: all

sjpadgett:
add portal dashboard card click events for tracking 
fix allowTelemetry logic

Fixes #8332

#### Does your code include anything generated by an AI Engine?  No

### DIFF
--- a/apis/dispatch.php
+++ b/apis/dispatch.php
@@ -155,7 +155,7 @@ if (!empty($tokenId)) {
 
 
 // recollect this so the DEBUG global can be used if set
-    $logger = new SystemLogger();
+$logger = new SystemLogger();
 
 $gbl::$apisBaseFullUrl = $GLOBALS['site_addr_oath'] . $GLOBALS['webroot'] . "/apis/" . $gbl::$SITE;
 $restRequest->setApiBaseFullUrl($gbl::$apisBaseFullUrl);
@@ -267,7 +267,7 @@ if ($isLocalApi) {
     if ($userRole == 'users') {
         $_SESSION['authUser'] = $user["username"] ?? null;
         $_SESSION['authUserID'] = $user["id"] ?? null;
-        $_SESSION['authProvider'] =  sqlQueryNoLog("SELECT `name` FROM `groups` WHERE `user` = ?", [$_SESSION['authUser']])['name'] ?? null;
+        $_SESSION['authProvider'] = sqlQueryNoLog("SELECT `name` FROM `groups` WHERE `user` = ?", [$_SESSION['authUser']])['name'] ?? null;
         if (empty($_SESSION['authUser']) || empty($_SESSION['authUserID']) || empty($_SESSION['authProvider'])) {
             // this should never happen
             $logger->error("OpenEMR Error: api failed because unable to set critical users session variables");
@@ -297,7 +297,7 @@ if ($isLocalApi) {
         $restRequest->setPatientRequest(true);
         $restRequest->setPatientUuidString($puuidStringCheck);
         $logger->debug("dispatch.php request setup for patient role", ['patient' => $puuidStringCheck]);
-    } else if ($userRole === 'system') {
+    } elseif ($userRole === 'system') {
         $_SESSION['authUser'] = $user["username"] ?? null;
         $_SESSION['authUserID'] = $user["id"] ?? null;
         if (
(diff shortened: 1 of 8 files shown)


sjpadgett:
I'm still testing and I need to be sure I included all my changed code in the commits. Don't trust myself anymore! :)
new events are labeled patientPortal and includes the window title, target id and href.

sjpadgett:
I'm happy with the project. It has evolved into pretty close to what we planned. What makes me the proudest is how well we all came together as a team without any serious bodily harm! lol
Attached is a json of what we report monthly.
Thank you everyone for all the help and yea, we finally got it done...
Let me know if any addition are needed

[telemetry_payload.json](https://github.com/user-attachments/files/20025629/telemetry_payload.json)

sjpadgett:
merging so on dev demos tomorrow.

portal/lib/track_portal_events.php
```diff
@@ -0,0 +1,75 @@
...
+    $telemetryRepo = new TelemetryRepository();
+    $versionService = new VersionService();
+    $telemetryService = new TelemetryService($telemetryRepo, $versionService);
+
+    $action = $data['action'] ?? '';
+    switch ($action) {
+        case 'portalCardClickData':
+            $telemetryService->reportClickEvent($data);
+            break;
+        case 'allUsageData':
```
bradymiller, line 64: I can't see where this is being used in the code. If not needed, wouldn't include it, since then giving access to a patient to view all usage data (and probably worse, have it truncated).

bradymiller: actually, looks like they won't be able to view since it just kicks off the report to the reporting engine. Still wouldn't allow patient access to do this since would mess up the reporting.

sjpadgett: yep, didn't need it. I think I left because I may have needed it to kick off reporting.
I'll remove.

src/Services/ProductRegistrationService.php
```diff
@@ -56,14 +56,14 @@ public function getProductStatus(): array
...
-                }
-                if ($telemetry_disabled == null) {
-                    $row['allowTelemetry'] = 1;
-                }
+            }
+            if ($optOut == null) {
+                $row['allowEmail'] = 1;
+            }
+            if (empty($telemetry_disabled)) {
+                $row['allowTelemetry'] = 1;
```
bradymiller, line 66: This will make allowEmail and allowTelemetry on the dialog no longer completely decoupled. In this case, if, optout for email is null, then the dialog will show for telemetry even if it has been opted out. I'll post a fix for this. Kicking myself that i should of made these variables more clear that only for whether to display in the dialog.

sjpadgett: I tested this and was working for me. Although I may have missed an update from just being tired. I'll test again but rename at will. I wished I had named telemetry_disable to telemetry enabled. That bit toggle just got me confused.

sjpadgett: btw telemetry_disabled is decoupled and L-65 is needed or reporting won't work i.e. 0 and or null.

sjpadgett: What am I missing Brady
image


image

sjpadgett: yep doesn't work if email is null and opt-out = 0 but I don't see why it'd get in that condition because null was the neutral state but maybe you can think of another event to add while there.:)

apis/dispatch.php
```diff
@@ -407,3 +408,17 @@
     $logger->debug("dispatch.php no route found for resource", ['resource' => $resource]);
     http_response_code(404);
 }
+
+// Report to Telemetry
+// Be nice to find a centralized place to catch failed requests.
```
bradymiller, line 413: I pull this stuff for the realworldtesting from the api log and log. (can get success, fails, from patient, from users, and resource) I could probably work this into the apiLog() function if you want (ie. put it into your tracking table in apiLog, then can record fails).

bradymiller: actually, not seeing fails being logged there. will research a bit more, but may be best to keep here if can't get centralized access to fails in apiLog().

bradymiller: Goodness, if above is true, then this means the Reals World Testing apis will never fail :)

sjpadgett:
Almost done with a sample API client to donate for our users to have a starting place. Kind of cool and I'll give to you guys once done. @adunsulag already has a super nice one he wrote but I hoped mine, while not full blown, would helpful to some.

image

bradymiller:

bradymiller:

Issue #8333, opened 2025-05-04
When updating any class that extends ORDataObject (e.g., Insurance Company), the record is deleted due to REPLACE INTO, causing the UUID to be regenerated


modanq:
### Describe the bug

Saving changes to an `ORDataObject` subclass (for example, `InsuranceCompany`) uses `REPLACE INTO`, which deletes the existing row and then inserts a new one. As a result, the original `uuid` is destroyed and a fresh UUID is generated on each save. This has previously been brought up by **adunsulag** in [Issue 4547](https://github.com/openemr/openemr/issues/4547#issuecomment-887951327).

### To Reproduce

1. Identify or create an `insurance_companies` record.
2. Execute this statement: `SELECT id, HEX(uuid) AS uuid, name, cms_id, ins_type_code FROM insurance_companies;` in MariaDB client _OR_ `SELECT id, BIN_TO_UUID(uuid) AS uuid, name, cms_id, ins_type_code FROM insurance_companies;` in MySQL client
3. Note the value of `uuid`.
4. Update that same record (via the UI) by changing any field (e.g., Name, Payer ID, or Payer Type).
5. Re-run the SELECT above and observe that `uuid` has changed.
6. Additionally, new rows are created in the `uuid_registry` table

<img width="872" alt="Image" src="https://github.com/user-attachments/assets/d8bac2a5-32a5-44a3-a86e-d157a7402130" />

**MAKE CHANGES TO INSURANCE COMPANY ON UI**

<img width="341" alt="Image" src="https://github.com/user-attachments/assets/42a19992-4aa8-402a-baf2-f279af7d7eeb" />

**THE UUID CHANGES**

<img width="869" alt="Image" src="https://github.com/user-attachments/assets/3f51da8a-d635-4c5a-be5f-79f3112ac995" />

### Expected behavior

The update operation should preserve the row’s original `uuid`. Only the modified fields should change; there should be no delete/reinsert cycle. I have an upcoming PR that proposes switching the `REPLACE INTO` with `INSERT … ON DUPLICATE KEY UPDATE` to address this issue.

### Client configuration

- Browser: Chrome 135.0.7049.116
- OpenEMR version: 7.0.4
- Operating system: macOS

modanq:
https://github.com/openemr/openemr/pull/8336

PR #8336, opened 2025-05-05
Switch REPLACE INTO with INSERT … ON DUPLICATE KEY UPDATE

modanq:
- Swap out `REPLACE INTO` (which issues DELETE+INSERT) for `INSERT … ON DUPLICATE KEY UPDATE` (in-place UPSERT)
- Introduce `$setClause` accumulator to build the column-value list once and replace manual `substr` trimming with `rtrim()`

Fixes [8333](https://github.com/openemr/openemr/issues/8333)

This patch fixes the unintended deletion/re-insertion cycle in all `ORDataObject` subclasses (e.g. InsuranceCompany) by replacing the `REPLACE INTO` logic with an in-place upsert (`INSERT … ON DUPLICATE KEY UPDATE`), thereby preserving the original `uuid` on every save.

- Swap out the existing `REPLACE INTO` statement for an `INSERT … ON DUPLICATE KEY UPDATE` in the `ORDataObject::save()` method.
- Introduce a `$setClause` string accumulator to build the `column = value` list once, rather than concatenating and then trimming with `substr`.

`src/Common/ORDataObject/ORDataObject.php` used *ChatGPT o4-mini-high*

### DIFF
--- a/src/Common/ORDataObject/ORDataObject.php
+++ b/src/Common/ORDataObject/ORDataObject.php
@@ -56,12 +56,12 @@ class ORDataObject
             return true;
         }
 
-        // NOTE: REPLACE INTO does a DELETE and then INSERT, if you have foreign keys setup the delete call will trigger
-        $sql = "REPLACE INTO " . $this->_prefix . $this->_table . " SET ";
+		$sql = "INSERT INTO " . $this->_prefix . $this->_table . " SET ";
         //echo "<br /><br />";
         $fields = QueryUtils::listTableFields($this->_table);
         $db = get_db();
         $pkeys = $db->MetaPrimaryKeys($this->_table);
+		$setClause = "";
 
         foreach ($fields as $field) {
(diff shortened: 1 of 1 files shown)


bradymiller:
hi @modanq . Thanks for reporting this and thanks for the PR! Since critical bug, will expedite this code review and plan to have the fix in next patch 7.0.3.4  (to likely be released next week).

adunsulag:
Pulling this down and doing testing.  I suspect that the history tables are going to have issues with this and will need some additional fixes as I traced that down as one of the main reasons this pattern was implemented.

adunsulag:
History tables were fine, realized those use their own sql mechanism and don't use ORData.  Tested all the sub classes of ORData and the data populates fine for everything (Document, Note, InsuranceCompany, InsuranceNumber, Prescription, FormVitals, FormVitalDetails, etc).

bradymiller:

PR #8345, opened 2025-05-09
fix(setup.php): use unchanged values in config check

### LINKED ISSUE #8344: setup.php display_errors and max_execution_time checks will always show fixed values
### Describe the bug

setup.php [sets display_errors off](https://github.com/openemr/openemr/blob/b761eacc6596b6802e52aa01645077e555970f17/setup.php#L41) early in the script, so when [it checks it later](https://github.com/openemr/openemr/blob/b761eacc6596b6802e52aa01645077e555970f17/setup.php#L1482) it can't be checking the legitimate value.

### To Reproduce

set display_errors on in php.ini, go through setup.php, note that it shows as "Off" and displays no warning.

### Expected behavior

display_errors should be correctly reported during setup.php

kojiromike:
Fixes #8344

Ensures that the values checked in the config check in setup.php aren't just the ones that were set earlier in the same file.

Capture the original ini values before setting them.

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -36,7 +36,10 @@ if ($response !== true) {
     die(htmlspecialchars($response));
 }
 
-// Set the maximum excution time and time limit to unlimited.
+// Capture the maximum execution time and display errors setting before changing them.
+$original_ini_max_execution_time = ini_get('max_execution_time');
+$original_ini_display_errors = ini_get('display_errors');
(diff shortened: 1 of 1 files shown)


stephenwaite (approved):
nice!

bradymiller:
Brought in rel-703 for patch 7.0.3.4: #8356

PR #8364, opened 2025-05-14
refactor(sql_patch.php,sql_upgrade.php): use insert_globals from Installer

### LINKED ISSUE #8363: Reuse the `insert_globals` function from Installer in sql_patch and sql_upgrade scripts
`sql_patch.php` and `sql_upgrade.php` have code that is redundant with a function in the Installer. It'll reduce maintenance overhead if we can eliminate duplication, and this is low-hanging fruit.

kojiromike:
Fixes #8363

Eliminates duplicate code between `sql_patch.php`, `sql_upgrade.php` and the Installer class.

### DIFF
--- a/sql_patch.php
+++ b/sql_patch.php
@@ -25,7 +25,6 @@ $ignoreAuth = true; // no login required
 
 require_once('interface/globals.php');
 require_once('library/sql_upgrade_fx.php');
-require_once('library/classes/Installer.class.php');
 
 use OpenEMR\Common\Uuid\UuidRegistry;
 use OpenEMR\Services\Utils\SQLUpgradeService;
@@ -68,8 +67,8 @@ $sqlUpgradeService = new SQLUpgradeService();
     $sqlUpgradeService->flush_echo();
 
     echo '<p style="font-weight:bold; text-align:left; color:green">',xlt('Updating global configuration defaults'),'...</p>';
-    $installer = new Installer([]);
-    $installer->insert_globals();
+    $sqlUpgradeService->flush_echo();
+    SQLUpgradeService::insertGlobals();
 
     $versionService = new VersionService();
     $currentVersion = $versionService->fetch();
--- a/sql_patch.php
+++ b/sql_patch.php
@@ -25,6 +25,7 @@ $ignoreAuth = true; // no login required
 
 require_once('interface/globals.php');
 require_once('library/sql_upgrade_fx.php');
+require_once('library/classes/Installer.class.php');
 
 use OpenEMR\Common\Uuid\UuidRegistry;
 use OpenEMR\Services\Utils\SQLUpgradeService;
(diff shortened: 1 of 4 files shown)


sql_upgrade.php
```diff
@@ -392,19 +392,9 @@ function pausePoll(othis) {
...
-                if (is_array($fldtype) || (substr($fldtype, 0, 2) !== 'm_')) {
-                    $row = sqlQuery("SELECT count(*) AS count FROM globals WHERE gl_name = '$fldid'");
-                    if (empty($row['count'])) {
-                        sqlStatement("INSERT INTO globals ( gl_name, gl_index, gl_value ) " .
-                            "VALUES ( '$fldid', '0', '$flddef' )");
-                    }
-                }
-            }
-        }
+        require_once('library/classes/Installer.class.php');
```
bradymiller, line 395: rec putting the require once near top of script below line 49

sql_upgrade.php
```diff
@@ -392,19 +392,9 @@ function pausePoll(othis) {
 
         echo "<p class='text-success'>" . xlt("Updating global configuration defaults") . "..." . "</p><br />\n";
         $skipGlobalEvent = true; //use in globals.inc.php script to skip event stuff
```
bradymiller, line 394: can also remove above line

sql_patch.php
```diff
@@ -68,19 +69,8 @@
 
     echo '<p style="font-weight:bold; text-align:left; color:green">',xlt('Updating global configuration defaults'),'...</p>';
     $skipGlobalEvent = true; //use in globals.inc.php script to skip event stuff
```
bradymiller, line 71: can also remove above line

bradymiller:
Will need to test since sql patch/upgrade scripts use the openemr sql library while the installer stuff uses native mysqli calls.

bradymiller:
For example, if the sql queries error out, they will report error in the installation, which i think is ok.

bradymiller:
Also, it appears to hard-code the collation, which could be a problem is edge cases. Another option would be to make this a static function in SQLUpgradeService class (then don't need to worry about the separate Installer function and needs which is assuming there is not always a installed database etc).

kojiromike:
@bradymiller I had two thoughts:

1. Could this not be re-written to use an upsert, like `INSERT … ON DUPLICATE KEY UPDATE`? and then…
2. We could generate SQL to put in sql/, instead of dynamically loading the globals.

This would increase flexibility for installers, reduce the complexity of the code in a few places. Basically, openemr is most of the way toward being instantly runnable without going through the currently complex entrypoint in the -devops repo:

If we stand up an instance of openemr such that…
1. sql/database.sql is loaded into the database (for example, via the /docker-entrypoint-initdb.d feature in the database)
2. sites/default/sqlconf.php has $config = 1 and the mysql configuration is available somehow (environment, perhaps)
3. loading these globals is the only remaining hurdle to installing without actively running some php. That's pretty desirable IMO, and the only php we need to run, actually just runs some sql that is (afaict) statically defined anyway.

bradymiller:
hi, Sorry I should of been more clear. Meant that Installer code will need to stay separate and then can centralize the code for the upgrade and patch scripts in SQLUpgradeService.php

kojiromike:
I may revisit this in the future…

PR #8371, opened 2025-05-18
fix: do not send usage report if telemetry is turned off

bradymiller:
fixes #8370

PR #8372, opened 2025-05-18
fix: do not send usage report if telemetry is turned off (#8371)

bradymiller:
for rel-703,  do not send usage report if telemetry is turned off

### DIFF
--- a/src/Telemetry/TelemetryService.php
+++ b/src/Telemetry/TelemetryService.php
@@ -121,6 +121,11 @@ class TelemetryService
      */
     public function reportUsageData(): int|bool
     {
+        if (empty(self::isTelemetryEnabled())) {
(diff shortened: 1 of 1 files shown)


sjpadgett:
@bradymiller The only way this should have happened is by not disabling in dialog after it was previously enabled. Like directly in database.
If it is still sending that means the background service is still running and would have been deleted if disabled by dialog entry.
So under normal operation this shouldn't be an issue but having your mod as safeguard can't hurt.

bradymiller:
@sjpadgett ,
Agreed. Figured that openemr instance admins may want to turn it off manually (since no way to do in gui interface, which sort of doing on purpose :) ) and that they may only toggle the setting in the database. Then it won't send that last batch of data and also won't keep sending the foundation database monthly empty reports :)

PR #8381, opened 2025-05-21
refactor: decouple SQLUpgradeService and SQLUpgradeEvent

### LINKED ISSUE #8380: SQLUpgradeService and SQLUpgradeEvent Circular Dependency
### Describe the bug

`OpenEMR\Events\Core\SQLUpgradeEvent` and `OpenEMR\Services\Utils\SQLUpgradeService` both have `use` statements to each other.

`SQLUpgradeService` calls methods and references stuff in `SQLUpgradeEvent`, but `SQLUpgradeEvent` only references `SQLUpgradeService` for its types. So it's safe, prima facie.

But I think it's still undesirable and confusing to read. I think this is a case where SQLUpgradeEvent should refer to types defined in an interface instead of an implementation.

kojiromike:
Fixes #8380

Resolves a circular dependency between SQLUpgradeService and SQLUpgradeEvent.
Opens the door to dependency injection.

- Create `ISQLUpgradeService`, an interface 
- Reference the interface instead of the implementation in `SQLUpgradeEvent`

### DIFF
--- a/src/Events/Core/SQLUpgradeEvent.php
+++ b/src/Events/Core/SQLUpgradeEvent.php
@@ -15,7 +15,7 @@
 
 namespace OpenEMR\Events\Core;
 
-use OpenEMR\Services\Utils\SQLUpgradeService;
+use OpenEMR\Services\Utils\Interfaces\ISQLUpgradeService;
 use Symfony\Contracts\EventDispatcher\Event;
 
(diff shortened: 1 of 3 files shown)


adunsulag (approved):
Looks good.  Thanks for the change!

PR #8384, opened 2025-05-21
chore: remove summernote-plugins

### LINKED ISSUE #8383: unpinned, unmaintained dependency on summernote-plugins repo
in package.json in the `napa` section, there's a reference to a summernote-nugget that comes directly from the github repo https://github.com/pHAlkaline/summernote-plugins. This causes npm install (via napa) to require git, which makes the build take longer. Also, that repo hasn't had a commit for 8 years, and it has never been tagged or released, so it is a risk in that regard.

I haven't been able to figure out how the summernote-plugins repo code is being used.

I think, in the immediate, we can change the git url to an archive url, pinned to the particular commit, and that will resolve the better part of the issue.

kojiromike:
Fixes #8383

Makes it a bit easier and faster to npm install, and prevents unintended consequences from upstream updating their repo.

- removes the reference to summernotes-plugins, which is obsolete.

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,6 @@
     "jquery-panelslider": "https://github.com/eduardomb/jquery-panelslider/archive/1.0.0.tar.gz",
     "jquery-ui": "https://jqueryui.com/resources/download/jquery-ui-1.12.1.zip",
     "jquery-ui-themes": "https://jqueryui.com/resources/download/jquery-ui-themes-1.12.1.zip",
-    "summernote-nugget": "https://github.com/pHAlkaline/summernote-plugins",
     "literallycanvas": "https://github.com/literallycanvas/literallycanvas/archive/v0.4.14.tar.gz",
     "react": "https://github.com/facebook/react/releases/download/v15.1.0/react-15.1.0.zip",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -120,7 +120,7 @@
     "jquery-panelslider": "https://github.com/eduardomb/jquery-panelslider/archive/1.0.0.tar.gz",
     "jquery-ui": "https://jqueryui.com/resources/download/jquery-ui-1.12.1.zip",
     "jquery-ui-themes": "https://jqueryui.com/resources/download/jquery-ui-themes-1.12.1.zip",
-    "summernote-nugget": "https://github.com/pHAlkaline/summernote-plugins",
+    "summernote-nugget": "https://github.com/pHAlkaline/summernote-plugins/archive/90b558b004f83aafac51f99ad51b0aabcac72491.zip",
```
sjpadgett, line 123: You may remove if want

sjpadgett (approved):

bradymiller:
hi @adunsulag ,
Dependency work/changes generally can't get into patches, so I removed the `Status: Needs Release Merge` label on this and will label the issue for 7.0.4
(just making sure this is ok and I am correct in that the `Status: Needs Release Merge` means to bring into a rel-* branch for a future patch)

Issue #8244, opened 2025-04-09
bug: Broken edit button for questionnaires


bban160:
### Describe the bug

Attempting to edit a questionnaire from the encounter summary does not work. The questionnaire must be deleted and recreated.

### To Reproduce

1. Start and save a questionnaire in an encounter.
2. From the encounter summary, click the Edit button for the questionnaire.
3. Receive a blank page and 500 Internal Server Error.

### Technical Details

```
PHP Fatal error:  Uncaught Error: Call to undefined function formFetch() in /var/www/localhost/htdocs/openemr/interface/forms/questionnaire_assessments/questionnaire_assessments.php:71
Stack trace:
#0 /var/www/localhost/htdocs/openemr/interface/forms/questionnaire_assessments/view.php(14): require()
#1 /var/www/localhost/htdocs/openemr/interface/patient_file/encounter/view_form.php(50): require_once('...')
#2 {main}
  thrown in /var/www/localhost/htdocs/openemr/interface/forms/questionnaire_assessments/questionnaire_assessments.php on line 71
```

Editing [interface/forms/questionnaire_assessments/questionnaire_assessments.php](https://github.com/openemr/openemr/blob/2708ff6f2ff451ac8a28179c244a03630df834b0/interface/forms/questionnaire_assessments/questionnaire_assessments.php) to add `require_once($GLOBALS["srcdir"] . "/api.inc.php");` seems to resolve the issue.

---

OpenEMR version: 7.0.3 (3)

bryanopenemr:
I wasn't able to reproduce this, is this perhaps already fixed?

adunsulag:
@bban160 are you using a custom questionnaire, or one of the standard LOINC assessments that you imported.  Looks like the api.inc.php missing library bug has been there for a long time (at least since 2022).

```
        if (($_REQUEST["formOrigin"] ?? null) == 1 && empty($isPortal)) {
            $form = sqlQuery("SELECT * FROM `form_questionnaire_assessments` Where id=? And activity = 1 Limit 1", [$formid]);
        } else {
            $form = formFetch("form_questionnaire_assessments", $formid);
        }
```
What I don't understand is how its just now being triggered.

adunsulag:
@bban160 I'm also unable to reproduce the bug following the steps you followed, I can see how the code could break, but I'm trying to figure out how you are triggering it.  Are you doing this actually from the portal instead of the encounter screen?  Are you logged into the portal at the same time you are logged in as a clinician?

PR #8394, opened 2025-05-28
fix(questionnaires): add missing api.inc.php require

bban160:
Fixes https://github.com/openemr/openemr/issues/8244

Under certain circumstances (not yet identified at the time of this PR), the edit button for questionnaire responses returns a 500 Internal Server Error and the following PHP error:

```
PHP Fatal error:  Uncaught Error: Call to undefined function formFetch() in /var/www/localhost/htdocs/openemr/interface/forms/questionnaire_assessments/questionnaire_assessments.php:71
Stack trace:
#0 /var/www/localhost/htdocs/openemr/interface/forms/questionnaire_assessments/view.php(14): require()
#1 /var/www/localhost/htdocs/openemr/interface/patient_file/encounter/view_form.php(50): require_once('...')
#2 {main}
  thrown in /var/www/localhost/htdocs/openemr/interface/forms/questionnaire_assessments/questionnaire_assessments.php on line 71
```

Explicitly include api.inc.php in questionnaire_assessments.php to guarantee `formFetch()` is always defined.

No

### DIFF
--- a/interface/forms/questionnaire_assessments/questionnaire_assessments.php
+++ b/interface/forms/questionnaire_assessments/questionnaire_assessments.php
@@ -19,6 +19,7 @@ use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Forms\CoreFormToPortalUtility;
 use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Core\Header;
+use OpenEMR\Core\OEGlobalsBag;
 use OpenEMR\Services\QuestionnaireResponseService;
 use OpenEMR\Services\QuestionnaireService;
 
@@ -32,6 +33,7 @@ if ($isPortal) {
 $patientPortalOther = CoreFormToPortalUtility::isPatientPortalOther($_GET);
 
 require_once(__DIR__ . "/../../globals.php");
+require_once(OEGlobalsBag::getInstance()->getString('srcdir') . "/api.inc.php");
 require_once("$srcdir/user.inc.php");
 // used for form generation utilities
 require_once("$srcdir/options.inc.php");


kojiromike:
Hi @bban160 - this looks like a valid fix. The error is clear and the one-line fix (adding the missing `require_once` for `api.inc.php`) makes sense.

To get this merged:

1. **PR title format** - Should follow [Conventional Commits](https://www.conventionalcommits.org/), e.g., `fix(questionnaires): add missing api.inc.php require`

2. **Rebase needed** - Will need to be rebased onto current master.

See [CONTRIBUTING.md](https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md) for guidelines.

Are you still available to update this? It's a small fix that would be easy to complete.

kojiromike:
## Analysis

I traced through the code to understand why this is intermittent:

**The bug**: In `questionnaire_assessments.php`, line 73 calls `formFetch()`:

```php
if (($_REQUEST["formOrigin"] ?? null) == 1 && empty($isPortal)) {
    $form = sqlQuery("SELECT * FROM ...", [$formid]);  // Dashboard path
} else {
    $form = formFetch("form_questionnaire_assessments", $formid);  // Normal edit path
}
```

But `formFetch()` is defined in `api.inc.php`, which is **not explicitly included**. The file only includes `globals.php`, `user.inc.php`, and `options.inc.php`.

**Why it's intermittent**: The bug only triggers when:
1. Editing from the **encounter screen** (not Dashboard, where `formOrigin=1` bypasses `formFetch()` entirely)
2. AND `api.inc.php` wasn't already loaded as a side-effect through another include chain

In most flows, `api.inc.php` gets pulled in transitively, so it works. But in certain session states or entry points, it's missing.

**The fix is correct**: Adding `require_once($GLOBALS["srcdir"] . "/api.inc.php")` ensures `formFetch()` is always available regardless of entry point. This is proper defensive coding - don't rely on transitive includes.

This is a valid, low-risk fix. 👍

interface/forms/questionnaire_assessments/questionnaire_assessments.php
sjpadgett, line 40: Since a global is used in path probably best to load after the globals.php is loaded.
Move require_once($GLOBALS["srcdir"] . "/api.inc.php"); L-37

kojiromike:
Thanks for the fix, @bban160! I've rebased to resolve the conflicts and implemented @sjpadgett's suggestion to move the require after `globals.php` is loaded. Will merge once the checks pass.

PR #8401, opened 2025-05-29
ci(test action): test matrix

kojiromike:
Simplify the code for running the tests in github actions by using a matrix to find all the docker-compose files instead of hand-coding each one.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,1213 +10,175 @@ permissions:
   contents: read
 
 jobs:
-
-  build_apache_82_114:
-    name: PHP 8.2 - Apache - MariaDB 11.4
+  collect:
     runs-on: ubuntu-22.04
-    env:
-      DOCKER_DIR: apache_82_114
-      OPENEMR_DIR: /var/www/localhost/htdocs/openemr
-      CHROMIUM_INSTALL: "apk update; apk add --no-cache chromium chromium-chromedriver; export PANTHER_CHROME_DRIVER_BINARY=/usr/lib/chromium/chromedriver"
     steps:
-      - uses: actions/checkout@v4
-
-      - name: Install PHP
-        uses: shivammathur/setup-php@v2
-        with:
-          php-version: '8.2'
-
-      - name: Report PHP Version
-        run: php -v
-
-      - name: Install npm package
-        uses: actions/setup-node@v4
-        with:
-          node-version: '20'
-
-      - name: Main build
-        run: |
-          source ci/ciLibrary.source
-          composer_github_auth
-          main_build
-
-      - name: CCDA build
-        run: |
-          source ci/ciLibrary.source
-          ccda_build
-
-      - name: Dockers environment start
-        run: |
-          source ci/ciLibrary.source
-          dockers_env_start
-          sleep 60
-
-      - name: Install and configure
-        run: |
-          source ci/ciLibrary.source
-          install_configure
-
-      - name: Unit testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_unit
-        if: ${{ success() || failure() }}
-
-      - name: E2e testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_e2e
-        if: ${{ success() || failure() }}
-
-      - name: Api testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_api
-        if: ${{ success() || failure() }}
-
-      - name: Fixtures testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_fixtures
-        if: ${{ success() || failure() }}
-
-      - name: Services testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_services
-        if: ${{ success() || failure() }}
-
-      - name: Validators testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_validators
-        if: ${{ success() || failure() }}
-
-      - name: Controllers testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_controllers
-        if: ${{ success() || failure() }}
-
-      - name: Common testing
-        run: |
-          source ci/ciLibrary.source
-          build_test_common
-        if: ${{ success() || failure() }}
-
-  build_apache_83_116:
-    name: PHP 8.3 - Apache - MariaDB 11.6 (Rolling version)
+    - uses: actions/checkout@v4
+    - name: Collect Docker Dirs
+      id: docker-dirs
+      ##
+      # Create a JSON array of the ci subdirectories that contain
+      # docker-compose.yml files. We'll use that output as a test matrix
+      # in github actions.
+      run: |
+        shopt -s nullglob
+        dirs=( ci/*/docker-compose.yml )
+        dirs=( "${dirs[@]%/docker-compose.yml}" )
+        dirs=( "${dirs[@]#ci/}" )
+        {
+          printf 'docker_dirs='
+          printf '%s\n' "${dirs[@]}" | jq -R . | jq -sc .
+        } >> "$GITHUB_OUTPUT"
+    outputs:
+      docker-dirs: ${{ steps.docker-dirs.outputs.docker_dirs }}
+  build:
     runs-on: ubuntu-22.04
+    needs: collect
+    strategy:
+      fail-fast: false
+      matrix:
+        docker_dir: ${{ fromJson(needs.collect.outputs.docker-dirs) }}
     env:
-      DOCKER_DIR: apache_83_116
-      OPENEMR_DIR: /var/www/localhost/htdocs/openemr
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -10,1213 +10,175 @@ permissions:
...
+        docker_dir: ${{ fromJson(needs.collect.outputs.docker-dirs) }}
     env:
-      DOCKER_DIR: apache_83_84
-      OPENEMR_DIR: /var/www/localhost/htdocs/openemr
-      CHROMIUM_INSTALL: "apk update; apk add --no-cache chromium chromium-chromedriver; export PANTHER_CHROME_DRIVER_BINARY=/usr/lib/chromium/chromedriver"
+      DOCKER_DIR: ${{ matrix.docker_dir }}
+      # OPENEMR_DIR: /var/www/localhost/htdocs/openemr
+      # CHROMIUM_INSTALL: "apt-get update; apt-get install -y chromium-driver; export PANTHER_CHROME_DRIVER_BINARY=/usr/bin/chromedriver"
+      # CHROMIUM_INSTALL: "apk update; apk add --no-cache chromium chromium-chromedriver; export PANTHER_CHROME_DRIVER_BINARY=/usr/lib/chromium/chromedriver"
+      # CHROMIUM_INSTALL: "apk update; apk add --no-cache chromium chromium-chromedriver; export PANTHER_CHROME_DRIVER_BINARY=/usr/lib/chromium/chromedriver"
```
kojiromike, line 46:
```suggestion
```

bradymiller:
hi @kojiromike ,
Some more neat stuff!
Couple things:
1. Is there a way to get readable titles in here (ie. somehow stuffing them in docker-compose maybe)
2. Sometime we need to turn off a test for a specific env (ie. sometime the e2e decides to entirely crap out on entire PHP versions, so then comment out the testing for awhile) (suppose would somehow need to stuff this in docker-compose)
3. Looks like an extra test called `collect`
4. Would need to deal with the sleep 60 issue (ie. only in the non-mariadb runs)

bradymiller:
hi @kojiromike ,

Awesome stuff!

There should only be 21 jobs, which is why i knew there was an extra one. Otherwise wouldn't of noticed it. It is `[Test / collect (pull_request)` on the output and appears to just be an empty Test job.

I see how we can simply turn off all testing for an environment via ci/foo/docker-compose.disabled.yml, which is helpful. The only tests that every really breaks for environments is the e2e testing. For example, right now the 2 tests are failing in 2 of the nginx environments because of systemic breaks in e2e which will likely fix itself via upstream at some point in future. In the past, I generally just commented out the e2e testing for these environments until they started working again. A very cool mechanism would be to have option of changing ci/foo/docker-compose.yml script to ci/foo/docker-compose.e2e.disabled.yml, which then still runs the environment, but skips the e2e tests (don't need this for any other section except e2e since this is the only one that breaks beyond our control).

kojiromike:
> Actually, looks like the collect Test is potentially meant to be there to ensure the docker dir collection works. If that is case, then disregard that part of my comment above.

Yes, collect isn't a test, it is the thing that builds the matrix used to define the individual tests.

.github/workflows/test.yml
```diff
@@ -9,1670 +9,177 @@ on:
...
+      run: |
+        . ci/ciLibrary.source
+        configure_coverage
+    - name: Unit testing
+      if: ${{ success() || failure() }}
+      run: |
+        . ci/ciLibrary.source
+        build_test unit
+    - name: E2e testing
+      if: ${{ (success() || failure()) && ! contains(fromJson(vars.SKIP_E2E_CONFIGS || '[]'), matrix.configurations.docker_dir) }}
```
kojiromike: @adunsulag @bradymiller If I did this right, then if you create a repo variable named `SKIP_E2E_CONFIGS` and its value is a valid json string that contains the docker dir names of the configs for which you want to disable e2e, then those tests should be skipped. 
image

bradymiller:
hi @kojiromike , didn't work. Per Claude looks like may be issue with the `||` and pasted what Claude told me when I asked what was wrong with the script:

The issue is with your conditional expression syntax. You have:

```yaml
if: ${{ (success() || failure()) && ! contains(fromJson(vars.SKIP_E2E_CONFIGS || '[]'), matrix.configurations.docker_dir) }}
```

The problem is `vars.SKIP_E2E_CONFIGS || '[]'` - this syntax doesn't work in GitHub Actions expressions.

## Fix Option 1: Use Separate Condition Check

```yaml
- name: Check if E2E should be skipped
  id: check-skip
  run: |
    if [ -z "${{ vars.SKIP_E2E_CONFIGS }}" ]; then
      echo "skip_configs=[]" >> $GITHUB_OUTPUT
    else
      echo "skip_configs=${{ vars.SKIP_E2E_CONFIGS }}" >> $GITHUB_OUTPUT
    fi

- name: E2e testing
  if: ${{ (success() || failure()) && !contains(fromJson(steps.check-skip.outputs.skip_configs), matrix.configurations.docker_dir) }}
```

## Fix Option 2: Handle Empty Variable in Repository Settings

Set your `SKIP_E2E_CONFIGS` repository variable to `[]` (empty JSON array) instead of leaving it undefined. Then your original syntax works:

```yaml
- name: E2e testing
  if: ${{ (success() || failure()) && !contains(fromJson(vars.SKIP_E2E_CONFIGS), matrix.configurations.docker_dir) }}
```

## Fix Option 3: Use Multiple Conditions

```yaml
- name: E2e testing
  if: ${{ (success() || failure()) && (vars.SKIP_E2E_CONFIGS == '' || !contains(fromJson(vars.SKIP_E2E_CONFIGS), matrix.configurations.docker_dir)) }}
```

## Recommendation

**Option 2 is the cleanest** - just ensure your repository variable `SKIP_E2E_CONFIGS` is always set to a valid JSON array (even if empty: `[]`). This way your conditional logic stays simple and readable.

The `||` operator for providing defaults doesn't work in GitHub Actions expressions the way it does in other contexts.

bradymiller:
I like option 3 since then won't break things if we forget to make the variable `[]` when it is empty (which i can see us doing after we forget in couple weeks :) )

kojiromike:
@bradymiller @adunsulag OK, I have renamed the variable `SKIP_E2E_DOCKER_DIRS` (because I felt it would be much clearer after we all forget how this works) and we can try again when one of you has time to make the config change.

stephenwaite:
hi @kojiromike , updated the variable name and re-ran the tests :crossed_fingers:

kojiromike:
> hi @kojiromike , updated the variable name and re-ran the tests 🤞

sadly, it did not work. I'm going to need to take a closer look at the repo vars feature before I circle back on this.

kojiromike:
OK, I retested on my fork, and this does work as expected. I added a debug step that makes it clear what variables are set, and in this repo, it seems the `SKIP_E2E_DOCKER_DIRS` variable is not set.

image

Perhaps it is set, but not in the right place.

It should be under Settings -> Security -> Secrets and variables -> Actions, then the Variables tab, and *Repository variables*, not Environment variables.

image

bradymiller:
Looks like PR's won't have access to these (just likw the secrets as we discussed on the last call):
Found this on the variables page in github:
`Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.`

sjpadgett:
I really like and am glad to see this P.R., thanks. FYI I recently had problems with scope when trying to define variables however I was accessing within a test. In case it matters never did get to work.

kojiromike:
> Looks like PR's won't have access to these (just likw the secrets as we discussed on the last call): Found this on the variables page in github: `Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.`

OHO! Nasty, GitHub. It's always fun to find a thing that works on a personal fork but suddenly doesn't work when you provide it upstream.

@bradymiller I think maybe we have to abandon this approach. I was hoping to find a way to disable E2E for specific configurations _without_ having to make a commit, but I think this is not it.

bradymiller:
Yep @kojiromike , may need to consider following strategy:
docker-compose.e2e.disabled.yml (run, but skip e2e)
docker-compose.disabled.yml (do not run anything) (already supported)
docker-compose.yml (standard run) (already supported)

kojiromike:
@bradymiller OK, this is ready for review again. I had to take a slight deviation from the above because if we rename the docker-compose to something like `docker-compose.e2e.disabled.yml` then we have the additional work of adding `-f` to the docker compose command, conditionally. If you rename the docker-compose file to anything other than `docker-compose.yml` it will disable that entire configuration. However, if you rename the directory so it ends with `_no-e2e`, then the e2e tests get skipped for that config.

I added a ci/README.md as well.

bradymiller:
@kojiromike , awesome stuff! the dir change is actually better since then easy to see which ones are skipping e2e (since at some point in future, we will then change it to see if it works). bringing in. and thanks for the super clear documentation!

PR #8417, opened 2025-06-05
ci: consolidate docker compose files

### LINKED ISSUE #8416: reduce repetition in ci docker-compose files
We can use docker compose file composition features to reduce the repetition of the ci/*/docker-compose files down to the essential differences.

kojiromike:
Fixes #8416

Use `extends` to reduce duplication in docker compose files in ci/

Create source docker-compose files for apache and nginx, and reference those files in the different test configurations.


bradymiller:
hi @kojiromike ,

Again, really cool stuff. Just did a bit of research on this.
Would `include` be better. For example (and then would basically omit the image: lines in the base files):
```
include:
  - ../docker-compose.apache-base.yml
services:
  mysql:
    image: mariadb:10.11 
  openemr:
    image: openemr/openemr:flex-3.20
```

EDIT:

new example with command: included
```
include:
  - ../docker-compose.apache-base.yml
services:
  mysql:
    image: mariadb:10.11 
    command: ['mysqld','--character-set-server=utf8mb4']
  openemr:
    image: openemr/openemr:flex-3.20
```

bradymiller:
Note in my above example, would also need an entry for command: in the mysql service since this differs between mysql and mariadb (which could also omit in the base).

EDIT:
There is even a difference in command for different mysql versions (because of the stupid required --default-authentication-plugin=mysql_native_password setting in mysql 8.0)

ANOTHER EDIT:
Even difference in command in mariadb versions where earlier ones use mysqldb rather than mariadb (betting the bug you have now that is breaking ci is using mysqldb rather than mariadb for the later mariadb versions). So definitely should separate out the command: into the above example and omit them from the base.

kojiromike:
@bradymiller

- I don't think we can use `include`. At least, the last time I tried, it didn't like me redefining services from included compose files.
- I think I have the commands right now.

There are other ways we can make these more flexible…
- We can use profiles to define services that don't start by default, and activate different profiles for different tests.
- We can use variables for docker image tags
- We can use yaml composition, as well as some more advanced features of compose like !override and !reset.

However, I think I should pause here and go back to working on inferno or making the -devops builds faster and more resilient.

kojiromike:
I'm happy to help resolve conflicts with #8415 in either merge order.

bradymiller:
Sounds good. Agree not to spend to much time on this :) at work but will
check out the pr this evening and plan to bring it on as long as working.

bradymiller:
hi @kojiromike , Looks good and testing well. Bringing in. Thanks for the improvement and the nice documentation! If i have time in future, i may play around with the include stuff (since seems like a neat thing to try out).

PR #8419, opened 2025-06-06
chore(ccdaservice): bump libxmljs2 to 0.37.0

### LINKED ISSUE #8418: ccdaservice doesn't build on node 24
The ccdaservice won't build on my macbook. I believe it's because I have node 24 installed. Further, I think it's because the service depends on libxmljs2, which only added support for node 24 in 0.37.0 yesterday. Perhaps we can update to that version and solve the issue.

kojiromike:
support nodejs 24

Fixes #8418

Fix that `npm install` fails in ccdaservice when the node version is 24.

Update libxmljs2 to 0.37.0, which works with node 24.

### DIFF
--- a/ccdaservice/README.md
+++ b/ccdaservice/README.md
@@ -19,19 +19,19 @@ To ensure the lastest libraries are installed, node version changes or the packa
 
 
 ## Ubuntu Setup
-* Latest version tested is node v20.10.0
+* Latest version tested is node v24.1.0
 
 If node is not already installed then do the following(Research the required installation for your environment):
 - `cd ~`
 - `apt-get remove --purge nodejs npm` // Ensures clean install and will allow chance to clean up.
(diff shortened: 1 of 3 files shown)


sjpadgett (approved):
Tested on Windows xampp ubuntu with node 24.0.1 and libxmljs2 version 37,  however on PHP 8.2.

Testing included build service, generate a couple CCDA's and import 200+ CCDA's from single zip.

Thanks for contribution!

kojiromike:
@bradymiller I'm going through my changes since April and adding the `Status: Needs Release Merge` label onto things I think should go in 7.0.3.next. Please let me know if you disagree with any of these.

sjpadgett:
@kojiromike Problem with these kinds of changes in a patch is legacy and having the knowledge that most user can't go to the command line if that is needed to support patch, especially this PR if they have to update their node version.
I always try to back port to production all I can for patches but it can be dangerous and patch testing is crucial.
Happy your thinking quick support to users.

PR #7780, opened 2024-10-24
Patch 2 rel-702

sjpadgett:
* Eligibility Fixes Change eligibilty password and login to use SFTP inputs. Rework MIME parsing.
Fix conditional preventing any request calls.

* Prevent auto select unassigned for multiselect if other items are selected. i.e only select for empty selects.

* add title for review encounters

* refactor out oeHttp

(cherry picked from commit 13d17525381c29bb567ad9ca8f25af44564fbcba)

Fixes #

PR #8424, opened 2025-06-06
Fixes #8423, #6889 logo customizations

### LINKED ISSUE #6889: bug: Portal login page logo path does not exist
The sites path `sites/default/images/portal/login/primary` does not exist, resulting in a failed customization of the login logo.

### LINKED ISSUE #8423: feat: Add ability for module writers to change location of logo files for site customization
### Describe the problem

If a module writer wants to handle all customization for a site installation from within the module including the logos displayed on the site, they currently cannot do so.

As a module writer I want an event that I can hook into to be able to change the web path locations of where the logos are stored.

adunsulag:
Adds a logo filter event to allow for customizing the web path location of the logo.  It still requires a valid path (so if using something dynamic from the db will need an php controller file that is publicly accessible to serve up the dynamic image).

Updates the LogoService to dispatch the filter event.

Also adds in folders for customizing the portal primary and secondary logos.

fixes #6889
fixes #8423

Partially addresses #7780 by allowing logo customizations but not in the way the issue discusses.


adunsulag:
Been digging into the tests failing. It doesn't make any sense on why they are failing outside of perhaps a resource issue on github.  I can run the same ci docker containers and it doesn't fail.  Ill try running them again tomorrow and see if its something github related.  Spent several hours profiling and there's performance improvements that could be done but it shouldn't be giving the 30 second timeout that the docker images are experiencing.

bradymiller:
Just tried to rerun them. Odd that this is only failing on this PR and not other runs.

bradymiller:
Maybe it will work in the morning :)

PR #8433, opened 2025-06-10
ci(tests): wait for healthcheck

### LINKED ISSUE #8432: Dockerfile healthchecks for openemr
We added healthchecks to some of the mariadb services in docker compose a while ago. Now I propose we add healthchecks for openemr services too.

If we add a healthcheck to openemr, we can tell when it's able to respond to http requests and mark services healthy after openemr.sh is complete. This is very useful information, and can also simplify tests if we run something like

```
docker compose up --detach --wait --wait-timeout 600
```

Which will cause docker compose to block for up to 10 minutes until all services (that have a healthcheck) are healthy.

kojiromike:
Fixes #8432

Use healthchecks against the openemr service to defer startup until openemr is ready to go

add a healthcheck to ci/ compose, and then --wait for it.

### DIFF
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -60,7 +60,7 @@ _exec() {
 }
 
 dockers_env_start() {
-    dc up --detach
+    dc up --detach --wait --wait-timeout 10m
 }
 
 actions_chmod() {
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -60,7 +60,7 @@ _exec() {
 }
 
 dockers_env_start() {
-    dc up --detach --wait --wait-timeout 10m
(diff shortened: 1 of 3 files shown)


ci/ciLibrary.source
```diff
@@ -60,7 +60,7 @@ _exec() {
 }
 
 dockers_env_start() {
-    dc up --detach
+    dc up --detach --wait --wait-timeout 10m
```
kojiromike:
```suggestion
    dc up --detach --wait --wait-timeout 300
```

kojiromike:
This is actually passing most tests. The failure is one of those flukes, I think.

bradymiller:
hi @kojiromike , looks good and bringing in.

PR #8448, opened 2025-06-13
ci(gh actions): phpstan

### LINKED ISSUE #8447: Scan Code with PHPStan
PHPStan is a static analyzer for php. As such it will catch certain classes of errors far more efficiently than tests.

kojiromike:
Fixes #6810
Fixes #8447

Introduce PHPStan into composer.json and as a GitHub action.

Add the PHPStan Github Action, mainly.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -1,0 +1,37 @@
+name: PHPStan
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+permissions:
+  contents: read
+
+jobs:
+  phpstan:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+    - name: Composer Cache
+      uses: actions/cache@v4
+      with:
+        path: /tmp/composer-cache
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+    - uses: php-actions/composer@v6
+      with:
+        php_extensions: gd soap zip
+    - name: Gather PHP Paths
+      id: gather-php-paths
+      run: |
+        # Gather all the directories that contain PHP files and all the root level php files, too.
+        {
+          printf 'paths='
+          git ls-files '*php' | awk -v ORS=' ' -F / '!$2 { print; next; } !s[$1]++ { print $1 }'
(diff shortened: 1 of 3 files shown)


bradymiller:
Is there an auto-fix :)

Is there a way to manage false positives?

kojiromike:
@bradymiller phpstan does not seem to have its own auto-fixer. There is a tool called [rector](https://getrector.com/documentation) I've been looking at that might fit the bill. In fact, rector even has a lot of options for managing the scope of what it changes, so the PRs don't have to be continent-size.

The closest thing I saw to a false positive might be phpstan confusing the sqlQuery function in admin.php with the one in library/sql.inc.php. I would actually prefer to fix that than to ignore it, but we can always ignore things. https://phpstan.org/user-guide/ignoring-errors

kojiromike:
@bradymiller I did a little PoC on my fork of rector, and it's pretty nice, I think. Take a look, if you think it's interesting.

The PR is https://github.com/kojiromike/openemr/pull/11. The first commit is me setting up the rector stuff, but the second commit was by rector, doing a relatively small set of php improvements. I left it at pretty conservative defaults, but we can ratchet it up over time.

kojiromike:
I pushed a change, but it didn't trigger the actions. I think there's a github outage happening right now.

kojiromike:
Github's working again.

contrib/forms/handwritten-notes/hw-notes.php
```diff
@@ -10,7 +10,7 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  **/
 
-require_once("../../globals.php");
+require_once("../../../interface/globals.php");
```
bradymiller: Would revert this since the path will be relative from interface/forms/handwritten-notes, and don't want it working from where it currently is.

Otherwise this PR is ready to come in :)

kojiromike: I guess we'll have to add ignore expressions for these off-path imports.

kojiromike: I updated the PR to remove the fixes. I'll open separate PRs for the fixes, and for any ignore/configuration we want to do.

bradymiller:
Looks good and bringing in. thanks!

PR #8465, opened 2025-06-19
fix(admin.php): rename sqlQuery to avoid naming conflict

### LINKED ISSUE #8464: Rename `sqlQuery` in `admin.php` to avoid confusion
the `sqlQuery` function in admin.php has a different signature from the function with the same name in library/sql.inc.php. Phpstan thinks the definition in admin.php is what's being used in very many places throughout the openemr codebase, so it reports a mismatch between the function signature and its use.

It might be possible to configure phpstan to assume the other one, but it still could be a point of confusion for humans, and since the `sqlQuery` in admin.php is essentially private, I think it makes more sense and is easiest to rename it.

kojiromike:
Fixes #8464

Fixes the conflict between the two sqlQuery function definitions in the codebase.

- rename `sqlQuery` to `adminSqlQuery` in admin.php

### DIFF
--- a/admin.php
+++ b/admin.php
@@ -29,7 +29,7 @@ if (stripos(PHP_OS, 'WIN') === 0) {
 
 $OE_SITES_BASE = "$webserver_root/sites";
 
-function sqlQuery($statement, $link)
+function adminSqlQuery($statement, $link)
 {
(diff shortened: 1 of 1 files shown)


adunsulag:
This looks like we can bring it in, re-ran 8.5 test but doesn't look like its related.

PR #8486, opened 2025-06-20
Openemr external pr 8409 automate inferno testing

adunsulag:
Took @kojiromike's PR and added fixes to that.

Fixed the db credentials for OpenEMR to connect to mysql using the
environment variables.  This makes the openemr-cmd snapshot installation
work as long as the auto_configure.php file is there.  Still need to
work on getting the auto_configure.php to be copied into the docker
container.

Fixed up the submodule to enable CORS on the nginx installation.

Wrote out a preliminary api test for the single patient api.  Found out
the us_core_311 suite is very different in its terminology support
compared to the g10_certification.  So tests passing in
g10_certification are failing in the us_core_311 which is very
frustrating.

I can get individual resource results on the us_core_311 suite whereas
on the g10_certification the most granular I can get to right now is the
entire single patient test suite (all 300+ tests).  I need to either
figure out a way to handle the reporting of the entire test suite in a
way that is easily visible from the phpunit test runner or figure out a
way to make the resource tests executable as individual groups.

I also discovered a bug in the redis docker installation where it fails
to change the owner to be redis:redis on the /data folder which causes
inferno test session instances to fail to create.  It looks like the
.keep file in the folder directory causes redis to skip the permission
change.  To fix it manually you have to run the following command which
I found on the redis docker instructions page.

docker run --rm -v /your/host/path:/data redis chown -R redis:redis /data

Next steps once I can get g10_certification to run consistently and the docker builds to come up completely is to enable the code coverage piece.  I'm thinking we turn on the coverage using a docker environment variable on the openemr image, we can then collect the xdebug coverage reports and send them up to codecov.


adunsulag:
Apparently I'm very tired.  This was supposed to go to my fork, not main repo as its only meant for draft.

PR #8490, opened 2025-06-23
fix(code_types): convert_type_id_to_key explicit return

### LINKED ISSUE #8489: `convert_type_id_to_key` is not total
The function `convert_type_id_to_key` defined in `code_types.inc.php` has a return type of `string`, but can inadvertently and implicitly return `null` when the provided $id is not found. This could lead to errors downstream that are hard to track down and debug.

kojiromike:
Fixes #8489

Fix the type of the function by clarifying that it can return null.

Clarify the type signature of the function and explicitly return `null` when no match is found. This does not change the actual runtime behavior of the function – just aids typecheckers and other static analyzers in understanding the behavior.

Also fix a small style error in the following function.

### DIFF
--- a/custom/code_types.inc.php
+++ b/custom/code_types.inc.php
@@ -312,7 +312,7 @@ function related_codes_are_used()
  * Convert a code type id (ct_id) to the key string (ct_key)
  *
  * @param  integer $id
- * @return string
+ * @return string|null
  */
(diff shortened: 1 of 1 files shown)


adunsulag (approved):

adunsulag:
Thanks for the improvement!

PR #8494, opened 2025-06-23
fix(ORDataObject): refactor for consistent initialization and id type

### LINKED ISSUE #8491: Inconsistent extensions of ORDataObject
`ORDataObject` itself calls `get_id` internally, but does not declare or implement it. However, declaring `get_id` with a reasonable type signature is problematic. Currently there are 33 classes that extend `ORDataObject`. All the subclasses implement `get_id` and `set_id`, but some of them require `$id` to be numeric, whereas others don't. In many cases, if the provided $id is not an expected type, the setter will silently drop the value without any warning or error.

Also, some implementations of `set_id` are fluent, meaning they return `$this`, whereas others are not, and return `void`.

Also, some of the implementations override the base constructor and don't call it.

kojiromike:
Fixes #8491

Makes all the subclasses of `ORDataObject` initialize consistently and have the same type of `$id`.

1. Make `ORDataObject` an abstract class.
2. Implement `get_id` and `set_id` consistently in `ORDataObject` and remove subclass implementations.
3. `set_id` is never fluent.
4. Enable implementations to supply `$_table` and optionally `$_prefix` as class properties instead of setting them in the constructor.
5. Make a consistent default constructor such that
  1. Default value initialization (`init`) and database population (`populate`) can happen for most classes without overriding the constructor.
  2. It's clear when something does need to override the constructor.
6. `ORDataObject`'s own call to `get_id` can reference a function that exists.
7. All subclasses call the constructor, explicitly or implicitly.


kojiromike:
*Note* I did not attempt to fix any implementations in contrib!

adunsulag:
@kojiromike did you want this to be a draft?  I noticed a lot of the tests are failing with these changes, wanted to check to see if you wanted us to review this or not.

kojiromike:
@adunsulag sorry about that, it should pass now.

kojiromike:
@openemr/admins If you're able, please review this

bradymiller:
hi @kojiromike , I'll check this out over weekend if nobody else gets to it by then.

bradymiller:
sorry for delay on this. looking at this now and plan to complete review over next couple days.

bradymiller:
hi @kojiromike ,
If users have created custom forms based on the prior mechanism, would they break?
If so, it looks like fixing them would be straightforward.
just making sure these changes are worth this?

kojiromike:
@bradymiller With type inconsistencies like this it's sorta hard to reason about whether or not someone could even create a form based on the prior mechanism -- at least, they'd have to find a path where the id type is an integer-like string anyway, but php, being weird, doesn't really mind that.

```
php > $x = '2';
php > echo gettype($x);
string
php > echo $x + 4;
6
```

So I think it wouldn't break, even in that case.

That said, because it's hard to reason about, it might still break. I can't be sure.

adunsulag:
I pulled this down to play with it from the GUI side, it looks like C_FormVitals::populate_object() was failing.  Sadly, due to our lack of sufficient tests we're going to need to do some tests on the other forms,  if this one was broken there are likely to be others :(

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -450,9 +450,15 @@ public function populate_object(&$obj)
         foreach ($_POST as $varname => $var) {
             $varname = preg_replace("/[^A-Za-z0-9_]/", "", $varname);
             $func = "set_" . $varname;
+            if ($func == "set_id") {
+                if (!empty($var)) {
+                    $var = (int)$var; // make sure we have an integer
+                } else {
+                    $var = null; // if we don't have an id then we set it to null
+                }
+            }
```
kojiromike, line 459: I take it 0 is not a valid value for the id?

adunsulag: I usually avoid it as 0 often has been used in the past for special state conditions in the codebase rather than the fact that there is no value and null is used to represent the absence of value.

kojiromike: It just seems it might confuse readers, because empty(0) is true, but 0 is a valid integer.

kojiromike:
> I pulled this down to play with it from the GUI side, it looks like C_FormVitals::populate_object() was failing. Sadly, due to our lack of sufficient tests we're going to need to do some tests on the other forms, if this one was broken there are likely to be others :(

Understood. I don't know when I'll get to it, but I'll see what I can do to write tests on the old version of the code, before taking this change through.

PR #8496, opened 2025-06-24
fix(shell): address SC2006

kojiromike:
Ran `shellcheck -i SC2006 -f diff | git apply` to automatically resolve the SC2006 errors, reducing the error count by 30.


bradymiller:
hi @kojiromike , Thanks for the fixes straightforward changes, so bringing in.

PR #8497, opened 2025-06-24
fix(shell): address SC2250

kojiromike:
Ran `shellcheck -i SC2250 -f diff | git apply` to automatically fix this type of finding, reducing the error count by 757.


bradymiller:
going to bypass testing here and bring in since the dependabot PR's will likely take awhile to go through.

bradymiller (approved):

PR #8517, opened 2025-06-24
chore(shellcheck): ignore SC1091

kojiromike:
We don't vendor /usr/share/debconf/confmodule for shellcheck to follow it.


bradymiller:
@kojiromike , fyi, these ubuntu packages files are going to be removed very soon (pretty much after i bring in your 2 PRs :) )

bradymiller:
going to bypass testing here and bring in since the dependabot PR's will likely take awhile to go through.

PR #8520, opened 2025-06-25
fix(backuplog.sh): fix all shellcheck complaints

kojiromike:
Fix all the shellcheck complaints in this file and some other cleanup.

1. Use a correct shebang.
2. Name the positional parameters.
3. Create a function to reduce duplication.
4. Quote the parameters.
5. Use long shell options for clarity.
6. Use curly-braces around parameter expansions.


bradymiller (approved):

bradymiller:
looks good and bringing in

PR #8522, opened 2025-06-25
ci(shellcheck): remove contrib scripts

### LINKED ISSUE #8521: Remove contrib/util/restore and clean up shellcheck
The ubuntu scripts in contrib were recently removed in #8518. The only remaining shell script in contrib is contrib/util/restore. I believe it is entirely superseded by features in openemr-cmd, and also shellcheck finds many problems with the script. Rather than fix the issues, I would like to remove it and clean up the shellcheck command.

kojiromike:
Fixes #8521

Resolves many shellcheck complaints by removing the only remaining shell script in contrib, and removes a bunch of custom file paths in the shellcheck workflow.

1. remove contrib/util/restore
2. remove special contrib paths from shellcheck
3. adds a shellcheck badge to the readme


bradymiller:
hi @stephenwaite and @adunsulag , In the remote past, there has been mention of use of the restore scripts on the forum. Do you think this is still being used or ok to shelve it?

sunsetsystems:
> hi @stephenwaite and @adunsulag , In the remote past, there has been mention of use of the restore scripts on the forum. Do you think this is still being used or ok to shelve it?

Greetings all! The issue says "I believe it is entirely superseded by features in openemr-cmd". I'd just want to see that confirmed, perhaps with an example.

kojiromike:
> > hi @stephenwaite and @adunsulag , In the remote past, there has been mention of use of the restore scripts on the forum. Do you think this is still being used or ok to shelve it?
> 
> Greetings all! The issue says "I believe it is entirely superseded by features in openemr-cmd". I'd just want to see that confirmed, perhaps with an example.

The restore script restores a backup of an openemr snapshot, right? Is that not the same thing as `openemr-cmd restore-snapshot`?

sunsetsystems:
I don't know, which is the point. I've been out of the community development loop for a while. But also I don't see openemr-cmd in the distribution, it appears to be only in the openemr-devops project and intended (?) for docker implementations. How will those who install or upgrade from a normal distribution download know that it exists, how to get it and how to use it?

kojiromike:
> I don't know, which is the point. I've been out of the community development loop for a while. But also I don't see openemr-cmd in the distribution, it appears to be only in the openemr-devops project and intended (?) for docker implementations. How will those who install or upgrade from a normal distribution download know that it exists, how to get it and how to use it?

I think it is intended for docker implementations. That's fair. I'm not sure what a "normal distribution" is, though. My understanding is OpenEMR wants to discourage, say, XAMPP installations, because they can lead to insecure deployments. I'll set this aside and just fix the shellcheck problems in this script directly.

bradymiller:
yep, the snapshot/capsule stuff was designed to be more specific to testing/development/dockers and is not the same as openemr's traditional backup(via gui in openemr or via backup script). The traditional backup will backup entire database and entire codebase. And the restore script will take this file and restore OpenEMR from it. The snapshot is more lightweight and does not backup the entire codebase, but instead just the sites directory. Also the devtools/openemr-cmd are really only supported in the flex docker series, so don't really reach the productions dockers.

PR #8523, opened 2025-06-25
fix(shellcheck): autofix

kojiromike:

Use `shellcheck -f diff | git apply` to automatically fix those remaining shellcheck issues that can be automatically fixed.

Mostly putting quotes around expansions.

This can cause issues in cases where whitespace was built up in a variable directly. I'm not sure how to test for that expressly, but we can address those if they should appear.

By automation, yes. By AI? No.


bradymiller:
thanks @kojiromike !

PR #8524, opened 2025-06-25
fix(shellcheck): remaining issues

kojiromike:

This should resolve the remaining shellcheck findings after #8522 and #8523

- Avoid using `find` when globs will do the trick.
- Ensure variables are defined or checked.
- Ensure exit conditions are honored.

### DIFF
--- a/bin/generate-ehi-documentation.sh
+++ b/bin/generate-ehi-documentation.sh
@@ -12,7 +12,6 @@ TEMPLATE_LOCATION=${EHI_ROOT_FILE}/layout/
 DOC_OUTPUT=../Documentation/EHI_Export/docs/
 JAR_FILES=${EHI_ROOT_FILE}/jars
 TYPE=mariadb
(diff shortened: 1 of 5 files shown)


stephenwaite (approved):
looks good, thanks @kojiromike

bradymiller:
is this ready to go out of draft mode?

kojiromike:
@bradymiller I wanted to resolve #8522 first before making this mergeable. Otherwise it won't actually resolve the remaining issues contrib/util/restore has the lion's share of shellcheck issues in the repo, and many of them would need to be resolved manually, so if we can't remove that script, it'll be a little bit before we can get a green shellcheck badge.

kojiromike:
I gave in and resolved the shellcheck issues in contrib/util/restore even though I don't think we can know if it's being used.

bradymiller:
hi @kojiromike , thanks for rehabilitating that script. looks good. bringing in.

PR #8550, opened 2025-07-01
Update dependabot.yml - changed rebase-strategy for composer to manual since the auto was overtaking the ci testing

bradymiller:
Update dependabot.yml - changed rebase-strategy for composer to manual since the auto was overtaking the ci testing

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,7 @@ updates:
       interval: "weekly"
       day: "monday"
       time: "08:00"
+    rebase-strategy: "disabled"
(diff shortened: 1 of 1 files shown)


bradymiller:
This should avoid the craziness that happens with the dependabot composer package updates (where when bring one in, all the other need to rebase which then overtakes the ci testing for a long time). After I bring this in, it will happen one more time. And then after that, we should be able to manually rebase the PR and then bring it in after the testing is complete (ie. auto rebasing will be off for the composer update PR's; note the npm will still auto update since these don't seem to be as problematic as the composer updates).

PR #8566, opened 2025-07-08
ci(tests): split e2e from other tests

kojiromike:
Fixes #


kojiromike:
This is more complicated than I thought because I still have to re-combine the coverage after doing both kinds of testing. Too bad, will revisit later.

PR #8580, opened 2025-07-10
fix(ci):fix edge case in functional testing when adding user and fix inferno skipping logic

bradymiller:
fixes #8604

PR #8582, opened 2025-07-10
test(e2e): publish videos of e2e tests

### LINKED ISSUE #8581: Provide video artifacts from e2e tests for easy debugging
Now that the e2e tests run with selenium since #8555 (thanks, Brady!) we can produce more debuggable output and even demo material in the form of videos produced through those tests.

Since Github doesn't limit the storage of test artifacts (it just deletes them after some time), it costs OpenEMR nothing to provide these videos as artifacts, as long as users know they are only available temporarily for any given test run. 

So I propose we make the actions produce the videos.

kojiromike:
Fixes #8581

Makes videos of E2E tests available for easy debugging and potentially feature demonstrations.

Publish videos of E2E tests as workflow artifacts.

### DIFF
--- a/ci/compose-shared-selenium.yml
+++ b/ci/compose-shared-selenium.yml
@@ -1,8 +1,12 @@
 services:
   selenium:
     image: selenium/standalone-chrome:134.0-chromedriver-134.0-grid-4.31.0-20250414    # Pinned to Chrome 134 and grid 4.31
+    # Services with profiles don't start by default.
+    profiles:
+    - selenium
+    - video-recording
     ports:
-    - 4444:4444    # Selenium WebDriver interface
+    - 4444:4444  # Selenium WebDriver interface
     environment:
       SE_SCREEN_WIDTH: 1920
       SE_SCREEN_HEIGHT: 1080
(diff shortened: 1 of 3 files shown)


ci/compose-shared-selenium.yml
```diff
@@ -11,13 +15,26 @@ services:
...
+      - CMD
+      - curl
+      - -f
+      - http://localhost:4444/wd/hub/status
       start_period: 30s
       interval: 30s
       timeout: 5s
       retries: 3
+  video:
+    image: selenium/video:ffmpeg-7.1-20250707
```
bradymiller, line 28: dance-bear

bradymiller:
@kojiromike , Awesome stuff! The non-headless mode and/or video processing do add several minutes to each run, but definitely worth it. Looks like is exacerbating the user add issue in functional testing, but that is an already existent bug that am working on in #8580 (also adding code there to nicely skip the inferno run when don't have secret), so no reason to hold this PR to wait for that (and having it happen more often will make it easier to debug/fix :) ). Bringing this in.

bradymiller (approved):

PR #8587, opened 2025-07-10
refactor(calendar): remove return/store polymorphism from pnHTML

### LINKED ISSUE #8586: postnuke html generator is unnecessarily polymorphic
The pnhtml object used I guess for the calendar has a feature where you can toggle whether each html generation method either

1. stores its output in an internal buffer and returns nothing, or
2. returns its output, and stores nothing

It is implemented via a `define` constant, plus an internal variable on the class. This not only confuses phpstan due to the return values not being clear, but it requires careful state checking before making any of these method calls. For the most part, this toggle-output feature isn't really used all that heavily anyway. I propose to replace the internal condition with two separate functions -- one to return the value and the other to store the state. Then, callers just need to call the function they want.

kojiromike:
Fixes #8586

The legacy `pnHTML` class used a mutable-state polymorphism: every generator method either *returned* its HTML or *stored* it in an internal buffer, depending on a `SetOutputMode()` toggle maintained across calls. The same method call did different things at different points in the request, making the code hard to reason about and easy to break.

- `pnHTML` generators are now pure functions that always return a `string`. Methods were renamed with a `generate*` prefix (`generateFormStart`, `generateFormSelectMultiple`, `generateText`, etc.) and the old names were deleted so PHPStan flags any stragglers.
- `GetOutput(string $body): string` and `PrintPage(string $body): void` now take the assembled body explicitly, replacing the internal output buffer and `SetOutputMode` toggle.
- All callers in `pnadmin.php`, `pnuser.php`, `pc_date_select`, `pc_filter`, and `interface/main/calendar/index.php` were updated to accumulate a `$body` string and pass it through.

Net effect: +191 / −438 lines across the calendar module, plus net −80 lines across the PHPStan baselines (more stale entries drained than new ones accrued — `return.missing.php` went 29 → 20, which required a ratchet cap update in `.phpstan/fatal-baseline-caps.php`).

### DIFF
--- a/.phpstan/baseline/assign.propertyType.php
+++ b/.phpstan/baseline/assign.propertyType.php
@@ -97,16 +97,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../interface/forms/vitals/C_FormVitals.class.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Property pnHTML\\:\\:\\$header \\(string\\) does not accept array\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../interface/main/calendar/includes/pnHTML.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Property pnHTML\\:\\:\\$output \\(string\\) does not accept mixed\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../interface/main/calendar/includes/pnHTML.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Property Smarty_Legacy\\:\\:\\$caching \\(int\\) does not accept false\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../interface/main/calendar/modules/PostCalendar/pnadmin.php',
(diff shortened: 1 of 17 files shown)


bradymiller:
Hi @kojiromike . That doesn't look like very fun coding :) If testing well for you (just let me know), then happy to bring in (after merge conflict and styling issue(s) fixed).

kojiromike:
This can wait until all the automatic rector changes are done. See #8601

kojiromike:
@aisle-analyzer

kojiromike:
Re: the CSRF finding from `aisle-research-bot` above — valid but out of scope for this PR, which is a pure HTML-generation refactor with no auth/session logic. The gap is systemic across `PostCalendar/pnadmin.php` (no `CsrfUtils` usage anywhere in the module), so I've filed #11829 to audit and wire `CsrfUtils::verifyCsrfToken()` into every state-changing admin handler as a separate security fix.

Firehed (approved):
I eagerly await the day we're post postnuke. :shipit:

PR #8594, opened 2025-07-11
Re-add couchdb volume to openemr service in development-easy-light setup

### LINKED ISSUE #8593: Missing couchdb volume causes rsync error during openemr startup in development-easy-light setup
### Describe the bug

Running `docker compose up` in the current `development-easy-light`setup causes an rsync error in the openemr container during startup:

```
rsync: [sender] link_stat "/couchdb/data" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.4.1]
```

It seems that after the couchdb service has been removed from the new `-light` setup, there is still a dependency to the couchdb volume (that has also been removed).

Should this volume just be added again for simplicity?

### To Reproduce

1. Clone current OpeneEMR project (Commit 9ae210a57eb5fd33452ea189970d340c695a359e)
2. Run `docker compose up` in `development-easy-light` setup.
3. See logs for error

### Expected behavior

Expected system to start successfully

### Client configuration

- Commit: 9ae210a57eb5fd33452ea189970d340c695a359e
- Operating system: macos 15.5

cburnicki:
The couchdb service is missing from this setup but openemr still depends on the volume being present for an rsync command.

Fixes #8593

Starting the development-easy-light setup using docker-compose currently causes an rsync error in the openemr container, preventing a successful startup.

This solution provides an (easy to understand) quick fix by just re-adding the volume to the container (as in the development-easy setup), although there is no couchdb service present in the easy-light setup.
Another solution, of course, would be to remove the rsync command that represents the broken dependency. But my guess is this would cause container images to diverge, so just adding the volume might be an easier and cleaner solution for the dev setup.

No

### DIFF
--- a/docker/development-easy-light/docker-compose.yml
+++ b/docker/development-easy-light/docker-compose.yml
@@ -40,6 +40,7 @@ services:
     - vendordir:/var/www/localhost/htdocs/openemr/vendor:rw
     - ccdanodemodules:/var/www/localhost/htdocs/openemr/ccdaservice/node_modules:rw
     - logvolume:/var/log
+    - couchdbvolume:/couchdb/data
     environment:
       DEBUG_COLORS: "true"
       TERM: xterm-256color
@@ -105,3 +106,4 @@ volumes:
   vendordir: {}
   ccdanodemodules: {}
   logvolume: {}
+  couchdbvolume: {}


docker/development-easy-light/docker-compose.yml
```diff
@@ -40,6 +40,7 @@ services:
     - vendordir:/var/www/localhost/htdocs/openemr/vendor:rw
     - ccdanodemodules:/var/www/localhost/htdocs/openemr/ccdaservice/node_modules:rw
     - logvolume:/var/log
+    - /couchdb/data
```
kojiromike: I think you should not use an anonymous volume in this case, because people will want to retain and manage their couchdb data. I don't think that's the case with the .git data, based on how the development dockers are designed.

kojiromike (approved):
I'm not an admin, but this looks fine to me

bradymiller (approved):

bradymiller:
hi @cburnicki , agree looks good. thanks for the fix! bringing in.

PR #8598, opened 2025-07-11
Bugfix/fix drug dispension

### LINKED ISSUE #8597: Drugs cannot be dispatched from inventory
### Describe the bug

### To Reproduce

**Preconditions**

Make sure drug sales are enabled and a drug/lot exists in the system:

1. Admin -> Config -> Features -> Drugs and Products is set to: _Inventory and sell drugs only_ (and saved)
2. Go to new top menu _Inventory_ -> _Management_
3. Fill in the "Add drug" form with any data and a template, click _Add_

**How to replicate**

1. Select a patient
2. Select _Prescriptions_, click Pencil/Edit symbol
3. Click _Add_
4. Select an existing drug template from the inventory in "in-house"
5. Enter unit and price on the top right
6. Click _Save and Dispense_
7. Error: **"Internal error, no drug ID specified!"**

### Expected behavior

I expect the drug to be added and the uints to be deducted from the inventory.

### Client configuration

- OpenEMR version: 7.0.4-dev

** I've already found the error and I'm working on a fix **

cburnicki:
Fixes #8597

Currently, drugs cannot be disposed from inventories because of a bug in a template file where instead of the value of a variable, the variable name is set as an option value in a select element.

Add `{{}}` around the variable so its value (instead of its name) is rendered as the option value.

I also improved an error message I encountered multiple times. The "sanity check" usually fails because no encounter has been selected or created, not because of an integrity issue. I improved the check and implemented an error message that clearly communicates the issue.

No

### DIFF
--- a/interface/drugs/drugs.inc.php
+++ b/interface/drugs/drugs.inc.php
@@ -14,6 +14,8 @@
 // openemr/library/classes/Prescription.class.php file.
 
 use PHPMailer\PHPMailer\PHPMailer;
+use OpenEMR\Services\EncounterService;
+
 
 // Decision was made in June 2013 that a sale line item in the Fee Sheet may
 // come only from the specified warehouse. Set this to false if the decision
@@ -44,6 +46,30 @@ function send_drug_email($subject, $body)
     }
 }
 
+/**
+ * Validates that an encounter exists for a given patient
+ *
+ * @param int $patient_id The patient ID
+ * @param int $encounter_id The encounter ID to validate
+ * @return bool|string Returns true if valid, or an error message if invalid
+ */
+function _validateEncounter($patient_id, $encounter_id)
+{
+    if (empty($encounter_id) || $encounter_id == 0) {
+        return xlt('No active encounter selected. To dispense medication, you must first select or create a valid encounter for this patient. Please return to the patient dashboard, select or create an encounter, then try again.');
+    }
+
+    $encounter = new EncounterService()->getOneByPidEid($patient_id, $encounter_id);
+    
+    if (empty($encounter)) {
+        // This is a system error - the encounter ID exists but doesn't match the patient
+        error_log("Drug dispensing error: Encounter $encounter_id does not exist or does not belong to patient $patient_id");
+        return xlt('System error: The selected encounter does not exist or does not belong to this patient. This may indicate a data integrity issue. Please contact support if this problem persists.');
+    }
+    
+    return true;
+}
+
 function sellDrug(
     $drug_id,
     $quantity,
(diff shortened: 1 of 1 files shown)


interface/drugs/drugs.inc.php
```diff
@@ -44,6 +44,33 @@ function send_drug_email($subject, $body)
...
 }
 
+/**
+ * Validates that an encounter exists for a given patient
+ *
+ * @param int $patient_id The patient ID
+ * @param int $encounter_id The encounter ID to validate
+ * @return bool|string Returns true if valid, or an error message if invalid
+ */
+function _validateEncounter($patient_id, $encounter_id)
```
stephenwaite, line 56: hi @cburnicki, can use `private function validateEncounter()` instead of old style `_validateEncounter`

interface/drugs/drugs.inc.php
```diff
@@ -44,6 +44,33 @@ function send_drug_email($subject, $body)
...
+    if (empty($encounter_id) || $encounter_id == 0) {
+        return xlt('No active encounter selected. To dispense medication, you must first select or create a valid encounter for this patient. Please return to the patient dashboard, select or create an encounter, then try again.');
+    }
+
+    $res = sqlQuery(
+        "SELECT count(*) AS count from form_encounter WHERE pid = ? AND encounter = ?",
+        array($patient_id, $encounter_id)
+    );
+if (mysqli_num_rows($res) > 0) {
+    if (empty($res['count'])) {
```
stephenwaite: would replace above sql with EncounterService, `use OpenEMR\Services\EncounterService;` before above PhpMailer 
```
$res = (new EncounterService())->getOneByPidEid($pid, $encounter);

File: templates/prescription/general_edit.html.twig [hunk 3]
```diff
@@ -190,7 +190,7 @@
             <label>&nbsp; {{'in-house'|xlt}}</label>
             <select class="input-sm form-control" name="drug_id" onchange="drugselected(this)">
                 {% for key,drug in DRUG_ARRAY_VALUES %}
-                    <option value="drug"
+                    <option value="{{ drug|attr }}"
```
stephenwaite, line 193: bird_jive

cburnicki:
Hi thank you @stephenwaite, I will make the suggested changes. Can you give me a hint why all these seemingly unrelated test pipes are failing on all my PRs? Am I missing something or is this just the way it is?

stephenwaite:
can ignore inferno and phpstan as they're works in progress

interface/drugs/drugs.inc.php
cburnicki: Hey @stephenwaite, the way I understand it, this is only possible in classes. I am not a php dev though so maybe I am getting it wrong.

I have also been thinking about refactoring the whole thing and make it a service since the functionality and design right now is... not ideal... and a lot of the functionality is quite error prone and also lacks proper error handling.

However, since this would take quite some time and I'm not yet sure if I have that much time, I wanted to just improve the approach already implemented.

stephenwaite: ok, thanks for correcting me on that

interface/drugs/drugs.inc.php
```diff
@@ -44,6 +44,30 @@ function send_drug_email($subject, $body)
...
+ * @param int $encounter_id The encounter ID to validate
+ * @return bool|string Returns true if valid, or an error message if invalid
+ */
+function _validateEncounter($patient_id, $encounter_id)
+{
+    if (empty($encounter_id) || $encounter_id == 0) {
+        return xlt('No active encounter selected. To dispense medication, you must first select or create a valid encounter for this patient. Please return to the patient dashboard, select or create an encounter, then try again.');
+    }
+
+    $encounter = new OpenEMR\Services\EncounterService()->getOneByPidEid($patient_id, $encounter_id);
```
stephenwaite: minor fix is to replace with 
```
use OpenEMR\Services\EncounterService;
```
at the top of the file
and then can call it without the path
```
$encounter = (new EncounterService())->getOneByPidEid($patient_id, $encounter_id);

PR #8610, opened 2025-07-13
fix: couple function void fixes

### LINKED ISSUE #8609: fix: couple function void fixes
couple function void fixes

bradymiller:
fixes #8609


bradymiller:
fixes broken syntax, so bringing in.

PR #8619, opened 2025-07-14
fix(php scripts): avoid unequivocal early exit

### LINKED ISSUE #8618: Early `exit` triggers DCE, inhibits static analysis
Several of the php scripts in openemr use an [early-exit pattern like this](https://github.com/openemr/openemr/blob/72928da7f1bbf8fdfa56962f26d99ef79e96b5ec/contrib/util/installScripts/InstallerAuto.php#L78):

```php
// This exit is to avoid malicious use of the script.
exit;
```

When rector or any analyzer with dead code elimination (DCE) reads this unequivocal end-of-script, it wants to delete everything after it. This will block us going to rector level 4 and up.

To be honest, I'm also skeptical of the utility of this kind of check in any case, but I propose changing these to a conditional exit based on an environment variable. That way, instead of having to do a `sed` command to rewrite the script, we can run the script in question by setting the env var.

kojiromike:
Fixes #8618

This can be merged after the two changes below:

- [x] update openemr-devops scripts except obsolete (openemr/openemr-devops#468)
- [x] update https://github.com/openemr/demo_farm_openemr/blob/master/demo_build.sh (openemr/demo_farm_openemr#75)

Avoid triggering dead code elimination and requiring `sed` to activate scripts.

- Replace `exit` with environment condition on scripts that need a check to start.
- Use that method in the ciLibrary

### DIFF
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -124,11 +124,9 @@ install_configure() {
     	actions_chmod -R 0777 sites/default/documents
     )
     # we need this to occur on the docker image of OpenEMR, on local systems this may be the same but in the ci engine its inside the docker container
-    _exec sh -c "sed -e 's@^exit;@ @' < ./contrib/util/installScripts/InstallerAuto.php > ./contrib/util/installScripts/InstallerAutoTemp.php"
     # until we get the devops repo setup we'll just grab the auto_configure.php script and use that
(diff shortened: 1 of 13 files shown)


bradymiller:
hi @kojiromike . looks good and bringing in. thanks!

PR #8623, opened 2025-07-16
ci(styling): actions annotations

### LINKED ISSUE #8624: Present Github actions findings as annotations
The following Github actions workflows produce output that isn't rendered into an annotation. We want to make the output available as annotations, especially so they can appear inline with the problematic line of code, and make action reports easier to read. Please add workflow summaries as well, for the most useful output for each workflow.

1. api-docs.yml
2. composer-require-checker.yml
3. composer.yml
4. syntax.yml

For implementation, create files that fail the checks for testing. Use [problem matchers](https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md) to convert the output of each tool into annotations.

kojiromike:
Fixes #8624 

Present styling action findings as github annotations

- [x] ci(styling): actions annotations
- [ ] test(styling): deliberate style breakages (to be removed before merge)

### DIFF
--- a/.github/workflows/styling.yml
+++ b/.github/workflows/styling.yml
@@ -19,7 +19,8 @@ jobs:
     strategy:
       matrix:
         # Single-element matrix provides named variable and job title
-        php-version: ['8.4']
+        php-version:
+        - '8.4'
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


kojiromike:
Sigh, TIL actions/toolkit#2105

PR #8627, opened 2025-07-16
feat(Crypto): Crypto Strategy

### LINKED ISSUE #8626: Alternative Encryption Strategies
I would like to open the door to the possibility of alternative methods of encryption besides the existing implementation in src/Common/Crypto/CryptoGen.php. The approach I will use is to implement a Strategy Pattern with dependency injection so that alternative strategies can be injected by users.

In the initial implementation, users must understand that there is no automatic migration method for transitioning existing data from one strategy to another. That would have to be worked out on a strategy-pair-by-strategy-pair basis.

kojiromike:
Fixes #8626

Todo:

- [x] introduce the ability to select crypto strategy at install time
- [x] document 
- [x] explore possibility of injecting alternative strategies via modules (lower priority?)
- [x] write explicit tests for strategies (as test coverage may not capture crypto calls through database serialized code)
- [ ] enable selecting crypto strategy in docker development environments (openemr/openemr-devops#470)
- [ ] additional manual testing

Implement the strategy pattern for CryptoGen and add a Null crypto strategy

- Create an encryption/decryption interface based on CryptoGen
- Move the implementation from CryptoGen into CryptoGenStrategy, which implements the interface
- Implement strategy injection into the existing CryptoGen, and use the relocated implementation
- Implement a null crypto implementation

### DIFF
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -61,6 +61,46 @@ class CryptoGen
     }
 
     /**
+     * Get the encryption strategy name from the keys table.
+     *
+     * @return string|null The strategy name, or null if not found
+     */
+    public static function getEncryptionStrategyName(): ?string
+    {
+        if (!self::ready()) {
+            return null;
+        }
+
+        $result = sqlQueryNoLog("SELECT value FROM keys WHERE name = 'encryption_strategy_name'");
+        return $result['value'] ?? null;
+    }
+
+    /**
+     * Set the encryption strategy name in the keys table.
+     *
+     * @param string $strategyName The strategy name to store
+     * @return bool True on success, false on failure
+     */
+    public static function setEncryptionStrategyName(string $strategyName): bool
+    {
+        if (!self::ready()) {
+            return false;
+        }
+
+        try {
+            // Use INSERT ... ON DUPLICATE KEY UPDATE for MySQL compatibility
+            sqlStatementNoLog(
+                "INSERT INTO keys (name, value) VALUES ('encryption_strategy_name', ?) ON DUPLICATE KEY UPDATE value = VALUES(value)",
+                [$strategyName]
+            );
+            return true;
+        } catch (Exception $e) {
+            error_log("CryptoGen: Failed to set encryption strategy name: " . $e->getMessage());
+            return false;
+        }
+    }
+
+    /**
      * Constructor - Initialize CryptoGen with database-stored encryption strategy.
      *
      * Loads the encryption strategy from the database global setting.
@@ -169,7 +209,7 @@ class CryptoGen
 
 
     /**
-     * Load encryption strategy from database global setting.
+     * Load encryption strategy from database keys table.
      *
      * Note: This method is only called once during construction and the result
      * is stored in $this->encryptionStrategy.
@@ -179,30 +219,28 @@ class CryptoGen
      */
     private function loadStrategyFromDatabase(): EncryptionStrategyInterface
     {
-        // Check if we have a database connection and the globals table exists
+        // Check if we have a database connection
         if (!self::ready()) {
             throw new CryptoGenException("Fatal error: No database connection available for loading encryption strategy");
         }
 
         $encryptionStrategySelector = new EncryptionStrategySelector();
 
-        // Query for the encryption strategy name from global setting
-        $result = sqlQueryNoLog("SELECT gl_value FROM globals WHERE gl_name = 'encryption_strategy_name' AND gl_index = 0");
+        // Get the encryption strategy name from keys table
+        $strategyName = self::getEncryptionStrategyName();
 
-        if (empty($result['gl_value'])) {
+        if (empty($strategyName)) {
             $this->systemLogger->debug("CryptoGen: No encryption strategy name stored in database, setting default");
             // Set the default strategy in the database
             $defaultStrategy = $encryptionStrategySelector->getDefaultStrategy();
             $defaultStrategyId = $defaultStrategy->getId();
-            sqlStatementNoLog("INSERT INTO globals (gl_name, gl_index, gl_value) VALUES ('encryption_strategy_name', 0, ?)", [$defaultStrategyId]);
+            self::setEncryptionStrategyName($defaultStrategyId);
             $this->systemLogger->debug(
                 sprintf("CryptoGen: Set default encryption strategy in database to '%s'", $defaultStrategyId),
                 ['strategy_name' => $defaultStrategyId]
             );
             return $defaultStrategy;
         }
-
-        $strategyName = $result['gl_value'];
 
         if (!is_string($strategyName)) {
             throw new CryptoGenException("Fatal error: Encryption strategy name from database is not a string");
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -1,121 +1,149 @@
 <?php
 
 /**
- * CryptoGen class.
+ * CryptoGen class - Facade for encryption/decryption operations.
  *
- *   OpenEMR encryption/decryption strategy:
- *    1. Two separate private key sets are used, one key set in the database and one key set on the drive.
- *    2. The private database key set is stored in the keys mysql table
- *    3. The private drive key set is stored in sites/<site-name>/documents/logs_and_misc/methods/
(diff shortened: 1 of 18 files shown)


src/Common/Crypto/NullEncryptionStrategy.php
```diff
@@ -0,0 +1,46 @@
...
+
+use OpenEMR\Common\Crypto\EncryptionStrategyInterface;
+
+class NullEncryptionStrategy implements EncryptionStrategyInterface
+{
+    private string $nullVersion = "000";
+
+    public function encryptStandard(?string $value, ?string $customPassword = null, string $keySource = 'drive')
+    {
+        return ($value === null) ? null : $this->nullVersion . base64_encode($value);
```
adunsulag: I'm assuming you did this so we could keep track of the fact that the field that was encrypted, instead of just doing an identity function here (return the exact same value)?  That way people could migrate the encryption to a different strategy if needed?

src/Common/Crypto/CryptoGen.php
```diff
@@ -55,8 +57,16 @@ class CryptoGen
     #  from the drive).
     private array $keyCache = [];
 
-    public function __construct()
+    private EncryptionStrategyInterface $encryptionStrategy;
+
+    public function __construct(?EncryptionStrategyInterface $encryptionStrategy = null)
+    {
+        $this->encryptionStrategy = $encryptionStrategy ?? new CryptoGenStrategy();
```
adunsulag: I suppose you could hit the $GLOBALS['kernel']->getEventDispatcher() here as I imagine injecting this everywhere CryptoGen is used is going to be a really big pain.  At some point perhaps we can get around the global, but that'd be what I'd do in the short term to be able to get a module to be able to override this.

src/Common/Crypto/CryptoGen.php
```diff
@@ -124,15 +103,7 @@ public function decryptStandard(?string $value, ?string $customPassword = null,
...
-
-        if (preg_match('/^00[1-6]/', $value)) {
-            return true;
-        } else {
-            return false;
-        }
+        return $this->encryptionStrategy->cryptCheckStandard($value);
     }
 
     /**
```
adunsulag, line 78: Do you plan on dropping the coreEncrypt function in this class?  I'm assuming so now that the strategy handles everything.

src/Common/Crypto/CryptoGen.php
adunsulag: I suppose another option would be as an environment variable.

kojiromike: I'll do an event, but we should consider what happens if someone introduces a new strategy via an event. Changing global encryption strategy after there's encrypted data could be catastrophic. How can we prevent it?

adunsulag: Hmm.. we'll have to think on this as that is a problem.  Event priority levels and telling the event to stop cascading could mitigate that if the module that defines the encryption is the first one loaded and executed on event dispatching.  Could be a brittle solution though unless the module is added as an always loaded module (this functionality right now is only available through a Zend module).

I thought about using a config variable that could store the FQCN of the strategy, but anything in the database can be easily changed via another module.

src/Common/Crypto/NullEncryptionStrategy.php
kojiromike: yeah, but it really isn't compatible with our goal of being able to use encryption outside of openemr altogether, so I'm not keeping it.

src/Common/Crypto/CryptoGen.php
kojiromike: yes, good catch

src/Common/Crypto/CryptoGen.php
kojiromike: I think I need to rethink this. We should not have an event that lets modules change the strategy at cryptogen constructor time. We should only have a utility for the installer to use that itself uses an event so that modules can register strategies at install time. The only chance to select a strategy will be at install time. Then, strategies must be Serializable, and we'll store the implementation in the database so that even after a module is uninstalled, data can still be decrypted.

kojiromike: Update: Now this system works by storing the implementation in the database at install time. Once that's done, the crypto code on the filesystem mostly doesn't matter.

adunsulag: @kojiromike I'm trying to think through this, if a module is uninstalled, won't the class definition still be required to unserialize the implementation?  For example in the case of a module being uninstalled and the class definitions are no longer present, I didn't think it could be unserialized as the class definition can no longer be found.  I don't think it can be unserialized as just an interface, but the actual class definition had to be loaded.  If it can just do it based on the interface, is this a new feature in PHP?

adunsulag: I will mention I agree on doing the *strategy selection* at installation time, I think that's a great idea and much better approach.

* edited to swap event out to strategy selection

kojiromike: @adunsulag you're too right about needing to always have the class definition to unserialize. I'm giving up on making this work with modules – it's too risky to people's data. If someone has a good enough need for a new encryption strategy, they can propose it go into core. I know we want a lot of things to be modules, but perhaps the encryption system, specifically, is too ingrained to be a good candidate.

library/globals.inc.php
```diff
@@ -2470,6 +2470,13 @@ function gblTimeZones()
             xl('Authentication SHA512 hash rounds number.')
         ),
 
+        'encryption_strategy_name' => array(
+            xl('Encryption Strategy Name') . '*',
+            'text',
+            'cryptogen',
+            xl('Name of the encryption strategy selected during installation. This setting should not be modified after installation.')
+        ),
```
bradymiller: Seems odd that we even want anybody to touch this since it is set on install. is this here just to store it? If so, prob best to store it somewhere else (maybe even the version table or a new specific table that can grow if need to support new strategies (ie. if need to store things for it).

src/Common/Crypto/README.md
```diff
@@ -0,0 +1,264 @@
...
+
+1. **Strategy Discovery** - `EncryptionStrategySelector` dispatches `EncryptionStrategyRegistrationEvent::STRATEGY_REGISTRATION`
+2. **Module Registration** - Modules can register available strategies via event listeners
+3. **Strategy Selection** - User/installer selects strategy (default: `cryptogen`)
+4. **Serialization** - Selected strategy is serialized and stored in `globals.encryption_strategy_serialized`
+5. **Runtime Loading** - `CryptoGen` loads and deserializes strategy from database
+
+### Available Installation Options
+
+- **Web Installation** - Strategy selection integrated into installation UI
```
bradymiller: I don't even think i would allow selection of this on the UI since those users have much less technical adeptness (if a user unknowingly chooses null then would be a complete disaster) and the autointaller via docker/orchestration is where this will almost always come into play. If needed, could support a get parameter via setup.php to open up the selector for edge cases.

src/Common/Crypto/CryptoGen.php
bradymiller: agreed. less moving and unknown parts, the better.

library/globals.inc.php
kojiromike: @bradymiller hmm, I don't want anyone to touch it. I just want admins to be able to know what it's set to, without being able to change it.

bradymiller:
hi @kojiromike ,
Just checking if the strategy is still being stored in the database (was unable to find it in the PR)?

bradymiller:
hi @kojiromike , Did some digging. Looks like maybe storing it in globals table (at least see where getting it at global encryption_strategy_name). Rec putting it outside globals table for 2 reasons:
1. Globals is meant to be paired with globals.inc.php and the config user interface in OpenEMR. I am guessing there are chances that things not in globals.inc.php may get removed somewhere in the codebase.
1. Globals gets lots of read/write and this is pretty critical (best for users that do not need to see it to not see it and if a module does something bad with globals then this could amplify issues).

My thought would be to consider storing it the selection in the keys table (is a vertical table) and then display the selection in the Admin->System->Diagnostics .
1. Then looking at Admin->System->Diagnostics can provide a sanity check.
2. And looking at keys table can also provide a sanity check.

kojiromike:
This PR is ostensibly passing tests, but it should not be -- once #8754 is resolved, it will fail "correctly" and the issue then can be fixed.

PR #8629, opened 2025-07-16
ci(actions): enable caching

### LINKED ISSUE #8625: Php and node actions should use caching
Several github actions would benefit from caching

kojiromike:
Fixes #8625

Enable caching for php, node actions

- Enable caching for php, node actions
- Remove redundant storage of result when it's the natural behavior of the shell
- Replace find with git ls-files.

### DIFF
--- a/.github/workflows/inferno-test.yml
+++ b/.github/workflows/inferno-test.yml
@@ -51,6 +51,22 @@ jobs:
               with:
                   node-version: '20'
 
+            - name: Cache Composer dependencies
+              uses: actions/cache@v4
+              with:
+                  path: ~/.composer/cache/files
+                  key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+                  restore-keys: |
+                      ${{ runner.os }}-composer-
+
+            - name: Cache node modules
+              uses: actions/cache@v4
+              with:
+                  path: ~/.npm
+                  key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+                  restore-keys: |
+                      ${{ runner.os }}-node-
+
             - name: Set up Docker Buildx
               uses: docker/setup-buildx-action@v3
 
--- a/.github/workflows/styling.yml
+++ b/.github/workflows/styling.yml
@@ -25,21 +25,19 @@ jobs:
       - name: Report PHP Version
         run: php -v
 
+      - name: Cache Composer dependencies
+        uses: actions/cache@v4
+        with:
+          path: ~/.composer/cache/files
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-composer-
+
       - name: Install php_codesniffer
-        run: |
-          failTest=false
-          composer global require "squizlabs/php_codesniffer=3.*" || failTest=true
-          if $failTest; then
-            exit 1
-          fi
+        run: composer global require "squizlabs/php_codesniffer=3.*"
 
       - name: Check PHP Styling
-        run: |
-          failTest=false
-          $HOME/.composer/vendor/bin/phpcs -p -n --extensions=php,inc --report-width=120 --standard=ci/phpcs.xml --report=full . || failTest=true
-          if $failTest; then
-            exit 1
-          fi
+        run: $HOME/.composer/vendor/bin/phpcs -p -n --extensions=php,inc --report-width=120 --standard=ci/phpcs.xml --report=full .
 
   css_styling:
     runs-on: ubuntu-24.04
@@ -52,21 +50,19 @@ jobs:
         with:
           node-version: '20'
 
+      - name: Cache node modules
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
       - name: npm install
-        run: |
-          failTest=false
-          npm install || failTest=true
-          if $failTest; then
-            exit 1
-          fi
+        run: npm ci
 
       - name: Check CSS Styling
-        run: |
-          failTest=false
-          npm run stylelint || failTest=true
-          if $failTest; then
-            exit 1
-          fi
+        run: npm run stylelint
 
   js_linting:
     runs-on: ubuntu-24.04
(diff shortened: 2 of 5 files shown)


.github/workflows/inferno-test.yml
```diff
@@ -51,6 +51,22 @@ jobs:
               with:
                   node-version: '20'
 
+            - name: Cache Composer dependencies
+              uses: actions/cache@v4
+              with:
+                  path: ~/.composer/cache/files
+                  key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
```
bradymiller: should this be composer.lock instead of composer.json?

.github/workflows/inferno-test.yml
```diff
@@ -51,6 +51,22 @@ jobs:
...
+                  restore-keys: |
```
bradymiller, line 59: What are your thoughts on not using the restore-keys? Looks like there is potential risk with this where the ci will not be giving same builds as the devs. Still going to get good performance when hit the same hash for the locks, which is what happens most of the time.

.github/workflows/syntax.yml
```diff
@@ -29,10 +29,5 @@ jobs:
...
       - name: Check PHP Syntax
-        run: |
-          failSyntax=false
-          if find . -type f -name "*.php" -exec php -l {} \; 2>&1 >/dev/null | grep "^"; then failSyntax=true; fi
-          if find . -type f -name "*.inc" -exec php -l {} \; 2>&1 >/dev/null | grep "^"; then failSyntax=true; fi
-          if $failSyntax; then
-            exit 1
-          fi
+        shell: bash
+        run: git ls-files -z '*.inc' '*.php' | xargs -0 php -l >/dev/null
```
bradymiller, line 33: kung-fu-baby-funny

kojiromike: One of the more obscure behaviors of github actions is that if you set `shell: bash` you get the shell bash, and it sets the `pipefail` option, but if you don't set a shell explicitly, you still get bash, and it doesn't set the pipefail option. https://github.com/actions/runner-images/issues/4459

.github/workflows/inferno-test.yml
kojiromike: I'm not sure I understand the concern, but if anything, using restore-keys will help resolve a possible issue where we need to cache two different composer.locks for different workflow timeframes. Suppose someone creates a PR to add a new dependency and their composer.lock doesn't match the hash. Then `restore-keys` will restore the nearest cache it can find. We'll get a cache miss, but composer can still take advantage of all the other packages in the cache. Then, if their job builds, there will be a new hash, and if they don't change the lockfile again, it'll be a cache hit on the next round.

bradymiller: Just seems like we are asking for problems if do the restore-keys across the different installs and branches (going up and down in package versions doesn't seem natural, lead to bloat of the cache as multiple versions of packages get cached, doesn't mimick the current dev workflow, and may lead to some not very fun debugging during breaks). composer.lock also does not change that often, so it really doesn't come with much benefit.

kojiromike: @bradymiller composer.lock should change whenever composer.json changes, _and_ whenever a transitive dependency gets updated, such as by dependabot. So it makes sense to use that for the cache key. But also, that's the key that's used for a perfect cache hit, when nothing has changed. That's exactly when we benefit the most from the cache. So in those circumstances, there's no reason not to take advantage, and also restore-keys isn't in play there. (Despite the outdated view of the above code, we are using composer.lock and not composer.json here.)

In cases where there's no cache hit for `${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}`, the cache action will restore the ~/.composer/cache/files directory for some cache entry that matches (probably the most recent one). This is also harmless, if less helpful than a perfect cache hit, because composer still takes charge to ensure that the packages that get installed are the ones from composer.lock. Packages that are in the restored cache in the correct version will get used, but ones that are outdated or missing will not. So composer will take advantage where it can, but not as much as a perfect hit.

Your concern seems to be about as if we were restoring the `vendor` directory directly, but that's not what's happening here. This PR is about restoring a substrate of composer's internal cache, so that it can install packages faster. Composer's still ultimately going to decide what to restore. With both Github and composer able to invalidate the cache, and composer able to invalidate it on a per-package basis, I don't think there's any reason to be concerned about stale packages impacting CI here.

bradymiller: hi @kojiromike ,

Thanks for the explanation and after looking through documentation and discussing with Claude agree safe to use (ie. only the correct packages will be used) restore-keys for composer and npm.

The only issue/question I have remaining for restore-keys is the cache space. Won't the cache just keep getting bigger and bigger (since unused packages are not removed from cache) over time. Since the cache will be used frequently enough, it will always use an older cache when the lock files are changed and add the new packages to it. Just seems like these indefinitely enlarging caches for composer and npm this will cause problems over time as we have more uses for github cache (looks like repo cache limit is 10GB).

bradymiller: Actually, to answer my own question, it looks like caches are cleared after 7 days if not used, so that will stop my issue above (since we have periods of time that last more than 7 days between package lock changes thus avoiding my scenario above). So disregard my note above.

bradymiller (approved):

bradymiller:
hi @kojiromike , thanks for the improvement! just brought in.

.github/workflows/styling.yml
```diff
@@ -25,21 +25,19 @@ jobs:
       - name: Report PHP Version
         run: php -v
 
+      - name: Cache Composer dependencies
```
bradymiller, line 28: I of course noted this issue after i brought this in since i am now a cache nerd. I don't think this is right since the composer here is not using the composer.lock .
'''
      - name: Install php_codesniffer
        run: composer global require "squizlabs/php_codesniffer=3.*"
'''
So in this case, I think the composer cache will just be for squizlabs/php_codesniffer and it's dependencies and won't jive with the composer cache in the inferno cache. And since github cache is immutable, on a composer.lock change, either the composer.lock stuff will get the cache (via inferno) as intended or the squizlabs/php_codesniffer will steal it. Inferno in that case would think it has a hit when it doesn't in future and basically need to download all the composer packages :) (if my brain is thinking through this correctly). If this is case, then squizlabs/php_codesniffer stuff prob needs it's own cache.

PR #8630, opened 2025-07-16
ci(cache): cache docker actions

### LINKED ISSUE #8625: Php and node actions should use caching

### LINKED ISSUE #8631: Docker actions should use caching
The actions that do docker builds could benefit from caching

kojiromike:
Fixes #8631

Introduces docker build caching to github actions

- Add caching to docker builds in github actions

### DIFF
--- a/.github/workflows/build-dev-php-fpm-docker.yml
+++ b/.github/workflows/build-dev-php-fpm-docker.yml
@@ -32,4 +32,5 @@ jobs:
           tags: openemr/dev-php-fpm:pre-build-dev-85
           platforms: linux/amd64
           push: true
-          no-cache: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/weekly-build-php-fpm-dockers.yml
+++ b/.github/workflows/weekly-build-php-fpm-dockers.yml
@@ -33,7 +33,8 @@ jobs:
           tags: openemr/dev-php-fpm:8.1
           platforms: linux/amd64
           push: true
-          no-cache: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
   build_8_1_redis:
     # Only run from master branch on the main repository
@@ -58,7 +59,8 @@ jobs:
           tags: openemr/dev-php-fpm:8.1-redis
           platforms: linux/amd64
           push: true
-          no-cache: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
   build_8_2:
     # Only run from master branch on the main repository
@@ -83,7 +85,8 @@ jobs:
           tags: openemr/dev-php-fpm:8.2
           platforms: linux/amd64
           push: true
-          no-cache: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
   build_8_2_redis:
     # Only run from master branch on the main repository
@@ -108,7 +111,8 @@ jobs:
           tags: openemr/dev-php-fpm:8.2-redis
           platforms: linux/amd64
           push: true
-          no-cache: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
   build_8_3:
     # Only run from master branch on the main repository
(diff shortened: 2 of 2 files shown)


bradymiller:
hi @kojiromike , Researched this cache thing a bit and seems like there can be issues where packages can stale out. Note these things get built pretty fast so no real significant performance issues with them.

I asked Claude about this caching and to look at one of the Dockerfiles and here is reply:

Yes, absolutely! This Dockerfile has several issues where dependencies can become stale even if the base image doesn't update:

**Problem layers:**
```dockerfile
FROM php:8.4-fpm        # Base stays same
RUN apt-get update      # ← Cached layer with old package index
RUN apt-get install -y mariadb-client imagemagick curl  # ← Uses stale index
```

**What can go wrong:**

1. **Stale package repositories**: The `apt-get update` gets cached, so you're installing packages from an old index that might reference packages that no longer exist or have security vulnerabilities.

2. **External script caching**: 
```dockerfile
ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
```
This downloads once and caches forever. If there's a newer version with bug fixes, you won't get it.

3. **Node.js setup script**:
```dockerfile
RUN curl -sL https://deb.nodesource.com/setup_20.x | bash -
```
This could be cached with an old setup script.

The cache helps performance but can definitely cause issues with dependency staleness in long-running CI environments.

kojiromike:
@bradymiller yes, those are problems with the Dockerfiles indeed, but they aren't a problem for this PR per se -- they're a lower level problem. I create #8635 and a PR to address the first issue. The other two issues probably deserve a conversation of their own, but I don't think they're urgent.

bradymiller:
hi @kojiromike , brought in the #8635.
But still can't see how this docker caching provides benefit considering the potential harm. Been researching how it works and appears high risk of stale packages.
For example, posted a sample Dockerfiles below, which will go through:
```
#
# php-fpm Dockerfile build for openemr development docker environment
# This docker is hosted here: https://hub.docker.com/r/openemr/dev-php-fpm/ <tag is 7.2>
#
FROM php:8.3-fpm

# Add mariadb-client package that is needed in the OpenEMR Backup gui, which does direct command mysql commands
# Add imagemagick that is needed for some image processing in OpenEMR
# Note this basically add 160MB of space to the docker, so would be nice for OpenEMR to not require this stuff
#  and instead rely on php scripts, if possible.
RUN apt-get update \
 && apt-get install -y curl \
                       imagemagick \
                       mariadb-client

# Install correct version nodejs
RUN curl -sL https://deb.nodesource.com/setup_20.x \
  | bash - \
 && apt-get update \
 && apt-get install -y nodejs

# Add the php extensions (note using a very cool script by mlocati to do this)
ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
RUN chmod uga+x /usr/local/bin/install-php-extensions \
 && sync \
 && install-php-extensions calendar \
                           gd \
                           gettext \
                           intl \
                           ldap \
                           mysqli \
                           pdo_mysql \
                           soap \
                           sockets \
                           tokenizer \
                           xmlreader \
                           xsl \
                           zip

# Needed to ensure permissions work across shared volumes with openemr, nginx, and php-fpm dockers
RUN usermod -u 1000 www-data

# Copy over the php.ini conf
COPY php.ini /usr/local/etc/php/php.ini
```

The FROM layer will work nicely and will use cache unless there is a new version out. If this updates to new version, then all downstream layers are redone and taken from cache.

The install of the packages curl etc. will use cache in Perpetuity thus high risk of stale packages (only reinstalls and creates new cache when the above FROM layer is redone).

The add of the install-php-extensions will never take from cache since it is a internet call (and thus all layers after it will also be redone and not use cache; basically is a cachebuster) (_actually, upon more research docker may cache ok on this and keep a hash of the file to see if need to break the cache_).

The run of install-php-extensions would also cache in Perpetuity thus high risk of stale packages. This is not the case though in this dockerfile since it is always redone since a prior layer (the add layer always trigger a cachebust).

The run usermod works fine (albeit always rebuilt secondary to prior add layer).

The copy layers works fine (albeit always rebuilt secondary to prior add layer).


If my analysis is correct (definitely feel free to correct me if wrong), then don't see merit in this. Especially if things stop working, then would need to rethink all of above (which i will forget by tomorrow :) ) while trying to fix.

PR #8637, opened 2025-07-17
fix(Installer): phpstan level 1

kojiromike:
Run phpstan --level 1 on the installer and fix the found issues.


bradymiller:
hi @kojiromike , looks good and bringing in. thanks!

PR #8657, opened 2025-07-20
CCDA updates

sjpadgett:
Fixes #

Laminas\Db\TableGateway\Exception\RuntimeException
C:\xampp\htdocs\openemr\vendor\laminas\laminas-db\src\TableGateway\Feature\GlobalAdapterFeature.php:46
Message: No database adapter was found in the static registry.

inject database adapter from config  service factory.


interface/forms/clinical_notes/templates/partials/new/_clinical-notes-row.html.twig
```diff
@@ -60,7 +60,7 @@
                     <!-- Last edited date -->
                     <div class="col-12">
                         <label class="h6">{{ 'Last Updated'|xlt }}:</label>
-                        <span>{{ obj.last_updated }}</span>
+                        <span>{{ obj.last_updated|text }}</span>
```
bradymiller, line 63: also need to do to line 58 :)

interface/modules/custom_modules/oe-module-faxsms/templates/phone_widget.html.twig
```diff
@@ -3,9 +3,9 @@
 <div id="rc-container" class="d-flex flex-row align-items-center justify-content-between">
     <div id="rc-header">
         <script>
-            const RC_CLIENT_ID = "{{ clientId }}";
-            const RC_CLIENT_SECRET = "{{ clientSecret }}";
-            const RC_JWT = "{{ jwt }}";
+            const RC_CLIENT_ID = {{ clientId|js_escape }};
+            const RC_CLIENT_SECRET = {{ clientSecret|js_escape }};
+            const RC_JWT = {{ jwt|js_escape }};
```
bradymiller, line 8: 2DV

sjpadgett:
bringing in so available for testing on dev-demo

bradymiller:
looks like several phpstan issues :)

sjpadgett:
phpstan is wrong on these

interface/modules/custom_modules/oe-module-faxsms/library/setup_services.php
```diff
@@ -332,11 +349,68 @@ function toggleSetup(id, type = 'single') {
...
-        <!-- TODO refactor this to have vendor list a global array for future vendor additions -->
         <div class="frame col-12" id="set-service">
             <form id="set_form" name="set_form" class="form" role="form" method="post" action="">
                 <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken()); ?>" />
-                <div class="">
+                <div class="form-group">
+                    <?php if (isset($permissions_saved) && $permissions_saved) { ?>
+                    <div class="alert alert-<?php echo attr($permissions_message_type ?? 'success'); ?> text-center alert-dismissible fade show" role="alert">
+                        <strong>
+                            <?php if ($permissions_message_type === 'warning') { ?>
```
bradymiller, line 359: try:
```
<?php if (($permissions_message_type ?? '') === 'warning') { ?>
```
(it looks like the $permissions_message_type = '' in above code may not always be run)

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -57,7 +57,7 @@ class EncounterccdadispatchTable extends AbstractTableGateway
     public $searchFiltered = false;
     private $encounterFilterList = [];
 
-    public function __construct()
+    public function __construct($dbAdapter)
     {
```
bradymiller, line 61: - Looks like the call to this class does not have a parameter in interface/modules/zend_modules/module/Carecoordination/config/module.config.php
- Also looks like the constructor does nothing with this parameter

interface/modules/zend_modules/module/Carecoordination/config/module.config.php
```diff
@@ -155,7 +155,7 @@
                 return new CarecoordinationTable();
             },
             EncounterccdadispatchTable::class =>  function (ContainerInterface $container, $requestedName) {
-                    return new EncounterccdadispatchTable();
+                    return new EncounterccdadispatchTable($container->get(\Laminas\Db\Adapter\Adapter::class));
```
bradymiller, line 158: oops, i see there is now a parameter :)

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
bradymiller: Should the constructor do something with the parameter (ie. call the parent constructor or something else?)

bradymiller: The third issue needs to be removed from the baseline (ie. you fixed something in module.config.php from the baseline of issues).
@kojiromike , what is best way for us to do this?

sjpadgett: I couldn't figure this out. If you remove the dbadapter from constructor original error happens.

bradymiller: here are options from Claude (if the parameter is truly a road to nowhere, then seem like option 4 makes sense):
The PHPStan error indicates that the `$dbAdapter` parameter in the constructor is not being used within the constructor method. Looking at the code, the constructor is indeed empty - it accepts the `$dbAdapter` parameter but doesn't do anything with it.

Here are a few ways to fix this issue:

## Option 1: Remove the unused parameter (if not needed)
If the `$dbAdapter` is not actually needed, simply remove it:

```php
public function __construct()
{
}
```

## Option 2: Use the parameter (most likely correct)
Since this class extends `AbstractTableGateway`, you probably need to pass the adapter to the parent constructor:

```php
public function __construct($dbAdapter)
{
    parent::__construct('your_table_name', $dbAdapter);
}
```

## Option 3: Store it as a property
If you need to use the adapter elsewhere in the class:

```php
private $dbAdapter;

public function __construct($dbAdapter)
{
    $this->dbAdapter = $dbAdapter;
    // or also call parent if needed:
    // parent::__construct('your_table_name', $dbAdapter);
}
```

## Option 4: Suppress the warning (not recommended)
If you must keep the parameter for interface compatibility but don't use it:

```php
public function __construct($dbAdapter)
{
    // Explicitly mark as unused to satisfy PHPStan
    unset($dbAdapter);
}
```

**Recommendation**: Option 2 is most likely what you want, since `AbstractTableGateway` typically requires a table name and database adapter. You'll need to specify the appropriate table name for your use case.

interface/modules/zend_modules/module/Carecoordination/config/module.config.php
sjpadgett: I meant in service you reference. Something weird about the ApplicationTable.

bradymiller: phpstan is going to make us age faster :)

bradymiller: gonna see what happens if remove both parameters :)

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
sjpadgett: option 2 makes sense. will check but I think parent may be this weird AbstractTableGateway I'll look into but doubt I'll do a PR.

interface/modules/zend_modules/module/Carecoordination/config/module.config.php
sjpadgett: It breaks ccda import init, trust me. That was original problem that Stephen did.

PR #8660, opened 2025-07-20
fix(phpstan): establish baseline

### LINKED ISSUE #8659: AuthAccount.php extends missing AccountDAO class
See…

- https://github.com/openemr/openemr/blob/9200c3f6c4b8a8e23f5c12ff51436bc266c6a18b/portal/patient/fwk/libs/verysimple/Phreeze/AuthAccount.php#L8
- https://github.com/openemr/openemr/blob/9200c3f6c4b8a8e23f5c12ff51436bc266c6a18b/portal/patient/fwk/libs/verysimple/Phreeze/AuthAccount.php#L45

PHPStan cannot analyze this class and I don't think it can be instantiated as-is. Should we remove it?

kojiromike:
Fixes #8659

Sets a baseline file for phpstan so that we can get a green checkmark. We can work through the issues listed in the file over time, while maintaining the passing checks for any new and changed code.

Note, this changeset also removes AuthAccount.php entirely, because it is missing a dependency and so has critical errors that cannot be ignored in the phpstan baseline.

- ci(phpstan): create phpstan.neon.dist as a baseline, so we can move past current issues
- fix(OemrUI): remove the render param – it is never used
- fix(AuthAccount): delete, it has a missing dependency and cannot work
- phpstan(PredisSessionHandler): add hints so it's compatible with base class


kojiromike:
This is frustrating. They all pass locally and then fail in actions.

bradymiller:
@kojiromike , This will be awesome when it's working. I have gotten to a point in my laziness where I just ask Claude to look at a PR and let me know what is wrong and how to fix it :)
(I tried this and it has some good debug pointers)

@sjpadgett , just making sure you are ok with removal of the portal/patient/fwk/libs/verysimple/Phreeze/AuthAccount.php which just appears to be a sample/example class that is not used in OpenEMR.

kojiromike:
> This is frustrating. They all pass locally and then fail in actions.

Apparently [this is a thing](https://phpstan.org/blog/phpstan-reports-different-errors-locally-ci-what-should-i-do).

kojiromike:
@bradymiller well, I flipped the script – now this passes in CI, but will still have errors locally. But anyway, even in that case it has fewer errors locally than it did before. Maybe we should merge it?

kojiromike:
OK, it's not how I like to do things in general, but I did figure out how to create a conditional include that uses PHP to decide which baseline to use.

bradymiller (approved):

bradymiller:
nice @kojiromike ! bringing in.

dance-bear

PR #8666, opened 2025-07-21
Allow for more millisecond digits in date search

mschoettle:

Allow for more millisecond digits in date search regex.

This matches the regex of the `instant` datatype: https://build.fhir.org/datatypes.html#instant

Also fix a typo in the error message related to the ISO format number.

No


adunsulag:
I think we should be fine taking this in.  Note there are inconsitencies in the specification.  On the fhir date search page it says the search format has 4 digit millisecond precision, but it also mentions supporting instant datetime which allows for 9 digit millisecond precision.  https://build.fhir.org/search.html#date

I'm hazarding a guess that I hard-coded it to 4 digit at the time because it didn't pass the inferno test suite.  If it fails the inferno test suite, we'll revert the change, but I think allowing for more precision and matching the instant date time on the specification is a good change.

adunsulag:
The change should have nothing to do with the failing test on php 8.2, but will double check.  If that passes will bring it in and see what happens with Inferno.

PR #8673, opened 2025-07-22
ci: fix caching and installation

### LINKED ISSUE #8672: Fixes to CI Caching
I realized too late is that we should be looking up the composer path instead of hard coding it -- that way we can restore a different path for the global cache than for the local path. See https://github.com/actions/cache/blob/main/examples.md#php---composer

Currently the logs are reporting

> Post job cleanup.
> Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

Also, in some cases the cache key is incorrect, such as when composer is being used to globally install phpcs in styling.yml

kojiromike:
Fixes #8672

Fix the issues preventing Github Actions from taking full advantage of caching.

- [x] chore(composer): add php_codesniffer to require-dev
- [x] ci(inferno-test): Remove php extension installation (not needed outside of docker)
- [x] ci: consistent whitespace and style
- [x] ci(js-test): Lookup npm cache dir
- [x] ci(phpstan): Lookup composer cache dir
- [x] ci(phpstan): Consistent composer install method
- [x] ci(rector): dry run against master, too
- [x] ci(rector): remove redundant composer specification
- [x] ci(styling): Use local composer phpcs
- [x] ci(styling): Lookup cache dir
- [x] ci(styling): consistent composer install command

### DIFF
--- a/.github/workflows/inferno-test.yml
+++ b/.github/workflows/inferno-test.yml
@@ -60,14 +60,23 @@ jobs:
           composer config cache-files-dir
         } >> "$GITHUB_OUTPUT"
 
-    - name: Cache Composer dependencies
-      uses: actions/cache@v4
+    - name: Composer Cache
+      uses: actions/cache/restore@v4
       with:
         path: ${{ steps.composer-cache.outputs.dir }}
         key: ${{ runner.os }}-composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.lock') }}
         restore-keys: |
           ${{ runner.os }}-composer-${{ matrix.php-version }}-
           ${{ runner.os }}-composer-
+
+    - name: Composer Install
+      run: composer install --prefer-dist --no-progress
+
+    - name: Save composer cache
+      uses: actions/cache/save@v4
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.lock') }}
 
     - name: Install npm package
       uses: actions/setup-node@v4
@@ -83,7 +92,7 @@ jobs:
         } >> "$GITHUB_OUTPUT"
 
     - name: Cache node modules
-      uses: actions/cache@v4
+      uses: actions/cache/restore@v4
       with:
         path: ${{ steps.npm-cache-dir.outputs.dir }}
         key: ${{ runner.os }}-node-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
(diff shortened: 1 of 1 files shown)


kojiromike:
The difference between a cache miss and a cache hit:

Cache miss:

image

Cache hit:

image

.github/workflows/phpstan.yml
```diff
@@ -17,15 +17,27 @@ jobs:
...
-        path: /tmp/composer-cache
+        path: ${{ steps.composer-cache.outputs.dir }}
         key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-composer-
     - name: Composer Install
-      uses: php-actions/composer@v6
-      with:
-        php_extensions: gd soap zip
+      run: composer install --prefer-dist --no-progress
```
bradymiller, line 45: For the composer cache to be compatible with inferno, would we also need `--prefer-dist` at https://github.com/openemr/openemr/blob/a26d6c49557e1c46ec52df1b8a3d88cf3d11c6e0/ci/ciLibrary.source#L91
(also, can these composer/npm caches be worked into test.yml stuff (guessing the ccda npm stuff will get it's own cache based on it own package.lock))?

kojiromike: I added the prefer-dist to ciLibrary, and the composer and npm caches.

The cache hash for npm is based on `'**/package-lock.json'` which IIUC will match all the package-lock.json files in the whole repo (including, it seems https://github.com/openemr/openemr/blob/master/interface/modules/custom_modules/oe-module-comlink-telehealth/public/assets/js/package-lock.json). `hashFiles('**/package-lock.json')` is a hash that will change if any of those three package-lock.json files change, which is pretty ideal, I think.

.github/workflows/test.yml
```diff
@@ -125,7 +125,10 @@ jobs:
           npm config get cache
         } >> "$GITHUB_OUTPUT"
     - name: Cache node modules
-      uses: actions/cache@v4
+      # Use explicit restore and save actions to control
+      # precisely when in the workflow the cache is
+      # restored and saved.
+      uses: actions/cache/restore@v4
```
stephenwaite, line 167: seal-dunk

.github/workflows/test.yml
```diff
@@ -110,6 +180,18 @@ jobs:
...
         . ci/ciLibrary.source
         ccda_build
+    - name: Save node cache
+      if: ${{ matrix.configurations.save_node_cache == 'true' }}
+      uses: actions/cache/save@v4
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ matrix.configurations.node_version }}-${{ hashFiles('**/package-lock.json') }}
+    - name: Save composer cache
+      if: ${{ matrix.configurations.save_composer_cache == 'true' }}
```
kojiromike, line 190: @adunsulag I was getting the same error in test.yml as you saw in the inferno tests, and I think it was a collision when too many jobs simultaneously tried to save the same cache/key. If you still can't get it to work on a retry, maybe we should just try to restore the cache, but never try to save it, for inferno.

bradymiller:
wowsers! bringing in. nice work @kojiromike !

PR #8676, opened 2025-07-23
Mws/integrate gcip auth

msummers42:
Draft for discussion

### DIFF
--- a/Documentation/GCIP_Integration.README.md
+++ b/Documentation/GCIP_Integration.README.md
@@ -49,5 +49,16 @@ This document outlines the plan for integrating Google Cloud Identity Platform (
 - [x] **Update Developer Documentation:**
   - [x] Document the new GCIP integration feature.
   - [x] Provide instructions on how to configure and use it.
-- [ ] **Update User Documentation:**
+- [x] **Update User Documentation:**
   - [ ] Explain how to use the GCIP login feature.
+
+## Phase 6: Security and Audit Logging
+
+- [x] **Security:**
+  - [x] Remove `google_client_secret` from the client management UI.
+  - [x] Encrypt `google_client_secret` in the database.
+- [x] **Audit Logging:**
+  - [x] Log GCIP configuration changes (create, update, delete).
+  - [x] Log GCIP login attempts, successes, and failures.
+  - [x] Log new user provisioning via GCIP.
+


kojiromike:
The phpstan errors have to be being caused by an autoloader problem that I haven't figured out yet (since it only happens in GH and not locally.)

kojiromike:
@msummers42 if you would, please…

- [ ] rebase upstream/master
- [ ] `composer install`
- [ ] autofix the style complaints (`php -d memory_limit=8G vendor/bin/phpcbf -p -n --extensions=php,inc --report-width=120 --standard=ci/phpcs.xml --report=full .`)
- [ ] autofix the rector complaints (`php -d memory_limit=8G vendor/bin/rector`)

src/Common/Auth/OpenIDConnect/Entities/ClientEntity.php
```diff
@@ -68,13 +68,49 @@ class ClientEntity implements ClientEntityInterface
...
     public function __construct()
     {
         $this->scopes = [];
         $this->skipEHRLaunchAuthorizationFlow = false;
         $this->isEnabled = false;
         $this->dsiType = self::DSI_TYPE_NONE;
+        $this->identityProvider = 'local';
+        $this->googleClientId = null;
+        $this->googleClientSecret = null;
+    }
```
kojiromike, line 84: IMO properties that can be set at the class should not be set in the constructor. Not only does PHP not have the same issue with mutables as Python does, but all properties are instance properties unless you explicitly use the `static` keyword. So I don't think this class needs a constructor at all.

src/Common/Auth/OpenIDConnect/Repositories/ClientRepository.php
```diff
@@ -293,4 +296,28 @@ public function saveSkipEHRLaunchFlow(ClientEntity $client, bool $skipFlow)
...
     }
+
+    public function persist(ClientEntity $client)
+    {
+        $googleClientSecret = $client->getGoogleClientSecret();
+        if (!empty($googleClientSecret)) {
+            $googleClientSecret = $this->cryptoGen->encryptStandard($googleClientSecret);
+        } else {
+            $existingClient = $this->getClientEntity($client->getIdentifier());
+            $googleClientSecret = $existingClient->getGoogleClientSecret();
```
kojiromike, line 307: Why is it implied that the secret is cleartext when the secret can be gotten on line 302, but it's encrypted when it's gotten from `$existingClient`?

src/RestControllers/AuthorizationController.php
```diff
@@ -581,6 +586,204 @@ public function oauthAuthorizationFlow(): void
...
+        ], JSON_THROW_ON_ERROR);
+
+        $authUrl = 'https://accounts.google.com/o/oauth2/v2/auth?' .
+            'response_type=code&' .
+            'client_id=' . urlencode($clientId) . '&' .
+            'redirect_uri=' . urlencode($redirectUri) . '&' .
+            'scope=' . urlencode($scope) . '&' .
+            'state=' . urlencode($state) . '&' .
+            'nonce=' . urlencode($nonce) . '&' .
+            'access_type=offline&prompt=consent';
```
kojiromike, line 630: I think this should use https://www.php.net/manual/en/function.http-build-query.php

src/RestControllers/AuthorizationController.php
```diff
@@ -581,6 +586,204 @@ public function oauthAuthorizationFlow(): void
...
+
+        header('Location: ' . $authUrl, true, 302);
+        exit;
```
kojiromike, line 633: Exit statements like this are an antipattern -- this function should return something. I think you can follow the $response object setup as in https://github.com/msummers42/openemr/blob/8f14a00733a6c28a9dec0363bbd86e16d531a653/src/RestControllers/AuthorizationController.php#L462 and set `withHeader` and then `withStatus(302)` or whatever, and return the response.

PR #8677, opened 2025-07-23
Mws/gcip auth custom module

msummers42:
Draft for discussion. This work is based on work in https://github.com/openemr/openemr/pull/8676 but the mutation to a custom-module was performed by Copilot as an experiment.

### DIFF
--- a/interface/modules/custom_modules/oe-module-gcip-auth/ModuleManagerListener.php
+++ b/interface/modules/custom_modules/oe-module-gcip-auth/ModuleManagerListener.php
@@ -1,0 +1,256 @@
+<?php
+
+/**
+ * GCIP Module Manager Listener
+ * 
+ * <!-- AI-Generated Content Start -->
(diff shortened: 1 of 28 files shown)


interface/modules/custom_modules/oe-module-gcip-auth/ModuleManagerListener.php
```diff
@@ -0,0 +1,256 @@
...
+namespace OpenEMR\Modules\GcipAuth;
+
+use OpenEMR\Events\ModuleManagerEvent;
+use OpenEMR\Modules\GcipAuth\Services\GcipAuditService;
+
+/**
+ * Module Manager Listener for GCIP Authentication
+ */
+class ModuleManagerListener
+{
```
sjpadgett, line 28: This won't work because Module Manager calls specific function names. See Weno module for an example.
Or you could improve how this class is called if wanted.

PR #8679, opened 2025-07-23
ci(phpstan): add phpstan-phpunit extension

### LINKED ISSUE #8682: Introduce phpunit extension to phpstan
The extension provides better precision as to the signatures and types of what are being analyzed. One problem we encountered was that Mocks are not the type of the thing they're mocking, but that causes type checks to fail when trying to pass a mock as an argument.

The extension provides various other features as well.

Details here: https://github.com/phpstan/phpstan-phpunit

kojiromike:
Fixes #8682

Requires #8683 

Some phpstan errors made me suspect that the missing phpunit extension might be responsible for a bunch of problems, but testing locally doesn't show me what I need to know, so trying it in CI... 

- [x] Add the phpstan-phpunit extension

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -88,11 +88,14 @@
         },
         "allow-plugins": {
             "openemr/oe-module-installer-plugin": true,
-            "php-http/discovery": false
+            "php-http/discovery": false,
+            "phpstan/extension-installer": true
         }
     },
     "require-dev": {
+        "phpstan/extension-installer": "^1.4",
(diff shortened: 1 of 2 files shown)


kojiromike:
Well, sadly, instead of invalidating a bunch of errors that are in the excludefile, this just came up with new errors.

bradymiller:
hi @kojiromike , looks good and bringing in. thanks!

PR #8681, opened 2025-07-24
fix: couple phpstan fixes

bradymiller:
fix: couple phpstan fixes

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/config/module.config.php
+++ b/interface/modules/zend_modules/module/Carecoordination/config/module.config.php
@@ -155,7 +155,7 @@ return array(
                 return new CarecoordinationTable();
             },
             EncounterccdadispatchTable::class =>  function (ContainerInterface $container, $requestedName) {
-                    return new EncounterccdadispatchTable();
+                return new EncounterccdadispatchTable($container->get(\Laminas\Db\Adapter\Adapter::class));
             },
             EncountermanagerTable::class =>  function (ContainerInterface $container, $requestedName) {
                     return new EncountermanagerTable();
--- a/interface/modules/zend_modules/module/Carecoordination/config/module.config.php
+++ b/interface/modules/zend_modules/module/Carecoordination/config/module.config.php
@@ -155,7 +155,7 @@ return array(
                 return new CarecoordinationTable();
             },
             EncounterccdadispatchTable::class =>  function (ContainerInterface $container, $requestedName) {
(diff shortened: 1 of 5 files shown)


interface/modules/zend_modules/module/Carecoordination/config/module.config.php
```diff
@@ -155,7 +155,7 @@
                 return new CarecoordinationTable();
             },
             EncounterccdadispatchTable::class =>  function (ContainerInterface $container, $requestedName) {
-                    return new EncounterccdadispatchTable($container->get(\Laminas\Db\Adapter\Adapter::class));
```
bradymiller, line 158: @sjpadgett ,
Just seeing what happens with phpstan here.
To confirm, the above line is needed for ccda to work, correct:
```
return new EncounterccdadispatchTable($container->get(\Laminas\Db\Adapter\Adapter::class));
```

sjpadgett: yep

sjpadgett: kind of a head scratcher, huh!

sjpadgett:
Now I really want to figure out why the adapter needs to be passed in config.

bradymiller:
Makes no sense since it literally does nothing :)
Passing phpstan by explicitly unsetting the variable which is never used in the first place :)

bradymiller:
I wonder if the class should not even extend AbstractTableGateway class???

sjpadgett:
Maybe right I'll test

bradymiller:
gotta go to bed over here. feel free to bring in or modify if needed. phpstan will not defeat me :)

sjpadgett:
Okay thanks.

sjpadgett:
I figured this out and it stems from the ApplicationTable being made global and called in several class as $table = new ApplicationTable(); For some reason unless an adapter is included in factory EncounterccdadispatchTable(adapter) the GlobalAdapterFeature::getStaticAdapter() in constructor will not return an adapter!

Unless we(I) want to do a bunch of refactoring several classes we will have to live with a global ApplicationTable().

sjpadgett:
Thanks Brady, I don't have time to do a bunch of refactoring to get rid of the Global schema used throughout zend project. It is not a recommended method and will eventually need to be address. Thanks for get a phpstan fix.

PR #8700, opened 2025-07-28
Restore Flow Board Filter

### LINKED ISSUE #8699: Error in the statistics provided by the flow chart
### Describe the bug
From forum issue: https://community.open-emr.org/t/error-in-the-statistics-provided-by-the-flow-chart/25913?u=sjpadgett

The problem is that in the statistics, regardless of the filter used (professionals, centers, etc.), the statistics don’t change. They’re always the same and refer to all the appointments that day, for all professionals and all centers, which seems incorrect.

### Client configuration

- Browser:
- OpenEMR version: 7.0.0+
- Operating system:

sjpadgett:
add back parameters to search

Fixes #8699


bradymiller:
2DV

PR #8709, opened 2025-07-30
chore: add php dependencies to composer json

### LINKED ISSUE #8708: chore: add php dependencies to composer json
chore: add php dependencies to composer json

bradymiller:
fixes #8708

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,6 @@
     },
     "require": {
         "php": ">=8.2.0",
-        "ext-apache2": "*",
         "ext-calendar": "*",
         "ext-ctype": "*",
         "ext-curl": "*",
@@ -112,7 +111,6 @@
     "config": {
         "platform": {
             "php": "8.2",
-            "ext-apache2": "8.2",
             "ext-calendar": "8.2",
             "ext-ctype": "8.2",
             "ext-curl": "8.2",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -12,6 +12,34 @@
     },
     "require": {
         "php": ">=8.2.0",
+        "ext-apache2": "*",
```
bradymiller: going to remove apache2 since this would not be required in nginx/fpm installs

composer.json
```diff
@@ -84,7 +112,34 @@
     "config": {
         "platform": {
             "php": "8.2",
-            "ext-intl": "8.2"
+            "ext-apache2": "8.2",
```
bradymiller: going to remove apache2 since this would not be required in nginx/fpm installs

bradymiller:
Realized apache2 should not be on list, so will remove.

As an aside, the flex/production dockers also bring in (specific to the dockers):
(soon to be ext-apache2)
ext-apache2
ext-fpm
ext-opcache
ext-pecl-apcu
ext-redis

Goal is to make it clear for folks what packages are needed when bringing in dependencies for operating systems etc.

Please let me know if any changes needed.

bradymiller:
testing well and bringing in

PR #8711, opened 2025-07-30
ci(phpstan): remove rector to fix phpstan

### LINKED ISSUE #8710: Produce phpstan baseline on failure
PHPStan currently behaves differently in actions than locally, so it can be a chore to generate the error ignores necessary to make progress. I propose that we always produce a baseline file when phpstan fails, so we can get those errors if they can't reasonably be fixed in a given PR.

### LINKED ISSUE #8714: Rector interferes with phpstan
@adunsulag discovered that in GitHub Actions (and seemingly only in GitHub Actions) the [rector TestCase](https://github.com/rectorphp/rector/blob/66f60d7aca890180f49488209a8fb765130db8a1/stubs-rector/PHPUnit/Framework/TestCase.php) overrides the phpunit one, which is what's disrupting so many phpstan checks.

### LINKED ISSUE #8715: Symfony PantherTestCase dynamic inheritance confuses phpstan
The dynamic inheritance [here](https://github.com/symfony/panther/blob/b7e0f834c9046918972edb3dde2ecc4a20f6155e/src/PantherTestCase.php#L19) makes phpstan reject all inherited testcase methods.

kojiromike:
Fixes #8710
Fixes #8714
Fixes #8715

- rector includes a testcase override that makes phpstan reject methods that should be inherited from phpunit's testcase class.
- even after removing rector, PantherTestCase dynamic inheritance causes phpstan to still reject phpunit testcase methods.
- even after fixing the two above, github and local are still different, so it's useful to publish a phpstan baseline file when phpstan fails in actions.

- [x] uninstall rector in the workflow for phpstan
- [x] update the workflow to publish baseline.neon when and only when phpstan analyze fails in GitHub.
- [x] create a class alias to work around [PantherTestCase's dynamic inheritance](https://github.com/symfony/panther/blob/b7e0f834c9046918972edb3dde2ecc4a20f6155e/src/PantherTestCase.php#L19)


adunsulag:
@kojiromike I didn't realize you were going to work on this.  I should have opened an issue as I had done some work on removing rector altogether, removing it during the action is likely a better approach then the one I was taking.

kojiromike:
@adunsulag I just threw the removing rector thing into a PR that I already had, which was just to publish a baseline when phpstan failed, since I kept needing one that only applied in github.

adunsulag:
@kojiromike if you want any assistance on this let me know.  I do have a working implementation where rector is removed that fully passes phpstan using just a single phpstan baseline file, but I think what you are doing with the panther alias and other pieces is probably the better approach.  If my stuff can be helpful let me know.

kojiromike:
@adunsulag I think this PR is ready to merge as is if you like. Then we can continue to iterate.

bradymiller:
Bringing in PR to tame phpstan as quick as possible :)

Just let me know if anybody needs a PR involving a PHPStan Stuffy and boxing gloves or for those more violently inclined explosives :) Maybe even therapy for those devs that have been severely impacted by phpstan :)

phpstanstuffy

kangaroo-boxing

tnt-funny

kangaroo-psych

PR #8724, opened 2025-08-01
fix(CryptoGen): prohibit passing null to encrypt

### LINKED ISSUE #8723: encrypting `null` deprecated in php 8
`openssl_encrypt(null)` treats that `null` as the empty string, but in php 8+ this is deprecated, and it may raise an exception some day.

(This intersects with other work I'm doing on CryptoGen, but I don't like to make arbitrary changes in a vacuum, so documenting this on its own.)

kojiromike:
Fixes #8723

Fixes php 8 deprecation warnings about passing null to encrypt functions.

Remove the nullable types for things going to encrypt.

### DIFF
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * CryptoGen class.
+ * CryptoGen class
  *
  *   OpenEMR encryption/decryption strategy:
  *    1. Two separate private key sets are used, one key set in the database and one key set on the drive.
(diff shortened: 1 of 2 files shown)


bradymiller:
hi @kojiromike ,
Wouldn't this break when send null though via typing check (ie. turn the deprecation into an error). Just not sure if any of the calls on this function send a null. Am wondering what actually gets encrypted when send a null. Am guessing a decrypt on it would return a blank/empty rather than null, but would need to test that; if that is case, would it make more sense for the functions to allow null, but then convert them to blank/empty?

bradymiller:
fyi @kojiromike , posed the question to Claude regarding what gets encrypted when null and Claude explained that blank is stored and will be given back on decrypt (ie. null gets converted to blank/empty). Of course, Claude could be hallucinating :)

kojiromike:
In a bunch of cases in php 7, passing null will be treated as the empty string. In php 8, it's deprecated, meaning it'll still work, but you'll get a deprecation warning. In any case, after making this change phpstan didn't complain, so I don't think there are any internal places where null is passed to encrypt.

bradymiller:
hi @kojiromike , Looked through the function calls and agree didn't find any that would bring in null. Two things:
1. The comments for the functions also need to be addressed (ie. remove the `|null` for the 1st parameter in both functions).
2. Won't the test for CryptoGenTest for testEncryptStandardWithNullValue now break?

bradymiller:
hi @kojiromike , I think incorporating the decrypt may have higher chance of bugs that may require some focal protections on the calls. Had Claude do some analysis on this:

# Summary: Breaking Changes in PR #8724

## **The Core Change**
```php
// BEFORE - Accepts null:
public function encryptStandard(?string $value, ...)
public function decryptStandard(?string $value, ...)

// AFTER - Rejects null:  
public function encryptStandard(string $value, ...)
public function decryptStandard(string $value, ...)
```

## **What Actually Breaks (New Issues Only)**

### ❌ **Issues that ALREADY existed (not new):**
- `file_get_contents()` returning `false` 
- Other functions returning `false`/`int`/`array` instead of string
- These already caused TypeErrors with `?string`

### ✅ **Issues INTRODUCED by PR #8724:**
Only **explicit `null` values** passed as first parameter

---

## **`decryptStandard()` Breaking Changes - HIGH IMPACT**

### **Critical Breaking Cases:**
```php
// 1. Missing POST/GET parameters
$data = $cryptoGen->decryptStandard($_POST['enc_data']); 
// Breaks if $_POST['enc_data'] is null/missing

// 2. Portal password reset  
$token = $crypto->decryptStandard($_GET['forward'], null, 'drive', 6);
// Breaks if $_GET['forward'] is null/missing

// 3. Global configuration values
$cred = $cryptoGen->decryptStandard($GLOBALS['erx_account_password']);
// Breaks if global is null/unset

// 4. Database query results
$secret = $cryptoGen->decryptStandard($registrationSecret);
// Breaks if $registrationSecret is null from failed DB query

// 5. CouchDB document content
$content = $this->cryptoGen->decryptStandard($respM->data, null, 'database');
// Breaks if $respM->data is null
```

### **Risk Assessment:**
- **HIGH RISK**: User-facing functionality (portal, payments, prescriptions)
- **MEDIUM RISK**: File operations, configuration loading
- **WIDESPREAD**: Many `$_POST`, `$_GET`, `$GLOBALS` usages

---

## **`encryptStandard()` Breaking Changes - LOWER IMPACT**

### **Potential Breaking Cases:**
```php
// 1. Configuration arrays with null values
foreach ($GLOBALS['more_secure'] as $key => $value) {
    $GLOBALS['more_secure'][$key] = $cryptoGen->encryptStandard($value);
    // Breaks if $value is null
}

// 2. Variables set to null from failed operations  
$secret = null; // from failed decryption
$encrypted = $cryptoGen->encryptStandard($secret);
// Breaks if explicitly null

// 3. Database fields that are NULL
$encrypted = $cryptoGen->encryptStandard($result['nullable_field']);
// Breaks if database field is NULL
```

### **Risk Assessment:**
- **LOWER RISK**: Fewer direct user inputs
- **MORE DEFENSIVE**: Most code validates before encrypting
- **FEWER INSTANCES**: Less common null-passing patterns

---

## **Most Critical Fixes Needed**

### **Top Priority (Production Breaking):**
1. **`paylib.php`**: `decryptStandard($_POST['enc_data'])`
2. **Portal password reset**: `decryptStandard($_GET['forward'])`  
3. **eRx credentials**: `decryptStandard($GLOBALS['erx_account_password'])`
4. **MFA authentication**: `decryptStandard($registrationSecret)`

### **Recommended Fix Pattern:**
```php
// OLD (breaks):
$result = $cryptoGen->decryptStandard($_POST['data']);

// NEW (safe):
$data = $_POST['data'] ?? '';
if (!empty($data)) {
    $result = $cryptoGen->decryptStandard($data);
} else {
    $result = '';
    // Handle missing data appropriately
}
```

---

## **Bottom Line**

- **`decryptStandard()`**: **High impact** - many user-facing breaking changes
- **`encryptStandard()`**: **Low impact** - fewer breaking changes  
- **Root cause**: Removing `?string` nullable type for first parameter only
- **Fix strategy**: Add null checks before calling encryption functions
- **Priority**: Fix user-facing decrypt operations first (payments, portal, auth)

The change is primarily about **preventing PHP 8 deprecation warnings** when null reaches OpenSSL functions, but creates **backward compatibility breaks** that need systematic fixing.

kojiromike:
@bradymiller ok, to lower risks, I'll break this PR up into much smaller chunks so we can digest it a byte at a time, and if we have to roll something back, it won't mean throwing out the baby with the bathwater. See #8739

bradymiller:
hi @kojiromike , agree with plan.

Issue #8732, opened 2025-08-05
docker/development-easy broken due to missing ARM64 support for the selenium/standalone-chrome image


msummers42:
### Describe the bug

In the docker/development-easy, the selenium/standalone-chrome image does not ship an ARM64 image thereby breaking the docker compose on ARM64 like MacOS.

### To Reproduce
Try to use the existing docker-compose.yml on an AMR64 system, like MacOS.

PR #8733, opened 2025-08-05
fix: Use selenium/standalone-chromium to enable support for both AMD64 & ARM64

msummers42:
This change pins to v4 of the upstream Selenium Grid with the latest compatible version of Chromedriver. It is using Chromium as the base versus Chrome. The primary issue with the selenium/standalone-chrome image is that it does not ship an ARM64 image thereby breaking the docker compose on ARM64 like MacOS.

Fixes # https://github.com/openemr/openemr/issues/8732

Supports AMD64 & ARM64 with the Selenium Chromedriver.

Use the upstream chromium flavor of the image that does support ARM64

### DIFF
--- a/docker/development-easy/docker-compose.yml
+++ b/docker/development-easy/docker-compose.yml
@@ -101,7 +101,7 @@ services:
       retries: 3
   selenium:
     restart: always
-    image: selenium/standalone-chrome:134.0-chromedriver-134.0-grid-4.31.0-20250414    # Pinned to Chrome 134 and grid 4.31
+    image: selenium/standalone-chromium:4 # Pins to Selenium Grid major version 4 using the latest compatible version of chrome driver, amd64 & arm64 support
     ports:
     - 4444:4444    # Selenium WebDriver interface
     - 7900:7900    # VNC port for viewing browser during tests (to work, will require SELENIUM_FORCE_HEADLESS to be "false" in openemr service)


docker/development-easy/docker-compose.yml
```diff
@@ -101,7 +101,7 @@ services:
       retries: 3
   selenium:
     restart: always
-    image: selenium/standalone-chrome:134.0-chromedriver-134.0-grid-4.31.0-20250414    # Pinned to Chrome 134 and grid 4.31
+    image: selenium/standalone-chromium:4 # Pins to Selenium Grid major version 4 using the latest compatible version of chrome driver, amd64 & arm64 support
```
bradymiller, line 104: try this instead:
```
selenium/standalone-chromium:135.0-chromedriver-135.0-grid-4.31.0-20250404 # Pinned to Chromium 135 and grid 4.31
```

If that works for you, then make the change in the PR and then I will bring it in. We need to pin versions here so the ci testing remains stable. Already tested that ci testing works well with this version at https://github.com/openemr/openemr/pull/8736

msummers42:
@bradymiller I am generally reluctant to ship outdated versions of containers, which your suggestion does. Chromium has already moved on to 138.x (139.x is coming soon) and Selenium Grid is up to 4.34.0.

My change does pin the image to a version, just not quite so specifically as your suggestion. The tag `4`, uses the exact same image as the tags `4.34`, `4.34.0`, `latest`, `138.0.7204.183-chromedriver-138.0.7204.183-grid-4.34.0-20250727` and several others, including `138.0-chromedriver-138.0-grid-4.34.0-20250727` which aligns with the tag naming pattern used previously as well as in your suggestion.

Interestingly, there is a CVE rated High on versions prior to to 138.0.7204.96/.97 for Windows, 138.0.7204.92/.93 for Mac and 138.0.7204.92 for Linux.

Relying on an older version creates a false sense of stability where tests pass even if the application is broken for actual users (who are likely auto-updating to the latest version of Chrome/Chromium). This outdated containers also contains known security vulnerabilities, posing an unnecessary, albeit fairly minor, risk. By regularly updating, we prevent the accumulation of technical debt and ensure the tests accurately reflect reality. The easiest means of regularly updating is to pin to the simple `4` tag in something like GH Actions. For the case of a local dev environment, the user would be responsible for pulling in any new versions of a tagged container. If this is an unacceptable outcome, with some minor effort we could enable dependabot to update the images in the various docker-compose files. Alternatively, we could write some checks into the openemr-cli to compare versions available in DockerHub versus what is available locally.

What do you think?

bradymiller:
hi  @*msummers42*
*Happy to try your strategy. I ended up being forced to migrate to selenium
grid (good thing in the end) since most recent chromedriver version was
basically broken for selenium, but github actions will pick that up if
still the case :) . Recommend putting your proposed tag also for github
actions stuff and insane dev environment (see my linked PR to see the 2
places where to do this) and will bring it in if passes github actions.*
*thanks! -brady*

msummers42:
I think they're all just different tags on the same image. For example, the following tags all point to the same container's sha256
```
138.0
138.0-20250727
138.0-chromedriver-138.0
138.0-chromedriver-138.0-20250727
138.0-chromedriver-138.0-grid-4.34.0-20250727
138.0.7204.183
138.0.7204.183-20250727
138.0.7204.183-chromedriver-138.0.7204.183
138.0.7204.183-chromedriver-138.0.7204.183-20250727
138.0.7204.183-chromedriver-138.0.7204.183-grid-4.34.0-20250727
4
4.34
4.34.0
4.34.0-20250727
latest
```

bradymiller:
hi @msummers42, would just do `latest`, then wouldn't need to worry about
it unless it breaks in future (which will be very obvious on github
actions), or `4` would be ok. (if strategy is to use most recent production
quality release without ever needing to worry about changing with newer
versions in future, then would go with `latest`)

kojiromike:
@bradymiller hmm, if we use `latest` and it breaks in the future, we won't have as much information about specifically which version broke. Could we use dependabot instead? that way, when it breaks, `master` will remain working, while the specific dependabot PR will show us the breakage?

bradymiller:
hi @kojiromike
Sure, but note there are 3 places that need updating (ci, easy dev
docker-compose, insane dev docker compose).

Note looks like can output the chromedriver and grid version within docker
in ci via (haven't tested, just got from google ai):
chromedriver -v
java -jar /opt/selenium/selenium-server-standalone.jar --version

May want to test either master or 4 now since it it fails miserably then
this is all moot :)

bradymiller:
correction:
meant test `latest` or `4` (not master)

bradymiller:
I think a good potential solution is having dependbot focus on the grid
version *.*.* (ignore all the other tags) (and even allow major version
updates) since this will avoid grid dev versions. And then report the
chromedrive versions during the ci run. But just thinking out loud.

bradymiller:
hi @kojiromike and @msummers42 , Got a PR working that uses grid version with dependabot incorporation (note there are a couple unknowns that likely will not know answers to until bring it in):
https://github.com/openemr/openemr/pull/8736

msummers42:
Closing in favor of https://github.com/openemr/openemr/pull/8736

PR #8747, opened 2025-08-07
refactor(CryptoGen): Validate KeySource via Enum

### LINKED ISSUE #8746: CryptoGen does not validate keySource
CryptoGen can take either 'database' or 'drive' for the value of keysource, but if it is not 'database' it is assumed to be 'drive'. Since the keySource is a string provided in the public API of the class, it should be checked.

kojiromike:
Fixes #8746

Validates any passed keySource against an enum of good possible values.

- [x] create an enum for KeySource
- [x] convert any incoming value into an enum instance, or die trying.

### DIFF
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -33,8 +33,6 @@
 namespace OpenEMR\Common\Crypto;
 
 use Exception;
-use OpenEMR\Common\Crypto\CryptoGenException;
-use OpenEMR\Common\Crypto\CryptoInterface;
 use OpenEMR\Common\Utils\RandomGenUtils;
 
 class CryptoGen implements CryptoInterface
@@ -74,7 +72,7 @@ class CryptoGen implements CryptoInterface
      */
     public function encryptStandard(?string $value, ?string $customPassword = null, string $keySource = 'drive'): string
     {
-        return $this->encryptionVersion . $this->coreEncrypt($value, $customPassword, $keySource, $this->keyVersion);
+        return $this->encryptionVersion . $this->coreEncrypt($value, $customPassword, KeySource::from($keySource), $this->keyVersion);
     }
 
     /**
@@ -104,13 +102,15 @@ class CryptoGen implements CryptoInterface
             }
         }
 
+        $keySourceEnum = KeySource::from($keySource);
+
         // Map the encrypt/decrypt version to the correct decryption function
         if ($encryptionVersion == 6) {
-            return $this->coreDecrypt($trimmedValue, $customPassword, $keySource, "six");
+            return $this->coreDecrypt($trimmedValue, $customPassword, $keySourceEnum, "six");
         } elseif ($encryptionVersion == 5) {
-            return $this->coreDecrypt($trimmedValue, $customPassword, $keySource, "five");
+            return $this->coreDecrypt($trimmedValue, $customPassword, $keySourceEnum, "five");
         } elseif ($encryptionVersion == 4) {
-            return $this->coreDecrypt($trimmedValue, $customPassword, $keySource, "four");
+            return $this->coreDecrypt($trimmedValue, $customPassword, $keySourceEnum, "four");
         } elseif (($encryptionVersion == 2) || ($encryptionVersion == 3)) {
             return $this->aes256DecryptTwo($trimmedValue, $customPassword);
         } elseif ($encryptionVersion == 1) {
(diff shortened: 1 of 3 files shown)


src/Common/Crypto/CryptoGen.php
```diff
@@ -69,11 +70,15 @@ class CryptoGen
...
      */
-    public function encryptStandard(?string $value, ?string $customPassword = null, string $keySource = 'drive')
+    public function encryptStandard(?string $value, ?string $customPassword = null, string $keySource = 'drive'): string
     {
-        $encryptedValue = $this->encryptionVersion . $this->coreEncrypt($value, $customPassword, $keySource, $this->keyVersion);
+        // Validate and convert keySource to enum
+        $keySourceEnum = KeySource::tryFrom($keySource);
+        if ($keySourceEnum === null) {
+            throw new CryptoGenException("Invalid key source: $keySource");
+        }
```
bradymiller: neat stuff. Still see in draft mode, but just doing a quick lookover until ready. Wouldn't using `from` instead of `tryFrom` get you what you want without needing to check it for null (looks like will throw a fatal error that will be logged with the invalid key source).

src/Common/Crypto/CryptoGen.php
```diff
@@ -449,84 +454,139 @@ public function aes256Decrypt_mycrypt(string $sValue): string
...
+                : $this->decryptStandard($storedFileContents, null, KeySource::DATABASE->value);
+            if ($key === $storedKey) {
+                return $key;
             }
         }
-
-        // Store key in cache and then return the key
-        $this->keyCache[$cacheLabel] = $key;
-        return $key;
+        throw new CryptoGenException("OpenEMR Error: The newly created key could not be stored, encoded or encrypted correctly.");
```
kojiromike, line 586: I'm slightly annoyed at myself that I couldn't figure out a nice way to do these round trips by recycling the `collect` methods – it's doable, of course, but the recursion and exception handling make it seem worse than the duplicated code.

adunsulag (approved):
I don't see any problems looking through this.  Looks sensible to me.

src/Common/Crypto/CryptoGen.php
```diff
@@ -449,84 +454,139 @@ public function aes256Decrypt_mycrypt(string $sValue): string
...
+        throw new CryptoGenException("OpenEMR Error: Key in drive is not compatible (ie. can not be decrypted) with key in database - Exiting.");
+    }
 
-        // Ensure have a key (if do not have key, then is critical error, and will exit)
+    /**
+     * Create a new crypto key in the drive.
+     * Produce a 256bit key (32 bytes equals 256 bits)
+     *
+     * @param  string $label   the version and sub-version label
+     * @param  string $version the key encryption version (not used in this implementation, but important for future compatibility)
```
bradymiller: $version is being used in this function

src/Common/Crypto/CryptoGen.php
bradymiller: round trip clear for me and clear that it really makes sure the new keys are where they should be since things will implode if that is not the case :)

src/Common/Crypto/CryptoGen.php
```diff
@@ -449,84 +454,139 @@ public function aes256Decrypt_mycrypt(string $sValue): string
...
+     * @param  string $version the key encryption version
+     * @return string the plaintext key value
+     * @throws CryptoGenException if key collection fails
+     */
+    protected function collectDriveKey(string $label, string $version): string
+    {
+        $keyPath = $GLOBALS['OE_SITE_DIR'] . "/documents/logs_and_misc/methods/" . $label;
+        $isLegacyVersion = in_array($version, ["one", "two", "three", "four"]);
+        if (!$this->fileExists($keyPath)) {
+            $key = $this->createDriveKey($label, $version);
```
bradymiller: looks like the $key variable here doesn't do anything (either skipped when key already exists or is overwritten below)
(agree function call definitely needed, just don't see the return $key being used)

bradymiller:
hi @kojiromike , Overall looks great! Just noted couple minor things on review.

src/Common/Crypto/CryptoGen.php
kojiromike: Right, thanks

kojiromike:
@bradymiller thanks for another thorough review. I think I've addressed everything.

bradymiller (approved):

bradymiller:
hi @kojiromike , look great and bringing in. thanks!

PR #8748, opened 2025-08-07
refactor(CryptoGen): optimize method signatures and control flow

### LINKED ISSUE #8750: CryptoGen private methods are too flexible
The private methods in CryptoGen have optional parameters with default values, sometimes return `false` on failure, and trigger phpstan and other checks as a result.

These kinds of things are easy and safe to fix for private methods without changing the interface.

kojiromike:
Fixes #8750 

The refactoring maintains full backward compatibility while reducing the
codebase and improving overall performance through
more efficient memory usage and simplified execution paths.

- Made private method signatures simpler and stricter.
- Unified parameter handling by using null coalescing operator (??) for cleaner code
- Replaced verbose if/else blocks with ternary operators and early returns
- Simplified `cryptCheckStandard()` from 8 lines to a single `preg_match` expression
- Eliminated unnecessary variable assignments and intermediate steps
- Merged similar error handling patterns for HMAC authentication failures
- Simplified key collection logic by removing redundant conditionals
- Streamlined encryption value completion logic
- Reduced method call overhead through parameter consolidation
- Eliminated temporary variable creation in multiple locations
- Optimized string concatenation patterns
- Improved parameter naming and ordering for logical consistency
- Reduced cyclomatic complexity through simplified branching
- Made error handling patterns more consistent across methods

### DIFF
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -22,9 +22,11 @@
  * @author    Ensoftek, Inc
  * @author    Brady Miller <brady.g.miller@gmail.com>
  * @author    Jerry Padgett <sjpadgett@gmail.com>
+ * @author    Michael A. Smith <michael@opencoreemr.com>
  * @copyright Copyright (c) 2015 Ensoftek, Inc
  * @copyright Copyright (c) 2018-2019 Brady Miller <brady.g.miller@gmail.com>
  * @copyright Copyright (c) 2024 Jerry Padgett <sjpadgett@gmail.com>
+ * @copyright Copyright (c) 2025 OpenCoreEMR Inc.
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -65,24 +65,24 @@ class CryptoGen implements CryptoInterface
     /**
      * Encrypts data using the standard encryption method
      *
-     * @param ?string $value          The data to encrypt
-     * @param ?string $customPassword If provided, keys will be derived from this password (standard keys will not be used)
-     * @param string  $keySource      The source of the standard keys. Options are 'drive' and 'database'
+     * @param ?string $value           The data to encrypt
+     * @param ?string $customPassword  If provided, keys will be derived from this password (standard keys will not be used)
+     * @param string  $keySource       The source of the standard keys. Options are 'drive' and 'database'
      * @return string The encrypted data
      */
     public function encryptStandard(?string $value, ?string $customPassword = null, string $keySource = 'drive'): string
     {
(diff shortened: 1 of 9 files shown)


src/Common/Crypto/CryptoGen.php
```diff
@@ -193,16 +189,14 @@ private function coreEncrypt(?string $sValue, ?string $customPassword = null, st
 
         $hmacHash = hash_hmac('sha384', $iv . $processedValue, $sSecretKeyHmac, true);
 
-        if ($sValue != "" && ($processedValue == "" || $hmacHash == "")) {
+        if ($sValue != "" && ($processedValue === "" || $hmacHash === "")) {
             throw new CryptoGenException("OpenEMR Error : Encryption is not working (encrypted value is blank or hmac hash is blank).");
         }
```
bradymiller, line 195: bookmark for ongoing review. should complete this tomorrow and then bring in.

kojiromike: Thanks, @bradymiller. I think #8754 is more urgent, as I have seen it mask failures in unit tests. That is, I think we should merge that then rerun tests on anything that touches the database, before merging.

kojiromike: That's merged now, so this is ready for review.

bradymiller:
hi @kojiromike , beautiful stuff. looks great. looks like couple failed unit tests to clean up.

src/Common/Crypto/CryptoGen.php
```diff
@@ -184,25 +182,23 @@ private function coreEncrypt(?string $sValue, ?string $customPassword = null, st
...
 
-        if (empty($customPassword)) {
-            // prepend the encrypted value with the $hmacHash and $iv
-            $completedValue = $hmacHash . $iv . $processedValue;
-        } else {
-            // customPassword mode, so prepend the encrypted value with the salts, $hmacHash and $iv
-            $completedValue = $sSalt . $hmacHash . $iv . $processedValue;
+        $completedValue = $hmacHash . $iv . $processedValue;
+        if (isset($salt)) {
+            // customPassword mode, so prepend the encrypted value with the salt
```
bradymiller:
```
// customPassword mode, so prepend the completed value with the salt
```

bradymiller:
Looks like couple more unit tests. so close :)
Will bring in when pass tests.

bradymiller:
hi @kojiromike ,
Actually, to be safe and not need to even worry about backward compatibility (ie. prior encrypted stuff not getting decrypted (in theory should not be happening but never know and we would be in a true pickle if it did happen :) ), would simply bump up the $encryptionVersion and $keyVersion to `007` and `seven` with concordant changes at line 112 (insert line), and line 139.

bradymiller:
To clarify why above helpful. If we do find a backward compatibility issue in future, then can create a specific decrypt function for 6 (ie. above will keep that option open).

PR #8755, opened 2025-08-07
Feat(Option to automatically open a pt's issues tab when pt demographics are open.  Issues tab auto closes when new pt is selected)

growlingflea:
…raphics is opened.  When a new pt is opened, all tabs activate to the new pt

Fixes #

Feature that allows the end user to automatically open a pt's issue tab when demographics is opened.  When a new pt is selected the previous tabs last pt closed and new tabs open for the new pt.

### DIFF
--- a/interface/main/tabs/js/frame_proxies.js
+++ b/interface/main/tabs/js/frame_proxies.js
@@ -47,6 +47,40 @@ left_nav.setPatient = function(pname, pid, pubpid, frname, str_dob)
         tabCloseByName('rev');
     });
 
+    if (AutoIssuesTab) {
+        loadPatientIssues("leftNav");
+
+
+//**ERX
+        left_nav.setPatientErx = function () {
+            navigateTab(webroot_url + "/interface/eRx.php", "er1");
+
+
+        };
+
+//Close the open tabs to previous pt.
+        left_nav.closeErx = function () {
+            tabCloseByName('er1');
+            tabCloseByName('er2');
+            tabCloseByName('er3');
+            tabCloseByName('enc2');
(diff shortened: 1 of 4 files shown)


growlingflea:
@stephenwaite  I went ahead and branched this off master.  Not sure why I'm getting PHP errors but I don't think its related to my new code.

interface/main/tabs/js/frame_proxies.js
```diff
@@ -47,6 +47,40 @@ left_nav.setPatient = function(pname, pid, pubpid, frname, str_dob)
         tabCloseByName('rev');
     });
 
+    if (AutoIssuesTab) {
```
CloCkWeRX, line 50: Bit of excess line breaks here - that would be good to clean up/make consistent.

interface/main/tabs/js/tabs_view_model.js
```diff
@@ -428,6 +431,15 @@ function clearPatient(openFinder = true)
 	});
 }
 
+
+function loadPatientIssues(fun = '')
```
CloCkWeRX, line 435: Is there a better variable name you could use?
fun, while it sounds both `fun` and `functional` can be a bit ambigious

CloCkWeRX (changes requested):
Could you have a quick go at the formatting, plus tweak the pull request title?

You might want to try `feat: Option to automatically open patient's tab...` rather than the braces - https://www.conventionalcommits.org/en/v1.0.0/ is the full explainer, but the short version is the automated tooling doesn't quite get what you mean the same way a human does without following the pattern strictly.

PR #8775, opened 2025-08-13
fix: more fixes to selenium dependabot

### LINKED ISSUE #8765: fix: dependabot fix for selenium
dependabot fixes for selenium (ongoing)

1. dependabot can not see the compose-shared-selenium.yml script (will try to first put a docker- infront of it. if that does not work, then will lable it docker-compose.yml and put it in a directory)
2. Will dependabot grouping work across directories (need to wait, but am doubting this. not end of world. just mean there will be 3 separate PRs when updated) (FIXED)
3. Will dependabot stick to the versioning tag or will it get confused by other tags (FIXED)

bradymiller:
fixes #8765

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -118,9 +118,9 @@ updates:
   # Docker selenium docker updates in 3 places in the repo as a group
   - package-ecosystem: "docker-compose"
     directories:
-      - "ci/compose-shared-selenium.yml"
-      - "docker/development-easy/docker-compose.yml"
-      - "docker/development-insane/docker-compose.yml"
+      - "docker/development-easy/"
+      - "docker/development-insane/"
+      - "ci/"
     schedule:
       interval: "weekly"
     allow:


.github/dependabot.yml
```diff
@@ -118,9 +118,9 @@ updates:
   # Docker selenium docker updates in 3 places in the repo as a group
   - package-ecosystem: "docker-compose"
     directories:
-      - "ci/compose-shared-selenium.yml"
-      - "docker/development-easy/docker-compose.yml"
-      - "docker/development-insane/docker-compose.yml"
+      - "docker/development-easy/"
+      - "docker/development-insane/"
+      - "ci/"
```
kojiromike, line 123:
```suggestion
      - "ci/"
      - "docker/development-easy/"
      - "docker/development-insane/"
```

Isn't dependabot going to go nuts trying to update all the different versions of things in -insane?

bradymiller: this is still a work in progress (the custom filename in ci is causing issues :) plan a debug commit soon to figure things out and then have a solution). Note using this to only focus on the selenium docker:
```
    allow:
      - dependency-name: "selenium/standalone-chromium"
```

PR #8777, opened 2025-08-13
fix: more selenium dependabot fixes

### LINKED ISSUE #8765: fix: dependabot fix for selenium

### LINKED ISSUE #8778: ci all of sudden missing 4 tests
Noted that ci is all of sudden missing 4 tests
(just started today)

Maybe a temporary github thing hopefully. Will see if corrects itself :)

looks like its the nginx ci tests. pretty sure i must of done something :)
(guessing i messed up the mechanism that finds the test directories by adding another unrelated directory :) )

bradymiller:
fixes #8765
fixes #8778


bradymiller:
That was fun :)
Turns out:
1. Dependabot only can see docker-compose.yml files, so required a bit of refactoring.
2. Turns our the ordering of docker-compose file merging is important. The project name is derived from the first item, so needed to re-order so the selenium one was not first (or else project name changed to compose-shared-selenium rather than ci and broke stuff)

Still need to ensure dependabot looks ok for this (via insights) after bring in, but pretty confident it will now work.

kojiromike:
@bradymiller there's a fairly recent addition to docker compose where you can get it to output the final config -- maybe we should be doing that in the test.yml if we aren't already.

https://docs.docker.com/reference/cli/docker/compose/config/

kojiromike:
Oh lol, nevermind, we are doing it already.

https://github.com/bradymiller/openemr/blob/82a0f21c626b412b8c8d1f82b5128f2d97cd452c/.github/workflows/test.yml#L216

carry on!

PR #8806, opened 2025-08-22
fix(import_ccda): don't use result of void function

### LINKED ISSUE #8805: Result of void function is used
Trivial phpstan fix:

The result of a void function is being used in `contrib/util/ccda_import/import_ccda.php`.

kojiromike:
Fixes #8805

Removes `echo` before function call on a void function.

- [x] remove use of void function output.
- [x] remove some phpstan baseline misses.


kojiromike:
This passed locally (bare-metal), locally (docker dev-easy) and in actions. Merging and bringing in.

PR #8819, opened 2025-08-23
fix(VoiceClient): make $authUser protected, not private

### LINKED ISSUE #8818: FaxSMS VoiceClient accesses private authUser property from parent
`OpenEMR\Modules\FaxSMS\Controller\AppDispatch\AppDispatch::$authUser` is declared private, but `OpenEMR\Modules\FaxSMS\Controller\VoiceClient\VoiceClient`, which extends the above, tries to access it in `getVoiceCredentials`. 

That said, `$authUser` is not really being used as an instance property. It always seems to be redefined in every method before being accessed. The way it's defined is not entirely consistent. I can make this illegal access problem "go away" by making authUser protected instead of private, but it seems more correct to define it once in the AppDispatch constructor and remove the code that defines it in various places.

The question is, would it be correct to define it as it's currently defined in the constructor, unconditionally, as it's defined in `getSetup`, with four different possible values, or as it's defined in `getEmailSetup`, with two possible values?

kojiromike:
Fixes #8818

This is the simplest possible fix for #8818. It just makes the accessed property legal. That said, I think it's not ideal that this property is accessed multiple times at all. But either way, it's likely this property needs to be not-private. 

- [x] make `$authUser` accessible by subclasses.

### DIFF
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
@@ -34,7 +34,7 @@ abstract class AppDispatch
     protected $_currentAction;
     protected $credentials;
     private $_request, $_response, $_query, $_post, $_server, $_cookies, $_session;
-    private $authUser;
+    protected $authUser;
 
     /**
      * @throws \Exception
--- a/phpstan.github.neon
+++ b/phpstan.github.neon
@@ -11724,10 +11724,6 @@ parameters:
     identifier: staticProperty.notFound
     count: 1
     path: interface/modules/custom_modules/oe-module-faxsms/src/Controller/VoiceClient.php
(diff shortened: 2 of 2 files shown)


interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
```diff
@@ -34,7 +34,7 @@ abstract class AppDispatch
     protected $_currentAction;
     protected $credentials;
     private $_request, $_response, $_query, $_post, $_server, $_cookies, $_session;
-    private $authUser;
+    protected $authUser;
```
sjpadgett, line 37: This propery is a work in progress to remove as I was cleaning up when adding VoiceService and deciding if I wanted to let service have separate credentials and in the end decide to keep with standard credentials.
After I reviewed you can remove as I think I've got it covered. If not leave protected and I'll remove next time in that module. thanks

sjpadgett (approved):
I'm fine with what you decide.

PR #8828, opened 2025-08-24
fix: `unset` on nonexistent variable

### LINKED ISSUE #8827: `unset` on undefined variables
There are five places in the codebase where `unset(...)` is called on names that have never been assigned in the current scope.

kojiromike:
Fixes #8827

Fix `unset` on undefined variables.

Remove `unset` calls where they are not doing anything.

### DIFF
--- a/src/Common/Auth/OneTimeAuth.php
+++ b/src/Common/Auth/OneTimeAuth.php
@@ -330,7 +330,6 @@ class OneTimeAuth
             }
         } catch (OneTimeAuthExpiredException $e) {
             $this->systemLogger->error("Failed " . $e->getMessage());
-            unset($auth);
             throw new OneTimeAuthException(xlt("Decode Authentication Failed! Contact administrator."));
         }
         if (!empty($auth['error'] ?? null)) {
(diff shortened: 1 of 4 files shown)


src/Common/Auth/OneTimeAuth.php
```diff
@@ -330,7 +330,6 @@ public function processOnetime($token, $redirect_token): array
             }
         } catch (OneTimeAuthExpiredException $e) {
             $this->systemLogger->error("Failed " . $e->getMessage());
-            unset($auth);
```
bradymiller, line 333: auth is set in line 323

bradymiller: oh, i think i get it. auth will not be set here since that means the function it called threw an error/exception. correct?

bradymiller (approved):

src/Common/Auth/OneTimeAuth.php
kojiromike: phpstan is being confusing here, because technically it could be set, but it might not be set, too. 

But the real question is why bother unsetting a variable (and deal with it potentially not being set) just before it goes out of scope?

kojiromike:
Discovering #8829 has made me a bit nervous about some of the phpstan findings. Marking this as draft until that's resolved.

PR #8833, opened 2025-08-25
fix: remove only-null annotations

### LINKED ISSUE #8829: `null` is never correct as an _only_ type
There are quite a few places in the code where a docblock has something like:

```
/**
 * …
 * @param null $name
 * …
 */
function haveFun($name = null) …
```

This makes phpstan think the variable can only ever be null. Then in the function body, a check like `($name ?? "fishy")` triggers an erroneous phpstan finding. Of course, we surely don't intend for `$name` here to be always null, or we wouldn't make it a parameter at all. So we probably need at least to remove all these docblock bugs, if not replace them with a more apt type.

There are also cases where a function is annotated as `@return false`, but it doesn't only return false.

kojiromike:
Fixes #8829

Fix cases where phpstan is misled by annotations saying a parameter is `null` or a function returns `false` when the type can be other things.

- [x] describe `null` and `false` union types more thoroughly
- [x] remove fixed phpstan complaints from baseline
- [x] fix new phpstan findings

### DIFF
--- a/src/Services/QuestionnaireResponseService.php
+++ b/src/Services/QuestionnaireResponseService.php
@@ -333,12 +333,12 @@ class QuestionnaireResponseService extends BaseService
     /**
      * @param       $response
      * @param       $pid
-     * @param null  $encounter
-     * @param null  $qr_id
-     * @param null  $qr_record_id
-     * @param null  $q
-     * @param null  $q_id
-     * @param null  $form_response
+     * @param ?int  $encounter
+     * @param ?string  $qr_id
+     * @param ?int  $qr_record_id
+     * @param string|array|null  $q
+     * @param ?string  $q_id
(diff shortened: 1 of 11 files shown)


src/Services/QuestionnaireResponseService.php
```diff
@@ -397,20 +397,18 @@ public function saveQuestionnaireResponse(
         }
 
         $version = 1;
-        if (!empty($fhirQuestionnaireOb)) {
```
kojiromike, line 400: were this value empty we would have hit the `throw` on 383.

adunsulag: I agree.

adunsulag (approved):
Looks good.  I didn't see any problems.

PR #8836, opened 2025-08-26
chore: Refactor observation files into MVC architecture

### LINKED ISSUE #8803: feat: Refactor form_observation to use ObservationService
### Describe the problem

We need to refactor the form_observation table to use a testable service.  This in preparation for changes to USCDI core for observation values

- [x] Refactor observation/save.php to use ObservationController which will use ObservationService
- [x] Refactor observation/report.php to use ObservationController which will use ObservationService
- [x] Refactor observation/new.php to use ObservationController which will use ObservationService
- [x] Write ObservationServiceTest class to unit test the observation functionality.

adunsulag:
Implemented a Controller for the new,report,save views of the observation form.  Implemented a service class to put observation retrieval and saving into.

Wrote unit tests to ensure the basic functionality works before we add additional data elements to the observations.

Fixes #8803


interface/forms/observation/report.php
```diff
@@ -16,68 +16,31 @@
...
+
+// Output the response
+    try {
+        // Create controller and handle request
+        $service = new ObservationService();
+        $controller = new ObservationController($service);
+        // This approach is consistent with the current design, even though the original report used session values.
+        $response = $controller->report($pid, $encounter, $cols, $id);
+        $response->send();
+        return;
```
kojiromike: Do you still need the `return;`s?

interface/forms/observation/save.php
```diff
@@ -14,125 +14,33 @@
...
+$logger = new SystemLogger();
+
+try {
+    // Create controller and handle request
+    $request = Request::createFromGlobals();
+    $service = new ObservationService();
+    $controller = new ObservationController($service);
+    $response = $controller->save($request);
+    $response->send();
+    exit;
```
kojiromike: `exit;`s are unconditional here. If you can't get rid of them, you could unify them below the block.

interface/forms/observation/report.php
adunsulag: No.  thanks for the catch.

interface/forms/observation/save.php
adunsulag: Should be safe to just remove them.

src/Services/FormService.php
```diff
@@ -57,6 +57,48 @@ public function getFormByEncounter(
...
+        if ($date == "NOW()") {
+            $sql .= "$date";
+        } else {
+            $sql .= "?";
+            array_push($arraySqlBind, $date);
+        }
+
+        $sql .= ", ?, ?, ?, ?, ?, ?, ?, ?, ?)";
+        array_push($arraySqlBind, $encounter, $form_name, $form_id, $pid, $user, $group, $authorized, $formdir, $therapy_group);
+        return sqlInsert($sql, $arraySqlBind);
```
kojiromike, line 99: Wanna use QueryUtils?

src/Services/ObservationService.php
```diff
@@ -0,0 +1,256 @@
...
+     *
+     * @param int $formId
+     * @param int $pid
+     * @param int $encounter
+     * @return array
+     */
+    public function getObservationsByFormId(int $formId, int $pid, int $encounter): array
+    {
+        $sql = "SELECT * FROM `form_observation` WHERE id=? AND pid = ? AND encounter = ?";
+        $res = sqlStatement($sql, array($formId, $pid, $encounter));
```
kojiromike, line 28: Consider using QueryUtils here. We maybe should create a Query interface and provide the implementation via dependency injection.

Alternatively, I have been considering a trait that provides any class that does queries with its own stubbable protected methods.

src/Services/ObservationService.php
```diff
@@ -0,0 +1,256 @@
...
+
+        $observations = [];
```
kojiromike, line 30:
```suggestion
        $observations = array_filter(sqlFetchArray($res));
```

src/Services/ObservationService.php
```diff
@@ -0,0 +1,256 @@
...
+                    } else {
+                        if ($key1 == $key) {
+                            $obUnitValue = $obUnit[$key1];
+                        }
+                    }
+                }
+            }
+        }
+
+        return $obUnitValue;
```
kojiromike, line 154: What about the default case when the code doesn't match? Consider a match expression here?

tests/Tests/Services/ObservationServiceTest.php
```diff
@@ -0,0 +1,289 @@
...
+
+        // In a real test environment with database:
+        // $result = $this->service->getObservationsByFormId(1, 1, 1);
+        // $this->assertIsArray($result);
+    }
+
+    public function testGetNextFormIdStructure(): void
+    {
+        // This test would require database mocking
+        $this->assertTrue(method_exists($this->service, 'getNextFormId'));
```
kojiromike, line 268: This kind of check could be done statically, right? If these are public methods, then php would fail if the method didn't exist if the class was supposed to implement an interface.

Or is this a workaround for php's lack of structural typing?

src/Services/ObservationService.php
adunsulag: Sound good, I'll put that in the next PR.  I wanted to get this initial form into a MVC format, and the next step will be overhauling the entire service class and the views to add in the pieces we need for US Core.

src/Services/ObservationService.php
adunsulag: I left this code in for now, but I'm scratching my head why its even in there.  I was tempted to just remove it, but it looks like it came in the original care coordination code back in 2015 but there's no documentation on what it is fixing, the spec it is using, or anything.  I can convert it to a match expression in the next PR, though I'm tempted to just chuck it but may need @sjpadgett on whether CareCoordination still needs this.

tests/Tests/Services/ObservationServiceTest.php
adunsulag: It was generated by AI, more as placeholders / reminders to implement the tests when we can mock the db.  I specifically wanted a unit test only w/ no DB interaction for this initial round of testing.   I didn't see any harm in leaving the methods but perhaps I'll just mark the tests as incomplete which conveys better my intent.

tests/Tests/Unit/Controllers/Interface/Forms/Observation/ObservationControllerTest.php
```diff
@@ -0,0 +1,485 @@
...
+        $this->mockService->method('getObservationTypes')->willReturn([]);
+
+        // Test that all controller methods return Response
+        $newResponse = $this->controller->new($this->mockRequest);
+        $viewResponse = $this->controller->view($this->mockRequest);
+        $reportResponse = $this->controller->report(1, 1, 2, 1);
+
+        $this->assertInstanceOf(Response::class, $newResponse);
+        $this->assertInstanceOf(Response::class, $viewResponse);
+        $this->assertInstanceOf(Response::class, $reportResponse);
```
kojiromike: Is there any way this kind of check could be done statically?

src/Controllers/Interface/Forms/Observation/ObservationController.php
```diff
@@ -0,0 +1,658 @@
...
+ *
+ * @package openemr
+ * @link      http://www.open-emr.org
+ * @author    Jacob T Paul <jacob@zhservices.com>
+ * @author    Vinish K <vinish@zhservices.com>
+ * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @author    Claude.AI on August 21st 2025
+ * @copyright Copyright (c) 2015 Z&H Consultancy Services Private Limited <sam@zhservices.com>
+ * @copyright Copyright (c) 2017-2019 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Public Domain for the pieces that were generated by Claude.AI (refactor from interface/forms/observation) view.php,new.php,save.php
```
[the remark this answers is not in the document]
kojiromike: oh, Copilot, stop being a cop.

src/Controllers/Interface/Forms/Observation/ObservationController.php
```diff
@@ -0,0 +1,658 @@
...
+        <head>
+            <?php Header::setupHeader(); ?>
+            <title><?php echo text($title); ?></title>
+        </head>
+        <body style="background: <?php echo attr_url($GLOBALS['backpic'])?>; margin: 10px;">
+        <script>
+        parent.closeTab(window.name, true);
+        </script>
+        </body>
+        </html>
```
kojiromike, line 125: Is this "view" code in the controller?

adunsulag: They'll be converted to twig files on the next iteration.

interface/forms/observation/new.php
```diff
@@ -14,265 +14,37 @@
...
+
+// Output the response
+try {
+    // Create controller and handle request
+    $request = Request::createFromGlobals();
+    $service = new ObservationService();
+    $controller = new ObservationController($service);
+    $response = $controller->new($request);
+    $response->send();
+    exit;
```
kojiromike: These exits are unconditional, too. If they can't be removed, there could be one after the block.

bradymiller (approved):

PR #8854, opened 2025-09-01
fix(sql.inc.php): check connection when setting

### LINKED ISSUE #8853: dbh connection check happens after db query
### Describe the bug

The db connection check in ['library/sql.inc.php#L93'](https://github.com/openemr/openemr/blob/master/library/sql.inc.php#L93) happens after some `$database->ExecuteNoLog` calls in the blocks that configure utf8 in the database around [line 64](https://github.com/openemr/openemr/blob/1c3d0a4940cc33cb0dc29178e37fd50868be01ec/library/sql.inc.php#L64). If the database connection is invalid, this leads to a less helpful die.

### Expected behavior

$GLOBALS['dbh'] should be validated immediately when it is being set.

kojiromike:
Fixes #8853

Fixes an issue setting db options when the db connection has not been successfully made.

Move the dbh check to just when it's assigned, before any queries are made.

### DIFF
--- a/library/sql.inc.php
+++ b/library/sql.inc.php
@@ -60,6 +60,16 @@ if ((!empty($GLOBALS["enable_database_connection_pooling"]) || !empty($_SESSION[
 $GLOBALS['adodb']['db'] = $database;
 $GLOBALS['dbh'] = $database->_connectionID;
 
+// This makes the login screen informative when no connection can be made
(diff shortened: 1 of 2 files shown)


kojiromike:
I'm confident about this change, so merging. We can revert if it causes an issue.

PR #8865, opened 2025-09-03
Patient Finder Update

### LINKED ISSUE #8864: Patient Finder Improvements
Patient finder is based on the DataTables library.  

There are features of datatables that are not being leveraged.

Features that would be useful in patient finder:
--Custom Column Widths
--Custom column contents
--Expanding patient info within the list.

psoas:
Fixes #8864 

Adds many features to patient finders, including increased customizability.

Please see community forum post for more details & screenshots:
https://community.open-emr.org/t/interface-improvements-patient-finder/26078/2

## Summary by CodeRabbit

* **New Features**
  * Patient Finder interface redesigned with expandable layout options and exact-match search toggle for improved navigation
  * New customizable patient list columns including address, next/last appointments, age, and insurance information
  * Patient detail view expansion capability for quick access to additional information
  * Enhanced responsive design across all screen sizes
  * Updated theme color palette for improved visual consistency

### DIFF
--- a/interface/main/finder/dynamic_finder.css
+++ b/interface/main/finder/dynamic_finder.css
@@ -1,0 +1,711 @@
+/* dynamic_finder.css - Styles for Patient Finder interface */
+
+
+
+/* ==========================================================================
+   Container
+   ========================================================================== */
+
+#container_div {
+    margin-top: 0.4rem !important;
+}
+
+#container_div.container #pt_table {
+    width: 100% !important;
+    min-width: 800px;
+}
+
+/* Expanded: Remove min-width constraint */
+#container_div.container-fluid #pt_table {
+    width: 100% !important;
+    min-width: auto;
+}
+
+/* When w-auto class is applied */
+#pt_table.w-auto {
+    width: auto !important;
+    min-width: auto !important;
+}
+
+
+
+/* ==========================================================================
+   Three-Section Header Layout (Tabs | OemrUI Icons | Actions)
+   ========================================================================== */
+
+.nav-tabs {
+    border-bottom: 1px solid #dee2e6;
+    margin-right: 1rem;
+}
+
+.nav-tabs .nav-link.active {
+    border-bottom-color: transparent;
+}
+
+/* Header Icons, Buttons */
+#expand_icon,
+#search_icon {
+    color: #007bff;
+    font-size: 1rem;
+    padding: 0.25rem 0.5rem;
+    margin-left: 1rem;
+    margin-right: 0rem;
+    text-decoration: none;
+    border-radius: 0.25rem;
+    transition: all 0.2s ease;
+}
+
+#expand_icon {
+    margin-left: 1rem;
+}
+#search_icon {
+    margin-left: 0rem;
+}
+
+#expand_icon:hover,
+#search_icon:hover {
+    background-color: rgba(0,123,255,0.1);
+    color: #0056b3;
+    text-decoration: none;
+}
+
+#create_patient_btn {
+    height: 2rem; 
+    line-height: 0;
+    margin-left: 2rem;
+    margin-top: 0rem;
+}
+
+
+
+/* OemrUI buttons styling next to tabs */
+#oemr-ui-buttons .oe-page-header-actions {
+    display: flex !important;
+    align-items: center;
+}
+
+#oemr-ui-buttons .oe-page-header-actions a {
+    margin-left: 0.5rem;
+    text-decoration: none;
+    color: #007bff; /* Bootstrap primary blue to match header */
+    padding: 0.125rem 0.25rem; /* Smaller padding */
+    border-radius: 0.25rem;
+    transition: all 0.2s ease;
+    font-size: 0.875rem; /* Smaller font size */
+}
+
+#oemr-ui-buttons .oe-page-header-actions a i {
+    font-size: 0.875rem !important; /* Make icons smaller */
+}
+
+#oemr-ui-buttons .oe-page-header-actions a:hover {
+    background-color: rgba(0,123,255,0.1);
+    text-decoration: none;
+    color: #0056b3; /* Darker blue on hover */
+}
+
+/* Far right section styling */
+#header-buttons-right {
+    flex-shrink: 0;
+}
+
+#header-buttons-right .btn {
+    white-space: nowrap;
+}
+
+
+/* Search elements styling in header */
+#search-elements {
+    flex-shrink: 0;
+}
+
+#search-elements #custom-search label {
+    display: flex;
+    align-items: center;
+    font-size: 0.875rem;
+    margin-bottom: 0;
+}
+
+#search-elements #custom-search input {
+    margin-left: 0.5rem;
+    width: 150px;
+}
+
+#search-elements .btn-select-search {
+    background: transparent;
+    border: none;
+    padding: 0.375rem 0.5rem;
+    margin-left: -1px;
+    border-left: 1px solid #ced4da;
+    border-radius: 0 0.25rem 0.25rem 0;
+}
+
+#search-elements .btn-select-search:focus {
+    outline: none;
+    box-shadow: 0 0 0 0.2rem rgba(0,123,255,.25);
+}
+
+#search-elements .dropdown-menu {
+    right: 0;
+    left: auto;
+}
+
+#custom-search .input-group {
+    position: relative;
+}
+
+#custom-search .input-group-append {
+    position: absolute;
+    right: 0;
+    top: 0;
+    bottom: 0;
+    z-index: 10;
+}
+
+#custom-search .btn-select-search {
+    background-color: #007bff !important;
+    color: white !important;
+    border: 1px solid;
+    border-color: #007bff !important;   
+    border-left: none !important;
+    padding: 0.25rem 0.5rem;
+    height: 100%;
+    border-radius: 0 0.25rem 0.25rem 0;
+}
+
+#custom-search .btn-select-search i {
+    color: white !important;
+}   
+
+#custom-search .btn-select-search:hover {
+    /* Darker blue on hover */
+    background-color: #0056b3 !important;
+}
+
+/* Adjust input padding to make room for button */
+#custom-search .form-control {
+    padding-right: 2rem;
+}
+
+/* Hide original search elements */
+#original-search,
+#search-moved-here {
+    display: none !important;
+}
+
+
+/* ==========================================================================
+   DataTables Integration & Overrides
+   ========================================================================== */
+
+/* Hide DataTable's built-in search and old header buttons */
+.dataTables_filter {
+    display: none !important;
+}
+
+div.dataTables_wrapper div#header-buttons {
+    display: none;
+}
+
+div.dataTables_wrapper #search {
+    display: none !important;
+}
+
+/* Style the moved length selector */
+.dataTables_length {
+    float: left;
+    margin-right: 1rem;
+}
+
+.dataTables_length select {
+    width: auto !important;
+    display: inline-block;
+    margin: 0 0.5rem;
+}
+
+
+/* ==========================================================================
+   Existing DataTables Style Overrides
+   ========================================================================== */
+
+/* Finder Processing style */
+div.dataTables_wrapper div.dataTables_processing {
+    width: auto;
+    margin: 0;
+    color: var(--danger);
+    transform: translateX(-50%);
+}
+
+/*div.dataTables_wrapper div#header-buttons {
+    padding-top: 0;
+    padding-left: 2rem;
+    display: inline-block;
+}*/
+
+div.dataTables_wrapper div#select-search {
+    padding-top: 0;
+    padding-left: 0;
+    display: inline-block;
+}
+
+div.dataTables_wrapper a#exp_cont_icon {
+    padding-top: 0rem;
+    padding-left: 1.5rem;
+    font-size: 1.5rem !important;
+    display: inline-block;
+    vertical-align: middle;
+    color: var(--primary) !important;
+}
+
+div.dataTables_wrapper i#show_hide {
+    padding-top: 0rem;
+    padding-left: 1rem;
+    font-size: 1.5rem;
+    display: inline-block;
+    vertical-align: middle;
+    color: var(--primary) !important;
+}
+
+div.dataTables_wrapper .my_bottom_div {
+    padding-left: 0.75rem;
+    padding-top: 0.5rem;
+}
+
+.card {
+    border: 0;
+    border-radius: 0;
+}
+
+
+/* Table Headers */
+#header {
+    margin: 0 !important;
+}
+
+table.dataTable #header-top {
+    padding: 0 !important;
+}
+
+
+/* Color Overrides for jQuery-DT */
+table.dataTable thead th,
+table.dataTable thead td {
+    /*border-bottom: 1px solid var(--gray900) !important;*/
+    border: 0 !important;
+    padding: 0.25rem 0.25rem;
+}
+
+table.dataTable thead #pt_table_header_search_boxes {
+    border-bottom: 0;
+    background-color:  #007bff !important;
+}
+
+table.dataTable thead #pt_table_header_search_boxes td {
+    padding-top: 5px;
+    padding-bottom: 5px !important;
+    padding-left: 10px;
+    padding-right: 10px;
+}
+
+table.dataTable thead #pt_table_header_search_boxes td input.search_init {
+    height: 1.7rem;
+}
+
+table.dataTable thead tr.header-labels th {
+    border-bottom: 0;
+    padding: 0.25rem 0.75rem;
+}
+
+table.dataTable thead tr.header-labels th.dt-control {
+    padding-left: 0.1rem;
+    padding-right: 0.1rem;
+}
+
+table.dataTable tfoot th,
+table.dataTable tfoot td {
+    /*border-top: 1px solid var(--gray900) !important;*/
+    border-top: 0;    
+}
+
+table.dataTable tbody tr {
+    background-color: var(--white) !important;
+    cursor: pointer;
+}
+
+table.dataTable tbody td {
+    padding: 0.1rem 0.75rem;
+    border-bottom: 1px solid var(--primary) !important;
+}
+
+table.dataTable tbody td.dt-control {
+    padding: 0.1rem 0.2rem;
+    border-bottom: 1px solid var(--primary) !important;
+}
+
+table.dataTable.row-border tbody th,
+table.dataTable.row-border tbody td,
+table.dataTable.display tbody th,
+table.dataTable.display tbody td {
+    border-top: 1px solid var(--gray300) !important;
+}
+
+table.dataTable.cell-border tbody th,
+table.dataTable.cell-border tbody td {
+    border-top: 1px solid var(--gray300) !important;
+    border-right: 1px solid var(--gray300) !important;
+}
+
+table.dataTable.cell-border tbody tr th:first-child,
+table.dataTable.cell-border tbody tr td:first-child {
+    border-left: 1px solid var(--gray300) !important;
+}
+
+table.dataTable.stripe tbody tr.odd,
+table.dataTable.display tbody tr.odd {
+    background-color: var(--light) !important;
+}
+
+table.dataTable tbody tr:hover,
+table.dataTable tbody tr:hover a,
+table.dataTable.display tbody tr:hover,
+table.dataTable.display tbody tr:hover a {
+    background-color: var(--secondary) !important;
+    text-decoration: none !important;
+    color: var(--light);
+}
+
+table.dataTable tbody a:hover,
+table.dataTable.display tbody a:hover {
+    text-decoration: none !important;
+    color: var(--light);
+}
+
+table.dataTable.order-column tbody tr>.sorting_1,
+table.dataTable.order-column tbody tr>.sorting_2,
+table.dataTable.order-column tbody tr>.sorting_3,
+table.dataTable.display tbody tr>.sorting_1,
+table.dataTable.display tbody tr>.sorting_2,
+table.dataTable.display tbody tr>.sorting_3 {
+    background-color: var(--light) !important;
+}
+
+table.dataTable.display tbody tr.odd>.sorting_1,
+table.dataTable.order-column.stripe tbody tr.odd>.sorting_1 {
+    background-color: var(--light) !important;
+}
+
+table.dataTable.display tbody tr.odd>.sorting_2,
+table.dataTable.order-column.stripe tbody tr.odd>.sorting_2 {
+    background-color: var(--light) !important;
+}
+
+table.dataTable.display tbody tr.odd>.sorting_3,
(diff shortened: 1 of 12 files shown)


psoas:
@bradymiller @adunsulag @kojiromike — could one of you give this PR a review?

What it does: it improves the patient finder, which is a tool nearly every OpenEMR user touches daily. The change is a meaningful usability improvement there, and it's small and self-contained — low risk to review and merge.

The history is the frustrating part. I first submitted this in 2022 and reopened it about 9 months ago. It hasn't been reviewed in that time, and it's now drifted out of sync with main, so I'm looking at refactoring it for the third time just to keep it mergeable.

I'm glad to do that refactor and address any review feedback promptly. All I need is for someone to put eyes on it, and assist me to get it approved/merged. Given how central the patient finder is, I think this earns a look.

I'll also name the broader thing gently: feature PRs from outside contributors have seemed to stall on other occasions, and I suspect it quietly discourages people from contributing features at all. I don't think that's anyone's intent — certification and infra work understandably dominate right now — but a habit of giving those contributing PRs more assistance to getting them approved would go a long way. Glad to take the process conversation to the forum if that's a better venue. Thanks.

interface/main/finder/dynamic_finder.css
```diff
@@ -0,0 +1,711 @@
+/* dynamic_finder.css - Styles for Patient Finder interface */
+
+
+
+/* ==========================================================================
+   Container
+   ========================================================================== */
+
+#container_div {
+    margin-top: 0.4rem !important;
```
kojiromike, line 10: We probably need to do something about having `!important` everywhere. Not sure this can be fixed in this pr, but it's a problem all over the repo.

bradymiller:
@coderabbitai review

bradymiller:
Claude review:

# Code Review: openemr/openemr#8865 — Patient Finder Update

Review for scoping the work needed before this can be merged. Rhe feature itself is a real usability win, but as written, this PR has several blockers that need fixing, and the work is non-trivial.

## TL;DR

**Roughly 1.5–2 weeks of focused contributor work to make this mergeable**, on top of a rebase. The split is approximately:

- **Security / regression (must-fix before merge)** — 3–4 days
- **Feature correctness (must-fix before merge)** — 3–5 days
- **Upgrade SQL rewrite (must-fix before merge)** — 1–2 days
- **Polish/cleanup (should-fix)** — 2–3 days

Three independent reviewers (Copilot, CodeRabbit, and this review) converge on the same blocker set, which raises confidence that nothing major is going unnoticed. Copilot caught ~18 issues. CodeRabbit independently confirmed most of those, validated the multi-level escape path through `display_layout_tabs_data()`, and added several findings nobody else flagged: dead CSS selectors that target IDs that don't exist in the Twig template, an explicit `prevSetting()` deprecation as of 7.0.3, a silent data-corruption path in the admin list editor when an existing `option_id` isn't in the new dropdown, a duplicate `id="finderTabs"` on the tab list and the tab-content container, and a click handler bound to a stale `#exp_cont_icon` that means the expand-to-fullwidth feature does not actually work in this PR. The new findings *this* review added beyond Copilot+CodeRabbit are the IDOR/CSRF gap on the detail endpoint, the destructive upgrade-SQL pattern that wipes site customizations, the stale-rebase regression that silently disables module-based patient blacklisting, and the realization that the PR is out of sync with master's `OEGlobalsBag` migration.

## Blockers — security & access control

### 1. `interface/main/finder/patient_data_ajax.php` has no ACL check
The endpoint reads `$_GET['pid']` and returns full `SELECT * FROM patient_data WHERE pid = ?` — SSN, DOB, contact info, billing notes — to the caller. There's no `AclMain::aclCheckCore('patients', 'demo', ...)`, so any authenticated user (regardless of patient-data ACL) can enumerate every patient by URL fuzzing. Master's sibling endpoints all gate on `patients|demo`. This must be added.

### 2. Same endpoint has no CSRF check
Most OpenEMR AJAX endpoints verify `CsrfUtils::verifyCsrfToken($_POST['csrf_token_form'])`. The companion `dynamic_finder_ajax.php` documents *why* it skips CSRF (line 29: "no state changes" + popup-from-new-window UX) — that reasoning carries over to `patient_data_ajax.php`, but it should be documented there too. The bigger issue is the IDOR above.

### 3. `dynamic_finder_ajax.php` silently breaks PatientFinderFilterEvent
Line 220 of the PR:
```php
//$patientFinderFilterEvent = $GLOBALS["kernel"]->getEventDispatcher()->dispatch(PatientFinderFilterEvent::EVENT_HANDLE, $patientFinderFilterEvent);
```
The dispatch is commented out. Master has it live:
```php
OEGlobalsBag::getInstance()->getKernel()->getEventDispatcher()->dispatch($patientFinderFilterEvent, PatientFinderFilterEvent::EVENT_HANDLE);
```
Modules subscribing to this event (notably the PatientFilter blacklist module, used to hide restricted patients from finders) will *silently* stop working after this PR merges. Two issues stacked: the dispatch was probably commented out for local debugging, AND the PR's base predates the `$GLOBALS["kernel"]` → `OEGlobalsBag::getInstance()->getKernel()` API migration. Needs a fresh rebase plus uncomment.

### 4. JS injection via `option_id` in `dynamic_finder.php:84`
```php
$colname_escaped = addcslashes($colname, "\t\r\n\"\\");
$coljson .= "{ name: '$colname_escaped', data: '$colname_escaped'";
```
`addcslashes` doesn't include the single quote, but the value is embedded into a single-quoted JS string. An admin who edits a `ptlistcols` `option_id` to contain `'` can break out of the JS context. Use `js_escape()` or `json_encode($colname)` — same pattern is used correctly at `dynamic_finder.php:314`.

### 5. `patient_data_ajax.php` returns `SELECT *`
Even if the ACL is added, returning the full row to `display_layout_tabs_data()` exposes every patient_data column — SSN, billing fields, employer fields, internal notes — to the JSON encoding/HTML buffer, regardless of which fields the layout actually renders. Restrict the SELECT to the columns the layout needs, or filter `$out` before rendering.

## Blockers — upgrade SQL is destructive and non-idempotent

### 6. `UPDATE list_options SET activity = 0 WHERE list_id = 'ptlistcols'` is data loss
```sql
-- sql/7_0_3-to-7_0_4_upgrade.sql:470
UPDATE `list_options` SET `activity` = 0 WHERE `list_id` = 'ptlistcols';

INSERT INTO `list_options` (...) VALUES
    ('ptlistcols', 'pubpid', 'ID', ...),
    ('ptlistcols', 'name', 'Full Name', ...),
    ...
ON DUPLICATE KEY UPDATE ...;
```
This deactivates every existing `ptlistcols` entry, including any custom column an OpenEMR site has added over time. The subsequent INSERT+ON DUPLICATE only re-activates the PR's hardcoded list. Sites with custom patient-finder columns will silently lose them on upgrade. The migration should be additive: only deactivate the specific legacy entries the PR is replacing.

### 7. No `#IfNotRow` guards anywhere in the patient-finder block
None of the inserts (layout_group_properties, layout_options, list_options for `sort_direction_full`) carry the upgrade-script idempotency guards that the rest of the file uses (`#IfNotRow list_options list_id ecqm_reporting_period` etc.). Re-running the upgrade — which happens in CI, on partial-failure retries, and during migration testing — will either error out on duplicate keys or repeatedly destroy and reinsert state. Wrap each chunk in `#IfNotRow`.

### 8. `sort_direction_full` is missing from the `lists` discovery row in `database.sql`
Confirmed: `database.sql` adds the two options at lines 5590–5591 but doesn't INSERT the `('lists', 'sort_direction_full', 'Sort Direction Full', ...)` row that makes the list discoverable in the List editor UI. Fresh installs will have the values but no listing. (The upgrade SQL has it; database.sql needs to mirror.)

### 9. List id casing mismatch
SQL inserts `'sort_direction_full'` (lowercase) but `edit_list.php:541` calls `generate_select_list("...", 'Sort_Direction_Full', ...)`. Result: empty dropdown on the column editor. Pick one casing and use it everywhere — the codebase generally goes lowercase for list_ids.

### 10. Sort direction values disagree with DataTables
```sql
('sort_direction_full', '0', 'Ascending', ...),
('sort_direction_full', '1', 'Descending', ...);
```
`dynamic_finder.php:70` reads the `title` (`Ascending` / `Descending`) and injects it directly into the DataTables `order` config, which expects `asc` / `desc`. The author's comment on that line — "Get the title 'asc' or 'desc'" — shows the intent, but the SQL doesn't match. Either change the titles to `asc` / `desc` (with display labels separate) or translate in PHP.

## Blockers — feature correctness

### 11. Sort by a column-view produces invalid SQL
```php
// dynamic_finder_ajax.php:80
$orderby .= $orderby ? ', ' : 'ORDER BY ';
...
} elseif (is_column_view($aColumns[$iSortCol])) {
    $orderby = substr($orderby, 0, -2);
}
```
When a column view is the first sort column, `$orderby` becomes `'ORDER BY '` (9 chars) and `substr(..., 0, -2)` truncates to `'ORDER B'` — invalid SQL. The "strip trailing separator" pattern only works when a separator was *just appended* by the current iteration; the column-view branch doesn't append anything. Either skip the separator append for column views, or map them to their underlying real columns (preferred, see #14).

### 12. Column-view search hardcodes `usertext8` for every view
```php
// dynamic_finder_ajax.php:159, 199
} elseif (is_column_view($colname)) {
    $where .= "usertext8 LIKE ?";
    array_push($srch_bind, ($sSearch . '%'));
}
```
`usertext8` is an unrelated user-defined text field. A search on `city_state` looks at `usertext8`, not city/state. Same hardcoded value in both global and column-specific filter blocks. Either build the WHERE from `column_view_select_list($colname)`'s underlying columns, or mark column-view columns non-searchable for now.

### 13. N+1 query pattern in `column_view_next_appointment` / `column_view_last_appointment`
Each runs one query per patient row. With page-size 250 + these columns enabled, that's ~500 extra queries per page load against `openemr_postcalendar_events`. On a real-sized practice this stalls the finder. Options: precompute via a JOIN/subquery in the main finder query, or only resolve on row expansion.

### 14. `column_view_last_appointment` crashes silently when there's no prior appointment
```php
// library/column_views.inc.php:208
$row = sqlFetchArray($result);
$event_date = $row['pc_eventDate'];
$event_time = $row['pc_startTime'];
$last_appointment_datetime = new DateTime($event_date . " " . $event_time);
```
No `if ($event_date && $event_time)` guard like the next-appointment version has. When there's no prior appointment, `new DateTime(" ")` returns the current time, so the column displays *now* as the "last appointment" — wrong and misleading.

### 15. `column_view_age` blows up on missing/invalid DOB
```php
$dob = $row['DOB'];
$age = date_diff(date_create($dob), date_create('now'))->y;
```
`date_create('')` and `date_create(invalid)` return `false`. `date_diff(false, ...)` errors, and dereferencing `->y` then crashes the AJAX response (JSON output gets corrupted, the table fails to render). Guard against missing/invalid DOB.

### 16. Two "column views" return empty stub strings
```php
function column_view_insurance_primary_provider_name($row) {
    // TODO: Implement insurance primary provider name retrieval
    return "";
}
function column_view_insurance_primary_policy_number($row) {
    // TODO: ...
    return "";
}
```
These are wired into the dropdown so admins can select them, but they produce a blank column. Either implement them or remove them from `$column_views` for now.

### 17. Triple-escaping in `column_views.inc.php` (worse than it looks)
Multiple `column_view_*()` functions return `attr($result)` — already HTML-escaped. Then `all_column_views()`, `column_view()`, and `column_views_list()` wrap the result in `attr(...)` again. CodeRabbit traced one more layer that I missed: when `display_layout_tabs_data()` receives the `$out` array in `patient_data_ajax.php`, it escapes a *third* time via `attr()` at line ~3906 of `library/options.inc.php`, and `generate_display_field()` then escapes again via `text()` for the JSON-encoded HTML. End result: `&` becomes `&amp;amp;amp;` in the expanded row. Pick one layer of escaping responsibility (recommend: leaf `column_view_*()` functions return raw, wrappers escape exactly once, callers trust the contract).

### 17b. The `age` column-view's `require` field is empty but it needs `DOB`
```php
"age" => array(
    "function" => "column_view_age",
    "require" => ""
),
```
`column_view_age()` reads `$row['DOB']`. With `require` empty, the SELECT list doesn't include `DOB`, so age is always wrong (or crashes per #15). Set `"require" => "DOB"`.

### 18. `column_view_name` produces "Last, Middle" when first name is empty
```php
$name = $row['lname'];
if ($name && ($row['fname'] || $row['mname'])) {
    $name .= ', ';
}
if ($row['fname']) { $name .= $row['fname']; }
if ($row['mname'] && $row['fname']) { $name .= ' '; }
$name .= $row['mname'];
```
If `mname` is set but `fname` is empty, output is `"Doe, Quincy"` — no name convention works that way. Don't add the comma unless first name exists, or always omit the middle when first is empty.

### 19. Hardcoded `colspan='7'` in `templates/patient_finder/finder.html.twig:172`
The column count is dynamic (driven by `ptlistcols`), but the expand-detail template assumes 7. Sites with fewer/more columns get a misaligned or truncated detail row. Either pass `colcount` to the template or compute the colspan in JS at insertion time.

### 20. `noHover { pointer-events: none }` disables interaction inside the expanded detail
`display_layout_tabs_data()` can render links and tabs in the detail — these become unclickable. The row-click handler already short-circuits when `pid` is empty, so disabling pointer events is redundant.

### 21. Accessibility *and* DOM-validity: tab semantics are broken
Two stacked bugs in `templates/patient_finder/finder.html.twig`:
- The inactive `recentTab` button has `aria-selected="true"` (line 23). Screen readers report two simultaneously-selected tabs (Copilot finding).
- Both the `<ul class="nav nav-tabs">` (line ~17) and the `<div class="tab-content">` (line ~80) carry `id="finderTabs"` — duplicate IDs in the same document. Bootstrap's tab JS uses those selectors and at least one of the two bindings won't behave correctly. Rename the tab-content container (e.g., `id="finderTabContent"`). (CodeRabbit finding.)

### 21b. Dead click handler — the expand-to-fullwidth feature does not work
`templates/patient_finder/finder.html.twig:187`:
```js
$("#exp_cont_icon").click(function () {
    $("#pt_table").removeAttr("style");});
```
The template renders the icon as `#expand_icon` (PHP-side handler defined in `dynamic_finder.php`), not `#exp_cont_icon`. The selector never matches and the callback never fires — the "expand to fullwidth" interaction was likely the centerpiece of the responsive UX rework but is silently dead. Rename to `#expand_icon`. (CodeRabbit finding.)

## High-priority bugs & smells

### 21c. The new admin column-editor can *silently rewrite* existing `ptlistcols` rows
`interface/super/edit_list.php` around line 419: the new `<select>` is populated only from the current `patient_data` columns + `column_views_list()`. If an `option_id` that already exists in the database isn't in that list (e.g., a site-custom column that the migration's destructive `UPDATE` deactivated, or an `option_id` added by a third-party module), the browser falls back to the *first* option (which after lines 410 + 412 is blank). When the admin clicks save, that row is silently rewritten with an empty `option_id`. This is admin-form-level data corruption — combine it with the destructive upgrade-SQL (#6) and a site's customizations could disappear in two steps: (1) upgrade SQL deactivates them, (2) admin opens the list editor to investigate and the option_id gets overwritten to blank on save.

Fix: before rendering the `<select>`, append `$option_id` to the candidates list if it's not already present, so the existing value always has a matching `<option>`. CodeRabbit's suggested patch is the right shape. (CodeRabbit finding.)

### 22. `edit_list.php:404` runs `information_schema.COLUMNS` per option row
```php
if ($list_id == 'ptlistcols') {
    $query = "SELECT COLUMN_NAME FROM information_schema.COLUMNS WHERE TABLE_NAME = 'patient_data'";
```
Two issues: (a) no `TABLE_SCHEMA = DATABASE()` filter — multi-tenant deployments with another `patient_data` table elsewhere get cross-contaminated results; (b) `writeOptionLine()` is called once per option row, so this query runs N times when editing the list. Cache once per request.

### 23. `console.log(PatientDetailHTML)` in production code
`dynamic_finder.php:236`. Leaves full patient demographics in the browser console on every row expand. Remove before merge.

### 24. `pid` not URL-encoded in fetch
`dynamic_finder.php:233`: `let url = "patient_data_ajax.php?pid=" + pid;`. Currently safe because `pid` is a numeric DB id parsed from the row id, but cheap insurance to `encodeURIComponent(pid)`.

### 25. `searchAny` re-explodes `sColumns` at line 229 after using a different column set at line 47
The WHERE clause was built using `$aColumns` derived from `layout_options`. Then the SELECT-list code re-reads `$_GET['sColumns']` and overrides. SELECT and WHERE end up referring to potentially different column sets. Either intentional and undocumented, or a real bug.

### 26. `subtype` column hijacked for CSS width strings
The PR stores values like `'10rem'` and `'1.5rem'` in `list_options.subtype`. That column's existing semantic varies by list (some use it for numeric subtypes, some leave it null). Reusing it for a free-form CSS string is fragile and confusing to future readers. Either pick a dedicated column or use `notes` with a documented prefix.

### 27. `prevSetting()` *is* deprecated as of 7.0.3 — must be replaced
`dynamic_finder.php:36-38`: the author left an inline `//Is prevSetting deprecated? does it work?` comment. CodeRabbit confirmed: yes, `prevSetting()` is the deprecated wrapper, and the canonical call is now `UserSettingsService::prevSetting()`. The two `prevSetting()` calls (`patient_finder_exact_search` and `dynamic_finder_xpd`) need to be migrated, with an added `use OpenEMR\Services\UserSettingsService;` at the top of the file. Removing the wrapper also kills the speculative comment.

### 28. PHPStan baseline change (`phpstan.github.neon`)
12 lines removed from the baseline. Verify those errors actually no longer fire on the changed code — if they're related to files the PR didn't fix, removing them from the baseline will turn red on CI.

### 29. Pile of commented-out code throughout `dynamic_finder.php`
Lines 86, 87–89, 171–172, 196–201, 220 (the critical regression). Cleanup pass needed. Don't merge commented-out code.

### 30. kojiromike's `!important` note is fair but out of scope
The new `dynamic_finder.css` file does lean heavily on `!important`. It mirrors the codebase's existing pattern (so blocking on it isn't reasonable), but adding 711 lines of `!important`-laden CSS expands the surface that any future Bootstrap-5-style migration has to retrofit (see #10119). Worth a comment to author asking whether scoped selectors could replace some of the `!important` uses.

### 31. A noticeable fraction of `dynamic_finder.css` is dead — selectors don't match the new Twig DOM
CodeRabbit flagged that the new stylesheet targets `#header-buttons-right` and `#oemr-ui-buttons .oe-page-header-actions`, but the Twig template defines `#header-right` and places action links directly under `#oemr-ui-buttons`. So the header/responsive rules at lines 105, 108–114, 654–657, 679–689, and 705–708 silently don't apply. The header probably still looks "okay" because Bootstrap defaults take over, but the carefully-engineered breakpoint behavior in the CSS is purely cosmetic. Either rename the selectors to match the template, or rename the template IDs to match the CSS — and verify visually after.

### 32. Inline `style="..."` overrides defeat the responsive breakpoints
`templates/patient_finder/finder.html.twig:37-44` hardcodes `border-right: none`, `border-left: none`, `background: white` on the search input and its button. Those inline styles override the new stylesheet's mobile breakpoints, so the responsive search-control behavior is partially undone before it gets a chance to apply. Move to classes and let the breakpoints win.

### 33. Docstring coverage flagged by CodeRabbit (11% vs 80% threshold)
Not a blocker for OpenEMR's bar, but a useful signal — the new `column_views.inc.php` and `patient_data_ajax.php` ship with essentially no function-level documentation. Worth a paragraph at the top of `column_views.inc.php` explaining the registry shape (`function` + `require`), the escaping contract (whatever you decide on under #17), and that adding a new view is one row + one function. The next contributor extending this will need that.

## Scope of work to merge

In rough order of dependency:

1. **Rebase against current master.** This is gating: bring back master's `OEGlobalsBag::getInstance()->getKernel()` dispatch, and uncomment the line — fix the silent regression in one step.
2. **Lock down `patient_data_ajax.php`**: add `aclCheckCore('patients', 'demo')`, validate `$pid` as a positive int (CodeRabbit's `filter_input(INPUT_GET, 'pid', FILTER_VALIDATE_INT)` is the right shape), restrict the `SELECT *` to columns the layout actually renders, handle the empty-row case, and decide on CSRF behavior (probably skip with documented rationale, matching the sibling endpoint).
3. **Rewrite the upgrade SQL block**: wrap each statement in `#IfNotRow2D` (CodeRabbit's suggested guards are correct), scope the destructive `UPDATE ... SET activity=0` to *just the option_ids the migration is replacing* (not all `ptlistcols`), fix the `sort_direction_full` ↔ `Sort_Direction_Full` casing across PHP + both SQL files, add the `('lists', 'Sort_Direction_Full', 'Sort Direction Full', ...)` registration row in `database.sql`, and decide whether the option values should be `asc/desc` (what DataTables needs) or `Ascending/Descending` with a PHP-side translation.
4. **Fix the escaping contract in `column_views.inc.php`**: decide on a single layer (leaf functions return raw, wrappers escape once is the cleanest), strip the now-triple `attr()` chain through `display_layout_tabs_data()`, add `"DOB"` to the `age` view's `require` field, and either implement or remove the two insurance stubs.
5. **Fix column-view feature correctness**: ORDER BY clipping when sort starts on a column view, replace hardcoded `usertext8 LIKE ?` with WHERE-clauses built from each view's `require` fields, address the N+1 appointment queries (precompute or load-on-expand), guard the appointment functions for empty result rows, guard `column_view_age` against missing/invalid DOB.
6. **Fix the JS and template issues**: `json_encode` instead of `addcslashes` for `option_id` interpolation, remove `console.log`, `encodeURIComponent(pid)` on the fetch URL, add error handling on the fetch, replace hardcoded `colspan='7'` with `{{ colcount|attr }}`, fix `aria-selected` and the duplicate `id="finderTabs"` on the tab container, rename `#exp_cont_icon` → `#expand_icon` in the click handler, drop the inline `style="border-right: none"` overrides on the search input.
7. **Fix the admin form** in `interface/super/edit_list.php`: preserve existing `option_id`s by appending them to the dropdown candidates when not already present (CodeRabbit's patch), cache the `information_schema.COLUMNS` lookup once per request, scope it to `TABLE_SCHEMA = DATABASE()`.
8. **Fix the dead CSS selectors** in `dynamic_finder.css` (lines 105 + the 4 sibling regions): rename to match the actual Twig DOM (`#header-right`, `#oemr-ui-buttons` directly), then verify the header looks right on each breakpoint.
9. **Cleanup pass**: remove commented-out code, replace `prevSetting()` with `UserSettingsService::prevSetting()`, decide on the `subtype`-as-CSS-width hijack vs a new column, add at least minimal docstrings to `column_views.inc.php`.
10. **Verify the PHPStan baseline removal** still passes.

Feature itself is worth doing — the dynamic-column / expandable-row UX is a real improvement, the column-view abstraction is the right shape, and the three independent reviews converge cleanly on the same fixable set. ~1.5–2 weeks of contributor work concentrated in 4 files. Worth investing the review bandwidth.

## On the contributor's process note

Their comment about feature PRs stalling is fair feedback worth acknowledging separately. The fix here is small enough that it's worth committing maintainer time to walk through these findings together rather than leaving another 9-month gap — they've shown willingness to address review feedback and the feature has clear user value.

PR #8871, opened 2025-09-05
ci: run against release branches

### LINKED ISSUE #8867: Automated tests don't run against release branches
Sorry I didn't realize this sooner. All the Github actions target `master`, or pull requests against `master`. That means all patchsets and whatnot going into `rel-.*` don't get any automated tests. I think we should correct this by adding release branches as targets to the actions.

kojiromike:
Fixes #8867

Ensure CI runs against release branches as well as master.

- [x] Update the targets for all the targeted actions so they run against `rel-*`.

### DIFF
--- a/.github/workflows/inferno-test.yml
+++ b/.github/workflows/inferno-test.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
     - master
+    - rel-*
   pull_request:
     branches:
     - master
+    - rel-*
 
 permissions:
   contents: read
--- a/.github/workflows/isolated-tests.yml
+++ b/.github/workflows/isolated-tests.yml
@@ -11,9 +11,11 @@ on:
   push:
     branches:
     - master
+    - rel-*
   pull_request:
     branches:
     - master
+    - rel-*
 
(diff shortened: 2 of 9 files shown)


bradymiller:
hi @kojiromike , This is nice. I didn't realize this would work. I have been changing those branch names when branch the code. I am assuming it then just runs the pertinent branch that the PR/merge is currently on, correct?

kojiromike:
> hi @kojiromike , This is nice. I didn't realize this would work. I have been changing those branch names when branch the code. I am assuming it then just runs the pertinent branch that the PR/merge is currently on, correct?

Right, if you open a pr against master, tests run. If you open a pr against rel-1234, currently no tests will run. After this change, they will. Also after this change, if you merge/push a change to rel-1234, tests will run on that branch.

Of course we're talking about the actions in the pr or branch in question, so if a particular set of tests are in master and not in rel-1234 or the pr, then those tests will only run in master.

bradymiller:
Nice! approved (unable to get to github until this evening, so approved via
comment)

PR #8876, opened 2025-09-06
Add Edit Event validation fix

### LINKED ISSUE #8783: QRDA III Zip Download fails
### Describe the bug

I committed a wrong file version of the QRDAReportController thus a new PR to merge correct file!
### To Reproduce

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version: dev master 704
- Operating system:

### LINKED ISSUE #8852: Can't add Provider events
### Describe the bug

Due to recent PR #8263 that turned off ability to add duration to In Office events because it didn't schedule the duration i.e pointless, duration was disabled however, validation won't let user save because duration is not populated.

So fix validation and continue to use duration set in Config or fix duration so user may adjust providers schedule how they see fix. e.g. 15 minutes on Monday and 25 minutes on Tuesday or Provider A wants 15 and Provider B wants 30 etc..

I could use some advice from clinical point of view.

- Browser:
- OpenEMR version:
- Operating system:

sjpadgett:
allow duration to be 0 for all day event scheduling.

Fixes #8852 
Fixes #8783

### DIFF
--- a/interface/main/calendar/add_edit_event.php
+++ b/interface/main/calendar/add_edit_event.php
@@ -1067,8 +1067,8 @@ function set_display() {
             style_prefcat.display = '';
             f.form_apptstatus.style.display = 'none';
             f.form_prefcat.style.display = '';
-            f.form_duration.disabled = true;
-            f.form_duration.value = '';
+            f.form_duration.disabled = false;
+            f.form_duration.value = 0;
             document.getElementById('tdallday4').style.color = 'var(--gray)';
         } else {
             style_prefcat.display = 'none';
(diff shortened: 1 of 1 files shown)


kojiromike (approved):
Awesome, thanks, @sjpadgett

interface/main/calendar/add_edit_event.php
```diff
@@ -1915,6 +1919,7 @@ function validateform(event,valu){
             message: "Duration is required"
         }
     };
+    }
```
kojiromike, line 1922: Should 1911-1921 be indented to make the blocks clearer?

sjpadgett: true @kojiromike but don't want to change forcing retests!:) I'd do till next time I'm there which for this feature seems like soon....

PR #8889, opened 2025-09-15
Openemr fix #8801 observation form us core

### LINKED ISSUE #8801: feat: Improve Observation Form for US Core
### Describe the problem

Currently our observation form does not handle any of the required linkage of questionnaire responses nor of parent observations (heirarchical).  We also are missing the ability to categorize observations into broad categories that we can search on in addition to querying off a specific observation code. 

I'd like to revamp the Observation Form interface to have the following features:
- [ ] Category dropdown for selecting an observation category (functional, cognitive, sdoh, etc)
- [ ] 0..1 linkage to the QuestionnaireResponse that created this observation, with the ability to link an observation to a QuestionnaireResponse for manual override / editing capability
- [ ] Ability to add 0..* sub-observations that are linked to the current observation.  
- - [ ] Sub-observations can be linked from existing observations that have no parent, or added directly inside the parent observation.  Linked issues must have the same type, category, and Y-m-d of the parent issue.
- - [ ] Sub-observations are added with an option code, value, and units
- - [ ] Sub observations can be deleted or unlinked (which turns them into parent issues)
- - [ ] Have a list screen w/ summary information of the observations for the encounter
- - [ ] Have an edit/add screen for the observation.
- - [ ] The add screen is opened by default if there are no existing observations for the encounter.
- - [ ] Saving an observation will return the user to the summary screen.

adunsulag:
Enhances the observation form #8801 

Updated the observation form GUI to have a list view of the current
observations and an edit view of the observation with the ability to
have sub-observations, and linkage to a Questionnaire.

Changed the observation delete mechanism.  Observations are now set to have the forms hidden instead of deleting from the database which is similar to how other data records work in the system.

Added ability for code description to be editable in case of adding your
own codes.

Added all of the new observation types needed for USCDI simple
observation conformity.

Added observation status field to allow status to be set.  Defaults to
preliminary status.

Date defaults to current date when adding an observation.

Added ability to have sub-observations to handle cases where you have a
panel observation.

Added search field on observations.  It is still very rudimentary and I
may take it out as it just handles it client side which isn't helpful
for dealing with lots of observation data.

Added in some basic form validation.

Removed the legacy syndrmoic surveillance hard-coded LOINC code types
which had no explanation of what they did.

Refactored HTML code to use twig.

Removed old html files

### DIFF
--- a/interface/forms/procedure_order/common.php
+++ b/interface/forms/procedure_order/common.php
@@ -647,7 +647,7 @@ if (!empty($row['lab_id'])) {
         // we want to setup our reason code widgets
         window.addEventListener('DOMContentLoaded', function () {
             if (oeUI.reasonCodeWidget) {
-                oeUI.reasonCodeWidget.init(<?php echo js_url($GLOBALS['webroot']); ?>, <?php echo js_url(collect_codetypes("problem", "csv")) ?>);
+                oeUI.reasonCodeWidget.init(<?php echo js_url($GLOBALS['webroot']); ?>, <?php echo js_url(collect_codetypes("medical_problem", "csv")) ?>);
             } else {
                 console.error("Missing required dependency reasonCodeWidget");
                 return;
--- a/src/Controllers/Interface/Forms/Observation/ObservationController.php
+++ b/src/Controllers/Interface/Forms/Observation/ObservationController.php
@@ -121,7 +121,7 @@ class ObservationController
                 'csrf_token' => CsrfUtils::collectCsrfToken(),
                 'apiCsrfToken' => CsrfUtils::collectCsrfToken('api'),
                 'title' => xl('Observation Form'),
-                'reasonCodeTypes' => $this->codeTypeService->collectCodeTypes("problem", "csv"),
+                'reasonCodeTypes' => $this->codeTypeService->collectCodeTypes("medical_problem", "csv"),
                 'linkedQuestionnaireResponse' => $observation['questionnaire_response'] ?? null,
                 // AI Generated: FHIR configuration for QuestionnaireResponse dialog
                 'fhir_config' => [


src/Services/CodeTypesService.php
```diff
@@ -429,4 +429,67 @@ public function parseCodesIntoCodeableConcepts($codes)
...
+                if ($ct_arr['problem']) {
+                    $return[] = $ct_key;
+                }
+            } elseif ($category == "drug") {
+                if ($ct_arr['drug']) {
+                    $return[] = $ct_key;
+                }
+            } else {
+                //return nothing since no supported category was chosen
+            }
```
kojiromike: consider a match expression here?

adunsulag: Fixed

src/Services/CodeTypesService.php
```diff
@@ -429,4 +429,49 @@ public function parseCodesIntoCodeableConcepts($codes)
...
+        if ($return_format == "csv") {
+            //return it as a csv string
+            return csv_like_join($return);
+        }
+
+        //$return_format == "array"
+        //return the array
+        return $return;
+    }
 }
```
kojiromike: `match` should return a value, so doing the assignment within the match expression itself is a bit not-idiomatic. I'd suggest assigning some `$result = match ...` and then checking if `$result === null`.

Or for a different approach, since this whole function is re-mapping anyway…

```
    public function collectCodeTypes($category, $return_format = "array"): string|array
    {
        global $code_types;
        
        $code_remap = [
            'active' => 'active',
            'clinical_term' => 'term',
            'diagnosis' => 'diag',
            'drug' => 'drug',
            'medical_problem' => 'problem',
            'procedure' => 'proc',
        ];
        $cat_code = $code_remap[$category] ?? null;
        
        $return = array_keys(array_filter($code_types, function($ct_arr) use ($cat_code) {
            return ($ct_arr['active'] ?? false) && ($ct_arr[$cat_code] ?? false);
        }

        return $return_format === 'csv' ? csv_like_join($return) : $return;
    }
```

Also consider adding a warning if $return_format isn't an expected value. This could be a good use for an enum…

adunsulag: Thanks for the suggestions.  Incorporated the changes.  Didn't do the enum but did type check return_format.

adunsulag:
I've had this happen twice now with the cryptogen randomly failing in one of the tests.  Everything I see this shouldn't be non-deterministic.  So bizarre.
```
 ✘ Hmac validation failure
   │
   │ Failed asserting that 'test data' is false.
   │
   │ /var/www/localhost/htdocs/openemr/tests/Tests/Unit/Common/Crypto/CryptoGenTest.php:606
   │
```

interface/forms/observation/templates/observation_edit.html.twig
```diff
@@ -0,0 +1,465 @@
...
+                    <div id="questionnaire-results-error" class="questionnaire-results d-none">
+                        <div class="alert alert-danger text-center">
+                            <i class="fa fa-exclamation-triangle fa-2x mb-3"></i>
+                            <h6>{{ 'Error Loading Questionnaire Responses'|xlt }}</h6>
+                            <p class="mb-2">{{ "Error loading questionnaire responses. Please try again."|xlt }}</p>
+                            <small class="text-muted">{{ "Technical details"|xlt }}: <span class="questionnaire-error-details"></span></small>
+                            <div class="mt-3">
+                                {# TODO: move this to observation.js #}
+                                <button type="button" class="btn btn-sm btn-outline-primary" onclick="observationForm.loadQuestionnaireData()">
+                                    <i class="fa fa-refresh"></i> ${xl('Retry')}
```
bradymiller: just making sure above line correct:
```
${xl('Retry')}
```

interface/forms/observation/templates/partials/_sub-observation.html.twig
```diff
@@ -0,0 +1,50 @@
+<div id="{{ nodeId|default("sub-observation-1")|attr }}" class="sub-observation bg-white border rounded p-3 mb-2">
+    <input type="hidden" name="sub_observation_id[]" value="{{ observation.id|default('')|attr }}">
+    <div class="d-flex justify-content-between align-items-center mb-2">
+        <strong>{{ "Sub-observation"|xlt }} <span class="sub-obs-number">{{ observation.id }}</span></strong>
```
bradymiller:
```
{{ observation.id|text }}
```

src/Common/Uuid/UuidRegistry.php
```diff
@@ -47,6 +47,7 @@ class UuidRegistry
         'form_clinical_notes' => ['table_name' => 'form_clinical_notes'],
         'form_encounter' => ['table_name' => 'form_encounter'],
         'form_vitals' => ['table_name' => 'form_vitals'],
+        'form_observation' => ['table_name' => 'form_observation'],
```
bradymiller, line 50: @sjpadgett , I forgot to let you know about making this change for table(s) that you have included uuid columns on.

src/Services/ObservationService.php
```diff
@@ -24,233 +139,586 @@ class ObservationService
...
+                        ,status AS questionnaire_status
+                        ,create_time AS questionnaire_date
+                        ,uuid AS questionnaire_response_uuid
+                    FROM questionnaire_response
+                ) r ON fo.questionnaire_response_id = r.response_id ";
+        $whereFragment = FhirSearchWhereClauseBuilder::build($search, $isAndCondition);
+        $sql .= $whereFragment->getFragment();
+        // TODO: @adunsulag need to implement pagination and sorting
+        $sql .= " ORDER BY fo.date DESC, fo.parent_observation_id, fo.id";
+        $records = QueryUtils::fetchRecords($sql, $whereFragment->getBoundValues());
```
bradymiller, line 334: 2DV

sql/database.sql
```diff
@@ -11897,7 +11928,17 @@ CREATE TABLE `form_observation` (
...
+  `parent_observation_id` bigint(20) DEFAULT NULL COMMENT 'FK to parent observation for sub-observations',
+  `category` varchar(64) DEFAULT NULL COMMENT 'FK to list_options.option_id for observation category (SDOH, Functional, Cognitive, Physical, etc)',
+  `questionnaire_response_id` bigint(21) DEFAULT NULL COMMENT 'FK to questionnaire_response table',
+  PRIMARY KEY (`id`),
+  KEY `idx_form_id` (`form_id`),
+  KEY `idx_parent_observation` (`parent_observation_id`),
+  KEY `idx_category` (`category`),
+  KEY `idx_questionnaire_response` (`questionnaire_response_id`),
+  KEY `idx_pid_encounter` (`pid`, `encounter`),
+  KEY `idx_date` (`date`)
```
bradymiller, line 11941: above 2 keys need to be added in the upgrade script

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -614,5 +614,89 @@ VALUES ('sdoh_instruments', 'hunger_vital_sign', 'Hunger Vital Sign (2-item)', 1
...
+
+-- now we hide the old values
+#IfNotRow3D list_options list_id observation_value_types option_id physical_exam_performed activity 0
+UPDATE `list_options` SET `activity`=0 WHERE `list_id`='Observation_Types' AND `option_id`='physical_exam_performed';
+#EndIf
+
+#IfNotRow3D list_options list_id observation_value_types option_id procedure_diagnostic activity 0
+UPDATE `list_options` SET `activity`=0 WHERE `list_id`='Observation_Types' AND `option_id`='procedure_diagnostic';
+#EndIf
+
```
bradymiller, line 690: missing adding of the `idx_pid_encounter` and `idx_date` keys

interface/forms/observation/table.sql
```diff
@@ -17,6 +19,25 @@ CREATE TABLE IF NOT EXISTS `form_observation` (
...
+  `ob_reason_status` varchar(32) DEFAULT NULL,
+  `ob_reason_code` varchar(64) DEFAULT NULL,
+  `ob_reason_text` text,
+  `ob_documentationof_table` varchar(255) DEFAULT NULL,
+  `ob_documentationof_table_id` bigint(21) DEFAULT NULL,
+  `date_end` DATETIME DEFAULT NULL,
+  `parent_observation_id` bigint(20) DEFAULT NULL COMMENT 'FK to parent observation for sub-observations',
+  `category` varchar(64) DEFAULT NULL COMMENT 'FK to list_options.option_id for observation category (SDOH, Functional, Cognitive, Physical, etc)',
+  `questionnaire_response_id` bigint(21) DEFAULT NULL COMMENT 'FK to questionnaire_response table',
+  PRIMARY KEY (`id`),
```
bradymiller, line 36: missing `idx_form_id` key

bradymiller:
hi @adunsulag , noted several minor issues. When fixed, then consider PR accepted by me. thanks for the awesome PR!

src/Services/CodeTypesService.php
```diff
@@ -429,4 +430,43 @@ public function parseCodesIntoCodeableConcepts($codes)
...
+
+        $code_remap = [
+            'active' => 'active',
+            'clinical_term' => 'term',
+            'diagnosis' => 'diag',
+            'drug' => 'drug',
+            'medical_problem' => 'problem',
+            'procedure' => 'proc',
+        ];
+        $cat_code = $code_remap[$category] ?? null;
```
kojiromike, line 464: Oops, there may be a bug here in that I didn't think through what should happen if `$cat_code` ends up `null`. Kinda feels like that's an error condition.

adunsulag: I'll have it throw an exception if the category is invalid.  The previous behavior was to just return an empty array, but I think we want people to know they've provided an invalid value.

interface/forms/observation/templates/observation_edit.html.twig
adunsulag: Scratching my head why that didn't throw a fatal exception in twig.  No its not correct, so will fix it.

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -614,5 +614,97 @@ VALUES ('sdoh_instruments', 'hunger_vital_sign', 'Hunger Vital Sign (2-item)', 1
 #EndIf
 -- ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 
+-- Observation Form Changes
+
+-- Fix the issue that we don't have a primary key on the form_observation table
```
sjpadgett, line 619: @adunsulag 
Missing SW's function tag for category

`
#IfEyeFormLaserCategoriesNeeded
#EndIf
`

I'll put back in my PR
lol you ping me, I'll ping you.

sjpadgett: I see you moved it to end of your changes so I guess it's okay to leave as is.

PR #8896, opened 2025-09-17
ci(isolated-tests): codecov coverage

### LINKED ISSUE #8895: Integrate Isolated Tests into CodeCov Coverage
CodeCov will automatically merge all coverage reports (IIUC) so we should just be able to add the coverage from different jobs and CodeCov will do the work.

kojiromike:
Fixes #8895

Add code coverage reporting to codecov to isolated-tests.yml

### DIFF
--- a/.github/workflows/isolated-tests.yml
+++ b/.github/workflows/isolated-tests.yml
@@ -77,10 +77,23 @@ jobs:
                            --coverage-html=htmlcov \
                            --testdox
 
+    - name: Check if coverage files exist
+      if: ${{ (success() || failure()) }}
+      id: check-files
+      run: |
+        echo "clover_exists=$(test -f clover.xml && echo true || echo false)" >> $GITHUB_OUTPUT
+        echo "htmlcov_exists=$(test -d ./htmlcov && echo true || echo false)" >> $GITHUB_OUTPUT
     - name: Upload coverage reports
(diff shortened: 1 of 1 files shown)


kojiromike:
I'm curious what Codecov will say when the rest of the tests finish…

bradymiller:
nice! 50% bump in coverage!

kojiromike:
@bradymiller note that from here on until we have really good coverage in general, we may have some confused contributors. The isolated tests finish and report to codecov much earlier than the slow tests. At that point, codecov will probably fail most PRs. Then, it may un-fail the PR when it gets the slow test report, much later.

bradymiller:
That is a bummer. Hopefully way to prevent that.
Will this work (got from claude):
```
- name: Upload coverage to Codecov
  uses: codecov/codecov-action@v3
  with:
    fail_ci_if_error: false
    override_status: success  # Always reports success regardless of coverage
```

bradymiller:
may make sense to create a .codecov.yml (and consider support, if possible, from the easy dev environment).

also looks like can make codecov wait with setting like:
```after_n_builds = 2```

bradymiller:
easy dev environment support may be bad idea though since it takes so long and sounds like would need the secret key.

kojiromike:
Ah `after_n_builds` sounds like a good thing to try.

PR #8899, opened 2025-09-17
Missing files from public/assets and public/themes

YashVerma-code:
Fixes #


kojiromike:
Hi, @YashVerma-code, all of this content is generated when you run `npm run build` after doing `npm install`. Thus, it does not need to be committed to the repository.

PR #8797, opened 2025-08-20
USCDI v3 CCDA

sjpadgett:
fix payer template issues from b1 test tool
add plan name and correct oid's
Is passing IG
- add new SDOH Interventions GUI and CCDA section activity components and Treatment Plan Intervention and planned procedures. Whew, say that three times.
- Add new Related Persons e.g. Granddad, spouse etc. to Demographics. Most likely a new form panel in Contacts.

Fixes #8851 
Fixes #8647 
Fixes #8894
Fixes #8909

PR #8901, opened 2025-09-17
ci(codecov): report after two builds

### LINKED ISSUE #8900: Make codecov wait to report
codecov first gets a report from the fast isolated tests, then much later, a report from the slower docker-based tests. So a PR can get a "fail" from codecov, followed 40 minutes later by an overall "pass". We should make codecov wait to report until all the results are in, to avoid confusion.

kojiromike:
Fixes #8900

Make codecov wait to report until the slower tests have reported, so it doesn't fail a PR just because of one set of tests.

- [x] add codecov.yml
- [x] configure codecov to wait for the second submission.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,3 +2,13 @@ codecov:
   notify:
     after_n_builds: 2
 
+coverage:
+  status:
+    project:
+      default:
+        target: 4%        # Require at least 4% overall project coverage
+        threshold: 0      # No tolerance below the 4% target
+    patch:
+      default:
+        target: 0%        # New code doesn't need any coverage
+        threshold: 0
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,4 @@
+codecov:
+  notify:
+    after_n_builds: 2
+
--- a/.github/workflows/validate-codecov.yml
+++ b/.github/workflows/validate-codecov.yml
@@ -1,0 +1,19 @@
+name: Validate codecov.yml
+
+on:
+  pull_request:
+    paths:
+    - codecov.yml
+  push:
+    paths:
+    - codecov.yml
+
+jobs:
+  validate-codecov:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Validate codecov.yml
+        run: curl -f -X POST --data-binary @codecov.yml https://codecov.io/validate


bradymiller:
per claude, may need this setting: codecov_yml_path: ./.codecov.yml
(i hope claude isn't completely hallucinating on these settings :) )

bradymiller:
actually, it should of found it. :question:

kojiromike:
@bradymiller I think it will only find it once it's in the default branch. Since this is pretty harmless anyway, I'm inclined to merge it and find out.

bradymiller:
maybe try this (put in target/threshold stuff that should never fail for now):
```
codecov:
  after_n_builds: 2

coverage:
  status:
    project:
      default:
        target: 4%        # Require at least 4% overall project coverage
        threshold: 0      # No tolerance below the 4% target
    patch:
      default:
        target: 0%        # New code doesn't need any coverage
        threshold: 0
```

codecov.yml
```diff
@@ -0,0 +1,14 @@
+codecov:
+  notify:
```
bradymiller, line 2: i think need to drop the above notify setting and nest the after_n_builds: 2 there instead (directly below codecov).
(nothing like a PR that requires 45 min. testing period :) )

kojiromike: I added a checker action. Let's see what it says.

sjpadgett:
Can someone give a short justification on purpose here. Are we needing to start submitting test for our PR's?

kojiromike:
@sjpadgett That's part of what we're trying to fix. If we have the configuration the intended way, then as long as the test coverage doesn't drop below 4%, the tests should pass. So there's no specific mandate implied here to test everything.

That said, if a change deletes a bunch of tests (shrinking the numerator), or introduces an enormous amount of new code without tests (increasing the denominator), then it could fail.

codecov.yml
kojiromike: @bradymiller it passed, and I think this is correct. I just suspect this is one configuration option that won't take effect until we merge this PR.

sjpadgett:
Cool so what action would I need to take concerning my recent PR #8797

kojiromike:
> Cool so what action would I need to take concerning my recent PR #8797

Wait until the coverage test job finishes to see what codecov reports after it has all the information.

bradymiller:
@sjpadgett , the settings are now set to not trigger a fail (the amount of code to make it drop from 6% to 4% would be huge). And we are not requiring tests at this time (unless there has been discussion i am not aware of). Short term and mid term goals are getting coverage of current codebase higher. Mid term and long term goals are getting testing on code brought in via PRs (which will be much easier when the testing coverage is much higher in future). I would consider the status informational now and don't think it needs to be completed before bringing in a PR at this time (if this is slowing you down).

codecov.yml
bradymiller: :crossed_fingers:

bradymiller:
To clarify my comment above: Getting testing on PRs at this time would be awesome (but not required). Don't want to write off testing, which @kojiromike and @adunsulag have done great job on.  Hopefully I can figure out how to help out on this front in near future also.

bradymiller:
@kojiromike , bummer that still getting the premature fail (on a PR i just submitted). Agree that the notify line is supposed to be there, so not clear what is going on. I wonder if could be another one of those PR's not being able to use SECRETS things (maybe that setting requires a secret; just guessing). We could just lower the Project coverage goal to basically 0% until we sort it out so don't get the initial fail. If you agree, then I can do it on this PR https://github.com/openemr/openemr/issues/8902 (if we decide to bring it in).

EDIT: I just noted you have a PR going for the codedev so will watch with :popcorn:

bradymiller:
more info on this. on codecov, shows that actually get 5 uploads in total (4 early on and then 5 with the long one). So i bet changing number to 5 will work. Also guessing this will increase with the ongoing PR that will upload more reports.

EDIT: btw this makes sense since there are 4 isolated tests runs that i am guessing have coverage, while there is only 1 standard test run (the long one) with coverage. And 4 + 1 = 5. Hard coding this may be problematic as isolated test runs may change over time and additional uploaded reports may change over time. But seems like most straightforward way at this time while things settle out.

kojiromike:
Yeah that's it. Need to change to 5

bradymiller:
changing to 5 tested well in unrelated PR at https://github.com/openemr/openemr/pull/8903 , so brought it in there.

PR #8903, opened 2025-09-17
fix: crypto encrypt fix to maintain behavior of encryption of empty string when empty string or null

### LINKED ISSUE #8902: fix: crypto encrypt fix to maintain behavior of encryption of empty string when empty string or null
There is now a discrepancy with how null/blank is handled in encrypt function between master and rel-703. In master, it will return blank string. In rel-703, it will return an encrypted string that will decrypt to a blank string. This is currently breaking encryption/decryption feature in FHIR Export context when decrypting the exported Binary files if they are empty.

fyi, here are the pertinent code snippets:

master: https://github.com/openemr/openemr/blob/master/src/Common/Crypto/CryptoGen.php#L189-L194

rel-703:
https://github.com/openemr/openemr/blob/rel-703/src/Common/Crypto/CryptoGen.php#L182-L188

PROPOSE to make the master encryption same behavior ad the prior encryption in rel-703. @kojiromike , I know we have discussed this type of stuff in past, but don't recall if we explicitly decided to change this behavior. If so, I may need reminding :)

bradymiller:
Fixes #8902

### DIFF
--- a/src/Common/Crypto/CryptoGen.php
+++ b/src/Common/Crypto/CryptoGen.php
@@ -186,8 +186,8 @@ class CryptoGen implements CryptoInterface
             throw new CryptoGenException("OpenEMR Error: Random Bytes error - exiting");
         }
 
-        $processedValue = empty($sValue) ? '' : $this->openSSLEncrypt(
-            $sValue,
+        $processedValue = $this->openSSLEncrypt(
+            $sValue ?? '',
             'aes-256-cbc',
             $sSecretKey,
             $iv
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,6 @@
 codecov:
   notify:
-    after_n_builds: 2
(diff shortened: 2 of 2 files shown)


src/Common/Crypto/CryptoGen.php
```diff
@@ -186,8 +186,8 @@ protected function coreEncrypt(?string $sValue, ?string $customPassword = null,
             throw new CryptoGenException("OpenEMR Error: Random Bytes error - exiting");
         }
 
-        $processedValue = empty($sValue) ? '' : $this->openSSLEncrypt(
-            $sValue,
+        $processedValue = $this->openSSLEncrypt(
+            $sValue ?? '',
```
stephenwaite, line 190: squiggly

bradymiller:
Testing well and bringing in.

PR #8916, opened 2025-09-22
feat: Add robust prior authorization unit tracking

trafficpest:
This commit allows the prior authorization module to accurately track units used based on billing data.

It introduces the following key changes:
- Accurate Unit Calculation: The module now calculates units used by summing billing data based on the prior authorization code, CPT codes, and the encounter date range.
- Table Removal: Removed dependencies on the custom `lbt_data` and `patient_status` tables, which were not standard to OpenEMR and caused fatal SQL errors.

Fixes # 
https://community.open-emr.org/t/misc-billing-options-mbo-pull-data-from-advanced-prior-auth/20183/4

Fixes SQL and TypeError issues in Prior Authorization module, adds true unit and code use tracking. Prior, only counted number of times and auth number was used.

No, I did ask google's gemini about the openemr codebase though table structs etc...


interface/modules/custom_modules/oe-module-prior-authorizations/welcome.php
```diff
@@ -13,7 +13,7 @@
 use OpenEMR\Core\Header;
 
 require_once dirname(__FILE__, 4) . "/globals.php";
-require_once __DIR__ . '/vendor/autoload.php';
+require_once dirname(__FILE__) . '/vendor/autoload.php';
```
kojiromike, line 16: Is this change intended?

sjpadgett: globals runs autoloader so remove

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/tags
```diff
@@ -0,0 +1,55 @@
+!_TAG_FILE_SORTED	2	/0=unsorted, 1=sorted, 2=foldcase/
```
kojiromike, line 1: Glad to see someone using ctags, but perhaps don't commit the tags file.

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php
```diff
@@ -14,17 +14,21 @@
...
-    private ?string $end_date = null;
-    private ?string $cpt = null;
-    private ?int $init_units = null;
-    private ?int $remaining_units = null;
 
-    public function storeAuthorizationInfo(): void
+    public function __construct()
+    {
+        //do epic stuff
+    }
```
kojiromike, line 30: Why introduce an empty constructor?

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php
```diff
@@ -14,17 +14,21 @@
...
 class AuthorizationService
 {
+    private $id;
+    private $pid;
+    private $auth_num;
+    private $start_date;
+    private $end_date;
+    private $cpt;
+    private $init_units;
+    private $remaining_units;
```
kojiromike, line 24: Why remove all the type information?

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -8,6 +8,8 @@
  *  All Rights Reserved
  */
 
+
+
 require_once dirname(__FILE__, 6) . "/globals.php";
 require_once dirname(__FILE__, 3) . '/vendor/autoload.php';
```
kojiromike, line 14: can remove

interface/modules/custom_modules/oe-module-prior-authorizations/welcome.php
trafficpest: I run and used the 7.0.3 stable as the base. Wasn't getting type errors then pulled and merged with master update and introduced new errors. So just reverted to 7.0.3 base (lacks types) with my changes and tested, is working on vanilla installation. No error log noticed.

I think that comes in the skeleton module.

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php
trafficpest: I started with 7.0.3 was working then merged and  typing broke it again returned null when expected array etc. function Code needs better error handling.

trafficpest:
If you want to wait for the merge I'll commit a tested merge version with the code on master but this plugin is/was broken in stable 7.0.3 and master with standard openemr db. In current status only works with this merge and 7.0.3 prior-auth base. Current master needs type / error handling in some functions

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php
trafficpest: That for sure is in the skeleton module it was in the 7.0.3 codebase remove it. I'm sure who made the plugin had it then removed it in current master. Also, why aren't these modules separated from core in their own repos?

sjpadgett:
@trafficpest I approved this PR to move forward with integration process.
I believe it could be a needed improvement but I still need to do a code run through.
This module was a private module that was integrated in openemr core because it provided a useful feature to users. So knowing this you need to treat changes as if basically part of core. For one, composer is not needed and besides we'd never allow temporarily lived composer files to be modified directly as if static i.e. remove unless justified which in this case I don't see.
Also @kojiromike is doing the difficult work of bringing our core code up to coding standards which means maintaining or adding annotations and correctly typing properties. So if data type is being violated then need to fix instead of removing such as edge case of boolean is some functions.

I hate reviewing others code but I feel this PR is important for proper reporting so will try to keep it going in the integration process. 
Thanks btw for the contribution. I know it can take a considerable effort.

trafficpest:
OK, I'm going to close this pull request. I'm removing this one because I have another branch ready for the current master. I've tested it(lightly), and it works all error output removed. I'll create a new pull request for it. This version still works for 7.0.3, so I will keep it. The new version only works on the master branch because of a new function call introduced in the src/menu class.

Thank you all for your work on this. This project helps my sister-in-law with her small practice a lot, and this prior authorization module was just what she needed.

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php
kojiromike: > why aren't these modules separated …

IIUC, that is the way things will be going forward for modules. This module probably predates that choice.

interface/modules/custom_modules/oe-module-prior-authorizations/welcome.php
kojiromike: I might create a standalone PR to just remove the extra autoload from this module. Small little PRs like that are quick and easy to merge, and that change is worth doing on its own.

sjpadgett: This composer has no dependencies except to check versions which is no longer needed because it is tied to release versions.

trafficpest:
this became new clean pull

https://github.com/openemr/openemr/pull/8919

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/tags
```diff
@@ -0,0 +1,55 @@
...
+AuthorizationService	/home2/traffid0/public_html/website_db2061ef/emr/interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php	/^class AuthorizationService$/;"	c	language:PHP
```
sjpadgett: remove

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php
```diff
@@ -43,131 +47,144 @@ public function storeAuthorizationInfo(): void
...
+                    WHERE
+                        f.form_name = 'Misc Billing Options'
+                        AND fmbo.prior_auth_number = ?
+                        AND fmbo.pid = ?
+                        AND b.code = ?
+                        AND fe.date BETWEEN ? AND ?";
+                
+      $binds = [$authnum, $pid, $cpt, $start_date, $end_date];
+      $result = sqlQuery($statement, $binds);
+      return (int) ($result['count'] ?? 0);
```
sjpadgett, line 69: I wouldn't case type in returns.

interface/modules/custom_modules/oe-module-prior-authorizations/src/Controller/AuthorizationService.php
sjpadgett: What function code? Is it module code? Can it not be fixed to return what calling routine is looking for or simply change data type if it is wrong but really shouldn't remove all w/o cause.

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/tags
```diff
@@ -0,0 +1,12 @@
+!_TAG_FILE_SORTED	2	/0=unsorted, 1=sorted, 2=foldcase/
+count	/home2/traffid0/public_html/website_db2061ef/emr/interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php	/^                $count = 0;$/;"	v	language:PHP
+data	/home2/traffid0/public_html/website_db2061ef/emr/interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php	/^$data = new AuthorizationService();$/;"	v	language:PHP
+insurance	/home2/traffid0/public_html/website_db2061ef/emr/interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php	/^                    $insurance = AuthorizationService::insuranceName($pid);$/;"	v	language:PHP
```
sjpadgett: Well I supposed tags are great if there is an engineering team to maintain them but IMO none of the admins current or pass would maintain and they would become dead at some point in near future.
Remove.

interface/modules/custom_modules/oe-module-prior-authorizations/public/patient_menu/custom_patient_menu.json
```diff
@@ -1,85 +1,84 @@
...
-        "label": "Dashboard",
-        "menu_id": "dashboard",
-        "target": "main",
-        "on_click": "top.restoreSession()",
-        "url": "interface/patient_file/summary/demographics.php",
-        "pid": "false",
-        "children": [],
-        "requirement": 0
+[	{
+    "label": "Dashboard",
```
sjpadgett, line 2: yep I don't like this! 
First, should emulate standard.json and not remove any standard menus that users are used to having just by turning on a module! Only add items effecting the module.
Second, url needs to be consistent and if not work for you, repair url only for your menu adds. These menu's, standard,  pathing is tried and true.
Third, if you want to change standard openemr permissions(ACL) put in a PR to change the standard.

Some of this isn't on you and should have been address when module was integration.
I don't have time or will to fix this module to be fully integrated and as far as I'm concerned should be removed from production until it is.
So please humor me and see if you can resolve this menu to keep existing features like Assessments required for certification.

interface/modules/custom_modules/oe-module-prior-authorizations/public/index.php
```diff
@@ -16,8 +16,10 @@
...
 
-$pid = $_SESSION['pid'] ?? null;
-function isValid($date, $format = 'Y-m-d'): bool
+$pid = $_SESSION['pid'];
+
+
+function isValid($date, $format = 'Y-m-d')
 {
     $dt = DateTime::createFromFormat($format, $date);
     return $dt && $dt->format($format) === $date;
```
sjpadgett, line 25: If dates are important for calculations then this should require better validation of dates.

sjpadgett:
Are you kidding me! I won't be reviewing the other one.

trafficpest:
I thought it would be easier Ill put it on the other then? I thought you wanted it to be the same as master base

Issue #8922, opened 2025-09-24
Patient Portal Logout - Not Working


infeg-khaleela:
### Describe the bug

The Patient Portal logout functionality is currently not working as expected. When a user attempts to log out, instead of being redirected to the login page, the system displays a "Requested URL not found" error.

Its reproduced in the live demo server also.

1. Login to Patient portal
2. Click on ' Logout' option
3. It will show 'Not Found
The requested URL was not found on this server.'

Here is the commit - for the simple fix for issue
https://github.com/infeg-khaleela/openemr_6_clone/commit/083ac5c39b39a6ef847b1176f4f44c1596b6f81f

kojiromike:
Fixed in #8923

PR #8923, opened 2025-09-24
 Updated Patient Portal logout URL point to portal/logout - for bug 8922

infeg-khaleela:
Fix for - https://github.com/openemr/openemr/issues/8922

Describe the bug
The Patient Portal logout functionality is currently not working as expected. When a user attempts to log out, instead of being redirected to the login page, the system displays a "Requested URL not found" error.

Its reproduced in the live demo server also.

Login to Patient portal
Click on ' Logout' option
It will show 'Not Found
The requested URL was not found on this server.'

### DIFF
--- a/templates/portal/home.html.twig
+++ b/templates/portal/home.html.twig
@@ -662,7 +662,7 @@
                                 <!-- Infeg: Updated logout URL to point to portal/logout.php -->
                                 {% include "portal/partial/_nav_icon.html.twig"
                                     with {
-                                    url:  webroot ~ "/portal/logout.php"
+                                    url:  "./logout.php"
                                     ,localLink: false
                                     , icon: "right-from-bracket"
                                     , navText: "Logout"|xl
--- a/templates/portal/home.html.twig
+++ b/templates/portal/home.html.twig
@@ -659,9 +659,10 @@
                                     , id: "help-go"
                                 } %}
                                 {{ fireEvent(eventNames.dashboardInjectCard) }}
+                                <!-- Infeg: Updated logout URL to point to portal/logout.php -->
                                 {% include "portal/partial/_nav_icon.html.twig"
                                     with {
-                                    url:  webroot ~ "./logout.php"
+                                    url:  webroot ~ "/portal/logout.php"
                                     ,localLink: false
                                     , icon: "right-from-bracket"
                                     , navText: "Logout"|xl


templates/portal/home.html.twig
```diff
@@ -659,9 +659,10 @@
                                     , id: "help-go"
                                 } %}
                                 {{ fireEvent(eventNames.dashboardInjectCard) }}
+                                <!-- Infeg: Updated logout URL to point to portal/logout.php -->
                                 {% include "portal/partial/_nav_icon.html.twig"
                                     with {
-                                    url:  webroot ~ "./logout.php"
+                                    url:  webroot ~ "/portal/logout.php"
```
kojiromike: This code is a little confusing, but I think it's already in the portal context, so a possibly more correct solution would be to remove the `webroot` part:


```suggestion
                                    url:  "./logout.php"
```

This would still use the url specified in the `portal_onsite_two_address` global setting, whereas `webroot ~ ...` may not.

infeg-khaleela: Yes Agreed

kojiromike (approved):

sjpadgett:
This is a bug. Not only is logout broken but so are other cards like Secure Messaging.
This is only broken if openemr is not on a domain alias into root e.g. https://localhost/openemr...
I don't know if I have time to fix all of these urls. This happened when we did a refactor to twig.
Most users run using a domain.
Thanks guys for the fix. Mikes commit is the fix. @infeg-khaleela your change would have broken running via a domain but work for you.
Merging

templates/portal/home.html.twig
```diff
@@ -659,9 +659,10 @@
                                     , id: "help-go"
                                 } %}
                                 {{ fireEvent(eventNames.dashboardInjectCard) }}
+                                <!-- Infeg: Updated logout URL to point to portal/logout.php -->
                                 {% include "portal/partial/_nav_icon.html.twig"
                                     with {
-                                    url:  webroot ~ "./logout.php"
```
bradymiller, line 664: I am pretty sure that replacing `webroot` with `webRoot` would better fix this (now it will break if webRoot is not blank)

EDIT: actually, I think i am wrong since looks like the `./` works by going from directory of the script and then no need to know the webRoot. So, just move along :)

bradymiller:
fyi, forgetting about webroot stuff is a common issue. Since on easy dev environment it is blank. There is a function in easy dev, which will allow to test this (webroot with an actual value):
```
webroot-management:
  cwb, change-webroot-blank          Change webroot to be blank (this is default setting of environment)
  cwo, change-webroot-openemr        Change webroot to be openemr
```

templates/portal/home.html.twig
sjpadgett: No, it won’t. It’s a relative path.

bradymiller: yep, figured it out after I typed in the post :) Figured i must of been missing something and looked it up (didn't even use AI this time; just a simple google search :) )

sjpadgett: Yeah, but you still have to be careful with that notation because it can get you in trouble

bradymiller: Agreed. I think why we generally try to do complete path. Would of the following have worked (since `webRoot` holds the actual javascript webroot variable in that script):
```
url:  webRoot ~ "/portal/logout.php"
```

sjpadgett: You’d think. I think I tested and it didn’t work. I think it has to do with how your virtual host is set up.

sjpadgett: Oh, I see what happened
The bug was Webb root was being combined with the relative path
So you are correct yay, Brady.

Issue #8780, opened 2025-08-14
fix: ignore submodules for composer.json dependabot


bradymiller:
fix: ignore submodules for composer.json dependabot

bradymiller:
#8781 did not fix this issue. Will need to revert and figure out a better way.

bradymiller:
Looks like now working so can close this

PR #8928, opened 2025-09-25
chore: update composer packages

bradymiller:
chore: update composer packages
(note dependabot updates of composer stuff is broken https://github.com/openemr/openemr/issues/8780 )


bradymiller:
testing well and bringing in. This will be last big composer update before next release (unless can get the dependabot fixed and going again for this stuff).

PR #8944, opened 2025-09-27
feat(ServiceLocator): TelemetryService Initial

### LINKED ISSUE #8929: Introduce Service Locator with module overrides
To modernize OpenEMR we want to iterate towards SOLID development principles. One way to make it more extensible is to introduce dependency injection. The module system is useful for introducing new features where there are suitable events. We can interlace the module system with the dependency injection system by introducing an event that happens when a class is instantiated.

However, this is a two or three-step problem. Most classes in OpenEMR are hard-coded (`$x = new Thing();`), hence the need for dependency injection. So I propose we create a ServiceLocator utility with configuration that lets it locate the built-in OpenEMR classes by default. Then we replace the `$x = new Thing();` with `ServiceLocator::get($name);`.

But `ServiceLocator`'s `get` must first dispatch an `EVENT_SERVICE_LOCATE`, so that modules can inject their own implementation in lieu of the built-in class. If `$dispatchedEvent->hasServiceInstance()`, then we use the injected implementation.

Finally, I think this only works if the built-in openemr classes all have interfaces. Then the ServiceLocator can reject injections that don't match the interface, which will prevent a lot of bugs.

kojiromike:
Fixes #8929

This change introduces a service locator approach to dependency injection. Instead of directly instantiating the class, we use the service locator to find registered implementations of the interface. The default implementation is the original class already in OpenEMR, as registered in `config/services.php`. Then we provide an event so that custom modules can inject alternative implementations, without needing to modify the register directly.

For the initial implementation we use the well-tested TelemetryService, which should let us know if this is fundamentally working.

### DIFF
--- a/Documentation/ServiceLocator.md
+++ b/Documentation/ServiceLocator.md
@@ -1,0 +1,562 @@
+# OpenEMR Service Locator Pattern
+
+The Service Locator pattern in OpenEMR allows for dependency injection and service replacement through an event-driven system. This enables modules to provide alternative implementations of core services while maintaining interface compatibility.
+
+## Overview
+
+The `ServiceLocator` class provides a centralized way to:
+- Locate and instantiate services
+- Allow modules to replace core services via events
+- Validate that replacement services implement required interfaces
+- Maintain backward compatibility with existing code
+
+## Core Components
+
+### 1. ServiceLocator Class
+**Location:** `src/Services/ServiceLocator.php`
+
+The main service locator that:
+- Dispatches events when services are requested
+- Validates service interfaces
+- Falls back to default implementations
+- Integrates with Laminas ServiceManager
+
+### 2. ServiceLocatorEvent
+**Location:** `src/Events/Services/ServiceLocatorEvent.php`
+
+Event fired when services are requested, allowing modules to provide alternatives.
+
+### 3. Service Interfaces
+Interfaces that define contracts for services (e.g., `TelemetryServiceInterface`).
+
+## Basic Usage
+
+### Getting a Service
+```php
+use OpenEMR\Services\ServiceLocator;
+use OpenEMR\Telemetry\TelemetryServiceInterface;
+
+// Get a service instance
+$service = ServiceLocator::get(TelemetryServiceInterface::class);
+$service->reportClickEvent($data);
+
+// With context (optional)
+$service = ServiceLocator::get(TelemetryServiceInterface::class, [
+    'environment' => 'test',
+    'user_id' => 123
+]);
+```
+
+## Service Configuration
+
+Core OpenEMR services are configured in `config/services.php`. This file defines:
+- **Interfaces and implementations** for core services
+- **Factory functions** for services that need dependency injection
+- **Aliases** for convenient service access
+
+```php
+// config/services.php
+return [
+    'factories' => [
+        TelemetryServiceInterface::class => function($container) {
+            return new TelemetryService();
+        },
+    ],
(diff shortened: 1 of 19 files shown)


Documentation/ServiceLocator.md
```diff
@@ -0,0 +1,562 @@
...
+**Laminas ServiceManager** excels at runtime service creation:
+```php
+// Easy runtime service location with context
+$service = ServiceLocator::get(TelemetryServiceInterface::class, [
+    'environment' => 'test',
+    'user_id' => 123
+]);
+```
+
+**Symfony DI** is optimized for compile-time resolution and makes runtime service replacement more complex.
```
adunsulag, line 437: This statement is very true...

Documentation/ServiceLocator.md
```diff
@@ -0,0 +1,562 @@
...
+- First listener to call `setServiceInstance()` wins
+- Event propagation stops after first replacement
+- Use listener priorities to control order
+
+## Performance Considerations
+
+- Event dispatching adds minimal overhead (~1ms)
+- Service validation is lightweight
+- Consider caching service instances for heavy objects
+- Use lazy loading patterns for expensive services
```
adunsulag, line 548: I'd add here to avoid hitting the database for any kind of service determination logic, or utilize memory or fs caching of db hits to avoid slowing the event loop down.

src/Events/Services/ServiceLocatorEvent.php
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * @var array Additional context for the service lookup
+     */
+    private $context;
+
+    /**
+     * @var bool Whether a service has been provided by a listener
+     */
+    private $serviceProvided = false;
```
adunsulag, line 33: Would it make sense to just leverage the stopPropagation property in the parent Event class here?  This property has more semantic meaning so probably better, but just a thought.

src/Services/ServiceLocator.php
```diff
@@ -0,0 +1,232 @@
...
+     * First checks if modules provide an alternative via event system
+     *
+     * @param string $serviceName The service name or interface
+     * @param array $context Additional context for service creation
+     * @return mixed The service instance
+     * @throws \Laminas\ServiceManager\Exception\ServiceNotFoundException
+     */
+    public static function get(string $serviceName, array $context = [])
+    {
+        // First, dispatch an event to allow modules to provide alternative services
```
adunsulag, line 94: Is there a way to hit the cache first before hitting the modules?  I haven't dug into the ServiceManager internals so there may be a danger of never hitting the modules altogether by first checking the service manager, just wondering if you thought about this in case any module misbehaves to minimize the damage of a heavy hit on fetching the service instance.

src/Validators/ImmunizationValidator.php
```diff
@@ -21,5 +23,21 @@ class ImmunizationValidator extends BaseValidator
     protected function configureValidator()
     {
         parent::configureValidator();
+
+        // insert validations
+        $this->validator->context(
```
adunsulag, line 28: Did this get leaked in from another PR?  Not sure how its connected to this work.

adunsulag (commented):
Had a couple comments, but overall the code looked good to me.

adunsulag:
Didn't realize this was a draft PR lol, oh well just a few comments on this then.

kojiromike:
superseded by the src/BC/ServiceContainer.php class.

PR #8949, opened 2025-09-28
refactor: rector apply str_starts_with

### LINKED ISSUE #8948: Convert to str_starts_with using rector
OpenEMR variously uses strpos and other methods to determine if a string starts with another string. PHP 8 has a `str_starts_with` function that is more suitable, and rector can convert all the cases for us.

kojiromike:
Fixes #8948

Apply the `str_starts_with` rule automatically using rector.

- [x] refactor: rector str_starts_with
- [x] refactor: rector apply str_starts_with
- [x] run checks

### DIFF
--- a/interface/forms/eye_mag/js/eye_base.php
+++ b/interface/forms/eye_mag/js/eye_base.php
@@ -3475,7 +3475,7 @@ $("body").on("click","[name^='old_canvas']", function() {
                                             }
                                             function startsWith($str, $needle)
                                             {
-                                                return substr($str, 0, strlen($needle)) === $needle;
+                                                return str_starts_with($str, $needle);
                                             }
                                             ?>
                                             submit_form("eye_mag");
--- a/ccr/createCCR.php
+++ b/ccr/createCCR.php
@@ -202,7 +202,7 @@ function gnrtCCR($ccr, $raw = "no", $requested_by = ""): void
                     displayError(xl("ERROR: Unable to Create Zip Archive."));
                     return;
         }
-    } elseif (substr($raw, 0, 4) == "send") {
+    } elseif (str_starts_with($raw, "send")) {
         $recipient = trim(stripslashes(substr($raw, 5)));
         $ccd_out = $ccr->saveXml();
         $result = transmitCCD($pid, $ccd_out, $recipient, $requested_by, "CCR");
@@ -303,7 +303,7 @@ function viewCCD($ccr, $raw = "no", $requested_by = ""): void
         }
     }
 
-    if (substr($raw, 0, 4) == "send") {
+    if (str_starts_with($raw, "send")) {
         $recipient = trim(stripslashes(substr($raw, 5)));
         $ccd_out = $ccd->saveXml();
         $result = transmitCCD($pid, $ccd_out, $recipient, $requested_by);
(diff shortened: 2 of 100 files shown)


sjpadgett:
Dangerous!

interface/forms/CAMOS/save.php
```diff
@@ -25,7 +25,7 @@
     }
 
     foreach ($_POST as $key => $val) {
-        if (substr($key, 0, 3) == 'ch_' and $val = 'on') {
+        if (str_starts_with($key, 'ch_') and $val = 'on') {
```
kojiromike: @sjpadgett any idea if this assignment `($val = 'on')` is intended and correct? Note this isn't a change that was made in this PR.

sjpadgett: No doubt a bug. In this case replacement is okay when fixed.

sjpadgett:
This really scares me but if you review each substitute I'll be okay with it.

interface/forms/eye_mag/js/eye_base.php
```diff
@@ -3475,7 +3475,7 @@ function show_by_setting() {
                                             }
                                             function startsWith($str, $needle)
                                             {
-                                                return substr($str, 0, strlen($needle)) === $needle;
+                                                return str_starts_with($str, $needle);
```
bradymiller, line 3478: so much simpler. nice.

portal/patient/fwk/libs/verysimple/Phreeze/QueryBuilder.php
```diff
@@ -175,7 +175,7 @@ private function RemoveWherePrefix($sql)
         $sql = trim($sql);
 
         // remove if the query is surrounded by parenths
-        while (substr($sql, 0, 1) == "(" && substr($sql, 0, - 1) == ")") {
+        while (str_starts_with($sql, "(") && substr($sql, 0, - 1) == ")") {
```
bradymiller: Your change is fine but the original code doesn't make sense. Doesn't really matter since pretty sure this code is never used. But looks like the second substr is not right (btw, if that is case, then could use str_ends_with() function for the second substr :) ). Asked Claude:

Now I see the complete loop! You're right to question this - there's definitely a bug.

**The Bug:**

`substr($sql, 0, -1)` returns everything **except** the last character. So this condition checks if the string (minus its last character) equals `")"`, which would only be true for the string `"()"` specifically.

**What it should be:**

```php
while (str_starts_with($sql, "(") && substr($sql, -1) == ")") {
    $sql = trim(substr($sql, 1, -1));
}
```

**What the corrected code does:**

Removes outer layers of parentheses from `$sql`:

- `"(((SELECT * FROM users)))"` → `"SELECT * FROM users"`
- `"( (nested) )"` → `"(nested)"`
- `"((a))"` → `"a"`

**Example of the bug in action:**

With the buggy version:
- `"(SELECT * FROM users)"` - The condition checks if `"(SELECT * FROM users"` (without last char) equals `")"` → **false**, loop never runs
- `"()"` - Checks if `"("` equals `")"` → **false**, loop never runs (actually this fails too!)

So the loop essentially never executes as intended. The correct version with `substr($sql, -1)` would properly check if the last character is `")"`.

bradymiller:
manually checked them all and they all looked good. fun, fun :) did note a likely unrelated bug in code that i am pretty sure openemr never uses though.

regarding the other array PR, this will likely take several hours. If we are all in agreement on this one (whether to do before release or after), just let me know and then i'll look through it.

portal/patient/fwk/libs/verysimple/Phreeze/QueryBuilder.php
kojiromike: I don't think even the string '()' minus its last character would be ')'. AI makes weird mistakes.

sjpadgett:
I thought we agreed on chat @bradymiller I was fine with your plan. So I'm okay with doing before release especially since most of what Stephen and I are doing is new code.

bradymiller (approved):

PR #8956, opened 2025-09-29
refactor(php): use array_key_first

### LINKED ISSUE #8955: `ArrayKeyFirstLastRector` - Use array_key_first() and array_key_last()

kojiromike:
Fixes #8955

Update rector to use `array_key_first` instead of `reset` and `key`, which mutates the internal array pointer.

- **chore(rector): set ArrayKeyFirstLastRector**
- **refactor: use array_key_first**


rector.php
```diff
@@ -30,16 +31,21 @@
     ])
     // uncomment to reach your current PHP version
     // ->withPhpSets()
+    ->withCodeQualityLevel(5)
+    ->withDeadCodeLevel(5)
     ->withPhpVersion(PhpVersion::PHP_82)
```
stephenwaite: active support for php8.2 ended 12/31/24, maybe we should bump this?

bradymiller:
Hi,
Would keep this php 8.2 since this is lowest version that next 7.0.4
version will support (I am guessing if you increment it higher, then rector
may change to things that are not supported on php 8.2). Also note php
world is doing something weird this year and not dropping a version:
https://www.php.net/supported-versions.php
(this means we will have more github actions :) )

bradymiller:
Let me correct that statement. They did something weird last year in not
dropping 8.1. which essentially means now supporting 4 versions at a time
rather than 3.

sjpadgett:
Just shoot me!:)

kojiromike:
Rector is actually [supposed to pick the php version from the composer.json.](https://getrector.com/documentation/php-version-features) But for some reason I had trouble with that, so I just hard-coded it in the rector.php for now.

PR #8973, opened 2025-09-30
refactor(php): change switch to match

### LINKED ISSUE #8972: `ChangeSwitchToMatchRector` - Prefer match to switch expressions

kojiromike:
Fixes #8972

Use rector to update switch statements to match expressions.

- [x] chore(rector): set ChangeSwitchToMatchRector
- [x] refactor(php): prefer match to switch

### DIFF
--- a/src/Billing/EDI270.php
+++ b/src/Billing/EDI270.php
@@ -259,16 +259,11 @@ class EDI270
         if ($qual == '102') {
             $DTP[3] = $row['date'];                 // Ins effective Date
         } else {
-            switch ($X12info['x12_dtp03']) {
-                case 'A':
-                    $dtp_date = !empty($row['pc_eventDate']) && $row['pc_eventDate'] > '20010101' ? $row['pc_eventDate'] : date("Ymd");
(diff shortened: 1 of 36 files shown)


kojiromike:
So many of these could be map lookups or something even simpler, but I'm trying not to do manual work here, just yet.

src/Billing/EDI270.php
```diff
@@ -325,17 +320,11 @@ public static function createIEA($row, $X12info, $segTer, $compEleSep)
...
-                return "19";
-                break;
-            case "self":
-            default:
-                return "S";
-        }
+        return match ($relationship) {
+            "spouse" => "01",
+            "child" => "19",
+            default => "S",
```
bradymiller, line 326: this seems odd that missing `self`. does the `self` above act like default since no break?

bradymiller: dang, i now see it. rector just smarter than me. since same as default, no reason to have a match for it :)
moving on

bradymiller (approved):

PR #8995, opened 2025-10-01
refactor(php): use first-class callables instead of arrays

### LINKED ISSUE #8994: `FirstClassCallableRector` - Use first-class callable syntax (PHP 8.1+)

kojiromike:
Fixes #8994

Use first class callables instead of arrays to pass callables.

- [x] chore(rector): set FirstClassCallableRector
- [x] refactor(php): use first-class callables instead of arrays
- [ ] run checks


kojiromike:
#9029 may eliminate some of these, so doing that one first.

bradymiller:
Hi, Got no idea why rector ci is failing, but code looks great. I had no idea what a callable function really did so was cool to learn about them.

custom/code_types.inc.php
```diff
@@ -56,32 +56,38 @@
 use OpenEMR\Events\Codes\ExternalCodesCreatedEvent;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 
+require_once(__DIR__ . "/../library/sql.inc.php");
```
kojiromike: @bradymiller this missing require is why we were seeing the error; however, it's nesty. Below, I had to add guards because of all the database calls we're making at `include` time. Rector can't avoid running those unless we implement an explicit guard.

bradymiller: to make it clear would encircle this code with the if `(!defined('OPENEMR_STATIC_ANALYSIS') || !OPENEMR_STATIC_ANALYSIS) {` stuff or else it will get removed in future when we forget why it is there.

bradymiller: also, is the above even still needed with the guards you placed below?

library/sqlconf.php
```diff
@@ -14,7 +14,8 @@
 use OpenEMR\Core\Kernel;
 use OpenEMR\Events\Core\SqlConfigEvent;
 
-require_once $GLOBALS['OE_SITE_DIR'] . "/sqlconf.php";
+$siteDir = $GLOBALS['OE_SITE_DIR'] ?? __DIR__ . '/../sites/default';
+require_once $siteDir . "/sqlconf.php";
```
bradymiller, line 27: to make it clear would encircle this code with the if `(!defined('OPENEMR_STATIC_ANALYSIS') || !OPENEMR_STATIC_ANALYSIS) {` stuff or else it will get removed in future when we forget why it is there.

Otherwise we will forget why did this. and would then only allow to change to sites/default during static analysis.

kojiromike: @bradymiller hmm? This just sets OE_SITE_DIR to 'default' if it isn't otherwise set. It seemed to me to be a useful change even if static analysis isn't going on.

custom/code_types.inc.php
kojiromike: @bradymiller I think this include is always required for `custom/code_types.inc.php`, but until rector came around we were always doing it implicitly. So IMO it should stay here, and not be guarded.

bradymiller: hi @kojiromike , this library is basically required for all scripts, but is brought in via globals.php for all scripts (ie. assumed to always be available when properly using scripts). And if a script does try to call it without globals.php then it will break horribly.

library/sqlconf.php
bradymiller: hi @kojiromike , If this is happening, then something is wrong and would best to break (ie. if no globals set for some reason). Otherwise run risk of a multisite user accessing/using the wrong database.

custom/code_types.inc.php
kojiromike: What if we explicitly include globals here, then?

bradymiller: hi @kojiromike , Still trying to wrap my brain around this. We generally only have globals.php include on top level scripts. Just doesn't seem like there is a bug here, so trying to figure out least invasive way to placate rector and ensure don't revert it in the future when we forget about it.

kojiromike: @bradymiller Hmm, it's a bit like some of the dependency issues that came up in the User API PR. In that PR, Igor is using some dependencies that we already have in the project, but only transitively: they're not being required explicitly. So we need to change that.

In this case, I think it's a bug for a class to have an implicit `include` that, if that include is missed, causes an error. In the worst case, this could actually lead to a much more serious problem where a file doesn't include authorization requirements, but does include SQL access, and thus exposes information to unauthorized users without an error.

You can see various errors related to this from time to time the the demo server logs, see https://github.com/openemr/openemr/issues/7598#issuecomment-3341942623. It's one of the reasons I think a front controller for the project is important. With a front controller we can eliminate direct access to arbitrary php files and this problem goes away.

I'd rather not have rector or static analysis guards at all, but they're the lesser of evils here. I can, of course, wrap anything you like in the rector guards now that they're here, but I don't really understand why it's harmful to include it without a guard here, since it is a direct dependency.

kojiromike: @bradymiller in any case, I did go ahead and guard it, as this doesn't need to be a blocking issue.

library/sqlconf.php
kojiromike: @bradymiller OK, I see that. I guarded this as well.

PR #9002, opened 2025-10-02
USCDI v5 Specimen Collection Implementation

### LINKED ISSUE #9000: Missing support for tests Specimen collection USCDI v5
Currently we have partial support in the HL7 order however specimen and transport type are populated for test by compendium import and usually proprietary to labs like Quest.
We need to support for In House and selected test from codes and USCDI
A win imo.

sjpadgett:
Adds a per-test Specimens panel to the Procedure Order form to capture multi-specimen details per procedure_order_code line.
Persists to new procedure_specimen (record-id refs; no DB FKs), mirroring existing naming and using current list options.
Captures USCDI v5 / US Core STU 8 fields (type, body site, collected date/time or period, condition, volume, identifiers/accession), enabling standards-compliant FHIR R4 Specimen resources and proper Observation.specimen references (incl. _include).
Keeps order-level specimen fields as convenient defaults → reduces duplication, improves auditability, and supports 24-hr collections.
UX matches the existing Reason panel for minimal training; future-proofs for LIS/eReq workflows without schema churn.

Fixes #9000

Overview
Adds comprehensive specimen collection tracking to OpenEMR's procedure order system to achieve USCDI v5 / US Core 8.0 compliance for laboratory observations.
Changes
Database Schema
New columns in procedure_order table:

date_collected_end - Collection period end time
specimen_type - Type of specimen (Blood, Urine, etc.)
specimen_location - Body site/collection location
specimen_volume - Volume in mL (DECIMAL)
specimen_condition - Acceptability status (acceptable, hemolyzed, etc.)

List options populated for:

Specimen types (15 common types: blood, serum, urine, etc.)
Collection sites (16 body locations: left arm, throat, finger, etc.)
Specimen conditions (12 HL7 v2 Table 0493 codes)

Backend (PHP)
common.php - Procedure Order Form:

Updated save logic to persist 5 new specimen fields
Added specimen collection form section with dropdowns and datetime pickers
Collection start uses existing date_collected field
Collection end uses new date_collected_end field
Validates collection period (end must be after start)

FhirObservationLaboratoryService.php:

Maps specimen data from procedure_order to FHIR observations
Implements 4 USCDI v5 required extensions:

specimen-source-site - Body location (CodeableConcept)
specimen-collection-period - Start/end times (Period)
specimen-volume - Volume with UCUM units (Quantity)
specimen-condition - HL7 v2 acceptability codes (CodeableConcept)

Auto-maps condition text to HL7 codes (e.g., "hemolyzed" → "HEM")
Supports specimen identifiers via extensions

Frontend (JavaScript)
Fixed lab_id_changed() race condition:

Uses setTimeout() to ensure DOM updates before popup
Properly clears existing procedures when lab changes
Auto-opens procedure selector after adding new line

USCDI v5 Compliance
✅ Specimen Type (required)
✅ Specimen Source Site (required if applicable)
✅ Specimen Collection Period (required)
✅ Specimen Volume (required if applicable)
✅ Specimen Condition Acceptability (required)
✅ Specimen Identifier (supported)


interface/forms/procedure_order/common.php
```diff
@@ -1311,17 +1403,7 @@ function doWait(e) {
                             <div class="col-md-2">
                                 <?php generate_form_field(array('data_type' => 10, 'field_id' => 'collector_id'), $row['collector_id'] ?? ''); ?>
                             </div>
-                            <label for='form_order_abn' class="col-form-label col-md-2"><?php echo xlt('ABN Status'); ?></label>
```
adunsulag, line 1314: Is ABN no longer needed?

interface/forms/procedure_order/common.php
```diff
@@ -1338,17 +1422,6 @@ class="col-form-label col-md-2"><?php echo xlt('Status'); ?></label>
                                     'list_id' => 'ord_status'), $row['order_status'] ?? '');
                                 ?>
                             </div>
-                            <label for="form_billing_type"
```
adunsulag, line 1341: Did AI remove this or is billing no longer needed either?

interface/forms/procedure_order/common.php
```diff
@@ -840,54 +860,42 @@ function addProcLine(flag = false) {
                     console.error("Failed to find node with selector ", selector);
                 }
             };
-            // once our node is in the DOM, we need to add event listeners to it.
+
             nullableFunction('.itemTransport', 'click', function (event) {
-                // we have to bind to our lineCount at the time of instantiation in case addProcLine is called again
```
adunsulag, line 845: It looks like a lot of context comments got removed.  I noticed that this is a common problem with AI stuff, not sure if you intentionally did that or if it was the AI system that did it, just wanted to make sure you intentionally removed it.

sjpadgett:
I had to rearrange the form for new fields. I didn't feel like fooling with comments today
image

interface/forms/procedure_order/common.php
sjpadgett: I asked Claude for cleanup after I moved fields around. It refactored JS but it made sense what was done so I kept changes but this is the difference I found between ChatGPT and Claude, ChatGPT would not have refactored in most occasions.
Anyway we need to keep comments so thanks.

interface/forms/procedure_order/common.php
sjpadgett: No I moved these to make room for needed specimen fields.

sjpadgett:
image

sjpadgett:
Latest may seem like overkill for specimen where we could get away with having just one order and one specimen for certification however that would not serve us well in the future
While reviewing some of my Quest and LabCorp classes and tests, I realized that these labs require ability to support multi specimen if order contains multi test/procedures. 
This follows SN's reason widget pattern.
So logic dictates that there is a good chance that DORN and these labs may require as well.

This also allows our FHIR resources requiring specimen to be more versatile and future proof, at least I hope.
This follows SN's reason widget pattern.

Sorry @bradymiller I guess your database nightmares have returned! lol

sjpadgett:
Still going to condense some more
image
image

image

sql/database.sql
```diff
@@ -9718,6 +9718,46 @@ CREATE TABLE `procedure_result` (
 
 -- -----------------------------------------------------------------------------------
 
+--
+-- Table structure for table `procedure_specimen`
+--
+
+DROP TABLE IF EXISTS `procedure_specimen`;
+CREATE TABLE `procedure_specimen` (
```
adunsulag, line 9726: Do we need to do a data migration from procedure_order now that specimens are tracked in this table?  I'm assuming yes.  I love how comprehensive this table is which is great, just wondering what your thoughts are on existing data.

sjpadgett: Since in the past specimen was only handled by when and by whom specimen where collected so nothing to migrate.
Custom lab integrations for LabCorp or Quest create their own specimen table and manage in their own way.
Over the years I have updated Rod's original lab project with many of the requirements learned from those integrations plus now with specimen, we are closer to being able to stand up either of those labs in our production releases.
Many may be unaware that we also have capability to auto assign storage, transportation and containers/color coding for the test to be performed. That came from a Quest integration module I wrote.
But now with DORN all for this has come together in that OpenEMR has a terrific capability.

Oh yeah, to answer your question, lol. No I don't think we need to touch legacy data but I will keep in mind going forward in testing. i.e. will load a previous DB with legacy labs that have been resulted. The new table can stand on its own with only new orders that have specimens will have dependency
BTW, I'm looking to add temperature to specimen if I can find if required in FHIR.

sjpadgett:
@bradymiller Looks like last PR to touch DB didn't advance version. Should I pass on 519 which would have been value with my commit bumping to 520 or should I keep sequential by using 519?

sjpadgett:
Usually when all the unit tests fail it is because of database error i've found.
However I am able to create the database from a direct import but my MariaDB is an old version so perhaps something in DB is version sensitive.
Loading up easy dev now so if DB loads with it then I'm not sure what is stopping mysql from running.

sjpadgett:
Can anyone make sense of this
`mkdir: can't create directory '/tmp/php-file-cache': File exists
2025-10-08 00:53:06 ERROR IN OPENEMR INSTALL: Unable to execute SQL: CREATE TABLE ccda_field_mapping (  id int(11) NOT NULL AUTO_INCREMENT,  table_id int(11) DEFAULT NULL,  ccda_field varchar(100) DEFAULT NULL,  PRIMARY KEY (id)) ENGINE=InnoDB AUTO_INCREMENT=1 due to: Table &#039;ccda_field_mapping&#039; already exists
2025-10-08 00:53:06 PHP Fatal error:  Uncaught Exception: ERROR: unable to execute SQL: 'CREATE TABLE ccda_field_mapping (  id int(11) NOT NULL AUTO_INCREMENT,  table_id int(11) DEFAULT NULL,  ccda_field varchar(100) DEFAULT NULL,  PRIMARY KEY (id)) ENGINE=InnoDB AUTO_INCREMENT=1' due to: Table 'ccda_field_mapping' already exists
2025-10-08 00:53:06  in /var/www/localhost/htdocs/auto_configure.php:62`

sjpadgett:
It looks like newer versions on sequel don't accept certain comments inserted in list inserts. Need to be block comments /* */ not -- this is comment. I need to upgrade my dev system again!

sjpadgett:
Finally was able to fix phpstan without having to change rule. Took a minute.:)
Will merge now because of so many changes hanging around and initial tests are testing well. I still need/want to do more extensive testing but I want to make some progress on FHIR resources dependent on this PR. So there it is.

All in all I'm happy I spent the time on this important form and am pleased with how it turned out.
I did add an auto collapse for details when working on order items.
I will be moving the javascript out on form to its own file the next time I do PR concerning this procedure_order form.

image

PR #9016, opened 2025-10-02
refactor(php): Change null to strict string defined function call args

### LINKED ISSUE #9015: `NullToStrictStringFuncCallArgRector` - Change null to strict string defined function call args

kojiromike:
Fixes #9015

Change null to strict string defined function call args

- [x] chore(rector): set NullToStrictStringFuncCallArgRector
- [x] refactor(php): Change null to strict string defined function call args
- [ ] run checks


kojiromike:
This is an enormous change, but one way to see it is to grep out all the cases where a line was added with `(string)`:

```
$ git show | grep '^\+ ' | grep -vF '(string)'
```

This should be empty. All this PR is doing is ensuring that every build-in function that requires a string argument is being passed a string, and not null. Adding `(string)` converts `null` to `''`, which is how PHP 8.1+ will behave implicitly, anyway, but we can then start to work on replacing these explicit casts with explicit type declarations instead.

PR #9026, opened 2025-10-02
 Add global front controller with security hardening

### LINKED ISSUE #8964: Introduce global front controller for openemr
OpenEMR lacks an overall front controller, so most deployments expect to be able to reach arbitrary php files by their code path directly. This is limiting for various purposes:

1. Security: The necessity of granting access to the entire webroot filesystem means providing clients with specific information about paths – even paths you might rather they not interact with directly, such as documents, temporary files and caches. A front controller would make such paths un-addressable. It would also clearly route between things that can be viewed anonymously and things that require login. Most things in OpenEMR require login.
2. Configuration: Separating the static content from the application content is made more complex by the fact that the web server needs detailed information about what files exist, and where. It should be straightforward to serve openemr with a web server forwarding php requests to php-fpm on another server. But right now, it's necessary to carefully configure a large number of individual paths expressly in the web server, and that configuration will need to be kept explicitly in sync with future code changes. A front controller would eliminate the need for the web server to know what all the paths in the codebase are.
3. Maintainability: Most standalone php paths in the codebase require globals pretty early. There are certainly php files in openemr that are not intended to be requested directly. The [demo site error log](https://demo.openemr.io/log/logPhp.txt) has errors that indicate that they are being requested, and error because they're missing some globals. (This could also be a security concern.) A front controller would ensure that only pages that should be reachable can be addressed, and that every reachable page has a consistent set of includes/headers/whatever.
4. Extension: A front controller would provide a much more consistent place for some very early and very late events, allowing potentially healthier places to introduce global error handling, logging, telemetry and instrumentation, and any other extension that makes sense.

Consider carefully what Symfony and other php frameworks do for front controllers. Then, in the OpenEMR codebase, review the index.php and other global, central files, and the .htaccess files in the rest of the repo. Consider these additional paths:

### Deny All (subpaths should always return 404)

- `/portal/patient/fwk/libs`
- `/sites/[^/]*/documents`
    
### Deny Sometimes

These paths are used for administrative purposes and are worth special handling – treat them like admin-only access.

`^/(admin|setup|rector|phpstan_panther_alias|acl_setup|acl_upgrade|sl_convert|sql_upgrade|gacl/setup|ippf_upgrade|sql_patch)\.php`

### Inner Front Controllers

These sub paths have controllers at index.php for their subdirectories, although some of them _still_ allow access to php files as well. See the .htaccess files.

- /apis/
- /gacl/admin/
- /interface/esign/
- /interface/main/calendar/
- /interface/modules/custom_modules/oe-module-claimrev-connect/public/
- /interface/modules/custom_modules/oe-module-comlink-telehealth/public/
- /interface/modules/custom_modules/oe-module-dorn/public/
- /interface/modules/custom_modules/oe-module-ehi-exporter/public/
- /interface/modules/custom_modules/oe-module-faxsms/
- /interface/modules/custom_modules/oe-module-prior-authorizations/
- /interface/modules/custom_modules/oe-module-prior-authorizations/public/
- /interface/modules/zend_modules/public/
- /interface/super/rules/
- /interface/therapy_groups/
- /oauth2/
- /portal/
- /portal/patient/

Also, please note that if the requested url lacks a trailing slash it should redirect (301) to the same url with the trailing slash.

### Likely Bugs

- /ccdaservice/packages/oe-cqm-service/index.php (except for comments, this file is empty. Should return 404 or be deleted outright.)
- /contrib/util/dupecheck/index.php (deprecated, should return 404)

### Static Assets

While it's not ideal for a php application to serve static assets, frameworks commonly provide various ways to handle them to make development easy. Consider what Symfony and other frameworks do to enable reasonable static asset management through the SDLC.

Static assets in OpenEMR are usually, but not exclusively in public/. Most are generated via `npm ci && npm run build`, and have these extensions:

`.(css|css\.map|js|js\.map|mjs|scss|png|jpg|jpeg|gif|bmp|ico|svg|eot|otf|ttf|woff|woff2|html|htm)`

`/swagger/*` should also be included, including its one .yaml file.

coding-sunshine:
Fixes #8964

Implements centralized routing architecture with multi-layer security:

Implements centralized routing architecture with multi-layer security:

  - Add home.php as main front controller for centralized routing
  - Add .htaccess with Apache rewrite rules for request routing
  - Add security-check.php as auto-prepended .inc.php blocker
  - Block direct access to .inc.php files (403 Forbidden)
  - Deny access to sensitive paths (/portal/patient/fwk/libs/, /sites/*/documents/)
  - Preserve existing inner front controllers (/apis/, /portal/, /oauth2/, /interface/*)
  - Add OPENEMR_ENABLE_FRONT_CONTROLLER feature flag (disabled by default)
  - Add extension hooks for early/late custom logic injection
  - Add multisite support with automatic site detection
  - Add path traversal prevention with realpath() validation
  - Add admin path detection for telemetry integration
  - Add comprehensive test suite (PHPUnit + shell scripts)
  - Add README-FRONT-CONTROLLER.md with deployment guide

  Security improvements:
  - Three-layer .inc.php blocking (.htaccess → security-check.php → home.php)
  - Request validation before routing
  - Error logging for security events
  - Debug mode via OPENEMR_FC_LOG_LEVEL

  Backward compatibility:
  - Feature disabled by default (opt-in via environment variable)
  - No breaking changes to existing routing
  - All existing front controllers preserved
  - Static assets passed through unchanged

  Files added:
  - .htaccess (Apache routing rules)
  - home.php (main front controller)
  - security-check.php (PHP prepend security layer)
  - README-FRONT-CONTROLLER.md (documentation)
  - custom/.gitkeep (extension directory)
  - tests/FrontController/SecurityTest.php
  - tests/FrontController/CompatibilityTest.php
  - tests/scripts/test_security.sh
  - tests/scripts/test_compatibility.sh
  - tests/scripts/test_performance.sh

  Files modified:
  - .env.example (feature flag documentation)


tests/FrontController/CompatibilityTest.php
kojiromike, line 1: Can these tests be integrated with one of the existing test setups? These tests are not being run by the CI system. If you go to https://app.codecov.io/gh/openemr/openemr/commit/adcfdafcc26e34ff20bf1a1b491c3f23ee725004 you can see the coverage reporting detects no coverage for the new code.

tests/FrontController/CompatibilityTest.php
```diff
@@ -0,0 +1,365 @@
...
+    public static function setUpBeforeClass(): void
+    {
+        self::$baseUrl = getenv('OPENEMR_TEST_URL') ?: 'http://localhost/openemr';
+    }
+
+    public function testIndexPhpAccessible(): void
+    {
+        $url = self::$baseUrl . '/index.php';
+
+        $ch = curl_init($url);
```
kojiromike: Consider the Guzzle HTTP client that's already a dependency.

kojiromike: There's also an APITestClient, but it may be specific to JSON APIs.

tests/FrontController/CompatibilityTest.php
```diff
@@ -0,0 +1,365 @@
...
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+        curl_exec($ch);
+        $httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+        curl_close($ch);
+
+        $this->assertContains(
+            $httpCode,
+            [200, 302],
+            'index.php should be accessible'
```
kojiromike: Why would we accept either a 302 _or_ a 200? Shouldn't we be specific in this case?

tests/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,262 @@
...
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m' # No Color
+
+# Create report directory
+mkdir -p "$REPORT_DIR"
+
+echo "Compatibility Test Report" | tee "$REPORT_FILE"
```
kojiromike: Consider combining all the `echo` statements into a single compound command to generate the file.

```
{
    echo "..."
    echo "...."
    echo "....."
} | tee "$REPORT_FILE"
```

tests/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,262 @@
...
+# Function to run a test
+run_test() {
+    local test_name="$1"
+    local url="$2"
+    local expected_codes="$3"
+    local description="$4"
+
+    TOTAL_TESTS=$((TOTAL_TESTS + 1))
+
+    echo -n "Testing: $test_name... " | tee -a "$REPORT_FILE"
```
kojiromike: Prefer `printf` to `echo -n`

```suggestion
    printf 'Testing: %s…' "$test_name" | tee -a "$REPORT_FILE"
```

tests/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,262 @@
...
+
+    # Make HTTP request and get status code
+    http_code=$(curl -s -o /dev/null -w "%{http_code}" -L "$url")
+
+    # Check if code matches any of the expected codes
+    if [[ "$expected_codes" == *"$http_code"* ]]; then
```
kojiromike: Avoid `==` in shell scripts except in arithmetic context.

```suggestion
    if [[ "$expected_codes" = *"$http_code"* ]]; then
```

tests/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,262 @@
...
+        echo -e "${GREEN}PASS${NC} (HTTP $http_code)" | tee -a "$REPORT_FILE"
+        PASSED_TESTS=$((PASSED_TESTS + 1))
+        echo "  ✓ $description" | tee -a "$REPORT_FILE"
+    else
+        echo -e "${RED}FAIL${NC} (Expected [$expected_codes], got $http_code)" | tee -a "$REPORT_FILE"
+        FAILED_TESTS=$((FAILED_TESTS + 1))
+        echo "  ✗ $description" | tee -a "$REPORT_FILE"
+        echo "  URL: $url" | tee -a "$REPORT_FILE"
+    fi
+    echo "" | tee -a "$REPORT_FILE"
```
kojiromike:
```suggestion
    echo | tee -a "$REPORT_FILE"
```

tests/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,262 @@
...
+
+if [ "$http_code" != "404" ] && [ "$http_code" != "403" ]; then
+    echo -e "${GREEN}PASS${NC} (HTTP $http_code)" | tee -a "$REPORT_FILE"
+    PASSED_TESTS=$((PASSED_TESTS + 1))
+    echo "  ✓ POST requests are processed correctly" | tee -a "$REPORT_FILE"
+else
+    echo -e "${RED}FAIL${NC} (HTTP $http_code)" | tee -a "$REPORT_FILE"
+    FAILED_TESTS=$((FAILED_TESTS + 1))
+    echo "  ✗ POST requests are being blocked" | tee -a "$REPORT_FILE"
+fi
```
kojiromike: - Please consistently use `[[` for testing and don't go back and forth with `[`.
- This would probably be easier to read by swapping the conditions.
- Prefer `printf` to `echo -e`.

tests/scripts/test_performance.sh
```diff
@@ -0,0 +1,252 @@
...
+set -e
+
+# Configuration
+BASE_URL="${1:-http://localhost/openemr}"
+ITERATIONS="${2:-100}"
+REPORT_DIR="$(dirname "$0")/../../reports"
+REPORT_FILE="$REPORT_DIR/performance-test-report-$(date +%Y%m%d-%H%M%S).txt"
+
+# Colors for output
+RED='\033[0;31m'
```
kojiromike: It's safer to use `tput` to generate escape sequences, because it will use the termcap table to identify if the terminal can support the given escapes.

tests/scripts/test_performance.sh
```diff
@@ -0,0 +1,252 @@
...
+echo "Date: $(date)" | tee -a "$REPORT_FILE"
+echo "URL: $BASE_URL" | tee -a "$REPORT_FILE"
+echo "Iterations: $ITERATIONS" | tee -a "$REPORT_FILE"
+echo "" | tee -a "$REPORT_FILE"
+
+# Check if Apache Bench (ab) is available
+if ! command -v ab &> /dev/null; then
+    echo -e "${YELLOW}WARNING: Apache Bench (ab) not found. Installing via brew...${NC}" | tee -a "$REPORT_FILE"
+    if command -v brew &> /dev/null; then
+        brew install httpd
```
kojiromike: We shouldn't automatically install things. Just exit with an error if the desired command isn't found.

tests/scripts/test_performance.sh
kojiromike, line 1: The performance testing is a good idea.

This might be a good case for a test we don't run automatically, so no need to integrate the performance test with the existing testing framework for now.

tests/scripts/test_performance.sh
```diff
@@ -0,0 +1,252 @@
...
+
+    # Extract key metrics
+    requests_per_sec=$(echo "$ab_output" | grep "Requests per second" | awk '{print $4}')
+    time_per_request=$(echo "$ab_output" | grep "Time per request.*mean\)" | head -1 | awk '{print $4}')
+    failed_requests=$(echo "$ab_output" | grep "Failed requests" | awk '{print $3}')
+
+    # Extract percentiles
+    p50=$(echo "$ab_output" | grep "50%" | awk '{print $2}')
+    p95=$(echo "$ab_output" | grep "95%" | awk '{print $2}')
+    p99=$(echo "$ab_output" | grep "99%" | awk '{print $2}')
```
kojiromike: Use herestrings instead of `echo | cmd`. Also, use awk to do the `grep` part, too. `awk /50%/{print $2}`.

.htaccess
kojiromike, line 1: The .htaccess may not interact well with home.php in the disabled condition.

home.php
```diff
@@ -0,0 +1,129 @@
...
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    OpenEMR Contributors
+ * @copyright Copyright (c) 2025 OpenEMR Contributors
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+// CLI not supported
+if (php_sapi_name() === 'cli') {
+    http_response_code(404);
```
kojiromike: Not sure what good it will do to return a 404 to the CLI.

home.php
```diff
@@ -0,0 +1,129 @@
...
+    exit('Front controller is web-only. Use direct file access for CLI scripts.');
```
kojiromike: It may be useful to let the cli sapi continue, for testing purposes etc.

home.php
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Load .env variables
+if (file_exists(__DIR__ . '/.env')) {
```
kojiromike: There is a Dotenv library in composer.json.

home.php
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Check feature flag
+if (!getenv('OPENEMR_ENABLE_FRONT_CONTROLLER') || getenv('OPENEMR_ENABLE_FRONT_CONTROLLER') !== '1') {
+    http_response_code(404);
+    exit('Front controller disabled');
+}
+
+// Early extension hook
+if (file_exists(__DIR__ . '/custom/front_controller_early.php')) {
```
kojiromike: Extension should be done using the module/event system in openemr. We should expect people to be able to extend openemr by composer installing a module, not by dropping code manually into the `./custom` directory.

See https://github.com/adunsulag/oe-module-custom-skeleton

home.php
```diff
@@ -0,0 +1,129 @@
...
+    require __DIR__ . '/custom/front_controller_early.php';
+}
+
+// Multisite selection
+$site_id = '';
+if (!empty($_GET['site'])) {
+    $site_id = $_GET['site'];
+} elseif (is_dir("sites/" . ($_SERVER['HTTP_HOST'] ?? 'default'))) {
+    $site_id = ($_SERVER['HTTP_HOST'] ?? 'default');
```
kojiromike: I'm not sure there's precedent for converting `$_SERVER['HTTP_HOST']` into the site name.

home.php
```diff
@@ -0,0 +1,129 @@
...
+$_GET['site'] = $site_id;
+
+// Extract route
+$route = $_GET['_ROUTE'] ?? '';
+if (($pos = strpos($route, '?')) !== false) {
+    $route = substr($route, 0, $pos);
+}
+
+// Remove trailing slash (redirect)
+if ($route !== '' && substr($route, -1) === '/') {
```
kojiromike: This is backwards. We always want there to be a trailing slash.

home.php
```diff
@@ -0,0 +1,129 @@
...
+
+// Only route .php files
+if (pathinfo($targetFile, PATHINFO_EXTENSION) !== 'php') {
+    http_response_code(404);
+    error_log("OpenEMR Front Controller: Non-PHP file: $route");
+    exit('Not Found');
+}
+
+// Late extension hook
+if (file_exists(__DIR__ . '/custom/front_controller_late.php')) {
```
kojiromike: The late extension should come after the requested content.

kojiromike (changes requested):
Good first pass. There are some things that need to be corrected, so I will request changes here.

One thing is that the routing is all hard-coded into home.php right now. But might there be a way to define the routing in a more extensible way?

I left specific comments where I thought it would be useful. It might also be useful to consider some of the ideas from https://symfony.com/doc/current/create_framework/front_controller.html because this project is heading in a Symfony-friendly direction.

One thing I found interesting from that doc was the idea of putting the front controller in a subdirectory and then setting the webroot to that subdirectory. So configured, all the content would only be reachable by the front controller via a controlled path traversal.

Another way to think of it might be, how could we build a front controller that would work if openemr were a phar?

sjpadgett:
I was moving towards moving portal to a module.
Still I have an old framework I worked on years ago that I used for the controller/model/view for two crud like features, onsite documents and Patient.
You'd be doing a great service for me and openemr if we could start to get rid of that framework(Phreeze).
lol should take about an hour or two!

kojiromike:
> I was moving towards moving portal to a module. Still I have an old framework I worked on years ago that I used for the controller/model/view for two crud like features, onsite documents and Patient. You'd be doing a great service for me and openemr if we could start to get rid of that framework(Phreeze). lol should take about an hour or two!

@sjpadgett I'm confused – did you mean to make this comment on this PR? I don't see a lot of interaction with Phreeze here.

tests/scripts/test_performance.sh
coding-sunshine: This is already the case - the performance test script (test_performance.sh) is a standalone script that can be run manually and is not integrated into the automated CI/CD pipeline or PHPUnit test suite.

.htaccess
coding-sunshine: @kojiromike agree. 

 The approach was: 
  1. .htaccess always routes PHP requests to home.php
  2. home.php checks if the feature is enabled
  3. If disabled, it returns 404, which causes Apache to fall back to the original file
  
 ## Now

 The `.htaccess` routing is designed to work in both states:

  ```apache
  # Physical files pass through directly (never hit home.php)
  RewriteCond %{REQUEST_FILENAME} -f
  RewriteRule ^ - [L]

  # Only non-existent files route to home.php
  RewriteCond %{REQUEST_URI} \.php$
  RewriteRule ^(.*)$ home.php?_ROUTE=$1 [QSA,L]

  When Front Controller is DISABLED:

  -  Existing PHP files (interface/login/login.php, etc.): Served directly by Apache via the -f check
  -  Virtual routes: Pass through home.php with transparent context preservation
  -  .inc.php files: Still blocked for security

  When Front Controller is ENABLED:

  -  All routes: Go through front controller with full security validation
  -  Security checks: Router validates paths, blocks forbidden files

  The Transparent Pass-Through

  For the edge case where disabled mode receives routed requests, home.php now acts as a transparent proxy:

  // Preserves correct execution context
  $_SERVER['SCRIPT_FILENAME'] = $targetFile;
  $_SERVER['SCRIPT_NAME'] = '/' . $route;
  $_SERVER['PHP_SELF'] = '/' . $route;
  chdir(dirname($targetFile));

  // Security maintained even when disabled
  - Directory traversal prevention
  - .inc.php blocking
  - File existence validation

  require $targetFile;  // __FILE__ and __DIR__ work correctly in target

  Tested and verified: __FILE__, __DIR__, relative paths, and $_SERVER variables all work correctly for the target file.

  The .htaccess + home.php combination now handles both enabled and disabled states correctly.

home.php
coding-sunshine: Agreed! Updated this to use OpenEMR's module/event system instead of the custom hook file approach.

  The code now dispatches proper events:
  - `front_controller.early` - for early initialization before routing
  - `front_controller.late` - for post-processing after content is loaded

  Modules can listen to these events following the pattern in https://github.com/adunsulag/oe-module-custom-skeleton

home.php
coding-sunshine: You're right. I've removed the HTTP_HOST-based site detection.

The updated code now follows OpenEMR's standard approach:
  - Uses explicit `?site=` parameter when provided
  - Falls back to 'default' site when no parameter is given
  - Avoids HTTP_HOST entirely

src/OpenCoreEMR/FrontController/RouteConfig.php
```diff
@@ -0,0 +1,221 @@
...
+     * @return bool True if forbidden
+     */
+    public function isForbidden(string $route): bool
+    {
+        foreach ($this->forbiddenPatterns as $pattern) {
+            if (preg_match($pattern, $route)) {
+                return true;
+            }
+        }
+        return false;
```
kojiromike:
```suggestion
        return array_any(
            $this->forbiddenPatterns,
            fn(string: $pattern): bool => (bool) preg_match($pattern, $route)
        );
```

src/OpenCoreEMR/FrontController/RouteConfig.php
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Check if route matches any admin-only pattern.
+     *
+     * @param string $route The route to check
+     * @return bool True if admin-only
+     */
+    public function requiresAdmin(string $route): bool
+    {
+        foreach ($this->adminPatterns as $pattern) {
```
kojiromike: use array_any

src/OpenCoreEMR/FrontController/RouteConfig.php
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Check if route is a deprecated path.
+     *
+     * @param string $route The route to check
+     * @return bool True if deprecated
+     */
+    public function isDeprecated(string $route): bool
+    {
+        foreach ($this->deprecatedPaths as $pattern) {
```
kojiromike: You can probably create a single private function that determines isDeprecated, requiresAdmin, and isForbidden using `array_any` and another parameter to determine which paths to reference.

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,141 @@
...
+{
+    private string $baseDir;
+    private string $route;
+    private string $siteId;
+    private RouteConfig $config;
+
+    public function __construct(string $baseDir, ?RouteConfig $config = null)
+    {
+        $this->baseDir = realpath($baseDir);
+        $this->config = $config ?? new RouteConfig();
```
kojiromike: To enable DI, the `RouteConfig` type should be an interface.

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,141 @@
...
+     * Extract and clean route from request
+     */
+    public function extractRoute(): string
+    {
+        $route = $_GET['_ROUTE'] ?? '';
+
+        // Remove query string
+        if (($pos = strpos($route, '?')) !== false) {
+            $route = substr($route, 0, $pos);
+        }
```
kojiromike:
```suggestion
        $route = strtok($route, '?');
```

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,141 @@
...
+     * Determine multisite ID
+     */
+    public function determineSiteId(): string
+    {
+        // explicit site parameter or session-based detection.
+        if (!empty($_GET['site'])) {
+            $this->siteId = $_GET['site'];
+        } else {
+            $this->siteId = 'default';
+        }
```
kojiromike: use the null coalescing operator here, please

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Handle trailing slash redirect
+     *
+     * We always want there to be a trailing slash for consistency
+     */
+    public function handleTrailingSlash(): void
+    {
+        // Add trailing slash if missing (and route is not empty)
+        if ($this->route !== '' && substr($this->route, -1) !== '/') {
+            header('Location: ' . $this->route . '/', true, 301);
```
kojiromike: Can we use Symfony\Component\HttpFoundation\RedirectResponse here?

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Handle trailing slash redirect
+     *
+     * We always want there to be a trailing slash for consistency
+     */
+    public function handleTrailingSlash(): void
+    {
+        // Add trailing slash if missing (and route is not empty)
+        if ($this->route !== '' && substr($this->route, -1) !== '/') {
```
kojiromike: use `str_ends_with` please

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    public function isForbiddenPath(): bool
+    {
+        // Check if route is deprecated (likely bugs - should return 404)
+        if ($this->config->isDeprecated($this->route)) {
+            return true;
+        }
+
+        // Check if route matches forbidden patterns
+        return $this->config->isForbidden($this->route);
```
kojiromike:
```suggestion
        return $this->config->isDeprecated($this->route) || $this->config->isForbidden($this->route);
```

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        $targetFile = realpath($this->baseDir . '/' . $this->route);
+
+        // Path traversal prevention
+        if ($targetFile === false || strpos($targetFile, $this->baseDir) !== 0) {
+            return null;
+        }
+
+        // Verify file exists
+        if (!file_exists($targetFile) || !is_file($targetFile)) {
```
kojiromike:
```suggestion
        if (!is_file($targetFile)) {
```

src/OpenCoreEMR/FrontController/SecurityValidator.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Check if running in CLI mode
+     */
+    public static function isCliMode(): bool
+    {
+        return php_sapi_name() === 'cli';
+    }
+
+    /**
+     * Check if route is an .inc.php file
```
kojiromike: There are also files ending in just `.inc`. I guess those don't matter because they'd already be excluded?

src/OpenCoreEMR/FrontController/SecurityValidator.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        error_log("OpenEMR Front Controller: " . $message);
+    }
+
+    /**
+     * Get debug message based on log level
+     */
+    public static function getDebugMessage(string $defaultMessage, string $debugMessage): string
+    {
+        return getenv('OPENEMR_FC_LOG_LEVEL') === 'debug' ? $debugMessage : $defaultMessage;
```
kojiromike: For consistency with existing code, I think these env vars should start with `OPENEMR__` (double underscore).

tests/reports/compatibility-test-report.json
```diff
@@ -0,0 +1,5 @@
+{
+    "test_date": "2025-10-03 12:12:10",
+    "base_url": "http:\/\/localhost\/openemr",
+    "compatibility_status": "All core functionality preserved"
+}
```
kojiromike, line 5: Please set up your editor to end all files with a newline.

tests/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,273 @@
...
+        echo "  URL: $url" | tee -a "$REPORT_FILE"
+    fi
+    echo | tee -a "$REPORT_FILE"
+}
+
+# Test 1: Core Entry Points
+echo "========================================" | tee -a "$REPORT_FILE"
+echo "TEST CATEGORY: Core Entry Points" | tee -a "$REPORT_FILE"
+echo "========================================" | tee -a "$REPORT_FILE"
+echo | tee -a "$REPORT_FILE"
```
kojiromike: Please apply the comments I made in my earlier review to every case where a similar thing happens. In this case, this should be…

```suggestion
{
    echo "========================================" | tee -a "$REPORT_FILE"
    echo "TEST CATEGORY: Core Entry Points" | tee -a "$REPORT_FILE"
    echo "========================================" | tee -a "$REPORT_FILE"
    echo
} | tee -a "$REPORT_FILE"
```

tests/scripts/test_performance.sh
```diff
@@ -0,0 +1,296 @@
...
+    {
+        echo "Front Controller Overhead:"
+        echo "  Baseline (direct): ${baseline_time}ms"
+        echo "  Front controller: ${fc_time}ms"
+        echo "  Overhead: ${overhead}%"
+        echo
+    } | tee -a "$REPORT_FILE"
+
+    # Performance verdict
+    if (( $(echo "$overhead < 5" | bc -l) )); then
```
kojiromike:
```suggestion
    if (( $(bc -l <<< "$overhead < 5") )); then
```

tests/scripts/test_performance.sh
```diff
@@ -0,0 +1,296 @@
...
+        {
+            printf "%s✓ EXCELLENT: Overhead < 5%%%s\n" "$GREEN" "$NC"
+            echo "  Front controller has negligible performance impact"
+        } | tee -a "$REPORT_FILE"
```
kojiromike: Consider writing a shell function to reduce duplication across all of these condition checks.

tests/scripts/test_security.sh
```diff
@@ -0,0 +1,268 @@
...
+    echo "========================================"
+    echo
+} | tee -a "$REPORT_FILE"
+
+TOTAL_TESTS=$((TOTAL_TESTS + 1))
+printf "Testing: Front controller responds when enabled... " | tee -a "$REPORT_FILE"
+
+response=$(curl -s -o /dev/null -w "%{http_code}" "$BASE_URL/home.php?_ROUTE=index.php")
+
+if [[ "$response" != "404" ]]; then
```
kojiromike: Probably the response should be more explicit than "not 404". It could be a 403, or a 500, and we wouldn't want those to pass either, right?

tests/Tests/FrontController/CompatibilityTest.php
```diff
@@ -0,0 +1,297 @@
...
+    {
+        $report = [
+            'test_date' => date('Y-m-d H:i:s'),
+            'base_url' => self::$baseUrl,
+            'compatibility_status' => 'All core functionality preserved',
+        ];
+
+        $reportPath = __DIR__ . '/../../reports/compatibility-test-report.json';
+        @mkdir(dirname($reportPath), 0755, true);
+        file_put_contents($reportPath, json_encode($report, JSON_PRETTY_PRINT));
```
kojiromike: If this file is edited by the tests, should it be in version control?

tests/Tests/FrontController/SecurityTest.php
```diff
@@ -0,0 +1,249 @@
...
+            'base_uri' => self::$baseUrl,
+            'http_errors' => false,
+            'allow_redirects' => true,
+        ]);
+        self::loadVulnerableIncFiles();
+    }
+
+    private static function loadVulnerableIncFiles(): void
+    {
+        $basePath = dirname(dirname(__DIR__));
```
kojiromike:
```suggestion
        $basePath = dirname(__DIR__, 2);
```

tests/Tests/FrontController/SecurityTest.php
```diff
@@ -0,0 +1,249 @@
...
+            'interface/login/login.php',
+        ];
+
+        foreach ($legitimateFiles as $file) {
+            $response = self::$client->head('/' . $file);
+            $httpCode = $response->getStatusCode();
+
+            $this->assertContains(
+                $httpCode,
+                [200, 302],
```
kojiromike: I'm still not sure why a 302 would be OK here.

tests/Tests/FrontController/SecurityTest.php
```diff
@@ -0,0 +1,249 @@
...
+
+        $this->markTestSkipped(
+            'This test requires environment variable control - manual testing required'
+        );
+    }
+
+    /**
+     * Test security headers are present
+     */
+    public function testSecurityHeadersPresent(): void
```
kojiromike: AFAICT we're testing the .htaccess here. We probably shouldn't use this test code to test things that aren't in the PHP.

.htaccess
```diff
@@ -0,0 +1,86 @@
...
+    # This approach simplifies deployment - no Apache environment variable configuration required
+    RewriteCond %{REQUEST_URI} \.php$
+    RewriteRule ^(.*)$ home.php?_ROUTE=$1 [QSA,L]
+</IfModule>
+
+# Security headers
+<IfModule mod_headers.c>
+    Header set X-Content-Type-Options "nosniff"
+    Header set X-XSS-Protection "1; mode=block"
+    Header set X-Frame-Options "SAMEORIGIN"
```
kojiromike: I have a feeling `X-Frame-Options` will not work well with openemr's current architecture. While these security headers might be a good idea in general, I think they're a bit off topic for this PR, and should be addressed in a separate changeset.

home.php
```diff
@@ -0,0 +1,131 @@
...
+use Dotenv\Dotenv;
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+// Load .env variables using Dotenv library
+$dotenv = Dotenv::createImmutable(__DIR__);
+$dotenv->safeLoad();
+
+// Check feature flag - when disabled, provide transparent pass-through
+if (!getenv('OPENEMR_ENABLE_FRONT_CONTROLLER') || getenv('OPENEMR_ENABLE_FRONT_CONTROLLER') !== '1') {
```
kojiromike: Looks like a good spot for the null coalesce operator.

home.php
```diff
@@ -0,0 +1,131 @@
...
+    require $targetFile;
+    exit;
+}
+
+// Early extension point via event system
+// Modules can listen to 'front_controller.early' event for early initialization
+// See https://github.com/adunsulag/oe-module-custom-skeleton for module development
+$GLOBALS['kernel']->getEventDispatcher()->dispatch(
+    new \Symfony\Component\EventDispatcher\GenericEvent(),
+    'front_controller.early'
```
kojiromike: Please create a new event entry in `src/Events/FrontController`.

home.php
```diff
@@ -0,0 +1,131 @@
...
+
+// Route to target file
+require $targetFile;
+
+// Late extension point via event system (after content is loaded)
+// Modules can listen to 'front_controller.late' event for post-processing
+$GLOBALS['kernel']->getEventDispatcher()->dispatch(
+    new \Symfony\Component\EventDispatcher\GenericEvent(),
+    'front_controller.late'
+);
```
kojiromike: We should think about what to do if the $targetFile exits or crashes. We could use `try/finally`, but that wouldn't necessarily handle `exit()`. Maybe `register_shutdown_function`?

I think we may want to even provide that context to the event, so that the event handler can do things based on whether the included script completed normally, threw an exception, or exited early.

tests/Tests/FrontController/CompatibilityTest.php
```diff
@@ -0,0 +1,297 @@
...
+    public function testGenerateCompatibilityReport(): void
+    {
+        $report = [
+            'test_date' => date('Y-m-d H:i:s'),
+            'base_url' => self::$baseUrl,
+            'compatibility_status' => 'All core functionality preserved',
+        ];
+
+        $reportPath = __DIR__ . '/../../reports/compatibility-test-report.json';
+        @mkdir(dirname($reportPath), 0755, true);
```
kojiromike: why do we want to hide warnings here?

custom/.gitkeep
kojiromike, line 1: is this directory still being used?

kojiromike (changes requested):
Some additional feedback. Also, I'm trying not to duplicate my comments, so please apply what I say to any place that a similar pattern occurs.

sjpadgett:
> I'm confused –
Yep sorry

kojiromike:
@coding-sunshine I didn't realize how recent array_any and array_all are. They're very new features to PHP (>=8.4). We'll need to go back to a foreach loop in those cases, as we still need to support php 8.2 and up here.

I apologize for the error.

src/Events/FrontController/FrontControllerEvent.php
```diff
@@ -0,0 +1,127 @@
...
+        /**
+         * The site ID (multisite identifier)
+         */
+        private readonly string $siteId = 'default',
+        /**
+         * Context data for the event
+         */
+        private array $context = []
+    ) {
+    }
```
kojiromike:
```suggestion
    public function __construct(
        private readonly string $route = '',
        private readonly string $siteId = 'default',
        private array $context = []
    ) {
    }
```

interface/login/login.php
```diff
@@ -25,9 +25,11 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
-// prevent UI redressing
+// prevent UI redressing and set security headers
 Header("X-Frame-Options: DENY");
 Header("Content-Security-Policy: frame-ancestors 'none'");
+Header("X-Content-Type-Options: nosniff");
+Header("X-XSS-Protection: 1; mode=block");
```
kojiromike: These should be done independent of this PR.

src/Events/FrontController/FrontControllerEvent.php
```diff
@@ -0,0 +1,127 @@
...
+     * Set context data
+     *
+     * @param array $context
+     * @return FrontControllerEvent
+     */
+    public function setContext(array $context): FrontControllerEvent
+    {
+        $this->context = $context;
+        return $this;
+    }
```
kojiromike:
```suggestion
    /**
     * Set context data
     *
     * @param array $context
     * @return static
     */
    public function setContext(array $context): static
    {
        $this->context = $context;
        return $this;
    }
```

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,136 @@
...
+    private readonly string $baseDir;
+    private string $route;
+    private string $siteId;
+    private readonly RouteConfigInterface $config;
+
+    public function __construct(string $baseDir, ?RouteConfigInterface $config = null)
+    {
+        $this->baseDir = realpath($baseDir);
+        // Ensure baseDir is valid
+        if ($this->baseDir === false || $this->baseDir === '') {
```
kojiromike:
```suggestion
        if (!$this->baseDir) {
```

`realpath` cannot return the empty string, so the only falsy value it can return is `false`, but we can be secure and concise just by using `!`.

src/OpenCoreEMR/FrontController/Router.php
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Resolve target file path
+     */
+    public function resolveTargetFile(): ?string
+    {
+        $targetFile = realpath($this->baseDir . '/' . $this->route);
+
+        // Path traversal prevention
+        if ($targetFile === false || !str_starts_with($targetFile, $this->baseDir . DIRECTORY_SEPARATOR)) {
```
kojiromike:
```suggestion
        $targetFile = realpath($this->baseDir . DIRECTORY_SEPARATOR . $this->route);

        // Path traversal prevention
        if ($targetFile === false || !str_starts_with($targetFile, $this->baseDir . DIRECTORY_SEPARATOR)) {
```

src/Events/FrontController/FrontControllerEvent.php
```diff
@@ -0,0 +1,127 @@
...
+ * request processing pipeline for logging, authentication, rate limiting, and
+ * other custom processing needs.
+ *
+ * AI DISCLOSURE: This file contains code generated using Claude AI (Anthropic)
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    OpenCoreEMR, Inc.
+ * @copyright Copyright (c) 2025 OpenCoreEMR, Inc.
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
```
kojiromike: Please use this `@license` line consistently everywhere, instead of sometimes shortcutting it with `GPLv3`.

home.php
```diff
@@ -0,0 +1,166 @@
...
+    $route = $_GET['_ROUTE'] ?? 'index.php';
+
+    // Security: Prevent directory traversal and null byte injection
+    if (strpos($route, '..') !== false || strpos($route, "\0") !== false) {
+        http_response_code(404);
+        exit('Not Found');
+    }
+
+    // Resolve and validate target file
+    $targetFile = __DIR__ . '/' . ltrim($route, '/');
```
kojiromike: Please use `DIRECTORY_SEPARATOR` consistently so we don't break compatibility with windows.

home.php
```diff
@@ -0,0 +1,166 @@
...
+    if (isset($GLOBALS['kernel']) && method_exists($GLOBALS['kernel'], 'getEventDispatcher')) {
+        $error = error_get_last();
+        $context = [
+            'completed_normally' => $error === null || !in_array($error['type'], [E_ERROR, E_PARSE, E_CORE_ERROR, E_COMPILE_ERROR]),
+            'error' => $error,
+        ];
+        $event = new FrontControllerEvent($route, $router->getSiteId(), $context);
+        $GLOBALS['kernel']->getEventDispatcher()->dispatch($event, FrontControllerEvent::EVENT_LATE);
+    }
+});
```
kojiromike: Would it make sense to create the shutdown function once and then use it both here and below on line 155?

index.php
```diff
@@ -4,6 +4,12 @@
 // modify it under the terms of the GNU General Public License
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
+
+// Set security headers
```
kojiromike: Any changes to security headers should be done in a separate PR.

kojiromike:
@coding-sunshine please resolve conflicts and also go through the old comments and mark anything resolved that is resolved.

src/OpenCoreEMR/FrontController/RouteConfig.php
coding-sunshine: @kojiromike 
 PHP Version Compatibility Issue:
  - array_any() was introduced in PHP 8.4 (released November 2024)
  - OpenEMR requires PHP ≥8.2.0 (line 14 in composer.json)
  - Using array_any() would break compatibility with PHP 8.2 and 8.3
  
  If I change this we will have issues with php 8.2 and 8.3
  Can you please confirm

kojiromike (commented):
I think you have a test that is checking for security headers that are no longer being set. Also, please address the composer normalize and rector issues, by running composer normalize and rector on the codebase.

coding-sunshine:
> I think you have a test that is checking for security headers that are no longer being set. Also, please address the composer normalize and rector issues, by running composer normalize and rector on the codebase.

Hi @kojiromike can you please check once again.

kojiromike:
Please run rector and composer normalize.

src/Events/FrontController/FrontControllerEvent.php
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Provides extension points for custom modules to hook into the front controller
+ * request processing pipeline for logging, authentication, rate limiting, and
+ * other custom processing needs.
+ *
+ * AI DISCLOSURE: This file contains code generated using Claude AI (Anthropic)
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    OpenCoreEMR, Inc.
```
kojiromike: @coding-sunshine you can put your own name here for the author.

src/OpenCoreEMR/FrontController/RouteConfig.php
kojiromike: Yes, please remove `array_any` per my comment here https://github.com/openemr/openemr/pull/9026#issuecomment-3367489069

kojiromike:
I think this is largely complete in #9943. If you want to improve the implementation, please let us know what you're going for.

PR #9027, opened 2025-10-03
Prior auth bug fixes and features

### LINKED ISSUE #9030: Prior Authorization Management Enhancements
### Describe the problem
Its hard to quickly understand the prior auth report it gets cluttered with old authorizations and its not quick to interpret remaining auths since every auth is different some have 3 others give 1000s.

Include a visual progress bar indicating unit consumption.
Include percentage
Allow filtering to hide expired authorizations

trafficpest:
fixes #9030

Prior Auth bug fixes and Features
fixed date type hint bug on a funtion when adding new auth

changed CPTs to CPT since you can only add one singular at a time

Removed duplicates rows in global report

added pie graph of % of all auths used on global report

Added bar graph and percent in tables of patient auths and global
auth reports to easily see how many auths you have remaining.
To know when to reauth quickly
It is color coded green 67-100%, yellow 34-66%, red 0-33%


interface/modules/custom_modules/oe-module-prior-authorizations/public/index.php
```diff
@@ -161,6 +162,29 @@ function refreshme() {
...
+                        }
+                        $barColor = '#4CAF50'; // Green: Full/Plenty remaining
+                        if ($percentRemaining <= 66) {
+                            $barColor = '#ffc107'; // Yellow: Getting low
+                        }
+                        if ($percentRemaining <= 33) {
+                            $barColor = '#dc3545'; // Red: Very low
+                        }
+
+                        print TABLE_TD;
```
kojiromike: echo is probably marginally better than print for these.

interface/modules/custom_modules/oe-module-prior-authorizations/public/index.php
```diff
@@ -161,6 +162,29 @@ function refreshme() {
...
+                        } else {
+                            $percentRemaining = 0;
+                        }
+                        $barColor = '#4CAF50'; // Green: Full/Plenty remaining
+                        if ($percentRemaining <= 66) {
+                            $barColor = '#ffc107'; // Yellow: Getting low
+                        }
+                        if ($percentRemaining <= 33) {
+                            $barColor = '#dc3545'; // Red: Very low
+                        }
```
kojiromike: Consider using a match expression here. Note that if $percentRemaining <= 33 it's also <= 66, so the assignment is done three times in that case.

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -26,6 +30,7 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title><?php echo xlt("List Exising Prior Auths Report"); ?></title>
+<script src="https://cdn.jsdelivr.net/npm/chart.js@4.4.0/dist/chart.umd.min.js"></script>
```
kojiromike: Please check indentation.

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -16,6 +16,10 @@
 $data = new AuthorizationService();
 $patients = $data->listPatientAuths();
 
+$hide_expired = isset($_GET['hide_expired']) && $_GET['hide_expired'] == '1';
```
kojiromike:
```suggestion
$hide_expired = ($_GET['hide_expired'] ?? '0') === '1';
```

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -92,21 +121,52 @@ function openNewTopWindow(newpid) {
                         print "<td></td>";
                     }
                     print "<td>" . text($iter['auth_num']) . "</td>";
+                    print "<td>" . text($iter['cpt']) . "</td>";
                     print "<td>" . text($iter['start_date']) . "</td>";
                     print "<td>" . text($iter['end_date']) . "</td>";
-                    if (($iter['end_date'] < date('Y-m-d')) && ($iter['end_date'] !== '0000-00-00') && !empty($iter['auth_num'])) {
+
+                    if (($iter['end_date'] < date('Y-m-d')) && ($iter['end_date'] !== '0000-00-00') && (!empty($iter['auth_num']))) {
```
kojiromike: Consider assigning the $iter['end_date'] check earlier to avoid duplication.

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -92,21 +121,52 @@ function openNewTopWindow(newpid) {
...
                         print "<td></td>";
+                        print "<td></td>";
                     } else {
-                        print "<td>" . text($iter['init_units']) . "</td>";
-                        $unitCount = $iter['init_units'] - $numbers;
+                        $initialUnits = (int)$iter['init_units'];
+                        print "<td>" . text($initialUnits) . "</td>";
+
+                        $usedUnits = $initialUnits - $numbers;
+                        $unitCount = $initialUnits - $numbers;
```
kojiromike:
```suggestion
                        $usedUnits = $initialUnits - $numbers;
                        $unitCount = $usedUnits;
```

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -92,21 +121,52 @@ function openNewTopWindow(newpid) {
...
+                            $percentRemaining = 0;
+                        }
+
+                        $barColor = '#4CAF50'; // Green: Full/Plenty remaining
+                        if ($percentRemaining <= 66) {
+                            $barColor = '#ffc107'; // Yellow: Getting low
+                        }
+                        if ($percentRemaining <= 33) {
+                            $barColor = '#dc3545'; // Red: Empty
+                        }
```
kojiromike:
```suggestion
                        $barColor = match(true) {
                            ($percentRemaining <= 33) => '#dc3545'; // Red: Empty
                            ($percentRemaining <= 66) => '#ffc107'; // Yellow: Getting low
                            default => '#4CAF50'; // Green: Full/Plenty remaining
                        };
```

You might also consider assigning the colors to variables or even an enum so the "empty/getting low/full" meaning can be gotten from the names.

kojiromike (commented):
Nice! I left some suggestions for simplifying some code and avoiding duplication.

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
trafficpest: Thank you I made all changes but the "$iter['end_date'] check earlier". Since I grouped the SQL all seems to be behaving good now. Thanks for the php match idea, I checked the php manual, and it shows commas, like arrays. So I changed it.

This PR should be ready now

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
trafficpest: ok I updated this too should be right

interface/modules/custom_modules/oe-module-prior-authorizations/public/index.php
```diff
@@ -150,22 +151,44 @@ function refreshme() {
...
+
+                        $barColor = match (true) {
+                            ($percentRemaining <= 33) => '#dc3545', // Red: Empty
+                            ($percentRemaining <= 66) => '#ffc107', // Yellow: Getting low
+                            default => '#4CAF50', // Green: Full/Plenty remaining
+                        };
+
+                        echo TABLE_TD;
+                        echo "<div style='background-color:#eee; height:20px; width:100px; border:1px solid #ccc; position:relative;'>";
+                        echo "<div style='background-color:{$barColor}; height:100%; width:{$percentRemaining}%; position:absolute;'></div>";
```
stephenwaite: hi @trafficpest, have to escape these variables:
```
                        echo "<div style='background-color:{" . attr($barColor) . "}; height:100%; width:{" . attr($percentRemaining) . "}%; position:absolute;'></div>";
```
there's at least one other spot in this PR

trafficpest: ok ill check now

kojiromike:
Is there an issue for this PR? We will need one, IIUC.

trafficpest:
> Is there an issue for this PR? We will need one, IIUC.

No, sorry. My sister-in-law had a few bugs and wanted some more features, then I thought I would share. I will make one in a bit. You guys are so fancy for me, lol.

I'm learning how you like things as I go. Thank you again

interface/modules/custom_modules/oe-module-prior-authorizations/public/index.php
trafficpest: ok did see two. I dont think i missed any more

interface/modules/custom_modules/oe-module-prior-authorizations/public/index.php
```diff
@@ -150,22 +151,44 @@ function refreshme() {
...
+                        $barColor = match (true) {
+                            ($percentRemaining <= 33) => '#dc3545', // Red: Empty
+                            ($percentRemaining <= 66) => '#ffc107', // Yellow: Getting low
+                            default => '#4CAF50', // Green: Full/Plenty remaining
+                        };
+
+                        echo TABLE_TD;
+                        echo "<div style='background-color:#eee; height:20px; width:100px; border:1px solid #ccc; position:relative;'>";
+                        echo "<div style='background-color:" . attr($barColor) . "; height:100%; width:" . attr($percentRemaining) . "%; position:absolute;'></div>";
+                        echo "<div style='position:absolute; top:0; width:100%; text-align:center; line-height:20px; color:#000; font-weight:bold; font-size:12px;'>{$percentRemaining}%</div>";
```
stephenwaite: this should be escaped too:
```
echo "<div style='position:absolute; top:0; width:100%; text-align:center; line-height:20px; color:#000; font-weight:bold; font-size:12px;'>" . text($percentRemaining) . "%</div>";

File: interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php [hunk 10]
```diff
@@ -73,41 +98,75 @@ function openNewTopWindow(newpid) {
...
+                        $barColor = match (true) {
+                            ($percentRemaining <= 33) => '#dc3545', // Red: Empty
+                            ($percentRemaining <= 66) => '#ffc107', // Yellow: Getting low
+                            default => '#4CAF50', // Green: Full/Plenty remaining
+                        };
+
+                        echo "<td>";
+                        echo "<div style='background-color:#eee; height:20px; width:150px; border:1px solid #ccc; position:relative;'>";
+                        echo "<div style='background-color:" . attr($barColor) . "; height:100%; width:" . attr($percentRemaining) . "%; position:absolute;'></div>";
+                        echo "<div style='position:absolute; top:0; width:100%; text-align:center; line-height:20px; color:#000; font-weight:bold; font-size:12px;'>{$percentRemaining}%</div>";
```
stephenwaite: and here :)

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -120,6 +179,49 @@ function openNewTopWindow(newpid) {
         </div>
         &copy; <?php echo date('Y') . " Juggernaut Systems Express" ?>
     </div>
+   <script>
+    const total_initial = <?php echo $total_initial_units; ?>;
```
stephenwaite: these have to be `js_escaped()`

adunsulag:
@juggernautsei can you take a look over this, looks like a good addition, but want to make sure functionality wise it will still work the way your other clients are using it.

trafficpest:
> Is there an issue for this PR? We will need one, IIUC.

https://github.com/openemr/openemr/issues/9030

trafficpest:
> @juggernautsei can you take a look over this, looks like a good addition, but want to make sure functionality wise it will still work the way your other clients are using it.

@adunsulag I did deactivate his code in my last commited PR for this module that required his custom table. Also, a main dev had me remove his custom patient menu overwrite for a simple patient menu append. I left his code for his custom table there just commented out that way vanilla OpenEMR users can use this plugin and he can always just uncomment.

sjpadgett:
I think this is going in a good direction.
The last PR was basically bug fixes and user interface consistencies. I do wanna pull it down and test later tonight.

interface/modules/custom_modules/oe-module-prior-authorizations/public/index.php
trafficpest: im so ignorant on these things I left $percentRemaining I thought since it returns a int it would be fine. thanks for teaching me

juggernautsei:
@adunsulag I will look this over this weekend. @kojiromike when the custom menu was conceived. At the time we had no way to inject into the dashboard menu bar. So, the custom bar was used. We should look into using some java script to inject the navigation into the native dashboard menu to get rid of the custom menu. I like what I read. I will try it out on one of our systems to see what it looks like. I like the idea of the graphs. Visuals are always good.

trafficpest:
@juggernautsei 
Note: the SQL count incorporates auth code + billed units of service code + service date to get an accurate count. My sister-in-law gets auth codes that include a date range and more than one service code each with different billed units. So now you enter amount for each code with auth. Thank you for the module, it was really helpful!

trafficpest:
This should be good to go now. I haven't tested since all these minor commits but I think it should be good.

juggernautsei:
I loaded the module and found in the index.php and the list_report.php needed to add the require autoloader in both files then it works. 
@sjpadgett  I noticed that there is a cdn reference. Is this permissible? 

image

juggernautsei:
There is so much I would restructure about this module now. Hopefully, when we find an API partner. I will do the restructuring of this module to bring it up to date.

interface/modules/custom_modules/oe-module-prior-authorizations/public/reports/list_report.php
```diff
@@ -26,6 +30,7 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title><?php echo xlt("List Exising Prior Auths Report"); ?></title>
+    <script src="https://cdn.jsdelivr.net/npm/chart.js@4.4.0/dist/chart.umd.min.js"></script>
```
bradymiller: can't do this. remove above line. Note you are in luck since OpenEMR has this library. You can include the chart.js library in this script by changing line 27 to:
```
<?php Header::setupHeader(['common', 'opener', 'chart']) ?>
```

trafficpest: done

trafficpest:
> I loaded the module and found in the index.php and the list_report.php needed to add the require autoloader in both files then it works. @sjpadgett I noticed that there is a cdn reference. Is this permissible?

@kojiromike Why does the module need the vendor autoload still? I thought core app will include registered modules autoloads? Should the include be there or not?

juggernautsei:
@sjpadgett  The module needs to be updated. This module was created prior to the work that @adunsulag did in the openemrbootstrap to autoload classes into the OpenEMR namespace. That will be part of my refactoring of this module to restructure it. 
@adunsulag said something to me a long time ago that stuck with me about polluting the OpenEMR namespace with my modules. So, from that point forward, I started putting my modules in their own namespace to keep it separated out from the OpenEMR namespace. Then he went and built the function/feature to load the module namespace into the main autoloader. Which frankly baffled me. But I still build my own namespaces most of the time.

sjpadgett:
So what's the question? 
- Should we be using autoloader? Depends! Does the module require dependencies that core doesn't provide? For instance, install a new widget library that may only be used by module and wouldn't serve core in any significant way however makes the module great. Then this would require an install mechanism where a package or composer run is needed.
- So what's in a namespace? If a core module I use openemr's module namespace and register from bootstrap(my preferred method). You can still have a composer json that describes the module and its PSR-4 pathing. I haven't yet needed to specifically run an autoloader specific to my modules. 

Our first module was my fax/sms that Stephen engineered the integration and added the namespace register and i've used and expanded on this method since. Check out my ModuleManagerListener where it allows the module to register and live in a namespace on install from module manager.
This method may have some limitations solved by composer PSR-4 handling. I do both. However IMO if you need to run an autoloader that resides in the module then we missed it in integrating module into core where it was a third party install previous.

Having @adunsulag module skeleton is terrific but remember, it is designed to show all the capabilities of modules and the preferred work flow for the various actions not necessarily all of which is required to be present.

So two thing caught my attention with this module: autoload and Patient Menu.
First autoload call shouldn't be needed and figure out and register your namespace. Second: whenever modifying the menus in openemr, existing menus shouldn't be removed or renamed unless a specific reason is given and admin is aware on module install. This is especially true for the Patient Menu and should stay intact due to items required for ONC. Add to but not take from!

I really like this module and I think it is necessary in a majority of installs. In fact it wouldn't hurt to enable by default! lol no please don't do that.
 @juggernautsei Are you going to pull down and help @trafficpest test and resolve the autoload issue? It'd be worthwhile instead of you doing mod's later after this is merged to help author resolve. I'll help if needed

juggernautsei:
@sjpadgett, I have already pulled down the module. Again, that is why I want to refactor this module to take advantage of the advances in the module ecosystem.  

When I first designed the module. I talked with Stephen briefly about it and the Patient Menu. Again, since there was no way at the time of injecting in the that menu. We opted to do the replacement which added the Prior Auth to the patient menu. Now, I should be able to expand the injection process to the patient menu. 

The other reason, I personally us an autoloader is sometimes I use PHP SlimFramwork in my modules. I like my modules having its own router. It gives me clean URL and decouples stuff like you using Phreezy. A lot of this stuff is just a matterr of preferences in my opinion. Of course we have to bow to the leaders of the gang of coders. There are still things I can't see yet but I will get there in time.  

So, I will refactor the openemrbootstrap.php file to include the class loader. Create a Bootstrap.php class in the /src and put in a really basic router. I am going to add and interface for the different prior auth providers or it may be better to use and abstract class. I will figure it out when I get closer to having a provider to work with.

sjpadgett:
@sjpadgett That's fine @juggernautsei it's your module and I'm really happy you contributed it. Can't wait to see outcome

juggernautsei:
@trafficpest, I have a question about the graphic at the top of the List report. 
The graphic is adding all of the units from every patient in the system together. Is that the way it's supposed to work? It doesn't seem helpful to me, but I am not the provider. 

image

@sjpadgett I have completed the refactoring of this module. I will post a repo and link to it. I will haven't found any insurance prior authorization providers yet. I am running into the same thing as with the most anything else. They want to know how many providers are there, what is the volume of blah blah blah

sjpadgett:
@trafficpest @juggernautsei 
Sherwin, I believe you should be able to push your changes to a branch in the authors repo where he reviews and merges into his PR repo.
Once that is done we can merge to core.

juggernautsei:
@sjpadgett I will work that out. 
I don't have access to push changes to @trafficpest  repo. I am stuck at this point

image

Here is what is looks like now.
https://komododecks.com/recordings/LgwT4CEEUwissyjkILWt

trafficpest:
> @trafficpest, I have a question about the graphic at the top of the List report. The graphic is adding all of the units from every patient in the system together. Is that the way it's supposed to work? It doesn't seem helpful to me, but I am not the provider.

Yes. The idea is to be able to have a quick, single view of authorizations as a whole. Sum all your active authorizations so you can tell normal ranges, etc. It wouldn't be a bad idea to capture some data points to show trends, but I think its not worth the effort. Main reason is to quickly see if people doing auths are doing their jobs.

trafficpest:
> @sjpadgett I will work that out. I don't have access to push changes to @trafficpest repo. I am stuck at this point
> 
@juggernautsei, sorry for the delay. I have sent an invite to you to be a collaborator and push code to my repo.
In the end what did you do with the module autoload? rewrite the bootstrap or just return the require in file?

Side note:
I have bad news for you guys: I got my wife's therapy practice to switch to OpenEMR after seeing it at my sister-in-law's ABA practice. Now you'll probably be dealing with me every once in a while, haha.

She already has me busy training employees at different locations, dealing with data migration, and, on top of that, I'm tasked with writing a Braintree payment integration module (her processor) that takes copays automatically on check-in. I'm tying it to arrival on the calendar, as I couldn't find an event for encounter.

trafficpest:
@juggernautsei, I see you're adding more than one CPT code in the string, and the way I changed the SQL to match the CPT and billed units, that would fail. I changed the label text from the plural CPTs to the singular CPT. Most authorizations have different units permitted for each service code, so the provider needs to add each one independently. The form could be changed, or the match condition could be changed from '=' to 'IN' in the SQL.

Screenshot From 2025-10-07 23-16-59

juggernautsei:
@trafficpest , We need to add an explainer on how to use the feature. So, people don't have to guess how the system is set up. I looked for the invite. Can I use your screenshot in the help file? I have pushed my changes.

trafficpest:
@juggernautsei the changes arent passing the project testing standards. Do you want to go through this, or do you want me to revert to a prior commit that was and just include the autoload to pass all tests and have  it working to close this PR?

sjpadgett:
@trafficpest Click the PHPStan and will take you to what is failing. Click Annotations at top to see specific.
Never backwards, always forward!:)
I'll try go through and test sometime tonight

interface/modules/custom_modules/oe-module-prior-authorizations/openemr.bootstrap.php
```diff
@@ -11,69 +11,17 @@
...
- * @var EventDispatcherInterface $eventDispatcher
- * @var array                    $module
- * @global                       $eventDispatcher @see ModulesApplication::loadCustomModule
- * @global                       $module          @see ModulesApplication::loadCustomModule
+ * @global EventDispatcher $eventDispatcher Injected by the OpenEMR module loader;
  */
-
-$eventDispatcher->addListener(MenuEvent::MENU_UPDATE, 'oe_module_priorauth_add_menu_item');
-$eventDispatcher->addListener(PatientMenuEvent::MENU_UPDATE, 'oe_module_priorauth_patient_menu_item');
+$bootstrap = new Bootstrap($eventDispatcher);
```
sjpadgett, line 26: We may need to add rule to ignore Stan error for missing var however I generally try to appease in code. In this case:
$eventDispatcher = $GLOBALS['kernel']->getEventDispatcher();
Do the same with classloader.

I still don't understand why using, too me, the confusing namespace Juggernaut\OpenEMR\Modules\PriorAuthModule. 
This module directly interacts with core and I don't see any dependencies needing to isolate from core namespace. I won't harp on this but we need to be consistent. Mostly for future if module is modified to use some dependency in core if you'd want to use core resources. If this was third party or a fee based module then okay.

There may be security concerns where staying outside openemr namespace has advantages. I don't know.

juggernautsei:
@trafficpest, do you have time to update this PR to fix PHPStan?

trafficpest:
Honesty I'm pretty slammed I'll try to to see usually those tests are ticky
tack stuff like formatting.

Have you tried to check?

interface/modules/custom_modules/oe-module-prior-authorizations/openemr.bootstrap.php
juggernautsei: @sjpadgett, the namespace was a comment that @adunsulag said to me a long time ago about not polluting the OpenEMR namespace with a module. So, I endeavoured to do that. I made sure my modules all had their own namespace. Then @adunsulag seemed to have changed his mind and added the bootstrap, which autoloaded the namespace with the core namespace. Nowadays, I don't use separate namespaces anymore.

juggernautsei:
@trafficpest, I understand. Time is always tight for most of us.

PR #9034, opened 2025-10-04
ci(rector): list applied rules

kojiromike:
Makes the list of applied rector rules visible in the CI job. This gives us important debugging information when dealing with rector.


kojiromike:
Since this only affects the rector workflow, I am going to kill the test job and merge this as is to save resources.

PR #9073, opened 2025-10-06
refactor(php): replace strpos() !== false and strstr() with str_contains()

### LINKED ISSUE #9072: `StrContainsRector` - Replace strpos() !== false and strstr() with str_contains()

kojiromike:
Fixes #9072

Apply `StrContainsRector` rule

- [x] chore(rector): set StrContainsRector**
- [x] refactor(php): replace strpos() !== false and strstr() with str_contains()**


library/api.inc.php
```diff
@@ -52,9 +52,9 @@ function formSubmit($tableName, $values, $id, $authorized = "0")
 
                 BillingUtilities::addBilling(date("Ymd"), 'CPT4', $code_array[0], $code_array[1], $_SESSION['pid'], $authorized, $_SESSION['authUserID']);
             }
-        } elseif (strpos($key, "diagnosis") == (strlen($key) - 10) && !(strpos($key, "diagnosis") === false )) {
-            //case where key looks like "[a-zA-Z]*diagnosis[0-9]" which is special, it is used to auto add ICD codes
-            //icd auto add ICD9-CM
+        } elseif ((bool) preg_match("/diagnosis\d$/", $key)) {
```
kojiromike, line 55: @adunsulag I made this change manually because after reading the comment it seems more correct than the previous version. LMK if you want me to change it back.

adunsulag: I think it looks ok, I'm not sure how much this code gets executed in the 3rd party forms, we'd need to trace down where its utilized and make sure behavior hasn't changed,

bradymiller: considering for icd9 stuff, guessing not even significant anymore

PR #9085, opened 2025-10-08
refactor(php): use `public`, not `var`

### LINKED ISSUE #9084: `VarToPublicPropertyRector` - change property modifier from var to public

kojiromike:
Fixes #9084

Apply `VarToPublicPropertyRector` rule.

- [x] chore(rector): set VarToPublicPropertyRector
- [x] refactor(php): use `public` instead of `var`

### DIFF
--- a/modules/sms_email_reminder/sms_clickatell.php
+++ b/modules/sms_email_reminder/sms_clickatell.php
@@ -66,7 +66,7 @@ class sms
     * Define SMS balance limit below class will not work
     * @var integer
     */
-    public $balace_limit = 0;
+    public $balance_limit = 0;
 
     /**
     * Gateway command sending method (curl,fopen)
@@ -167,7 +167,7 @@ class sms
     {
 
         /* Check SMS credits balance */
-        if ($this->getbalance() < $this->balace_limit) {
+        if ($this->getbalance() < $this->balance_limit) {
             die("You have reach the SMS credit limit!");
         };
 


library/classes/smtp/sasl.php
```diff
@@ -110,20 +110,20 @@ functions that may fail and only check the failure condition after
 	</variable>
 {/metadocument}
 */
-	var $encode_response=1;
+	public $encode_response=1;
 
 	/* Private variables */
```
kojiromike, line 115: This is suspicious. These were always public, because `var` isn't a visibility modifier.

library/classes/NumberToText.class.php
```diff
@@ -33,17 +33,8 @@
 
 class NumberToText
 {
-    var $number;
-    var $currency;
-    var $capatalize;
-    var $and;
-
-    function __construct($number, $currency = false, $capatalize = false, $and = false)
+    function __construct(public $number, public $currency = false, public $capitalize = false, public $and = false)
```
bradymiller, line 36: wow, did rector even auto fix the mispelled $capatalize ??

library/classes/smtp/sasl.php
bradymiller: would ignore the comment :)

modules/sms_email_reminder/sms_clickatell.php
```diff
@@ -42,61 +42,61 @@ class sms
...
     */
-    var $use_ssl = false;
+    public $use_ssl = false;
 
     /**
     * Define SMS balance limit below class will not work
     * @var integer
     */
-    var $balace_limit = 0;
+    public $balance_limit = 0;
```
bradymiller, line 69: wow, another mispell fix

bradymiller: couldn't this one lead to breakage though since guessing it is set directly by whatever is using this class?

bradymiller: not seeing where it is directly modified in codebase, so should be good.

library/classes/NumberToText.class.php
kojiromike: lol, nope. :)

modules/sms_email_reminder/sms_clickatell.php
kojiromike: Yep, I searched the codebase for this too.

PR #9087, opened 2025-10-08
refactor(php): add `never` annotations

### LINKED ISSUE #9086: `ReturnTypeNeverRector` - annotate functions that never return

kojiromike:
Fixes #9086

Apply `ReturnTypeNeverRector` rule.

- [x] chore(rector): set ReturnNeverTypeRector
- [x] refactor(php): add `never` annotations

### DIFF
--- a/tests/Tests/Unit/Common/Auth/OpenIDConnect/Repositories/ScopeRepositoryTest.php
+++ b/tests/Tests/Unit/Common/Auth/OpenIDConnect/Repositories/ScopeRepositoryTest.php
@@ -229,7 +229,7 @@ class ScopeRepositoryTest extends TestCase
         $this->assertEquals("patient/Patient.ruds", $allowedScopes[2]->getIdentifier(), "Allowed scope should be patient/Patient.ruds");
     }
 
-    public function finalizeScopesWithInvalidSubScopeWillFail()
+    public function finalizeScopesWithInvalidSubScopeWillFail(): never
     {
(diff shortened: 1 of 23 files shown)


tests/Tests/Unit/Common/Auth/OpenIDConnect/Repositories/ScopeRepositoryTest.php
```diff
@@ -229,7 +229,7 @@ public function testFinalizeScopesWithValidSubScopeWillPass(): void
         $this->assertEquals("patient/Patient.ruds", $allowedScopes[2]->getIdentifier(), "Allowed scope should be patient/Patient.ruds");
     }
 
-    public function finalizeScopesWithInvalidSubScopeWillFail()
+    public function finalizeScopesWithInvalidSubScopeWillFail(): never
```
bradymiller, line 232: interesting that rector knows that the following line will throw an exception:
```
$this->markTestIncomplete("ScopeRepository::finalizeScopes() needs to be implemented and tested.");
```

kojiromike: It's because upstream is also a `never`: https://github.com/sebastianbergmann/phpunit/blob/1f5777c430671863ae8646c0a66e37de6a863cd5/src/Framework/Assert.php#L2982

PR #9117, opened 2025-10-11
Openemr fix 8787 fhir condition encounter diagnosis

### LINKED ISSUE #8787: feat: US Core Condition Encounter Diagnosis Profile
### Describe the problem

Implement US Core Condition Encounter Diagnosis Profile 7.0.0

- [ ] assertedDate 0..1 (dateTime)  - Date the condition was first asserted
- [ ] clinicalStatus 0..1 CodeableConcept - active | recurrence | relapse | inactive | remission | resolved
- [ ] verificationStatus 0..1 	CodeableConcept	unconfirmed | provisional | differential | confirmed | refuted | entered-in-error
- [ ] category 	1..* CodeableConcept
-- [ ] coding 1..*
--- [ ] system=http://terminology.hl7.org/CodeSystem/condition-category
--- [ ] code=encounter-diagnosis
- [ ] code 1..1 	CodeableConcept	Identification of the condition, problem or diagnosis (http://hl7.org/fhir/us/core/ValueSet/us-core-condition-code
 which includes snomed-ct findings, context-dependent categories, events, icd10-cm and icd9-cm)
 -- new instance resources must use the codings from the current value-set, but old codes should not be translated. 
- [ ] subject 1..1 Reference (Patient)
- [ ] encounter 0..1 Reference (encounter)
- [ ] onset[x]  0..1
-- [ ]  onsetDateTime  dateTime
- [ ]  abatement[x] 0..1 - 	When in resolution/remission
-- [ ] abatementDateTime dateTime
- [ ] recordedDate 0..1 dateTime - Date record was first recorded
- Required Search Terms
-- [ ] patient
-- [ ] patient+category

Optional Requirements
- Required Search Terms
-- [ ] patient+clinical-status
-- [ ] patient+category+clinical-status
-- [ ] patient+category+encounter
-- [ ] patient+code
-- [ ] patient+onset-date
-- [ ] patient+asserted-date
-- [ ] patient+recorded-date
-- [ ] patient+abatement-date
-- [ ] patient+_lastUpdated

adunsulag:
Fixes #8787 

This commit implements the baseline functionality needed for Conditions
with a type of encounter-diagnosis.  In 3.1.1 we were incorrectly
labeling them as problem-items even if they had an attached encounter
diagnosis.

I had to overhaul the db structure of the conditions as the 3.1.1
implementation would create duplicate condition records if a
medical_problem lists record was attached to multiple encounters via the
issue_encounter join table.  In order to represent the 'instance' nature
of the condition linked to an encounter per FHIR spec I had to add a
uuid to the issue_encounter join relationship.  To be backwards
compatible I added a cross-over date check on the condition so that
after our deployment on November 15th conditions will use the uuid from
the issue_encounter table and fallback to the old uuid for anything
before that date.

In order to accomplish this I had to centralize all the insert queries
everywhere issue_encounter is created (fee sheet, eye form, patient
issues). This allowed me to generate the uuids properly.

Our UUIDRegistry does not handle composite primary keys so I had to add
a surrogate primary key to the issue_encounters table that
auto_increments in order to get the UUIDRegistry to work properly
without overhaulings its functionality.  If a system has a large number
of linked encounters to issues this may increase the runtime of the
sql_upgrade.php.

I added in the FhirConditionProblemsHealthConcernService as well but
need to do further testing there before I'll consider that class
complete.

### DIFF
--- a/interface/patient_file/problem_encounter.php
+++ b/interface/patient_file/problem_encounter.php
@@ -27,7 +27,7 @@ use OpenEMR\Services\PatientIssuesService;
 /**
  * @global $pid  pid should always be defined but to deal with phpstan issues we'll put this statement here
  */
-$pid = $pid ?? null;
+$pid ??= null;
 $patdata = getPatientData($pid, "fname,lname,squad");
 
 $thisauth = ((AclMain::aclCheckCore('encounters', 'notes', '', 'write') ||


adunsulag:
@bradymiller or @stephenwaite I'm still working through this PR but if you get a chance to look over the DB changes especially dealing with issue_encounters I'd appreciate it.  FHIR requires us to represent a specific instance that a condition is linked to an encounter rather than any ongoing health problems that span across multiple conditions (its a 1:1 relationship from condition to encounter).  I had to extend the join table and I want to make sure I'm handling that properly.

Note it will slow down the sql_upgrade slightly on a patch if someone has a large number of linked encounters due to the ALTERS, primary key changes, and uuid generation.  Haven't benchmarked it but don't think it will be too bad.

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1297,3 +1297,14 @@ ALTER TABLE `employer_data` ADD COLUMN `created_by` INT DEFAULT NULL COMMENT 'fk
 #IfMissingColumn employer_data uuid
 ALTER TABLE `employer_data` ADD COLUMN `uuid` binary(16) DEFAULT NULL COMMENT 'UUID for this employer record, for data exchange purposes';
 #EndIf
+
+#IfMissingColumn issue_encounter uuid
+ALTER TABLE `issue_encounter` ADD COLUMN `uuid` binary(16) DEFAULT NULL COMMENT 'UUID for this issue encounter record, for data exchange purposes';
```
bradymiller, line 1388: generally need to also add a unique key for uuid when add uuid. something like below.
```
  UNIQUE KEY `uuid` (`uuid`),
```

This will ensure key is either null or unique.

I just realized I have been forgetting to recommend this on recent reviews.

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1297,3 +1297,14 @@ ALTER TABLE `employer_data` ADD COLUMN `created_by` INT DEFAULT NULL COMMENT 'fk
...
+#EndIf
+
+#IfMissingColumn issue_encounter id
+-- we
```
bradymiller: looks like unintended comment line above

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1297,3 +1297,14 @@ ALTER TABLE `employer_data` ADD COLUMN `created_by` INT DEFAULT NULL COMMENT 'fk
...
+ALTER TABLE `issue_encounter` ADD UNIQUE INDEX `uniq_issue_key`(`pid`,`list_id`,`encounter`);
+ALTER TABLE `issue_encounter` ADD COLUMN `id` BIGINT(20) NOT NULL AUTO_INCREMENT FIRST;
+ALTER TABLE `issue_encounter` DROP PRIMARY KEY;
```
bradymiller, line 1393: do then need to add id as a primary key?

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1297,3 +1297,14 @@ ALTER TABLE `employer_data` ADD COLUMN `created_by` INT DEFAULT NULL COMMENT 'fk
 #IfMissingColumn employer_data uuid
 ALTER TABLE `employer_data` ADD COLUMN `uuid` binary(16) DEFAULT NULL COMMENT 'UUID for this employer record, for data exchange purposes';
```
bradymiller, line 1298: also need to do this in database.sql

bradymiller: disregard above comment; realized not related to your code.

adunsulag: That employer_data was my code on a different PR, so can take care of it here.

bradymiller: Hi, I am guessing you already brought it into database.sql though.

adunsulag:
image

PR #9128, opened 2025-10-11
refactor(sl_eob_process)

kojiromike:
This was going to start out as simply adding named parameters to the file so that I could identify the mismatched calls, but it got away from me. But it's still good, I think.

### DIFF
--- a/interface/billing/sl_eob_process.php
+++ b/interface/billing/sl_eob_process.php
@@ -25,23 +25,38 @@ use OpenEMR\Billing\ParseERA;
 use OpenEMR\Billing\SLEOB;
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Core\Header;
-use OpenEMR\Services\InsuranceService;
 
+/** @var int $debug */
 $debug = $_GET['debug'] ? 1 : 0; // set to 1 for debugging mode
-$paydate = parse_date($_GET['paydate']);
+/** @var string $paydate */
+$paydate = parseDate($_GET['paydate']);
+/** @var int $encount */
 $encount = 0;
 
+/** @var string $last_ptname */
 $last_ptname = '';
+/** @var string $last_invnumber */
 $last_invnumber = '';
+/** @var string $last_code */
 $last_code = '';
+/** @var float $invoice_total */
 $invoice_total = 0.00;
-$InsertionId; // last inserted ID of
+/** @var array<string, int> $InsertionId last inserted ID of */
+$InsertionId = [];
+/** @var string $StringToEcho a manual buffer */
+$StringToEcho = '';
 
 ///////////////////////// Assorted Functions /////////////////////////
 
-function parse_date($date)
+/**
+ * Parse a date string into YYYY-MM-DD format.
+ *
+ * @param string $date The date string to parse
+ * @return string The formatted date string or empty string if invalid
+ */
+function parseDate(string $date): string
 {
-    $date = substr(trim((string) $date), 0, 10);
+    $date = substr(trim($date), 0, 10);
     if (preg_match('/^(\d\d\d\d)\D*(\d\d)\D*(\d\d)$/', $date, $matches)) {
         return $matches[1] . '-' . $matches[2] . '-' . $matches[3];
     }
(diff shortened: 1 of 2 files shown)


interface/billing/sl_eob_process.php
```diff
@@ -770,9 +847,7 @@ function era_callback(&$out): void
     <?php
     global $InsertionId;
 
-    $eraname = $_REQUEST['eraname'];
-    $alertmsg = ParseERA::parseERAForCheck($GLOBALS['OE_SITE_DIR'] . "/documents/era/$eraname.edi");
-    $alertmsg = ParseERA::parseERA($GLOBALS['OE_SITE_DIR'] . "/documents/era/$eraname.edi", 'era_callback');
+    $alertmsg = ParseERA::parseERA($era_file_path, 'era_callback');
```
kojiromike: Were the two `alertmsg` supposed to be appended? Because I just removed the first one.

bradymiller:
hi @kojiromike , code looks good to me. Will be good to ensure @stephenwaite takes a look at it.

interface/billing/sl_eob_process.php
```diff
@@ -49,117 +64,161 @@ function parse_date($date)
...
-            $bgcolor,
-            'olddetail',
-            $ptname,
-            $invnumber,
-            $code,
-            $ddate,
-            $description,
-            $amount,
-            $invoice_total
+        yield getDetailLine(
```
stephenwaite, line 192: seal-dunk

interface/billing/sl_eob_process.php
```diff
@@ -395,12 +463,20 @@ function era_callback(&$out): void
...
+                    dos: $service_date,
+                    code: $codekey,
+                    bgcolor: $bgcolor
+                ));
+                echo implode('', $oldDetails);
                 // Check for sanity in amount charged.
                 $prevchg = sprintf("%.2f", $prev['chg'] + ($prev['adj'] ?? null));
-                if ($prevchg != abs($svc['chg'])) {
-                    writeMessageLine(
+                if ($prevchg !== abs($svc['chg'])) {
```
stephenwaite: hi @kojiromike, on testing the strict string comparison here will now fail and the payments won't get posted,
`$prevchg` string shouldn't get the sprintf treatment :)

kojiromike: oh good catch. Thanks!

interface/billing/sl_eob_process.php
```diff
@@ -379,14 +451,10 @@ function era_callback(&$out): void
...
-                    if (in_array($svc['code'], $cpts ?? [])) {
-                        foreach ($cpts as $v) {
-                            if ($v == $codekey) {
-                                $codekey = $cpt . ':' . implode(':', $mods[$v] ?? []);
-                            }
-                        }
+            if (!$prev && !$svc['mod'] && in_array($svc['code'], $cpts)) {
+                foreach ($cpts as $v) {
+                    if ($v === $svc['code']) {
+                        $codekey = $v . ':' . implode(':', $mods[$v] ?? []);
```
stephenwaite, line 459: nice

interface/billing/sl_eob_process.php
kojiromike: @stephenwaite ^^

interface/billing/sl_eob_process.php
kojiromike: fixed…

interface/billing/sl_eob_process.php
stephenwaite: supposed to be appended, if `parseERAForCheck()` isn't run then `$InsertionId[]` doesn't get loaded

kojiromike: @stephenwaite ok, should they have a newline between them? what should go into the alert() that pops up?

stephenwaite: pretty sure that's all taken care of if you append it

if it isn't I'll play around with it in a later with some improvements that hope to make based on your work, thanks

stephenwaite (approved):

interface/billing/sl_eob_process.php
```diff
@@ -25,141 +25,200 @@
...
+        yield getDetailLine(
+            bgcolor: $bgcolor,
+            class: 'olddetail',
+            ptname: $ptname,
+            invnumber: $invnumber,
+            code: $code,
+            date: $ddate,
+            description: $description,
+            amount: $amount,
+            balance: $invoice_total
```
stephenwaite, line 201: named arguments 💯

PR #9138, opened 2025-10-12
refactor(php): use classname constant instead of string

### LINKED ISSUE #9137: `StringClassNameToClassConstantRector` - use `::class` instead of literal string name for classes

kojiromike:
Fixes #9137

Apply `StringClassNameToClassConstantRector` rule

- [x] chore(rector): set StringClassNameToClassConstantRector
- [x] refactor(php): use classname constant instead of string


interface/modules/custom_modules/oe-module-prior-authorizations/vendor/composer/autoload_classmap.php
```diff
@@ -6,5 +6,5 @@
 $baseDir = dirname($vendorDir);
 
 return [
-    'Composer\\InstalledVersions' => $vendorDir . '/composer/InstalledVersions.php',
+    \Composer\InstalledVersions::class => $vendorDir . '/composer/InstalledVersions.php',
```
bradymiller: these composer things don't seem right since class may not exist yet and pretty sure are auto-generated

bradymiller:
hi @kojiromike , just seems like some weirdness on the composer change (since class may not exist yet and autogenerated) and then the huge fhir changes (guessing autogenerated upstream). Guessing that module with composer issues can be sorted out and guessing the fhir stuff can simply be re-done by rector in future if replace via autogenerated code.

interface/modules/custom_modules/oe-module-prior-authorizations/vendor/composer/autoload_classmap.php
kojiromike: No fear, I learned that `::class` is magic and doesn't trigger the autoloader or actually care if the class exists at all.

```
php > echo fishy_gawumpus::class;
fishy_gawumpus
php > echo '#@($&@*(#&$-unlikely classname'::class;
#@($&@*(#&$-unlikely classname
```

That said, I'm skeptical that this autoloader does anything useful anyway. This module is the only one with a `vendor` directory and an autoloader, which I think is unused, because the openemr main autoloader is already handling it.

@adunsulag am I wrong here? I think this whole vendor directory can be safely deleted.

kojiromike: I created #9142 to track this as a separate thing. LMK in that ticket if I'm offbase.

library/classes/Installer.class.php
```diff
@@ -2065,7 +2065,7 @@ protected function createTotpInstance(string $secret, string $user): Totp
      */
     protected function cryptoGenClassExists(): bool
     {
-        return class_exists('OpenEMR\Common\Crypto\CryptoGen');
+        return class_exists(\OpenEMR\Common\Crypto\CryptoGen::class);
```
kojiromike, line 2068: @bradymiller if `::class` didn't do its "ignorance is bliss" magic, then this would be trouble, heh.

interface/modules/custom_modules/oe-module-prior-authorizations/vendor/composer/autoload_classmap.php
juggernautsei: @kojiromike when I orignaly buit this module. There were other forces at play and the vendor directory and autoloader was prior to the openemr.bootstrap autoloader functionality. The vendor folder can be removed. It has not functionality at this point and if needed in the future. It can just be put back.

PR #9146, opened 2025-10-13
feat(QueryUtils): transaction abstraction

### LINKED ISSUE #9110: Create a transaction abstraction for database queries
It seems that when running sql in a transaction, openemr usually does something like:

```php
sqlStatement("SET autocommit=0");
sqlStatement("START TRANSACTION");
// ... a bunch of individual sqlStatement or other sql commands
sqlStatement("COMMIT");
sqlStatement("SET autocommit=1");
```

Often, there aren't any `try/except` or other handling of errors, so it's not entirely obvious what happens to the transaction if there's a failure condition.

I think it would be better if QueryUtils or something had a function that took a callback and wrapped it in a transaction, ensuring the rollback and cleanup criteria are always run as needed.

kojiromike:
Fixes #9110

Create a utility for running queries in a transaction.

Create a wrapper for running a callable as a transaction.

### DIFF
--- a/library/classes/Installer.class.php
+++ b/library/classes/Installer.class.php
@@ -16,11 +16,13 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
-use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Common\Database\DatabaseQueryTrait;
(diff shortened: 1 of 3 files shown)


kojiromike:
closing in favor of #10177

PR #9166, opened 2025-10-15
chore(rector): reduce parallel load

### LINKED ISSUE #9165: rector jobs are timing out
We're seeing an uptick in `Child process timed out after 120 seconds` in rector. I hope we can fix it by applying the guidance on https://getrector.com/documentation/troubleshooting-parallel

kojiromike:
Fixes #9165

Reduce load on actions to prevent parallel job timeouts


kojiromike:
This only changes rector.php, so not going to wait for tests.

PR #9178, opened 2025-10-16
test(procedure_order): test different hl7 generations

kojiromike:

Generated code that tests `interface/forms/procedure_order/common.php`


kojiromike:
Superseded and split in two:

- #13544 — the fixture managers, brought up to current conventions. The `.json` fixtures here never actually loaded: since #11025, `BaseFixtureManager::getFixturesFromFile()` `require`s the path, and a `require` on JSON returns `1` rather than an array. Also fixes a double patient install that broke `form_encounter`'s pid subquery.
- #13545 — the tests, rewritten. #9207 gave the four generators distinct names and a uniform signature, which removed the function-name collision that forced the four `markTestSkipped` placeholders here. The replacement has 40 cases and no skips, including the LabCorp requisition assertions this PR had to comment out.

This PR can be closed in favor of those.

PR #9186, opened 2025-10-16
ci: drop path coverage

### LINKED ISSUE #9184: Major performance regression in covered Services tests
Uncovered, the services tests take 10 seconds.
Covered, they take ~30 minutes.

Even the E2e tests only take 8-10 minutes.

Time to figure out what is causing those tests to take such a big hit.

kojiromike:
Fixes #9184

Path coverage covers every possible path through the code, which is exponential, if more detailed. It's more than OpenEMR needs right now.

### DIFF
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -24,7 +24,6 @@ coverage_args=(
     --coverage-filter src
     --coverage-filter tests
     --coverage-text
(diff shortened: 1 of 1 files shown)


kojiromike:
That's better, 2m8s for services testing this round. Now just need to see if the coverage report is adequate.

PR #9207, opened 2025-10-20
fix(gen_hl7_order): consistent signature

### LINKED ISSUE #9206: Labcorp gen_hl7_order compatibility break
There are several implementations of `gen_hl7_order`, but the LabCorp one requires a parameter that the others don't. This breaks static analysis.

---

So this is a problem with how OpenEMR in the past has handled polymorphic functions.  In this case $reqStr is actually required in this function call as its used as a parameter in labcorp/gen_hl7_order.inc.php  but is NOT used in the default orders/gen_hl7_order.inc.php

_Originally posted by @adunsulag in https://github.com/openemr/openemr/pull/9126#discussion_r2436348515_

kojiromike:
Fixes #9206

Use a consistent function signature for all the `gen_hl7_order` implementations so that static analysis can work.

- [x] create a new `Hl7OrderResult` class. (Currently basically a struct, but these function implementations could be moved into it some day.) 
- [x] rewrite all the `gen_hl7_order` implementations to always take an int, return an `Hl7OrderResult`, and throw exceptions instead of taking an int and various references, and returning a string that indicates an error.
- [x] clean up the header docblocks to use consistent layout for author, copyright and license.

### DIFF
--- a/interface/forms/procedure_order/common.php
+++ b/interface/forms/procedure_order/common.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Encounter form for entering procedure orders.
+ * Encounter form for entering procedure orders
  *
  * @package   OpenEMR
  * @link      http://www.open-emr.org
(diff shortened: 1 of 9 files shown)


bradymiller (approved):
nice!

PR #9212, opened 2025-10-21
chore(ci/ciLibrary.source): bump phpcov to 11.0.2

kojiromike:


kojiromike:
While phpcov is not very well documented, it seems to be the case that while 11.0.0 works fine for us, 11.0.2 depends on phpunit 12, and phpunit 12 requires php 8.3+. So giving up on this for now.

bradymiller:
yep, seem like phpunit always has their most recent production version set to not work with the lowest supported php version :)

PR #9216, opened 2025-10-21
ci(test): make phpinfo available from testing

kojiromike:

Create a phpinfo.php file for the web api, and make it an artifact for easier observability during testing.

Create ci/phpinfo.php, use curl to fetch it, and upload it to Github during testing.

### DIFF
--- a/ci/phpinfo.php
+++ b/ci/phpinfo.php
@@ -1,0 +1,3 @@
+<?php
+
+phpinfo();
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -247,6 +247,17 @@ jobs:
         . ci/ciLibrary.source
         install_configure
 
+    - name: PHP info
+      run: |
+        curl -fsSL phpinfo.html http://localhost/ci/phpinfo.php | tee phpinfo.html
+
+    - name: Upload PHP info to GitHub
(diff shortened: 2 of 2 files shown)


ci/phpinfo.php
```diff
@@ -0,0 +1,3 @@
+<?php
```
bradymiller, line 1: hi @kojiromike , won't this allow anybody to look at this information for production instances?

bradymiller: to followup on this, couldn't this be done without needing to have this script?

kojiromike: @bradymiller interesting point, we should not allow access to ci/* from deployed instances.

kojiromike: @bradymiller to follow up on your followup: I don't want to use an SAPI=='cli' solution for this, because it might behave differently from the e2e webserver runtime. So I think phpinfo is the right way to go. But we could produce it dynamically from ci/ciLibrary.source.

bradymiller: yep, that's what i meant to create it dynamically. having phpinfo available would be considered a security issue since can get bunch of server info from it to see if vulnarable to stuff. everything else in ci is safe (ie. outside user can't do anything with it unless they want to set up their own ci :) ).

PR #9219, opened 2025-10-22
ci: create phpinfo dynamically

### LINKED ISSUE #9218: avoid having phpinfo in ci
yep, that's what i meant to create it dynamically. having phpinfo available would be considered a security issue since can get bunch of server info from it to see if vulnarable to stuff. everything else in ci is safe (ie. outside user can't do anything with it unless they want to set up their own ci :) ).

_Originally posted by @bradymiller in https://github.com/openemr/openemr/pull/9216#discussion_r2449700392_

kojiromike:
Fixes #9218

Guard ci/phpinfo.php and other ci php with an environment variable, preventing it from being accessed accidentally.

Add an environment variable guard to ci/phpinfo.php.
Add the environment variable to the docker compose files.

### DIFF
--- a/ci/compose-shared-apache.yml
+++ b/ci/compose-shared-apache.yml
@@ -6,8 +6,9 @@ services:
     volumes:
     - ../:/var/www/localhost/htdocs/openemr
     environment:
-      FORCE_NO_BUILD_MODE: "yes"
       EMPTY: "yes"
+      FORCE_NO_BUILD_MODE: "yes"
+      OPENEMR_E2E_ENABLE_CI_PHP: "1"
       SELENIUM_BASE_URL: "http://openemr"
     healthcheck:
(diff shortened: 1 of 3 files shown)


kojiromike:
@bradymiller after some thought I realized I need to put some other php in `ci`, so a guard makes more sense than dynamically generating all of it. There's also precedence, since we effectively call https://github.com/openemr/openemr/blob/06cbbae994d48ec183fcc83f5c604572ecf8ed04/contrib/util/installScripts/InstallerAuto.php#L78 during ci as well.

This may be a better way to go because I still need to introduce more php into `ci/` in #9217.

bradymiller (approved):

bradymiller:
hi @kojiromike  , looks great!

PR #9220, opened 2025-10-22
ci(test): break into reusable workflow

### LINKED ISSUE #9222: Create individually-runnable ci test jobs
Currently test.yml is a monolith that iterates over ci directories to create configurations for running lots of different test configurations in a matrix. When working to improve the tests, I frequently create a bunch of wasteful test runs in duplicate. I'd like to be able to trigger a single test with a custom configuration without having to necessarily push code.

To that end, I propose to move the test code to its own dispatchable workflow, and call it. The end result will be much the same as it is today, except we'll also be able to run a single test on its own, if we want.

kojiromike:
Fixes #9222

Break up test.yml into a reusable, dispatchable workflow, so that we can run or re-run individual tests when needed.

I modified test.yml so that it uses workflow call/dispatch to run a single test. To call it, I broke up collect into two parts: one to collect the ci docker dirs, and the other to parse the config from that. I put the parse config step into the test.yml, and then created test-all.yml to collect the docker dirs and call the individual test.ymls for each one.

So test-all.yml is called for pull requests, and will behave as usual, with some slight display differences.
But if we need to debug a single test configuration, we can run test.yml with workflow dispatch with much lower overhead.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,160 +1,86 @@
 ##
-# OpenEMR has two different kinds of tests:
-# - "Isolated tests": Tests that run without secondary services, a data layer or significant initialization requirements
-# - "Tests": Tests that require a database and initialization before they can run or pass.
-# This workflow runs the kind that requires initialization.
+# Reusable workflow for running OpenEMR tests
+# Accepts a docker directory name and parses all configuration from it
 
 name: Test
 
+run-name: ${{ inputs.display_name || inputs.docker_dir }}
+
 on:
-  push:
-    branches:
-    - master
-    - rel-*
-  pull_request:
-    branches:
-    - master
-    - rel-*
+  workflow_call:
+    inputs:
+      docker_dir:
+        required: true
+        type: string
+        description: 'Docker directory name (e.g., apache_84_114)'
+      display_name:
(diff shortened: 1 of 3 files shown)


ci/parse_docker_dir.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
kojiromike: Sometimes copilot says things that are pure nonsense. This is one of them.

.github/workflows/test.yml
```diff
@@ -1,160 +1,86 @@
...
+      docker_dir:
+        description: 'Docker directory name'
+        required: true
+        type: choice
+        options:
+          - apache_84_114
+          - apache_84_57
+          - apache_83_114
+          - nginx_84_114
+          - nginx_83_114
```
bradymiller, line 47: hi @kojiromike , Just trying to figure out what the hard-coded options above are for.

kojiromike: These create a dropdown so it's easy to pick from. I also don't love the fact that they're hard-coded, but not sure if there's a better way to make a picker.

PR #9231, opened 2025-10-23
ci(test): more aggressive caching

### LINKED ISSUE #9239: Cache more to avoid repetitive steps in CI testing
We can improve the speed of our ci tests by caching more. We can cache the vendor directory itself, as well as the outputs of npm ci/build and npm ci in ccdaservice.

kojiromike:
Fixes #9239 

Use caching to speed up tests.

Cache vendor, node_modules, public/... directories directly instead of just caching the node and composer downloads

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -232,7 +232,9 @@ jobs:
       id: ccda-build-cache-restore
       uses: actions/cache/restore@v4
       with:
-        path: ccdaservice/node_modules/
+        path: |
+          ccdaservice/node_modules/
+          ccdaservice/packages/oe-cqm-service/node_modules/
         key: ${{ steps.cache-keys.outputs.ccda-build }}
         restore-keys: |
           ${{ runner.os }}-ccda-build-${{ steps.parse.outputs.node_version }}-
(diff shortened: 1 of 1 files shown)


kojiromike:
This is the strangest thing. I am trying an experiment in caching not just the package manager downloads, but the `vendor` and build results as well. All the tests pass, but this E2e one:

```
 ✘ Main menu link with data set "Modules -> Carecoordination menu link"
   │
   │ [Care Coordination] tab load FAILED
   │ Failed asserting that two strings are identical.
   │ --- Expected
   │ +++ Actual
   │ @@ @@
   │ -'Care Coordination'
   │ +'CCDA'
   │
   │ /var/www/localhost/htdocs/openemr/tests/Tests/E2e/Base/BaseTrait.php:120
   │ /var/www/localhost/htdocs/openemr/tests/Tests/E2e/HhMainMenuLinksTest.php:62
   │
```

I can't quite see why caching would affect this.

kojiromike:
> I can't quite see why caching would affect this.


The problem was that I missed capturing a subdirectory of ccdaservice in the cache.

kojiromike:
@openemr/admins It may not seem too big, but this is shaving 1-2 minutes off most every test job. WDYT?

kojiromike:
I suppose it's worth mentioning; currently all our tests are only running on Linux. If we ever wanted to test on mac or win, we'd have to expand the caching method to allow for the same php or node version to be cached with different platforms.

kojiromike:
This makes testing faster and doesn't impact runtime. I think it's safe to merge. Please let me know if you need me to roll anything back.

bradymiller:
neat stuff!

PR #9232, opened 2025-10-23
API - Swagger docs - Freshness checking

### LINKED ISSUE #9223: [API][Docs][CI] Check docs freshness by CI for PRs
### Describe the problem

I noticed that `openemr-api.yaml` can be outdated.

So we should check automatically that docs regenerated for each PR (and in sync with codebase / new API changes)

igormukhingmailcom:
Fixes #9223
Related: https://github.com/openemr/openemr/pull/9197

### When `openemr-api.yml` is in sync with codebase

image

### When `openemr-api.yml` is not in sync with codebase - job fail

image

Fails when docs not refreshed after code changes.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -1,0 +1,65 @@
+name: API Docs Freshness Checks
+
+on:
+  pull_request:
+    branches:
+    - master
+    - rel-*
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        php-version: ['8.2']
+    name: API Docs Freshness Check
+    steps:
+    - name: "Checkout Code"
+      uses: actions/checkout@v5
+      with:
+        fetch-depth: 0
+
+    - name: "Setup PHP"
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php-version }}
+        coverage: none
+
+    - name: "Get composer cache directory"
+      id: composer-cache
+      run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
+    - name: "Composer Cache"
+      uses: actions/cache@v4
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-composer-${{ matrix.php-version }}-
+          ${{ runner.os }}-composer-
+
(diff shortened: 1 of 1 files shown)


kojiromike:
See #9351

igormukhingmailcom:
@kojiromike Thanks for that fix :)
Now this PR is ready for merging

swagger/openemr-api.yaml
```diff
@@ -5932,6 +6048,141 @@ paths:
...
+                  whenHandedOver: '2021-09-18T00:00:00+00:00'
+        400:
+          $ref: '#/components/responses/badrequest'
+        401:
+          $ref: '#/components/responses/unauthorized'
+        404:
+          $ref: '#/components/responses/uuidnotfound'
+      security:
+        -
+          openemr_auth: {  }
```
kojiromike, line 6185: I don't understand why there's so much whitespace here.

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,61 @@
...
+      run: |
+        if [[ -n $(git diff --name-only) ]]; then
+          echo "Looks like openemr-api.yaml is not in sync with your changes."
+          echo "Please, run:"
+          echo "php bin/console openemr:create-api-documentation --skip-globals"
+          echo "and push updated openemr-api.yaml file into your PR"
+          exit 1
+        else
+          echo "openemr-api.yaml is in sync with your changes"
+        fi
```
kojiromike:
```suggestion
        if git diff --quiet; then
          echo 'openemr-api.yaml is in sync with your changes'
          exit 0
        fi
        echo "Looks like openemr-api.yaml is not in sync with your changes."
        echo "Please, run:"
        echo "php bin/console openemr:create-api-documentation --skip-globals"
        echo "and push updated openemr-api.yaml file into your PR"
        exit 1
```

swagger/openemr-api.yaml
igormukhingmailcom: Looks good to me. What exactly wrong?

kojiromike: @igormukhingmailcom there's a `-` for a list entry, then a newline and a bunch more indentation, then `{` for object entry, then a bunch of space, then `}`. IMO it should be:

```yaml
          - openemr_auth: {}
```

igormukhingmailcom: The only issue here is that array (`[]`) somehow converted to object (`{}`) during normalization, but http://localhost:8300/swagger/ still renders well even with that issue.

New lines are not an issue - it was like that before (normalizer was introduced) as well. Just the way formatter formats things to have same level keys aligned.

See diff at https://github.com/openemr/openemr/pull/9197/files when normalization was introduced:

image

PR #9243, opened 2025-10-24
ci(inferno): fix missing main_build

### LINKED ISSUE #9239: Cache more to avoid repetitive steps in CI testing

kojiromike:
Fixes #9239

`main_build` was removed in #9239 and that broke the inferno tests. This fixes that.

### DIFF
--- a/ci/inferno/run.sh
+++ b/ci/inferno/run.sh
@@ -70,37 +70,35 @@ initialize_openemr() {
     echo 'Initializing OpenEMR'
     local -x DOCKER_DIR=inferno
     local -x OPENEMR_DIR=/var/www/localhost/htdocs/openemr
-    (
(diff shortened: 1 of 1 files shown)


ci/inferno/run.sh
```diff
@@ -70,35 +70,35 @@ initialize_openemr() {
     echo 'Initializing OpenEMR'
     local -x DOCKER_DIR=inferno
     local -x OPENEMR_DIR=/var/www/localhost/htdocs/openemr
-    (
```
kojiromike, line 73: @adunsulag These subshells were just for managing the directory, right? They're triggering a shellcheck issue that's a little obnoxious, and I think the easiest fix is to remove the subshell. WDYT?

PR #9271, opened 2025-10-28
ci(composer-require-checker): add

### LINKED ISSUE #9113: [Tools] Composer require checker
### Describe the problem

The project currently relies on manual review or runtime errors to detect missing or unused Composer dependencies.

As the codebase grows, it becomes easy to accidentally use packages not explicitly declared in composer.json, or keep unused dependencies that increase maintenance burden.

This can lead to broken builds, unexpected runtime errors, and bloated dependency trees.

Integrate https://github.com/maglnet/ComposerRequireChecker into the project’s development tooling to automatically detect missing or incorrect Composer dependencies.

ComposerRequireChecker analyzes PHP code to ensure that all used symbols come from properly declared dependencies. This helps maintain a clean and accurate composer.json file.

### Suggested implementation
- Add as a dev dependency `composer require --dev maglnet/composer-require-checker`
- Add a configuration file (`composer-require-checker.json`) specifying ignored symbols or allowed extensions if needed
- Add a Composer script, e.g.:
```json
"scripts": {
    "check-requirements": "composer-require-checker check composer.json"
}
```
- Integrate into CI to automatically validate dependency correctness on pull requests

### Benefits

- Prevents missing dependency declarations
- Helps identify unused packages
- Keeps dependency management clean and reliable
- Reduces runtime dependency errors

kojiromike:
Fixes #9113

Add composer-require-checker to CI

Add composer-require-checker to CI


.composer-require-checker.json
```diff
@@ -0,0 +1,187 @@
...
+        "Symfony\\Component\\HttpKernel\\Controller\\ControllerResolver",
+        "Symfony\\Component\\HttpKernel\\Controller\\ControllerResolverInterface",
+        "Symfony\\Component\\HttpKernel\\Event\\ExceptionEvent",
+        "Symfony\\Component\\HttpKernel\\Event\\RequestEvent",
+        "Symfony\\Component\\HttpKernel\\Event\\TerminateEvent",
+        "Symfony\\Component\\HttpKernel\\Event\\ViewEvent",
+        "Symfony\\Component\\HttpKernel\\Exception\\HttpException",
+        "Symfony\\Component\\HttpKernel\\Exception\\NotFoundHttpException",
+        "Symfony\\Component\\HttpKernel\\Exception\\UnauthorizedHttpException",
+        "Symfony\\Component\\HttpKernel\\HttpKernel",
```
kojiromike, line 104: @stephenwaite To get this to pass I'm ignoring all the current issues, but I think this one is #9272

composer.json
```diff
@@ -37,6 +41,7 @@
         "ext-xmlreader": "*",
         "ext-xmlwriter": "*",
         "ext-xsl": "*",
+        "ext-zend-opcache": "*",
```
bradymiller, line 44: it's weird that these don't break builds in easy dev (ie. and all based alpine dockers) since pretty sure not all of these new extensions are installed. Even if did, not end of world; would just need to update them, which I will plan to do anyways at some point after research them all a bit.

bradymiller: to answer my own question, makes sense that build is fine since assigning them in the platform. Still going to research them and ensure they are included in dockers.

.github/workflows/composer-require-checker.yml
```diff
@@ -0,0 +1,130 @@
...
+        - '4.16.1'
+    steps:
+    - name: "Checkout Code"
+      uses: actions/checkout@v5
+
+    - name: "Setup PHP"
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php-version }}
+        extensions: calendar, ctype, curl, dom, fileinfo, filter, gd, iconv, imagick, intl, json, ldap, libxml, mbstring, mysqli, opcache, openssl, pdo, pdo_mysql, phar, session, simplexml, soap, sockets, sodium, tokenizer, xml, xmlreader, xmlwriter, xsl, zip, zlib
```
bradymiller, line 46: willing to bet this line is not needed since the platform stuff in composer.json already simulates their existence on the build

composer.json
bradymiller: here are some notes on these added php extensions (ongoing research):
- ext-dom (php84-dom alpine package will be installed in dockers)
- ext-filter (this has been in php core for a long time (per Claude since PHP 5.2))
- ext-imagick (this is an experimental extension that is only used in src/Services/ImageUtilities/HandleImageService.php (for faxsms module); not available package in alpine. will need to use pecl to bring it in and build it, which is complicated) (may need to not force this since only for certain cases where using the faxsms module and those installs could then install the extension if needed) (note imagemagick is used in other places in OpenEMR, but is doing command line calls rather than using imagick extension/class calls)
- ext-libxml (this has been in php core for a long time (per Claude since PHP 5.1))
- ext-zend-opcache (this is the opcache extension; is already installed in the docker and unclear why required extension in codebase) Only used during initial installing if it installed/enabled (code block below) so it does not break install if it is installed/enabled (ie. would not consider it to be needed in this case):
 ```
        // Tell PHP that its cached bytecode version of sqlconf.php is no longer usable.
        // @codeCoverageIgnoreStart
        if (function_exists('opcache_invalidate')) {
            opcache_invalidate($this->conffile, true);
        }
        // @codeCoverageIgnoreEnd
```

kojiromike: I can whitelist the `opcache_invalidate` symbol so composer-require-checker ignores it. Would you like me to whitelist imagick, too?

As for the other extensions, I think there's no harm in keeping a comprehensive list in composer.json. WDYT?

bradymiller: agree with whitelist of opcache_invalidate to drop the opcache requirement. I found a php alpine package with imagick (it is a pecl thing), so plan to bring that into dockers along with ext-dom (so can keep imagick as a dependency :) ).

~~The only thing left are those 2 core ones, ext-filter and ext-libxml; still gonna do some more research on these (and what core exactly means (ie. is it dependent on OS etc)), agree that if there isn't complete clarity on them being a solid part of core php functionality, that they should stay in the dependency listings.~~

bradymiller: After researching a bit more, agree to keep ext-filter and ext-libxml in composer.json :)

.github/workflows/composer-require-checker.yml
kojiromike: I had an interesting experience yesterday that led to https://github.com/openemr/openemr/issues/9308. But basically I'm experimenting with a "meta package" install of openemr, so I need to composer install openemr itself. But I didn't have imagick or intl on my mac, so composer balked. So maybe I'm not understanding you, but I think these are needed.

PR #9296, opened 2025-10-29
fix(missing dependencies): PSR HTTP interfaces

### LINKED ISSUE #9283: Add explicit dependencies for PSR HTTP interfaces
## Problem

The codebase directly uses **13 PSR HTTP interfaces** that are not explicitly declared in `composer.json`:

### PSR-7 HTTP Message Interfaces (12)
- `Psr\Http\Message\RequestFactoryInterface`
- `Psr\Http\Message\RequestInterface`
- `Psr\Http\Message\ResponseFactoryInterface`
- `Psr\Http\Message\ResponseInterface`
- `Psr\Http\Message\ServerRequestFactoryInterface`
- `Psr\Http\Message\ServerRequestInterface`
- `Psr\Http\Message\StreamFactoryInterface`
- `Psr\Http\Message\StreamInterface`
- `Psr\Http\Message\UploadedFileFactoryInterface`
- `Psr\Http\Message\UploadedFileInterface`
- `Psr\Http\Message\UriFactoryInterface`
- `Psr\Http\Message\UriInterface`

### PSR-18 HTTP Client Interface (1)
- `Psr\Http\Client\ClientInterface`

## Analysis

These interfaces are currently provided transitively by:
- `symfony/psr-http-message-bridge` → depends on `psr/http-message`
- `nyholm/psr7` → implements PSR-7
- `guzzlehttp/guzzle` → depends on `psr/http-client`

However, following best practices, packages that directly use these interfaces should explicitly declare them as dependencies.

## Solution

Add explicit dependencies to `composer.json`:

```json
"psr/http-message": "^1.1|^2.0",
"psr/http-client": "^1.0",
"psr/http-factory": "^1.0"
```

## Tasks

- [x] Add PSR interface packages to composer.json
- [x] Run `composer update` to verify compatibility
- [x] Remove these 13 symbols from composer-require-checker whitelist
- [x] Verify no conflicts with existing implementations

kojiromike:
Fixes #9283

Some PSR HTTP interfaces are being used, but not directly included in composer.json

Add the missing deps to composer.json


composer.json
```diff
@@ -91,6 +91,9 @@
         "phpoffice/phpspreadsheet": "5.2.0",
         "phpseclib/phpseclib": "3.0.47",
         "predis/predis": "3.2.0",
+        "psr/http-client": "^1.0",
+        "psr/http-factory": "^1.1",
+        "psr/http-message": "^1.1",
```
bradymiller, line 96: hi @kojiromike , our practice in composer.json has been to pin versions. We've done this to avoid inadvertently increasing package versions. What are you thoughts on this? I think we can be more open to this since now have dependabot doing update, but will need to ensure devs don't upgrade by completely dropping the lock file and rebuilding (i generally do this on updates for lots of packages, but again now that dependabot is taking care of that, should not be issue for my workflow).

bradymiller: btw, one of the worst cases would be if a version is inadvertently brought in that is incompatible with a lower php version. I generally check all new versions on packagist to ensure this is not the case.

bradymiller: also noted that pinning version also still opens door to dependabot increasing the versions.

kojiromike: Strict pinning can lead to really brittle installation problems, where it's difficult to update anything because there's no flexibility in the system for constraints. In that principle, I am in favor of a little bit of stretch in the pins. If you combine this with the github feature of being able to get a "rich view" of composer.lock, you should be able to tell pretty easily if the updates are concise or completely new.

So in my opinion, we should not be pinning to exact versions.

bradymiller: Hi @kojiromike (sorry in advance for potentially dumb questions :) ), 
1. Is it accurate to state (since going away from pinning) that we should never delete the lock file and never do a broad (ie. only do composer update on specific packages) composer update anymore?
2. In  cases where doing something like going to a new symfony long term version in future, I am assuming this means needs to specifically update each package.
3. If this is the case (ie. never remove lock file or do broad composer update), is there a chance that dependencies (ie. things that are not explicitly set in composer.json) may go stale (or does dependabot update these also)?

kojiromike: > Is it accurate to state (since going away from pinning) that we should never delete the lock file and never do a broad (ie. only do composer update on specific packages) composer update anymore?

Never say never, but it shouldn't be necessary. It should not be a normal practice, and should only be used in unusual cases, such as if we have to manually work through a dependency gridlock, or are doing a major refactor.

> In cases where doing something like going to a new symfony long term version in future, I am assuming this means needs to specifically update each package.

It's likely we'll need to update all the symfony dependencies at the same time across the major boundary. You can always script it, something like

```
< composer.json jq --arg newversion v7.2.0 -r '
    .require |
    to_entries[] |
    select(.key != "symfony/event-dispatcher-contracts" and (.key | startswith("symfony/"))) |
    "\(.key):\($newversion)"
'
```

> If this is the case (ie. never remove lock file or do broad composer update), is there a chance that dependencies (ie. things that are not explicitly set in composer.json) may go stale (or does dependabot update these also)?

Dependabot will update all dependencies, as long as they are in the .github/dependabot.yml file, and not being ignored there. IMO this is the right way to do it, because it alerts us to one dependency bump at a time, and automation should test the change for us.

bradymiller: thanks for the info. super helpful.

Issue #8251, opened 2025-04-10
feat: upgrade zircote/swagger-php 5.0.7 (currently stuck at 3.3.7)


adunsulag:
### Describe the problem

upgrade zircote/swagger-php 5.0.7 (currently stuck at 3.3.7)

Need to make sure as part of the upgrade that the swagger documentation is updated and tested.

PR #9303, opened 2025-10-29
fix(missing dependencies): zircote/swagger-php

### LINKED ISSUE #9289: Resolve zircote/swagger-php dev dependency used in production code
**Parent Issue:** #9278

## Problem

The codebase uses `OpenApi\Generator` from the `zircote/swagger-php` package, which is currently declared in `require-dev` but appears to be used in production code.

## Analysis

**Current status:** Package is in `require-dev` section of composer.json
**Issue:** Dev dependencies are not available in production environments

This creates a potential runtime error if:
- OpenAPI generation happens at runtime in production
- The class is referenced without checking for its existence first

## Solution

Choose one of the following approaches:

### Option A: Move to Production Dependencies
If OpenAPI generation happens at runtime, move the package:

```json
"require": {
  "zircote/swagger-php": "^3.0"
}
```

### Option B: Make Code Defensive
If only used in dev/testing environments, ensure code checks for class existence:

```php
if (class_exists('OpenApi\Generator')) {
    // OpenAPI generation code
}
```

## Tasks

- [ ] Search codebase for `OpenApi\Generator` usage
- [ ] Determine if used at runtime in production or only in dev/testing
- [ ] If production: Move `zircote/swagger-php` to `require` section
- [ ] If dev only: Add class_exists() checks or ensure it's only in dev contexts
- [ ] Remove this symbol from composer-require-checker whitelist

## Search Commands

```bash
grep -r "OpenApi\\\\Generator\|use OpenApi" --include="*.php"
```

---

**Title:** Resolve zircote/swagger-php dev dependency used in production code

kojiromike:
Fixes #9289

Add missing dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -119,7 +119,8 @@
         "twilio/sdk": "8.8.4",
         "vlucas/phpdotenv": "5.6.2",
         "waryway/php-traits-library": "1.0.4",
-        "yubico/u2flib-server": "1.0.2"
+        "yubico/u2flib-server": "1.0.2",
+        "zircote/swagger-php": "^3.0"
     },
     "require-dev": {
         "ergebnis/composer-normalize": "^2.48",
@@ -130,8 +131,7 @@
         "phpunit/phpunit": "^11.0",
         "rector/rector": "^2.1",
         "squizlabs/php_codesniffer": "^4.0",
-        "symfony/panther": "^2.0",
-        "zircote/swagger-php": "^3.0"
+        "symfony/panther": "^2.0"
     },
     "repositories": [
         {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -119,7 +119,8 @@
         "twilio/sdk": "8.8.4",
         "vlucas/phpdotenv": "5.6.2",
         "waryway/php-traits-library": "1.0.4",
-        "yubico/u2flib-server": "1.0.2"
+        "yubico/u2flib-server": "1.0.2",
+        "zircote/swagger-php": "^5.5"
```
bradymiller: this will break :)
https://github.com/openemr/openemr/issues/8251

bradymiller: (fyi, more context here; it will break the /swagger ui stuff)

kojiromike: fixed

Issue #9248, opened 2025-10-26
feat: US Core RelatedPerson Profile


adunsulag:
### Describe the problem

Implement the US Core 8.0.0 RelatedPerson Profile for FHIR

### Describe the solution you'd like
- [x] Add ability to create related persons to a patient from the OpenEMR UX
- - [x] Implement RelatedPerson in demographics
- - [x] Implement RelatedPerson connection in CareTeam
- [x] Implement FHIR US Core Endpoints
- [x] Implement RelatedPerson service

adunsulag:
Jerry has implemented the UX for RelatedPerson but a number of improvements are currently being done by @psoas that will make it easier for us to link the related person inside the CareTeam functionality.

PR #9333, opened 2025-11-02
Feat openemr #9248 related persons

adunsulag:
Feat #9248 

Working on the related persons.  Got everything rebased and data now displays properly.  Still a number of things outstanding.  Will also add in CareTeam to this since CareTeam needs to point to a related person

- [ ] Save Telecom and Address information for Related Person
- [ ] Fix Address save for Patient contact information
- [ ] Fix Telecom save for Patient contact information (optional)
- [ ] Fix new_comprehensive_save.php to handle all the new additions for RelatedPerson saving
- [ ] Implement CareTeam linkage to RelatedPerson

@bradymiller I wanted you to see the DB schema here.  I drop the old RelatedPerson schema and handle a migration since its been long enough in Master I think that'd be an important thing to have.

Note most of this work has been done by @psoas but I've gone in and fixed a number of things in order to make it so we can move forward with FHIR.

### DIFF
--- a/templates/patient/demographics/telecom_display.html.twig
+++ b/templates/patient/demographics/telecom_display.html.twig
@@ -96,14 +96,14 @@
 
 <script type="text/javascript">
     function toggleInactiveTelecomsDisplay(tableId) {
-        const checkbox = document.getElementById('show_inactive_telecoms_display_' + tableId);
-        const table = document.getElementById(tableId);
+        const table = document.getElementById(tableId);        
 
         if (!table) {
             console.error('Telecom table not found:', tableId);
             return;
         }
 
+        const checkbox = document.getElementById('show_inactive_telecoms_display_' + tableId);
         const inactiveTelecoms = table.querySelectorAll('.telecom-inactive-display');
 
         console.log('Toggle inactive telecoms:', {
--- a/library/templates/relation_display.php
+++ b/library/templates/relation_display.php
@@ -83,7 +83,7 @@ try {
 
             if ($record['person_id']) {
                 // Get addresses for this person's contact
-                $addressRecords = $addressService->getAddressesForContact($targetContactId, false);
+                $addressRecords = $addressService->getAddressesForContact($targetContactId, true);
                 foreach ($addressRecords as $addr) {
                     $relatedPerson['addresses'][] = [
                         'contact_address_id' => $addr['contact_address_id'] ?? $addr['id'],
@@ -102,7 +102,7 @@ try {
                 }
 
                 // Get telecoms for this person's contact
-                $telecomRecords = $telecomService->getTelecomsForContact($targetContactId, false);
+                $telecomRecords = $telecomService->getTelecomsForContact($targetContactId, true);
                 foreach ($telecomRecords as $telecom) {
                     $relatedPerson['telecoms'][] = [
                         'contact_id' => $targetContactId,
@@ -111,7 +111,7 @@ try {
                         'use' => $telecom['use'] ?? 'home',
                         'value' => $telecom['value'] ?? '',
                         'rank' => $telecom['rank'] ?? 1,
-                        'status' => $tel['status'] ?? 'A',
+                        'status' => $telecom['status'] ?? 'A',
                         'is_primary' => $tel['is_primary'] ?? 'N',
                         'notes' => $tel['notes'] ?? ''
                     ];
--- a/library/templates/relation_form.php
+++ b/library/templates/relation_form.php
@@ -97,11 +97,12 @@ try {
 
             if ($record['person_id']) {
                 // Get addresses for this person's contact
-                $addressRecords = $addressService->getAddressesForContact($targetContactId, false);
+                $addressRecords = $addressService->getAddressesForContact($targetContactId, true);
                 foreach ($addressRecords as $addr) {
                     $relatedPerson['addresses'][] = [
                         'contact_id' => $targetContactId,
                         'contact_address_id' => $addr['contact_address_id'] ?? $addr['id'],
+                        'addresses_id' => $addr['addresses_id'],
                         'use' => $addr['use'] ?? 'home',
                         'type' => $addr['type'] ?? 'postal',
                         'line1' => $addr['line1'] ?? '',
@@ -117,11 +118,11 @@ try {
                 }
 
                 // Get telecoms for this person's contact
-                $telecomRecords = $telecomService->getTelecomsForContact($targetContactId, false);
+                $telecomRecords = $telecomService->getTelecomsForContact($targetContactId, true);
                 foreach ($telecomRecords as $telecom) {
                     $relatedPerson['telecoms'][] = [
+                        'id' => $telecom['contact_telecom_id'] ?? $telecom['id'],
                         'contact_id' => $targetContactId,
-                        'contact_telecom_id' => $telecom['contact_telecom_id'] ?? $telecom['id'],
                         'system' => $telecom['system'] ?? 'phone',
                         'use' => $telecom['use'] ?? 'home',
                         'value' => $telecom['value'] ?? '',
--- a/templates/patient/demographics/address_form.html.twig
+++ b/templates/patient/demographics/address_form.html.twig
@@ -185,7 +185,33 @@
 </template>
 
 <script type="text/javascript">
+
+    // IE Compatibility Polyfills - Run immediately
     (function() {
+        if (!Element.prototype.matches) {
+            Element.prototype.matches = Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
+        }
+        if (!Element.prototype.closest) {
+            Element.prototype.closest = function(s) {
+                var el = this;
+                do {
+                    if (el.matches(s)) return el;
+                    el = el.parentElement || el.parentNode;
+                } while (el !== null && el.nodeType === 1);
+                return null;
+            };
+        }
+    })();
+
+    document.addEventListener('DOMContentLoaded', function() {
+        const tableId = {{ table_id|js_escape|raw }}
+        const table = document.getElementById(tableId);
+        if (!table) {
+            console.error('Table not found:', tableId);
+            return;
+        }
+        const container = table.querySelector('.table_edit_addresses')
+
         let ADDRESS_ACTION_VALUES = {
             'INACTIVATE': 'INACTIVATE',
             'ADD': 'ADD',
@@ -196,57 +222,26 @@
 
         {{ DateToYYYYMMDD_js()|raw }}
 
-        let tableConfig = {
-            tableId: {{ table_id|js_escape|raw }},
-            listAddressTypes: {{ list_address_types|json_encode|raw }},
-            listAddressUses: {{ list_address_uses|js_escape|raw }}
-        };
+        init(tableId);
 
-        (function() {
-            init(tableConfig.tableId);
-        })();
 
         function init(tableId) {
-            var tableElement = document.getElementById(tableId);
-            if (!tableElement) {
-                console.error('Address table element not found:', tableId);
-                return;
-            }
-            
-            // Mark as initialized
-            tableElement.dataset.addressInitialized = 'true';
             console.log('Initializing address table:', tableId);
 
-            // IE Compatibility
-            if (!Element.prototype.matches) {
-                Element.prototype.matches = Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
-            }
-
-            if (!Element.prototype.closest) {
-                Element.prototype.closest = function(s) {
-                    var el = this;
-                    do {
-                        if (Element.prototype.matches.call(el, s)) return el;
-                        el = el.parentElement || el.parentNode;
-                    } while (el !== null && el.nodeType === 1);
-                    return null;
-                };
-            }
-
             // Event listeners for form changes
-            tableElement.addEventListener('keyup', changeEdit);
-            tableElement.addEventListener('mouseup', changeEdit);
-            tableElement.addEventListener('touchend', changeEdit);
+            table.addEventListener('keyup', changeEdit);
+            table.addEventListener('mouseup', changeEdit);
+            table.addEventListener('touchend', changeEdit);
 
-            tableElement.addEventListener('input', addressUpdated);
-            tableElement.addEventListener('change', addressUpdated);
+            table.addEventListener('input', addressUpdated);
+            table.addEventListener('change', addressUpdated);
 
             // Setup existing address buttons and datepickers
-            tableElement.querySelectorAll('.addresses_group .display_addresses').forEach(function(row) {
+            table.querySelectorAll('.addresses_group .display_addresses').forEach(function(row) {
                 setupAddressesRowButtonEventListeners(row);
             });
 
-            tableElement.querySelectorAll('.form_addresses').forEach(function(row) {
+            table.querySelectorAll('.form_addresses').forEach(function(row) {
                 setupContainerDatePickers(row);
                 setupContainerListAddButtons(row);
             });
@@ -254,10 +249,7 @@
             // Setup show inactive checkbox
             var showInactiveCheckbox = document.getElementById('show_inactive_addresses_' + tableId);
             if (showInactiveCheckbox) {
-                var newCheckbox = showInactiveCheckbox.cloneNode(true);
-                showInactiveCheckbox.parentNode.replaceChild(newCheckbox, showInactiveCheckbox);
-                
-                newCheckbox.addEventListener('change', function() {
+                showInactiveCheckbox.addEventListener('change', function() {
                     toggleInactiveAddresses(tableId);
                 });
             }
@@ -265,15 +257,12 @@
             // Setup add button with event delegation
             var addButton = document.getElementById('btn_add_addresses_' + tableId);
             if (addButton) {
-                console.log('Found add button:', addButton.id);
-                var newButton = addButton.cloneNode(true);
-                addButton.parentNode.replaceChild(newButton, addButton);
-                
-                newButton.addEventListener('click', function(event) {
+                console.log('Setting up add button:', addButton.id);
+                addButton.addEventListener('click', function(event) {
                     console.log('Add address button clicked:', tableId);
-                    addAddress(event);
                     event.preventDefault();
-                    return false;
+                    event.stopPropagation();
+                    addAddress(event);
                 });
             } else {
                 console.warn('Add address button not found for table:', tableId);
@@ -282,7 +271,7 @@
 
         function toggleInactiveAddresses(tableId) {
             const checkbox = document.getElementById('show_inactive_addresses_' + tableId);
-            const inactiveAddresses = document.querySelectorAll('#' + tableId + ' .address-inactive');
+            const inactiveAddresses = table.querySelectorAll('.address-inactive');
 
             inactiveAddresses.forEach(function(address) {
                 if (checkbox.checked) {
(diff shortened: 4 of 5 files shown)


interface/patient_file/summary/person_search_ajax.php
```diff
@@ -0,0 +1,696 @@
...
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+// CRITICAL: Set error reporting for debugging
+error_reporting(E_ALL);
+ini_set('display_errors', '0');
+ini_set('log_errors', '1');
```
bradymiller: remove above at some point

library/templates/address_display.php
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ *
+ * @author    David Eschelbacher <psoas@tampabay.rr.com>
+ * @author    Stephen Nielson <snielson@discoverandchange.com>
+ * @copyright Copyright (c) 2022 David Eschelbacher <psoas@tampabay.rr.com>
+ * @copyright Copyright (c) 2022 Discover and Change, Inc. <snielson@discoverandchange.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
```
bradymiller, line 15: odd to have a script that runs without globals
(if case, would at least ensure it matches some sort of variable from the calling script (assuming this is a template); would also consider this for any other scripts in this PR that do the same thing)

library/templates/address_save_handler.php
```diff
@@ -0,0 +1,288 @@
...
+use OpenEMR\Common\ORDataObject\ContactAddress;
+use OpenEMR\Common\ORDataObject\Address;
+
+/**
+ * Save addresses for a patient from form data
+ *
+ * @param int $pid Patient ID
+ * @param array $addressData Address form data
+ * @return array Array of saved ContactAddress records
+ */
```
bradymiller, line 30: seems like this belongs in /library

sql/database.sql
```diff
@@ -1182,16 +1182,122 @@ DROP TABLE IF EXISTS `contact_address`;
...
+    `marital_status` VARCHAR(31) DEFAULT NULL,
+    `race` VARCHAR(63) DEFAULT NULL,
+    `ethnicity` VARCHAR(63) DEFAULT NULL,
+    `preferred_language` VARCHAR(63) DEFAULT NULL COMMENT 'ISO 639-1 code',
+    `communication` VARCHAR(254) DEFAULT NULL COMMENT 'Communication preferences/needs',
+    `ssn` VARCHAR(31) DEFAULT NULL COMMENT 'Should be encrypted in application',
+    `active` TINYINT(1) DEFAULT 1 COMMENT '1=active, 0=inactive',
+    `inactive_reason` VARCHAR(255) DEFAULT NULL,
+    `inactive_date` DATETIME DEFAULT NULL,
+    `notes` TEXT DEFAULT NULL,
```
bradymiller: make TEXT bare

sql/database.sql
```diff
@@ -1182,16 +1182,122 @@ DROP TABLE IF EXISTS `contact_address`;
...
+    `active` BOOLEAN DEFAULT TRUE,
+    `role` VARCHAR(63)  DEFAULT NULL,
+    `relationship` VARCHAR(63)  DEFAULT NULL,
+    `contact_priority` INT DEFAULT 1 COMMENT '1=highest priority',
+    `is_primary_contact` BOOLEAN DEFAULT FALSE,
+    `is_emergency_contact` BOOLEAN DEFAULT FALSE,
+    `can_make_medical_decisions` BOOLEAN DEFAULT FALSE,
+    `can_receive_medical_info` BOOLEAN DEFAULT FALSE,
+    `start_date` DATE,
+    `end_date` DATE,
```
bradymiller: would ensure dates are set to NULL by default (if not auto set to something like DATE below)

sql/database.sql
```diff
@@ -3497,8 +3603,7 @@ INSERT INTO layout_group_properties (grp_form_id, grp_group_id, grp_title, grp_m
 INSERT INTO layout_group_properties (grp_form_id, grp_group_id, grp_title, grp_mapping) VALUES ('DEM', '4', 'Employer'    , ''    );
```
bradymiller, line 3616: PLACEHOLDER FOR ONGOING REVIEW

library/templates/address_display.php
adunsulag: These files are only included from options.inc.php.  We can add in safety by checking for the presence of the variables and failing otherwise.  My preference would be to convert all of these to class widgets but David's built off the original address work so I can see why he went this direction and timing I don't see that happening.

interface/new/new_comprehensive_save.php
```diff
@@ -81,13 +83,25 @@
...
+            // if we need to save other kinds of addresses we could do that here with our field column...
+            if (!empty($addressFieldData)) {
+                $contactService = new ContactService();
+                $contact = $contactService->getOrCreateForEntity('patient_data', $pid);
+                $contactAddressService = new ContactAddressService();
+                $contactAddressService->saveAddressesForContact($contact->get_id(), $addressFieldData);
+            }
+        }
+    } catch (Exception $e) {
+        (new SystemLogger())->error("Fatal error in address processing", [
```
kojiromike, line 100: If this is fatal, should the program continue?

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1739,3 +1630,521 @@ ALTER TABLE prescriptions ADD COLUMN diagnosis TEXT COMMENT 'Diagnosis or reason
...
+#EndIf
+
+-- -------------------------------------------------------------------------------------------------------------------------------------------------------
+-- relatedperson-relationshiptype Valuesets
+-- https://terminology.hl7.org/6.5.0/ValueSet-v3-PersonalRelationshipRoleType.html
+
+#IfNotRow2D list_options list_id lists option_id related_person_relationship
+INSERT INTO list_options (list_id,option_id,title, seq, is_default, option_value)
+    VALUES ('lists','related_person_relationship','Related Person Relationships',0, 1, 0);
+
```
sjpadgett, line 1999: Remove the Relationship list, personal_relationship, or append to it but definitely don't want duplicate lists. 
You can replace personal_relationship with this list unsure refactor hits' to fix. However, why isn't the existing list (sql/4_2_0-to-4_2_1_upgrade.sql starts at L-417) acceptable?
Please don't make me have to do later because I'm a little busy at the moment. Suggest dedup the two lists by code/option_id and then fix existing list of related persons. 
I believe I've already pointed this out prior message.

sjpadgett: I feel this is important to resolve

library/templates/address_save_handler.php
sjpadgett: /library/ajax if an ajax, I didn't look. It's where I always put these type of scripts.

sql/7_0_3-to-7_0_4_upgrade.sql
psoas: We discussed pro's/con's of this at last tues call.  I can type in details later or discuss on phone.

sjpadgett: umm still was my concern and just leaving as is is not an option for me.
I looked into replacing list with yours by simply deleting the Relationship list. I currently don't see where else used except in ccda for my related persons which I'll take care of. I like your list as more comprehensive but really should also add code in notes because original list has done so.

Easiest way forward would be to simply delete original the add new under same list name, however; if list is currently only used in a few places, just delete original and i'll take care or i'll show you where to refactor. I'm looking into now for places used and will get back to use very soon.

I'm trying to get my PR merged so you can resolve your upcoming conflicts but first I need to fix if refactor needed for your list resolution.

I hope you understand that I'm trying to prevent duplicate content in lists and is something I'm trying to fix ongoing including rechecking all my new lists before release if I can find the time.

At any rate it looks like you're going to make deadline so congratulations and very happy to see this effort, thanks.

I'll be back soon or i'll give you a call if I come across something needing your input.

sjpadgett: @psoas @adunsulag Looks like only touched in interface/super/edit_list.php L-1338
```
 if ($list_id == 'language') {
          echo xlt('ISO 639 Code');
      } elseif (in_array($list_id, ['personal_relationship', 'religious_affiliation', 'ethnicity', 'race', 'drug_route'])) {
          echo xlt('HL7-V3 Concept Code');
      } elseif ($list_id == 'Immunization_Completion_Status') {
```

So how do you want to handle or give me a call to discuss?

sjpadgett:
First know that there are several left over artifacts not being addresses like new service I put in src/Services/DemographicsRelatedPersonsService.php that I don't know if new related person is using, if not, needs to be addressed. Also it's used in demographics and demografic_full. Since I told David that if he does this replacement, i'd clean up my abandoned implementation so i'll do that.

Second, it's not right that because people decide to use dev-master in production where us developers have to concern ourselves with maintaining legacy for code that hasn't been released! This becomes a giant  unnecessary amount of work for us that develop new code. No users are paying for me to break my development to cater to them using development code in production. 
I mean seriously, I find it disrespectful for users to even consider this as our responsibility. If user decides to use master in production then they need to accept that responsibility as theirs and adjust accordingly!

I'll be doing the list cabal and cleanup for removal of my related persons and push to David's repo. Stay tune.

adunsulag:
CareTeam Updates w/ Related Person
image

image

sjpadgett:
Working:
```
 SET @group_id = (SELECT max(`grp_group_id`)+1 FROM layout_group_properties WHERE grp_form_id='DEM')
INSERT INTO layout_group_properties(grp_form_id, grp_group_id, grp_title, grp_mapping)VALUES ('DEM', @group_id, 'Related','')

The above statement failed: Duplicate entry 'DEM-1' for key 'PRIMARY'
Upgrading will continue.
SET @group_id = (SELECT max(`grp_group_id`) FROM layout_group_properties WHERE grp_form_id='DEM' AND grp_title='Related')
SET @seq_add_to = (SELECT max(seq) FROM layout_options WHERE group_id = @group_id AND form_id='DEM')
INSERT INTO `layout_options`(`form_id`, `field_id`, `group_id`, `title`, `seq`, `data_type`, `uor`, `fld_length`, `max_length`, `list_id`, `titlecols`, `datacols`, `default_value`, `edit_options`, `description`, `fld_rows`)VALUES ('DEM','related_persons',@group_id,'',@seq_add_to+1,56,1,0,0,'',4,4,'','["J","SP"]','Related Persons',0)

The above statement failed: Column 'group_id' cannot be null
Upgrading will continue.
```

adunsulag:
CareTeam validating with the changes from #9249.

image

We still need to pass in the FHIR search for RelatedPerson.  Debating on whether that should go into DemographicsRelatedPersonService (probably would rename it to RelatedPersonService at this point or into ContactRelationService (would need to pass in a search where the foreign_table_name=person via the FHIR service).

sjpadgett:
> CareTeam validating with the changes from #9249.
> 
> image
> We still need to pass in the FHIR search for RelatedPerson. Debating on whether that should go into DemographicsRelatedPersonService (probably would rename it to RelatedPersonService at this point or into ContactRelationService (would need to pass in a search where the foreign_table_name=person via the FHIR service).

I'd do whatever is easier. I had planned to delete DemographicsRelatedPersonService but I forget what unused methods I may have added for additional support besides add/save and delete.

sjpadgett:
@psoas please give me access to you repo....

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -176,20 +176,18 @@ ALTER TABLE `track_events` DROP INDEX `unique_event_label_url`;
 ALTER TABLE `track_events` ADD UNIQUE `unique_event_label_target` (`event_label`, `event_url`(255), `event_target`(255));
 #EndIf
 
-#IfNotTable care_teams
-CREATE TABLE `care_teams` (
+#IfNotTable care_team
+CREATE TABLE `care_team` (
```
sjpadgett: @psoas I see that database has the wrong name but care_teams is the table name I used throughout care teams implementation.
This rename breaks CCDA. Looks like care_team was used for this related person implementation so I'll need to refactor back to care_teams.
Also note I resolved database to master for conflicts and will reposition new related person where it belongs so don't freak out.
I'll push new once I resolve an upgrade error and fix care_team

adunsulag: This is my fault, I thought I got the CCDA change to use care_team but must have missed something.

sjpadgett: nope my fault because named wrongly in database. So I'm keeping care_team and I'll fix in registry and CCDA

sjpadgett: Most certainly an interesting mission!:)

sjpadgett: woops, if we rename to care_team then we'd have to to a migration from existing care_teams to care_team table and while I can do I think best to stay with legacy names.
Boy I hope I don't have to make any major decisions soon. Seems I'm a little wishy washy these days!:)

sjpadgett:
@psoas I'm fixing the rector and styling errors but I'll leave the phpstan for you. 
Don't know why tests are failing but must be database. IME usually is.
I upgraded from a v702 database fine and tested the database install as import fine so????
I'm able to get into app okay. No testing yet

src/Common/Uuid/UuidRegistry.php
```diff
@@ -67,7 +68,6 @@ class UuidRegistry
         'procedure_specimen' => ['table_name' => 'procedure_specimen', 'table_id' => 'procedure_specimen_id'],
         'questionnaire_repository' => ['table_name' => 'questionnaire_repository'],
         'questionnaire_response' => ['table_name' => 'questionnaire_response'],
-        'patient_related_persons' => ['table_name' => 'patient_related_persons', 'table_id' => 'pid'],
         'form_history_sdoh' => ['table_name' => 'form_history_sdoh'],
         'users' => ['table_name' => 'users'],
         'care_teams' => ['table_name' => 'care_teams'],
```
sjpadgett: Missing tables that need to be registered. Can anyone point me to the tables if know off top of head?

sjpadgett:
@kojiromike Isn't this getting a little on the nitpicking side of things. I assume for performance. I personally don't set variable like this but still.
```
-$smallform = $smallform ?? '';
+$smallform ??= '';
```

Also why is it that rector gave me a report which I fixed then give another with new errors. Why not all errors so don't have to do a back and forth thing?

sjpadgett:
@kojiromike I think this is an issue because the funtion signature makes the second agr optional
[PHPStan Analyze: src/Services/PersonPatientLinkService.php#L486](https://github.com/openemr/openemr/pull/9333/files#annotation_41274917737)
Static method OpenEMR\Common\Database\QueryUtils::querySingleRow() invoked with 1 parameter, 2 required.

kojiromike:
> @kojiromike Isn't this getting a little on the nitpicking side of things. I assume for performance. I personally don't set variable like this but still.
> 
> ```
> -$smallform = $smallform ?? '';
> +$smallform ??= '';
> ```
> 
> Also why is it that rector gave me a report which I fixed then give another with new errors. Why not all errors so don't have to do a back and forth thing?

Rector will fix the things for you if you run it before you commit.

src/Services/PersonPatientLinkService.php
```diff
@@ -0,0 +1,519 @@
...
+        // Links by method
+        $methodCounts = QueryUtils::sqlStatementThrowException(
+            "SELECT link_method, COUNT(*) as cnt 
+             FROM person_patient_link 
+             WHERE active = 1 
+             GROUP BY link_method"
+        );
+
+        $stats['by_method'] = [];
+        while ($row = QueryUtils::fetchArray($methodCounts)) {
```
kojiromike:
```suggestion
        while ($row = QueryUtils::fetchRecords($methodCounts)) {
```

src/Services/CareTeamService.php
```diff
@@ -296,44 +359,329 @@ public function getOne($uuid, $puuidBind = null)
         return $this->search($search);
     }
 
+
+    public function hasActiveCareTeam($pid)
+    {
+        $result = QueryUtils::sqlStatementThrowException(
+            "SELECT COUNT(*) as count FROM " . self::CARE_TEAM_TABLE . " WHERE pid = ? AND (status = 'active' OR status IS NULL)",
+            [$pid]
+        );
```
stephenwaite:
```
$count = QueryUtils::fetchSingleValue(
    "SELECT COUNT(*) as count FROM " . self::CARE_TEAM_TABLE . " WHERE pid = ? AND (status = 'active' OR status IS NULL)",
    'count',  
    [$pid]
);
```
found it in the php log at [up for grabs demo](https://four.openemr.io/b/openemr/interface/login/login.php?site=default)

> [07-Nov-2025 14:25:28 UTC] PHP Fatal error:  Uncaught Error: Cannot use object of type ADORecordSet_mysqli as array in /var/www/localhost/htdocs/b/openemr/src/Services/CareTeamService.php:370
Stack trace:
#0 /var/www/localhost/htdocs/b/openemr/src/Patient/Cards/CareTeamViewCard.php(117): OpenEMR\Services\CareTeamService->hasActiveCareTeam(1)
#1 /var/www/localhost/htdocs/b/openemr/interface/patient_file/summary/demographics.php(1242): OpenEMR\Patient\Cards\CareTeamViewCard->getTemplateVariables()
#2 {main}
  thrown in /var/www/localhost/htdocs/b/openemr/src/Services/CareTeamService.php on line 370

src/Services/PersonPatientLinkService.php
sjpadgett: thanks

src/Services/CareTeamService.php
psoas: @sjpadgett @stephenwaite
I had error on my docker after I pulled Jerry's code.  I fixed and comitted it.   Actually, I just saw your post AFTER doing the fix.

sjpadgett:
oh if only I was a drinking man! lol

sjpadgett:
@psoas Your forms aren't following theme. Need to address most likely twigs.
I haven't done much testing and will leave that to you but you're now green and ready after further peer reviews.
Since a lot of your code has been touched for phpstan and rector I highly recommend a through testing of database and user UI.
In dark theme one can't see UI forms so a must fix.

I'm just too tired to do anymore here unless you need my help then of course reach out.
You've done a great job David and appreciate you getting this in for certification release.

Us Floridians do okay don't we! lol

library/ajax/person_search_ajax.php
```diff
@@ -0,0 +1,695 @@
...
+            ],
+            'how_to_get_token' => [
+                'From meta tag' => 'document.querySelector(\'meta[name="csrf-token"]\')?.content',
+                'From OpenEMR global' => 'csrf_token_js (if available)',
+                'From data attribute' => 'document.body.dataset.csrfToken',
+            ]
+        ]
+    ], JSON_PRETTY_PRINT);
+    exit;
+}
```
bradymiller: users shouldn't be able to see above output

library/ajax/person_search_ajax.php
```diff
@@ -0,0 +1,695 @@
...
+    echo json_encode([
+        'success' => false,
+        'message' => 'Unauthorized - No session',
+        'debug' => [
+            'session_id' => session_id(),
+            'session_data' => array_keys($_SESSION ?? [])
+        ]
+    ], JSON_PRETTY_PRINT);
+    exit;
+}
```
bradymiller: users shouldn't be able to see above output

library/ajax/person_search_ajax.php
```diff
@@ -0,0 +1,695 @@
...
+        'found' => !empty($value),
+        'length' => $value ? strlen((string) $value) : 0,
+        'first_10' => $value ? substr((string) $value, 0, 10) : 'null',
+        'last_10' => $value ? substr((string) $value, -10) : 'null',
+    ];
+}
+
+// ===== STEP 2: SESSION CHECK =====
+$debugInfo['session'] = [
+    'session_id' => session_id(),
```
bradymiller: i wouldn't put session_id anywhere (not in output or any logs)

library/templates/address_display.php
bradymiller: yes, would check for presence of passed variable from calling script to ensure secure.

library/templates/address_save_handler.php
bradymiller: hi, but this is just a list of functions.

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1922,3 +1811,556 @@ INSERT INTO `preference_value_sets`
...
+    `marital_status` VARCHAR(31) DEFAULT NULL,
+    `race` VARCHAR(63) DEFAULT NULL,
+    `ethnicity` VARCHAR(63) DEFAULT NULL,
+    `preferred_language` VARCHAR(63) DEFAULT NULL COMMENT 'ISO 639-1 code',
+    `communication` VARCHAR(254) DEFAULT NULL COMMENT 'Communication preferences/needs',
+    `ssn` VARCHAR(31) DEFAULT NULL COMMENT 'Should be encrypted in application',
+    `active` TINYINT(1) DEFAULT 1 COMMENT '1=active, 0=inactive',
+    `inactive_reason` VARCHAR(255) DEFAULT NULL,
+    `inactive_date` DATETIME DEFAULT NULL,
+    `notes` TEXT DEFAULT NULL,
```
bradymiller: TEXT should be bare (ie. remove the DEFAULT NULL)

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1922,3 +1811,556 @@ INSERT INTO `preference_value_sets`
...
+    `active` BOOLEAN DEFAULT TRUE,
+    `role` VARCHAR(63)  DEFAULT NULL,
+    `relationship` VARCHAR(63)  DEFAULT NULL,
+    `contact_priority` INT DEFAULT 1 COMMENT '1=highest priority',
+    `is_primary_contact` BOOLEAN DEFAULT FALSE,
+    `is_emergency_contact` BOOLEAN DEFAULT FALSE,
+    `can_make_medical_decisions` BOOLEAN DEFAULT FALSE,
+    `can_receive_medical_info` BOOLEAN DEFAULT FALSE,
+    `start_date` DATE,
+    `end_date` DATE,
```
bradymiller: rec defaulting dates to NULL (unless they are automatically set to something already like the created_date beloe)

library/ajax/person_search_ajax.php
sjpadgett: I'm taking all that out

sjpadgett (commented):
Add Telecom Contacts isn't working! There isn't an click event attached to the add button
```
<div id="btn_add_telecom_telecoms_624" class="fas fa-plus-square text-primary pl-3" style="display: inline-block; line-height: 1.5; cursor: pointer;"></div>
```
I fixed Brady's review items so I'm back on FHIR

templates/patient/demographics/telecom_display.html.twig
```diff
@@ -0,0 +1,123 @@
+{#
+  Telecom Display
+  Displays telecom records for a contact in read-only mode
+#}
+
+<style>
+    .telecom-inactive-display {
+        opacity: 0.6;
+        background-color: #f8f9fa;
```
sjpadgett, line 9: Breaks themes. Inline in general should not override themes. Formatting in related person may be necessary to condense view.
Also note that when rendering twigs in dashboard, bootstrap is already in scope. If you don't specifically set no_bootstrap in a header e.g.
```
 {% block head %}
    {{ setupHeader(['no_bootstrap','datetime-picker']) }}
 {% endblock %}
```
Problems with views can ensue. Dark theme is a popular theme with us old folks so save so skin later and address:)
I found this for all the new cards I added to dashboard but may not be true always. Just need to check what and when dependencies are injected.

psoas: I have been working on "Add Telecom Contacts isn't working!".  Regarding add button, your changes somehow broke it.  However, the way it was coded was flawed, also messing up activate/inactivate.  Initially, I was in a rush and solved problem with claude.  My own analysis resulted in my fixing issues by adding uniqid() to twig template functions (simpler, better, less convoluted than claude solution.)  I am now working on further fixes in javascript related to this.

sjpadgett: It was a rector auto fix. Not sure I trust rector to do fixes but this may be a one off due to some twig template reference but I'm not that curious to run it down as to why. Good catch!

psoas:
@sjpadgett 
Your changes to "person_search_ajax.php" seems to have broken it!

sjpadgett:
No, I just haven’t pushed the changes.

sjpadgett:
works for me. whats wrong

psoas:
"person_search_ajax.php" works for me now.  I don't know what I did to create errors before but I cannot reproduce the errors now.

So, one last major bug:
--If a related person already exists, you can add addresses and telecoms to it.
--However, when a new related person is added or selected, you cannot add addresses and telecoms because there is no event listener attached to the plus button.  

However, if you click on save after adding and editing the person and go back, then you can add addresses and telecoms.  There must be something wrong with the template used to add a related person, or some info the twig file is not sending to the included nested twig file.

PR #9342, opened 2025-11-03
ci(phpstan): forbid using deprecated db functions

### LINKED ISSUE #9341: New phpstan rules to limit deprecated code
As we progress towards modernizing the codebase, we run the risk that well-meaning contributors can introduce code that uses things we're trying to deprecate. For example, we know laminas-db is problematic in #9308, but it's widely used, particularly Carecoordination, and we can't remove it right now. So someone else may decide to use it. We need to prevent that.

I propose to introduce some custom phpstan rules to prevent using deprecated code in new contributions. We can start with:

1. Prevent using laminas-db
2. Prevent using sql.inc.php functions directly. (We can steer contributors toward QueryUtils.)

We'll add the existing uses to a phpstan baseline file, so they won't error constantly, but we can iterate away from big dependencies by limiting their further use.

kojiromike:
Fixes #9341

Introduce new phpstan rules to prevent using deprecated db functions.

- [x] Move phpstan configuration into `.phpstan` subdir.
- [x] Introduce phpstan extension to find and block laminas-db
- [x] Introduce phpstan extension to find and block sql.inc.php


bradymiller:
hi @kojiromike , may need to revert the db stuff for now since don't want it slowing down ONC stuff. Thoughts @sjpadgett and @stephenwaite (oops, meant @adunsulag ) ?

kojiromike:
@bradymiller if it slows down the ONC certification, sure we can put it on hold. Has it been doing that?

sjpadgett:
No guys it’s already done. It just caught me off guard and I was tired. I
apologize. I pretty much already have it factored.

PR #9344, opened 2025-11-04
fix: remove requirement of opcache php extension from composer

### LINKED ISSUE #9343: fix: remove requirement of opcache php extension from composer
remove requirement of opcache php extension from composer

bradymiller:
fixes #9343

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "6f235daaafafebb2adfddf5e38464ce4",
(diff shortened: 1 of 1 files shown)


bradymiller:
hi @kojiromike ,
Very strange indeed for phpstan to fail (even though not touching code) :)
```
Error: Use QueryUtils::fetchRecords() or QueryUtils::fetchArrayFromResultSet() instead of sqlFetchArray().
Error: Use QueryUtils::fetchRecords() or QueryUtils::fetchArrayFromResultSet() instead of sqlFetchArray().
Error: Process completed with exit code 1.
```

Looks like this started with the commit that brought in the 'ci(phpstan): forbid using deprecated db functions'. I am assuming this is not the intention though.

kojiromike:
@bradymiller that should be fixed now

PR #9357, opened 2025-11-05
Search handling for FHIR ServiceRequest

### LINKED ISSUE #9355: feat: US Core 8.0.0 ServiceRequest
### Describe the problem

ServiceRequest FHIR api currently has the following outstanding issues

- [ ] code search with no system - Does not respond to searching on the code search parameter without any system component (since procedure_order_code.procedure_code usually has a code type in the search parameter this fails)
- [ ] code search with system component - Does not respond to searching on a code with a system component such as code=http://snomed.info/sct|56251003
- [ ] category search fails - categories for ServiceRequest can be either a string or a numeric SNOMED-CT code.  searching on the numeric code fails as its not being mapped to the underlying OpenEMR data.
- [ ] Occurrence fails when there is both a date and a datePeriod (start and end date) as occurrence can only have one or other not both for validation.

Features are implemented for us core profile.

### LINKED ISSUE #9356: bug: FHIR token search does not handle FHIR system component
### Describe the bug

Not sure how we've gone on for so long but a token search that includes the system component does not handle the search properly.  This works when the system is always known (such as a particular field always being a LOINC code or a SNOMED-CT code) but if a search value can work across multiple code systems (such as ICD10, SNOMED-CT, CPT4, etc), the searching does not work with OpenEMR's way of representing code sets.  I discovered this with ServiceRequest which can have ICD10,CPT4, or SNOMED-CT codes in them so a request like the following fails:

```
GET https://localhost:8523/apis/default/fhir/ServiceRequest?patient=96506861-511f-4f6d-bc97-b65a78cf1995&code=56251003&code=http%3A%2F%2Fsnomed.info%2Fsct%7C56251003
```

### Expected behavior
Using a **search-param=system|code**  approach should successfully return results for the correct system by translating the underlying code to be CODE_TYPE:code for the OpenEMR field search.

### Client configuration

- Browser: Chrome
- OpenEMR version: OpenEMR 7.0.3+
- Operating system: Ubuntu 24.0 LTS

adunsulag:
Fixes #9355 Service Request US Core 8.0.0
Fixes #9356 FHIR Token Search with system component

Fixed the issues with occurrence not being handled properly on the date formatting.

Fixed the issues with the code and category searches not working given the underlying OpenEMR procedure_order_code.procedure_code data field using the codetype:code syntax.

Added a new suffix search modifier for dealing with naked codes (code w/o a system component) that will allow ServiceRequest to search across all search types.  I may need to add a new type of SearchField just for codes as I think Condition and the Procedure endpoints may have this same problem.

image

### DIFF
--- a/src/Services/Search/SearchFieldStatementResolver.php
+++ b/src/Services/Search/SearchFieldStatementResolver.php
@@ -270,23 +270,18 @@ class SearchFieldStatementResolver
         $modifier = $searchField->getModifier();
         $values = $searchField->getValues();
         foreach ($values as $value) {
-            if ($modifier === 'prefix') {
-                array_push($clauses, $searchField->getField() . ' LIKE ?');
-                $searchFragment->addBoundValue($value . "%");
-            } else if ($modifier === 'suffix') {
-                array_push($clauses, $searchField->getField() . ' LIKE ?');
-                $searchFragment->addBoundValue("%" . $value);
-            } else if ($modifier === 'contains') {
-                array_push($clauses, $searchField->getField() . ' LIKE ?');
-                $searchFragment->addBoundValue('%' . $value . '%');
-            } else if ($modifier === 'exact') {
-                // not we may want to grab the specific table collation here in order to improve performance
-                // and avoid db casting...
-                array_push($clauses, "BINARY " . $searchField->getField() . ' = ?');
-                $searchFragment->addBoundValue($value);
-            } else if ($modifier == SearchModifier::NOT_EQUALS_EXACT) {
-                array_push($clauses, "BINARY " . $searchField->getField() . ' != ?');
-                $searchFragment->addBoundValue($value);
+            $result = match ($modifier) {
+                SearchModifier::PREFIX => [$searchField->getField() . ' LIKE ?', $value . '%'],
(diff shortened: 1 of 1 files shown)


src/Services/Search/SearchFieldStatementResolver.php
kojiromike:
```suggestion
$result = match ($modifier) {
    'prefix' => [$searchField->getField() . ' LIKE ?', $value . '%'],
    'suffix' => [$searchField->getField() . ' LIKE ?', '%' . $value],
    'contains' => [$searchField->getField() . ' LIKE ?', '%' . $value . '%'],
    'exact' => ["BINARY " . $searchField->getField() . ' = ?', $value],
    SearchModifier::NOT_EQUALS_EXACT => ["BINARY " . $searchField->getField() . ' != ?', $value],
    default => null,
};

if ($result !== null) {
    [$clause, $bound] = $result;
    $clauses[] = $clause;
    $searchFragment->addBoundValue($bound);
}
```

adunsulag: @kojiromike I had the thought at Midnight that if I don't turn this into a match clause Michael's going to make a comment, so I did at least think about it :)  Tiredness won out though haha. I'll make the change.

PR #9384, opened 2025-11-09
Feat openemr #9373 bulk fhir

### LINKED ISSUE #9373: feat: SMART on FHIR V2 Multipatient / Bulk Export
### Describe the problem

Need to fix / implement pieces that are needed for the the SMART on FHIR V2 Multipatient implementation.
The following Areas need to be addressed

- [X] Bulk Data Authorization
- [X] Group Compartment Export Tests STU2
- [ ] Group Compartment Validation Tests
- - [x] Patient Resources
- - [X] AllergyIntolerance
- - [X] CarePlan
- - [x] CareTeam
- - [X] Condition
- - [X] Device
- - [x] DiagnosticReport
- - [x] DocumentReference
- - [X] Goal
- - [x] Immunization
- - [x] MedicationRequest
- - [ ] Observation
- - [x] Procedure
- - [X] Encounter
- - [X] Organization
- - [X] Practitioner
- - [X] Provenance
- - [X] ServiceRequest
- - [x] RelatedPerson
- - [x] Coverage
- - [x] MedicationDispense
- - [x] Specimen
- - [x] Location
- [x] Group Compartment Export Cancel Tests
- [x] Group Compartment Export Parameters Tests

### LINKED ISSUE #9382: bug: MedicationRequest with drugs inventory does not handle RXCUI code correctly.
### Describe the bug

The drugs table is populated via the inventory module (or loaded in manually from import).  The drugs.drug_code follows the OpenEMR standard format of **code_type:code**.  This is how the inventory management populates the drug code.  However, the FHIR api prepends **RXCUI:** to the drugs.drug_code if there is a value.  This is incorrect behavior as the RXCUI: should already be in the table based on the way the rest of OpenEMR functions.  

This results in FHIR returning a code value of 'RXCUI' instead of the numeric value as the drug_code value ends up being RXCUI:RXCUI:numeric_value.

### To Reproduce

1. Add a drug via the Inventory screen and give it a template name
Image

Image

2. Add a prescription using the template name

Image

3. Make http request with valid bearer token to GET /fhir/MedicationRequest/<uuid> and see the error

Image

### Expected Behavior

Result should look like this:

Image

**Note fixing this will cause anyone who has done an import directly into the drugs.drug_code table w/o the codesystem present to have bad data**
### Client configuration

- OpenEMR version: OpenEMR 7.0.3 patch 3

adunsulag:
Fixes #9373 bulk on fhir export

Fixes #9382 

Implements Bulk on FHIR spec compliance for the following resources
- Patient
- Location
- MedicationAdherence
- MedicationDispense
- Encounter
- Immunization
- DocumentReference
- Observation(category=vitals,laboratory)

Fixes _outputFormat,_since Bulk on FHIR export.

Still have a number of issues working on.  Can see #9373 for details.

### DIFF
--- a/tests/Tests/Services/FHIR/FhirMedicationRequestServiceUSCore8Test.php
+++ b/tests/Tests/Services/FHIR/FhirMedicationRequestServiceUSCore8Test.php
@@ -99,7 +99,9 @@ class FhirMedicationRequestServiceUSCore8Test extends TestCase
             'medication_adherence_date_asserted' => '2023-01-15 01:00:00',
             'medication_adherence_information_source' => 'patient',
             'medication_adherence_information_source_title' => 'Patient',
-            'medication_adherence_information_source_codes' => 'SNOMED-CT:116154003'
+            'medication_adherence_information_source_codes' => 'SNOMED-CT:116154003',
+            'reporting_source_type' => 'user',
+            'reporting_source_uuid' => 'practitioner-uuid-789'
         ];
 
         // Minimal US Core compliant data (required fields only)


adunsulag:
Will come back to this on Monday but wanted to push this out for people to have a chance to see it.  Its very fascinating to me that the BULK on FHIR profile validation seems to be more conformant to spec than the API Single Patient.  It caught a number of issues that the API Single Patient validation did not via Inferno.  Will have to come back and do a sweep in API Single Patient and hope that there isn't conflicting validation requirements.

Also noticed that the Inferno Validator java server (on docker) appears to have a memory leak.  Memory usage has gone from 4GB of memory up to 12GB of memory over a 2 day period with nothing being freed.  Doing a docker compose restart hl7_validator_service appears to free everything up once again.

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1922,3 +1922,8 @@ INSERT INTO `preference_value_sets`
     ('95541-9','OTH','http://terminology.hl7.org/CodeSystem/v3-NullFlavor','Other (see free text)',100,1),
     ('81338-6','OTH','http://terminology.hl7.org/CodeSystem/v3-NullFlavor','Other (see free text)',100,1);
 #EndIf
+
+
+#IfMissingColumn immunizations encounter_id
+ALTER TABLE `immunizations` ADD COLUMN `encounter_id` BIGINT(20) DEFAULT NULL COMMENT 'fk to form_encounter.encounter to link immunization to encounter record';
+#EndIf
```
bradymiller, line 2488: also need change in database.sql

library/classes/POSRef.class.php
```diff
@@ -19,122 +19,38 @@
...
-        $pos[] =  ["code" => "94","title" => xl("Unassigned"), "description" => "N/A"];
-        $pos[] =  ["code" => "95","title" => xl("Unassigned"), "description" => "N/A"];
-        $pos[] =  ["code" => "96","title" => xl("Unassigned"), "description" => "N/A"];
-        $pos[] =  ["code" => "97","title" => xl("Unassigned"), "description" => "N/A"];
-        $pos[] =  ["code" => "98","title" => xl("Unassigned"), "description" => "N/A"];
-        $pos[] =  ["code" => "99","title" => xl("Other Place of Service"), "description" => "Other place of service not identified above. "];
+
+        // Iterate through all PlaceOfServiceEnum cases and build the array
+        // AI Generated
+        foreach (PlaceOfServiceEnum::cases() as $posEnum) {
```
kojiromike, line 41: 😅

src/Common/Forms/Types/EncounterListOptionType.php
```diff
@@ -0,0 +1,129 @@
...
+
+use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Common\Layouts\LayoutsUtils;
+use OpenEMR\Services\EncounterService;
+
+class EncounterListOptionType implements IOptionFormType
+{
+    private ?EncounterService $encounterService = null;
+
+    public function __construct(private $pid = null)
```
kojiromike, line 24: The interface doesn't require a constructor, and there's at least one other implementation of this interface with a different constructor sig. Just making sure this is what you want.

src/Services/LocationService.php
```diff
@@ -74,12 +76,15 @@ public function getAll($search = [], $isAndCondition = true)
                     null as website,
                     email,
                     `date` AS last_updated,
-                    "' . self::TYPE_PATIENT . '" AS `type`
+                    "' . self::TYPE_PATIENT . '" AS `type`,
+                    \'\' AS location_role_type
```
kojiromike, line 80: suggest you use a heredoc to avoid the leaning toothpicks

src/Common/Forms/Types/EncounterListOptionType.php
adunsulag: Yea, I'm trying to provide a pathway to have some conformity with the options.inc.php and the building of widgets (would be lovely for this to eventually become web components).  The constructor I'm leaving up to the implementation to decide how to handle any additional state that is required.  EncounterListOptionType relies on a global pid value in options.inc.php and a pid that's in context in the other location its used, whereas LocalProviderListType doesn't rely on any additional state than what is passed in through the interface function methods.

adunsulag:
- This leaves outstanding PractitionerRole which I think is what is causing Practitioner to fail validation
- Smoking Status 
- and Screening Assessment Tests seems to be failing and not picking up some of the categories

That should then pass everything with single patient api and bulk on fhir api.  Leaving just the granular security scopes & patient context to fix.

sql/7_0_3-to-7_0_4_upgrade.sql
adunsulag: Fixed

adunsulag:
This has gotten too big, once I've got resolved any failing tests and rector (rector suggests a fix that phpstan won't accept), my plan is to bring this in.

The only thing outstanding at this point for BULK Fhir export is fixing up the observation laboratory resources for whatever reason inferno isn't picking them.

adunsulag:
Going to have to try on a fresh database as I'm not getting these same errors.

adunsulag:
Bringing this in so I can focus on granular scope access.

PR #9391, opened 2025-11-10
style(patient_select): use match/array instead of if/elseif

### LINKED ISSUE #9390: patient finder select should use match/array instead of elif
A small refactor to patient finder select to simplify a couple of blocks.

kojiromike:
Fixes #9390

Small style refactor to patient selector to skip some repetitive elseif blocks.

### DIFF
--- a/interface/main/finder/patient_select.php
+++ b/interface/main/finder/patient_select.php
@@ -272,20 +272,14 @@ if ($popup) {
     echo "<input type='hidden' name='patient' value='" . attr($patient) . "' />\n";
     echo "<input type='hidden' name='findBy'  value='" . attr($findBy) . "' />\n";
 
-    if ($findBy == "Last") {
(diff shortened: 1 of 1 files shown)


interface/main/finder/patient_select.php
```diff
@@ -571,11 +562,11 @@ function submitList(offset) {
 // will set the pid and load all the other frames.
     objID = eObj.id;
     var parts = objID.split("~");
-    <?php if (!$popup) { ?>
+    <?php if ($popup) { ?>
+        dlgclose("srchDone", parts[0]);
+    <?php } else { ?>
         top.restoreSession();
         document.location.href = "../../patient_file/summary/demographics.php?set_pid=" + parts[0];
-    <?php } elseif ($popup) { ?>
```
kojiromike, line 577: `if ($a) { … } elseif (!$a) { … }` doesn't make sense to me. I tried to think if maybe there's an obscure way `$popup` could be mutated, but I don't think it can be.

PR #9405, opened 2025-11-11
FHIR fixes #9399,#9397,#9398 inferno tests

### LINKED ISSUE #9397: bug: FHIR MedicationRequest cannot have reportedBoolean and reportedReference at same time
### Describe the bug

FHIRMedicationRequest is failing because both the reportedBoolean and reportedReference are being populated at the same time.  Specification has reported[x] and only one property should be populated.

### To Reproduce

1. Run the certification tests ```./vendor/bin/phpunit --testsuite certification --filter SinglePatient311APITest
```
<img width="1537" height="503" alt="Image" src="https://github.com/user-attachments/assets/8009d465-4511-45af-86a3-1a8652075b35" />

### Expected behavior
Tests should pass and either reportedBoolean is populated or if we have a user who did the reporting, reportedReference, but not both.


### Client configuration

- Browser:
- OpenEMR version: master branch
- Operating system:

### LINKED ISSUE #9398: bug: FHIR CareTeam Profile should not include RelatedPerson references on version 3.1.1
### Describe the bug

RelatedPerson will not validate as a valid option for a reference in CareTeam profile on version 3.1.1

### To Reproduce

Image

1. Run phpunit tests ```./vendor/bin/phpunit --testsuite certification --filter SinglePatient311APITest
```
### Expected behavior
RelatedPerson is excluded when max FHIR profile version is set to 3.1.1 and included when it is version 7.0.0 or 8.0.0



### Client configuration

- OpenEMR version: master branch

### LINKED ISSUE #9399: bug: FHIR Observation labs and DiagnosticReport labs should support searching with system prefix
### Describe the bug

We fixed a bug where the system property was being ignored but now in areas like labs which are coded to just LOINC codes for their results the system specifier is throwing everything off and fails to return any results when giving a system specifier

### To Reproduce

Image

1. Run the unit tests with ``` ./vendor/bin/phpunit --testsuite certification --filter SinglePatient311APITest```

### Expected behavior
Searching with **/fhir/Observation?code=http://loinc.org|5767-9** should return results if there are lab results with that code identifier.  Since LOINC is the only code set supported for lab result codes we should support those and continue to return empty results for other code systems.

### Client configuration

- Browser:
- OpenEMR version: master
- Operating system:

adunsulag:
Fixes #9399 searching with system prefix
Fixes #9397 disabled reportedReference (will fix in bulk export PR) 
Fixes #9398 Disable RelatedPerson when using 3.1.1 profile setting

### DIFF
--- a/apis/routes/_rest_routes_fhir_r4_us_core_3_1_0.inc.php
+++ b/apis/routes/_rest_routes_fhir_r4_us_core_3_1_0.inc.php
@@ -667,14 +667,16 @@ return [
      *      security={{"openemr_auth":{}}}
      * )
      */
-    "GET /fhir/CareTeam" => function (HttpRestRequest $request) {
+    "GET /fhir/CareTeam" => function (HttpRestRequest $request, OEGlobalsBag $globalsBag) {
         $getParams = $request->getQueryParams();
+        $restController = new FhirCareTeamRestController();
+        $restController->setOEGlobals($globalsBag);
         if ($request->isPatientRequest()) {
             // only allow access to data of binded patient
-            $return = (new FhirCareTeamRestController())->getAll($getParams, $request->getPatientUUIDString());
+            $return = $restController->getAll($getParams, $request->getPatientUUIDString());
         } else {
             RestConfig::request_authorization_check($request, "patients", "med");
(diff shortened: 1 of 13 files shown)


adunsulag:
Bringing this in to fix broken tests.

PR #9410, opened 2025-11-12
Twig - TwigExtension - OEGlobalsBag usage fix

igormukhingmailcom:

During development of other feature I found that `TwigExtension` uses service `GlobalsService` which is designed for completely other purpose. And it worked just because `GlobalsService` had no Setting metadata format checks. 

`OEGlobalsBag` suit better, so rewrote to use it instead.

### DIFF
--- a/tests/Tests/Isolated/Common/Twig/TwigContainerIsolatedTest.php
+++ b/tests/Tests/Isolated/Common/Twig/TwigContainerIsolatedTest.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * @package   OpenEMR
+ *
+ * @link      http://www.open-emr.org
+ *
+ * @author    Igor Mukhin <igor.mukhin@gmail.com>
+ * @copyright Copyright (c) 2025 OpenCoreEMR Inc
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Tests\Isolated\Common\Twig;
+
+use OpenEMR\Common\Twig\TwigContainer;
+use OpenEMR\Common\Twig\TwigExtension;
+use OpenEMR\Core\OEGlobalsBag;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\CoversMethod;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+#[Group('isolated')]
+#[Group('twig')]
+#[CoversClass(TwigContainer::class)]
+#[CoversMethod(TwigContainer::class, 'getTwig')]
+#[CoversMethod(TwigExtension::class, 'getGlobals')]
+class TwigContainerIsolatedTest extends TestCase
+{
+    protected function setUp(): void
+    {
+        $GLOBALS['fileroot'] = __DIR__ . '/../../../../../'; // @todo Remove this workaround after removal from TwigContainer
+        $GLOBALS['date_display_format'] ??= 0;
+    }
+
+    #[Test]
+    #[DataProvider('renderDataProvider')]
+    public function renderTest(
+        array $globals,
+        string $templateAsString,
+        string $expectedRenderedHtml,
+    ): void {
+
+        $globalsBag = OEGlobalsBag::getInstance();
+        foreach ($globals as $key => $value) {
+            $globalsBag->set($key, $value);
+        }
+
+        $twigContainer = new TwigContainer();
+        $twigEnvironment = $twigContainer->getTwig();
+        $template = $twigEnvironment->createTemplate($templateAsString);
+        $this->assertEquals($expectedRenderedHtml, $template->render());
+    }
+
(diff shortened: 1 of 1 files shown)


src/Common/Twig/TwigExtension.php
```diff
@@ -53,25 +53,21 @@ protected function getOemrUiInstance($oemrSettings = [])
...
-    public function __construct(GlobalsService $globals, ?Kernel $kernel)
+
+    public function __construct(OEGlobalsBag $globals, ?Kernel $kernel)
     {
-        $this->globals = $globals->getGlobalsMetadata();
+        $this->globals = $globals;
         $this->kernel = $kernel;
     }
 
     public function getGlobals(): array
```
kojiromike, line 57: It doesn't matter for this PR, but I find myself wondering why, if the set of names is fixed and known anyway, why return an array instead of making these available as individual properties.

kojiromike (commented):
This looks good to me - is there any test coverage here?

src/Common/Twig/TwigExtension.php
igormukhingmailcom: Its interface method from Twig. Twig call it to get vars that should be passed to all templates (together with ones passed to render function directly). Thats why they are globals - because they global for all templates and you can use them without passing explicitly at `render` method.

```php
namespace Twig\Extension;

/**
 * Allows Twig extensions to add globals to the context.
 *
 * @author Fabien Potencier <fabien@symfony.com>
 */
interface GlobalsInterface
{
    /**
     * @return array<string, mixed>
     */
    public function getGlobals(): array;
}
```

igormukhingmailcom: They will be available individually inside templates as `{{ array_key }}`, for example, `Asset version: <b>{{ assetVersion }}</b>`

igormukhingmailcom:
No coverage required I believe

kojiromike:
@igormukhingmailcom we want to cover as much as possible, particularly with a change. For example, without cloning the branch, starting up docker, and walking through known twig-using parts of the project, how can I tell this PR didn't break anything?

I mean, I think this is right, but then again, in the future, when there's an update to twig, I want those tests to run again, automatically.

igormukhingmailcom:
@kojiromike I noticed rector converting multiline constructor to single line. I believe multiline is way better and probably should be required when arguments > 2. As multiline arguments producing less conflicts / allow easier rebases when, for example, 1st parameter and 3rd parameters are changed in different branches... WDYT?

```diff
-    public function __construct(
-        OEGlobalsBag $globals,
-        ?Kernel $kernel = null
-    ) {
-        $this->globals = $globals;
-        $this->kernel = $kernel;
+    public function __construct(protected OEGlobalsBag $globals, protected ?Kernel $kernel = null)
+    {
     }
```

igormukhingmailcom:
> how can I tell this PR didn't break anything?

As we had no tests involving `twig::render` - probably nothing would indicate something is wrong, yes.
So added test where render is used, involving some globals at rendering.

kojiromike:
> @kojiromike I noticed rector converting multiline constructor to single line. I believe multiline is way better and probably should be required when arguments > 2. As multiline arguments producing less conflicts / allow easier rebases when, for example, 1st parameter and 3rd parameters are changed in different branches... WDYT?
> 
> ```diff
> -    public function __construct(
> -        OEGlobalsBag $globals,
> -        ?Kernel $kernel = null
> -    ) {
> -        $this->globals = $globals;
> -        $this->kernel = $kernel;
> +    public function __construct(protected OEGlobalsBag $globals, protected ?Kernel $kernel = null)
> +    {
>      }
> ```

I don't think rector cares about the formatting. What rector appears to be doing here is converting the syntax into property promotion. If you reformat the constructor, but keep the property promotion, rector will leave it alone.

igormukhingmailcom:
Ah, thanks. I did not noticed what it actually did. Makes sense.

kojiromike:
@adunsulag @bradymiller this looks like a good fix to me. Want it in 7.0.4?

kojiromike:
@igormukhingmailcom The TwigContainer coverage is pretty good now, but the TwigExtension coverage is still small. Please add coverage of at least the parts of the class that you changed.

kojiromike:
@igormukhingmailcom I would like to merge this so I can work on another problem with the TwigExtension. What do you think? Is it ready to go?

igormukhingmailcom:
Hi, @kojiromike 
Yeah, it is ready for merging I believe
I pushed changes you requested earlier (few more test cases) a moment ago

igormukhingmailcom:
Checks will be green after https://github.com/openemr/openemr/pull/9606 merged

kojiromike:
> Checks will be green after https://github.com/openemr/openemr/pull/9606 merged

Merged. Please rebase.

kojiromike:
@igormukhingmailcom Is it possible to fix the phpstan issues listed?

sjpadgett:
Is this going to be an issue: https://github.com/openemr/openemr/pull/9667/files#r2593638647

igormukhingmailcom:
@kojiromike Rebased

sjpadgett:
@kojiromike Did you look at my changes to OEGlobalsBag not working prompting me to make changes to class/glocals and fix compatibility mode? https://github.com/openemr/openemr/pull/9667/files#r2593638647

PR #9454, opened 2025-11-15
Fix incorrect totals in Financial Summary report (#8651)

### LINKED ISSUE #8651: Financial Summary by Service Code report incorrect totals
The codes database table allows for multiple rows with the same code but varied modifiers.

This clause in the query to populate the Financial Summary by Service Code report does not account for that:

https://github.com/openemr/openemr/blob/551ea97338336d52e9d0e880d8c52c62081d14c8/interface/reports/svc_code_financial_report.php#L224

This results in an incorrect calculation of the totals included in the report, as the amounts for a particular code will be multiplied by how many times it appears in the codes table.

Modifying the line to the following would fix the issue:

`"LEFT JOIN (select code, max(financial_reporting) as financial_reporting from codes group by code) AS c ON c.code = b.code " .`

teresanguyen503:
Fixes #8651 

The Financial Summary by Service Code report was calculating totals incorrectly when a code had multiple rows in the codes table with different modifiers. This change updates the query to correctly calculate totals.

Updating query in file openemr/interface/reports/svc_code_financial_report.php on line 224.

### DIFF
--- a/interface/reports/svc_code_financial_report.php
+++ b/interface/reports/svc_code_financial_report.php
@@ -221,7 +221,7 @@ if (!empty($_POST['form_refresh']) || !empty($_POST['form_csvexport'])) {
     "JOIN (select pid, encounter, code, sum(pay_amount) as paid, sum(adj_amount) as adjust " .
     "from ar_activity WHERE deleted IS NULL group by pid, encounter, code) as ar_act " .
     "ON ar_act.pid=b.pid and ar_act.encounter=b.encounter and ar_act.code=b.code " .
-    "LEFT OUTER JOIN codes AS c ON c.code = b.code " .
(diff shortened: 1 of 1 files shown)


kojiromike:
https://github.com/openemr/openemr/blob/879a5fbbf3b886ef9dcbfe9a9031c159f30eb272/interface/super/load_codes.php seems to imply that `(code, code_type)` are intended to be distinct as a pair, although there's no constraint. I'm also concerned how this will interact with https://github.com/openemr/openemr/blob/551ea97338336d52e9d0e880d8c52c62081d14c8/interface/reports/svc_code_financial_report.php#L243

I realize the lack of constraints and guarantees are a huge issue. I just want to make sure we don't break the behavior that the user can select rows with or without financial reporting by selecting the `Important Codes` checkbox.

kojiromike:
Addressed in #10438

Issue #9401, opened 2025-11-11
Flow Board Colors are unset when you set them


kojiromike:
From the forums: https://community.open-emr.org/t/flow-board-colors-disappeared/26327?u=kojiromike

### Describe the bug

If you change a color in the flow board, all the colors stop having any effect, and the flow board contents just stay white.

### To Reproduce

> I was doing some changes in the “Appointment statuses” list under “Admin - Forms” and the second I change the color of a category, or even add a new one, all colors disappear from the entire flow board. Status shows, but only in text, but no colors, all white. I tried going back to defaults, changing theme, but nothing is bringing the colors back. any idea?

PR #9461, opened 2025-11-15
Trim leading '#' from color value in patient tracker

### LINKED ISSUE #8190: bug: Flow board does not change color
### Describe the bug

The flow board should reflect the appointment status color

### To Reproduce

Open flowboard. the background is white.

### Expected behavior
Background color of each appointment should change with status color.

### Client configuration

- Browser: Chrome
- OpenEMR version: 7.0.3 Demo
- Operating system: Unknown (Possible ubuntu)

ophthal:
This fixes the Flow Board color problem when apptstat colors are changed.  Base DB list_option.notes has hex values without # but appointment status color updates all color/notes values to include the #.  That is probably the other place to look should this also be a Calendar problem.

Fixes #8190, #9401

Alternatively, the left trim could occur higher up in src/Services/PatientTrackerService.php

 the function 

image

### DIFF
--- a/interface/patient_tracker/patient_tracker.php
+++ b/interface/patient_tracker/patient_tracker.php
@@ -573,7 +573,7 @@ if (!($_REQUEST['flb_table'] ?? null)) {
                         $newarrive = collect_checkin($tracker_id);
                         $newend = collect_checkout($tracker_id);
                         $colorevents = (collectApptStatusSettings($status));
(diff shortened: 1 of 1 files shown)


kojiromike:
Closing in favor of #9485, which fixes this at the service layer so all callers benefit. Thanks @ophthal!

Issue #9414, opened 2025-11-12
Current Bugs Demographic Related Person


sjpadgett:
@psoas 
I'm listing bugs here for new related person feature as I run across them.
If working on any of these please note that you are here so we don't duplicate effort.

- Edit changes for existing related person are not being saved. Appears insert works okay thus far. Appears using global date format dd/mm/yyyy fails in date util.
- Feature is not following selected theme.
- Clean up left over file artifacts for removed features e.g. DemographicsRelatedPersonsService.php

sjpadgett:
@psoas To fix your theme issue simply remove all the inline styles for background-color:  in your twig templates, about eight I believe, and you should be golden.

<img width="1920" height="1015" alt="Image" src="https://github.com/user-attachments/assets/90bf227c-284b-47f9-9ca9-a8937a71bc5a" />

psoas:
When creating a person/relationship, unable to add telecoms or add addresses in the subsequent input form.

psoas:
@sjpadgett 
Removing the inline styles will remove needed background colors.
For appearance to look right, I need background colors.
So, I need a replacement.

Are the following replacements ok:
.bg-info			
.bg-light	
.bg-info .bg-gradient	
background-color: var(--gray300)
border-color: var(--gray300)
border-color: var(--info)

Bootstrap 4 does not have enough color classes available, so I need to at least be able to use variables.
Boostrap 5 has enough color classes.  Will we be transitioning to boostrap 5 any time soon?

I hope that variables are ok.

sjpadgett:
@psoas You don't have to remove all styles, just the background-color.
This feature sits in the middle of the Dashboard demo view and can not be fixed to a light style and must follow current theme.
As long as your colors follow theme background and text colors, I'm fine. If you use a lighter color for background than selected dark theme then you end up with a flashlight effect in middle of demo view.
Simply remove just the 
- background-color: #f8f9fa;
- background-color: #fff;
- any background-color inline
from your display twigs in templates/patient/demographics twigs.

We spent many months setting up our themes so we have a consistent look and feel plus a concerted effort to minimized inline styles so that we narrow where we have to go to add and change themes.
When developing in LBF and Dashboard it is sometimes necessary to use inline for such as positioning and sizing.

By removing the background-color overrides you won't need to replace with BS classes but if you want to use them, that is perfectly fine and the preferred choice. i.e. bg-light in dark theme is dark etc.

psoas:
More Bugs:
[...]

PR #9469, opened 2025-11-16
Fixes OpenEMR #9414 (partial):  Related person, telecom, and address form inits

psoas:
Partial fix to #9414

Addresses and telecoms can now be added in form that displays after creating a new person/relationship.


sjpadgett:
@psoas after I merge this it will cause conflicts in your other you'll need to fix I'm guessing. I'm not worried about datetime as UI datepicker is using time so onward. Thanks for this effort. It does improve the project nicely. I still recommend you collapse by default the person being displayed in dashboard view.
image

This will need to be pulled into 705 release branch by one of us admins.

bradymiller:
@copilot, please create a PR with this code for the rel-704 branch

PR #9484, opened 2025-11-16
Fixes OpenEMR #9414:  Related person save for alternative date formats

### LINKED ISSUE #9414: Current Bugs Demographic Related Person

psoas:
Fixes #9414 

Related person did not save if a date format different from default is selected.

This fixes that plus fixes background styling problems with alternate themes.


sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1866,8 +1866,8 @@ CREATE TABLE `contact_relation` (
     `is_emergency_contact` BOOLEAN DEFAULT FALSE,
     `can_make_medical_decisions` BOOLEAN DEFAULT FALSE,
     `can_receive_medical_info` BOOLEAN DEFAULT FALSE,
-    `start_date` DATE DEFAULT NULL,
-    `end_date` DATE DEFAULT NULL,
+    `start_date` DATETIME DEFAULT NULL,
```
adunsulag, line 1869: @sjpadgett will this impact anything in ccda? I'm guessing no, but wanted to check.

sjpadgett: No but why datetime. Time precision not needed here! Unless not storing datetime and only date but I don't see where the datetime-picker format has changed to collect time.
Also doesn't our date picker have a setting to follow the format set in locale global? I can never remember.
Anyway, if it works I guess okay

@psoas Have you run this in dark theme after changes to style? Does it look okay?

psoas: @sjpadgett @adunsulag 
contact_telecom and contact_address both use datetime for the same columns.

It is because the DateFormatterUtils.php uses datetime and not date:
public static function dateStringToDateTime(string $DateValue, bool $includeSeconds = false)
...
return $dateTime;

I think that I would need to add another function to DateFormatterUtils.php.
Is it preferred to use Date instead of datetime?
can datetime class in php output a date format?

--when I tested it, dark theme worked, since it modifies the color variables, and i switched my background color definitions to the css variables instead of hard-coded colors.

sjpadgett: I believe sending a date only to datetime column is okay and will fill with a zeroed time. However, DateFormatterUtils has the public static function oeFormatShortDate($date = 'today', $showYear = true)
Just thoroughly test these changes after changing date to datetime.

sjpadgett:
@psoas You can't edit person name in related person once created. Need this to work

psoas:
@adunsulag @sjpadgett 
You did not comment on my other PR, #9469.
They fix 2 separate things.  Are you able to bring that one in while I do more work on this one?

sjpadgett:
looks like the date picker is set up for datetime in UI so we'll leave this be! may merge if you're okay with @psoas 
Will look over other PR and decide best way to order merges.

sjpadgett:
@psoas Let me know when you're ready.

psoas:
@sjpadgett 
I rebased and took care of conflicts.  I am ready for you to merge now before I work on my next set of essential bug fixes.

sjpadgett:
@psoas we release 27 or 28th so...

PR #9485, opened 2025-11-17
fix(tracker): trim leading '#' from color setting in PatientTrackerService

### LINKED ISSUE #8190: bug: Flow board does not change color

ophthal:
Fixes #8190, #9401

The Flow Board color problem when appointment status colors are changed. The base DB `list_option.notes` has hex values without `#` but appointment status color updates include the `#`. This caused color mismatches.

- Trim leading `#` from color value in `PatientTrackerService::collectApptStatusSettings()` at the service layer, so all callers benefit

Supersedes #9461 which fixed it at the view layer.

### DIFF
--- a/src/Services/PatientTrackerService.php
+++ b/src/Services/PatientTrackerService.php
@@ -292,7 +292,7 @@ class PatientTrackerService extends BaseService
         }
 
         [$color_settings['color'], $color_settings['time_alert']] = explode("|", (string) $row['notes']);
-        $color_setting['color'] = ltrim($color_settings['color'],'#');
+        $color_settings['color'] = ltrim($color_settings['color'],'#');
         return $color_settings;
     }
 


src/Services/PatientTrackerService.php
```diff
@@ -292,6 +292,7 @@ public static function collectApptStatusSettings($option)
         }
 
         [$color_settings['color'], $color_settings['time_alert']] = explode("|", (string) $row['notes']);
+        $color_setting['color'] = ltrim($color_settings['color'],'#');
```
kojiromike:
```suggestion
        $color_settings['color'] = ltrim($color_settings['color'],'#');
```

PR #9492, opened 2025-11-17
fix: asset loading errors for header

### LINKED ISSUE #9441: Asset loading errors for exclusion tokens and empty strings
Two related but distinct asset loading issues causing error logs.

**Issue 1: "no_fontawesome" exclusion not counted properly**

Error:
```
Not all selected assets were included in header {"selectedAssets":["no_bootstrap","no_fontawesome","no_main-theme","no_dialog"],"foundAssets":[...]}
```

**Root cause:** Header.php lines 223-226 only increment `$excludedCount` when asset has `allowNoLoad: true`. Fontawesome doesn't have this flag in config.yaml, so "no_fontawesome" isn't counted as excluded, causing mismatch.

Assets with allowNoLoad: jquery, bootstrap, utility, main-theme, compact-theme, textformat, dialog
Missing allowNoLoad: fontawesome (and others)

**Issue 2: Empty string asset requested**

Error:
```  
Not all selected assets were included in header {"selectedAssets":[""],"foundAssets":[...]}
```

**Root cause:** Code somewhere calls `Header::setupHeader('')` with empty string. Line 182-184 converts string to array: `[''\]`, causing empty string to be in selectedAssets.

**Related to but different from #9435** (popper asset not found).

kojiromike:
Fixes #9441

This commit fixes two related asset loading issues in Header.php.

1. **Exclusion token counting for assets without allowNoLoad flag**
   - Previously, only assets with allowNoLoad: true had their exclusion tokens (no_<asset>) counted
   - This caused false error logs when using "no_fontawesome" and similar exclusions
   - Now all exclusion tokens are counted upfront if the corresponding asset exists in the map
   - Example: "no_fontawesome" is now properly recognized even though fontawesome doesn't have allowNoLoad: true

2. **Empty string asset handling**
   - Empty strings and whitespace-only strings passed to setupHeader() are now filtered out
   - Prevents "Not all selected assets were included" errors from empty asset requests
   - Uses trim() to catch whitespace-only strings like ' ' or '  '

The fix maintains backward compatibility while eliminating spurious error log entries.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/Core/Header.php
+++ b/src/Core/Header.php
@@ -216,20 +216,8 @@ class Header
         $foundAssets = [];
         $excludedCount = 0;
 
-        // Count exclusion tokens (assets starting with "no_") where the corresponding asset exists in map
-        foreach ($selectedAssets as $selectedAsset) {
-            $selectedAsset = (string) $selectedAsset;
-            if (str_starts_with($selectedAsset, 'no_')) {
-                $assetName = substr($selectedAsset, 3); // Remove "no_" prefix
-                if (array_key_exists($assetName, $map)) {
-                    $excludedCount++;
-                }
-            }
-        }
-
         foreach ($map as $k => $opts) {
             $autoload = $opts['autoload'] ?? false;
-            $allowNoLoad = $opts['allowNoLoad'] ?? false;
             $alreadyBuilt = $opts['alreadyBuilt'] ?? false;
             $loadInFile = $opts['loadInFile'] ?? false;
             $rtl = $opts['rtl'] ?? false;
@@ -237,6 +225,7 @@ class Header
             if ((self::$isHeader === true && $autoload === true) || in_array($k, $selectedAssets) || ($loadInFile && $loadInFile === self::getCurrentFile())) {
                 // Skip loading if exclusion token (no_<asset>) is present in selectedAssets
                 if (in_array("no_" . $k, $selectedAssets)) {
+                    $excludedCount++;
                     continue;
                 }
                 $foundAssets[] = $k;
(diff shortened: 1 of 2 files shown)


src/Core/Header.php
```diff
@@ -220,11 +235,9 @@ private static function parseConfigFile($map, $selectedAssets = [])
             $rtl = $opts['rtl'] ?? false;
 
             if ((self::$isHeader === true && $autoload === true) || in_array($k, $selectedAssets) || ($loadInFile && $loadInFile === self::getCurrentFile())) {
-                if ($allowNoLoad === true) {
```
kojiromike, line 223: @openemr/admins What is `allowNoLoad` supposed to do?

bradymiller: pretty sure it was just for the things that are autoloaded. As I recall there are use cases in portal (and possibly elsewhere) where did not want to bring in items that are autoloaded.

kojiromike: @bradymiller I'm trying to understand if Claude did the right thing by removing the check here. The bug was that `$excludedCount` would only increment if `$allowNoLoad === true`, and thus anything else that had `no_` before it would be excluded, but not counted, I think?

So Claude separated the counting from the continuing, which I think is right, but there's a change in behavior: Previously, it would only `continue` if both `$allowNoLoad===true` and the asset started with `no_`. Now it will `continue` regardless of `$allowNoLoad`, only if the asset starts with `no_`. Does that seem right to you?

bradymiller: hi @kojiromike ,
 - For the exclusion counter, would need to ensure it only counts items with autoload set to true. Since doing a `no_` on something that is not autoloaded would throw off the count (unless both did package with a `no_` in same line :) ).
 - I think we can drop the entire exclusion counter block and do `$excludedCount++;` above line 225.
 - And completely drop all the allowNoLoad (line 232) and in config/config.yaml

PR #9510, opened 2025-11-19
Feat openemr fix fhir observations us core8

### LINKED ISSUE #9489: bug: FHIR US Core Care Treatment Preferences Inferno does not pass validation
### Describe the bug

US Core Care Treatment Preferences will not validate with inferno as our US Care Treatment Preference general preference code of LOINC:95541-9 is the only code that Inferno looks at for care-experience instead of all the other care experience categories we have available.  Code 95541-9 only has a general text option and no valueCodeableConcept which FHIR is looking for and so the test fails.  We need a valueCodeableConcept for the general care experience we can select from.

Even though the specification specifically calls for supporting other more specific code types the actual specification implementation ONLY allows for code 95541-9.

Image

Image

### To Reproduce

Image

Image

1. Add a '....'

### Expected behavior

We should add a treatment preference record for the general findings or a code selector widget to allow any snomed code to be selected for the general experience.  Quickest option to fix is to just add another option.  I found in the FHIR Behavioral Health IG an example where they use a snomed code of treatment by a professional as a health care experience that I'll go ahead and add.

### LINKED ISSUE #9494: feat: Smoking Status Observation is missing a valueQuantity field.
### Describe the problem

We need to be able to report on the number of smoking packs per day.  This could be done with a standard observation but would deviate pretty hard from the Lifestyle History places where we record smoking status.

Add to lifestyle history the ability to designate additional fields missing from smoking observations.

adunsulag:
Fixes #9489  by adding additional treatment preference so
valueCodeableConcept for FHIR observation gets populated

Fixes label on ADI Warning

Fixes #9494 Adds quantity to smoking observation

With these changes BULK on FHIR export are fully passing.  Single Patient API is also passing but local inferno is skipping over validating DiagnosticReport.performer(type=Organization) for some reason.  Will attempt to test on ONC public inferno instance to see if I can reproduce the problem.  If its a glitch on our local instance then this PR makes g10 fully code complete for ONC.

### DIFF
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -15372,3 +15372,6 @@ CREATE TABLE `preference_value_sets` (
     ('81364-2','309687009','http://snomed.info/sct','Baptist',11,1),
     ('81364-2','160540005','http://snomed.info/sct','Sikh',12,1),
     ('81364-2','LA14063-6','http://loinc.org','Prefer not to answer',98,1);
+    -- General Preferences
+    INSERT INTO preference_value_sets(`loinc_code`,`answer_code`,`answer_system`,`answer_display`,`sort_order`,`active`) VALUES
+    ('95541-9', 314433002, 'http://snomed.info/sct', 'Preference for health professional (finding)', 1, 1);
--- a/src/Services/FHIR/DiagnosticReport/FhirDiagnosticReportClinicalNotesService.php
+++ b/src/Services/FHIR/DiagnosticReport/FhirDiagnosticReportClinicalNotesService.php
@@ -78,7 +78,7 @@ class FhirDiagnosticReportClinicalNotesService extends FhirServiceBase implement
     {
         return  [
             'patient' => $this->getPatientContextSearchField(),
-            'code' => new FhirSearchParameterDefinition('type', SearchFieldType::TOKEN, ['code']),
+            'code' => new FhirSearchParameterDefinition('code', SearchFieldType::TOKEN, ['code']),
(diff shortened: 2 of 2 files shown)


sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -2582,3 +2582,11 @@ UPDATE categories
...
 UPDATE categories SET codes='LOINC:42348-3' WHERE name='Advance Directive' AND codes='LOINC:LP173418-7';
 #EndIf
+
+-- FHIR mandates that the general care experience code support a valueCodeableConcept so we add at least one to the list
+-- this was found as an example on the Behavioral Health Guides here: https://www.fhir.org/guides/astp/bhp/Observation-provider-preference-observation-example-2.xml.html
+-- so adding it to our list of care experience preferences so we can pass certification testing, a future improvement would be to give
+-- a general code selector or something for the general care experience so anything in SNOMED-CT could be used.
+#IfNotRow2D preference_value_sets loinc_code 95541-9 answer_code 314433002
+INSERT INTO preference_value_sets(loinc_code,answer_code,answer_system,answer_display,sort_order,active) VALUES('95541-9', 314433002, 'http://snomed.info/sct', 'Preference for health professional (finding)', 1, 1);
+#EndIf
```
bradymiller, line 2592: ensure in database.sql

bradymiller:
review done. just noted missing change in database.sql. otherwise looks good.

adunsulag:
@sjpadgett this fixes your DocumentReference issue you were running into with g6.

PR #9515, opened 2025-11-19
Fixes #8639 granular scopes resource filter

### LINKED ISSUE #8639: feat: SMART on FHIR V2.2.0 - support new scope syntax for granular permissions
### Describe the problem

Implement the SMART on FHIR v2.2.0 new scope syntax for granular permissions as detailed here: https://build.fhir.org/ig/HL7/smart-app-launch/scopes-and-launch-context.html

We need to support the bare minimum *context*/*resource*.cruds syntax where C = create, R = read, U=update, D=delete, and S=search syntax for restricting permissions.  We will first target just read and search permission support with the plan to add in CUD support later once we've established this functionality.

We also need to support the minimum granular scopes for Condition, Observation, and DocumentReference as seen here:
```
            'Condition' => [
                'category=http://hl7.org/fhir/us/core/CodeSystem/condition-category|health-concern'
                ,'category=http://terminology.hl7.org/CodeSystem/condition-category|encounter-diagnosis'
                ,'category=http://terminology.hl7.org/CodeSystem/condition-category|problem-list-item'
            ]
            ,'Observation' => [
                'category=http://hl7.org/fhir/us/core/CodeSystem/us-core-category|sdoh'
                ,'category=http://terminology.hl7.org//CodeSystem-observation-category|social-history'
                ,'category=http://terminology.hl7.org/CodeSystem/observation-category|laboratory'
                ,'category=http://terminology.hl7.org/CodeSystem/observation-category|survey'
                ,'category=http://terminology.hl7.org/CodeSystem/observation-category|vital-signs'
            ]
            ,'DocumentReference' => [
                'category=http://hl7.org/fhir/us/core/CodeSystem/us-core-documentreference-category|clinical-note'
            ]
```

adunsulag:
The query filter approach on the search with the resource constraints did not work on edge cases.  One example is observations linked to a questionnaire.  Since there are two categories you could get a category of survey and one of 'social-history'.  With a constrained scope inferno would still try to grab results for survey and expect you to have observations with the survey+social-history category.  However, the query filter would block the request as survey is not an allowed constrained scope.

This new approach puts in a policy enforcement decision point where we collate the resources to be returned in the response.  This guarantees that regardless of the search parameters used an observation or condition resource cannot be returned that doesn't match the right category id.

The advantage of this approach is surety that no data leakage happens as long as the resource is correctly tagged.  The downside is that it adds additional computer usage as filtering happens AFTER the search rather than BEFORE.  As Observation can go through hundreds of resource items when the permission filtering at the end may return just a few resources this has a performance penalty.  In order to pass inferno we decided to go with that and will have to see what we can do for optimization later.

Currently only Condition and Observation have support for the resource filtering as it leverages the FhirGenericRestController.  Eventually we'll want all the resources to have the same capability and this can be done by either using the FhirGenericRestController or lverage the ResourceConstraintFilterer class.

I had to standardize the way Observation.category is populated across the resources using as we had some cases where a string value was used and others where a FHIRString and FhirURI was used.

Fixes #8639.

### DIFF
--- a/src/Common/Session/SessionConfigurationBuilder.php
+++ b/src/Common/Session/SessionConfigurationBuilder.php
@@ -90,8 +90,8 @@ class SessionConfigurationBuilder
         return (new self())
             ->setName(SessionUtil::OAUTH_SESSION_ID)
             ->setCookiePath((!empty($webRoot)) ? $webRoot . SessionUtil::OAUTH_WEBROOT : SessionUtil::OAUTH_WEBROOT)
-            ->setCookieSameSite('Lax')
-            ->setCookieSecure(false)
+            ->setCookieSameSite('None')
+            ->setCookieSecure(true)
             ->build();
     }
 


src/Common/Session/SessionConfigurationBuilder.php
```diff
@@ -90,8 +90,8 @@ public static function forOAuth(string $webRoot = ''): array
         return (new self())
             ->setName(SessionUtil::OAUTH_SESSION_ID)
             ->setCookiePath((!empty($webRoot)) ? $webRoot . SessionUtil::OAUTH_WEBROOT : SessionUtil::OAUTH_WEBROOT)
-            ->setCookieSameSite('None')
-            ->setCookieSecure(true)
+            ->setCookieSameSite('Lax')
+            ->setCookieSecure(false)
```
bradymiller: above changes meant to be?

adunsulag: No. sorry, will revert it.  Thanks for catching that. On my local inferno I have to have those settings to get the cookies to work with the ehr launch over a non-localhost ip.

bradymiller:
hi @adunsulag , Just noted issue in changes in SessionConfigurationBuilder.php (posted above), otherwise looks good.

src/Common/Session/SessionConfigurationBuilder.php
adunsulag: Fixed.

PR #9529, opened 2025-11-20
Procedure Order save specimen fix and other bug fixes

### LINKED ISSUE #9528: bug: Save & Continue Procedure Order with Specimen did not save specimen for new order, had to enter and save data again on existing order for save to work

### LINKED ISSUE #9530: bug: Clinical Notes Add Item for existing record clones previous author in author field instead of using current user.
When adding a new clinical note, it copies the Author from the previous Author that the add button was clicked on.  This will be very concerning to users who care about legal provenance.  We need to make sure the Author is the current user creating the note.

Image

Steps to reproduce.
1. Create a clinical note with one user.
2. Log out and login as a different user
3. Edit the existing clinical note
4. Hit the add button
5. See the previous author was copied.

Expected results is that the current user is populated instead.

sjpadgett:
- Procedure Order save specimen fix for when a new specimen in a new order doesn't save. Cause is flag for new test uses -1 for key which is missed in save. This -1 key is a carry over from way back in forms history.
- Fixed minor misleading different author than user shown in new note in Clinical Notes form.

Fixes #9528
Fixes #9530 

adjust for new specimen flag of -1 in save.

### DIFF
--- a/src/Events/Services/DornLabEvent.php
+++ b/src/Events/Services/DornLabEvent.php
@@ -18,15 +18,24 @@ class DornLabEvent extends Event
     public const GEN_HL7_ORDER = 'dorn.gen_hl7_order';
     public const GEN_BARCODE = 'dorn.gen_barcode';
     public const SEND_ORDER = 'dorn.send_order';
-    private ?string $hl7;
-    private ?string $reqStr;
     private array $messages = [];
     private $sendOrderResponse;
+    private ?string $hl7 = null;
+    private ?string $reqStr = null;
 
-    public function __construct(private int $formid, private int $ppid, ?string &$hl7 = null, ?string &$reqStr = null)
+    public function __construct(
+        private readonly int $formid,
+        private readonly int $ppid,
+        ?string &$hl7 = null,
+        ?string &$reqStr = null
+    )
     {
-        $this->hl7 = &$hl7;
-        $this->reqStr = &$reqStr;
+        if ($hl7 !== null) {
+            $this->hl7 =& $hl7;
+        }
+        if ($reqStr !== null) {
+            $this->reqStr =& $reqStr;
+        }
     }
 
     public function setSendOrderResponse($response): void


sjpadgett:
I restyled sql_upgrade to give a better wider view of process details text and slipped in on this PR.
image

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -176,21 +176,86 @@ ALTER TABLE `track_events` DROP INDEX `unique_event_label_url`;
...
+    `role`,
+    `facility_id`,
+    `provider_since`,
+    `status`,
+    `date_created`,
+    `date_updated`,
+    `note`
+)
+SELECT
+    ct_new.`id` AS `care_team_id`,
```
adunsulag, line 243: I'm a bit confused, isn't the care_teams we want to migrate from the patient_data.care_team_provider, patient_data.care_team_facility and patient_data.care_team_status?

We would also migrate patient_history.care_team_provider and patient_history.care_team_facility?

I have an idea for how I'd do this so I'll put something together, I don't think we can do it with a standard SQL query due to the denormalized pipe (|) syntax we use for this.  So I'll put together a SQLUpgrade function to handle it.

sjpadgett: No, I mentioned in chat that there are two migrations. one is my care team from few months ago that you added related person and two the demographic LBF care team that is in the wild. If you want to do two, please do.

sjpadgett:
Back to you my friend! lol

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -1940,8 +2010,8 @@ CREATE TABLE `person_patient_link` (
         (`form_id`, `field_id`, `group_id`, `title`, `seq`, `data_type`, `uor`, `fld_length`, `max_length`, `list_id`, `titlecols`, `datacols`, `default_value`, `edit_options`, `description`, `fld_rows`)
     VALUES
         ('DEM','additional_telecoms',@group_id,'',@max_seq+9,55,0,0,0,'',4,4,'','[\"J\",\"SP\"]','Additional Telecoms',0);
-    #Endif
-#Endif
+    #EndIf
```
bradymiller, line 2013: Have we tested this to ensure nesting works (I don't think nesting works, but if it is working on your testing, then it must :) ).

sjpadgett: While I didn't add this one you call this out on me a couple years ago when I used and was working but I removed because I didn't research in code to be 100%.
I saw this one and paid attention on testing and seems to work.
Then again I though misspelled `#Endif` use to break run but didn't see here or happenstance just allowed success.
Since parse is sequential and an open/close tags I'd think it would work.

adunsulag: I thought I removed all of David's nested ifs, on my testing they didn't work, so maybe something has changed and now they work.

sjpadgett: Don't go by me guys because i'm not 100% but will look at in code because I know it worked for me in the past the time I tried. Still if you want I will refactor unless you can get to @adunsulag

sjpadgett: Statement will always execute so it is broken. Sorry. Thanks for review Brady.

PR #9563, opened 2025-11-22
feat openemr 9551 migrate careteam

### LINKED ISSUE #9551: feat: migrate old care team data
### Describe the problem

We are not currently handling the old patient data care team data in demographics and both features exist simultaneously (new care teams and patient demographics care teams).

We need to migrate the patient_data.care_team_provider, patient_data.care_team_facility and patient_data.care_team_status fields to the care_team table.

We also need to migrate historical care_team data from patient_history.care_team_provider and patient_history.care_team_facility.

For the API's to work we should store the UUID from the uuid_mapping table in as the care_team.uuid.  We'll need to set the uuid_registry.mapped value to be 0 for the uuid and clean up the entries in the uuid_mapping.

We should leave the old fields alone and perhaps add a column comment to say the fields are deprecated.   
At a later date those fields could be removed.

If someone needs to revert / rollback from the upgrade the data elements would still exist but new mapped uuids would end up getting generated.

adunsulag:
- **Fixes #9551 careteam migration from patient_data**
- **Fix unit tests, remove unused code**

### DIFF
--- a/sql/7_0_3-to-7_0_4_upgrade.sql
+++ b/sql/7_0_3-to-7_0_4_upgrade.sql
@@ -2634,7 +2634,7 @@ ALTER TABLE form_history_sdoh DROP COLUMN `pregnancy_gravida`, DROP COLUMN `preg
 -- we don't want to destroy any patient data entered in these deprecated fields but we do want to stop them from appearing on the DEM form
 #IfNotRow2D layout_options form_id DEM field_id care_team_facility uor 0
 UPDATE layout_options SET uor=0 WHERE form_id='DEM' AND field_id IN ('care_team_facility', 'care_team_provider', 'care_team_status') AND uor=1;
-ALTER TABLE `patient_data` MODIFY COLUMN `care_team_provider` text DEFAULT NULL COMMENT 'Deprecated field, use care_team_member table instead';
-ALTER TABLE `patient_data` MODIFY COLUMN `care_team_facility` text DEFAULT NULL COMMENT 'Deprecated field, use care_team_member table instead';
-ALTER TABLE `patient_data` MODIFY COLUMN `care_team_status` text DEFAULT NULL COMMENT 'Deprecated field, use care_team_member table instead';
+ALTER TABLE `patient_data` MODIFY COLUMN `care_team_provider` text COMMENT 'Deprecated field, use care_team_member table instead';
+ALTER TABLE `patient_data` MODIFY COLUMN `care_team_facility` text COMMENT 'Deprecated field, use care_team_member table instead';
+ALTER TABLE `patient_data` MODIFY COLUMN `care_team_status` text COMMENT 'Deprecated field, use care_team_member table instead';
 #EndIf


adunsulag:
@sjpadgett let me know what you think this handles the patient_data migration.

src/Services/Utils/SQLUpgradeService.php
```diff
@@ -1440,4 +1455,169 @@ private function updateLayoutEditOptions($mode, $form_id, $add_option, $values):
...
+            $userRoleFacilities = QueryUtils::fetchRecords("SELECT fui.facility_id AS role_facility_id, "
+                . " u.facility_id FROM users u LEFT JOIN facility_user_ids fui ON fui.uid = u.id WHERE fui.field_id='provider_id' AND u.id = ? ", [$providerId]);
+            $userRoleFacilityIds = array_map(fn($item) => intval($item['role_facility_id'] ?? $item['facility_id']), $userRoleFacilities);
+            if (!empty($facilitiesIds)) {
+                // intersect user role facilities with care team facilities
+                $facilitiesIds = array_values(array_intersect($facilitiesIds, $userRoleFacilityIds));
+            } else {
+                // assign all user role facilities to care team facilities
+                $facilitiesIds = $userRoleFacilityIds;
+            }
```
sjpadgett, line 1552: Nice! Covers the bases. I'd, being the lazy guy I am, would simply drop facility reference and just used users facility and left blank if they didn't have one assigned. We don't force facility in new care team pretty sure plus it ensures current user's facility is referenced and not one they have since moved from but move hasn't been updated in the patient_data.
Plus billing is more closely tied to the users facility in users.
Really having the multiselect for facility in demographics is pointless and I admit it was a poor design on my part.

This works just as well and is artfully done.
I'm going to pull down and test against a 702 backup. Just need to know which of us is going to fix the conflict. I don't mind. I just need to have my fixes for DORN in by tomorrow so guys looking over will have success.

sql/7_0_3-to-7_0_4_upgrade.sql
```diff
@@ -2591,3 +2591,14 @@ UPDATE categories SET codes='LOINC:42348-3' WHERE name='Advance Directive' AND c
...
+
+#IfCareTeamsV1MigrationNeeded
+#EndIf
+
+-- we don't want to destroy any patient data entered in these deprecated fields but we do want to stop them from appearing on the DEM form
+#IfNotRow2D layout_options form_id DEM field_id care_team_facility uor 0
+UPDATE layout_options SET uor=0 WHERE form_id='DEM' AND field_id IN ('care_team_facility', 'care_team_provider', 'care_team_status') AND uor=1;
+ALTER TABLE `patient_data` MODIFY COLUMN `care_team_provider` text DEFAULT NULL COMMENT 'Deprecated field, use care_team_member table instead';
+ALTER TABLE `patient_data` MODIFY COLUMN `care_team_facility` text DEFAULT NULL COMMENT 'Deprecated field, use care_team_member table instead';
+ALTER TABLE `patient_data` MODIFY COLUMN `care_team_status` text DEFAULT NULL COMMENT 'Deprecated field, use care_team_member table instead';
```
bradymiller: above 3 TEXT fields should be bare (ie. no DEFAULT)

adunsulag: Fixed

src/Services/Utils/SQLUpgradeService.php
sjpadgett: I was looking at results from import and while it appears to duplicate already created members from my migration they really aren't fully duplicates because though have same user the new member has different facility which as far as I'm concerned is great because may may work in both facilities. So I say this tests well.

sjpadgett (approved):
good work

src/Services/Utils/SQLUpgradeService.php
```diff
@@ -20,11 +20,14 @@
...
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Database\SqlQueryException;
+use OpenEMR\Common\Uuid\UuidRegistry;
 use OpenEMR\Events\Core\SQLUpgradeEvent;
 use OpenEMR\Services\Utils\Interfaces\ISQLUpgradeService;
+use OpenEMR\Services\VersionService;
 
 class SQLUpgradeService implements ISQLUpgradeService
 {
+    const CARE_TEAMS_V1_MIGRATION_VERSION = 527;
```
sjpadgett, line 30: If this works the way I think it works, we need to ensure this const gets changed to reflect the actual database version on release.
The only other way to validate that I can think of now may be to set the care team status in patient data to 'ported' or similar then check in the shouldExecuteVersionUpdate()

adunsulag: Since I'm comparing it to any version less than or equal to 527, I would think the migration should always run unless I'm missing something.  No one should have a db version that would be greater than 527 unless they got the latest code updates right?

I could update the constant just before release, but was thinking we'd be safe with this approach.  We could go with the other option you mentioned also with a feature flag.

sjpadgett: We're currently on 528 aren't we? The method is fine I just wanted to be sure we don't miss if we update DB some more which I expect because I still don't trust update from my testing. I plan to slowly and with pinpoint assurance of no lingering table artifacts from previous tests go through a last time once you settle on this PR. Don't have to merge just if sql upgrade is fixed. Man I'm chatty today!

PR #9588, opened 2025-11-25
Use constants instead of their values in module management UI

### LINKED ISSUE #9587: chore: use named constants instead of their underlying values in modules management screen
While trying to create and install a new module, I ran into some UI code that's switching and filtering on custom modules vs Zend/Laminas modules, and it's referencing values like `$moduleResult->type == 0`. I think referencing the backing constants would make that code much clearer.

I've already got a minor adjustment on a local branch and will send a PR shortly.

Relatedly, there are two seemingly-equivalent pairs of constants to accomplish this which perhaps would be good to consolidate, but addressing that likely requires more input.

```
src/Core/ModulesApplication.php
28:    const MODULE_TYPE_CUSTOM = 0;
29:    const MODULE_TYPE_LAMINAS = 1;

interface/modules/zend_modules/module/Installer/src/Installer/Model/InstModuleTable.php
48:    public const MODULE_TYPE_ZEND = 1;
49:    public const MODULE_TYPE_CUSTOM = 0;
```

Firehed:
Fixes #9587

Changes some conditionals in the modules management UI to reference constants instead of their backing values

No logic changes, this only improves readability.

I'd be open to additional refactoring in here (I came across it while starting development of a new module) but this seems like it should be relatively non-controversial.

No

### DIFF
--- a/interface/modules/zend_modules/module/Installer/view/installer/installer/index.phtml
+++ b/interface/modules/zend_modules/module/Installer/view/installer/installer/index.phtml
@@ -162,7 +162,7 @@ $help_items = [
                         if ($moduleResult->modName == 'Acl') {
                             continue;
                         }
-                        if ($moduleResult->type == 1) {
+                        if ($moduleResult->type == ModulesApplication::MODULE_TYPE_LAMINAS) {
                             continue;
                         }
                         $count++;
@@ -201,7 +201,7 @@ $help_items = [
                                 ?>
                             </td>
                             <td>
-                                <?php echo $this->escapeHtml(($moduleResult->type == 1) ? "Laminas" : "Custom"); ?>
+                                <?php echo $this->escapeHtml(($moduleResult->type == ModulesApplication::MODULE_TYPE_LAMINAS) ? "Laminas" : "Custom"); ?>
                             </td>
                             <td>
                                 <?php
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/module/Installer/view/installer/installer/index.phtml
```diff
@@ -160,7 +162,7 @@ $help_items = [
                         if ($moduleResult->modName == 'Acl') {
                             continue;
                         }
-                        if ($moduleResult->type == 1) {
+                        if ($moduleResult->type == ModulesApplication::MODULE_TYPE_LAMINAS) {
```
kojiromike, line 165:
```suggestion
                        if ($moduleResult->type === ModulesApplication::MODULE_TYPE_LAMINAS) {
```

(Any reason not to?)

Firehed: I was hesitant to change too much in case there was unexpected or insufficient type-casting from the database records (I'm not super familiar with the broader internals yet), but locally it does appear that it should be safe. 

Ideally, most of this would move into the `Installer\Model\InstModule` class with its properties getting type annotations, and possibly replaced with `$moduleResult->isCustom()` or similar. More than happy to address that either here or in a fast-follow PR.

kojiromike: OK, I'm a fan of small PRs, especially if there's followup.

PR #9612, opened 2025-11-27
Modernization sales by item report

### LINKED ISSUE #9611: Modernization of Sales by Item Report
As a developer on this project. I felt it was my duty to pick up the shovel and throw some new code in the pot!

juggernautsei:
Fixes #9611

Modernizing the code that runs the report.

### DIFF
--- a/interface/reports/sales_by_item_new.php
+++ b/interface/reports/sales_by_item_new.php
@@ -1,0 +1,127 @@
+<?php
+
+/**
+ * Sales by Item Report - Modernized Controller
+ *
+ * This is a report of sales by item description.
+ * Modernized version using service layer and Twig templates.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Rod Roark <rod@sunsetsystems.com>
+ * @author    Terry Hill <terry@lillysystems.com>
+ * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @author    Stephen Waite <stephen.waite@cmsvt.com>
+ * @author    Sherwin Gaddis <sherwingaddis@gmail.com>
+ * @copyright Copyright (c) 2006-2016 Rod Roark <rod@sunsetsystems.com>
+ * @copyright Copyright (c) 2015-2016 Terry Hill <terry@lillysystems.com>
+ * @copyright Copyright (c) 2017-2018 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2025      Stephen Waite <stephen.waite@cmsvt.com>
+ * @copyright Copyright (c) 2025      Sherwin Gaddis <sherwingaddis@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once("../globals.php");
+require_once("$srcdir/patient.inc.php");
+require_once "$srcdir/options.inc.php";
+
+use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Twig\TwigContainer;
+use OpenEMR\Core\Header;
+use OpenEMR\Reports\SalesByItems\Services\DataPreparationService;
+use OpenEMR\Reports\SalesByItems\Services\CsvExportService;
+use OpenEMR\Reports\SalesByItems\Services\ChartDataService;
+use OpenEMR\Reports\SalesByItems\Repository\SalesByItemsRepository;
+
+// Check ACL permissions
+if (!AclMain::aclCheckCore('acct', 'rep') && !AclMain::aclCheckCore('acct', 'rep_a')) {
+    echo (new TwigContainer(null, $GLOBALS['kernel']))->getTwig()->render(
+        'core/unauthorized.html.twig',
+        ['pageTitle' => xl("Sales by Item")]
+    );
+    exit;
+}
+
+// CSRF protection
+if (!empty($_POST)) {
+    if (!CsrfUtils::verifyCsrfToken($_POST["csrf_token_form"] ?? '')) {
+        CsrfUtils::csrfNotVerified();
+    }
+}
+
+// Get filter parameters
+$form_provider = $_POST['form_provider'] ?? null;
+if (!AclMain::aclCheckCore('acct', 'rep_a')) {
+    // Only allow user to see their own data
+    $form_provider = $_SESSION['authUserID'] ?? null;
+}
+
+$form_facility = $_POST['form_facility'] ?? null;
+$form_from_date = isset($_POST['form_from_date'])
+    ? DateToYYYYMMDD($_POST['form_from_date'])
+    : date('Y-m-d');
+$form_to_date = isset($_POST['form_to_date'])
+    ? DateToYYYYMMDD($_POST['form_to_date'])
+    : date('Y-m-d');
+$form_details = !empty($_POST['form_details']) ? true : false;
+$form_csvexport = !empty($_POST['form_csvexport']) ? true : false;
+
+// Initialize services
+$repository = new SalesByItemsRepository();
+$dataService = new DataPreparationService($repository);
+$csvService = new CsvExportService();
+$chartService = new ChartDataService();
+
+// Handle CSV export
+if ($form_csvexport) {
+    // Prepare data
+    $reportData = $form_details
+        ? $dataService->prepareDetailedReport($form_from_date, $form_to_date, $form_facility, $form_provider)
+        : $dataService->prepareSummaryReport($form_from_date, $form_to_date, $form_facility, $form_provider);
+
+    // Build CSV using service
+    $csv = $form_details
+        ? $csvService->buildDetailedCsv($reportData['rows'])
+        : $csvService->buildSummaryCsv($reportData['rows']);
+
+    // Send CSV response
+    header("Pragma: public");
+    header("Expires: 0");
+    header("Cache-Control: must-revalidate, post-check=0, pre-check=0");
+    header("Content-Type: application/force-download");
+    header("Content-Disposition: attachment; filename=sales_by_item.csv");
+    header("Content-Description: File Transfer");
+
+    echo $csv;
+    exit;
+}
+
+// Prepare report data for HTML rendering
+$reportData = null;
+$chartCategory = null;
+$chartTopItems = null;
+if (!empty($_POST['form_refresh'])) {
+    $reportData = $form_details
+        ? $dataService->prepareDetailedReport($form_from_date, $form_to_date, $form_facility, $form_provider)
+        : $dataService->prepareSummaryReport($form_from_date, $form_to_date, $form_facility, $form_provider);
+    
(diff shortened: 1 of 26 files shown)


juggernautsei:
I had Warp to run PHPstan 8.4 on the code and it came back with this report. 

PHPStan analysis completed on the modernization-sales-by-item branch. The analysis found 10,639 errors across the codebase, running PHPStan version 2.1.31 at level 8.

The errors are primarily in these categories:
•  FHIR resource classes with isset() checks on non-nullable properties
•  Missing return type specifications for array iterables
•  Return type mismatches in XML serialization methods
•  Missing type specifications throughout various classes

Most errors appear to be in the src/FHIR/R4/ directory with auto-generated FHIR resource classes.

interface/reports/sales_by_item_new.php
```diff
@@ -0,0 +1,127 @@
...
+ * @copyright Copyright (c) 2015-2016 Terry Hill <terry@lillysystems.com>
+ * @copyright Copyright (c) 2017-2018 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2025      Stephen Waite <stephen.waite@cmsvt.com>
+ * @copyright Copyright (c) 2025      Sherwin Gaddis <sherwingaddis@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+require_once("../globals.php");
+require_once("$srcdir/patient.inc.php");
+require_once "$srcdir/options.inc.php";
```
kojiromike, line 26: Why parentheses on line 25, but none on line 26?

To make the phpstan warning go away, ask warp to add an appropriate `/** @var */` declaration, since srcdir is defined in ../globals.php.

interface/reports/sales_by_item_new.php
```diff
@@ -0,0 +1,127 @@
...
+    // Send CSV response
+    header("Pragma: public");
+    header("Expires: 0");
+    header("Cache-Control: must-revalidate, post-check=0, pre-check=0");
+    header("Content-Type: application/force-download");
+    header("Content-Disposition: attachment; filename=sales_by_item.csv");
+    header("Content-Description: File Transfer");
+
+    echo $csv;
+    exit;
```
kojiromike, line 97: To be frank, I think this may be a good step forward, but as long as we're manipulating data directly in a template and using `exit` statements after supplying some headers and echoing some output, we can't really call it "modern".

Modern, IMO, would mean using something like Symfony to provide the expected result. Symfony MVC would mean separating the data from the template. It might seem more complicated, but it'd be a lot more testable and it would allow us easily to provide the same data via more than one interface: API, CLI, WebUI, etc.

src/Reports/SalesByItems/Model/SalesGroup.php
```diff
@@ -0,0 +1,84 @@
...
+    public function getItems(): array
+    {
+        return $this->items;
+    }
+
+    public function addItem(SalesItem $item): void
+    {
+        $this->items[] = $item;
+        $this->quantity += $item->getQuantity();
+        $this->total += $item->getAmount();
```
kojiromike, line 62: This is great. I'd only suggest using our own interface to make as many changes as possible.

```suggestion
        $this->addToQuantity($item->getQuantity());
        $this->addToTotal($item->getAmount());
```

src/Reports/SalesByItems/Services/BillingItemService.php
kojiromike, line 1: I'm a little concerned about how we overload the term 'Service' in OpenEMR. I know naming things is hard, but we have a `BaseServiceInterface` and a `BaseService`, and I wondered if this class should use those. I don't think they should, because src/Services/* are relatively database-heavy, but then, what is a Service, and why is this Service not that kind of Service?

One thing we might still want from src/Services/* are the event dispatchers. We should do more of that, to enable better extensibility.

src/Reports/SalesByItems/Services/ChartDataService.php
```diff
@@ -0,0 +1,246 @@
...
+        '#76b7b2', // Teal
+        '#59a14f', // Green
+        '#edc949', // Yellow
+        '#af7aa1', // Purple
+        '#ff9da7', // Pink
+        '#9c755f', // Brown
+        '#bab0ab', // Gray
+        '#1f77b4', // Dark Blue
+        '#2ca02c', // Dark Green
+    ];
```
kojiromike, line 32: Is there a better, more flexible way to declare and configure these?

src/Reports/SalesByItems/Services/ChartDataService.php
```diff
@@ -0,0 +1,246 @@
...
+    {
+        $colors = [];
+        $paletteCount = count($this->palette);
+
+        for ($i = 0; $i < $count; $i++) {
+            $colors[] = $this->palette[$i % $paletteCount];
+        }
+
+        return $colors;
+    }
```
kojiromike, line 217: Instead of walking from 0 to $count, consider this:

```suggestion
    private function generateColors(int $count): array
    {
        $paletteCount = count($this->palette);
        $repeats = (int)ceil($count / $paletteCount);  // how many copies of $palette do we need?
        $repeated = array_merge(...array_fill(0, $repeats, $this->palette)); // make that many copies
        return array_slice($repeated, 0, $count);  // trim back our copied array to the desired length
    }
```

src/Reports/SalesByItems/Services/CsvExportService.php
```diff
@@ -0,0 +1,245 @@
...
+                    $row['quantity'],
+                    $row['total']
+                );
+            } elseif ($row['type'] === 'grand_total') {
+                $csv .= $this->formatGrandTotalRow(
+                    $row['quantity'],
+                    $row['total']
+                );
+            }
+        }
```
kojiromike, line 206:
```suggestion
        foreach ($rows as $row) {
            $csv .= match ($row['type']) {
                'item' => $this->formatDetailedRow($row['item']),
                'product_total' => $this->formatProductTotalRow(
                    $row['category'],
                    $row['product'],
                    $row['quantity'],
                    $row['total']
                ),
                'category_total' => $this->formatCategoryTotalRow(
                    $row['category'],
                    $row['quantity'],
                    $row['total']
                ),
                'grand_total' => $this->formatGrandTotalRow(
                    $row['quantity'],
                    $row['total']
                ),
                default => '',
            };
  }
```

src/Reports/SalesByItems/Services/CsvExportService.php
```diff
@@ -0,0 +1,245 @@
...
```
kojiromike, line 241: This could be a match expression, too.

src/Reports/SalesByItems/Services/DataPreparationService.php
```diff
@@ -0,0 +1,226 @@
...
+        string $toDate,
+        bool $detailed = false,
+        ?int $facilityId = null,
+        ?int $providerId = null
+    ): array {
+        if ($detailed) {
+            return $this->prepareDetailedReport($fromDate, $toDate, $facilityId, $providerId);
+        } else {
+            return $this->prepareSummaryReport($fromDate, $toDate, $facilityId, $providerId);
+        }
```
kojiromike, line 140:
```suggestion
        return $detailed
            ? $this->prepareDetailedReport($fromDate, $toDate, $facilityId, $providerId)
            : $this->prepareSummaryReport($fromDate, $toDate, $facilityId, $providerId);
```

src/Reports/SalesByItems/Services/DataPreparationService.php
```diff
@@ -0,0 +1,226 @@
...
+                $item = $this->drugService->processItem($item);
+            } else {
+                // Billing
+                $item = $this->billingService->processItem($item);
+            }
+
+            $processed[] = $item;
+        }
+
+        return $processed;
```
kojiromike, line 204: I suggest using `match` here just in case it makes sense to add more categories later. Then the whole thing is really array_map.

```suggestion
        return array_map(
            fn($item) => match ($item->getCategory()) {
                'Products' => $this->drugService->processItem($item),
                default => $this->billingService->processItem($item),
            },
            $items
        );
```

interface/reports/sales_by_item_new.php
juggernautsei: There is no reason that I have ever found other than PHP accepts both, and it depends on who originally wrote the code and no one has ever changed it.

src/Reports/SalesByItems/Services/GroupingService.php
```diff
@@ -0,0 +1,230 @@
...
+            $product = $item->getDescription() ?: 'Unknown';
+
+            if (!isset($grouped[$category])) {
+                $grouped[$category] = [];
+            }
+
+            if (!isset($grouped[$category][$product])) {
+                $grouped[$category][$product] = new SalesGroup($category, $product);
+            }
+
```
kojiromike, line 56: Suggest eliminating isset checks with the null coalescing assignment operator.

```suggestion
            $grouped[$category][$product] ??= new SalesGroup($category, $product);
```

src/Reports/SalesByItems/Services/GroupingService.php
```diff
@@ -0,0 +1,230 @@
...
+     * @param SalesGroup[] $productGroups Array of SalesGroup objects in a category
+     * @return array ['quantity' => int, 'total' => float]
+     */
+    public function getCategoryTotals(array $productGroups): array
+    {
+        $quantity = 0;
+        $total = 0.0;
+
+        foreach ($productGroups as $group) {
+            $quantity += $group->getQuantity();
```
kojiromike, line 75: The SalesGroup has a facility for adding up quantities and totals, but that work is repeated here.

src/Reports/SalesByItems/Services/TotalsService.php
```diff
@@ -0,0 +1,198 @@
...
+            $product = $item->getDescription() ?: 'Unknown';
+
+            if (!isset($byGroup[$category])) {
+                $byGroup[$category] = [];
+            }
+
+            if (!isset($byGroup[$category][$product])) {
+                $byGroup[$category][$product] = [];
+            }
+
```
kojiromike, line 129: Consider using ??= here.

src/Reports/SalesByItems/Services/TotalsService.php
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Format currency for display
+     *
+     * @param float $amount The amount to format
+     * @param string $format Format string (default: US format)
+     * @return string Formatted currency
+     */
+    public function formatCurrency(float $amount, string $format = '$%.2f'): string
```
kojiromike, line 151: Currency format should be based on localization settings for the system, and should be handled in a utility service across all of openemr.

interface/reports/sales_by_item_new.php
juggernautsei: I will switch this to a Symfony Response and StreamedResponse

src/Reports/SalesByItems/Services/BillingItemService.php
juggernautsei: I will keep this in mind in my next next project.

PR #9613, opened 2025-11-28
feat(prescription): add Save and Dispense workflow for new prescriptions

cburnicki:
After fixes by @adunsulag  to the drug prescriptions were made in #9192, the "Save & Dispatch" feature that would immediately dispatch a prescribed drug from the inventory was removed.

As of now, a workaround is required where a prescription has to be added first and then edited again to dispatch the drug from the inventory. This is not ideal for two reasons:

1. It removes functionality that is already being used in clinics without any clear indication of where it was moved.

2. It replaces a one-step "Save and dispatch" with a much more complex workflow that might lead people to forget to dispatch a prescribed drug.

This change re-enables the single-step "Save and Dispense" workflow:

- Show dispense options in the Add Prescription dialog when in-house pharmacy is enabled
- Add server-side handling to save the prescription and immediately dispense using DrugSalesService
- Disable the "Save and Dispense" button until an in-house drug template is selected, with a helpful info message
- Update dispense_drug_id hidden field when selecting in-house drugs

I believe this also fixes a bug in encounterSelectList Twig function that was calling buildDisplayView() instead of buildFormView(), causing the encounter dropdown to not render properly. (@adunsulag can you verify this, I think this came in in 4af4c827f)

Fixes #

### DIFF
--- a/templates/prescription/general_list.html.twig
+++ b/templates/prescription/general_list.html.twig
@@ -369,7 +369,9 @@ $(function () {
                                                 {% else %}
                                                     {{prescription.drug|text}}
                                                 {% endif %}&nbsp;
-                                                <br />{{prescription.note|text}}
+                                                <br />{% if prescription.get_dispensation_count() > 0 %}
+                                                    <span class="badge badge-success mr-1" style="font-size: 0.7em; margin-top: 0.3em; display: inline-block;">{{ 'dispensed'|xlt }}</span>
+                                                {% endif %}{{prescription.note|text}}
                                             </td>
                                         {% else %}
                                             <td>&nbsp;</td>
@@ -379,7 +381,9 @@ $(function () {
                                                 {% else %}
                                                     {{prescription.drug|text}}
                                                 {% endif %}&nbsp;
-                                                <br />{{prescription.note|text}}
+                                                <br />{% if prescription.get_dispensation_count() > 0 %}
+                                                    <span class="badge badge-success mr-1" style="font-size: 0.7em; margin-top: 0.3em; display: inline-block;">{{ 'dispensed'|xlt }}</span>
+                                                {% endif %}{{prescription.note|text}}
                                             </td>
                                         {% endif %}
                                         <td id="{{prescription.id|attr}}">


cburnicki:
After discussing with the doctors from our clinic, the topic has become somewhat clearer to me. The feedback was:

- As the function is currently implemented (Drug Disposition via edit dialog), it's essentially unusable in the stressful clinical routine.
- Ideally, there would be three action buttons: "Save", "Save and print", and "Save and dispense"

Accordingly, these are the three scenarios in reality: medications are either only prescribed, or additionally dispensed from the clinic's own pharmacy, or a prescription is printed so that the patient obtains the medication themselves (no dispense).
It was also mentioned multiple times that every additional step in this process—for example, having to close the PDF that opens but is unnecessary when using "Save and dispense"—leads to errors and delays in the hectic clinical routine.

Does anyone have thoughts on this? Otherwise I would start to implement a solution based on this feedback.

controllers/C_Prescription.class.php
```diff
@@ -353,10 +353,70 @@ function edit_action_process()
...
+                    $fee,
+                    $patientId,
+                    $encounterId,
+                    $prescriptionId
+                );
+
+                if (!$saleId) {
+                    $dispenseError = xl('Inventory is not available for this order');
+                }
+            } catch (\Exception $e) {
```
kojiromike:
```suggestion
            } catch (\Throwable $e) {
```

We should be catching Throwable since PHP 7.

templates/prescription/general_edit.html.twig
```diff
@@ -396,14 +409,29 @@
...
         }
+
+        // Existing prescription - use the popup dispense flow
         let promise = dlgopen('interface/drugs/dispense_drug.php' +
             '?drug_id=' + encodeURIComponent(drugId) +
-            '&prescription=' + encodeURIComponent(f.id.value) +
+            '&prescription=' + encodeURIComponent(prescriptionId) +
             '&quantity=' + encodeURIComponent(f.disp_quantity.value) +
             '&fee=' + encodeURIComponent(f.disp_fee.value) +
             '&encounter=' + encodeURIComponent(f.dispense_encounter_id.value),
```
kojiromike:
```suggestion
       const params = new URLSearchParams({
           "drug_id": drugId,
           "prescription": prescriptionId,
           "quantity": f.disp_quantity.value,
           "fee": f.disp_fee.value,
           "encounter": f.dispense_encounter_id.value
       });
       let promise = dlgopen('interface/drugs/dispense_drug.php?' + params.toString(),
```

sjpadgett:
> Does anyone have thoughts on this?

Scenario sounds completely reasonable and feedback from actual day to day users is very valuable. I can't see any reson to not implement. Perhaps give admin a way to choose which to use i.e. current(v1) vs. your version.

cburnicki:
I updated the PR with two changes:

- **"Save and print" button**: This will save the prescription and show/download the pdf prescription. It will not dispense from the inventory. This is based on feedback from the clinic: In many cases, The patient is required to acquire the drug themselves for which they need to print the prescription.
- **Check/Uncheck all feature for prescription list**: Also based on feedback from the clinic: Once you edit the prescription list, all prescriptions are checked by default. This is inconvenient if, e.g. only one prescription should be printed. This feature lets the user easily check/uncheck all items in the list.

I also rebased the branch on master to include @adunsulag's bugfix

If you agree, I think this would be ready to be merged.

kojiromike:
Please run the instructions in the job summary for that failed API Docs Freshness Checks job.

cburnicki:
@kojiromike Thanks for the hint but I need a bit of help here... Should I just run `php bin/console openemr:create-api-documentation --skip-globals`? it changes the swagger file, but seemingly unrelated to my feature.

kojiromike:
@cburnicki ah, is it possible that your PR is not up to date with master? Otherwise, yes, please run that command and commit the changes.

kojiromike:
> @kojiromike Thanks for the hint but I need a bit of help here... Should I just run `php bin/console openemr:create-api-documentation --skip-globals`? it changes the swagger file, but seemingly unrelated to my feature.

Aha! it wasn't you at all. See #9671

cburnicki:
> Aha! it wasn't you at all. See #9671

Okay! Should I undo the OpenAPI commit then? What a bout the last failing test? It doesn't seem related to my changes and I remember from previous PRs, that some tests might be broken.

kojiromike:
> > Aha! it wasn't you at all. See #9671
> 
> Okay! Should I undo the OpenAPI commit then? What a bout the last failing test? It doesn't seem related to my changes and I remember from previous PRs, that some tests might be broken.

Yes, if you remove your commit and rebase against master, the test should pass.

> some tests might be broken.

Some of the E2E tests are flaky. If all your tests pass but one, we can ignore it or rerun it, and it'll usually pass.

cburnicki:
Done. With one check failing. But I cant see how that would be related to my changes. Thanks for the support!

cburnicki:
can this be merged?

kojiromike:
Not my area of expertise, but it looks OK to me. Can another @openemr/admins advise if this is good to merge?

sjpadgett:
I'll try to pull it down a little later today. I want to test.

sjpadgett:
@cburnicki Are you around?

sjpadgett:
@cburnicki 
It's been several year since I've looked at Inventory so I can't compare but I get the Save and Dispense button and after click I get the print window where on close I have a blank dialog waiting to be closed(x out). After close I'm returned to the main prescription dialog. It should close after print.

Let me know if I've missed something but for now need to wait on merging until at least the unruly dialog artifact is addressed.

sjpadgett:
Rethinking, leaving the main prescription dialog open is okay in case there is more to be prescribed.

cburnicki:
Thanks! I didn't realize this mixed behavior. I had another call and screenshare with a doctor from our clinic and the feedback was:

- Leave the prescription list open. In many cases they want to prescribe more than one med.
- Additional feedback: Increase dialog width for larger screens since there are many columns
- And they really wished for an indicator in the table to show whether a drug has been dispensed (so they can see which ones might need a printed prescriptions and which ones don't).

I made all three changes and added a "dispensed" indicator.

Screenshot 2025-12-17 at 09 10 08

sjpadgett:
@cburnicki 
Are you around to converse?

sjpadgett:
@cburnicki looks okay to me and sounds like you're happy with process so will merge so can be include in release build.
Thanks for the contribution.

cburnicki:
@sjpadgett thanks for merging and for your help. I am currently in Berlin TZ. If you still wanna reach out, feel free to send me an email!

PR #9673, opened 2025-12-05
docs(README): add missing badges

kojiromike:

Add badges to the README for actions that are missing.


kojiromike:
This only affects the README. The checks themselves do not need to run for this PR.

PR #9676, opened 2025-12-05
Adjust styling to provide clearer row separation

### LINKED ISSUE #9675: Add "zebra striping" to configuration UIs
### Describe the problem

In the config UIs (app-wide and personal settings), there are many rows of configuration options. While everything is aligned correctly, it's easy to change the wrong setting since there's very little visual differentiation.

Image

Add a little CSS to those pages providing alternating row background colors, or some other comparable approach that provides clearer separation.

I've got a small patch locally that produces this (which does work _reasonably_ well across all of the included themes), which in my opinion makes it a lot clearer which setting you're changing. I'll send a PR momentarily.

Image

Firehed:
Fixes #9675 

Adds a background to every other row in the settings UIs (Admin>Config and (User Icon)>Settings) to make it clearer which label goes with which control.

Small style adjustment with a CSS class to opt-in to the design change. That class has been applied to the config pages. Nothing else _should_ be affected.

To minimize the changes here, I've reused the existing `--light` CSS variable, which seems to produce the intended behavior across all packaged themes.

Note that there are `table table-striped` classes that seem to be supported by the existing stylesheets, but they're scoped to actual `<table>` elements, and the config forms are `<div>`-based.

No

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -351,6 +351,10 @@ function checkBackgroundServices(): void
           width: 100%;
         }
       }
+
+      .striped .row.form-group:nth-child(even) {
+        background: var(--light);
+      }
     </style>
     <?php
     $heading_title = ($userMode) ? xl("Edit User Settings") : xl("Edit Configuration");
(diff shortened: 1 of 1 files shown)


kojiromike:
Tested well enough.

PR #9682, opened 2025-12-06
Fix legacy mode.in new OEGlobalsBag

### LINKED ISSUE #9681: Unable to set Compatibility mode flag OEGlobalsBag
### Describe the bug

Unable to fetch $GLOBAL keys from module because ParameterBag wasn't being populated in global initialization.
Two issues, read only flag mishandled and compatibility property is never set.

A recent PR modified this class so @kojiromike take a look to see if I rebroke its PR intent.
I need this so will merge. I reverted my changes in my PR in favor of the merged solution and opted for singular PR so we can track and bring back to release if my PR doesn't make it.
### Client configuration

- Browser: All
- OpenEMR version: 7.0.4+
- Operating system: all

sjpadgett:
Fixes #9681

#### Does your code include anything generated by an AI Engine? Yes check my work and format class.

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -159,7 +159,9 @@ if (empty($restRequest)) {
     $restRequest = HttpRestRequest::createFromGlobals();
 }
 if (empty($globalsBag)) {
-    $globalsBag = new OeGlobalsBag([], true);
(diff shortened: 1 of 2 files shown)


kojiromike:
cc @igormukhingmailcom

PR #9693, opened 2025-12-08
fix: Use current date for CC dropdown values and reuse templates

### LINKED ISSUE #9692: InHouse CC expiration date field outdated and incorrect
### Describe the bug

There are a couple forms that provide a dropdown field for a card expiration date. Both (that I've seen) have values from the past, and one has a misalignment where the `<option>` value is different than the label.

### To Reproduce

Visible in source code `portal/portal_payment.php` and `interface/patent_file/front_payment.php`

Will send a PR to fix shortly.

Firehed:
Fixes #9692 

Fixes CC# expiration year starting in the past, and in some cases having year misalignment e.g. `<option value="2028">2030</option>`.

Moves the CC month and expiration year dropdowns to Twig templates, and updates the year path to be based on the current year. Now+12 year was picked somewhat arbitrarily to be very forgiving; in practice most cards tend to be issed within ~5years in my experience.

I also pulled out Twig into a reusable variable, since I hope to perform some additional refactors and improvements on these screens that can benefit from further Twig template use.

Note that this change does _not_ attempt to resolve any of the (many) possible downstream implications of directly accepting CC data - it justs fixes the forms and reduces some code duplication.

Screenshots show no visual changes for month field, and current year values. I also cross-checked the actual HTML.

Screenshot 2025-12-08 at 1 08 00 PM
Screenshot 2025-12-08 at 1 07 52 PM

No


Firehed:
I'll tweak the PHPStan baseline shortly, not sure that's avoidable without a _much_ more significant rework that doesn't make sense to include in scope for this change.

kojiromike (approved):

templates/forms/exp_year_dropdown.html.twig
```diff
@@ -0,0 +1,3 @@
+{% for y in range("now"|date("Y"), "now"|date("Y") + 12) %}
+  <option value="{{ y }}">{{ y }}</option>
```
bradymiller, line 2: to be safe, would escape:
```
<option value="{{ y | attr }}">{{ y | text }}</option>
```

bradymiller: note these should only be numbers, so can keep as is if wish. I generally escape everything (sometimes overkill) that gets outputted though since then never need to worry about it if it gets modified in future (like if decide to support another calendar or run it through a filter for unknown reasons).

PR #9703, opened 2025-12-09
Fixes #9702 swagger bugs

### LINKED ISSUE #9702: bug: Swagger tool v2 scopes missing, invalid $docref operation, trusted user save bug
### Describe the bug

If you go to the swagger tool (/swagger/ endpoint) and do a Select All scopes and attempt to authorize the application it returns an invalid scope for the system/DocumentReference.$docref.  

Also the v2 scope syntax are all missing from the swagger scopes.  Instead of showing the v1 and v2 which gets very confusing we should just show the v2 syntax.

Finally it appears that the swagger is caching the authorization_code parameter if the authorization request fails, this exposed a bug with our trusted user service where our REPLACE INTO was not changing the value and returning the id of the row since nothing was changed.

### Client configuration

- OpenEMR version: 7.0.4

adunsulag:
Fixed a bug where the trusted user save was not returning an identifier if the authorization code was a duplicate within the same valid timespan for the code, the request failed but gave a different error message which masked the core problem.

Fixed the swagger scopes that show up to use the new v2 syntax instead of the old v1 syntax.

Added the Questionnaire to our valid scopes which wasn't being handled properly.

Fixed an issue where the system/DocumentReference.$docref was not supported when the underlying code supports it.

Fixes #9702

### DIFF
--- a/src/Common/Auth/OpenIDConnect/Entities/ServerScopeListEntity.php
+++ b/src/Common/Auth/OpenIDConnect/Entities/ServerScopeListEntity.php
@@ -121,7 +121,9 @@ class ServerScopeListEntity
 
             $fhirScopes[] = 'patient/DocumentReference.$docref';
             $fhirScopes[] = 'user/DocumentReference.$docref';
-            $fhirScopes[] = 'system/DocumentReference.$docref';
+            if ($systemEnabled) {
+                $fhirScopes[] = 'system/DocumentReference.$docref';
+            }
 
             $this->v1ResourceFhirScopes = $fhirScopes;
         }


adunsulag:
I'm ignoring the API Docs freshness test as that is unrelated to this PR.  Once we fix the swagger library we can deal with that.

bradymiller:
Does this need to come in before release?

stephenwaite:
37ee997c918178f83272c107c00aedbf6c804649 covers the failing check, okay to bring this in @adunsulag ? then will backport it

Issue #9705, opened 2025-12-09
Refactor payments in user portal for improved flexibility


Firehed:
Today, there are a small number of payments drivers available for use within the client portal; the current internals make adding either first- or third-party support (e.g. as a module) of other payments backends challenging and likely error-prone.

I'd like to take a pass at making some improvements here, with the following goals:

- Adding support of new payments providers to be used in the client portal should be easier
- Reduced inter-mixing of PHP, HTML, and JS; better leverage templates (i.e. twig)
- Reduced code repetition
- Each "driver" can provide its own UI for the checkout process (e.g. improved support of PCI-compliant card data collection components), isolated processing, etc.
- Resolve some high-risk areas around API key handling

I expect this will take a highly-iterative approach across a number of PRs that I'll have reference this issue.

kojiromike:
@Firehed if it's not too much trouble, consider using sub-issues of this issue for each PR. That will track with the current release process. You can create and manage sub-issues fairly easily [using the api](https://docs.github.com/en/rest/issues/sub-issues?apiVersion=2022-11-28).

Firehed:
Can do!

PR #9716, opened 2025-12-09
Refactor payments UI in client portal

Firehed:
Progress on #9705.

Refactors the payments UI and rendering logic in the client portal.

While there's more work to do before I'd call the issue complete, this makes progress in the following way:

- Introduces constants for supported payment providers; updates config system to reference them
- Moves the provider-specific JS on the payment screen to dedicated, provider-specific JS files (cut-and-paste)
- Moves most inline provider-agnostic JS into a separate JS file (also cut-and-paste). This omits a few functions that rely on inlined PHP
- Migrates the table showing outstanding billable encounters to a Twig template
- Refactors the table footer summing logic to avoid `globals` and mixing rendering with side-effects

There should be no appreciable logic changes, and I've tested that the existing behavior is unchanged. You can still specify amounts in the different table `<input>` fields and the sum of that makes it into the payment processing step.

I'll aim to further refactor this in a future PR, especially regarding some of the inlined SQL queries that can lead to performance issues. This is already pushing it as reviewability goes due to its size, though the majority is pure cut-and-paste.

No

### DIFF
--- a/portal/portal_payment.php
+++ b/portal/portal_payment.php
@@ -786,7 +786,7 @@ if (($_POST['form_save'] ?? null) || ($_REQUEST['receipt'] ?? null)) {
 
             <?php
             // This logic was in the middle of HTML rendering; will get
-            // consoldiated further in another PR.
+            // consoldiated further in another PR. See #9705.
             $encs = [];
             $rows = [];
             // Get the unbilled service charges and payments by encounter for this patient.


portal/portal_payment.authorizenet.js
```diff
@@ -0,0 +1,72 @@
...
+        return response.text();
+    }).then(function(data) {
+        if(data !== 'ok') {
+            alert(data);
+            return;
+        }
+        alert(chargeMsg);
+        window.location.reload(false);
+    }).catch(function(error) {
+        alert(error)
```
kojiromike, line 69: no semicolon required here? (sorry my ECMA is rusty.)

portal/portal_payment.js
```diff
@@ -0,0 +1,229 @@
...
+                else if (paying == patient_balance) {
+                    document.getElementById('paying_' + i).style.background = '#ffffff';
+                }
+            }
+            else {
+                document.getElementById('paying_' + i).style.background = '#ffffff';
+            }
+        }
+        else {
+            break;
```
kojiromike, line 22: thoughts on making this a guard?

portal/portal_payment.stripe.js
```diff
@@ -0,0 +1,83 @@
...
+// Add an instance of the card Element into the `card-element` <div>.
+card.mount('#card-element');
+// Handle real-time validation errors from the card Element.
+card.addEventListener('change', function (event) {
+    let displayError = document.getElementById('card-errors');
+    if (event.error) {
+        displayError.textContent = event.error.message;
+    } else {
+        displayError.textContent = '';
+    }
```
kojiromike, line 30: how do you feel about ternaries?

Firehed:
Closing in favor of several smaller, more-scoped PRs.

PR #9724, opened 2025-12-09
refactor(payments): Extract gateway-specific payment JS code to standalone files

### LINKED ISSUE #9717: Extract payment-provider JS into separate files for user portal
Part of #9705.

Firehed:
Fixes #9717

Moves JS code into `.js` files, which are loaded through a normal `<script>` tag.

As part of a broader refactor of the client portal payments code, this moves the JS specific to Stripe and Authorize.net into separate files. 

Sphere is already following this _sort_ of pattern (though through a different approach; I'll likely try to standardize them later on) and InHouse doesn't seem to have any special handling.

The new JS files are a pure cut-and-paste; there are zero logic or formatting changes at this time (beyond lowering the indentation level).

No


kojiromike:
Do you know how this will interact with the npm build system? Will the js file get minified/combined and put in public/*? How would we work with it if an operation wanted to serve it from a CDN?

Firehed:
Truthfully I'm not 100% sure; I'd assume it would have the same behavior as any other `.js` file currently in the project. How that works in a broader sense is a much bigger discussion, but this generally follows existing patterns I've seen (for better or worse)

portal/portal_payment.php
```diff
@@ -1460,165 +1460,11 @@ function validateCC() {
...
-                        return;
-                    }
-                    alert(chargeMsg);
-                    window.location.reload(false);
-                }).catch(function(error) {
-                    alert(error)
-                });
-            }
-        </script>
+        <script src="portal_payment.stripe.js"></script>
```
bradymiller: incorporate cache bust in above 2 script calls

.phpstan/baseline/loader.php
```diff
@@ -1,6 +1,6 @@
 <?php declare(strict_types = 1);
 
-// total 19074 errors
+// total 19076 errors
```
kojiromike, line 3: "I need more gif"

Firehed: I hope to snipe that and the related ones with #10156.

PR #9726, opened 2025-12-09
refactor: Extract gateway-agnostic payment JS to standalone file

### LINKED ISSUE #9718: Move payment provider-agnostic JS out of PHP code in user portal
Part of #9705.

Firehed:
Fixes #9718

Moves common JS out of PHP rendering into a standalone `.js` file in the client portal's payments screen.

Part of a broader refactoring effort to ease future development. There are **no logic or formatting changes** to the code at this time, just a cut-and-paste. 

No


portal/portal_payment.js
```diff
@@ -0,0 +1,224 @@
...
+                else if (paying == patient_balance) {
+                    document.getElementById('paying_' + i).style.background = '#ffffff';
+                }
+            }
+            else {
+                document.getElementById('paying_' + i).style.background = '#ffffff';
+            }
+        }
+        else {
+            break;
```
kojiromike, line 22: Thoughts on making this a guard?

portal/portal_payment.js
```diff
@@ -0,0 +1,224 @@
...
+        var elem = f.elements[i];
+        var ename = elem.name;
+        if (ename.indexOf('form_upay[') == 0) {
+            elem.focus();
+            break;
+        }
+    }
+}
+
+function make_it_hide_enc_pay() {
```
kojiromike, line 50: Any chance you could use the jquery starts with selector? If not `id^='td_head_'` then maybe at least to avoid the loop?

Firehed: I'm in favor of such a change but would prefer to do it separately since this is currently just moving code around. If you're ok with it I'll open a separate issue to track it (same for above)

portal/portal_payment.js
Firehed: I'm omitting any changes to the JS to keep this easy to test for now, but I'm in favor of this kind of cleanup in a future pass.

portal/portal_payment.php
```diff
@@ -531,6 +531,7 @@ function notifyPatient() {
     </style>
     <script src="<?php echo $globalsBag->getString('assets_static_relative'); ?>/jquery-creditcardvalidator/jquery.creditCardValidator.js"></script>
     <script src="<?php echo $globalsBag->getString('webroot') ?>/library/textformat.js?v=<?php echo $v_js_includes; ?>"></script>
+    <script src="portal_payment.js"></script>
```
bradymiller, line 534: rec considering incorporating cache buster:
```
<script src="portal_payment.js?v=<?php echo $v_js_includes; ?>"></script>
```

Firehed: #10156, thanks for the callout!

PR #9729, opened 2025-12-10
Use QueryUtils::generateId() over direct access to `adodb`

### LINKED ISSUE #9728: Prefer generate_id function over inlined equivalents
While looking at some database connection stuff, I noticed a handful of places that are duplicating the logic contained within the `generate_id()` function. Updating that code to use the existing function keeps the logic easier to follow and reduces direct touch-points to the database.

Firehed:
Fixes #9728

Replaces `$conn->GenID("sequences")` calls with `QueryUtils::generateId()`, which does the same thing. I've also updated most calls to `generate_id()` to use that same method.

This is purely a refactor - no logic changes. Reducing access to `$GLOBALS` will ease future improvements to database interactions. While `generate_id` still relies on globals, this makes iterative progress towards encapsulating it.

I've also added a PHPStan rule per @kojiromike's suggestion that should prevent more use of the old path going forward.

No


kojiromike:
@Firehed care to write a phpstan check for this? (Also the tests should pass if you rebase.)

I should say, if we can write a rector rule it'd be even better, but a phpstan rule if rector is too complex.

Firehed:
Sure, I'll have a go at it.

PR #9743, opened 2025-12-10
various fixes and cleanup

### LINKED ISSUE #9644: New Dashboard Context Manager
The Problem I’m Solving

We’ve all been there - the patient dashboard shows everything at once, but depending on your role or the type of visit, you really only need to see specific information. An emergency visit needs vitals and allergies front and center. A routine primary care visit benefits from the full picture. Behavioral health focuses on different elements than pediatrics.

Currently, everyone sees the same dashboard regardless of context. That’s a lot of scrolling and visual clutter when you’re trying to focus on what matters for this patient, this visit.

What Dashboard Context Manager Does

This module lets you switch between predefined “care contexts” that show only the widgets relevant to your current workflow.

sjpadgett:
peer review
escaping and cleanup

Fixes #9644

### DIFF
--- a/interface/modules/custom_modules/oe-module-dashboard-context/src/Bootstrap.php
+++ b/interface/modules/custom_modules/oe-module-dashboard-context/src/Bootstrap.php
@@ -14,6 +14,7 @@ namespace OpenEMR\Modules\DashboardContext;
 
 use OpenEMR\Common\Logging\SystemLogger;
 use OpenEMR\Events\PatientDemographics\RenderEvent;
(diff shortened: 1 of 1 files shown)


sjpadgett:
@kojiromike I can make this module 7.0.3 compatible if it is worth the effort. It may force folks to upgrade to 7.0.4 if they really want the feature. Let me know while I have it opened back up for peer review

kojiromike:
@sjpadgett I'm not prepared to say who would want it in 7.0.3 right now, but I think this is related to my advocating for making it a composer-installable module in any case, because that has all kinds of additional benefits besides being able to install it in other versions of OpenEMR.

Issue #9757, opened 2025-12-10
Dependencies - Remove unused - waryway/php-traits-library


igormukhingmailcom:
Dependency `waryway/php-traits-library` is used only in one place and I believe we can just include similar code inside project and remove dependency to have smaller dependency list.

PR #9759, opened 2025-12-11
chore(deps): remove waryway/php-traits-library

igormukhingmailcom:
Fixes: https://github.com/openemr/openemr/issues/9757 by implementing own (a bit better and up-to-date version of `SingletonTrait`)

## Motivation

Main motivation was to remove library we using just in one place.

But after implementing own version of that trait I realized waryway's one had limitations like private final constructor which means we cannot override it and use `__construct(private readonly string $argument) {}`  best practice.
Also I noticed its pain to test classes with waryway singleton - we had to inject mocks via `ReflectionClass` rather than just pass them via constructor.

Probably that’s why it was not used much.

## Usage

New singleton allows making constructors like this:

```php
class Singleton
{
    use SingletonTrait;

    protected function __construct(
        private readonly string $argument
    ) {}

    protected static function createInstance(): static
    {
        return new self('argument');
    }
}
```

And its very useful:
- Having everything as `private readonly string $argument`  at constructor (no `new Something()`  inside constructor) - is good for testing as we want instantiate new/own instance with mocks
- Having `protected static function createInstance(): static { ... }` allows to quickly initialize everything we need
- When we don’t want singleton to be instantiated directly - we can make constructor protected or even private at `Singleton`  class - not having that strict limitation inside `SingletonTrait` give us flexibility
- This singleton approach help us with quick/easy instantiation - `Singleton::getInstance()` - and avoids services duplicates while following `__construct(private readonly string $argument) {}` best practice before we have DI

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -122,7 +122,6 @@
         "twig/twig": "^3.22.2",
         "twilio/sdk": "^8.10.0",
         "vlucas/phpdotenv": "^5.6.3",
-        "waryway/php-traits-library": "^1.0.4",
         "yubico/u2flib-server": "^1.0.2",
         "zircote/swagger-php": "^3.0"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9ef74f3f920b40474c5b03d34161da9d",
+    "content-hash": "6f0904ea3920bce476256d0a25322759",
     "packages": [
         {
             "name": "academe/authorizenet-objects",
@@ -11584,46 +11584,6 @@
             "time": "2025-12-27T19:49:13+00:00"
         },
         {
-            "name": "waryway/php-traits-library",
-            "version": "1.0.4",
-            "source": {
-                "type": "git",
-                "url": "https://github.com/Waryway/PhpTraitsLibrary.git",
-                "reference": "ec0b630770bdd84168d2cd5e61bd93f45095e035"
-            },
-            "dist": {
-                "type": "zip",
-                "url": "https://api.github.com/repos/Waryway/PhpTraitsLibrary/zipball/ec0b630770bdd84168d2cd5e61bd93f45095e035",
-                "reference": "ec0b630770bdd84168d2cd5e61bd93f45095e035",
-                "shasum": ""
-            },
-            "require-dev": {
-                "phpunit/phpunit": "6.4.3"
-            },
-            "type": "library",
-            "autoload": {
-                "psr-4": {
-                    "Waryway\\PhpTraitsLibrary\\": "src/"
-                }
-            },
-            "notification-url": "https://packagist.org/downloads/",
-            "license": [
-                "MIT"
-            ],
-            "authors": [
-                {
-                    "name": "Kyle Wiering",
-                    "email": "kyle@waryway.com"
-                }
-            ],
-            "description": "General use traits",
-            "support": {
-                "issues": "https://github.com/Waryway/PhpTraitsLibrary/issues",
-                "source": "https://github.com/Waryway/PhpTraitsLibrary/tree/1.0.4"
-            },
-            "time": "2017-12-23T14:22:07+00:00"
-        },
-        {
             "name": "webimpress/safe-writer",
             "version": "2.2.0",
             "source": {
@@ -14974,5 +14934,5 @@
         "ext-zlib": "8.2",
         "php": "8.2"
     },
-    "plugin-api-version": "2.9.0"
+    "plugin-api-version": "2.6.0"
 }


src/Common/Logging/EventAuditLogger.php
```diff
@@ -36,17 +33,18 @@ class EventAuditLogger
...
+    private ?bool $breakglassUser = null;
+
+    protected static function createInstance(): static
     {
-        if (!isset($this->cryptoGen)) {
-            $this->cryptoGen = new CryptoGen();
-        }
-        return $this->cryptoGen;
+        return new self(
+            new CryptoGen()
```
kojiromike: Because encryption may not be in use, we don't want to instantiate CryptoGen at all unless we need it. That's why this was lazy before.

src/Common/Logging/EventAuditLogger.php
```diff
@@ -36,17 +34,18 @@ class EventAuditLogger
...
-        if (!isset($this->cryptoGen)) {
-            $this->cryptoGen = new CryptoGen();
-        }
-        return $this->cryptoGen;
+        return new self(
+            new CryptoGen(),
+        );
+    }
+
+    public function __construct(
```
kojiromike, line 46: If the constructor is public, what prevents someone from making a new instance?

igormukhingmailcom: Having it public allow easier testing and some flexibility in other cases.
I would not force singletons to have private constructors. At least for now.

src/Core/Traits/SingletonTrait.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+/**
+ * @package   OpenEMR
+ *
+ * @link      http://www.open-emr.org
```
kojiromike, line 6: Please add this link as well:

```suggestion
 * @link      http://www.open-emr.org
 * @link      https://opencoreemr.com/
```

tests/Tests/Isolated/Core/Traits/SingletonTraitTest.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+/**
+ * @package   OpenEMR
+ *
+ * @link      http://www.open-emr.org
```
kojiromike, line 6:
```suggestion
 * @link      http://www.open-emr.org
 * @link      https://opencoreemr.com/
```

igormukhingmailcom:
@kojiromike Any chance to merge this one?

kojiromike:
Yeah, OK. To be honest I still have reservations about this change, but I decided this is pragmatic and appropriate right now. That said, let me lay down what I'm thinking.

*Con:*

This isn't really a singleton. We're telling people we have a singleton trait, but there's no enforcement, and it'd be pretty easy to accidentally introduce a public constructor in an implementation. That could lead to a misunderstanding at some point.

*Pro:*

It's going to fix some obnoxious problems, particularly with testability.

> Probably that’s why it was not used much.

A Singleton should _not_ be used much, and we should not encourage the use of things that keep global state hanging around everywhere. But in OpenEMR land, this is currently the least of our problems, so I'm good with it.

kojiromike:
It occurs to me that we could add a phpstan rule to flag if anyone creates a public constructor on a class that uses this singleton. Probably overkill, but it'd be a nice clean static way to prevent accidental overloading without enforcing it at runtime.

composer.json
```diff
@@ -122,7 +122,6 @@
         "twig/twig": "^3.22.2",
         "twilio/sdk": "^8.10.0",
         "vlucas/phpdotenv": "^5.6.3",
-        "waryway/php-traits-library": "^1.0.4",
```
bradymiller, line 125: fare thee well :)

igormukhingmailcom:
> Yeah, OK. To be honest I still have reservations about this change, but I decided this is pragmatic and appropriate right now. That said, let me lay down what I'm thinking.
> 
> _Con:_
> 
> This isn't really a singleton. We're telling people we have a singleton trait, but there's no enforcement, and it'd be pretty easy to accidentally introduce a public constructor in an implementation. That could lead to a misunderstanding at some point.
> 
> _Pro:_
> 
> It's going to fix some obnoxious problems, particularly with testability.
> 
> > Probably that’s why it was not used much.
> 
> A Singleton should _not_ be used much, and we should not encourage the use of things that keep global state hanging around everywhere. But in OpenEMR land, this is currently the least of our problems, so I'm good with it.

Agree its not true singleton, but I think about it as **temporary tool** that make life easier, help transitioning to DI and after implementing DI - we can remove this at all.

Right now I use it to be able to quickly instantiate single instance with all dependencies (and it become a bit less pain after using this approach tbh):

```php
namespace OpenEMR\Validators\Checker;

use OpenEMR\Common\Database\Repository\User\UserRepository;
use OpenEMR\Core\Traits\SingletonTrait;

class UserEmailChecker
{
    use SingletonTrait;

    protected static function createInstance(): static
    {
        return new self(
            UserRepository::getInstance(),
        );
    }

    public function __construct(
        private readonly UserRepository $userRepository,
    ) {
    }

    public function isEmailTaken(string $email): bool
    {
        return 0 !== $this->userRepository->countBy(['email' => $email]);
    }
}
```

Later, when we have DI and autowiring, - we just remove singleton trait and instantiation from every class and have everything working as expected:

```php
namespace OpenEMR\Validators\Checker;

use OpenEMR\Common\Database\Repository\User\UserRepository;
use OpenEMR\Core\Traits\SingletonTrait;

class UserEmailChecker
{
    public function __construct(
        private readonly UserRepository $userRepository,
    ) {
    }

    public function isEmailTaken(string $email): bool
    {
        return 0 !== $this->userRepository->countBy(['email' => $email]);
    }
}
```

igormukhingmailcom:
> It occurs to me that we could add a phpstan rule to flag if anyone creates a public constructor on a class that uses this singleton. Probably overkill, but it'd be a nice clean static way to prevent accidental overloading without enforcing it at runtime.

Yeah, phpstan rule will be helpful for soft-deny direct instantiating singletons (via `new ClassName`) at main codebase, but allowing it at tests (as that is the main idea/reason to have public constructor - be able to pass mocks to constructor at tests).

Would you like me to implement that rule?

PR #9760, opened 2025-12-11
refactor(twig): Turned TwigContainer into TwigFactory

### LINKED ISSUE #9704: Should TwigContainer be a singleton?
Instantiating TwigContainer is apparently fairly heavy, and it doesn't need to vary on state. Perhaps we should find a way to make sure it doesn't get instantiated in different ways in different places.

igormukhingmailcom:
Fixes: https://github.com/openemr/openemr/issues/9704

## Merging notes

- [x] Should be rebased and merged after https://github.com/openemr/openemr/pull/9759
- [ ] Should be rebased after merging https://github.com/openemr/openemr/pull/10514


igormukhingmailcom:
@kojiromike

I can see that our baseline placed at separate files like `phpstan-globals-baseline.neon`, etc.
Is that means we can't update them automatically?

I have 230 errors (as omitted `$GLOBALS` usage at many places) and trying to understand how to do baseline update faster...

```bash
Ignored error pattern #^Direct access to \$GLOBALS is forbidden\. Use OEGlobalsBag\:\:getInstance\(\)\->get\(\) instead\.$# (openemr.forbiddenGlobalsAccess) in path ...  is expected to occur 8 times, but occurred only 7 times
         
[ERROR] Found 230 errors 
```

kojiromike:
Try doing a phpstan check just for that rule. I think

```
php -d memory_limit=8g ./vendor/bin/phpstan analyze \
    --configuration=phpstan.neon.dist \
    --identifier=openemr.forbiddenGlobalsAccess \
    --generate-baseline=.phpstan/phpstan-globals-baseline.neon
```

src/Common/Twig/TwigContainer.php
```diff
@@ -1,65 +1,67 @@
...
+        );
+    }
 
-        if (null !== $kernel) {
-            $this->kernel = $kernel;
-        }
+    protected function __construct(
+        private readonly Kernel $kernel,
+        private array $paths,
+    ) {
```
kojiromike: With the singleton, `addPath()` permanently mutates the shared instance — paths added by one caller (e.g., AuthorizationController adding `/oauth2/`) persist for every subsequent `getTwig()` call from any other caller. Before, each `new TwigContainer(...)` had isolated paths.

Is this accumulation intentional? It's probably fine in practice since template names tend to be unique per module, but it's a behavior change from the isolated-instance model. If two modules happened to have a template with the same filename, the first one loaded would win.

src/Common/Twig/TwigContainer.php
```diff
@@ -1,65 +1,67 @@
...
-    public function __construct(?string $path = null, ?Kernel $kernel = null)
+    protected static function createInstance(): static
     {
-        $this->paths[] = $GLOBALS['fileroot'] . '/templates';
-
-        if (!empty($path)) {
-            $this->addPath($path);
-        }
+        $globals = OEGlobalsBag::getInstance();
+        return new self(
```
kojiromike: Is the singleton pulling its weight here? Since `OEGlobalsBag` is already a singleton, constructing a new `TwigContainer` from it is cheap — there's no heavy initialization being cached. And `addPath()` mutates the shared instance, so paths added by one caller (e.g., AuthorizationController adding `/oauth2/`) persist for every subsequent caller.

A static factory method like `TwigContainer::create()` would give the same ergonomic win (no `$GLOBALS['kernel']` boilerplate at 244 call sites) while keeping each caller's paths isolated — which is how it worked before.

Not blocking — the change is consistent and CI passes — but 246 files is a big diff for a pattern that doesn't clearly improve on a factory method. Worth considering whether the churn is justified.

kojiromike (commented):
The mechanical replacement is clean and CI is green, but I'm not sure the singleton is the right pattern here — a static factory would achieve the same thing without the shared mutable state. See inline comment.

src/Common/Twig/TwigContainer.php
igormukhingmailcom: You're right

Even more - just realized `TwigContainer` should be turned into `TwigFactory`, as we don't need instance of stateful `TwigContainer` (where more and more paths can be added before calling `getTwig()`). 

`TwigFactory::createInstance` now bootstraping Twig Environment with passed paths:

```php
/**
 * Usage:
 *   TwigFactory::createInstance()->render('template.html.twig', ['var' => 'value]); // Default path
 *   TwigFactory::createInstance(__DIR__)->render('template.html.twig', ['var' => 'value]); // __DIR__ as additional path
 *   TwigFactory::createInstance([$path1, $path2])->render('template.html.twig', ['var' => 'value]); // Multiple additional paths
 */
```

PR #9864, opened 2025-12-15
fix(dclint): npx hides arguments from command

kojiromike:
Switch to docker to avoid npx entanglements.

From #9776

### DIFF
--- a/.github/workflows/docker-compose-lint.yml
+++ b/.github/workflows/docker-compose-lint.yml
@@ -31,8 +31,6 @@ jobs:
   dclint:
     name: Docker Compose Linting
     runs-on: ubuntu-latest
-    container:
-      image: zavoloklom/dclint:3.1.0
     steps:
     - name: Checkout code
       uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


kojiromike:
@copilot using `--` to separate npx args from dclint args didn't resolve the issue. If I run the exact npx commandline that I see in the workflow output, either

```
npx dclint@3.1.0 --exclude .github ci/ docker/
```

or the specific changed-files case, maybe something like:

```
npx dclint@3.1.0 --exclude .github ci/inferno/compose.yml
```

It works fine locally, but in github actions, we get the "usage" output for dclint, followed by `Not enough non-option arguments: got 0, need at least 1`.

There are also these other annotations in the workflow, although only the dclint step is showing a failure exit status:

```
Please ensure that submodules are initialized and up to date. See: https://github.com/actions/checkout#usage
Get changed docker compose files
fatal: Invalid revision range b6541e68afde0c465205dccb9feb6741f053e2c4..b3f53874d72d0466c46949234a867375acfb63dd
Get changed docker compose files
Set 'fetch_additional_submodule_history: true' to fetch additional submodule history for: ci/inferno/inferno-files
Get changed docker compose files
Unable to find merge base between b6541e68afde0c465205dccb9feb6741f053e2c4 and b3f53874d72d0466c46949234a867375acfb63dd
Get changed docker compose files
Unable to find merge base between b4e671dadacc4ade833c9ee32c9968c259a7cdc0 and 51e6c802ee289ea8a60a5b5d8453692d83b3b9d6
```

Please resolve these issues. If you know what to do to fix it, go ahead. Otherwise, try replacing npx and setup-node, and instead run the job using the `zavoloklom/dclint` docker image.

Issue #9867, opened 2025-12-15
Adapt `OEGlobalsBag` to implement PSR-11


Firehed:
I'd like to start a discussion about leveraging PSR-11 (and, separately, other [PHP-FIG](https://www.php-fig.org)-originated standards) in OpenEMR.

Reference: [PSR-11 spec](https://www.php-fig.org/psr/psr-11/)

Why:

- Code that implements an interface becomes easier to test and integrate into other tests, since that interface can be trivially mocked and controlled using native tooling
- Code referencing an interface rather than a concrete class is more loosely coupled; this provides a lot of added flexibility when making changes. This adds the ability (but not requirement!) to, for example, use a different container entirely without any changes to the code that relies on it
- `PSR-11` specifically is part of the `FIG` group of standards, which is well-supported across the PHP ecosystem. This can simplify adoption of vetted libraries

What this process _could_ look like:

1) Declare `OEGlobalsBag` as conforming to `Psr\Container\ContainerInterface` (it already de-facto does, though the `$default` parameter on `get()` is non-conformant but not problematic at runtime)
2) Where possible, update typehints for `OEGlobalsBag` to be `ContainerInterface`. Where not possible (there's a number of possible reasons for this), track these places in an issue for future updates. Discourage new code from referencing OEGlobalsBag in favor of ContainerInterface.
3) Iteratively fix issues revealed during step 2. At a glance, this looks mostly like using the `set()` method (which should happen in application bootstrapping) and reliance on the `$default` parameter on `get()` (update to use `has()`)

Long-term, the goal I'd propose would be to have new services follow the general pattern of dependency injection. Doing so _dramatically_ increases code testability, and makes the application more reliable and easier to develop. This could be greatly eased by adopting an "auto-wiring" container, which simplifies both reading and writing code. _This_ is a bit out of scope for the initial discussion.

---

As the code exists today, step 1 is basically free. OpenEMR already implicitly pulls in the `psr/container` package, so the whole change should be ~3LOC. Most of the interesting discussion would be around next steps & long-term goals.

kojiromike:
cc @adunsulag for comment

adunsulag:
[...]

PR #9868, opened 2025-12-15
Replace $GLOBALS with OEGlobalsBag in Portal

zmilan:
Fixes #

Replace `$GLOBALS[]` with `OEGlobalsBag` in the Portal

### DIFF
--- a/.phpstan/phpstan-remaining-baseline.neon
+++ b/.phpstan/phpstan-remaining-baseline.neon
@@ -15620,19 +15620,11 @@ parameters:
       identifier: variable.undefined
       count: 2
       path: ../portal/add_edit_event_user.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 3
-      path: ../portal/add_edit_event_user.php
     - message: '#^Variable \$starttime might not be defined\.$#'
       identifier: variable.undefined
       count: 1
       path: ../portal/add_edit_event_user.php
     - message: '#^Variable \$providerid might not be defined\.$#'
-      identifier: variable.undefined
-      count: 1
-      path: ../portal/find_appt_popup_user.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
       identifier: variable.undefined
       count: 1
       path: ../portal/find_appt_popup_user.php
@@ -15657,10 +15649,6 @@ parameters:
       count: 1
       path: ../portal/get_medications.php
     - message: '#^Variable \$pid might not be defined\.$#'
-      identifier: variable.undefined
-      count: 1
-      path: ../portal/get_patient_documents.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
       identifier: variable.undefined
       count: 1
       path: ../portal/get_patient_documents.php
@@ -15704,10 +15692,6 @@ parameters:
       identifier: variable.undefined
       count: 8
       path: ../portal/home.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 2
-      path: ../portal/home.php
     - message: '#^Variable \$defaultLangID might not be defined\.$#'
       identifier: variable.undefined
       count: 1
@@ -15736,14 +15720,6 @@ parameters:
       identifier: variable.undefined
       count: 3
       path: ../portal/index.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 2
-      path: ../portal/lib/doc_lib.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 1
-      path: ../portal/lib/patient_groups.php
     - message: '#^Variable \$ccaudit might not be defined\.$#'
       identifier: variable.undefined
       count: 1
@@ -15764,10 +15740,6 @@ parameters:
       identifier: empty.variable
       count: 1
       path: ../portal/messaging/handle_note.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 1
-      path: ../portal/messaging/handle_note.php
     - message: '#^Variable \$pid might not be defined\.$#'
       identifier: variable.undefined
       count: 1
@@ -15779,10 +15751,6 @@ parameters:
     - message: '#^Variable \$resultusers on left side of \?\? always exists and is not nullable\.$#'
       identifier: nullCoalesce.variable
       count: 1
-      path: ../portal/messaging/messages.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 3
       path: ../portal/messaging/messages.php
     - message: '#^Instantiated class OpenEMR\\PatientPortal\\Chat\\ChatController not found\.$#'
       identifier: class.notFound
@@ -16024,18 +15992,10 @@ parameters:
       identifier: variable.undefined
       count: 1
       path: ../portal/portal_payment.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 5
-      path: ../portal/portal_payment.php
     - message: '#^Variable \$v_js_includes might not be defined\.$#'
       identifier: variable.undefined
       count: 1
       path: ../portal/portal_payment.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 1
-      path: ../portal/report/document_downloads_action.php
     - message: '#^Variable \$adj_amt might not be defined\.$#'
       identifier: variable.undefined
       count: 1
@@ -16071,10 +16031,6 @@ parameters:
     - message: '#^Variable \$prevIssueType might not be defined\.$#'
       identifier: variable.undefined
       count: 1
-      path: ../portal/report/portal_custom_report.php
-    - message: '#^Variable \$srcdir might not be defined\.$#'
-      identifier: variable.undefined
-      count: 8
       path: ../portal/report/portal_custom_report.php
     - message: '#^Variable \$tmp_files_remove might not be defined\.$#'
       identifier: variable.undefined


portal/home.php
```diff
@@ -36,6 +31,19 @@
 use Twig\Error\RuntimeError;
 use Twig\Error\SyntaxError;
 
+// Need access to classes, so run autoloader now instead of in globals.php.
+require_once(__DIR__ . "/../vendor/autoload.php");
+$globalsBag = OEGlobalsBag::getInstance(true);
+$globalsBag->set('already_autoloaded', true);
+
+require_once('verify_session.php');
+require_once("$srcdir/patient.inc.php");
```
kojiromike, line 42: Can we use $globalsBag->get('srcdir') here?

portal/questionnaire_render.php
```diff
@@ -12,12 +12,15 @@
...
 use OpenEMR\Core\Header;
+use OpenEMR\Core\OEGlobalsBag;
 use OpenEMR\Services\QuestionnaireService;
 
 // block of code to securely support use by the patient portal
 // Need access to classes, so run autoloader now instead of in globals.php.
-$GLOBALS['already_autoloaded'] = true;
 require_once(__DIR__ . "/../vendor/autoload.php");
+$globalsBag = OEGlobalsBag::getInstance(true);
+$globalsBag->set('already_autoloaded', true);
```
kojiromike: If globals aren't otherwise used in this file, why does this need to be set at all?

portal/verify_session.php
```diff
@@ -27,12 +27,15 @@
...
 use OpenEMR\Common\Session\SessionUtil;
+use OpenEMR\Core\OEGlobalsBag;
 
 //continue session
 // Will start the (patient) portal OpenEMR session/cookie.
 // Need access to classes, so run autoloader now instead of in globals.php.
-$GLOBALS['already_autoloaded'] = true;
 require_once(__DIR__ . "/../vendor/autoload.php");
+$globalsBag = OEGlobalsBag::getInstance(true);
+$globalsBag->set('already_autoloaded', true);
```
kojiromike: If globals aren't otherwise used in this file, why does this need to be set at all?

portal/questionnaire_render.php
zmilan: I was removing each usage of the `$GLOBALS`, but it is not mandatory. Do you suggest not changing such files?

portal/verify_session.php
zmilan: I was following a "No $GLOBALS" rule, but you are right... I will revert changes in such cases.

portal/questionnaire_render.php
kojiromike: My question is about whether globals are needed in this file at all.

zmilan: It was there only to set `already_autoloaded`, but with `require_once`, I think it is not really needed

portal/messaging/messages.php
```diff
@@ -51,7 +53,8 @@
 use OpenEMR\Core\Header;
 use OpenEMR\Events\Messaging\SendSmsEvent;
 
-if (!(isset($GLOBALS['portal_onsite_two_enable'])) || !($GLOBALS['portal_onsite_two_enable'])) {
+$portal_onsite_two_enable = $globalsBag->get('portal_onsite_two_enable');
```
kojiromike: GlobalsBag is a subclass of symfony parameterbag, which has a `getBoolean`. We should use that when we know a thing is a boolean.

zmilan: Sure, I will change it now

zmilan: Fixed in all places

portal/sign/assets/signer_modal.php
```diff
@@ -11,11 +11,13 @@
...
 
 use OpenEMR\Common\Session\SessionUtil;
+use OpenEMR\Core\OEGlobalsBag;
 
 // this script is used by both the patient portal and main openemr; below does authorization.
 // Need access to classes, so run autoloader now instead of in globals.php.
-$GLOBALS['already_autoloaded'] = true;
 require_once(__DIR__ . "/../../../vendor/autoload.php");
+$globalsBag = OEGlobalsBag::getInstance(true);
+$globalsBag->set('already_autoloaded', true);
```
Firehed, line 20: I was looking at removing all of the `already_autoloaded` paths in a separate PR, since it's already handled via `require_once`. I don't think there's any technical need to move it into globalsbag - but if you've found that it _is_ required I'd love to know why :)

portal/report/portal_custom_report.php
```diff
@@ -39,6 +41,9 @@
 $ignoreAuth_onsite_portal = true;
 global $ignoreAuth_onsite_portal;
 
+$web_root = $globalsBag->getString('web_root');
+
+$srcdir = $globalsBag->getString('srcdir');
 require_once('../../interface/globals.php');
```
Firehed, line 47: I'm going mostly on memory/intuition here rather than actually having run the code in your PR, but wouldn't this line need to happen before the `OEGlobalsBag::getInstance(true)` above so everything is populated?

(this applies to several files)

portal/portal_payment.php
```diff
@@ -1296,8 +1298,8 @@ function getAuth() {
                     <!--<button type="button" class="close" data-dismiss="modal">&times;</button>-->
                 </div>
                 <div class="modal-body">
-                    <?php if ($GLOBALS['payment_gateway'] != 'Stripe' && $GLOBALS['payment_gateway'] != 'Sphere') { ?>
-                    <form id='paymentForm' method='post' action='<?php echo $GLOBALS["webroot"] ?>/portal/lib/paylib.php'>
+                    <?php if ($globalsBag->get('payment_gateway') !== 'Stripe' && $globalsBag->get('payment_gateway') !== 'Sphere') { ?>
```
Firehed, line 1305: Note: these may conflict with some other open PRs that are refactoring this logic, but don't worry about it here. Shouldn't be a difficult merge if it comes up at all.

kojiromike:
I suppose we could defer this to another PR, but I would appreciate a look for other cases where using typed getters would make sense. ParameterBag supports `getBoolean`, `getInt`, `getString`, `getEnum`, etc.

https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpFoundation/ParameterBag.php

Also, aside, @zmilan @Firehed should we consider adding a few specific typed results ourselves? For example, a typed `getKernel` might be useful, no?

Firehed:
> should we consider adding a few specific typed results ourselves? For example, a typed getKernel might be useful, no?

Useful, certainly, but I'm inclined to avoid this in favor of a proper DI container (#9867, etc). Seeing that OEGlobalsBag already contains a handful of services in addition to pure config values, it'd make tracking down service access easier if/when that time comes. If my grep-fu is on point, there's not a whole lot of code accessing `kernel` from the bag and I don't really want to encourage that behavior by making it easier.

Not really sure where that balancing act lies, but in general I'd rather see energy going towards DI+service autowiring over maintaining the globals bag if we have to make a choice.

Regardless, I'd punt on it from _this_ PR.

kojiromike:
> I'd rather see energy going towards DI+service autowiring over maintaining the globals bag if we have to make a choice.


From the _once bitten, twice shy_ department:

I 100% agree that a DI container is the better way to architect this, but that's going to be a lift and we'll probably find some other things we want to change while we're working on it.

OTOH, adding a `getKernel` method with type declarations is practically free and it's already in GLOBALS, just not typed. There may be a handful of other things in GLOBALS that are used everywhere, specific types (not primitives) and easy to add to globalsBag.

> Regardless, I'd punt on it from _this_ PR.

That's completely understandable.

Firehed:
Added #9942 to discuss further, but after thinking on it a big longer I tend to agree it's probably worth it in the short term.

portal/patient/fwk/libs/verysimple/Phreeze/Reporter.php
kojiromike:
```suggestion
    private static array $PublicPropCache = [];
```

This should make rector happy.

zmilan: I am not 100% sure, but I think it still does it. I was playing with it on the Portal Symfony Session branch, since Rector broke it a few times while I was working on it...

zmilan:
> I suppose we could defer this to another PR, but I would appreciate a look for other cases where using typed getters would make sense. ParameterBag supports `getBoolean`, `getInt`, `getString`, `getEnum`, etc.
> 
> https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpFoundation/ParameterBag.php
> 
> Also, aside, @zmilan @Firehed should we consider adding a few specific typed results ourselves? For example, a typed `getKernel` might be useful, no?

Just to be sure that I understand, the suggestion is to add `getKernel()` without parameters and type response to `OpenEMR\Core\Kernel`?

portal/sign/assets/signer_modal.php
zmilan: I agree about removing `already_autoloaded`.

portal/report/portal_custom_report.php
zmilan: I was not sure if there are some cases where we don't wish to use `compatibilityMode`, and in some of files it was already without it so I was trying to follow... as far as I tested it works, but if there is any special reason for not using `compatibilityMode` I will change it and use it everywhere

zmilan:
@kojiromike I added `getKernel()` and `setKernel(Kernel)` to the `OEGGlobalsBag` and replaced all usage in the Portal code.

portal/patient/fwk/libs/verysimple/Phreeze/Reporter.php
zmilan: I was wrong, that really solved it... thanks

kojiromike:
> @kojiromike I added `getKernel()` and `setKernel(Kernel)` to the `OEGGlobalsBag` and replaced all usage in the Portal code.

@zmilan sorry, I guess there's some controversy about that. So as not to hold up this PR, please revert it and we can address it in a different scope.

kojiromike:
@zmilan I think you can regenerate a baseline for a single phpstan rule by adapting this code:

```
php -d memory_limit=8g ./vendor/bin/phpstan analyze \
    --configuration=phpstan.neon.dist \
    --identifier=openemr.forbiddenGlobalsAccess \
    --generate-baseline=.phpstan/phpstan-globals-baseline.neon
# and because neon uses tabs for indentation
expand .phpstan/phpstan-globals-baseline.neon |
    yq -P . > temp
mv temp .phpstan/phpstan-globals-baseline.neon
```

portal/lib/paylib.php
```diff
@@ -16,7 +16,6 @@
 
 // Will start the (patient) portal OpenEMR session/cookie.
 // Need access to classes, so run autoloader now instead of in globals.php.
-$GLOBALS['already_autoloaded'] = true;
```
bradymiller, line 19: @kojiromike and @zmilan , It probably makes sense to completely remove this `$GLOBALS['already_autoloaded'] = true;` mechanism completely in codebase since where it is called again in globals.php is using a require_once. I think I was being overly cautious when i added this mechanism (added it when placing the Session stuff under umbrella of autoload.php rather than previously Session was isolated and called prior to autoload.php).

portal/patient/libs/Controller/OnsitePortalActivityController.php
```diff
@@ -108,11 +111,10 @@ public function Read()
             $pk = $this->GetRouter()->GetUrlParam('id');
             $onsiteportalactivity = $this->Phreezer->Get('OnsitePortalActivity', $pk);
             // only allow patient to update onsiteportalactivity about themself
-            if (!empty($GLOBALS['bootstrap_pid'])) {
-                if ($GLOBALS['bootstrap_pid'] != $onsiteportalactivity->PatientId) {
-                    $error = 'Unauthorized';
-                    throw new Exception($error);
-                }
+            $bootstrapPid = OEGlobalsBag::getInstance(true)->get('bootstrap_pid');
```
bradymiller, line 114: @kojiromike and @zmilan , I probably missed further discussion on this, but I thought that there was discussion that unable to have a singleton with a parameter in the constructor.

kojiromike: @bradymiller it's true. The (de-facto) fix is in https://github.com/openemr/openemr/pull/9880. We gotta get that merged. I'll get on it.

portal/sign/assets/signer_modal.php
bradymiller: just saw this and agree.

bradymiller:
super neat stuff!

PR #9874, opened 2025-12-15
fax service and eRx titles

### LINKED ISSUE #9866: Fax service new vendor add
As a developer, I create a better user experience for faxing, so I added a new vendor that is more self service. This will allow clinics to get up and running faster.

juggernautsei:
Fixes #9866

### DIFF
--- a/interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
@@ -24,6 +24,7 @@ require_once(__DIR__ . "/../../../../globals.php");
 use OpenEMR\Common\Crypto\CryptoGen;
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Http\oeHttp;
+use OpenEMR\Common\Http\oeHttpRequest;
 use OpenEMR\Modules\FaxSMS\Controller\FaxDocumentService;
 
 // Capture raw input first (can only be read once)
@@ -80,9 +81,9 @@ if ($vendor !== 'signalwire' || $type !== 'fax') {
     exit('Invalid request');
 }
 
-// Get webhook payload
+// Get webhook payload - initialize all variables explicitly for PHPStan
 $faxSid = $_POST['FaxSid'] ?? $_POST['Sid'] ?? '';
-$status = $_POST['Status'] ?? $_POST['FaxStatus'] ?? '';
+$status = $_POST['Status'] ?? $_POST['FaxStatus'] ?? 'unknown';
 $from = $_POST['From'] ?? $_POST['RemoteStationId'] ?? '';
 $to = $_POST['To'] ?? $_POST['OriginalTo'] ?? '';
 $numPages = $_POST['NumPages'] ?? $_POST['Pages'] ?? 0;
@@ -207,7 +208,7 @@ function downloadAndStoreFaxMedia(
         // Download the fax media with authentication using oeHttp
         // SignalWire files.signalwire.com requires Bearer token, not Basic auth
         try {
-            $httpRequest = oeHttp::newArgs(oeHttp::client());
+            $httpRequest = oeHttpRequest::newArgs(oeHttp::client());
 
             // Set up headers based on URL type
             if (str_contains($mediaUrl, 'files.signalwire.com')) {
@@ -225,7 +226,8 @@ function downloadAndStoreFaxMedia(
             $response = $httpRequest->get($mediaUrl);
             $httpCode = $response->status();
             $mediaContent = $response->body();
-            $contentType = $response->header('Content-Type') ?? 'application/pdf';
+            $contentTypeHeader = $response->header('Content-Type');
+            $contentType = !empty($contentTypeHeader) ? $contentTypeHeader : 'application/pdf';
 
             if ($httpCode !== 200 || empty($mediaContent)) {
                 return;
@@ -248,7 +250,7 @@ function downloadAndStoreFaxMedia(
             $mediaContent,
             $fromNumber,
             $patientId,
-            $contentType ?? 'application/pdf'
+            $contentType
         );
 
         // Update queue with storage info
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
@@ -387,7 +387,8 @@ abstract class AppDispatch
             $spaceUrl = $this->getRequest('space_url');
             $projectId = $this->getRequest('project_id');
             $apiToken = $this->getRequest('api_token');
-            $faxNumber = $this->formatPhoneForSave($this->getRequest('fax_number'));
+            $faxNumberRaw = $this->getRequest('fax_number');
+            $faxNumber = !empty($faxNumberRaw) ? $this->formatPhoneForSave($faxNumberRaw) : '';
 
             $setup = [
                 'username' => "$username",
@@ -409,7 +410,7 @@ abstract class AppDispatch
                 'space_url' => $spaceUrl ?? '',
                 'project_id' => $projectId ?? '',
                 'api_token' => $apiToken ?? '',
-                'fax_number' => $faxNumber ?? '',
+                'fax_number' => $faxNumber,
             ];
         }
 
@@ -616,6 +617,7 @@ abstract class AppDispatch
      */
     public function mailEmail($email, $from_name, $body, $subject = '', $htmlContent = ''): string
     {
+        $status = 'Error: ' . xlt('Unknown error occurred');
         try {
             $mail = new MyMailer();
             $smtpEnabled = $mail::isConfigured();
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/SignalWireClient.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/SignalWireClient.php
@@ -130,14 +130,17 @@ class SignalWireClient extends AppDispatch
             ]);
         }
 
-        // Get request parameters
+        // Get request parameters - initialize all to avoid PHPStan warnings
         $isContent = $this->getRequest('isContent');
-        $file = $this->getRequest('file');
+        $fileParam = $this->getRequest('file');
+        $file = !empty($fileParam) ? $fileParam : '';
         $docId = $this->getRequest('docid');
-        $phone = $this->formatPhone($this->getRequest('phone'));
+        $phoneParam = $this->getRequest('phone');
+        $phone = !empty($phoneParam) ? $this->formatPhone($phoneParam) : '';
         $recipientName = $this->getRequest('name') . ' ' . $this->getRequest('surname');
         $recipientName = trim($recipientName) ?: 'Unknown'; // Default if empty
-        $isDocuments = (int)$this->getRequest('isDocuments');
+        $isDocumentsParam = $this->getRequest('isDocuments');
+        $isDocuments = !empty($isDocumentsParam) ? (int)$isDocumentsParam : 0;
         $email = $this->getRequest('email');
         $hasEmail = $this->validEmail($email);
         $globals = OEGlobalsBag::getInstance();
@@ -145,17 +148,17 @@ class SignalWireClient extends AppDispatch
         $user = $this::getLoggedInUser();
 
         // DEBUG: Log parameters received in sendFax
-        error_log("SignalWireClient.sendFax(): DEBUG - Received file path: " . ($file ?? 'EMPTY'));
+        error_log("SignalWireClient.sendFax(): DEBUG - Received file path: " . $file);
         error_log("SignalWireClient.sendFax(): DEBUG - isContent: " . ($isContent ?? 'EMPTY'));
-        error_log("SignalWireClient.sendFax(): DEBUG - isDocuments: " . ($isDocuments ?? 'EMPTY'));
-        error_log("SignalWireClient.sendFax(): DEBUG - Phone: " . ($phone ?? 'EMPTY'));
-        error_log("SignalWireClient.sendFax(): DEBUG - File exists: " . (file_exists($file) ? 'YES' : 'NO'));
+        error_log("SignalWireClient.sendFax(): DEBUG - isDocuments: " . $isDocuments);
+        error_log("SignalWireClient.sendFax(): DEBUG - Phone: " . $phone);
+        error_log("SignalWireClient.sendFax(): DEBUG - File exists: " . (!empty($file) && file_exists($file) ? 'YES' : 'NO'));
         if (!empty($file) && file_exists($file)) {
             error_log("SignalWireClient.sendFax(): DEBUG - File size: " . filesize($file) . " bytes");
         }
 
         // Handle file path
-        if (empty($isContent)) {
+        if (empty($isContent) && !empty($file)) {
             if (str_starts_with((string) $file, 'file://')) {
                 $file = substr((string) $file, 7);
             }
@@ -241,7 +244,7 @@ class SignalWireClient extends AppDispatch
         try {
             // DEBUG: Log before upload
             error_log("SignalWireClient.uploadFileForFax(): DEBUG - baseDir: " . ($this->baseDir ?? 'EMPTY'));
-            error_log("SignalWireClient.uploadFileForFax(): DEBUG - File parameter: " . ($file ?? 'EMPTY'));
+            error_log("SignalWireClient.uploadFileForFax(): DEBUG - File parameter: " . $file);
             error_log("SignalWireClient.uploadFileForFax(): DEBUG - isDocuments: " . ($isDocuments ? 'YES' : 'NO'));
 
             // Use public web root for uploads so SignalWire can access via HTTP


interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -0,0 +1,309 @@
+<?php
+
+/**
+ * SignalWire Webhook Receiver
+ * Handles incoming fax notifications from SignalWire
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    SignalWire Integration
+ * @copyright Copyright (c) 2024
```
kojiromike: 2024?

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -0,0 +1,309 @@
...
+use OpenEMR\Modules\FaxSMS\Controller\FaxDocumentService;
+
+// Capture raw input first (can only be read once)
+$rawInput = file_get_contents('php://input');
+
+// Get site ID from query parameter
+$siteId = $_SESSION['site_id'] ?? $_GET['site'] ?? 'default';
+
+// Log all incoming webhook data for debugging
+error_log("=== SignalWire Webhook START ===");
```
kojiromike: I think we should use SystemLogger on any new code going forward. Or an injectable PSR-3 logger (which SystemLogger is one)

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -0,0 +1,309 @@
...
+    if ($direction === 'inbound' && $status === 'received' && !empty($mediaUrl)) {
+        downloadAndStoreFaxMedia($faxSid, $mediaUrl, $from, $siteId, $existingFax['patient_id'] ?? 0);
+    }
+
+    // Respond with success
+    http_response_code(200);
+    header('Content-Type: application/xml');
+    echo '<?xml version="1.0" encoding="UTF-8"?><Response></Response>';
+    
+} catch (Exception $e) {
```
kojiromike: If you want to catch everything, catch Throwable, not Exception.

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -0,0 +1,309 @@
...
+
+        if (empty($projectId) || empty($apiToken)) {
+            error_log("SignalWire Webhook: Invalid credentials for media download");
+            return;
+        }
+
+        // Download the fax media with authentication
+        // SignalWire files.signalwire.com requires Bearer token, not Basic auth
+        $ch = curl_init($mediaUrl);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
```
kojiromike: Please use oeHttpRequest or Guzzle Http client, not curl. Using multiple different http clients leads to trouble and security vulnerabilities.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
kojiromike: This long nested switch block can be reduced into a relatively simple nested configuration array that represents the factories for different telecomclients.

```suggestion
    $factoryMap = [
        'sms' => [
            1 => fn() => new RCFaxClient(),    // The value is a callable
            2 => fn() => new TwilioSMSClient(),
            5 => fn() => new ClickatellSMSClient(),
        ],
        'fax' => [
            1 => fn() => new RCFaxClient(),
            3 => fn() => new EtherFaxActions(),
            6 => fn() => new SignalWireClient(),
        ],
        'email' => [
            // Note: If EmailClient() is a function call, this is where you'd put it
            4 => fn() => EmailClient(),
        ],
        'voice' => [
            6 => fn() => new VoiceClient(),
        ],
    ];
    $factory = $factoryMap[$type][$s] ?? null;
    if (is_callable($factory)) {
        return $factory();
    }
```

interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
```diff
@@ -597,7 +616,11 @@ public static function getLoggedInUser(): array
      */
     public function validEmail($email): bool
     {
-        return ValidationUtils::isValidEmail($email);
+        if (preg_match("/^[_a-z0-9-]+(\.[_a-z0-9-\+]+)*@[a-z0-9-]+(\.[a-z0-9-]+)*(\.[a-z]{2,3})$/i", $email)) {
```
kojiromike: Please don't drop ValidationUtils. We're trying to make the project more consistent with centralizing these things. Is there something wrong with how ValidationUtils does it?

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
```diff
@@ -0,0 +1,377 @@
...
+    {
+        try {
+            // Get fax from queue
+            $fax = sqlQuery(
+                "SELECT * FROM oe_faxsms_queue WHERE job_id = ? AND site_id = ?",
+                [$faxSid, $this->siteId]
+            );
+            
+            if (empty($fax)) {
+                throw new Exception("Fax not found: {$faxSid}");
```
kojiromike: It's important that exceptions be scoped, so you can identify when something breaks, where the breakage comes from. So if your module is going to throw exceptions, they should be a custom exception type that is specific to your module, not a generic `Exception`.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    /**
+     * Get file extension from MIME type
+     *
+     * @param string $mimeType
+     * @return string
+     */
+    private function getExtensionFromMimeType(string $mimeType): string
```
kojiromike: Consider using src/Common/Utils/FileUtils.php instead of defining this yourself.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
```diff
@@ -0,0 +1,377 @@
...
+        return $map[$mimeType] ?? 'pdf';
+    }
+
+    /**
+     * Format phone number for display
+     *
+     * @param string $phone
+     * @return string
+     */
+    private function formatPhoneDisplay(string $phone): string
```
kojiromike, line 312: I don't know if there's a standard way to do this in openemr right now, but there ought to be, so that the presentation is consistent.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
```diff
@@ -0,0 +1,377 @@
...
+        return $phone;
+    }
+
+    /**
+     * Attempt to auto-match fax to patient by phone number
+     *
+     * @param string $fromNumber Sender's phone number
+     * @return int Patient ID if found, 0 otherwise
+     */
+    public function findPatientByPhone(string $fromNumber): int
```
kojiromike, line 333: This, too, should be a standard utility in openemr. While I'm a big fan of us modularizing the application as much as possible, functions like this should be shared so our modules are consistent and as small as possible.

interface/modules/custom_modules/oe-module-faxsms/messageUI.php
```diff
@@ -29,6 +29,7 @@
 $title = $service == "2" ? xlt('Twilio SMS') : $title;
 $title = $service == "3" ? xlt('etherFAX') : $title;
 $title = $service == "4" ? xlt('Email') : $title;
+$title = $service == "6" ? xlt('SignalWire Fax') : $title;
```
kojiromike: It'd be nice to have an enum mapping these numbers to the names.

interface/modules/custom_modules/oe-module-faxsms/WARP.md
```diff
@@ -0,0 +1,674 @@
...
+
+1. **Review the specific issue** - What's not working? Error messages?
+2. **Check relevant controller** - Which vendor service is affected?
+3. **Verify configuration** - Are credentials set and valid?
+4. **Check logs** - OpenEMR logs and vendor API logs
+5. **Test in isolation** - Use vendor's test tools before debugging module code
+
+---
+
+**Last Updated**: December 2024
```
kojiromike: year?

interface/modules/custom_modules/oe-module-faxsms/WARP.md
```diff
@@ -0,0 +1,674 @@
+# WARP.md - OpenEMR FaxSMS Module
```
kojiromike, line 1: I'm confused about WARP.md being in a module that's part of core. Shouldn't the modules be shipped as separate repos and installable via packagist, and then individualized agents instructions makes sense?

interface/modules/custom_modules/oe-module-faxsms/src/Controller/SignalWireClient.php
```diff
@@ -0,0 +1,926 @@
...
+        return $filepath;
+    }
+
+    /**
+     * Format phone number for saving
+     *
+     * @param string $number
+     * @return string
+     */
+    public function formatPhoneForSave($number): string
```
kojiromike, line 932: Hmm, I think this should be a central utility function, outside of modules, too.

interface/modules/custom_modules/oe-module-faxsms/composer.json
kojiromike, line 1: I'm confused about packaging files in modules -- how do those get included when openemr is deployed? Wouldn't it make more sense to separate the module into its own repo and install it via composer?

interface/modules/custom_modules/oe-module-faxsms/src/Controller/SignalWireClient.php
kojiromike: OK, there are already several. See https://github.com/openemr/openemr/issues/9875

interface/eRx.php
```diff
@@ -121,7 +121,7 @@ function array_key_exists_default($key, $search, $default = null)
 
         if (count($errors) > 0) {
             ?>
-        <strong><?php echo xlt('NewCrop call failed'); ?></strong>
+        <strong><?php echo xlt('Ensora call failed'); ?></strong>
```
kojiromike, line 124: Hmm, I think we need to be careful changing translated strings. If there's already a translation of 'NewCrop call failed' somewhere, changing 'NewCrop' to 'Ensora' will break the lookup. We should probably figure out a way to avoid proper nouns in our translations, for the most part.

I'd be curious about @bradymiller 's take here, since I think this probably goes mostly unnoticed.

juggernautsei: I understand your concern. I thought we had people working on translations. There were volunteers who went through the codebase and updated the translations.  I have never really thought about it too much. I just know that end the end. Things get translated or left. And since Ensora is a proper name it should not be translated like no one translates a persons name it remains the same in every language. Like Google.

interface/modules/custom_modules/oe-module-faxsms/composer.json
sjpadgett: Not our core modules. Besides I'd rather see a module installer that doesn't place onus on user having to get into command line use.
This module is a core communication module such that it is and is allowed to be more closely integrated with core.
I still have some refactors of the core of the module to get better consistency or services. So of which you pointed out in review.

sjpadgett:
@juggernautsei I have to agree with @kojiromike review and changes, many of which helps the overall module and not only your additional service. I can help if you want.
I have some core module refactors that I've been wanting to do anyway. Sorry but for some reason I didn't get notification of this PR. I'd like to see this completed and ready for our certification release next month. Let me know when you re focus on this.

Good review Mike!

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
juggernautsei: @kojiromike, the closest the system gets to standardizing the phone number is in the formatting functions in the library files.

interface/modules/custom_modules/oe-module-faxsms/WARP.md
juggernautsei: You raise a valid point about the architectural approach. You're correct that the long-term ideal would be:

1. Separate repositories - Each module as its own independent repo
2. Composer/Packagist distribution - Installable via composer require openemr/oe-module-faxsms
3. Decoupled WARP rules - Module-specific agent instructions only in that repo

Current State:
However, this module is currently shipped as part of OpenEMR core (interface/modules/custom_modules/), so it exists in this monorepo. The WARP.md serves as documentation for developers working on this integrated module within the larger codebase.

Suggested Approach:

1. Short term: Keep WARP.md as module-level documentation since it's part of core
2. Long term: When/if the module is extracted to a separate repository:
◦  Move the module to its own repo
◦  The WARP.md becomes the module's root documentation
◦  Individual agent instructions make complete sense at that point

Alternative: If the module stays in core, consider whether WARP.md should be moved to a location like /Documentation/MODULES/FAXSMS.md or consolidated into the main module documentation structure rather than at the module root.

This acknowledges the lead's architectural concerns while being pragmatic about the current state of the codebase.

juggernautsei:
@sjpadgett @kojiromike let me know if there is anything else you would like me to do. @sjpadgett I hope this helps with meeting the release schedule.

interface/modules/custom_modules/oe-module-faxsms/src/Enums/ServiceType.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+/**
+ * Service Type Enum
+ * Maps service type IDs to their display names and identifiers
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    SignalWire Integration
+ * @copyright Copyright (c) 2025
```
stephenwaite, line 10: hi @juggernautsei, shouldn't you get credit as the author and maybe the company has a claim to the copyright?

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
kojiromike: This comment is about moving this function into a shared utility class instead of defining it in the module.

interface/modules/custom_modules/oe-module-faxsms/composer.json
kojiromike: The problem with composer.json and package.json and the like in modules is that there's no static guarantee that the dependencies in the module align with the dependencies in openemr. Ironically, if the module were in its own repo/packagist, then composer would make those checks at install time, which would be safer. Furthermore, scanners like google's goog-vulnz will recursively search for dependency files and flag docker images for dependencies listed in them, even if they're not installed. So it's problematic to have dependency files laying around that aren't in use.

By the way, in my modules I've taken to including openemr/openemr in dev requires. It's not perfect, but it makes it harder to accidentally introduce an incompatible dependency, like symfony v7 etc.

interface/eRx.php
kojiromike: Hmm, this makes me wonder if we use BiDi markers properly to handle not-transliterated proper names when going from a LTR to a RTL language.

interface/modules/custom_modules/oe-module-faxsms/src/Enums/ServiceType.php
juggernautsei: Hi @stephenwaite, I added my name as author. Signal Wire should not have claims to copyrights fo the enums?

sjpadgett:
Hi Sherwin. As you know we have some breathing room on this now but I'd like to have some time to restructure and clean up the javascript and usability for the UI. Let me know if I can help move along. I'm looking forward to this and know how hard you worked to contribute. Also nice to hear is working on a production site.

Btw I knew you've been around openemr longer than me but reviewing your repo I see back to v3.2.0!  Wow we have to come up with some sort of badge for that!:)

sjpadgett:
@juggernautsei Since the rector and phpstan changes I rec you do a quick test of your service. When ready give me the word and we'll merge.

sjpadgett:
@juggernautsei Merging! So happy to have another option for our users so thanks much....

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    SignalWire Integration
+ * @copyright Copyright (c) 2025
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+// Allow webhook access without authentication
+$ignoreAuth = true;
```
bradymiller, line 15: Just making sure we are ok with this script running without any auth.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
```diff
@@ -0,0 +1,365 @@
...
+                return [
+                    'success' => true,
+                    'document_id' => $documentId,
+                    'media_path' => $mediaPath,
+                    'patient_id' => $patientId
+                ];
+            } else {
+                // Store in unassigned directory
+                $mediaPath = $this->receivedFaxesPath . '/unassigned/' . $filename;
+                file_put_contents($mediaPath, $mediaContent);
```
bradymiller, line 114: Should always use standard documents functions to store documents. note they can be stored with patient id 0 (ie. same as unassigned) and can also set a directory (ie. this can all be done safely an securely via standard documents functions). recommend checking out how emrdirect witht the standard document function which does this same thing.

bradymiller: real concern especially since there appears to be a mechanism to push in documents via an unauthenticated mechanism (in the first script in this PR)

juggernautsei: I understand the concern. 
Thoughts:
The fax needs to accept inbound data without anyone being logged into the system. I know it can be done via the command line, where a user is assigned, but that would require the user to know how to do that. A layperson will have to hire someone to configure the CLI to perform the inbound fax without someone being logged in. So, I hear you saying check out how direct mail solved for this, but the difference is that in faxes. There is no way to know who the faxes are to or about. They are unknown documents at the time of their arrival, and there is no patient, in my mind, to attach the document to. I have not looked, but I can see direct mail having a patient to attach the inbound document to. Please enlighten me about how the unassigned document can find a home in the program. Without auth has limitations as to what it can do. I have run up against all of them.  I know the biggest concern is what queries can be executed and what code can be written and retrieved without some type of security audit to make sure that nothing can be executed that is malicious in nature. I know I can't say I was just following what was there. That is no excuse. I can revisit the code from a security standpoint to see what I can find and correct.

sjpadgett: Faxes should not be stored to documents until disposed by user(audit). IMO 80% of faxes will not need to be charted as patient history, We already have a fax repository queue table that can be duplicated plus classes that manage. We may need to move some methods to base abstract class. This stuff is already done in module
Query for new, queue in table, show user and then give user option on how to dispose. Workflow is already done.

bradymiller: hi @sjpadgett and @juggernautsei ,

Sounds reasonable to not make them official documents then since already have a process. 

What is the standard process for an incoming fax in codebase/module? are they all going to documents/unassigned?

Also, are their more fax related incoming scripts that bypass authentication?

Just trying to get an idea of the flow here, so can provide better input to ensure we don't open up any security vulnerabilities.

sjpadgett: Hi @bradymiller Sorry, I didn't review structure. I'm not very good at reviews but you're correct we missed file flows in new service. Security and file handling for the other vendors has been previously worked out over the years.

EtherFAX → Transport Provider
Think of it like a fax “pipeline.”
Moves data through a secure network with no storage footprint.

RingCentral → Fax Mailbox Provider
More like Gmail for faxes.
Stores the fax content unless you explicitly delete it.

When I supported Twilio it acted in the same way as RC for fax retention.

So originally I let the vendor be my storage/backup and provided portal so user can purge when and how they want.
As you can see with etherFAX I had to handle document management locally. Once I download it, I own it. No do overs!

So I manage in a fax queue table whereas the fax is stored with it's meta data. I provide the user with several features to manage such as send in a pnote message, forward in pnote or email, chart to existing patient(use our document classes), view/download and delete.

RC I store meta and use RC Cloud server to hold documents.

Let me know if this is what you're after.

sjpadgett: Oh no on authentication I auth and needs to be addressed here. Pretty standard.

juggernautsei:
@bradymiller, because I was doing this at a breakneck speed to get a clinic operational. There was no looking around at the infrastructure, trying to see what existed. If it was not obvious, it was overlooked. Now that production is up and running, most of this is retrospective for me. This is where we make improvements on what was initially laid down. I never looked to see what @sjpadgett did in connecting to RingCentral or Twilio. Their path was not going to be SignalWire. 
I hope understanding this workflow helps. The inbound fax is a two-step process. The fax documents go into SignalWire first and land on their server. The webhook receives a notification that there is a new fax. The URL to pick up the file from SignalWire is contained within the notification message.  @sjpadgett could re-route the inbound PDF to store it somewhere else. 
I looked, and RingCentral external access is the same way. 
image

sjpadgett:
The premise is if SignalWire allows you to retain fax doc for history on their servers then document handling is brief and needs to go through an audit for disposal. If not, fetch and store in queue table. It is structured for different accounts.

Btw: when bypass auth in glbal the transaction will have a previous auth i.e. csrf

bradymiller:
Hi @sjpadgett and @juggernautsei ,

Thanks for the info. Super helpful.

rc_sms_notification.php only does $ignoreAuth = true for command line use
voice_webhook,php does $ignoreAuth = true, but has a token to control auth

however, for the script, webhook_receiver.php, in this PR, I do not see any auth happening. It appears anybody can hit that endpoint and dump documents into OpenEMR. Is that correct?

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -0,0 +1,272 @@
...
+    exit('Invalid request');
+}
+
+// Get webhook payload - initialize all variables explicitly for PHPStan
+$faxSid = $_POST['FaxSid'] ?? $_POST['Sid'] ?? '';
+$status = $_POST['Status'] ?? $_POST['FaxStatus'] ?? 'unknown';
+$from = $_POST['From'] ?? $_POST['RemoteStationId'] ?? '';
+$to = $_POST['To'] ?? $_POST['OriginalTo'] ?? '';
+$numPages = $_POST['NumPages'] ?? $_POST['Pages'] ?? 0;
+$mediaUrl = $_POST['MediaUrl'] ?? '';
```
bradymiller, line 90: hi @juggernautsei and @sjpadgett , I think this variable may be the main vulnerability.
My take on this script is that somebody (should be SignalWire) uses this to let you know there is a fax to pick up from them. You then connect to SignalWire (with your SignalWaire credentials) to get it via the above mediaUrl path. Is there a way to ensure the mediaUrl path only goes to the SignalWire server? (ie. a bad actor coul put in their own evil url for this post variable and then you would send the evil url your SignalWire credentials to try to pick it up).

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -0,0 +1,272 @@
...
+                    'Authorization' => 'Bearer ' . $apiToken
+                ]);
+            } else {
+                // Use Basic authentication for API endpoints
+                $httpRequest->setOptions([
+                    'auth' => [$projectId, $apiToken]
+                ]);
+            }
+
+            $response = $httpRequest->get($mediaUrl);
```
bradymiller, line 226: here's where the request could go to the evil url

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
```diff
@@ -0,0 +1,365 @@
...
+                
+                return [
+                    'success' => true,
+                    'document_id' => $documentId,
+                    'media_path' => $mediaPath,
+                    'patient_id' => $patientId
+                ];
+            } else {
+                // Store in unassigned directory
+                $mediaPath = $this->receivedFaxesPath . '/unassigned/' . $filename;
```
bradymiller, line 113: Hi @sjpadgett and @juggernautsei , Highly recommend running the file through encrypt before storing it in the documents folder (assuming you can decrypt it when go through process of placing it in a patient chart (where it will again be decrypted via the standard document import)). Since a simple apache gaffe can make the documents folder public (with no need for auth), goal is to encrypt all files with PHI and/or uploaded files to prevent loss of PHI and uploading of a web shell.

juggernautsei: The encryption and decryption have been built, but I think I have to open a new repo to push those changes to.

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
bradymiller: hi @juggernautsei and @sjpadgett , Just checking if this issue has been addressed

Issue #9877, opened 2025-12-15
feat: Add automated tests for ccda node service


adunsulag:
### Describe the problem

We don't have automated tests for the ccda document generation that verifies the node service is working properly.  We need to add basic tests that will verify the node serveccda.js file is operating properly when sent an OpenEMR xml data generation request.

We also need to add in tests for certification to make sure the CCDA is generated properly when called from the G9 Data API Access for onc certification.  

This way we can do updates to the underlying node packages and dependencies without worrying that things will break.

If we can add initial tests to cover the common case scenario then we can build on that foundation.

PR #9878, opened 2025-12-15
feat: openemr 9877 ccda tests

adunsulag:
Implemented a phpunit test for verifying certification data didn't
change.  I went with Alice who is the most comprehensive test patient
for certification.  I could do Happy Child and Jeremy but I think this
is sufficient without us going and implementing the entire ETT test
suite for onc certification. If someone doesn't have the private files
test suite and attempts to run the certification suite it just skips the
test.

I also implemented a much faster test that verifies the node serveccda.js
without requiring that we work on the certification test data set.  It
uses the OpenEMR xml generated CCDA file that is sent to the serveccda
which formats it to the ccda specification format.  This verifies that
the node server gets started and that it properly generates the correct
xml file.

Fixes https://github.com/openemr/openemr/issues/9877

Also, the serveccda.js was seg faulting due to invalid node version 20 when
version 22 is now required for package dependencies.  I updated the
version.  I left it at 8.3 which is what inferno has been running on.
We could bump it up to 8.4 if we wanted to in the future.


adunsulag:
@kojiromike What needs to be added here for the docker lint stuff?  Looks like this is something new here that was added but the error log would suggest something missing in the docker image.  I thought I picked the same image as one of the other CI tests so not sure what is going on here.

kojiromike:
> @kojiromike What needs to be added here for the docker lint stuff?

@adunsulag that's my bad. There's something fishy about npx where it's hiding the parameters from the called command. It's not been entirely reproducible either, which is super. I can't reproduce it at all locally, and it doesn't always happen in actions. Anyway, I'll probably remove npx from the equation and see if that addresses it.

kojiromike:
@adunsulag if you rebase the dclinting should be ok now.

On a different note, once this is done, how do you feel about testing and then merging the dependabot PRs on ccdaservice?

adunsulag:
@kojiromike will try to get to it, I'm skeptical we'll get to the ccda stuff as I'm trying to wrap another thing and will circle back to this as soon as I finish.

kojiromike:
@adunsulag I am not saying we have to do this before 7.0.4. I'm just asking if the testing will make you feel confident about dependabot.

adunsulag:
Discovered that the node version installed on openemr/dev-php-fpm:8.* is node 20 when the minimum supported is nodejs 22.  

See here https://github.com/openemr/openemr/blob/master/docker/library/dockers/dev-php-fpm-8-2/Dockerfile

We'll need to bump up the nodejs version on the dockers to fix these issues.  Will open a separate issue for this.

adunsulag:
Forgot that the weekly php-fpm workflow has to be triggered to rebuild the images.  Then I can re-run these workflows.

PR #9880, opened 2025-12-16
Remove compatibility mode from OEGlobalsBag

### LINKED ISSUE #9876: OEGlobalsBag won't reliably respect `$compatibilityMode` when using singleton
### Describe the bug

`OEGlobalsBag::getInstance()` returns a singleton. The implementation is such that subsequent calls with a different value of the `$compatibilityMode` parameter will have that value effectively ignored.

I'm not sure what impact this has, if any. It may be a lurking area of danger without any current practical implications. Certainly the common-case is that the "default" call in `interface/globals.php` passes `true`, so unless an endpoint initializes it earlier in a different way, it likely won't matter. I get the impression that most of the application would fail catastrophically if this is ever used _without_ `$compatibilityMode = true`.

This may not be worth addressing, especially in the context of #9867 and some possible longer-term decisions from there. But I spotted it while looking at something else and wanted to make sure this behavior is documented even if it's left alone.

Firehed:
Fixes #9876 

Removes compatibility mode option from OEGlobalsBag, makes it always-on.

Removes globals compatibility mode flag from OEGlobalsBag::getInstance() - it created a path where the parameter could get ignored and result in unexpected behavior. See linked issue for more context.

While the intent behind the flag is solid, it doesn't _really_ help in the migration away from `$GLOBALS` and in the meantime could lead to strange bugs.

There are a couple call-site adjustments to account for the change.

No


src/Core/OEGlobalsBag.php
```diff
@@ -51,16 +50,15 @@ public function set(string $key, mixed $value): void
...
-            $GLOBALS[$key] = $value;
-        }
+        // Push the value into GLOBALS for backwards compatibility. Eventually
+        // this should be removed.
+        $GLOBALS[$key] = $value;
     }
 
     public function get(string $key, mixed $default = null): mixed
     {
         // If in compatibility mode and key not in bag, try $GLOBALS
```
kojiromike: Will the phrase "compatibility mode" make sense to a future reader of this comment?

src/Core/OEHttpKernel.php
```diff
@@ -18,7 +18,7 @@ class OEHttpKernel extends HttpKernel
     public function __construct(EventDispatcherInterface $dispatcher, ControllerResolverInterface $resolver, ?RequestStack $requestStack = null, ?ArgumentResolverInterface $argumentResolver = null, bool $handleAllThrowables = false)
     {
         parent::__construct($dispatcher, $resolver, $requestStack, $argumentResolver, $handleAllThrowables);
-        $this->globalsBag = new OEGlobalsBag([], true); // set compatibility mode to true until we can get rid of it
+        $this->globalsBag = new OEGlobalsBag([]);
```
kojiromike: Should this not be the singleton instance?

src/Core/OEGlobalsBag.php
Firehed: Oops, meant to pull that out too. Will fix.

src/Core/OEHttpKernel.php
Firehed: That was my initial reaction when seeing it, but I don't have enough context of the surrounding code yet to know if that's a safe assumption. I don't think much code goes through this path yet, but I'd be more than happy to test such a change if someone can point me in the right direction.

kojiromike: @adunsulag what do you think? Does OEHttpKernel need a dedicated globalsbag instance that is not the one we get from globals.php?

Firehed:
I'm kicking this back to draft for now since we've got a couple open directional questions and some internal tweaks required to get everything green.

kojiromike:
@adunsulag can you take another look at the (hidden) comments in this PR? I think this PR is pretty important, and we should try to get it into 8.0, but there are some open design questions.

src/Core/OEHttpKernel.php
adunsulag: No, its fine to be the singleton.  In terms of the request lifecycle the globals don't get setup until SiteSetupListener executes.  They should all share the same object.

Firehed: While following up on this, I discovered this will be a _little_ funky in practice:

1) `OEHttpKernel` gets the singleton
2) `SiteSetupListener` runs, causing `interface/globals.php` to get `require`d.
3) globals.php does all of its various things, which ends with a) $GLOBALS getting populated, b) the existing OEGlobalsBag singleton getting wiped, and c) the OEGlobalsBag singleton getting reinitialized with the new state of $GLOBALS

Consequently, the bag instance in the Kernel is _no longer_ the same one managed by the singleton. Perhaps it's now a dualton? In any case, since both are in BC mode, it'll just proxy `$GLOBALS` and the end result is the same - both should reliable get and set the same data, even if they won't pass reference equality.

I don't want to spend too much time trying to fix _that_ problem (`resetInstance` should get reworked to just reload the global data) since we want to get away from the globals and singleton stuff anyway.

In short: this should have no runtime impact, but it could lead to some really cryptic debugging situations. I'll make a follow-up issue specifically for that

kojiromike:
https://github.com/openemr/openemr/pull/10128

PR #9881, opened 2025-12-16
refactor: Update internals of EventAuditLogger to use DB tooling

### LINKED ISSUE #9773: Migrate EventAuditLogger from raw ADODB to sql wrapper
The internals of this class are relying on a direct database connection rather than using the wrappers used throughout the rest of the application. The actual touch points are manually escaping parameters, rather than using prepared statements (which are supported by the function it used to actually execute the SQL)

For both safety and encapsulation purposes, these should be switched to using the existing DB tooling and prepared statements.

Firehed:
Fixes #9773

Avoids `$GLOBALS['adodb']['db']` access in favor of the existing database wrappers

The database tooling that the EventAuditLogger calls supports prepared statements. Instead of manually escaping values which is _generally_ error-prone (although in this instance looks fine), a few touch points have been migrated to using the existing prepared statements.

No

### DIFF
--- a/.phpstan/phpstan-database-baseline.neon
+++ b/.phpstan/phpstan-database-baseline.neon
@@ -5664,11 +5664,11 @@ parameters:
       identifier: openemr.deprecatedSqlFunction
       count: 1
       path: ../src/Common/ORDataObject/Contact.php
-    - message: '#^Use QueryUtils\:\:sqlStatementThrowException\(\) or QueryUtils\:\:fetchRecords\(\) instead of sqlStatement\(\)\.$#'
+    - message: '#^Use QueryUtils\:\:querySingleRow\(\) or QueryUtils\:\:fetchRecords\(\) instead of sqlQuery\(\)\.$#'
       identifier: openemr.deprecatedSqlFunction
       count: 1
       path: ../src/Common/ORDataObject/ORDataObject.php
-    - message: '#^Use QueryUtils\:\:querySingleRow\(\) or QueryUtils\:\:fetchRecords\(\) instead of sqlQuery\(\)\.$#'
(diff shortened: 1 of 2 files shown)


src/Common/Logging/EventAuditLogger.php
```diff
@@ -692,13 +692,17 @@ public function auditSQLAuditTamper($setting, $enable)
...
     public function recordDisclosure($dates, $event, $pid, $recipient, $description, $user)
     {
-        $adodb = $GLOBALS['adodb']['db'];
-        $sql = "insert into extended_log ( date, event, user, recipient, patient_id, description) " .
-            "values (" . $adodb->qstr($dates) . "," . $adodb->qstr($event) . "," . $adodb->qstr($user) .
-            "," . $adodb->qstr($recipient) . "," .
-            $adodb->qstr($pid) . "," .
-            $adodb->qstr($description) . ")";
-        sqlInsertClean_audit($sql);
+        $sql = "INSERT INTO extended_log ( date, event, user, recipient, patient_id, description) " .
```
kojiromike: As long as we're writing multi-line raw sql in here, what do you think about using heredocs?

Firehed: I tend to do so in my own projects, but saw it's not too widely used in this codebase so I stuck with the existing formatting. Ideally I'd like to see this replaced with proper DBAL tooling and sidestep formatting entirely, but that's out of scope for now.

If you'd prefer this as heredoc I'm happy to change it!

kojiromike: Please do. Also, is anything stopping us using QueryUtils here?

Firehed: Updated.

`QueryUtils` doesn't expose a wrapper for `sqlInsertClean_audit` and that's only used by this file so I'd rather not mess with exposing it in favor of putting effort towards fixing more of the DB-related fundamentals, but I did update the `getLastInsertId()` since that exists already.

PR #9884, opened 2025-12-16
Fix the swagger file Fixes #9702

### LINKED ISSUE #9702: bug: Swagger tool v2 scopes missing, invalid $docref operation, trusted user save bug

adunsulag:
Fixes #9702 swagger file which wasn't updated properly on the merge.

### DIFF
--- a/swagger/openemr-api.yaml
+++ b/swagger/openemr-api.yaml
@@ -5161,6 +5161,53 @@ paths:
       security:
         -
           openemr_auth: {  }
+  /fhir/DocumentReference/$docref:
+    post:
+      tags:
+        - fhir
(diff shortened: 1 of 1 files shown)


stephenwaite:
super, will bring in, thanks @adunsulag !

PR #9943, opened 2025-12-19
feat(routing): Initial front-controller implementation

### LINKED ISSUE #8964: Introduce global front controller for openemr

Firehed:
Fixes #8964

Creates a front controller file and a sample `.htaccess` for opt-in adoption. It should handle all of the `.htaccess` rules currently in the codebase.

This is the first of probably many PRs that will get all OpenEMR web traffic to flow through a front-controller. That goal has been discussed at length so I won't rehash it or its goals here.

What this DOES do:
- Create a front controller file
- Perform basic application bootstrapping
- Mirror the various `.htaccess` rules into PHP (deny and rewrite)
- Performs PHP-side routing to existing paths
- Sets up an example root-level rewrite rule to send a lot of web traffic to `public/index.php`[^apache]
- Tweaks PHPUnit configs for better support of DI tooling that the front controller uses 

What this DOES NOT do:
- Send any traffic to the front controller (!)
- Break any existing code
- Complete comprehensive request filtering
- Make any changes regarding use of globals, database connections, etc
- Set up any path for a "new" controller/request handler system[^req]. This is a logical next-step to come later, to be inserted in the same front controller file.
- Change low-level Apache rules, or other webserver configs (officially-supported or otherwise)

_Following_ this, we'd likely want to:
- Create "official" webserver configs that push traffic through the FC
- Update the dev docker environments, etc, with a way to run traffic through the FC
- Insert a well-defined place that new code not relying on globals, etc, can make new routes accessible (see comments in the FC file itself)[^di]
- Improve filtering if needed
- In some sort of BC break release, _require_ that the webserver routes php traffic

#### Testing Guidance

This (mostly) works with the built-in PHP webserver! The easiest way to smoke-test it without other changes is:

1) Run the docker development-easy and let it sit there (should have a generally working OE install). This is just to have a DB set up.
2) On the HOST machine, set `sites/default/sqlconf.php` to point at its database:
```php
$host   = '127.0.0.1';
$port   = '8320';
$login  = 'openemr';
$pass   = 'openemr';
$dbase  = 'openemr';
```
3) On the HOST machine, make sure dependencies are all set and assets are built
```
composer install
npm ci
npm run build
```
4) On the HOST machine, run the PHP webserver:
```
php -S 0.0.0.0:8305 -t . public/index.php
```
5) Browse to `http://localhost:8305`

#### Does your code include anything generated by an AI Engine?
A couple tests and lots of discussion

[^req]: Using any sort of router, PSR-7 or Symfony request tooling, etc.
[^apache]: So far, I've only tested this through the `docker/development-easy` path which uses Apache. More to come later.
[^di]: The DI mechanics to start on this are largely in place. The only soft blocker is ORM (basically ready, #10944) + DBAL (landed but dormant), and _using_ the latter is itself soft-blocked by the cryptogen refactor (#11267, etc) which enables the audit logging middleware (#11129). Yeah, OpenEMR is kind of complex.


Firehed:
This is still I think directionally correct, but largely blocked by some packaging issues.

One thing that I think _may_ help addressing this is some sort of shutdown hook check. Something to the general effect of:
```php
// public/index.php
const USED_FRONT_CONTROLLER = true;

// interface/globals.php, probably
register_shutdown_function(function() {
  if (!defined('USED_FRONT_CONTROLLER')) {
    getLogger()->error('Request did not go through front controller file, this means your web server config is incorrect. See <some page> for how to correct this');
  }
});
```

The "official" infrastructure (openemr-devops, I guess) should have the webserver configured correctly, along with first-party docker images.

src/Common/Http/HttpRestRequest.php
```diff
@@ -747,11 +747,6 @@ public function withMethod($method): self
         return $clonedRequest;
     }
 
-    public function getUri(): string
```
Firehed, line 753: While working on the test hooks, I discovered that this class had an issue where the default initialization wasn't enough and getUri() would return null and crash due to the typehint. The parent method does it just fine, so I'm removing the override.

tests/Tests/Isolated/BC/FrontControllerRoutingTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * E2E tests for front controller routing using PHP's built-in server.
+ *
+ * These tests validate that requests are routed to the correct entry point.
+ * They do NOT validate full application behavior (which requires a database).
+ */
+#[CoversClass(FallbackRouter::class)]
+#[Large]
+class FrontControllerRoutingTest extends TestCase
```
Firehed, line 32: Strictly speaking this shouldn't be in the "isolated" tests since it's doing some funky process work, but it doesn't require the DB to work and is still fast

apis/dispatch.php
```diff
@@ -17,13 +17,15 @@
...
 
+use OpenEMR\BC\FallbackRouter;
 use OpenEMR\Common\Http\HttpRestRequest;
 use OpenEMR\RestControllers\ApiApplication;
 use Symfony\Component\HttpFoundation\Response;
 
 // create the Request object
 try {
     $request = HttpRestRequest::createFromGlobals();
+    FallbackRouter::handleRoutingTestIfRequested($request->getRequestUri(), 'apis');
```
Firehed, line 28: As the name implies, this is a brief hook for integration testing.

config/app.php
```diff
@@ -0,0 +1,19 @@
...
+ * @copyright Copyright (c) 2026 Eric Stern
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+return [
+    // Important: this must be a closure rather than a static definition so
+    // that a compiled container is guaranteed to align with the runtime.
+    'installRoot' => fn () => dirname(__DIR__),
```
Firehed, line 18: Aside: this actually won't be required one all code runs through the front-controller since there's a clear reference for the install root, but it's needed transitionally.

oauth2/authorize.php
```diff
@@ -10,15 +10,18 @@
...
 require_once "../vendor/autoload.php";
 
 // TODO: @adunsulag at some point we can have the .htaccess file just hit
 // everything in the dispatch.php file and then we can remove this file
-use OpenEMR\Common\Http\HttpRestRequest;
-use OpenEMR\RestControllers\ApiApplication;
 // create the Request object
 try {
     $request = HttpRestRequest::createFromGlobals();
+    FallbackRouter::handleRoutingTestIfRequested($request->getRequestUri(), 'oauth2');
```
Firehed, line 24: (same as api/dispatch.php)

src/BC/FallbackRouter.php
```diff
@@ -0,0 +1,222 @@
...
+ * .htaccess files:
+ *
+ * Deny rules:
+ * - [x] ./bin/.htaccess
+ * - [x] ./portal/patient/fwk/libs/.htaccess
+ * - [x] ./sites/default/documents/.htaccess
+ *
+ * Rewrites:
+ * - [x] ./apis/.htaccess
+ * - [x] ./interface/modules/custom_modules/oe-module-faxsms/.htaccess (no path rewrite needed; only adds query param)
```
Firehed: If there are other modules with custom .htaccess rules... there's not a great way to handle that here. They'd already be either broken or unpredictable in nginx as well.

This will become irrelevant soon enough,

tests/Tests/Isolated/BC/FrontControllerRoutingTest.php
```diff
@@ -0,0 +1,124 @@
...
+
+        self::$process = new Process([
+            'php',
+            '-S', sprintf('%s:%d', $host, $port),
+            $router,
+        ]);
+        self::$process->start();
+
+        // Give the server time to start
+        usleep(500_000);
```
Firehed, line 53: I don't like this one bit, but haven't found it to be flaky in practice.

PR #9947, opened 2025-12-19
refactor(php): prefer interfaces to implementations

### LINKED ISSUE #9946: refactor: use interfaces for type declarations where possible
Currently, a lot of cases where types are declared in openemr specify the implementation of a thing that has an interface. For example, we might say

```php
public function __construct(SystemLogger $logger) ...
```

But [SystemLogger implements LoggerInterface](https://github.com/openemr/openemr/blob/487b13f910bd4859d8586b9e45a42947a631c95a/src/Common/Logging/SystemLogger.php#L19C31-L19C46), so `LoggerInterface` would be a better choice there, enabling easier testing and a low-cost step towards dependency injection.

This issue is to look for the low-hanging fruit places we can make this kind of improvement.

kojiromike:
Fixes #9946

Replace type declarations that declare the implementation with ones that declare the interface, where feasible.

- refactor: use EventDispatcherInterface instead of concrete EventDispatcher
- refactor: use EventDispatcherInterface in remaining type declarations
- refactor: use LoggerInterface instead of concrete SystemLogger
- refactor: use PSR-11 ContainerInterface and SessionInterface

### DIFF
--- a/interface/patient_file/encounter/forms.php
+++ b/interface/patient_file/encounter/forms.php
@@ -40,7 +40,6 @@ use OpenEMR\Common\Forms\FormReportRenderer;
 
 use OpenEMR\Services\EncounterService;
 use OpenEMR\Services\UserService;
-use Symfony\Component\EventDispatcher\EventDispatcher;
 
 $expand_default = (int)$GLOBALS['expand_form'] ? 'show' : 'hide';
 $reviewMode = false;
@@ -59,14 +58,7 @@ if ($is_group && !AclMain::aclCheckCore("groups", "glog", false, ['view', 'write
     exit();
 }
 
-if ($GLOBALS['kernel']->getEventDispatcher() instanceof EventDispatcher) {
-/**
- * @var EventDispatcher
- */
-    $eventDispatcher = $GLOBALS['kernel']->getEventDispatcher();
-} else {
-    throw new Exception("Could not get EventDispatcher from kernel", 1);
-}
+$eventDispatcher = $GLOBALS['kernel']->getEventDispatcher();
 // instantiate the locator at the beginning so our file caching can be re-used.
 $formLocator = new FormLocator();
 ?>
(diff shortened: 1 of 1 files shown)


src/Core/OEHttpKernel.php
```diff
@@ -30,14 +31,14 @@ public function getEventDispatcher(): ?\Symfony\Component\EventDispatcher\EventD
     {
         return $this->dispatcher;
     }
-    public function getSystemLogger(): ?SystemLogger
+    public function getSystemLogger(): ?LoggerInterface
```
kojiromike:
```suggestion
    public function getSystemLogger(): LoggerInterface
```

This can't actually ever be null.

kojiromike:
@Firehed @rollingventures I can't directly request your review because of another regrettable GitHub design decision, but I can still spam you. Please review this changeset.

kojiromike:
To kick off any discussion of LSP:

- Return types changed from concrete → interface (widening), which is safe. Existing code still receives the same concrete objects.
- Parameter types changed from concrete → interface (widening), which is valid contravariance.
- Interface implementations (BaseServiceInterface) were updated consistently with their implementers.

The only gotcha I see is that `EventDispatcherInterface` type inconsistency in `OEHttpKernel` (Contracts vs Component), but that's a pre-existing issue.

rollingventures:
Tested locally with existing and dev modules, all passing. One note for future cleanup: interface/patient_file/encounter/forms.php and interface/main/finder/dynamic_finder.php still have `instanceof` EventDispatcher checks and inline `addListener()` calls that rely on the concrete class.  A follow-up PR could either swap those to `Symfony\Component\EventDispatcher\EventDispatcherInterface` (which declares those methods) or refactor to the EventSubscriberInterface pattern at bootstrap. Not blocking — just related tech debt worth tracking.

kojiromike:
@rollingventures oh yeah, good find. Also I think the defensive conditions that those instanceof checks were used for are irrelevant – `getEventDispatcher` will throw an error if it can't return the thing, so we don't need to double-check the return type. We can get rid of them entirely.

PR #9950, opened 2025-12-20
Refactor Portal CCDA actions

### LINKED ISSUE #9949: Refactor API passed session requirement to access Portal CCDA actions.
### Describe the problem

Currently for CCDA actions the CCM is accessed using an API call passing the current auth session to CCM to authorize the request.

Need to refactor to use CCM services directly. This kind of limits future move toward portal as a stand alone app(if desired).
Hopefully I'm able to separate concerns in CCM for this process enough to duplicate the process without the CCM dependency.

Let's see how it goes.

sjpadgett:
Remove API request for CCDA view to use CCM services.

Fixes #9949

Doing this one action at a time. This is rough so far. I want to do some consolidation in CCM too.
comments welcome.

### DIFF
--- a/.composer-require-checker.json
+++ b/.composer-require-checker.json
@@ -5,6 +5,8 @@
     "BillingExport",
     "C_Document",
     "Carecoordination\\Model\\CarecoordinationTable",
+    "Carecoordination\\Model\\CcdaGenerator",
+    "Carecoordination\\Model\\EncounterccdadispatchTable",
     "FPDF_FONTPATH",
     "GenericProductRegistrationException",
     "IS_WINDOWS",


src/Services/CDADocumentService.php
```diff
@@ -173,21 +179,54 @@ public function generateCCDXml($pid): string
...
+    private function XmlToHtmlContent($content): false|string
+    {
+        $xml = simplexml_load_string($content);
+        $xsl = new DOMDocument();
+        // cda.xsl is self-contained with bootstrap and jquery.
+        $sheet = __DIR__ . '/../../interface/modules/zend_modules/public/xsl/cda.xsl';
+        $xsl->load($sheet);
+        $proc = new XSLTProcessor();
+        $proc->importStyleSheet($xsl); // attach the xsl rules
+        $outputFile = sys_get_temp_dir() . '/out_' . time() . '.html';
```
Sjord: Would it be better/more secure to use [`tempnam`](https://www.php.net/manual/en/function.tempnam.php) here?

sjpadgett: kind of a nit pick isn't it?:)

kojiromike: @sjpadgett you don't think this might be an insecure creation of a temporary file, like https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File? tempnam would help avoid that.

sjpadgett: Nope

sjpadgett:
@adunsulag I hook same place your interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Listener/CCDAEventsSubscriber.php do.
May want to check me...

No clue how this is supposed to work here! [Composer Require Checker / composer-require-checker (8.2, 4.16.1) (pull_request)

sjpadgett:
Laminas Zip is deprecated and has been for quit awhile so while I'm here will refactor out the only two place I can find.
We've been caught before so away I go!:)

sjpadgett:
I've tested this pretty well and appears to be solid. Didn't test subscriber events though but I used same hook events do so I don't foresee issue.
Going to merge so session project can continue. I'll make an effort to stay out of portal but I have a feature I want to add to medication list for next release so perhaps we can get the session PR done and merged soon.

ccdaservice/ccda_gateway.php
```diff
@@ -7,101 +7,120 @@
...
 // Will start the (patient) portal OpenEMR session/cookie.
 // Need access to classes, so run autoloader now instead of in globals.php.
 $GLOBALS['already_autoloaded'] = true;
-require_once(__DIR__ . "/../vendor/autoload.php");
+require_once __DIR__ . "/../vendor/autoload.php";
 SessionUtil::portalSessionStart();
 
 $sessionAllowWrite = true;
-if (isset($_SESSION['pid']) && isset($_SESSION['patient_portal_onsite_two'])) {
+if (isset($_SESSION['pid'], $_SESSION['patient_portal_onsite_two'])) {
```
bradymiller, line 28: neat change on the isset. didn't know isset could do that.

ccdaservice/ccda_gateway.php
```diff
@@ -7,101 +7,120 @@
...
     define('IS_PORTAL', $_SESSION['pid']);
 } else {
     SessionUtil::portalSessionCookieDestroy();
     $ignoreAuth = false;
-    require_once(__DIR__ . "/../interface/globals.php");
-    if (!isset($_SESSION['authUserID'])) {
-        $landingpage = "index.php";
-        header('Location: ' . $landingpage);
+    require_once __DIR__ . "/../interface/globals.php";
+    if (empty($_SESSION['authUserID'])) {
```
bradymiller, line 38: Rec keeping the !isset since empty is not the same and this code block is used in may places (ie. goal to keep it uniform)

(unless there is a reason for the change needing to be here of course)

src/Services/CDADocumentService.php
```diff
@@ -6,250 +6,319 @@
...
-        if ($networkUtils->isLoopbackAddress($this->serverUrl)) {
-            // Loopback address - traffic never leaves the local machine
-            // SSL verification is always disabled for loopback (no security benefit, often fails)
-            $this->verifySsl = false;
-            $this->caCert = false;
-        } else {
-            // Non-loopback address (e.g., nginx sidecar, docker compose, kubernetes)
-            $this->verifySsl = (bool) ($GLOBALS['http_verify_ssl'] ?? true);
-            $this->caCert = $GLOBALS['http_ca_cert'] ?? false; // Use custom CA cert for self-signed certificates
-        }
```
bradymiller, line 56: nice!

bradymiller:
2DV-2

PR #9957, opened 2025-12-22
perf: optimize xl() translation function for speed

### LINKED ISSUE #9956: xl() function is unnecessarily slow
It's crucial that `xl` (the translation function) be fast, because there are 10,362 calls to it in the codebase. Every time a string can be translated, it calls this function. Yet, the function has several problems:

1. It makes a sql query with a join, and that join is missing an index on `lang_id`, the heart of the `WHERE` clause.
2. There is no caching of the result. Two translations of the same string = 2 database calls.
3. The database returns 0 rows for English language lookups. English is the default language, so xl should not call out to the database at all.

Consider:

1. Add a configuration setting so that deployments with sufficient memory can load translations into cache early. (Or, at least, come up with some kind of caching to avoid unnecessary database calls.)
2. Add a short circuit for English, so that xl() just becomes an identity function when the language is English.
3. Add a composite index on (lang_id, cons_id).

kojiromike:
Fixes #9956

The `xl()` translation function is called over 10,000 times per page load but was making individual database queries for each call, with a JOIN missing an index on the key filter column.

- Add composite index `(lang_id, cons_id)` on `lang_definitions` table to optimize the JOIN query
- Add in-memory caching to avoid repeated DB lookups for the same strings within a request
- Add optional `translate_skip_english_lookup` setting to bypass DB entirely for English (for sites with no custom English terms)
- Add optional `translate_preload_cache` setting to warm the cache at startup with a single query loading all translations
- Optimize SQL to select only the needed column instead of `SELECT *`
- Log warning via SystemLogger if both cache preload and skip English are enabled (mutually exclusive settings)

Yes

The code was generated with Claude Code (Claude Opus 4.5). The commit message includes the standard Claude Code attribution footer.


sql/patch.sql
```diff
@@ -47,3 +47,7 @@
 
 --  #EndIf
```
bradymiller, line 48: hi @kojiromike , master doesn't use patch.sql . this should go into 7_0_4-to-7_0_5_upgrade.sql

bradymiller:
neat stuff!

sql/patch.sql
kojiromike: @bradymiller thanks, will do

kojiromike:
@Firehed

library/translation.inc.php
```diff
@@ -1,9 +1,124 @@
...
+ * When 'translate_skip_english_lookup' is set and the language is English
+ * When cache is empty (no translations)
+ * When first access to database finds there are no translations for that language.
+ */
+function xlskip(int $lang_id): bool
+{
+    $cache = &xlGetCache();
+    $globals = OEGlobalsBag::getInstance(true);
+    return ($globals->getBoolean('temp_skip_translations')
+        || ($lang_id === 1 && $globals->getBoolean('translate_skip_english_lookup'))
```
Firehed: Is there a constant we can reference for english instead of `=== 1` here? Even if it alone gets a special-cased addition I think it'd add clarity here.

library/translation.inc.php
```diff
@@ -1,9 +1,124 @@
...
+ * When 'temp_skip_translations' is set
+ * When 'translate_skip_english_lookup' is set and the language is English
+ * When cache is empty (no translations)
+ * When first access to database finds there are no translations for that language.
+ */
+function xlskip(int $lang_id): bool
+{
+    $cache = &xlGetCache();
+    $globals = OEGlobalsBag::getInstance(true);
+    return ($globals->getBoolean('temp_skip_translations')
```
Firehed: I'd request breaking this down into smaller blocks for enhanced readability, e.g. 

```php
if ($globals->getBoolean('temp_skip_translations')) {
  return true;
}
if ($lang_id === 1 && $globals->getBoolean('translate_skip_english_lookup')) {
  return true;
}
//...
```

library/translation.inc.php
```diff
@@ -1,9 +1,124 @@
...
 
+use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Common\Logging\SystemLogger;
+use OpenEMR\Core\OEGlobalsBag;
+
+if (!(function_exists('xlGetCache'))) {
+    /**
+     * Returns a reference to the shared translation cache
+     */
+    function &xlGetCache(): array
```
Firehed: If we can get away with it structurally, moving this stuff into a class (even if it's gotta be a singleton that gets wrapped for convenience, a la QueryUtils/sql.inc.php) would be a lot nicer, and should make it easy (easier?) to avoid this static-by-reference business.

library/translation.inc.php
```diff
@@ -1,9 +1,124 @@
...
+        $rows = QueryUtils::fetchRecordsNoLog($sql, [$lang_id]);
+        $cache[$lang_id] = array_column($rows, 'definition', 'constant_name');
+    }
+}
+
+/**
+ * Checks if there are any translations for the given language.
+ * Caches the result: empty array means "checked, no translations exist".
+ */
+function xlLangHasTranslations(int $lang_id): bool
```
Firehed: Presumably if the above `function_exists` checks above are required (I'd guess due to this file erroneously getting included multiple times) this probably needs it too. Unsure with the limited context I have. Also largely irrelevant if this moves into a class per above comment.

sql/patch.sql
```diff
@@ -47,3 +47,7 @@
...
 --    all blocks are terminated with and #EndIf statement.
+
+#IfNotIndex lang_definitions lang_cons
+CREATE INDEX `lang_cons` ON `lang_definitions` (`lang_id`, `cons_id`);
```
Firehed: With a slightly different set of internals we might be able to avoid a second index in favor of simply expanding the existing one (`KEY cons_id (cons_id, lang_id)`, maybe renamed), which would save a little bit of resource usage on the DB side. But if that makes the code too fiddly, it might not be worth it.

Firehed (commented):
I've gotta go a deeper on the internals still, but right now I'd say it's a huge improvement (when turned on) that could go a little further.

The big thing for me is putting this logic into the global function namespace rather than a scoped class. Encapsulating this further ought to be a better foundation for future improvements.

I'd also be somewhat inclined, for prewarming, to do a two-phase load avoiding a join:
- `SELECT * FROM lang_constants`, turn into a dict of `constant_name => cons_id`
- `SELECT * FROM lang_definitions WHERE lang_id = ?` and then munge this with above into the cached value for similar lookups.  Far from critical now, but I think it'll make future work easier, both generally (db tooling) and in the context of i18n. This is certainly not a blocker for me.

Pseudocode, I'd probably do something like:
```php
class TranslationCache {
  private ?array $cache = null;
  public function __construct(
    private int $langId
  ) {}
  public function getTranslation(int $definitionId): string
  {
    // hydrate the cache if needed, return the value
  }
}

class Translator {
  public function __construct(
    private int $langId,
    private ?TranslationCache $cache,
    private bool $skipEnglish,
  ) {}
  function translate(string $source): string
  {
    // if langId == english && skipEnglish, return source
    // if caching enabled, delegate to it, otherwise fall back to current path
  }

  public static function instance(): Translator
  {
    // read globals/session/etc and populate singleton, pass cache or null depending on config
    // xl() function family updated to use this path
  }
}
```

Not fully fleshed out here but hopefully you get the general idea. Feel free to take it or leave it, or do something with it in a future PR.

(Thinking further, we could define a `TranslatorInterface` with both the db- and cache-backed implementations use, but that may be excessive for now)

library/translation.inc.php
kojiromike: I don't really know why they're there too, especially since not _every_ definition in this file has a guard.

kojiromike:
I'm going to do this as a few distinct PRs to avoid having to wait for improvements.

PR #9974, opened 2025-12-23
perf: add composite index on lang_definitions for translation lookups

### LINKED ISSUE #9971: add a composite index on (lang_id, cons_id)
this should make xl lookups much faster.

kojiromike:
Fixes #9971

The `xl()` translation function performs a JOIN query between `lang_definitions` and `lang_constants` tables, filtering by `lang_id`. Without an index on `lang_id`, MySQL must scan the entire `lang_definitions` table for each translation lookup.

- Add composite index `(lang_id, cons_id)` on `lang_definitions` table in `database.sql` for fresh installations
- Add migration in `7_0_4-to-7_0_5_upgrade.sql` to create the index for existing installations

This index allows MySQL to efficiently filter by language and join on `cons_id` in a single index scan, significantly improving translation query performance.

Yes

The changes were generated with Claude Code (Claude Opus 4.5). The commit message includes the standard Claude Code attribution footer.

### DIFF
--- a/sql/7_0_4-to-7_0_5_upgrade.sql
+++ b/sql/7_0_4-to-7_0_5_upgrade.sql
@@ -112,3 +112,7 @@
 --  #IfMBOEncounterNeeded
 --    desc: Add encounter to the form_misc_billing_options table
 --    arguments: none
+
+#IfNotIndex lang_definitions lang_cons
+CREATE INDEX `lang_cons` ON `lang_definitions` (`lang_id`, `cons_id`);
+#EndIf
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -3537,7 +3537,8 @@ CREATE TABLE `lang_definitions` (
   `lang_id` int(11) NOT NULL default '0',
   `definition` mediumtext,
   UNIQUE KEY `def_id` (`def_id`),
-  KEY `cons_id` (`cons_id`)
+  KEY `cons_id` (`cons_id`),
+  KEY `lang_cons` (`lang_id`, `cons_id`)
 ) ENGINE=InnoDB;
 
 -- --------------------------------------------------------
--- a/version.php
+++ b/version.php
@@ -28,7 +28,7 @@ $v_realpatch = '0';
 // is a database change in the course of development.  It is used
 // internally to determine when a database upgrade is needed.
 //
-$v_database = 529;
+$v_database = 530;
 
 // Access control version identifier, this is to be incremented whenever there
 // is a access control change in the course of development.  It is used


sql/7_0_4-to-7_0_5_upgrade.sql
kojiromike, line 1: @bradymiller will some process automatically rename this to `sql/7_0_4-to-8_0_0_upgrade.sql` if necessary?

bradymiller: that process will be me :)
(after we get 7.0.4 out, I'll change the 7.0.5 things to 8.0.0; fyi, and when @Jmevorach 7.0.5 dockers changes get in, then will move it to 8.0.1 and it will use alpine 3.23, nodejs 24, yada yada yada (since 4 weeks is not enough time to test all those changes) )

bradymiller:
@kojiromike , also forgot to mention that need to increment $v_database counter in version.php when make database changes (also should do when add new globals in globals.inc.php)
(fyi this is for those ambitious souls that are using development code in production, so the admin.php script will tell them they need to update their database if they update their development codebase)

version.php
```diff
@@ -28,7 +28,7 @@
 // is a database change in the course of development.  It is used
 // internally to determine when a database upgrade is needed.
 //
-$v_database = 529;
+$v_database = 530;
```
kojiromike, line 31: @bradymiller thanks for the reminder. Easy to forget.

sql/database.sql
```diff
@@ -3537,7 +3537,8 @@ CREATE TABLE `lang_definitions` (
   `lang_id` int(11) NOT NULL default '0',
   `definition` mediumtext,
   UNIQUE KEY `def_id` (`def_id`),
-  KEY `cons_id` (`cons_id`)
+  KEY `cons_id` (`cons_id`),
+  KEY `lang_cons` (`lang_id`, `cons_id`)
```
bradymiller, line 3541: hi @kojiromike ,
It just hit me that this will not work for the translation table since it's overwritten by the language import (note the upgrade stuff in the upgrade script will work).
Definitely need to keep this change for clarity sake, but will need changes in 3 other places to ensure persistence:
1. lang_definitions table in https://raw.githubusercontent.com/openemr/openemr/refs/heads/master/contrib/util/language_translations/currentLanguage_utf8.sql
2. https://github.com/openemr/openemr/blob/master/contrib/util/language_translations/buildLanguageDatabase.pl#L554-L555
3. lang_definitions table in rel-704 (same file as step 1 above, but in rel-704 branch; this is important since we will likely be copying this over for 8.0.0 release)

PR #9978, opened 2025-12-23
perf: add in-memory translation cache

### LINKED ISSUE #9977: create a cache for translations
The xl function does a hefty database lookup, sometimes with 0 results. There are a large number of such lookups for any given request. If we cache the results, we can avoid the cost of a repeated lookup. Better still, if we can build the cache once and reuse it between requests, we can pay the cost of pre-warming that cache. (But, we will need a way to invalidate the cache if someone updates the translation tables in the database.)

kojiromike:
Fixes #9977

The xl() function performs a database query for every translation lookup. With 10,000+ calls per page, this creates significant overhead. This PR adds an in-memory cache to avoid repeated queries.

- Add `TranslationCache` class in `src/Common/Translation/` to manage cached translations
- Add `xlWarmCache()` function to preload all translations for the current language
- Add `translation_preload_cache` global setting (Appearance > Locale) to enable eager loading
- Modify `xl()` to check cache before querying database
- When cache warming is disabled, translations are still cached lazily (per-request memoization)
- Update phpstan baseline for new `$GLOBALS` access

**Performance modes:**
1. **Cache warming disabled (default):** Translations are cached as they're looked up. First lookup hits DB, subsequent lookups use cache.
2. **Cache warming enabled:** One query loads all translations at request start. All subsequent `xl()` calls are pure memory lookups.

Yes

The TranslationCache class and modifications were generated with Claude Code (Claude Opus 4.5). The commit message includes the standard Claude Code attribution footer.

### DIFF
--- a/library/translation.inc.php
+++ b/library/translation.inc.php
@@ -49,10 +49,16 @@ if (!(function_exists('xl'))) {
             $string = '';
         } else {
             // Cache not warmed, query database
-            $sql = "SELECT * FROM lang_definitions JOIN lang_constants ON " .
-            "lang_definitions.cons_id = lang_constants.cons_id WHERE " .
-            "lang_id=? AND constant_name = ? LIMIT 1";
-            $res = sqlStatementNoLog($sql, [$lang_id,$constant]);
+            $sql = <<<'SQL'
+                SELECT lang_definitions.definition
+                  FROM lang_definitions
+                  JOIN lang_constants
+                 USING (cons_id)
+                 WHERE lang_definitions.lang_id = ?
+                   AND lang_constants.constant_name = ?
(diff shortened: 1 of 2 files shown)


kojiromike:
@Firehed I isolated the three different kinds of changes into separate PRs to avoid mixing scope. For this one, I took your idea of using a class into some consideration. I really don't want to mess too much with the translation php file in this changeset. I think I took it too far in the other one. So I created a static class in a separate place, just for the cache handler. Later on, I'd like to make it handle more of the translation features, but baby steps.

For other improvements that I want to do, but are not necessary for this scope, I created #9979.

I'm also kinda not a fan of static classes, but it seemed like the right incremental step here. WDYT?

PR #10021, opened 2025-12-30
fix: remove extra quotes around attr_js() in URLSearchParams

### LINKED ISSUE #10020: Fix double-quoting bug with attr_js() in URLSearchParams refactor
## Description

PR #9741 introduced a bug where `attr_js()` is used inside JavaScript string literals when building `URLSearchParams` objects. 

The `attr_js()` function calls `json_encode()` which adds quotes around string values. When used inside already-quoted JavaScript strings like:

```javascript
    csrf_token_form: '<?php echo attr_js(CsrfUtils::collectCsrfToken()); ?>',
});
```

This produces double-quoted values:
```javascript
    csrf_token_form: '"abc123token"',  // Value includes literal quote characters
});
```

## Affected Files

1. `interface/main/dated_reminders/dated_reminders.php:112`
2. `interface/patient_file/history/encounters.php:228`

## Fix

Replace `attr_js()` with `attr()` in these cases. Since the value is already inside JavaScript string quotes, we only need HTML entity escaping (which `attr()` provides). `URLSearchParams` will handle URL encoding when converting to a query string.

## Related

- #9741 - Original URLSearchParams refactoring PR

kojiromike:
## Summary

- Fixes double-quoting bug where `attr_js()` was used inside JavaScript string literals
- `attr_js()` uses `json_encode()` which adds its own quotes, so outer quotes are not needed

## Test plan

- [ ] Verify dated reminders add functionality works (navigate to dated reminders, click to edit one)
- [ ] Verify document viewing from encounters history works

Fixes #10020
Related to #9741

### DIFF
--- a/interface/main/dated_reminders/dated_reminders.php
+++ b/interface/main/dated_reminders/dated_reminders.php
@@ -109,7 +109,7 @@ function openAddScreen(id){
   } else {
     top.restoreSession();
     const params = new URLSearchParams({
-        csrf_token_form: '<?php echo attr_js(CsrfUtils::collectCsrfToken()); ?>',
+        csrf_token_form: <?php echo js_escape(CsrfUtils::collectCsrfToken()); ?>,
         mID: id
     });
     dlgopen('<?php echo $GLOBALS['webroot']; ?>/interface/main/dated_reminders/dated_reminders_add.php?' + params, '_drAdd', 700, 500);
--- a/interface/patient_file/history/encounters.php
+++ b/interface/patient_file/history/encounters.php
@@ -225,7 +225,7 @@ function todocument(docid) {
   const params = new URLSearchParams({
     doc_id: docid,
     document: '',
-    patient_id: '<?php echo attr_js($pid); ?>',
+    patient_id: <?php echo js_escape($pid); ?>,
     view: ''
   });
   h = '<?php echo $GLOBALS['webroot'] ?>/controller.php?' + params;


kojiromike:
@bradymiller I think these were the only two that had the quoting issue.

interface/main/dated_reminders/dated_reminders.php
```diff
@@ -109,7 +109,7 @@ function openAddScreen(id){
   } else {
     top.restoreSession();
     const params = new URLSearchParams({
-        csrf_token_form: '<?php echo attr_js(CsrfUtils::collectCsrfToken()); ?>',
+        csrf_token_form: <?php echo attr_js(CsrfUtils::collectCsrfToken()); ?>,
```
bradymiller: this looks like pure js context, so would use js_escape instead of attr_js

interface/patient_file/history/encounters.php
```diff
@@ -225,7 +225,7 @@ function todocument(docid) {
   const params = new URLSearchParams({
     doc_id: docid,
     document: '',
-    patient_id: '<?php echo attr_js($pid); ?>',
+    patient_id: <?php echo attr_js($pid); ?>,
```
bradymiller: this looks like pure js context, so would use js_escape instead of attr_js

interface/main/dated_reminders/dated_reminders.php
bradymiller: check out here for example of attr_js:
https://www.open-emr.org/wiki/index.php/Codebase_Security#Cross-Scripting_Prevention

Image

bradymiller: just realized the wiki has typo in the function (js_attr instead of attr_js). will fix it now.

PR #10028, opened 2025-12-31
fix(sql): add migration for care_team_roles primary_care option_id

### LINKED ISSUE #9983: Migration miss in linear master branch
It looks like the changes in https://github.com/openemr/openemr/pull/9208 change `list_options.option_id` from `'primary_care'` to `'primary_care_provider'`, but there is no sql/7_0_3-to-7_0_4_upgrade.sql in that PR.

If I understand the migration process correctly, if someone installed OpenEMR after
4da82ad60dd5a34bd1df70d9930a908dffb95236 and before 231443a23d6bb4b2518784e51d68b174ca5d46e5
and then, after 231443a23d6bb4b2518784e51d68b174ca5d46e5 ran an upgrade to any later version, they'd still have 'primary_care', not 'primary_care_provider' in the database.

I know openemr doesn't tend to support migrations between releases in master, but this one ought to be easy to fix, right?

```
UPDATE `list_options` SET option_id = 'primary_care_provider'
WHERE list_id = 'care_team_roles' AND option_id = 'primary_care';
```

Wouldn't it be worth doing, for continuity and confidence?

kojiromike:
## Summary
- Adds missing migration to update `list_options.option_id` from `'primary_care'` to `'primary_care_provider'` for the `care_team_roles` list
- Uses `#IfRow2D` conditional to only run the update if the old value exists

## Test plan
- [ ] Verify migration runs successfully on a database with the old `primary_care` option_id
- [ ] Verify migration is skipped (no error) on databases already using `primary_care_provider`

Fixes #9983

### DIFF
--- a/sql/patch.sql
+++ b/sql/patch.sql
@@ -47,3 +47,7 @@
 
 --  #EndIf
 --    all blocks are terminated with and #EndIf statement.
+
+#IfRow2D list_options list_id care_team_roles option_id primary_care
+UPDATE `list_options` SET `option_id` = 'primary_care_provider' WHERE `list_id` = 'care_team_roles' AND `option_id` = 'primary_care';
+#EndIf


sql/patch.sql
```diff
@@ -47,3 +47,7 @@
 
 --  #EndIf
 --    all blocks are terminated with and #EndIf statement.
+
+#IfRow2D list_options list_id care_team_roles option_id primary_care
+UPDATE `list_options` SET `option_id` = 'primary_care_provider' WHERE `list_id` = 'care_team_roles' AND `option_id` = 'primary_care';
+#EndIf
```
bradymiller, line 53: hi @kojiromike ,
This script only gets populated in the release branches. If needs to be addressed in master (ie. the next release 8.0.0), then should go to 7_0_4-to-7_0_5_upgrade.sql. If want it in next 7.0.4 patch (don't plan a patch, but it won't hurt to prep in case), then should instead make this same change on the rel_704 branch.

PR #10037, opened 2026-01-02
feat(erx): add Ensora subscription options page

### LINKED ISSUE #10036: Enhance user eRx experience
When people are testing openemr, and they go and enable the eRx. Then they click the Med Entry button in the patient chart. The system shows Not Authorized. It would be better if the user were presented with the subscription option. This way, they don't have to go searching for what to do next.

juggernautsei:
Fixes #10036

 User experience of enabling the NewCrop/Ensora eRx and getting an error message that is not helpful. 

Show the subscription screen to the user to be able to make the purchase


interface/eRx.php
```diff
@@ -20,6 +20,9 @@
 require_once($GLOBALS['fileroot'] . '/interface/eRxXMLBuilder.php');
 require_once($GLOBALS['fileroot'] . '/interface/eRxPage.php');
 
+use OpenEMR\ERx\CredentialValidator;
+use OpenEMR\Common\Globals\OEGlobalsBag;
```
kojiromike:
```suggestion
use OpenEMR\Core\OEGlobalsBag;
```

src/ERx/CredentialValidator.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $missing = [];
+
+        foreach ($requiredFields as $field => $label) {
+            if (!isset($globals[$field]) || empty($globals[$field])) {
+                $missing[$field] = $label;
+            }
+        }
+
+        return $missing;
```
kojiromike: This is a good use case for `array_filter`, since all it really does is filter $requiredFields by a predicate. That way, instead of building up an array, you just do an assignment once.

```suggestion
        $missing = array_filter(
            $requiredFields,
            fn($field) => !isset($globals[$field]) || empty($globals[$field]),
            ARRAY_FILTER_USE_KEY
        );
        return $missing;
```

interface/eRx.php
```diff
@@ -141,11 +141,11 @@ function array_key_exists_default($key, $search, $default = null)
                 ?>
         <strong><?php echo xlt('Ensora call failed'); ?></strong>
         <ul>
-            <?php foreach ($errors as $message) {
-                echo '<li>' . text($message) . '</li>';
-            } ?>
+                <?php foreach ($errors as $message) {
+                    echo '<li>' . text($message) . '</li>';
+                } ?>
         <ul>
```
kojiromike: is this supposed to be a closing tag?

```suggestion
        </ul>
```

juggernautsei:
Copilot wrote:
GitHub Copilot Chat Assistant

Below is a ready-to-apply patch that fixes the immediate cause of job 59345329728 (Apache failing because /etc/ssl/certs/webserver.cert.pem is missing), by adding a runtime entrypoint that ensures a self-signed cert/key pair exists before Apache starts and wiring that entrypoint into the image Dockerfile.

- The job logs (ref 30f772e0fcbc2dd2bb28fe6a96bcda35dceb9876) show:
  "SSLCertificateFile: file '/etc/ssl/certs/webserver.cert.pem' does not exist or is empty" and Apache exits with a syntax error, causing the openemr service to stop and CI tests to fail.

Patch (unified diff)
- Adds docker/openemr/entrypoint.sh (new)
- Modifies docker/openemr/Dockerfile to install the entrypoint and ensure it runs

Please review the patch below. If you want, I can push this in a branch and open a PR.

--- a/docker/openemr/entrypoint.sh
+++ b/docker/openemr/entrypoint.sh
+#!/bin/sh
+set -e
+
+# Entry script to ensure SSL cert/key exist before starting Apache.
+# Generates a self-signed certificate if needed and then execs the main CMD.
+
+CERT="/etc/ssl/certs/webserver.cert.pem"
+KEY="/etc/ssl/private/webserver.key.pem"
+
+# Ensure target directories exist and have reasonable permissions
+mkdir -p "$(dirname "$CERT")" "$(dirname "$KEY")"
+chmod 755 "$(dirname "$CERT")" || true
+chmod 700 "$(dirname "$KEY")" || true
+
+# If cert or key missing/empty, generate a temporary self-signed certificate
+if [ ! -s "$CERT" ] || [ ! -s "$KEY" ]; then
+  echo "Generating self-signed SSL certificate at $CERT and $KEY..."
+  # Use openssl to create a 2048-bit RSA self-signed certificate for localhost
+  # Suppress interactive prompts with -subj
+  openssl req -x509 -nodes -newkey rsa:2048 \
+    -days 365 \
+    -keyout "$KEY" \
+    -out "$CERT" \
+    -subj "/CN=localhost" >/dev/null 2>&1 || {
+      echo "ERROR: openssl failed to generate certificate"; exit 1;
+    }
+  chmod 600 "$KEY" || true
+  chmod 644 "$CERT" || true
+fi
+
+# Print certificate info for CI debugging (non-fatal)
+echo "Certificate subject and validity:"
+openssl x509 -noout -subject -dates -in "$CERT" || true
+
+# Avoid accidental failure when mounted host file hides a generated cert:
+if [ ! -s "$CERT" ] || [ ! -s "$KEY" ]; then
+  echo "ERROR: Certificate or key is missing or empty at $CERT / $KEY"
+  echo "If you mount host paths over /etc/ssl/certs or /etc/ssl/private, ensure"
+  echo "the host files are present and non-empty. Failing to start to avoid"
+  echo "starting apache with invalid SSL config."
+  exit 1
+fi
+
+# Execute the main container command
+exec "$@"

--- a/docker/openemr/Dockerfile
+++ b/docker/openemr/Dockerfile
@@
 # (existing Dockerfile content above)
 
+# Copy entrypoint that guarantees cert/key exist
+COPY docker/openemr/entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+# Use the entrypoint to prepare certs before starting Apache.
+# If the original Dockerfile already had an ENTRYPOINT/CMD, this replaces them.
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+CMD ["apache2ctl", "-D", "FOREGROUND"]
+
 # (existing Dockerfile content may continue below)
 

Notes and guidance for applying the patch
- If the repository's Dockerfile is different in structure (already sets ENTRYPOINT or CMD), merge the entrypoint generation logic into the existing entrypoint rather than duplicating ENTRYPOINT/CMD lines.
- If the image is built with an alternate Apache startup (supervisord or other wrapper), ensure the entrypoint runs before whatever starts Apache.
- Ensure openssl is available in the image. If it's not already installed in the Dockerfile, add a package install step (example for Debian-based images):
  RUN apt-get update && apt-get install -y --no-install-recommends openssl ca-certificates && rm -rf /var/lib/apt/lists/*

CI improvements (recommended, but not required for immediate fix)
- The CI failure output shows the suite attempts docker compose exec while the openemr service is not running. Add a short wait-for-health check and more informative log dumps in the CI script before failing early, for example:
  for i in $(seq 1 30); do
    status=$(docker compose inspect -f '{{ .State.Health.Status }}' openemr 2>/dev/null || echo unknown)
    [ "$status" = "healthy" ] && break
    echo "waiting for openemr health... ($i)"
    sleep 2
  done
  if [ "$status" != "healthy" ]; then
    echo "openemr failed to become healthy. Dumping logs:"
    docker compose logs --no-color openemr || true
    docker compose ps -a || true
    exit 1
  fi

Why this fixes the failing job
- The Apache syntax error in the job (ref 30f772e0fcbc2dd2bb28fe6a96bcda35dceb9876) was caused by SSLCertificateFile pointing to a missing/empty file. The entrypoint ensures these files exist and are non-empty before Apache starts, preventing the syntax error and allowing the container to remain running so the phpunit steps can execute.

Next steps
- Tell me if you want me to:
  1) Create a branch (suggested name: ci/fix-ssl-cert-generation-10037), commit these two files, and open a PR against openemr/openemr.
  2) Or only provide the patch here for you to apply manually.
- If you want the branch/PR, confirm the repo owner is openemr (I will use openemr/openemr) and I will push the change.

adunsulag:
@juggernautsei I don't understand you're last message.  Is the bug due to some kind of ssl error?  I don't understand how its related to the newcrop stuff.

juggernautsei:
@adunsulag, that is my point. The rector error is related to a Docker SSL issue that has nothing to do with my repo. My question is, why am I getting the error if my repo has nothing to do with Docker?
BTW, this is not that I don't understand. The issue is that the rector errors should be specific to my repo. I think this is a general, broad effort by the "lead developers" to force the community to clean up the code. This seems like an all-hands-on-deck move to illicitly move the codebase where the steering committee has decided to push towards. 
But will my repo sit out there if I don't fix the codebase errors?

adunsulag:
> The rector error is related to a Docker SSL issue that has nothing to do with my repo. My question is, why am I getting the error if my repo has nothing to do with Docker?

Why didn't you ask that question in your post?  I'm still confused on whether you are getting this locally, or if you are getting this error on your PR, or what?  I'm very confused on the context on where you're getting this error as you're post didn't give any details.  From my end of things it looks like all of the tests have passed and your PR can be merged in if the code passed review.

As far as your repo sitting out there?  What are you meaning?  Are you asking if your PR won't be merged in if it doesn't pass the CI tests? As far as I can tell your tests passed.

If there's a bug with the CI tests we'll get those fixed if your code isn't causing the issue.  We've run into that before where a change in the linter / tool suite has caused things to fail.  It usually gets caught pretty quickly and fixed but sometimes it doesn't and we just need to be informed that the bug isn't related to your code.

juggernautsei:
@adunsulag I am not getting these error locally. The error only appear in the repo. I run rector and phpstan locally before pushing to the repo. 
Apologies for being a pain.

adunsulag:
@juggernautsei is this in your juggernautsei repo?  On the OpenEMR repo here it appears to be passing all of the tests.

juggernautsei:
Yes, we are good now. Thanks!

interface/eRx.php
adunsulag: As far as I can tell this never got fixed right?  I think this was the only thing holding up this PR.

juggernautsei:
@adunsulag bump

juggernautsei:
@adunsulag, are you still waiting on info?

kojiromike:
@aisle-analyzer please review

kojiromike:
@bradymiller — feedback on this AISLE run in case we want to share it with upstream.

The XXE finding on `simplexml_load_string` is a false positive. Modern PHP (8.0+) with libxml ≥ 2.9 (2012) disables external entity loading by default — `libxml_disable_entity_loader()` was deprecated in 8.0 for exactly this reason, and OWASP lists PHP ≥ 8.0 as XXE-safe by default for SimpleXML/DOM. We've had several GHSAs filed with this framing that closed as not-actionable. Worth gating the rule on detected PHP version, or reframing as defense-in-depth rather than a vulnerability.

The underlying recommendation to use `XmlUtils` was good (project-wide consistency), so leading with that rather than the XXE framing would be more accurate.

kojiromike:
@juggernautsei — quick update on this PR. I picked it up to push the small mechanical fixups @adunsulag and I had been waiting on (the `</ul>` typo, the auth-error substring matching, etc.) so it could land. While reviewing, I noticed the new Twig template renders a full `<!DOCTYPE html>...</html>` document, but `interface/eRx.php` is already inside its own `<html><body>` wrapper — so on the missing-credentials and authentication-failed paths we'd be emitting one HTML document nested inside another (with the inline `<style>` block ending up in the outer `<body>`).

That fix isn't mechanical — the cleanest path is to have the Twig template extend the standard layout and have `interface/eRx.php` short-circuit the `<html>` wrapper for these branches, which is more of a design decision than a fixup. Rather than guess at your intent, I'm going to stop here and let you take it from here when you have time. Three commits are pushed (XmlUtils adoption, `rel="noopener"`, `</ul>` fixes + dropping `xlt` from price literals). Please take a look and decide how you want to handle the layout question. Thanks for your patience on this one.

juggernautsei:
@kojiromike sometimes I step into something that others have worked on. I don't always know the full extent of my ignorance. However, looking at some of the structure of this eRx.php. I can see some things that need to be changed. I was only focused on the link to subscribing to Ensora. I did not investigate the entire page. But I will do my best to fix what you have pointed out. 
The other reason that I am writing is that the Globals are going through a transformation. I understand the transformation path. My concern is about the redundancy of GLOBALS also, when I came back into the file now. The use OpenEMR\Core\OEGlobalsBag; is in there twice above and below the require statements. I don't think that is proper coding technique. I am going to spend some time reviewing the code and structure today.

juggernautsei:
@coderabbitai fix-ci

juggernautsei:
@bradymiller bump, I think this is ready to be merged

juggernautsei:
@adunsulag @kojiromike What info is needed to bring this into the codebase?

PR #10040, opened 2026-01-02
revert: remove type declarations from translation functions

kojiromike:
## Summary
Reverts the runtime type hints (`string` parameter and return types) from `xl()` and wrapper functions added in #9960. The docblock comments are retained for IDE/static analysis support.

This allows for gradual adoption of stricter typing without breaking existing callers that may pass non-string values.

## Test plan
- [ ] Verify translations work correctly
- [x] Run PHPStan and confirm no new errors introduced

### DIFF
--- a/library/translation.inc.php
+++ b/library/translation.inc.php
@@ -25,7 +25,7 @@ if (!(function_exists('xl'))) {
      * @param string $constant The text constant to translate
      * @return string The translated string
      */
-    function xl(string $constant): string
+    function xl($constant)
     {
         if (!empty($GLOBALS['disable_translation']) || !empty($GLOBALS['temp_skip_translations'])) {
             return $constant;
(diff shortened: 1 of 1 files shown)


src/ClinicalDecisionRules/Interface/RuleLibrary/ReminderIntervalDetail.php
```diff
@@ -34,7 +34,7 @@ function __construct(public $intervalType, public $intervalRange, public $amount
     function display()
     {
         $display = xl($this->intervalRange->lbl) . ": "
-            . xl($this->amount) . " " . xl($this->timeUnit->lbl);
+            . $this->amount . " " . xl($this->timeUnit->lbl);
```
kojiromike: @bradymiller here are three places where what's passed to xl are not literals. I removed one that was a number, but there are two more… should they stay?

kojiromike: This isn't what this PR is for anyway, so I'm going to move this to a separate issue.

Issue #9754, opened 2025-12-10
Composer - Specify type


igormukhingmailcom:
### Describe the bug

When we run `composer validate` - we can see `The version field is present, it is recommended to leave it out if the package is published on Packagist.` warning. Its because we should have `"type": "project"` to be specified at `composer.json`

PR #10065, opened 2026-01-03
fix(composer): to pass validation by removing version

igormukhingmailcom:
Fixes: https://github.com/openemr/openemr/issues/9754

Before:

image

After:

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,6 @@
     "description": "OpenEMR is a Free and Open Source electronic health records and medical practice management application",
     "license": "GPL-3.0-or-later",
     "type": "project",
-    "version": "8.0.0",
     "support": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -3,7 +3,6 @@
     "description": "OpenEMR is a Free and Open Source electronic health records and medical practice management application",
     "license": "GPL-3.0-or-later",
     "type": "project",
-    "version": "8.0.0",
```
kojiromike, line 6: @bradymiller does this impact the release process?

bradymiller: hi @kojiromike 
no, it will just be 1 less thing to worry about :)
(will remove needing to make this change from the release howto docs)

PR #10068, opened 2026-01-03
fix(composer): memory limit increased for composer code quality scripts

igormukhingmailcom:
... as `1g` is not enough


kojiromike:
I empathize with this, but I think a limit is still necessary. Maybe we should set it to 4g and see if that does the trick. We don't want to allow it to grow without knowing that it's growing.

igormukhingmailcom:
@kojiromike Adjusted

kojiromike:
There's no intersection between the tests and this change, so I'm going to merge it without waiting.

PR #10072, opened 2026-01-03
fix(phpstan): indent size for neon files at .editorconfig

igormukhingmailcom:
Related: https://github.com/openemr/openemr/issues/10078
Related: https://github.com/openemr/openemr/issues/10077

Draft reason - we should decide which formatting suit best for us first. I have feeling we should migrate to php (for performance reasons) and this PR will be not needed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
 
-[*.{json,jsn,sh,yaml,yml}]
+[*.{json,jsn,sh,yaml,yml,neon,neon.dist}]
 indent_size = 2
 
 [composer.{json,lock}]


.editorconfig
```diff
@@ -10,5 +10,8 @@ trim_trailing_whitespace = true
 [*.{json,jsn,sh,yaml,yml}]
 indent_size = 2
 
+[*.{neon,neon.dist}]
+indent_size = 2
```
kojiromike, line 11: I intend to treat neon as yaml as much as possible, so let's add this to the above list of json/yaml.

igormukhingmailcom:
@kojiromike Done

kojiromike:
@igormukhingmailcom this is cheap. Let's merge it and then have the discussion.

igormukhingmailcom:
Ok, @kojiromike

PR #10082, opened 2026-01-04
chore: remove RestConfig

### LINKED ISSUE #10079: chore: remove RestConfig
Seems like the only thing we actually use at `RestConfig` is `request_authorization_check` which should be extracted to some `AuthChecker` class.

Everything else is not used and never instantiated/initialized.

`RestConfigCaseSensitivityTest` should also be removed as we don't use `is_*_request` methods.

igormukhingmailcom:
Fixes: https://github.com/openemr/openemr/issues/10079

## Merging notes

- [x] Should be rebased after https://github.com/openemr/openemr/pull/9759


igormukhingmailcom:
@kojiromike This one ready for merge I believe.

Last commit (where `RestConfig` removed) is questionable - if we need to keep `RestConfig` for BC - we can drop it.

igormukhingmailcom:
@adunsulag @bradymiller Could you review please as well?

src/RestControllers/Authorization/RestApiAclPermissionChecker.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
```
adunsulag, line 2: Missing copyright header here.

src/RestControllers/Authorization/RestApiAclPermissionChecker.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+class RestApiAclPermissionChecker
+{
+    use SingletonTrait;
+
+    public function check(
+        HttpRestRequest $request,
+        string $section,
+        string $value,
+        string $aclPermission = '',
```
adunsulag: $aclPermission can accept an array or a string so you need to handle that with the type check here.

adunsulag (changes requested):
So this is a breaking change and I believe it will need to go into 8.0.  People in the past used the RestConfig::$ROUTE_MAP and other static vars in their custom code.  

Two minor changes needed.  Otherwise I think this is fine.

src/RestControllers/Authorization/RestApiAclPermissionChecker.php
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Usage:
+ *   RestApiAclPermissionChecker::getInstance()->check($request, 'admin', 'users')
+ */
+class RestApiAclPermissionChecker
+{
+    use SingletonTrait;
+
+    public function check(
```
bradymiller, line 30: Hi. Just curious.
Isn't this basically just a static function call?
Is using the Singleton part of a future plan?

bradymiller:
looks ok to me (after issues by @adunsulag addressed)

src/RestControllers/Authorization/RestApiAclPermissionChecker.php
igormukhingmailcom: I believe new code should not have static functions (except factories) to be able to cover them with tests.

(This particular class not tested and not able to be tested as uses static call `AclMain::aclCheckCore`, but I just following that idea).

src/RestControllers/Authorization/RestApiAclPermissionChecker.php
igormukhingmailcom: Thanks. Fixed

src/RestControllers/Authorization/RestApiAclPermissionChecker.php
igormukhingmailcom: Thanks. Added.

PR #10087, opened 2026-01-05
fix: js error when samesite param missing

### LINKED ISSUE #10096: JavaScript error when session cookie samesite parameter is empty
## Description

A JavaScript syntax error occurs in `library/restoreSession.php` when the `samesite` session cookie parameter is empty or not set.

## Current Behavior

When `$scparams['samesite']` is empty, the code outputs:
```javascript
var oemr_scp_samesite = ;
```
This is invalid JavaScript syntax and causes a script error.

## Expected Behavior

The variable should be assigned a valid JavaScript value even when the samesite parameter is empty.

## Steps to Reproduce

1. Configure OpenEMR with a session cookie that has no `samesite` attribute
2. Load a page that includes `library/restoreSession.php`
3. Observe JavaScript error in browser console

## Affected File

- `library/restoreSession.php` (line 37)

## Environment

- OpenEMR version: latest/master

mdsupport:
Fixes #10096

Fixes a JavaScript syntax error that occurs when the session cookie `samesite` parameter is empty or not set. The generated JavaScript was outputting an empty value without quotes, resulting in invalid syntax.

- Changed the fallback value from empty string `''` to string `'false'` when `$scparams['samesite']` is empty in `library/restoreSession.php`

No

N/A

### DIFF
--- a/library/restoreSession.php
+++ b/library/restoreSession.php
@@ -34,7 +34,7 @@ var oemr_scp_lifetime = <?php echo js_escape($scparams['lifetime']); ?>;
 var oemr_scp_path = <?php echo js_escape($scparams['path']); ?>;
 var oemr_scp_domain = <?php echo js_escape($scparams['domain']); ?>;
 var oemr_scp_secure = <?php echo js_escape($scparams['secure']); ?>;
-var oemr_scp_samesite = <?php echo empty($scparams['samesite']) ? '' : js_escape($scparams['samesite']); ?>;
+var oemr_scp_samesite = <?php echo empty($scparams['samesite']) ? 'false' : js_escape($scparams['samesite']); ?>;
 var oemr_cookie = '';
 var oemr_change_count = 0; // debugging
 


library/restoreSession.php
```diff
@@ -34,7 +34,7 @@
 var oemr_scp_path = <?php echo js_escape($scparams['path']); ?>;
 var oemr_scp_domain = <?php echo js_escape($scparams['domain']); ?>;
 var oemr_scp_secure = <?php echo js_escape($scparams['secure']); ?>;
-var oemr_scp_samesite = <?php echo empty($scparams['samesite']) ? '' : js_escape($scparams['samesite']); ?>;
+var oemr_scp_samesite = <?php echo empty($scparams['samesite']) ? 'false' : js_escape($scparams['samesite']); ?>;
```
kojiromike, line 37: good fix. One minor suggestion:

```suggestion
var oemr_scp_samesite = <?php echo js_escape($scparams['samesite'] ?: false); ?>
```

Both produce identical output, but the second version uses js_escape() for both branches, which is more consistent with the above implementations.

Issue #10078, opened 2026-01-04
style(phpstan): baselines has different formatting


igormukhingmailcom:
Related: https://github.com/openemr/openemr/issues/10077

### Describe the bug

When we generate baseline - it has 4 spaces when what we have committed at repo has 2 spaces.

We should choose one format to be able to update baselines faster / without needing to fix that manually.

At repo:

<img width="554" height="197" alt="Image" src="https://github.com/user-attachments/assets/be2887ed-0f9f-400e-b0d1-c2c903db75b7" />

Generated:

<img width="525" height="266" alt="Image" src="https://github.com/user-attachments/assets/02fa4e02-3771-4d53-99a5-c5f9dd91d110" />

kojiromike:
Usually I do `--generate-baseline` followed by `expand | yq -P .` to reformat the output. I must've forgotten in this case.

Issue #10076, opened 2026-01-04
docs(phpstan): add one-liners to update baseline at contributing guide


igormukhingmailcom:
Depends on: https://github.com/openemr/openemr/issues/10077

Like this one:

```bash
php -d memory_limit=8g ./vendor/bin/phpstan analyze \
    --configuration=phpstan.neon.dist \
    --generate-baseline=.phpstan/phpstan-globals-baseline.neon
```

kojiromike:
```
php -d memory_limit=8g ./vendor/bin/phpstan analyze \
    --configuration=phpstan.neon.dist \
    --generate-baseline=.phpstan/temp.neon
expand .phpstan/temp.neon |
    yq -P . > .phpstan/phpstan-globals-baseline.neon
rm .phpstan/temp.neon
```

We could put these in composer.json too, of course.

PR #10099, opened 2026-01-05
feat(phpstan): openemr-related rules baselines update script

igormukhingmailcom:
Fixes: https://github.com/openemr/openemr/issues/10076
Fixes: https://github.com/openemr/openemr/issues/10078

Script is updating **ALL** baseline files:

- `.phpstan/phpstan-curl-baseline.neon`
- `.phpstan/phpstan-database-baseline.neon`
- `.phpstan/phpstan-globals-baseline.neon`
- `.phpstan/phpstan-remaining-baseline.neon`

## Usage

```bash
composer run phpstan-baseline
```

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -69,15 +69,14 @@ jobs:
     # Generate baseline only if analyze step failed
     - name: PHPStan Baseline
       if: steps.phpstan-analyze.outcome == 'failure'
-      run: vendor/bin/phpstan --memory-limit=8G --generate-baseline=baseline.neon
+      run: composer run phpstan-baseline
       continue-on-error: true
-    # Upload baseline only if baseline.neon file exists
     - name: Upload PHPStan Baseline
-      if: hashFiles('baseline.neon') != ''
+      if: steps.phpstan-analyze.outcome == 'failure'
       uses: actions/upload-artifact@v6
       with:
         name: phpstan-baseline-php${{ matrix.php-version }}
-        path: baseline.neon
+        path: .phpstan/phpstan-*-baseline.neon
       continue-on-error: true
     - name: Check PHPStan Results
       if: steps.phpstan-analyze.outcome == 'failure'


kojiromike:
> we should move ignoreErrors from it to separate file, like .phpstan/phpstan-main-baseline.neon.

Yes, good

kojiromike:
@zmilan you can test this script out to address the phpstan baseline trips in #9868

zmilan:
@kojiromike @igormukhingmailcom I have tested this, and it works well

igormukhingmailcom:
> > we should move ignoreErrors from it to separate file, like .phpstan/phpstan-main-baseline.neon.
> 
> Yes, good

Done. Ready for merge.

kojiromike:
@igormukhingmailcom one more thing -- if the phpstan github action fails, it produces a baseline, but this is better. Should we update the action to upload all the updated baselines? This is supposed to be convenient for people who are contributing code via copilot or codespaces.

igormukhingmailcom:
> @igormukhingmailcom one more thing -- if the phpstan github action fails, it produces a baseline, but this is better. Should we update the action to upload all the updated baselines? This is supposed to be convenient for people who are contributing code via copilot or codespaces.

Yeah, also was thinking about that...
Will adjust in a moment.

.github/workflows/phpstan.yml
```diff
@@ -69,15 +69,14 @@ jobs:
     # Generate baseline only if analyze step failed
     - name: PHPStan Baseline
       if: steps.phpstan-analyze.outcome == 'failure'
-      run: vendor/bin/phpstan --memory-limit=8G --generate-baseline=baseline.neon
+      run: composer run phpstan-baseline
       continue-on-error: true
-    # Upload baseline only if baseline.neon file exists
     - name: Upload PHPStan Baseline
-      if: hashFiles('baseline.neon') != ''
+      if: steps.phpstan-analyze.outcome == 'failure'
```
igormukhingmailcom, line 75: I changed condition to be consistent across 3 steps

kojiromike: OK. What if the outcome is failure, but phpstan still couldn't update the baseline?

kojiromike: We could get the hashFiles before and after, and compare them, I suppose.

igormukhingmailcom: I guess it unable to update baseline only when original neon files are broken, but in that case things fail on diagnose step

kojiromike (approved):

PR #10113, opened 2026-01-05
fix: deselect initial empty option in generate_select_list

### LINKED ISSUE #10112: bug: deselect initial empty option in generate_select_list
### Describe the bug

Undefined array key 0 in /var/www/localhost/htdocs/openemr/library/options.inc.php on line 249

code doesn't check if array is empty

### To Reproduce

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

stephenwaite:
Fixes #10112

### DIFF
--- a/library/options.inc.php
+++ b/library/options.inc.php
@@ -23,7 +23,7 @@
  * @copyright Copyright (c) 2026 Stephen Waite <stephen.waite@open-emr.org>
  * @copyright 2026 OpenCoreEMR Inc
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
- * 
+ *
  * Note: there are translation wrappers for the lists and layout labels
  * at library/translation.inc.php. The functions are titled xl_list_label()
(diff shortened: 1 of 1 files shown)


library/options.inc.php
```diff
@@ -246,7 +246,7 @@ function generate_select_list(
                 (strlen($currvalue ?? '') > 0 && in_array($lrow['option_id'], $selectedValues))
             ) {
                 // Deselect the initial empty option if a real selection is made.
-                if (!$multiple && $_options[0]['value'] === '' && $_options[0]['isSelected']) {
+                if (!$multiple && !empty($options) && $options[0]['value'] === '' && $_options[0]['isSelected']) {
                     $_options[0]['isSelected'] = false;
                 }
```
kojiromike: You don't actually need to check if isSelected is true, since assigning `false` won't change the value if it's already false.

```suggestion
    if (!$multiple && ($_options[0]['value'] ?? null) === '') {
        $_options[0]['isSelected'] = false;
    }
```

Issue #9981, opened 2025-12-24
Discussion: module overhaul


Firehed:
## Goal

Make adding functionality to OpenEMR via modules the default easy path, and do so in a way that sets the project up for success around maintainability and ease of contribution. Make core development easier, faster, and more secure by separating optional functionality into modules. Be in a position to potentially slim down core so end-users can have simpler installations.

## Summary

Define a series of PHP interfaces that modules can use to expose different types of functionality. For new modules, implementing these interfaces becomes the only supported way to add features. This encourages writing high-quality, easy-to-test code.

There can be first- and third-party modules; both behave the same way in terms of installation and usage. Existing bundled modules should be migrated to this system, and a portion of "core" code should also get extracted out into modules.

Early in the application lifecycle (this is effectively gated on the front-controller work, see #9943) the bootstrapping process would detect and initialize any modules that are installed and active, so that e.g. their CLI commands, API routes, or data layers would be available. this would be done in a context-aware way; e.g. CLI commands wouldn't initialize on web requests, and API routes wouldn't be configured on the command line.

## High-level concept

- Define interfaces for first- and third-party pluggable components. Each should exist as a composer package.
- Create a code-driven path for bootstrapping modules  
- Create CLI utilities for detecting/installing modules into the code path  
- Create API (reachable through CLI and probably web) for enabling/disabling modules  
- Idea: modules at runtime are given their own database+user and can only interact with "core" tables through clear API boundaries - basically have a way to limit the blast radius of an insecure module  
- Ideal case: modules can be developed and tested (relatively) independently of core

### Caveats

- Names are conceptual and not intended to be taken literally. For discussion purposes, let's focus on the underlying concepts.  
- Executing on this as described requires a number of other foundational aspects to happen first

## Open Questions

- What *is* a module? What belongs in core vs an optional package?  
- What (other) types of modules exist?  
- How (if at all) can a module "plug in" to core UI  
- What does the transition path look like?  
- Can a module indicate that it supports an air-gapped installation?  
- What level of database isolation can modules support? Where exactly do we draw the boundaries to core?  
- Can modules talk to each other at all? If so, what are the API boundaries?  
- How can a module have sub-modules? Example: a top-level payments module, with individual gateways (stripe, authorize.net, etc) being "child" modules

## Interfaces

Modules should have a top-level entrypoint that vend their services to the rest of the application, similar to how this works today. The mechanics of doing so are likely to change.

[...]

PR #10114, opened 2026-01-06
Draft: CLI module system

Firehed:
Towards #9981

Discussion-level prototype of a new module foundation & CLI modules

This demonstrates what new modules could look like according to the vision document in #9981. It's fully backwards-compatible with existing modules.

- Defines `ModuleInterface` and `CliModuleInterface`
- Adds a low-level bootstrap script to normalize the runtime
- Adds a new CLI entrypoint[^cli]
- Includes a `Module Manager` module, which adds CLI commands to list and manage modules[^manager]
- Sets up some PSR-11 DI tooling with autowiring support[^di]

Of note, see the CLI command registration process in `bin/cli:L17-21`. We'd do something quite similar for API modules (add routes into router), DB migrations, etc.

Also see https://github.com/Firehed/openemr-module-sample - this is more representative of what a module developer would actually do.

Where I hope this goes:

- Other module types, of course (most require some broader consensus on tooling)
- The module interfaces will move out to an external Composer package[^interfaces], probably `openemr/module-interfaces`. OpenEMR will require this package, as will modules themselves. See how various PSR package work.
- The module-management module might also move out to a separate package, also a core dependency. Right now it's kinda half-module, half-core.
- The bootstrap would also get included by a future front-controller
- Module validation tooling, template repo, etc. This would include tools for module developers as well as OpenEMR admins as part of a certification process
- I don't love the composer `extra` setup path[^module-bootstrap] for a number of reasons and want to refine this

What needs work (mostly niceties rather than blockers):

- Basically everything with `getModuleManager()`
- The internals of module detection are... funky.
- Tidy up composer.json, etc
- Unit test coverage
- As-is, this should be split into several smaller PRs. It's intentionally not right now in order to demonstrate the broad concept
- `ModuleInfo` feels somewhat duplicative. It's an internal concern and thus not a design-blocker, but I don't love it
- `ModuleInterface::getConfigFiles()` can probably be removed if we're able to be strict about conventions, and I'm inclined to do so.

Looking forward to discussing more, both on admin calls and in comments.

#### Does your code include anything generated by an AI Engine? 
No

[^cli]: For BC reasons, this intentionally ignores the existing CLI entrypoints.
[^manager]: Enable/disable doesn't actually do anything yet, since it's not really important for demonstrating the concept. It'll be done prior to landing.
[^interfaces]: I think we only want to do this once we're very confident the interfaces are stable. It can be done incrementally.
[^di]: I used my own library since it was the fastest way for me to prototype and I know it'll do what we need, but most any PSR-11 container with autowiring could work. This DOES likely impact the `getConfigFiles()` mechanic or its replacement
[^module-bootstrap]: The existing system uses `openemr.bootstrap.php`. I want to avoid a procedural/side-effecty module loader, but something convention-driven might be good, and probably faster. The composer file parsing took ~240ms on a M2 Pro, hence adding the cache layer which cut things to about 1ms.


adunsulag:
What are you're concerns with Backwards Compatibility with using the existing command runner?   I would think that the OpenEMR SymfonyCommandRunner could be extended to grab all of the module commands for both existing modules and the new modules if there are specific conventions we follow for the new system vs old.

Also we'll likely want to merge what you've done with your sample into this repo https://github.com/openemr/oe-module-custom-skeleton or deprecate the skeleton repo as we've pointed people to that in our documentation for a sample repo to get started with a module.

Its too bad the laminas stuff is not really maintained that much anymore as your config file felt a lot like the laminas application bootstrap process.  Do you prefer the PHP file format versus the symfony yaml configs used for service injection?

Overall I'm not seeing a ton of problems but just a few thoughts here.

Firehed:
@adunsulag thanks for the feedback!

> What are you're concerns with Backwards Compatibility with using the existing command runner?

Mostly I didn't want to dig too deep while prototyping. I have a hunch that some of the commands in that path rely on legacy paths that I don't want to load at all (e.g. literally everything with globals) but I've done nothing to verify that's an actual problem. Collapsing these would be a fast-follow

> ...  https://github.com/openemr/oe-module-custom-skeleton  ...

Yes! Once this is a bit more solid, I'd want to see all of that centralized and tidied up. My vision here is a Github template repo so module developers have the option to just hit a button and get going. Since docs are already referencing that one, my default would be to get it updated to use the new system. However, there may be some cart-before-the-horse in this approach since we'd likely be adding functionality iteratively, so perhaps a separate repo would be a better choice.

>  Do you prefer the PHP file format versus the symfony yaml configs used for service injection?

Strongly. YAMLs are IMO too disconnected from static analysis tools, plus it requires either installing the PECL extension or having a pure-PHP parser which is a performance concern since this happens on every request. I actually tried to prototype this with the symfony di tooling on the PHP path (the dependency is already installed!) but kept getting misled by the docs, so paused on that for the sake of making progress. Fully onboard with revisiting this - that decision isn't really intended as part of the proof-of-concept.

kojiromike:
@Firehed probably unsurprisingly, I'm finding it immediately useful to have a local cli tool for installing and enabling arbitrary modules in openemr, especially for e2e testing of a module. IIUC, this code does that, albeit a lot more to boot. What do you think about taking that chunk out of this PR and making it one of the first things to implement and merge?

Firehed:
This has been effectively replaced by #12643. It is spiritually similar, but better builds on foundation laid since this was first prototyped.

PR #10116, opened 2026-01-06
fix(health): skip auth for readyz endpoint to prevent login redirect

### LINKED ISSUE #10115: Health endpoint /meta/health/readyz redirects to login instead of returning health status
## Description

The `/meta/health/readyz` endpoint redirects to the login page instead of returning a JSON health status. This makes it unsuitable for use as a Docker/Kubernetes readiness probe.

## Expected Behavior

```bash
curl -sk https://localhost/meta/health/readyz
# Should return: {"status":"ok","checks":{...}}
```

## Actual Behavior

```bash
curl -sk https://localhost/meta/health/readyz
# Returns JavaScript that redirects to login:
# <script>
#  w.top.timed_out = true;
#  w.top.location.href = '/interface/login_screen.php?error=1&site=default';
# </script>
```

## Root Cause

The `/meta/health/readyz` endpoint in `meta/health/index.php` loads `interface/globals.php` to bootstrap the database connection:

```php
require_once __DIR__ . "/../../interface/globals.php";
```

However, `globals.php` initializes session handling and authentication checks, which trigger the login redirect when no valid session exists.

## Workaround

The `/meta/health/livez` endpoint works correctly since it doesn't load `globals.php`.

## Suggested Fix

The readyz endpoint needs a lighter bootstrap path that:
1. Loads database configuration without full session/auth initialization
2. Or explicitly disables session handling before loading globals
3. Or uses a dedicated minimal bootstrap for health checks

## Environment

- OpenEMR: development (flex docker image)
- Discovered while implementing #10085 (split setup/startup in docker)

kojiromike:
Fixes #10115

The `/meta/health/readyz` endpoint redirects to the login page instead of returning JSON health status, making it unusable as a Docker/Kubernetes readiness probe.

- Set `$ignoreAuth = true` before loading `globals.php` in the health check endpoint
- This follows the existing OpenEMR pattern for endpoints that don't require authentication
- The endpoint now returns proper JSON health status without requiring a session

Yes

The fix was generated with Claude Code (Claude Opus 4.5). The change is a 3-line addition that follows existing codebase patterns.

### DIFF
--- a/tests/Tests/Api/HealthEndpointTest.php
+++ b/tests/Tests/Api/HealthEndpointTest.php
@@ -1,0 +1,100 @@
+<?php
+
+/**
+ * Health Endpoint Tests
+ *
+ * Verifies that health check endpoints return proper JSON responses
+ * without requiring authentication (no login redirect).
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2025 OpenCoreEMR Inc
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Tests\Api;
+
+use GuzzleHttp\Client;
+use PHPUnit\Framework\TestCase;
+
+class HealthEndpointTest extends TestCase
+{
+    private Client $client;
+
+    protected function setUp(): void
+    {
+        $baseUrl = getenv("OPENEMR_BASE_URL_API", true) ?: "https://localhost";
+        $this->client = new Client([
+            "verify" => false,
+            "base_uri" => $baseUrl,
+            "timeout" => 10,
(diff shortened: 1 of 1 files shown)


meta/health/index.php
```diff
@@ -55,6 +55,9 @@
             // NOTE: This loads the full OpenEMR framework on each probe request.
             // For high-frequency probes, a lighter bootstrap path that only
             // initializes the database connection could improve performance.
+            //
+            // Skip authentication - health checks must work without a session
+            $ignoreAuth = true;
```
bradymiller, line 60: hi @kojiromike ,
Just checking if it makes sense to guard this mechanism with an environment variable (since only pertinent to docker/orchestration).

kojiromike: @bradymiller it kinda does, but I have questions. Can we talk about it at the admin meeting today?

bradymiller: healthchecks on agenda :)

PR #10134, opened 2026-01-07
ci: skip slow tests for draft PRs with label

### LINKED ISSUE #10132: Skip slow test-all.yml workflow for draft PRs with specific label
## Problem

When iterating on a PR, we often want to run the quick CI checks (phpstan, styling, isolated-tests, etc.) without waiting for the slow `test-all.yml` workflow that runs the full test suite across multiple PHP/webserver/database configurations.

Currently, our options are:
- Use `[no ci]` in commit messages — but this skips **all** CI, not just the slow tests
- Wait for all tests to run on every push

## Proposed Solution

Skip `test-all.yml` when **both** conditions are met:
1. The PR is in **draft** mode
2. A specific label is applied (e.g., `skip-slow-tests` or `quick-ci`)

This allows contributors to iterate quickly while still getting linting and static analysis feedback, then enable full tests when ready for review.

kojiromike:
Fixes #10132

## Description
Allow contributors to skip the slow `Test All Configurations` workflow by adding a `Skip-Slow-Tests: true` trailer to their commit message.

- Add `check-skip` job that detects the trailer and fails immediately with a clear error
- Document the feature in `ci/README.md`

## AI Disclosure
Yes

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -24,8 +24,31 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
+  check-skip:
+    runs-on: ubuntu-24.04
+    # Check for skip marker (see issue #10132)
+    if: github.event_name == 'pull_request'
+    outputs:
+      skip: ${{ steps.check.outputs.skip }}
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v6
+      with:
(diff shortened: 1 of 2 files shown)


Firehed:
The only potential issue here is that contributors that aren't org members can't actually edit their own labels :/

kojiromike:
> The only potential issue here is that contributors that aren't org members can't actually edit their own labels :/

aw, crud, you're totally right. I'll just go back to using a commit message signal, even though they're obnoxious.

PR #10151, opened 2026-01-09
fix: fax module php fatal error memory size exhausted

### LINKED ISSUE #10150: bug: fax module php fatal error allowed memory size exhausted
### Describe the bug

>PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 4096 bytes) in /var/www/localhost/htdocs/openemr/vendor/adodb/adodb-php/drivers/adodb-mysqli.inc.php on line 1271, referer: /interface/modules/custom_modules/oe-module-faxsms/messageUI.php?type=fax

stephenwaite:
Fixes #10150 

duplicate faxes were getting stored in the table

return the id of the existing fax in the table

### DIFF
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/EtherFaxActions.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/EtherFaxActions.php
@@ -795,7 +795,7 @@ class EtherFaxActions extends AppDispatch
         $uid = $_SESSION['authUserID'];
         $jobId = $faxDetails->JobId;
 
-        // avoid duplicates
+        // avoid duplicates, Claude helped craft the query
         $existing = QueryUtils::querySingleRow("SELECT `id` FROM `oe_faxsms_queue` WHERE `job_id` = ? LIMIT 1", [$jobId]);
         if ($existing !== false && !empty($existing['id'])) {
             return (int)$existing['id'];  // Return existing ID instead of inserting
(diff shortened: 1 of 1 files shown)


bradymiller (approved):

interface/modules/custom_modules/oe-module-faxsms/src/Controller/EtherFaxActions.php
```diff
@@ -793,6 +794,13 @@ public function insertFaxQueue($faxDetails): int
         $account = $this->credentials['account'];
         $uid = $_SESSION['authUserID'];
         $jobId = $faxDetails->JobId;
+
+        // avoid duplicates, Claude helped craft the query
+        $existing = QueryUtils::querySingleRow("SELECT `id` FROM `oe_faxsms_queue` WHERE `job_id` = ? LIMIT 1", [$jobId]);
```
sjpadgett, line 799: hmm this shouldn't happen unless a fax was not marked as downloaded on upstream on previous insert.
Kind of worried previous insert was incomplete or errored.
It'd be safer to replace existing.

kojiromike (changes requested):
You already said it yourself. I don't think this can really be done effectively without a unique id. There's basically a TOCTOU race condition in this implementation. I suggest you add the inevitable SQL change.

kojiromike:

Reviewers have requested changes and the PR has been stale. Wait for a refreshed revision before backporting.

PR #10171, opened 2026-01-12
feat(faxsms): add SignalWire integration

### LINKED ISSUE #10150: bug: fax module php fatal error allowed memory size exhausted

### LINKED ISSUE #10169: FaxSMS service introduction updates
@bradymiller asked that this new PR be opened for https://github.com/openemr/openemr/pull/10035#issuecomment-3737206415

juggernautsei:
Fixes #10169
Fixes #10150

### DIFF
--- a/.phpstan/baseline/openemr.deprecatedSqlFunction.php
+++ b/.phpstan/baseline/openemr.deprecatedSqlFunction.php
@@ -2198,7 +2198,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Use QueryUtils\\:\\:querySingleRow\\(\\) or QueryUtils\\:\\:fetchRecords\\(\\) instead of sqlQuery\\(\\)\\.$#',
-    'count' => 8,
+    'count' => 9,
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-faxsms/src/BootstrapService.php',
 ];
 $ignoreErrors[] = [


interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -19,7 +19,7 @@
 $_GET['auth'] = 'portal';  // Enable site selection
 $_GET['site'] ??= 'default';
```
sjpadgett: If on a multisite defaulting to default site might be an issue that you're working from wrong site. If a site id is expected here shouldn't you fail?

juggernautsei: @sjpadgett, you are right. The system should die if the site is not specified.

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -183,6 +275,12 @@ function downloadAndStoreFaxMedia(
     int $patientId = 0
 ): void {
     try {
+        // Validate mediaUrl to prevent SSRF attacks
+        if (!isValidSignalWireUrl($mediaUrl)) {
```
bradymiller: dance-bear

bradymiller:
hi @juggernautsei , Security issue looks solved. Will bring this in when have the site issue resolved.

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
juggernautsei: I don't understand the dancing bear for !isValidSignalWireUrl. This function does not exist. Do you want it created? Then we can all dance?

bradymiller: dancing since it will shut down the vulnerability :)
(your isValidSignalWireUrl() function is with your other validation functions)

sjpadgett:
@juggernautsei I strongly recommend storing and handling you fax documents in the table along with its meta data. For etherfax in bundle the fax download fetch by bundling the meta and fax in a json and stored in table. Wouldn't be too hard as you've done a good job of coverage and could clone some of my methods.
I learned from my Twilio fax server where besides easier for bad actors to access file system its just a maintenance nightmare managing IMO, plus could be more expensive on cloud for volume costs.
Sorry to hit you at this late stage of PR and of course it's up to as looks like you addressed the encryption concerns by Brady.
Let me know if I can help or your thoughts otherwise happy to merge this version.

juggernautsei:
I can adjust the project to accommodate the request. It's like when the construction industry got air hammers. The newbies could nail as fast as the best manual hammerer.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/EtherFaxActions.php
```diff
@@ -785,23 +791,108 @@ public function getCallLogs()
...
-     * @return int
+     * Insert inbound fax into queue with document storage
+     *
+     * Uses FaxDocumentService for consistent document handling and patient matching.
+     *
+     * @param object $faxDetails EtherFax FaxResult object with fax details
+     * @return int Queue record ID
+     * @throws FaxDocumentException
      */
     public function insertFaxQueue($faxDetails): int
```
sjpadgett: You need to revert this and create this action in your service.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/EtherFaxActions.php
```diff
@@ -900,7 +1002,12 @@ public function fetchFaxFromQueue($jobId, $id = null): mixed
      */
     public function fetchQueueCount(): int
     {
-        return (int)sqlQuery("SELECT COUNT(id) as count FROM `oe_faxsms_queue` WHERE deleted = 0")['count'] ?? 0;
+        $siteId = $_SESSION['site_id'] ?? 'default';
```
sjpadgett: This breaks legacy and you need to clone to actions in your service.
Do for all methods you are sharing with this services actions.
Also need to get out of the habit of defaulting to default for site_id

interface/modules/custom_modules/oe-module-faxsms/src/Controller/EtherFaxActions.php
```diff
@@ -793,106 +793,20 @@ public function getCallLogs()
     /**
      * Insert inbound fax into queue with document storage
      *
-     * Uses FaxDocumentService for consistent document handling and patient matching.
+     * Delegates to FaxDocumentService for consistent document handling and patient matching.
      *
      * @param object $faxDetails EtherFax FaxResult object with fax details
      * @return int Queue record ID
      * @throws FaxDocumentException
      */
```
sjpadgett: @juggernautsei Sherwin, completely restore the etherFaxActions class to master version or the version before you touched it. You should not use this service for your service. One issue is you break legacy data using site_ifd in this class. I use accounts and jobId.
This class is specific to etherFax and not a module wide service.

sjpadgett:
@juggernautsei 
- Bump module version
- You can add an upgrade table script to sql directory fn to match upgrade versions. This will auto show an upgrade sql button in module manager.
- I don't see a setup for new service
- Menu for top Service menu
- No big deal but was wondering why tracking site_id in queue table? Site is global to the database on multi site so site id should always be the same for records.
- I've tested this PR for RC and etherFAX okay at this stage.
image

image

image

Let me know if I can help. If this is as far as you want to take this PR then let me know and we'll get merged.

juggernautsei:
image

interface/modules/custom_modules/oe-module-faxsms/table.sql
```diff
@@ -45,33 +45,3 @@ UPDATE categories_seq SET id = (select MAX(id) from categories);
...
-#IfMissingColumn oe_faxsms_queue status
-ALTER TABLE `oe_faxsms_queue` ADD `status` varchar(50) DEFAULT NULL COMMENT 'Fax status (queued, sent, delivered, received, failed, etc)';
-#Endif
-
-#IfMissingColumn oe_faxsms_queue direction
-ALTER TABLE `oe_faxsms_queue` ADD `direction` varchar(20) DEFAULT 'inbound' COMMENT 'inbound or outbound';
-#Endif
-
-#IfMissingColumn oe_faxsms_queue site_id
-ALTER TABLE `oe_faxsms_queue` ADD `site_id` varchar(63) DEFAULT 'default' COMMENT 'Site identifier for multi-site support';
```
sjpadgett: Leave for new install! The upgrade scripts are used so user doesn't have to reinstall to update.

sjpadgett:
@juggernautsei I can't find where you've committed the setup_services.php. I've pulled down your repo with both branches(current and previous PR) so my testing is stalled! Did you or I miss?

juggernautsei:
@sjpadgett I can't push the changes to this repo

sjpadgett:
> I can't push the changes to this repo

I don't understand. You can't push at all or only the setup_services changes?

sjpadgett:
Okay, I'll have to test tomorrow. I wiped tonight. This should get us close to finish line then Mike and I have PR's for some fixes to get in.
Btw Sherwin, thanks for the refactor to queue table.

sjpadgett:
@juggernautsei I enable SignalWire in setup but in top menu Services shows RC. Also the Setup Fax is empty.
In other words SignalWire is broken in setup and menu.
Check the move service id = 9 change

sjpadgett:
@juggernautsei If you want leave SignalWire option = '6' and move voice to '9' or I can do and push. up to you but let me know soon!

sjpadgett:
@juggernautsei We should be squared away now with new service inclusion. I'm testing credentials now.

sjpadgett:
@juggernautsei Have you tested? Will you test your service from init to actions for fax including ensuring credentials stick for both.
Also noted the credential setup titles are wrong but dialog is correct. You know just simple stuff like that.
Once you give me green light I'll merge
image

sjpadgett:
For those that don't know we have a Comm widget for ring central available. Pretty cool and makes available any SMS, Fax or calls from one widget. I had to be talked into this feature but it turns out to be very usable.

image

sjpadgett:
> What is the purpose of having credentials per person? Why?

Because a practice may have users that have their own accounts.

juggernautsei:
> Because a practice may have users who have their own accounts.

That does not compute. But there's no need for any more discussion. I can't test this in production. My production is older than the module is now, and there is no backward compatibility.

juggernautsei:
Is there a way to stop this? Why does this have to list everyone in the address book? These are external referring physicians who won't be using the system. There are 35 referring physicians in the address book. 
image

sjpadgett:
What doesn't compute. Large practice may have separate billing and or split communication accounts. I have folks that use.


> Is there a way to stop this?

I'll add a delimit in query and it doesn't come from address book but users. The whole purpose for feature is so admin can limit user permissions to only use service they would require. This is not just about fax! Admin may hav SMS, Fax and voice where voice is restricted to the reception or certain others but admin doesn't want to let them fax or text.

Why are you just now complaining!

juggernautsei:
> Why are you just now complaining!

Complaining because it is new to me. I have never used this module until now. I have other modules for this. I thought I would try my hand at this module now that I have some more engineering under my belt. You may have people who are using this, but a one-provider practice has no use for this feature, the way it is. I can hear the question why from people who don't have that requirement. I hear questions often about features that don't apply. I have to explain over and over.  I can see having to explain this feature.  
Yes, I know the names come from the users table. I wrote the address book to differentiate between staff and non-staff.

sjpadgett:
Well try reading the several usage documents in directory. If you don't like then I recommend sticking with your module. This module has been expanding feature to meet larger practice needs essentially becoming a communication module. Permissions was becoming a pain with trying to explain how to do with ACL. This was users concerns and why User Permissions is there. Just leave the defaults and should work as if not there.
OpenEMR has a wide audience range with many different communication needs. Take the fax actions to handle documents, should I exclude and treat as simple fax machine? 
You're complaining that you have to answer questions about available features! I'd rather that than questions on why don't you have this feature.

juggernautsei:
Oh, so now you are catering to enterprise??  

LOL. 
It's not that serious. :-) 
Lighten up. I was not complaining. I was just asking questions.
I know OpenEMR has a wide range of clinics that use it. I have talked to my share. It is

sjpadgett:
> Oh, so now you are catering to enterprise??

@juggernautsei First, sorry I didn't mean to be harsh. My communication skills when having to type it out leaves much to be desired. This is why I prefer phone confabs.
Next and as you know, I'm simply trying to provide a generally useful module. It follows the same story of the diverse nature of openemr and why I started down the road of Context Management. To provide workflow that will tailor to work disciplines of user/app setting.
Your concerns are valid and the module has grown to the point that some restructure needs to happen which I've been trying to get at but ran out of time for my goal of v8.0.0.
We talked about refactoring messageUI to pull out the shared panels to twig templates that only supports it's service.
We can also setup the module to run in a context similar to dashboard.

I need to get this merged soon as both Mike and I have some additional changes in general. If you don't mind, I can push my bug fixes here and save a PR. They won't effect SignalWire.
I just need to know you're okay with the new service if I merge. I'll fix the external users so nice catch there.

juggernautsei:
@sjpadgett I am happy just to be able to play in the same sandbox. Do what you feel is best. I am good with it all. I have no agenda. 
I was listening to a speaker the other day who said, " If you are the smartest guy in the room, you are in the wrong room". I am not the smartest guy in the room here. 

One more, Dirty Harry said: "A man gots to know his limitations".

sjpadgett:
image

sjpadgett:
> I am not the smartest guy in the room here.

There's a difference between the smartest or more experienced!

sjpadgett:
@juggernautsei We are going to get some new phpstan errors for your web_hook script. A merge master to this branch should fix then we're done.

sjpadgett:
@juggernautsei Since we recently moved phpstan level to 10 we went from pass to fail here. I refactored webhook to split out the validate functions to a class.
I ensured to keep the methods the same so if worked before should work now but, I rec you pull down changes and test or review and let me know.

sjpadgett:
@stephenwaite Once I got into the etherFax I remembered why the duplicates and that they only happened for same account testing because of same JobId. Now in README:

`
When sending a fax to a number within the same etherFax account (common for testing/loopback), EtherFax may reuse the same JobId for the outbound (Sent) job and the inbound (Received) record. Since OpenEMR enforces unique (account + job_id) and does not allow duplicates, the inbound “Received” update can overwrite the existing Sent queue entry — meaning you may not see a full Sent-history for that test fax once it is received.
`
@juggernautsei I had to fix your upgrade sql and the table sql to fix standard upgrade sql formate. Added new unique key to queue table for account+job_id so verify if it effects your use case. 
Note I also set User Permissions off by default.

sjpadgett:
@juggernautsei Here is a claude eval on the new unique id in queue table and how you'll need to address to prevent database save issues.

Yes, SignalWire Fax Has a Unique Job ID Equivalent
Both SignalWire and etherFAX use unique identifiers for fax jobs, though with slightly different naming conventions:
etherFAX

Identifier Name: JobId
Format: GUID (Globally Unique Identifier)
Example: "833fd712-2ca2-4c7b-8b26-8166629963de"
Usage:

Returned immediately when sending a fax
Used to track fax status via GetFaxStatus(JobId)
Used to retrieve received faxes via GetFax(JobId)
Used to mark faxes as received via SetFaxReceived(JobId)



SignalWire

Identifier Name: sid (Segment ID) or FaxSid
Format: UUID (Universally Unique Identifier)
Example: "b3877c40-da60-4998-90ad-b792e98472fx"
Usage:

Returned as sid in REST API responses
Also called "Fax Segment ID" in the dashboard
Used to retrieve fax details via /Faxes/{FaxSid}
Used to delete faxes via DELETE /Faxes/{FaxSid}
Used to track throughout the fax lifecycle



Key Similarities

Both are UUIDs/GUIDs: Both systems use universally unique identifiers
Globally Unique: Each fax gets a unique identifier that persists throughout its lifecycle
Returned Immediately: Both return the ID when you initiate a fax send
Used for Tracking: Both can be used to query fax status and retrieve details
Used in Logs/Reports: Both appear in dashboard logs and reports for searching

API Response Comparison
etherFAX Response:
json{
  "JobId": "833fd712-2ca2-4c7b-8b26-8166629963de",
  "FaxResult": "InProgress",
  "ConnectTime": 56,
  "PagesReceived": 2
}
SignalWire Response:
json{
  "sid": "b3877c40-da60-4998-90ad-b792e98472fx",
  "status": "queued",
  "from": "+13103383454",
  "to": "+13104456789",
  "direction": "outbound",
  "num_pages": null,
  "date_created": "2018-11-26T20:00:00Z"
}
For OpenEMR Integration
If you're migrating from etherFAX to SignalWire or supporting both, you can map:

etherFAX JobId ↔ SignalWire sid (or FaxSid)
etherFAX FaxResult ↔ SignalWire status

Both identifiers serve the exact same functional purpose and can be stored in your database's fax tracking table in the same column.

sjpadgett:
merging once test pass

interface/modules/custom_modules/oe-module-faxsms/library/webhook_receiver.php
```diff
@@ -11,31 +11,149 @@
...
+use OpenEMR\Common\Crypto\CryptoGen;
+use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Common\Http\oeHttp;
+use OpenEMR\Common\Http\oeHttpRequest;
+use OpenEMR\Modules\FaxSMS\Controller\FaxDocumentService;
+use OpenEMR\Modules\FaxSMS\Utils\SignalWireWebhookValidator;
+
 // Allow webhook access without authentication
 $ignoreAuth = true;
 $sessionAllowWrite = true;
```
bradymiller, line 25: Looks like above 3 lines no longer needed (since globals.php already included above)

sjpadgett: yeah looks like I broke

interface/modules/custom_modules/oe-module-faxsms/src/Utils/SignalWireWebhookValidator.php
```diff
@@ -0,0 +1,145 @@
...
+        // Sanitize to prevent path traversal and injection attacks
+        $sanitized = preg_replace('/[^a-zA-Z0-9_-]/', '', $siteId);
+        return !empty($sanitized) ? $sanitized : 'default';
+    }
+
+    /**
+     * @param string $url
+     * @return bool
+     */
+    public static function isValidSignalWireUrl(string $url): bool
```
bradymiller, line 114: looks like this function is duplicated in 2 places

sjpadgett: I know you are busy Brady but I had a heck of a time getting green and I wasn't ready to merge but got tired of fighting conflicts where I can't merge master to fix because not my branch.  Why wait until I merged?

bradymiller:
Shouldn't be tough to encrypt the stuff (with no patient id) stored in unassigned. Recommend:
1. Encrypting all stuff stored in unassigned (as long as the global to allow this is on)
2. Make a new database row in the queue table that toggles whether is encrypted or not (default to no or null depending on if have been encrypting some stuff already)
3. When store it in unassigned, set this encrypted variable to on if that global to turn off encryption is not set
4. When collect the document from unassigned, then can use the the encrypted setting in queue table to find out if it is encrypted or not (btw, this mechanism makes is backward compatible and supports folks that decide to turn it off) 
5. Do not use the cryptCheckStandard() function on the file. This is only meant for when have no other way to check and there are edge cases where it will fail (if have control much better to have a mechanism to record if it is encrypted or not especially in cases where huge numbers of files will be coming through).

interface/modules/custom_modules/oe-module-faxsms/src/Utils/SignalWireWebhookValidator.php
sjpadgett: @juggernautsei will need to fix because new service won't work anyway because
- API composer dependency that has to be worked out
- Still referring fax documents to patient documents
- save to queue table need refactor

I can't test these things.

bradymiller: agree was definitely in good enough shape to bring into codebase (and it does fix the main vulnerability).

sjpadgett:
No need to mess with encrypting documents as they should be saved to database or flagging if encrypted or not.
Still I well know your concerns with doc handling and storage and I don't disagree but will need to address when module new service is re addressed which I'll get to. I always appreciate  your input but I needed to get my working services with bugs pointed to by others fixed and in.
Also if the cryptCheckStandard() isn't reliable then should be removed. I've been using for years in this or another module

bradymiller:
Sounds good.
cryptCheckStandard() meant for only cases where can't build out tracking mechanism (ie. those old log files). it is simply checking first 3 characters to see if match. So if have high volume use, there is always a chance the actual file may have those characters without actually being encrypted.

PR #10178, opened 2026-01-12
fix(ci): use valid GitHub Actions expression for conditional job

### LINKED ISSUE #10173: bug: recent test all configurations aren't running on master branch
### Describe the bug

hi @kojiromike, wanted to check if [recent test all runs](https://github.com/openemr/openemr/actions/workflows/test-all.yml) should be running on master branch, thank you.

### To Reproduce

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

kojiromike:
Fixes #10173

## Description
PR #10174 attempted to fix the test-all workflow not running on master pushes, but used `skipped()` which is not a valid GitHub Actions function. This caused the workflow to fail with a "workflow file issue" error.

- Replace invalid `if: success() || skipped()` with explicit job result checking
- Use `always()` to ensure the condition is evaluated even when dependencies are skipped

The valid status check functions in GitHub Actions are: `success()`, `failure()`, `always()`, and `cancelled()`. To check if a specific job was skipped, you must use `needs.<job_id>.result == 'skipped'`.

## AI Disclosure
Yes

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -48,7 +48,7 @@ jobs:
   collect:
     runs-on: ubuntu-24.04
     needs: check-skip
-    if: success() || skipped()
+    if: ${{ always() && (needs.check-skip.result == 'success' || needs.check-skip.result == 'skipped') }}
     steps:
     - name: Checkout Code
       uses: actions/checkout@v6


mschoettle:
@kojiromike Randomly saw this. A tool like [`actionlint`](https://github.com/rhysd/actionlint) could be used to lint workflow files that would detect something like this for example.

This is from one of our workflow files where I changed `!cancelled()` to `!skipped()`:

```shell
  .github/workflows/ci.yml:45:14: undefined function "skipped". available functions are "always", "cancelled", "contains", "endswith", "failure", "format", "fromjson", "hashfiles", "join", "startswith", "success", "tojson" [expression]
     |
  45 |         if: '!skipped()'
```

If this is of interest to the OpenEMR project I'd be happy to contribute adding this to CI.

kojiromike:
@mschoettle yes indeed, that'd be very welcome. I took the liberty of creating #10210 from your comment

PR #10203, opened 2026-01-14
fix(session): Set cookie secure flag based on install settings

### LINKED ISSUE #10200: bug(session): undefined global var $_SESSION if web_root is null (pt. portal)
### Describe the bug

Login in patient portal fails following new session overhaul.

```
openemr-1  | [Tue Jan 13 23:26:52.523569 2026] [php:warn] [pid 203:tid 203] [client 192.168.65.1:27386] PHP Warning:  Undefined global variable $_SESSION in /var/www/localhost/htdocs/openemr/src/Common/Session/PHPSessionWrapper.php on line 40, referer: http://localhost:8300/portal/index.php?site=default
```

Looking at the surrounding code, it seems as though the path to `session_start()` only happens if `webroot` is non-null.

I have absolutely no idea why mine is null or how anything functions in this scenario, but here we are. More to come...

### To Reproduce

docker/development-easy on head of master (currently d9158765714d800b087775d3f06768b46963a431)

enable client portal

Visit `http://localhost:8300/portal`

Attempt to log in

### Expected behavior

Log in works, redirected into portal with normal access.

### Client configuration

n/a, backend issue

Firehed:
Fixes #10200

In installs where `https` is not configured (_hopefully_ only `localhost` for development!!), the `setcookie()` call in `SessionUtil::setAppCookie()` would do nothing since it forced `secure=true` and the value would be ignored by any well-behaved user-agent. In a normal deployment that's correctly configured with `https`, it shouldn't have any effect.

Replaces `secure=true` with the value from `session_get_cookie_params()`, mirroring some other cookie handling. This isn't really an ideal source of the value, but it seems reliable, consistent, and doesn't require any of the globals bootstrapping to have occurred yet since it's sourced from `php.ini`.

Even if that cookie erroneously doesn't get the `httponly` flag set on it, it still has all of the necessary flags for general protection _and_ doesn't contain anything sensitive.

> [!NOTE]
> Since it uses `path=/`, I suspect there could be some weird clashing in edge cases where a user tries to use both the main and the patient portal at the same time; the values might collide or fight with each other. I don't _think_ this is a security concern, but I'm not intimately familiar with the session handling.

No

### DIFF
--- a/src/Common/Session/SessionUtil.php
+++ b/src/Common/Session/SessionUtil.php
@@ -313,14 +313,14 @@ class SessionUtil
 
     public static function setAppCookie(string $appType): void
     {
-        ['secure' => $secure] = session_get_cookie_params();
         setcookie(
             self::APP_COOKIE_NAME,
             $appType,
             [
                 'expires' => time() + 3600,
                 'path' => '/',
-                'secure' => $secure,
+                // This permits the app cookie to work in non-https dev environments. It's not a sensitive value.
+                'secure' => false,
                 'httponly' => true,
                 'samesite' => Cookie::SAMESITE_STRICT
             ]


Firehed:
Also cc @zmilan - I can't tag you as a reviewer for whatever reason, but seeing that you authored the session overhaul, you should see this too!

kojiromike (approved):

src/Common/Session/SessionUtil.php
```diff
@@ -313,13 +313,14 @@ private static function standardSessionCookieDestroy(): void
...
     {
+        ['secure' => $secure] = session_get_cookie_params();
         setcookie(
             self::APP_COOKIE_NAME,
             $appType,
             [
                 'expires' => time() + 3600,
                 'path' => '/',
-                'secure' => true,
+                'secure' => $secure,
```
bradymiller: Would just make it false to fall in line with the other openemr session cookies (at some point we will likely make these all true)

src/Common/Session/SessionUtil.php
```diff
@@ -313,13 +313,14 @@ private static function standardSessionCookieDestroy(): void
 
     public static function setAppCookie(string $appType): void
     {
+        ['secure' => $secure] = session_get_cookie_params();
         setcookie(
             self::APP_COOKIE_NAME,
             $appType,
             [
                 'expires' => time() + 3600,
```
bradymiller, line 320: hi @zmilan , what happens to somebody that is logged in for more than 1 hour?

zmilan: @bradymiller I didn't really think carefully about that scenario... I am sorry for that... in general if they are using Portal they will be logged out, but if they are using Core app they would be ok (meaning they will not be logged out because of missing cookie). Decision which session wrapper to use is decided early for benefit to the Core application.
Do we use some standard time for how long our session live?

zmilan:
> Also cc @zmilan - I can't tag you as a reviewer for whatever reason, but seeing that you authored the session overhaul, you should see this too!

I will check it now.

src/Common/Session/SessionUtil.php
zmilan: I am checking now code in `src/Common/Session/SessionWrapperFactory.php` and it seems that it should be changed from current:
```
if ($app !== SessionUtil::PORTAL_SESSION_ID) {
    $session = new PHPSessionWrapper();
} else if (SessionUtil::isPredisSession()) {
    SessionUtil::portalPredisSessionStart();
    $session = new PHPSessionWrapper();
} else {
    $session = new SymfonySessionWrapper(SessionUtil::portalSessionStart());
}
```
To something like this, so it first check for Redis session:
```
if (SessionUtil::isPredisSession()) {
    SessionUtil::portalPredisSessionStart();
    $session = new PHPSessionWrapper();
} else if ($app !== SessionUtil::PORTAL_SESSION_ID) {
    $session = new PHPSessionWrapper();
} else {
    $session = new SymfonySessionWrapper(SessionUtil::portalSessionStart());
}
```

Firehed: I don't even see an explicitly-set value for the session cookies, and my install (development-easy docker) is, somewhat oddly, `28000` seconds, or a bit under 8 hours. I'm assuming it was intended to be `28800` or 8 hours exactly, but that's beside the point. 

They should probably be the same lifetime, regardless of what that lifetime is. Perhaps even `0` so it's a session cookie instead of something with a fixed lifetime? IDK here.

kojiromike: There are these settings in the global config. They should mean something, right? 
image

Firehed: https://github.com/openemr/openemr/issues/10215

bradymiller: Those timeout values in globals should be working (if not, then a bug).
This App cookie is different though and I would rec simply having it never expire. It is not a session token and just there to ensure session machinery knows to use a portal versus core session (it gets switched at login screens and other strategic places. my guess is that this will go away when symonfy sessions are worked into core).

zmilan: I agree with @bradymiller that this cookie is different and they are switched on core/portal login screens, so I think also we can make them very long/never expire. I am not 100% that we can remove them when symfony session is added to core (but that was intention for sure... I will just need to see if that could work)
Should we change this cookie timeout on this branch, or should I make a new one?

zmilan: I have fixed cookie `expires` time and also proper selection of the session wrapper in a new PR: https://github.com/openemr/openemr/pull/10227

PR #10212, opened 2026-01-14
feat(ci): add pre-commit workflow to run actionlint on GitHub Actions workflows

### LINKED ISSUE #10210: implement actionlint to validate changes to github actions
@kojiromike Randomly saw this. A tool like [`actionlint`](https://github.com/rhysd/actionlint) could be used to lint workflow files that would detect something like this for example.

This is from one of our workflow files where I changed `!cancelled()` to `!skipped()`:

```shell
  .github/workflows/ci.yml:45:14: undefined function "skipped". available functions are "always", "cancelled", "contains", "endswith", "failure", "format", "fromjson", "hashfiles", "join", "startswith", "success", "tojson" [expression]
     |
  45 |         if: '!skipped()'
```

If this is of interest to the OpenEMR project I'd be happy to contribute adding this to CI.

_Originally posted by @mschoettle in https://github.com/openemr/openemr/issues/10178#issuecomment-3749780652_

mschoettle:
Fixes #10210 

Add a workflow for actionlint to catch issues in GitHub Actions workflows and run scripts (via shellcheck).

Added as a separate workflow when any workflow file changed. Runs on all workflow files.

No

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -31,4 +31,4 @@ jobs:
 
     - uses: j178/prek-action@v1
       with:
-        extra_args: "--all-files actionlint"
+        extra_args: "--all-files actionlint-docker"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -83,13 +83,7 @@ repos:
 - repo: https://github.com/rhysd/actionlint
   rev: v1.7.10
   hooks:
-  - id: actionlint
-    language: golang
-    additional_dependencies:
-      # actionlint has a shellcheck integration which extracts shell scripts in `run:` steps from GitHub Actions
-      # and checks these with shellcheck.
-      # see also: https://github.com/rhysd/actionlint/pull/482
-      - "github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.11.1"
+  - id: actionlint-docker
 
 - repo: local
   hooks:


.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,39 @@
...
+      uses: actions/checkout@v6
+
+    - name: Add problem matcher
+      run: |
+        echo "::add-matcher::.github/actionlint-matcher.json"
+
+    - name: Run actionlint on all files
+      run: |
+        echo "Linting all Github Action workflow files in the repository..."
+        docker run --rm -v "$(pwd):/repo" --workdir /reop rhysd/actionlint -color
```
kojiromike:
```suggestion
        docker run --rm -v "${PWD}:/repo" --workdir /reop rhysd/actionlint -color
```

Bash always populates the magic variable `PWD` with the current working directory anyway.

.github/actionlint-matcher.json
```diff
@@ -0,0 +1,17 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "actionlint",
+      "pattern": [
+        {
+          "regexp": "^(?:\\x1b\\[\\d+m)?(.+?)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*: (?:\\x1b\\[\\d+m)*(.+?)(?:\\x1b\\[\\d+m)* \\[(.+?)\\]$",
```
kojiromike: This is why we love github. It puts us in a position to write the most beautiful code.

mschoettle:
I addressed all violations I could (see this run: https://github.com/openemr/openemr/actions/runs/21009044957/job/60398995237?pr=10212). The one remaining one is this here: 

```shell
Error: .github/workflows/shellcheck.yml:91:7: shellcheck reported issue in this script: SC2188:warning:50:3: This redirection doesn't have a command. Move to its command (or use 'true' as no-op) [shellcheck]
```

https://github.com/openemr/openemr/blob/daed090bb37a3148f413ef44b643e18804b6b11c/.github/workflows/shellcheck.yml#L141

Not sure what to do here. Is `touch"/tmp/shellcheck-hashes/$hash"` the same?

.github/workflows/test.yml
```diff
@@ -890,4 +890,4 @@ jobs:
           echo "| Status | PHP | Webserver | Database | E2E | Coverage | Docker Dir |"
           echo "| --- | --- | --- | --- | --- | --- | --- |"
           echo "| ${status_emoji} ${{ job.status }} | ${{ steps.parse.outputs.php }} | ${{ steps.parse.outputs.webserver }} | ${{ steps.parse.outputs.database }} ${{ steps.parse.outputs.db }} | ${{ steps.parse.outputs.e2e_enabled }} | ${{ env.ENABLE_COVERAGE }} | ${{ steps.parse.outputs.docker_dir }} |"
-        } >> $GITHUB_STEP_SUMMARY
+        } >> "$GITHUB_STEP_SUMMARY"
```
kojiromike, line 814: is it running shellcheck on the steps? I love that.

kojiromike:
@mschoettle can I help with this? It seems a bit stuck.

.github/workflows/test.yml
mschoettle: Yep it is!

mschoettle:
> @mschoettle can I help with this? It seems a bit stuck.

I was waiting on an answer to my comment above: https://github.com/openemr/openemr/pull/10212#issuecomment-3751944065

I have now replaced it with `touch`, so all checks should now pass.

.github/workflows/pre-commit.yml
kojiromike, line 1: I think there's a rhysd actionlint action we can use to simplify this. It has the problem matcher built-in. See https://github.com/openCoreEMR/oce-module-template/blob/bad9263c01a17faf7267f5535c688b3da7e20c85/.github/workflows/actionlint.yml for example.

mschoettle: I had looked in the Actions marketplace and could not find an official one. So I was surprised that the one you referenced worked. Looking at the logs I noticed that it builds an image from the Dockerfile every time: https://github.com/openCoreEMR/oce-module-template/actions/runs/21229473523/job/61084614927#step:2:2

I think it should be possible to use the [official image](https://hub.docker.com/r/rhysd/actionlint) right away using `container` instead of `runs-on`: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idcontainer

Alternatively, we could use the [`eifinger/actionlint-action`](https://github.com/eifinger/actionlint-action#actionlint-action). I cannot see, however, whether it includes shellcheck.

A third option is to run it via a pre-commit hook directly. That is what we do. It also makes it easy/possible to run the same config locally while developing. We now use [`prek`](https://prek.j178.dev/) which is a drop-in replacement to `pre-commit` (one advantage being that it is much faster).

I can try that out the image later.

kojiromike: @msummers42 will find this thread interesting, especially the bit about prek. Thanks, @mschoettle!

kojiromike: We probably should build a workflow that runs pre-commit.

msummers42: Certainly have been enjoying using `rhysd/actionlint@v1.7.10` in GHA. It does the job. 

Also locally via `docker run --rm -v "$PWD:/workdir" -w /workdir rhysd/actionlint:latest -color -verbose` from a repo root.

One thing I have yet to dig into is using it's shellcheck integration. Could be useful.

mschoettle: The image contains shellcheck so it should work automatically.

For example, this is the `actionlint-docker` hook executed with `prek`:

```console
$ prek run --all-files --verbose actionlint-docker
Lint GitHub Actions workflow files.......................................Failed
- hook id: actionlint-docker
- duration: 2.34s
- exit code: 1

  .github/workflows/isolated-tests.yml:112:7: shellcheck reported issue in this script: SC2086:info:1:74: Double quote to prevent globbing and word splitting [shellcheck]
      |
  112 |       run: |
```

mschoettle: > We probably should build a workflow that runs pre-commit.

I added the `actionlint-docker` hook to the existing pre-commit config and changed the workflow to be for `pre-commit` (via `prek`). I tried the `actionlint` with the additional `shellcheck` dependency but it runs much slower than via the container.

For now, this workflow only executes the `actionlint-docker` hook but it could be extended to more in the future.

kojiromike: There are some bugs in the .pre-commit config. Out of scope for this PR, but I'll document them so we can address them soon.

.github/workflows/isolated-tests.yml
```diff
@@ -51,15 +51,15 @@ jobs:
...
           composer config cache-files-dir
-        } >> $GITHUB_OUTPUT
+        } >> "$GITHUB_OUTPUT"
 
     - name: Composer Cache
       uses: actions/cache/restore@v5
       with:
         path: ${{ steps.composer-cache.outputs.dir }}
-        key: ${{ runner.os }}-composer-${{ matrix.configurations.php }}-${{ hashFiles('**/composer.lock') }}
+        key: ${{ runner.os }}-composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.lock') }}
```
kojiromike, line 60: This was a good catch, but out of scope here. I fixed it in another PR.

bradymiller:

PR #10217, opened 2026-01-14
fix(insurance): rel-704 insurance edit fails if dashboard context mod…

stephenwaite:
…ule enabled

Fixes #


adunsulag:
Bringing this in, Thanks @stephenwaite !

PR #10218, opened 2026-01-14
refactor(ci): Change phpstan baseline files to PHP & integrate per-error-code splitter tool

### LINKED ISSUE #10077: perf(phpstan): migrate baselines to php format
This docs mention php format is faster for huge baselines (and so beneficial for us):

https://phpstan.org/user-guide/baseline#php-baseline-format-instead-of-neon

### LINKED ISSUE #10189: `composer phpstan-baseline` script can generate invalid files
Since yaml isn't _quite_ neon, the `yq`-based tooling can generate some invalid data which can't be read back out in subsequent runs.

Specifically, I was trying to add the official phpstan-deprecation-rules package, and the baseline got addtions like this:

```diff
--- a/.phpstan/phpstan-remaining-baseline.neon
+++ b/.phpstan/phpstan-remaining-baseline.neon
@@ -160,6 +160,12 @@ parameters:
       identifier: variable.undefined
       count: 2
       path: ../ccr/transmitCCD.php
+    - message: |-
+        ' #^Call to deprecated method createFromPath\(\) of class League\\Csv\\Reader\: since version 9\.27\.0
+        Returns a new instance from a file path\.$# '
+      identifier: staticMethod.deprecated
+      count: 1
+      path: ../contrib/util/billing/load_fee_schedule.php
     - message: '#^Variable \$duplicates on left side of \?\? always exists and is not nullable\.$#'
       identifier: nullCoalesce.variable
       count: 1
```

That's valid yaml, but not valid neon.

This might be an argument in favor of #10077, though it's also an argument in favor of simply dropping the post-processing and using the generated file as-is.

CC @kojiromike

Firehed:
Fixes #10189, fixes #10077.

Fixes a problem where sufficiently-long error messages could get mangled due to the subtle differences between yaml and neon formats, which are relevant to the PHPStan baseline generator wrapper.

1) Converts the baseline files to PHP (side-steps the line-length issues from `yq`, fixing the issue)
2) Replaces the custom file-per-code splitting logic with a dedicated tool that does the same thing (maintains and expands the existing logic)
3) Updates the composer.json wrappers to use that tool

As a bonus, generating the baseline now also prints a count of errors grouped per-code, so we can easily have a sense of how much ground we're gaining (or losing) on entire classes of problems. There should be a negligible performance improvement thanks to the format shift; see comments in #10077.

Hat tip to @kojiromike for finding https://github.com/shipmonk-rnd/phpstan-baseline-per-identifier

> [!NOTE]
> I needed to move the baselines to their own directory due to the way the baseline splitter tool works - it seems to delete _all_ `.php` files in the directory when it runs, so it would delete the custom rules.

No

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Mark all PHPStan baseline files as generated
+.phpstan/baseline/ linguist-generated=true


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Mark all PHPStan baseline files as generated
+.phpstan/baseline/ linguist-generated=true
```
Firehed, line 2: This should cause the baseline files to render differently in code review (similar to what composer.lock gets), but GH doesn't seem to do it reliably on the PR that first configures it.

composer.json
```diff
@@ -270,19 +271,10 @@
...
-            "expand tmp-phpstan/tmp.neon | yq -P '.parameters.ignoreErrors |= map(select(.identifier == \"openemr.forbiddenCurlFunction\"))' > .phpstan/phpstan-curl-baseline.neon",
-            "expand tmp-phpstan/tmp.neon | yq -P '.parameters.ignoreErrors |= ((map(select(.identifier == \"openemr.deprecatedSqlFunction\"))) + (map(select(.identifier == \"openemr.deprecatedLaminasDb\"))))' > .phpstan/phpstan-database-baseline.neon",
-            "expand tmp-phpstan/tmp.neon | yq -P '.parameters.ignoreErrors |= map(select(.identifier == \"openemr.forbiddenGlobalsAccess\"))' > .phpstan/phpstan-globals-baseline.neon",
-            "expand tmp-phpstan/tmp.neon | yq -P '.parameters.ignoreErrors |= map(select(.identifier != \"openemr.forbiddenCurlFunction\" and .identifier != \"openemr.deprecatedSqlFunction\" and .identifier != \"openemr.deprecatedLaminasDb\" and .identifier != \"openemr.forbiddenGlobalsAccess\"))' > .phpstan/phpstan-remaining-baseline.neon",
-            "rm tmp-phpstan/tmp.neon"
+            "phpstan analyze --memory-limit=8G --configuration=phpstan.neon.dist --generate-baseline=.phpstan/baseline/loader.php",
```
kojiromike, line 276: Will phpstan completely regenerate the baseline, even if one already exists?

Firehed: Yes, that's the intended usage of the flag.

PR #10227, opened 2026-01-15
fix(session): increase app cookie expire time and better wrapper selection logic

### LINKED ISSUE #10215: Standardize session timeout mechanics
There are some security settings in the globals-based config. The new portal session settings don't seem to use them. I'm not saying we must use them, but if we don't, the session lifetime should be well-defined and respected by the codepaths that handle them.

---

There are these settings in the global config. They should mean something, right? 
image

_Originally posted by @kojiromike in https://github.com/openemr/openemr/pull/10203#discussion_r2691631421_

zmilan:
Fixes #10215

While switching to Symfony Session in the Portal application, two potential bugs are introduced:
1. The `expires` time for the App type cookie is set to 1h. We agree to make it without expiration, since it is set/updated on the login page of both core and portal applications.
2. Logic for choosing the proper session wrapper is wrong, and it first needs to check for the Redis session before it chooses between the core or portal one.

1. Set `expires` to be 1 year
2. First check for the Redis session, then for core, and last one should be portal.

### DIFF
--- a/src/Common/Session/SessionWrapperFactory.php
+++ b/src/Common/Session/SessionWrapperFactory.php
@@ -36,10 +36,10 @@ class SessionWrapperFactory
     private function findSessionWrapper(array $initData = []): SessionWrapperInterface
     {
         $app = SessionUtil::getAppCookie();
-        if ($app !== SessionUtil::PORTAL_SESSION_ID) {
-            $session = new PHPSessionWrapper();
-        } else if (SessionUtil::isPredisSession()) {
+        if (SessionUtil::isPredisSession()) {
             SessionUtil::portalPredisSessionStart();
+            $session = new PHPSessionWrapper();
+        } else if ($app !== SessionUtil::PORTAL_SESSION_ID) {
             $session = new PHPSessionWrapper();
         } else {
             $session = new SymfonySessionWrapper(SessionUtil::portalSessionStart());
--- a/src/Common/Session/SessionUtil.php
+++ b/src/Common/Session/SessionUtil.php
@@ -317,7 +317,7 @@ class SessionUtil
             self::APP_COOKIE_NAME,
             $appType,
             [
-                'expires' => time() + 3600,
+                'expires' => time() + 31536000, // 1 year
                 'path' => '/',
                 'secure' => true,
                 'httponly' => true,


Firehed:
Forgive my ignorance here on some of the mechanics - if I'm following correctly, this seems to have logic that's handling both the storage mechanism of the session (php-native e.g. files vs predis) and which session cookie gets read based on which page you're on. Is that intended?

It seems to me we'd want the "is predis" path to determine which storage mechanism to use on the backend, and the app cookie to determine which session identifier is used.

(I'm also not seeing anything that actually interacts with `predis`, but maybe that's intentional for now?)

Also, minor note, #10203 is a separate PR - this seems to target some sub-issues discussed in its changes. Perhaps this fixes #10215, one of the spin-offs?

zmilan:
> Forgive my ignorance here on some of the mechanics - if I'm following correctly, this seems to have logic that's handling both the storage mechanism of the session (php-native e.g. files vs predis) and which session cookie gets read based on which page you're on. Is that intended?
> 
> It seems to me we'd want the "is predis" path to determine which storage mechanism to use on the backend, and the app cookie to determine which session identifier is used.
> 
> (I'm also not seeing anything that actually interacts with `predis`, but maybe that's intentional for now?)
> 
> Also, minor note, #10203 is a separate PR - this seems to target some sub-issues discussed in its changes. Perhaps this fixes #10215, one of the spin-offs?

So, `PHPSessionWrapper` is just a simple wrapper around `$_SESSION`, to be able to use the same code in shared files between the core and the portal application. For core (and Redis, since it is set to use `$_SESSION` too), it returns `PHPSessionWrapper`  and actually works with `$_SESSION`, so it doesn't interrupt other parts of the application. For the portal, it uses `SymfonySessionWrapper`, which is again a simple wrapper around Symfony Session, so it can have a few more functions to be able to handle some specific things inside files shared between applications.  And switching between the core and the portal is detected based on the `App` cookie, which is set on each application login page.

So, when it comes to the Redis session, it do same session initialization as before my changes, it just uses this wrapper in shared files.

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -36,11 +36,11 @@ public function getWrapper(array $initData = []): SessionWrapperInterface
     private function findSessionWrapper(array $initData = []): SessionWrapperInterface
     {
         $app = SessionUtil::getAppCookie();
-        if ($app !== SessionUtil::PORTAL_SESSION_ID) {
-            $session = new PHPSessionWrapper();
-        } else if (SessionUtil::isPredisSession()) {
+        if (SessionUtil::isPredisSession()) {
```
bradymiller, line 39: hi @zmilan , Will predis session work for both the portal and core sessions?

zmilan: I didn't test it locally, but since predis is set up to use `$_SESSION`  and `PHPSessionWrapper` is just a few functions around it, I don't see why it would not work. And in that case, there is no difference between core and portal, they will both fail to `PHPSessionWrapper` since we first check for that (with this last change).

PR #10234, opened 2026-01-15
feat: Fixes #10233 cqm reporting period and deps updates

### LINKED ISSUE #10233: feat: Update ecqm reporting period
### Describe the problem

We need to update the reporting period for the ecqm library.  The default is set to 2022 in the demo database and we should at a minimum set it to 2025 until the 2026 measures are released.  

As part of this work we should update some of the cqm dependencies and log an error if the reporting file is not found.  Currently it throws a fatal exception in the care coordination module if the reporting period files are not installed.

adunsulag:
Update a number of dependencies in the cqm, ccda, and schematron pieces to bring them up to date.

Log an error if a cqm reporting file is not found.

Update the default reporting period to be at least 2025 instead of 2022 which no longer even exists in the project files.

### DIFF
--- a/ccdaservice/packages/oe-cda-schematron/package.json
+++ b/ccdaservice/packages/oe-cda-schematron/package.json
@@ -21,7 +21,7 @@
     "grunt": "^1.0.1",
     "grunt-contrib-jshint": "^3.2.0",
     "grunt-mocha-test": "^0.13.2",
-    "mocha": "^11.7.0"
+    "mocha": "^11.7.5"
   },
   "dependencies": {
     "@xmldom/xmldom": "0.9.8",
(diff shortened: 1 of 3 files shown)


ccdaservice/packages/oe-cda-schematron/package.json
```diff
@@ -21,7 +21,7 @@
     "grunt": "^1.0.1",
     "grunt-contrib-jshint": "^3.2.0",
     "grunt-mocha-test": "^0.13.2",
-    "mocha": "^11.7.5"
+    "mocha": "^11.7.0"
```
sjpadgett: Is this reverting

adunsulag: Thanks for catching that.

adunsulag:
Going to bring this in as we need to verify it against another system check.

PR #10237, opened 2026-01-16
chore(phpstan): bump to level 10, pin to PHP 8.2

### LINKED ISSUE #10236: chore(phpstan): Bump to level 10 with baselined errors
## Summary

Bump PHPStan analysis from level 1 to level 10 (the strictest level) and baseline all existing errors. This prevents new type-safety issues from being introduced while allowing incremental cleanup of legacy code.

## Changes

- **PHPStan level**: 1 → 10
- **Baselined errors**: ~159,600 across 147 categorized baseline files
- **Baseline organization**: Errors are split by error type (e.g., `argument.type.php`, `return.type.php`, `variable.undefined.php`) for easier tracking and incremental fixes

## Benefits

1. **Prevents regressions**: New code must pass level 10 analysis
2. **Incremental improvement**: Baseline files can be reduced over time
3. **Categorized errors**: Easy to identify and prioritize which error types to fix first
4. **CI enforcement**: Pre-commit hooks and CI will catch new violations

## Baseline Categories (largest)

| Category | Errors | Description |
|----------|--------|-------------|
| `missingType.parameter` | ~10,600 | Function parameters without type hints |
| `variable.undefined` | ~8,300 | Potentially undefined variables |
| `missingType.return` | ~7,400 | Functions without return type hints |
| `argument.type` | ~6,200 | Type mismatches in function arguments |
| `missingType.property` | ~4,800 | Class properties without type hints |

## Follow-up Work

Individual baseline files can be tackled incrementally. Priority should be given to:
1. Security-sensitive code paths
2. Core services and APIs
3. Frequently modified files

### LINKED ISSUE #10238: ci(phpstan): baseline is not being uploaded
When switching from using .neon to .php baselines, we inadvertently didn't update the github action, so it isn't uploading the baselines anymore.

kojiromike:
Fixes #10236
Fixes #10238

## Summary

Bump PHPStan analysis from level 1 to level 10 (the strictest level) and baseline all existing errors to prevent new type-safety issues from being introduced. Pin analysis to minimum supported PHP version (8.2) for consistent results across environments.

- **PHPStan level**: 1 → 10
- **PHP version**: Pin to 8.2 (minimum from composer.json) in both config and CI workflow
- **Baselined errors**: ~159,600 across 147 categorized baseline files  
- **Pre-commit config**: Added exclusion for `.phpstan/baseline/` (some files exceed 500KB)
- **Baseline organization**: Errors are split by error type (e.g., `argument.type.php`, `return.type.php`) for easier tracking

## Test plan

- [ ] PHPStan passes with `composer phpstan`
- [ ] Pre-commit hooks pass
- [ ] CI checks pass

---

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -27,11 +27,14 @@ jobs:
         # issues affecting all supported versions.
         # Must match phpVersion in phpstan.neon.dist for consistent results.
         php-version: ['8.2']
-        # Test both architectures to investigate baseline discrepancies
-        runner: [ubuntu-24.04, ubuntu-24.04-arm]
+        # Test architectures to investigate baseline discrepancies
+        # - ubuntu-24.04: x64 Linux
+        # - ubuntu-24.04-arm: ARM Linux (Ampere Altra server chips)
+        # - macos-14: Apple Silicon M1 (matches local Mac dev environment)
+        runner: [ubuntu-24.04, ubuntu-24.04-arm, macos-14]
         # Test with and without parallelism to check if it affects results
         parallel: [true, false]
-    name: PHP ${{ matrix.php-version }} (${{ matrix.runner == 'ubuntu-24.04-arm' && 'ARM' || 'x64' }}, ${{ matrix.parallel && 'parallel' || 'single-threaded' }})
+    name: PHP ${{ matrix.php-version }} (${{ matrix.runner == 'macos-14' && 'macOS M1' || (matrix.runner == 'ubuntu-24.04-arm' && 'Linux ARM' || 'Linux x64') }}, ${{ matrix.parallel && 'parallel' || 'single-threaded' }})
     runs-on: ${{ matrix.runner }}
     steps:
     - name: Checkout Code
@@ -46,8 +49,12 @@ jobs:
         mkdir -p .phpstan
         php -i > .phpstan/php-config.txt
         php -m > .phpstan/php-modules.txt
-        ldd --version 2>&1 | head -1 > .phpstan/glibc-version.txt || true
-        uname -m > .phpstan/arch.txt
+        if command -v ldd &> /dev/null; then
+          ldd --version 2>&1 | head -1 > .phpstan/glibc-version.txt
+        else
+          echo "N/A (macOS)" > .phpstan/glibc-version.txt
+        fi
+        uname -a > .phpstan/arch.txt
     - name: Upload PHP Configuration
       uses: actions/upload-artifact@v6
       with:


kojiromike:
@Firehed you should see this…

phpstan is failing because it does analysis differently in github than locally. This seems to result in it either not finding particular names, or, when a function is defined with the same name in multiple places (especially with different signatures), deciding that the wrong signature is in play even when the runtime is fine.

If this is passing by the time you see it, it may be because I copied the baseline that the github action produced. That may pass, but it seems like a very rickety fix, at best. Do you have any suggestions?

Firehed:
Yes, fix this:

> function is defined with the same name in multiple places

Half-joking, half-serious. Any code like that is one weird `include` change away from being a fatal error at runtime. I'll see if I can AI a rule to find them; I know of a couple within payments, but this is obviously easier said than done w/ the ad-hoc autoloading.

IMO, anything else like trying to force a specific file processing order is just a lousy workaround.

sjpadgett:
Okay for new code but scares me for global! A less impactful ramp up over next couple releases is more appropriate imo.

kojiromike:
@sjpadgett this is for new code, and the baseline is there to prevent it impacting old code…

sjpadgett:
Works for me, thanks

Firehed:
When I check out 9ccf93c3cff1a97207216996b2f56f6e40cd6da1 locally I also get no errors, which I assume is the case on your machine as well. Something is indeed weird or subtly different in CI.

If I merge `upstream/master` into the branch (as `actions/checkout` does by default) I do get a few errors popping, but not the 75-ish being reported in the failing CI step.

```
$ composer phpstan
 3994/3994 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   tests/Tests/E2e/Base/BaseTrait.php (in context of class OpenEMR\Tests\E2e\BbCreateStaffTest)
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  85     Ignored error pattern #^Cannot call method until\(\) on mixed\.$# in path /Users/firehed/dev/github/openemr/smith-oemr/tests/Tests/E2e/BbCreateStaffTest.php is expected to occur 9 times, but occurred 10 times.
  85     Ignored error pattern #^Cannot call method wait\(\) on mixed\.$# in path /Users/firehed/dev/github/openemr/smith-oemr/tests/Tests/E2e/BbCreateStaffTest.php is expected to occur 12 times, but occurred 13 times.
  126    Ignored error pattern #^Cannot call method waitFor\(\) on mixed\.$# in path /Users/firehed/dev/github/openemr/smith-oemr/tests/Tests/E2e/BbCreateStaffTest.php is expected to occur 24 times, but occurred 25 times.
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------------
  Line   tests/Tests/E2e/User/UserAddTrait.php (in context of class OpenEMR\Tests\E2e\BbCreateStaffTest)
 ------ -------------------------------------------------------------------------------------------------
  101    Cannot call method until() on mixed.
         🪪  method.nonObject
  158    Cannot call method waitFor() on mixed.
         🪪  method.nonObject
  185    Cannot call method wait() on mixed.
         🪪  method.nonObject
 ------ -------------------------------------------------------------------------------------------------


 [ERROR] Found 6 errors


Script phpstan analyze --memory-limit=4G --configuration=phpstan.neon.dist handling the phpstan event returned with error code 1
```

Since I'm running PHP 8.4 locally, I tried the same thing inside `php:8.2-alpine-cli` and saw the same results.

Firehed:
I have a completely unfounded hunch that there's something weird happening with one of the cache steps that's polluting CI and resulting in the errors. I can't push to your branch since it's on your own fork, but maybe try commenting them out? It could also be worth dumping a `git status`, `git diff` and a little bit of `git log` to the job output as well just in case there's something spooky happening that has nothing at all to do with this change.

Firehed:
Low but nonzero chance https://github.com/openemr/openemr/pull/10289 is the culprit here (probably combined with aforementioned caching)

Firehed:
I did some additional testing over at https://github.com/Firehed/openemr/pull/1 (to avoid polluting the main repo)

So far:

- not the result cache
- not the stray neon file
- not a randomly different version of phpstan being installed
- not uninstalling rector

The delta seems _stable_ at a glance, so that's something.

My next thing to try tomorrow is bisecting my way to the phpstan level bump that causes a problem, under the vauge assumption that's at least part of the problem. If it's a contributing factor, I think we can use that to bump to where we can ASAP and solve the delta later. Can't imagine why it would be, but this issue isn't making much sense to me in the first place.

kojiromike:
@Firehed I don't think the level bump is the cause. I have seen this at level 1.

kojiromike:
## PHPStan Baseline Investigation

I've been investigating (using Claude) why PHPStan produces different results locally vs in CI, causing baseline mismatches.

### Experiments Conducted

#### Experiment 1: PHP Version Mismatch
**Hypothesis**: Different PHP runtime versions cause different results.
**Result**: Confirmed. Local PHP 8.5.2 produced 70 errors, CI PHP 8.4 produced 89 errors.
**Action**: Pinned `phpVersion: 80200` in `phpstan.neon.dist`.

#### Experiment 2: phpVersion Configuration
**Hypothesis**: Setting `phpVersion` in config makes results consistent regardless of runtime.
**Result**: Partially confirmed. Helped but didn't fully resolve discrepancies.

#### Experiment 3: Baseline Generation After PHP Version Pinning
**Hypothesis**: After pinning PHP version, baselines should match.
**Result**: CI still failed - baseline had errors CI didn't find, indicating local found MORE errors.

#### Experiment 4: Docker with PHP 8.2 Runtime (Debian)
**Hypothesis**: Same PHP version in Docker should match CI.
**Result**: Still different. Both PHP 8.2.30, but:
- `echo.nonString.php`: CI 3351, Docker 3309 (+42 in CI)
- `argument.type.php`: CI 25693, Docker 25748 (+55 in Docker)

#### Experiment 5: Ubuntu 24.04 Docker (matching CI OS)
**Hypothesis**: OS differences (Debian vs Ubuntu) cause discrepancies.
**Result**: Built Ubuntu 24.04 Docker with ondrej/php PPA. Same glibc (2.39), same OS. ARM Docker produced 0 errors against local baseline.

#### Experiment 6: PHP Extension Comparison
**Hypothesis**: Different extensions affect type inference.
**Result**: CI has 82 PHP modules, Ubuntu Docker has 57. Missing extensions are mostly DB drivers and caching - unlikely to affect the specific files showing discrepancies.

#### Experiment 7: Architecture Differences (KEY FINDING)
**Hypothesis**: x86_64 vs ARM architecture affects PHPStan analysis.
**Result**: CONFIRMED.

| Environment | Architecture | Errors vs Local Baseline |
|-------------|--------------|--------------------------|
| Local (Apple Silicon) | ARM (aarch64) | 0 |
| Docker (ARM) | ARM (aarch64) | 0 |
| Docker (x86_64 emulated) | x86_64 | 7 errors |
| GitHub CI | x86_64 | ~7 errors |

### Total Error Count
- CI baseline: 159,599 errors
- Local baseline: 159,598 errors
- **Net difference: 1 error**

Errors are distributed differently across files, causing baseline mismatches even though totals are nearly identical.

### Files Most Affected
- `interface/orders/order_manifest.php` (CI: 24, Local: 4)
- `interface/orders/single_order_results.inc.php` (CI: 37, Local: 16)
- `interface/orders/list_reports.php` (CI: 8, Local: 7)

### Current Experiment
This commit adds an ARM runner (`ubuntu-24.04-arm`) alongside the existing x86_64 runner to reproduce the experiment in CI:

**If architecture is the root cause, the ARM runner should PASS (0 errors) while x86_64 fails.**

GitHub provides free ARM runners for public repositories (Public Preview since Jan 2025).

### Investigation Details
Full investigation documented in `.phpstan/baseline-investigation.md`

kojiromike:
## PHPStan Baseline Investigation (Updated)

I've been investigating (using Claude) why PHPStan produces different results locally vs in CI, causing baseline mismatches.

### Experiments Conducted

#### Experiment 1: PHP Version Mismatch
**Hypothesis**: Different PHP runtime versions cause different results.
**Result**: Confirmed. Local PHP 8.5.2 produced 70 errors, CI PHP 8.4 produced 89 errors.
**Action**: Pinned `phpVersion: 80200` in `phpstan.neon.dist`.

#### Experiment 2: phpVersion Configuration
**Hypothesis**: Setting `phpVersion` in config makes results consistent regardless of runtime.
**Result**: Partially confirmed. Helped but didn't fully resolve discrepancies.

#### Experiment 3: Baseline Generation After PHP Version Pinning
**Hypothesis**: After pinning PHP version, baselines should match.
**Result**: CI still failed - baseline had errors CI didn't find, indicating local found MORE errors.

#### Experiment 4: Docker with PHP 8.2 Runtime (Debian)
**Hypothesis**: Same PHP version in Docker should match CI.
**Result**: Still different. Both PHP 8.2.30, but:
- `echo.nonString.php`: CI 3351, Docker 3309 (+42 in CI)
- `argument.type.php`: CI 25693, Docker 25748 (+55 in Docker)

#### Experiment 5: Ubuntu 24.04 Docker (matching CI OS)
**Hypothesis**: OS differences (Debian vs Ubuntu) cause discrepancies.
**Result**: Inconclusive. Built Ubuntu 24.04 Docker with ondrej/php PPA (same glibc 2.39, same OS as CI). Baseline passed, but since this ran on ARM (Apple Silicon Mac), couldn't determine if it passed due to OS match or architecture match. Variable not properly isolated.

#### Experiment 6: PHP Extension Comparison
**Hypothesis**: Different extensions affect type inference.
**Skipped**: CI has 82 PHP modules, Ubuntu Docker has 57. Missing extensions are mostly DB drivers and caching - unlikely to affect the specific files showing discrepancies. Did not test further.

#### Experiment 7: Architecture Differences (KEY FINDING)
**Hypothesis**: x86_64 vs ARM architecture affects PHPStan analysis.
**Method**: Built Ubuntu Docker with `--platform linux/amd64` to force x86_64 emulation on ARM Mac.
**Result**: CONFIRMED.

| Environment | Architecture | Baseline Result |
|-------------|--------------|-----------------|
| Local (Apple Silicon) | ARM (aarch64) | Passed |
| Docker (ARM, native) | ARM (aarch64) | Passed |
| Docker (x86_64, emulated) | x86_64 | Failed (7 errors) |
| GitHub CI | x86_64 | Failed (~7 errors) |

### Total Error Count
- CI baseline: 159,599 errors
- Local baseline: 159,598 errors
- **Net difference: 1 error**

Errors are distributed differently across files, causing baseline mismatches even though totals are nearly identical.

### Files Most Affected
- `interface/orders/order_manifest.php` (CI: 24, Local: 4)
- `interface/orders/single_order_results.inc.php` (CI: 37, Local: 16)
- `interface/orders/list_reports.php` (CI: 8, Local: 7)

### Current Experiment
This commit adds an ARM runner (`ubuntu-24.04-arm`) alongside the existing x86_64 runner to reproduce the experiment in CI:

**If architecture is the root cause, the ARM runner should PASS while x86_64 fails.**

GitHub provides free ARM runners for public repositories (Public Preview since Jan 2025).

### Investigation Details
Full investigation documented in `.phpstan/baseline-investigation.md`

kojiromike:
## Experiment Results: Hypothesis DISPROVED

The CI run completed with both ARM and x86_64 runners. Results:

| Environment | Architecture | Total Errors |
|-------------|--------------|--------------|
| CI ARM | aarch64 | 159,582 |
| Local ARM (Mac) | aarch64 | 159,598 |
| CI x86_64 | x86_64 | 159,599 |

### Key Finding
**My hypothesis was wrong.** CI ARM finds **16 fewer errors** than local ARM, not the same.

All three environments produce different results:
- CI ARM: 159,582 (fewest)
- Local ARM: 159,598 
- CI x86_64: 159,599 (most)

### What This Means
Architecture alone doesn't explain the discrepancy. The difference is likely caused by:
1. `shivammathur/setup-php` configures PHP differently than Docker
2. Different PHP compilation flags between environments
3. Different system libraries (even with same glibc version)
4. Possibly different parallelism or file ordering behavior

### Next Steps
The investigation continues. Possible approaches:
1. Compare detailed PHP configs between CI and local Docker
2. Try using the exact same PHP binary from shivammathur in Docker
3. Accept that baselines must be generated in CI and downloaded

Added `fail-fast: false` to let both jobs complete independently for cleaner comparisons.

kojiromike:
## PHPStan Baseline Investigation Update

I (using Claude) have been investigating why PHPStan produces different results locally vs CI. Here's a summary of all experiments:

### Completed Experiments

| # | Hypothesis | Result |
|---|------------|--------|
| 1 | PHP version mismatch | ✅ Confirmed - PHP version affects results |
| 2 | phpVersion config | ⚠️ Partial - config helps but insufficient alone |
| 3 | Baseline after version pinning | ❌ Still mismatched |
| 4 | Docker PHP 8.2 | ❌ Same PHP version, different results |
| 5 | Ubuntu 24.04 Docker | ⚠️ Inconclusive - OS match didn't isolate variable |
| 6 | PHP extensions | ⏭️ Skipped - unlikely to affect specific files |
| 7 | Architecture (local Docker) | ⚠️ Preliminary - x86_64 vs ARM showed differences |
| 8 | GitHub ARM runner | ❌ **DISPROVED** - all 3 environments differ |
| 9 | Parallelism | ❌ **DISPROVED on x64** - parallel/single-threaded identical (159,599) |
| 10 | Rector removal | ❌ **DISPROVED** - no effect after clearing cache |

### Experiment 9 Partial Results (Parallelism)

| Configuration | Errors |
|---------------|--------|
| Linux x64, parallel | 159,599 |
| Linux x64, single-threaded | 159,599 |
| Linux ARM (Ampere), parallel | 159,582 |
| Linux ARM (Ampere), single-threaded | *pending* |

**Key finding**: Parallelism does NOT affect results on x64.

### Experiment 11: macOS M1 (IN PROGRESS)

Added `macos-14` runner to test Apple Silicon specifically. GitHub's Linux ARM runners use Ampere Altra (server ARM), not Apple Silicon - this could explain why CI ARM results differ from local Mac ARM results.

New CI matrix (6 jobs):
- Linux x64 (parallel, single-threaded)
- Linux ARM/Ampere (parallel, single-threaded)
- macOS M1/Apple Silicon (parallel, single-threaded)

### Current Theory

The discrepancy is likely caused by differences in the PHP build from `shivammathur/setup-php` vs local/Docker PHP installations, not architecture or parallelism per se.

Firehed:
> @Firehed I don't think the level bump is the cause. I have seen this at level 1.

What I've seen so far (trying level 5, but not fully bisecting) doesn't make me think it's the case either, nor does it make any sort of intrinsic sense. 

_However_, after something of a wild goose chase due to rector, I was able to more-or-less confirm one of our earlier suspicions:
```
 ------ ----------------------------------------------------------------------- 
  Line   interface/orders/order_manifest.php                                    
 ------ ----------------------------------------------------------------------- 
         Ignored error pattern #^Parameter \#1 \$s of function myCellText       
         expects string, int given\.$# in path                                  
         /home/runner/work/openemr/openemr/interface/orders/order_manifest.php  
         was not matched in reported errors.                                    
 ------ -----------------------------------------------------------------------
```

There are three definitions (of course) of `myCellText`, _one_ of which has docblock-provided type information; the other two are mixed all the way through. Analyzing just the one file results in that line not having a problem, but in the context of the whole project it gets tagged again (well, sort of the inverse with the baseline).

I don't know the exact internal mechanics of file discovery in PHPStan's analysis path (it uses Symfony's Finder component), but this _strongly_ implies that duplicate, conflicting function definitions are the cause of the problem. The higher PHPStan level yields much more type info which can propagate throughout analysis. My educated guess is that it's something goofy with how GHA tends to mangle file timestamps by default causing different results in CI vs locally.

I attempted to use [this action](https://github.com/marketplace/actions/git-restore-mtime) to restore the timestamps, but no dice. Filesystems (🛎️) are weird and there's no explicit sort order that I can see within PHPStan, so it was a long shot.

Removing the conflicting definitions and moving the one that seemed to best align with the baseline (the one with types) into the autoload path **fixed the issue**.

tl;dr: I'm almost positive that this is all fallout from duplicate, conflicting function definitions, and that they get analyzed in a different order in CI.

kojiromike:
@Firehed  I'm starting to wonder if there isn't just more than one phpstan issue at play. Or, to put it another way, if the problem is indeed functions with the same name, that isn't by itself a satisfactory answer, because even with duplicate function names, if phpstan ran consistently on different systems, it'd yield the same results.

So, could it be duplication function names, triggered by different run ordering on…
1. different architectures?
2. different parallelizations?
3. something completely stochastic? (IDK, this one would mean if you run it twice on your machine you could get different results depending on … something)

kojiromike:
I think I get what you're saying about timestamps. I'll see if I can build an experiment around it.

Firehed:
Yeah, I think because the file finder isn't specifying an ordering scheme, there's some system-specific thing at play. So it's probably consistent across (some spooky filesystem thing I don't want to dig in to) that's stable _on_ a machine but not stable _across_ machines. Could be timestamps, CPU architecture, byte order, lunar phase, meme dankness index, or anything else.

But I'm willing to bet (a small amount) that if we fix the dupes and block more from occurring, it'll be stable again.

PR #10244, opened 2026-01-16
feat(session): porting core and portal apps to HttpSessionFactory

### LINKED ISSUE #10259: Replace Core application current PHP native session implementation with Symfony Session
As part of the OpenEMR session handling improvement, documented here https://github.com/openemr/openemr/issues/2556, we wish to replace the current session handling, using PHP native session, with the Symfony Session implementation.

zmilan:
Fixes #10259, fixes #11168

Migrate from native PHP sessions to Symfony Session

  Replace direct $_SESSION and session_*() usage across the entire codebase
  with Symfony's Session component, routed through SessionUtil.

  Key changes:
  - Introduce SessionUtil as the central API for all session operations
  - Support `read_and_close` mode for read-only pages (including portal)
  - Add reopen monitoring and auto-disable `read_and_close` for Redis
  - Add a PHPStan rule to detect direct session writes bypassing SessionUtil
  - Remove dead switchToOAuthSession code
  - Fix session destroy handling for `read_and_close` sessions
  - Prevent lock leaks on multiple `read_and_close` writes
  - Update baselines and fix static analysis / semgrep findings

Most changes are related to replacing `$_SESSION` with `$session` syntax. Changes that are not related to the session are introduced to fix issues reported by PHPStan and Semgrep OSS (there were a lot of them since the PR touched a lot of files)

### DIFF
--- a/src/RestControllers/Authorization/LocalApiAuthorizationController.php
+++ b/src/RestControllers/Authorization/LocalApiAuthorizationController.php
@@ -25,7 +25,7 @@ class LocalApiAuthorizationController implements IAuthorizationStrategy
     public function shouldProcessRequest(Request $request): bool
     {
         if ($request->headers->has("APICSRFTOKEN")) {
-            $request->attributes->set("_is_local_api", true);
+            $request->attributes->set("is_local_api", true);
             // this is a local api request, so we should process it
             return true;
         }


zmilan:
I made good progress with this feature, and most of the pages work, as far as I can test them for now. `SessionWrapperFactory` is refactored to work with `HttpSessionFactory` to create SymfonySessions for both the portal and the core app. I also fixed the issue with most/all tests that were dealing with the session. This also means that we do not use/have any `$_SESSION` use in the code (except if I miss something for now, or they came through rebasing and I didn't spot them yet). And that brings us to the challenge I would like to tackle now and test. 

**The challenge:**
In case of API/OAuth authentication, what should `SessionWrapperFactory` return in files shared among different parts of the application? Before these changes, it was returning a  `PHPSessionWrapper` instance that was just a wrapper around using `$_SESSION` directly. Since we get rid of `$_SESSION` fully, we should return a proper instance of SymfonySession created with the same parameters as we do in API/OAuth authentication. At least that is my understanding at the moment.
To phrase it differently. How can I detect that a request is coming from an API, and how to properly create a SymfonySession for such a request? 

@adunsulag, I noticed that you were involved in a discussion about the API part in my portal session work. Could you help with this?

@kojiromike, do you know where I can look for some example/code that could help me with this?

kojiromike:
@zmilan - I think this might be simpler than detecting API context.

For API/OAuth requests, `SiteSetupListener` already creates the session before shared code runs. So the question isn't "is this an API request?" but "is there already an active session?"

When `session_status() === PHP_SESSION_ACTIVE`, you can bridge to it without knowing the session type:

```php
if (session_status() === PHP_SESSION_ACTIVE) {
    $sessionFactory->setUseExistingSessionBridge(true);
    return $sessionFactory->createSession();  // bridges to whatever's running
}
```

See `SiteSetupListener` for a real-world example of this pattern:
https://github.com/openemr/openemr/blob/d3d4a80/src/RestControllers/Subscriber/SiteSetupListener.php#L165-L175

Is there a case where you'd need to distinguish API from web app when a session is already active, or am I missing the point?

zmilan:
@kojiromike, I see, but what I was thinking of (I could be wrong) is this. Until now, we had `$_SESSION` in code directly or through `PHPSessionWrapper`. Now we do not have it anywhere, and I refactored `SessionWrapperFactory` to return Symfony's `SessionInterface` instance based on which application is in use (core and portal for now). I create them using `HttpSessionFactory`, and it all works well for core and portal at the moment. But currently, `SessionWrapperFactory` doesn't know how to return a session for the API. 
`HttpSessionFactory` has types for both API and OAuth (`SESSION_TYPE_API` and `SESSION_TYPE_OAUTH`), and I can instantiate them and return as active session in these shared files if requests comes that require them... but at the moment, I do not know how to instantiate them properly and also when I need to return them instead of core session for example.
I don't know if this makes sense at all, or if I am thinking with wrong assumptions. But, for example, most files currently use code like this to get session `$session = SessionWrapperFactory::getInstance()->getActiveSession();`. This `getActiveSession()` currently return already instantiated session or creates a new one based on what it understands who is making the request (core or portal).

zmilan:
@kojiromike, Do you think that I can use this kind of check before I create a new instance for core and portal to actually return a proper session that is already started?

```
if (session_status() === PHP_SESSION_ACTIVE) {
    $sessionFactory->setUseExistingSessionBridge(true);
    return $sessionFactory->createSession();  // bridges to whatever's running
}
```

That will solve part of my question. The second part is some parameters that I need for creating a session with `HttpSessionFactory`. For example, creating a proper `HttpRestRequest` instance and getting a proper `$webroot`.

kojiromike:
Rather than having `SessionWrapperFactory` detect and re-create API sessions, can the code that *already knows* the session type register it:

```php
// SessionWrapperFactory
public function setActiveSession(SessionInterface $session): void
{
    $this->activeSession = $session;
}

public function getActiveSession(): SessionInterface
{
    if ($this->activeSession !== null) {
        return $this->activeSession;
    }
    // ...existing core/portal dispatch...
}
```

```php
// SiteSetupListener, after creating the session
$session = $sessionFactory->createSession();
$request->setSession($session);
SessionWrapperFactory::getInstance()->setActiveSession($session);
```

I think this avoids the bag key mismatch, the lost request reference, and the need to detect session type from within the factory.

---

There's a subtle coupling in this idea to the destroy*Session() functions. I'm not sure how serious they are, but I thought I should just post what I have so far.

zmilan:
@kojiromike that is a very good point. I think it will not work for the core and portal distinction. But it could be a great way of dealing with API/OAuth. Thanks for the suggestion, I will investigate it more today.

src/Common/Csrf/CsrfUtils.php
```diff
@@ -52,9 +44,9 @@ public static function setupCsrfKey(?SessionInterface $session = null)
     //  $subject allows creation of different csrf tokens:
     //    Using 'api' for the internal api csrf token
     //    Using 'default' for everything else (for now)
-    public static function collectCsrfToken($subject = 'default', ?SessionInterface $session = null)
+    public static function collectCsrfToken($subject = 'default', ?SessionInterface $session = null): false|string
     {
-        $privateKey = !empty($session) ? $session->get('csrf_private_key', null) : $_SESSION['csrf_private_key'] ?? null;
+        $privateKey = $session?->get('csrf_private_key', null);
```
kojiromike: With the `$_SESSION` fallback removed, this returns `null` when `$session` is `null` — causing `collectCsrfToken()` to log an error and return `false`. Several callers still omit the session parameter:

- `interface/patient_tracker/patient_tracker.php` lines 833, 845, 886, 957, 1100, 1110 — `$session` is available at line 32 but not passed in these JS-embedded PHP blocks
- `interface/billing/era_payments.php` line 82 (`verifyCsrfToken`) — `$session` is available at line 74
- `src/Common/Twig/TwigExtension.php` line 189 — `csrfTokenRaw` uses a first-class callable that doesn't bind `$session` (see separate inline comment)

Should `$session` be made required (like `setupCsrfKey`), or should the method fall back to `SessionWrapperFactory::getInstance()->getActiveSession()` when null?

src/Common/Twig/TwigExtension.php
```diff
@@ -181,7 +181,7 @@ function ($subject = 'default', $fieldName = "_token") use ($session) {
                     if (empty($subject)) {
                         $subject = 'default';
                     }
-                    return sprintf('<input type="hidden" name="%s" value="%s">', $fieldName, attr(CsrfUtils::collectCsrfToken($subject, $session->getSymfonySession())));
+                    return sprintf('<input type="hidden" name="%s" value="%s">', $fieldName, attr(CsrfUtils::collectCsrfToken($subject, $session)));
```
kojiromike: The `csrfToken` function correctly captures `$session` via `use ($session)`. But the `csrfTokenRaw` function on the next line uses `CsrfUtils::collectCsrfToken(...)` as a first-class callable, which doesn't capture `$session`. This affects 12+ templates that use `{{ csrfTokenRaw() }}`. Would something like this work?

```suggestion
                    return sprintf('<input type="hidden" name="%s" value="%s">', $fieldName, attr(CsrfUtils::collectCsrfToken($subject, $session)));
                }
            ),
            new TwigFunction(
                'csrfTokenRaw',
                function ($subject = 'default') use ($session) {
                    return CsrfUtils::collectCsrfToken($subject, $session);
                }
```

interface/globals.php
```diff
@@ -227,19 +226,15 @@ function GetCallingScriptName()
...
-$globalsBag->set('vendor_dir', $GLOBALS['vendor_dir'] ?? "$webserver_root/vendor");
+$globalsBag->set('vendor_dir', $GLOBALS['vendor_dir'] ?? "$webserver_root/vendor"); // @phpstan-ignore nullCoalesce.offset ($GLOBALS key may not exist yet)
 $globalsBag->set('restRequest', $restRequest);
-$globalsBag->set('OE_SITES_BASE', $GLOBALS['OE_SITES_BASE'] ?? "$webserver_root/sites");
-$globalsBag->set('debug_ssl_mysql_connection', $GLOBALS['debug_ssl_mysql_connection'] ?? false);
+$globalsBag->set('OE_SITES_BASE', $GLOBALS['OE_SITES_BASE'] ?? "$webserver_root/sites"); // @phpstan-ignore nullCoalesce.offset
+$globalsBag->set('debug_ssl_mysql_connection', $GLOBALS['debug_ssl_mysql_connection'] ?? false); // @phpstan-ignore nullCoalesce.offset
 $globalsBag->set('eventDispatcher', $eventDispatcher ?? null);
 $globalsBag->set('ignoreAuth_onsite_portal', $ignoreAuth_onsite_portal);
 $read_only = empty($sessionAllowWrite);
```
kojiromike, line 235: This `$read_only` value is no longer used — `getActiveSession()` doesn't accept it, and `createCoreSession()` in `SessionWrapperFactory` always creates sessions with `readOnly=false`.

More importantly, even if it were threaded through, Symfony's `NativeSessionStorage::setOptions()` silently drops `read_and_close` because it's not in the hardcoded `$validOptions` whitelist (see `vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php:359-367`). And Symfony calls `session_start()` with no arguments — `read_and_close` only works when passed directly to `session_start(['read_and_close' => true])`.

This means sessions are now held open (file-locked) for the entire request. In the multi-frame UI where `main.php` spawns multiple iframes, concurrent requests from the same user will serialize waiting for the session lock. The old `$sessionAllowWrite` mechanism that ~26 files depend on is now inert.

The `setSession()`/`unsetSession()` methods also lost their open-write-close pattern — the old code called `session_write_close()` after each write. The new code just calls `$session->set()` and the session stays open.

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -16,39 +16,110 @@
...
-            $session = new PHPSessionWrapper();
-        } else {
-            $session = new SymfonySessionWrapper(SessionUtil::portalSessionStart());
+        if ($app === SessionUtil::PORTAL_SESSION_ID) {
+            return $this->getPortalSession();
+        }
+        return $this->getCoreSession();
+    }
+
+    public function getPortalSession(bool $reset = false): SessionInterface
```
kojiromike, line 103: Since the factory stores a single `$activeSession`, calling `getPortalSession()` after `getCoreSession()` (or vice versa) returns whichever was created first. Is the intent that only one session type is ever active per request? If so, might be worth documenting that. If not, would a keyed map (e.g., `$sessions['core']`, `$sessions['portal']`) be more appropriate?

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -16,39 +16,110 @@
...
+        if (!$request->hasSession()) {
+            $sessionFactory = new HttpSessionFactory($request, $web_root, HttpSessionFactory::SESSION_TYPE_PORTAL);
+            $this->activeSession = $sessionFactory->createSession();
+        } else {
+            $this->activeSession = $request->getSession();
+        }
+        return $this->activeSession;
+    }
+
+    private function createCoreSession(): SessionInterface
```
kojiromike, line 171: `HttpRestRequest::createFromGlobals()` creates a fresh Request every time. Since a newly-created Request never has a session attached, `$request->hasSession()` on the next line will always be `false` — making the `else` branch dead code. Should this be cleaned up, or is it anticipating a future pattern where the Request is injected?

kojiromike (commented):
The architectural direction is right, and the overall structure (singleton factory, Symfony `SessionInterface` everywhere, removal of the bridging `PHPSessionWrapper`/`SymfonySessionWrapper`) is clean. Great work pushing through such a large change.

I found a few issues that need attention, some functional and one architectural.

### 1. CSRF token breakage (functional bug)

`collectCsrfToken()` and `verifyCsrfToken()` still accept `$session = null` but now use `$session?->get(...)` with no fallback. Callers that don't pass a session will silently generate invalid tokens / fail validation. See inline comments for the specific callers.

### 2. Session locking / `read_and_close` regression (performance)

This is the big one. OpenEMR's multi-frame UI relies on `read_and_close` to prevent session file lock serialization. When a user loads the main UI, 3-5 iframes fire concurrent requests. Today, most pages open the session read-only (`read_and_close => true`), immediately release the lock, and only momentarily re-acquire it for writes via `SessionUtil::setSession()`'s open-write-close dance.

The `pid` flow illustrates this: when a user selects a patient, `PatientSessionUtil::setPid()` calls `SessionUtil::setUnsetSession()` which opens the session writable, writes `pid`, and immediately calls `session_write_close()`. Other frames read `pid` via read-only sessions without blocking.

Symfony's `NativeSessionStorage::setOptions()` has a hardcoded whitelist of valid options (line 359-367 in the Symfony source) and `read_and_close` is **not** in it — it gets silently dropped. Furthermore, Symfony calls `session_start()` with no arguments, but `read_and_close` only works when passed directly to `session_start(['read_and_close' => true])`. So:

- `$read_only` in `globals.php` is computed but never used
- `SessionConfigurationBuilder::setReadOnly()` sets the config but it's silently ignored by Symfony
- The new `SessionUtil::setSession()` just calls `$session->set()` without the open-write-close pattern
- Sessions are held open (locked) for the entire request lifetime

This will cause visible UI delays when loading the multi-frame interface. The `$sessionAllowWrite` mechanism that ~26 files rely on becomes inert.

Possible approaches:
- Use a database or Redis session handler (eliminates file locking entirely — predis-sentinel support already exists)
- Implement a custom `SessionStorageInterface` that passes `read_and_close` to `session_start()` directly
- For the write path, replicate the open-write-close pattern using `$session->save()` after writes

### 3. Remaining `$_SESSION` direct access

`interface/modules/zend_modules/public/index.php:39` reads `$_SESSION['pid']`, `$_SESSION['sessionUser']`, and `$_SESSION['portal_init']` directly. This file was not modified by the PR. Since Symfony Session manages its own storage bags, `$_SESSION` may not be populated the same way, potentially breaking portal-to-Zend-module integration.

### 4. Single session slot in factory

`SessionWrapperFactory` stores one `$activeSession`. Calling `getCoreSession()` then `getPortalSession()` returns the core session for both (since `$this->activeSession` is already set). The `$reset` parameter exists but callers have to know to use it. Is the intent that only one session type is ever active per request?

### Not blocking

- `HttpRestRequest::createFromGlobals()` in `createCoreSession()`/`createPortalSession()` creates a fresh Request, so `$request->hasSession()` is always false — the `else` branch is dead code.
- 81 commits could be squashed for a cleaner merge history.

zmilan:
@kojiromike about 1. CSRF token breakage (functional bug): I agree, and I made it mandatory initially, but phpstan was not happy since we have a first parameter with a default value and a second as mandatory... so I reverted it to be optional... I will rethink how to change it better without changing the order and updating all places in the code.

zmilan:
@kojiromike about **Remaining $_SESSION direct access**: I left it intentionally for now, since I was thinking that it is not in use anymore. I remember it being an issue with generating some reports from the portal, and then we changed that in some other PR, so it was not using that file at all. It would not be possible to pass the session id and recreate the session as it is done in that file - that is why I added a JWT token to try to solve it... But later, we removed it since we decided not communicate that way at all.

rollingventures:
I am performing E2E smoke testing of this PR against the `openemr/openemr:flex` Docker image. 

Testing revealed one critical issue that causes HTTP 500 errors on all endpoints, but I think it was just a file/touch creation accident and you can just delete it unless I am mistaken.

### Empty `global.php` Config File

**File:** `interface/modules/zend_modules/config/autoload/global.php`
**Commit:** `713459fd7`

The PR creates an empty (0 bytes) `global.php` file. When Laminas includes this file during module initialization, PHP returns `1` (the success indicator for `include`) instead of an array.

This causes:
Config being merged must be an array, implement the Traversable interface,
or be an instance of Laminas\Config\Config. integer given.

```
<?php

/**
 * Global autoload configuration for Laminas modules.
 *
 * @package   OpenEMR
 * @link      https://www.open-emr.org
 */

return [];
```

### Before
| Endpoint | Result |
|----------|--------|
| Core Login | ❌ HTTP 500 |
| FHIR API `/apis/default/fhir/metadata` | ❌ HTTP 500 |
| Main Interface | ❌ HTTP 500 |
| Patient Portal | ❌ HTTP 500 |

### After
| Endpoint | Result |
|----------|--------|
| Core Login | ✅ HTTP 302 (redirect to main) |
| FHIR API `/apis/default/fhir/metadata` | ✅ HTTP 200, valid CapabilityStatement |
| Main Interface | ✅ HTTP 200 |
| Patient Portal | ✅ HTTP 200 |

zmilan:
@kojiromike I just added a [commit](https://github.com/openemr/openemr/pull/10244/commits/f8ed51398aac8eff0695b427b0a13f8edb15f10a) that handles `read_and_close` with Symfony Session. Your help in verifying that with these changes, we do not have blocked requests would be a big help.

**Important Note:** Currently, this branch doesn't work at all, since `interface/modules/zend_modules/config/autoload/global.php` file is removed and it is still referenced. For my initial testing of these last changes, I just created that file again and returned an empty array from it.

rollingventures:
Ran comprehensive unit tests for all session types (Core, API, OAuth, Portal) via SessionWrapperFactoryTest.php (building on tests from PR #10404), plus new tests for read_and_close behavior in `ReadAndCloseNativeSessionStorageTest.php` and `SessionUtilReadAndCloseTest.php`. 

All 54 session-related tests pass.

One remaining discussed issue:
`interface/modules/zend_modules/config/autoload/global.php` is empty (0 bytes) and must contain <?php return []; or be deleted to prevent Laminas config merge errors.

@zmilan Awesome work!

src/Common/Session/SessionWrapperFactory.php
zmilan: The intention was to have only one active session per request. Do we have a use case to handle multiple sessions per request?

interface/globals.php
zmilan: This should be solved now.

src/Common/Twig/TwigExtension.php
zmilan: This is resolved

src/Common/Csrf/CsrfUtils.php
zmilan: This is fixed now, parameters are swapped, and $session is required. And missing lines are covered (they usually come after rebase with the latest master).

library/MedEx/API.php
```diff
@@ -2019,7 +2022,8 @@
                         $has_recall = !empty($processed['ALL']);
                         $this->recall_board_top($has_recall);
                         if ($has_recall) {
-                            echo $processed['ALL'];
+                            // Output pre-built HTML (escaped at construction in recall_board_process)
+                            echo is_array($processed) ? ($processed['ALL'] ?? '') : ''; // nosemgrep: echoed-request
```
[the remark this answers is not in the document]
zmilan: As the message above states, output pre-built HTML (escaped at construction in recall_board_process)

zmilan:
@rollingventures @kojiromike I have resolved all issues after adding support for `read_and_close`. It is again all green. So it is ready for review again.
I also cleaned git history (but because of phpstan and semgrep reports fixing, it is still more than 1 commit)

It would be a great help if someone with more experience with APIs could test this branch. I will do some by myself, but I have zero experience with it at the moment.

kojiromike:
@zmilan many of the conflicts are likely due to #11048. That PR has a rector rule that can probably resolve them for you.

zmilan:
> @zmilan many of the conflicts are likely due to #11048. That PR has a rector rule that can probably resolve them for you.

I am working on resolving it...

zmilan:
I converted it again to a draft since I discovered that I will need to put some more work and checks related to the `read_and_close` use cases. Once I cover that fully, I will open it again.

interface/globals.php
zmilan: I rewrote a lot of things around `read_and_close` functionality, even for the Portal part of the application. It should work correctly now, and by default, it is read-only.

src/Common/Session/SessionWrapperFactory.php
zmilan: It is fixed now. I am sorry for delaying... I thought I already did it...

interface/modules/zend_modules/public/index.php
```diff
@@ -37,10 +37,6 @@
         session_id($_REQUEST['me']);
         SessionUtil::sessionStartWrapper();
```
kojiromike: This `sessionStartWrapper()` starts a raw PHP session via `session_start()`. When `globals.php` runs on line 50 and calls `SessionWrapperFactory::getInstance()->getActiveSession()`, it creates a new `ReadAndCloseNativeSessionStorage` which checks `session_status()`. Since the raw PHP session is still active, it throws:

```
RuntimeException: Failed to start the session: already started by PHP.
```

On master, a double `session_start()` just triggers a PHP notice and continues. With the Symfony wrapper, it's a fatal crash. This affects portal CCDA endpoints and API-to-zend-module requests that pass the `me` parameter.

Would it make sense to either:
- Close the raw session before globals.php runs (`session_write_close()` after line 38/47), or
- Have the factory detect an already-active PHP session and bridge to it (similar to the old `setUseExistingSessionBridge` pattern)?

interface/login/login.php
```diff
@@ -45,6 +46,8 @@
 
 SessionUtil::setAppCookie(SessionUtil::CORE_SESSION_ID);
 
+$session = SessionWrapperFactory::getInstance()->getCoreSession();
```
kojiromike: This creates the session before `$sessionAllowWrite = true` is set on line 53. At this point, `SessionWrapperFactory::readOnly` is still `true` (the default), so the session is created with `read_and_close => true`.

When globals.php later calls `setSessionReadOnly(false)`, the session has already been created — the flag only affects future session creations.

This means the direct `$session->set()` calls later in this file (lines 185, 189, 227, etc.) write to Symfony's in-memory bags but the native session is already closed. Those writes aren't persisted to the session file — they're lost when the request ends.

Compare with `portal/index.php` which correctly calls `setSessionReadOnly(false)` before `getPortalSession()`. Should this file follow the same pattern?

```suggestion
SessionWrapperFactory::getInstance()->setSessionReadOnly(false);
$session = SessionWrapperFactory::getInstance()->getCoreSession();
```

src/RestControllers/Subscriber/SiteSetupListener.php
```diff
@@ -216,9 +218,9 @@ protected function setupCoreSessionBridge(HttpRestRequest $request, string $webr
             $webroot,
             HttpSessionFactory::SESSION_TYPE_CORE
         );
-        $sessionFactory->setUseExistingSessionBridge(true);
         $session = $sessionFactory->createSession();
```
kojiromike, line 221: The old `setUseExistingSessionBridge(true)` was removed, but this code path still needs to handle an already-active PHP session.

In the `APICSRFTOKEN` flow:
1. `loadApplicationGlobals()` runs globals.php, which starts a writable core session (since `$sessionAllowWrite = true`)
2. This method then creates a new `ReadAndCloseNativeSessionStorage` and calls `createSession()`
3. When the new session is accessed, `start()` finds `PHP_SESSION_ACTIVE` and throws `RuntimeException`

The bridge session also doesn't pass its storage to `setActiveSession()`, so `withWritableSession()` can't detect it as a `ReadAndCloseNativeSessionStorage`.

Would it make sense to check `session_status()` here and bridge to the existing session instead of creating a new one? Or should the globals.php session be closed first?

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -10,45 +10,172 @@
  * @package   OpenEMR
  * @link      https://www.open-emr.org
  * @author    Milan Zivkovic <zivkovic.milan@gmail.com>
+ * @author    Claude Code AI
```
kojiromike: `@author Claude Code AI` — OpenEMR file headers use human author attribution. Same issue in `ReadAndCloseNativeSessionStorage.php` and `ForbidDirectSessionWriteRule.php`.

src/Common/Session/PHPSessionWrapper.php
```diff
@@ -1,105 +0,0 @@
-<?php
```
kojiromike, line 1: This file was emptied to 0 bytes rather than deleted. Since `PHPSessionWrapper` is fully replaced by the Symfony `SessionInterface`, should this file be removed entirely? An empty PHP file could cause unexpected behavior if anything includes it.

interface/modules/zend_modules/public/index.php
kojiromike: **Correction:** I tested this in Docker and the crash does not actually happen. Symfony's `NativeSessionStorage::getBag()` detects an already-active PHP session (`session_status() === PHP_SESSION_ACTIVE`) and calls `loadSession()` instead of `start()`, bridging to the existing session without collision. The `RuntimeException` only fires on an explicit `start()` call, but `HttpSessionFactory::createSession()` never calls `start()` — it relies on lazy initialization through `getBag()`.

So this code path works correctly. Apologies for the false alarm.

src/RestControllers/Subscriber/SiteSetupListener.php
kojiromike: **Same correction applies here.** The bridge session created by `createSession()` is never explicitly `start()`-ed — it starts lazily through `getBag()`, which detects the already-active PHP session and bridges to it via `loadSession()`. Verified in Docker: a second `Session` object backed by a new `ReadAndCloseNativeSessionStorage` correctly reads and writes to the same underlying `$_SESSION` without crashing.

src/Common/Session/PHPSessionWrapper.php
zmilan: They were appearing after rebase... but now I will pay attention and double check after rebasing so I can be sure that they are not present again... thanks for removing them.

interface/modules/zend_modules/public/index.php
zmilan: @kojiromike, so I did an investigation and as a result removed the whole part of this code, since it is not in use anymore and is a security risk to keep it in the codebase.
The "me" parameter accepted a session ID from the request data and resumed it via session_id(). This was used by the old CDADocumentService HTTP client flow, removed in c4d18356e. The handler is now dead code and a session fixation vector.

src/RestControllers/Subscriber/SiteSetupListener.php
zmilan: @kojiromike, thanks for this feedback. I have investigated it more and fixed it with commit https://github.com/openemr/openemr/pull/10244/commits/22d2b532ee3b9af971bd4bb46b9618a78f9b4ee7

I will keep this not resolved so you can review it again.

kojiromike:
### Bug: `SessionCleanupListener` attribute name mismatch destroys sessions after internal API calls

I pushed a failing test in `6b813bb` that demonstrates this.

**How we found it:** While smoke-testing this PR in Docker, clicking through a patient's Demographics → History → Ledger → Insurance edit flow triggered a `400 Site ID is missing from session data!` error. After that, every subsequent request returned 400. Capturing a HAR file showed the session died immediately after `insurance_edit.php` made an internal API call to `/apis/default/api/patient/{uuid}/insurance`.

**Root cause:** `LocalApiAuthorizationController::shouldProcessRequest()` sets the attribute `_is_local_api` (with underscore prefix, line 28). But `SessionCleanupListener::onRequestTerminated()` checks for `is_local_api` (without underscore, line 26). The attribute is never found, so the listener treats every internal API request as external and calls `$session->invalidate(0)`, destroying all session data.

The insurance edit page is one of the few places in the core UI that makes an internal API call from a browser page, which is why normal page navigation works fine but this specific flow breaks.

**Fix:** Either change `SessionCleanupListener` to check `_is_local_api`, or change `LocalApiAuthorizationController` to set `is_local_api`. The test (`SessionCleanupListenerTest::testLocalApiRequestSessionIsPreserved`) will pass once the names match.

kojiromike:
Update: pushed `998deb5` with an improved version of the test. Instead of hardcoding the `_is_local_api` attribute name, the test now runs `LocalApiAuthorizationController::shouldProcessRequest()` to mark the request, then passes that same request to `SessionCleanupListener::onRequestTerminated()`. The test fails if the two classes disagree on the attribute name, without baking either name into the test itself.

kojiromike:
### Testing update

We spun this up in Docker and walked through the main flows:

- **Session locking (read_and_close):** Confirmed working. HAR captures show the multi-frame UI loading concurrently — no session lock serialization. This is a real improvement over master.
- **Patient selection + tab navigation:** Works. Session context preserved across Demographics, History, Ledger, Encounters.
- **Message sending / CSRF:** Works. Chart notes submit successfully, CSRF tokens validate correctly.
- **Encounter creation:** Works. Session writes for encounter context persist correctly.
- **Insurance edit → internal API call:** **Broken.** Session destroyed after the `/apis/default/api/patient/{uuid}/insurance` call. See my earlier comments and the failing test for the `SessionCleanupListener` attribute name mismatch.

### Remaining items

Beyond the session destruction bug (which needs to be fixed before merge), there are several other things we noticed. We'd like to get this merged soon — keeping a 780-file PR passing PHPStan and conflict-free is a tax on everyone. So: which of these would you rather address here, and which can be follow-up PRs?

1. **`globals.php` manual reopen bypass** (lines 278-284) — Manually checks `instanceof ReadAndCloseNativeSessionStorage` and calls `reopenForWriting()`/`save()` instead of using `SessionUtil::clearSession()`, which exists for exactly this purpose.

2. **Dead `SessionUtil::sessionStartWrapper()`** — No longer called anywhere. All session creation goes through `HttpSessionFactory` now.

3. **Dual enforcement strategy** — Both `WriteThroughSession` (runtime safety net) and `ForbidDirectSessionWriteRule` (PHPStan rule with ~20 whitelisted paths) exist to prevent the same problem. Worth clarifying which is primary. The large whitelist on the PHPStan rule undermines its value.

4. **Remaining `$_SESSION` references** — ~15 application files still use `$_SESSION` directly (e.g., `eye_mag_functions.php`, `calendar/index.php`, `date_functions.php`, `FormService.php`). Some may be intentionally left, but worth confirming.

5. **`CsrfUtils` breaking signature change** — Parameter order flipped and `$session` became required. Correct change, but module developers calling `collectCsrfToken()` or `verifyCsrfToken()` will break. Needs a note in upgrade docs.

6. **Portal messaging username display** — `getAuthPortalUsers()` now correctly uses `portal_username`, but the message summary shows that internal identifier instead of the patient's name. Filed as #11202 — this is pre-existing UX, not a regression.

7. **`/meta/health/index.php` reopening for writes** — The Docker health check triggers a session reopen-for-write on every poll (~1/min). Not harmful but noisy in the logs.

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -16,39 +16,165 @@
...
+    private ?SessionStorageInterface $activeStorage = null;
+
+    private bool $readOnly = true;
+
+    /**
+     * Whether Redis is used for session storage. When true, read_and_close
+     * is disabled because Redis has no file lock contention — the reopen
+     * cycle would only add unnecessary round-trips.
+     */
+    private ?bool $isRedisSession = null;
```
bradymiller, line 42: hi @zmilan ,
redis does have a locking mechanism. I was under impression from @adunsulag , that symfony session supported this for redis.

zmilan: I was checking for this, and this is only what I found:
"The `redis.session.locking_enabled` `php.ini` directive is applied when the `session.save_handler` is `redis` (for rediscluster session locking may not work properly). Symfony `RedisSessionHandler` has no locking mechanism. The documentation says: "Symfony will now use your Redis server to read and write the session data. The main drawback of this solution is that Redis does not perform session locking, so you can face race conditions when accessing sessions. "

Issue #10219, opened 2026-01-15
refactor(payments): Create service for payments internals


Firehed:
Today, there are a number of places that re-implement similar logic around payments handling - both in terms of API calls and especially data storage and retrieval. INSERT statements alone for just the `ar_activity` and `ar_session` tables have roughly a dozen queries each, with substantially similar implementations. There are also several related tables - `payments`, `onsite_portal_activity`, `billing`, and possibly others that I don't yet understand.

I'm not sure what this looks like yet - it's certainly fairly complex given the nature of payments in general (never mind in the healthcare space). The changes will probably need to be incremental - start by centralizing the write queries, then switch some strings to constants, then go from there. Centralizing the logic to retrieve the data will also be a huge win. Ideally some dedicated data-types instead of untyped arrays, but let's tackle one thing at a time.

Firehed:
@sjpadgett Maybe you have thoughts here? You're certainly more familiar than I am with payments right now, plus you'd brought up getting payments stuff moved into modules. I'll generally defer to you on specifics. I want to be mindful of how there are currently three external providers, plus `InHouse` (which I assume is some sort of "other"/cash accounting type box); OCE is working on adding another and someone in the forums expressed a desire on yet another.

I've got some high-level thoughts on a "driver" model, and I want to be sure that we record which provider was used for any given payment for accounting and general handling purposes, even if it's not actively used yet. This is getting more into enhancements territory than purely refactoring.

sjpadgett:
@Firehed InHouse is a pathway for user to collect credit and payment and so user can manually verify card and post payment.
As far as the rest, my plan was to move/clone to a module and work on improvements then hook and switch on when ready then pull old code.
I'll leave it up to you because you certainly can handle the security stuff better than me. I'll help you with the workflow details if you need to break out in service plus I recommend a reengineering of the gateway. As is it really doesn't follow its original intent(got sidetracked).
Just thought all of this would be easier in a module.
However you approach I'm sure it'd be great when done!

PR #10246, opened 2026-01-16
refactor(payments): Centralize some of the data insertion logic

Firehed:
Part of #10219, but that's really too big to get fixed in one.

Duplicated logic for inserting to `ar_activity` is consolidated

- Adds a new class that contains the logic for adding the next-sequential record into `ar_activity` for a given patient+encounter
- Converts most (but not all) places that were doing this to use the new helper

In doing so:
- The sequence number thing is now centralized
- There's a path to improving type safety for several of the fields (e.g. switching to enums, `Money`)
- Some redundant date handling has been eliminated

No

### DIFF
--- a/src/PaymentProcessing/Recorder.php
+++ b/src/PaymentProcessing/Recorder.php
@@ -62,7 +62,7 @@ class Recorder
                 `payer_claim_number`
             ) VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?);
             SQL;
-        QueryUtils::inTransaction(function () use ($query, $data) {
+        QueryUtils::inTransaction(function () use ($query, $data): void {
             $now = date('Y-m-d H:i:s');
             $next = $this->getNextSequenceNumber(
                 patientId: $data['patientId'],


src/PaymentProcessing/Recorder.php
```diff
@@ -82,8 +86,10 @@ public function recordActivity(array $data): void
                 $data['memo'] ?? '',
                 $data['accountCode'] ?? '',
                 $data['followUp'] ? 'y' : '',
-                $data['followUpNote'] ?? '',
-                $data['reasonCode'] ?? '',
+                $data['followUpNote'] ?? null,
+                $data['reasonCode'] ?? null,
+                $data['postDate'] ?? null,
+                $data['payerClaimNumber'] ?? null,
```
kojiromike, line 92: Some day all this type-duality will be behind us.

Firehed: I hope so! There's a LOT more to do here (this should be a service that takes a _couple_ of values, not ~15) but progress is progress.

interface/patient_file/pos_checkout_ippf.php
```diff
@@ -1613,31 +1615,21 @@ function markTaxes($taxrates): void
...
-                    $encounter_id,
-                    $sequence_no['increment'],
-                    $code_type,
-                    $code,
-                    $session->get('authUserID'),
-                    $this_bill_date,
-                    $postdate,
-                    $memo,
-                    $adjust
+                $recorder = new Recorder();
```
sjpadgett, line 1618: FYI ippf is the Planned Parenthood custom feature. I don't know if they are still using but you'll see this designation several other places in core.

Firehed: I was wondering about that, thanks for the explanation!

interface/billing/edit_payment.php
```diff
@@ -231,26 +221,20 @@
...
-                        "', session_id = '" . trim(formData('payment_id')) .
-                        "', modified_time = '" . trim(add_escape_custom($created_time)) .
-                        "', pay_amount = '" . 0 .
-                        "', adj_amount = '" . 0 .
-                        "', memo = '" . "Deductible $" . trim(formData("Deductible$CountRow")) .
-                        "', account_code = '" . "Deduct" .
-                        "'");
-                    sqlCommitTrans();
+                    $recorder->recordActivity([
+                        'patientId' => trim(formData("HiddenPId$CountRow")),
```
kojiromike, line 225: Do we not have a cbf rule that makes this…

```suggestion
                        'patientId' => trim(formData("HiddenPId{$CountRow}")),
```

?

interface/billing/edit_payment.php
```diff
@@ -158,28 +161,20 @@
...
-                        "', post_user = '" . trim(add_escape_custom($user_id)) .
-                        "', session_id = '" . trim(formData('payment_id')) .
-                        "', modified_time = '" . trim(add_escape_custom($created_time)) .
-                        "', pay_amount = '" . trim(formData("Payment$CountRow")) .
-                        "', adj_amount = '" . 0 .
-                        "', account_code = '" . add_escape_custom($AccountCode) .
-                        "'");
-                    sqlCommitTrans();
+                    $recorder->recordActivity([
+                        'patientId' => trim(formData("HiddenPId$CountRow")),
```
bradymiller, line 165: looks like there will be a lot of double escaping going on
(formData will escape for sql)
(then the binding in queryutils)

To migrate this old sql escaping mechanism, in past I have basically removed formData and directly collected from the POST parameters.

Firehed: Oh drat, that's probably true. I can adapt these to `trim($_POST['whatever'] ?? '')` unless there's a more-preferred approach in place. I'll send a follow up PR following your guidance on this.

kojiromike: Given how much `trim(formData("thing"))` is in here, an abstraction would not be out of place anyway.

bradymiller:

interface/billing/edit_payment.php
bradymiller: hi, ok with abstraction or the direct POST call for formData (note the add_escape_custom calls just get removed)

PR #10247, opened 2026-01-16
feat: rel-800 Fixes #10233 cqm reporting period and deps updates (#10234)

### LINKED ISSUE #10233: feat: Update ecqm reporting period

stephenwaite:
* Fixes #10233 cqm reporting period and deps updates

Update a number of dependencies in the cqm, ccda, and schematron pieces to bring them up to date.

Log an error if a cqm reporting file is not found.

Update the default reporting period to be at least 2025 instead of 2022 which no longer even exists in the project files

* Fix mocha package, phpstan error

Fixes #

### DIFF
--- a/ccdaservice/package-lock.json
+++ b/ccdaservice/package-lock.json
@@ -12,20 +12,19 @@
                 "packages/*"
             ],
             "dependencies": {
+                "body-parser": "1.20.4",
                 "file": "./oe-blue-button-generate",
                 "libxmljs2": "0.37.0",
                 "lodash": "4.17.21",
                 "moment": "2.30.1",
                 "net": "1.0.2",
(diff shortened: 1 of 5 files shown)


adunsulag:
@stephenwaite odd, I swear I pushed a PR with this backport earlier today but now I see that you did it.  Strange, but thanks for bringing this in.

PR #10253, opened 2026-01-17
feat: Fixes #10251 Related Person save existing patient when locale date format is not YYYY-MM-DD.

### LINKED ISSUE #10251: bug: Related Person form fails to add relation for existing patient when global date format is not yyyy-mm-dd
### Describe the bug

The related person feature of looking up an existing patient in the system to add as a related person fails if the localization of the date format is set to dd/mm/yyyy or mm/dd/yyyyy.  This is because the patient controller fails the validation when the calling code retrieves the patient's DOB which is default formatted as YYYY-MM-DD.  

### To Reproduce

1. Set the Admin -> Config -> Locale -> Date Display Format to be DD/MM/YYYY format
2. Add two new patients via the Finder -> New Patient / Search
3. Edit an existing patient's demographic information
4. Select the Related Person tab
5. Click the plus icon next to related person
6. Enter in the first name of the other patient you created
7. When the patient pops up in the search box, hit the Select box
8. Verify the Related Person form is populated with the patient, hit the save button at the top of the screen
9. See in the Dashboard screen's demographics page that the Related Person was not saved.

### Expected behavior
The Related Person that is added is saved.

### Client configuration

- Browser: Firefox and Chrome
- OpenEMR version: OpenEMR 7.0.4
- Operating system: Ubuntu LTS 24

This bug was originally reported by Open Plan IT.

adunsulag:
Using an existing patient was failing to save due to the DOB validation on the person controller.  The person controller was assuming the DOB it receives was formatted in a localized format.  Whereas the ContactRelationService was grabbing the DOB directly from the patient_data.DOB field which is formatted in the YYYY-MM-DD format.  So if a localized format is used, the existing patient fails to add / save.

Fixes #10251

### DIFF
--- a/src/Services/ContactRelationService.php
+++ b/src/Services/ContactRelationService.php
@@ -728,7 +728,8 @@ class ContactRelationService extends BaseService
             'middle_name' => $patient['mname'] ?? '',
             'title' => $patient['title'] ?? '',
             'suffix' => $patient['suffix'] ?? '',
-            'birth_date' => $patient['DOB'] ?? '',
+            // birth date is expected to be in the format the frontend entered it so we have to convert it to an OpenEMR date
+            'birth_date' => DateFormatterUtils::oeFormatShortDate($patient['DOB'] ?? ''),
             'gender' => $patient['sex'] ?? '',
             'ssn' => $patient['ss'] ?? '',
             'email_direct' => $patient['email_direct'] ?? '',


bradymiller (approved):

kojiromike:
@adunsulag is this a feature or a fix?

PR #10260, opened 2026-01-17
fix(medex): recall event iteration

### LINKED ISSUE #10257: is_countable issue
### Describe the bug

Problem: In [API.php]
 
$camps = '0';
if (is_countable($events)) {
        
Really what we want is $events to not be an empty array.

$camps = '0';
if (! empty($events) && is_array($events)) { 

does not seem to work either.  I will look deeper.

ophthal:
Fixes #10257 and the recall board.  

Problem:

Display::show_progress_recall($recall, $events) is called from multiple paths with different $events shapes.
In some cases $events is a single associative event array (not a list). Iterating that with foreach yields scalar values, so checks like $event['M_group'] break and recall progression/campaign handling fails.
Solution:

Normalize $events into a consistent $eventList:
If $events looks like a single event (keys like M_group, M_type, C_UID), wrap it as [$events].
If it’s already a list of events, use it as-is.
If it’s empty/non-array, safely no-op.
Impact:

Recall Board and background/cron paths handle “no events”, “single event”, and “multiple events” consistently without PHP warnings or broken progression logic.
Files:

[API.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Notes:

AI-generated portion is clearly marked in the code (normalization block).

### DIFF
--- a/library/MedEx/API.php
+++ b/library/MedEx/API.php
@@ -2324,8 +2324,22 @@ class Display extends Base
         $show['progression'] .= $show['EMAIL']['text'] . $show['SMS']['text'] . $show['AVM']['text'];
 
         $camps = '0';
-        if (is_countable($events)) {
-            foreach ($events as $event) {
+        // BEGIN AI-GENERATED CODE (GitHub Copilot)
+        // Normalize $events: callers may pass either a single event array or a list of event arrays.
+        // If a single event array is iterated as-is, foreach() yields scalar values and breaks lookups
+        // like $event['M_group'].
+        $eventList = [];
+        if (is_array($events)) {
+            if (isset($events['M_group']) || isset($events['M_type']) || isset($events['C_UID'])) {
+                $eventList = [$events];
+            } else {
+                $eventList = $events;
+            }
+        }
+        // END AI-GENERATED CODE (GitHub Copilot)
+
+        if (!empty($eventList) && is_array($eventList)) {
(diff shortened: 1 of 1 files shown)


library/MedEx/API.php
```diff
@@ -2324,8 +2324,22 @@ public function show_progress_recall($recall, $events = '')
...
+            if (isset($events['M_group']) || isset($events['M_type']) || isset($events['C_UID'])) {
+                $eventList = [$events];
+            } else {
+                $eventList = $events;
+            }
+        }
+        // END AI-GENERATED CODE (GitHub Copilot)
+
+        if (!empty($eventList) && is_array($eventList)) {
+            foreach ($eventList as $event) {
```
kojiromike, line 2342: Would you consider simplifying this block? `array_is_list()` (available since PHP 8.1) can detect whether `$events` is a sequential list vs an associative single-event array, without sniffing for specific keys. And the `if (!empty(...))` guard isn't needed since `foreach` on an empty array is a no-op.

This would also fix the PHPStan and Rector CI failures.

```suggestion
        // Normalize $events: callers may pass either a single event array or a list of event arrays.
        $eventList = match (true) {
            !is_array($events) => [],
            array_is_list($events) => $events,
            default => [$events],
        };

        foreach ($eventList as $event) {
```

PR #10263, opened 2026-01-18
feat(phpstan): adopt openemr-phpstan-rules package

### LINKED ISSUE #10262: feat(phpstan): adopt openemr-phpstan-rules package for consistent static analysis
## Summary

Adopt the `opencoreemr/openemr-phpstan-rules` package to provide consistent PHPStan rules across OpenEMR core and third-party modules.

## Rationale

Currently, OpenEMR has custom PHPStan rules in `.phpstan/` that enforce coding standards. By extracting these rules into a standalone Composer package, we enable:

1. **Module developers** can install the same rules in their modules, ensuring consistency with OpenEMR core
2. **Standalone module analysis** - modules can be analyzed without having OpenEMR core installed
3. **Centralized rule maintenance** - rule updates benefit both core and modules simultaneously

## Rules Provided

The package enforces:
- **Database**: Forbid legacy `sql.inc.php` functions and Laminas-DB in favor of `QueryUtils`
- **Globals**: Forbid direct `$GLOBALS` access in favor of `OEGlobalsBag`
- **HTTP**: Forbid raw `curl_*` functions in favor of PSR-18 HTTP clients
- **Testing**: Forbid `@covers` annotations that exclude code from coverage
- **Controllers**: Require Symfony Request/Response objects (for modules)
- **Exceptions**: Require catching `Throwable` instead of `Exception`
- **Deprecations**: Report usage of deprecated code via `@deprecated` annotations

## Implementation

- Install `opencoreemr/openemr-phpstan-rules` package
- Add `@deprecated` annotations to legacy `sql.inc.php` functions
- Local `.phpstan/*.Rule.php` files can be removed once the package is validated

## Links

- Package: https://packagist.org/packages/opencoreemr/openemr-phpstan-rules
- Repository: https://github.com/opencoreemr/openemr-phpstan-rules

kojiromike:
Fixes #10262

## Summary

Adopt the `opencoreemr/openemr-phpstan-rules` package to provide consistent PHPStan rules across OpenEMR core and third-party modules.

- Add `opencoreemr/openemr-phpstan-rules` package as dev dependency
- Add `@deprecated` annotations to legacy `sql.inc.php` functions
- Simplify `.phpstan/README.md` to reference package documentation

## AI Disclosure

Yes

### DIFF
--- a/.phpstan/baseline/function.deprecated.php
+++ b/.phpstan/baseline/function.deprecated.php
@@ -1,6 +1,6 @@
 <?php declare(strict_types = 1);
 
-// total 6138 errors
+// total 6136 errors
 
 $ignoreErrors = [];
 $ignoreErrors[] = [
@@ -9263,18 +9263,6 @@ $ignoreErrors[] = [
 Use QueryUtils\\:\\:fetchRecordsNoLog\\(\\) instead$#',
     'count' => 1,
     'path' => __DIR__ . '/../../src/Common/Crypto/CryptoGen.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Call to deprecated function edi_generate_id\\(\\)\\:
-Use QueryUtils\\:\\:ediGenerateId\\(\\) instead$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../src/Common/Database/QueryUtils.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Call to deprecated function generate_id\\(\\)\\:
-Use QueryUtils\\:\\:generateId\\(\\) instead$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../src/Common/Database/QueryUtils.php',
 ];
 $ignoreErrors[] = [
     'message' => '#^Call to deprecated function sqlBeginTrans\\(\\)\\:
--- a/.phpstan/baseline/loader.php
+++ b/.phpstan/baseline/loader.php
@@ -1,6 +1,6 @@
 <?php declare(strict_types = 1);
 
-// total 27705 errors
+// total 27701 errors
 
 return ['includes' => [
     __DIR__ . '/arguments.count.php',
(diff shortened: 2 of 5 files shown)


Firehed:
> Add `@deprecated` annotations to legacy sql.inc.php functions

I'm in favor of this, but I'm a little lost as to the implication of `@deprecated` (or `#[Deprecated]`, though that one actually has runtime implications) vs the comment [here](https://github.com/opencoreemr/openemr-phpstan-rules?tab=readme-ov-file#why-custom-rules-instead-of-just-deprecated)

kojiromike:
> > Add `@deprecated` annotations to legacy sql.inc.php functions
> 
> I'm in favor of this, but I'm a little lost as to the implication of `@deprecated` (or `#[Deprecated]`, though that one actually has runtime implications) vs the comment [here](https://github.com/opencoreemr/openemr-phpstan-rules?tab=readme-ov-file#why-custom-rules-instead-of-just-deprecated)

Are you asking why do both? It isn't necessary today, but it's an aid to a future idea. I want to create a tool in openemr-phpstan-rules that automatically builds new rules like this. That is, by providing that tool access to openemr, it can figure out what static rules to create. Then _those rules_ can be applied to a module without having openemr available at check time.

Furthermore, we can combine that with e.g. https://github.com/openCoreEMR/openemr-phpstan-rules/issues/4 to provide version-specific rulesets based on actually looking at the code in a given version of openemr, automatically, so if we deprecate something in the future, we don't flag it in modules looking to support earlier versions of openemr.

Firehed:
Works for me. I'd strongly like to see the phpstan-deprecation-rules make its way in ASAP (either through this or standalone) as it'll greatly simplify a bunch of DB improvements.

Longer-term I think having the rules package move into the main openemr org would be good, but hardly a blocker for me.

kojiromike:
btw, this particular PR still isn't perfectly aligned with the previous findings. I'm not sure why. I will probably have to come back to it.

adunsulag:
@kojiromike do you want to convert this to draft if you're still working on it?

PR #10279, opened 2026-01-20
feat(db): Introduce DBAL-based wrapper

### LINKED ISSUE #9944: Introduce doctrine/dbal support
I'd like to introduce support for the `doctrine/dbal` library, with the near-term goal of moving the existing DB wrapper tooling to use it and off of ADOdb. This would be opaque to the rest of the application - no breaks or code changes for existing first-party or third-party code. It would not remove ADOdb nor break any tooling directly relying on it.

## Why?

There's a tremendous wealth of knowledge and tooling support built around `doctrine/dbal`; it's by far the most widely used DBAL system in modern PHP code. While `ADOdb` appears to get active maintenance, the way its built doesn't lend itself well to writing testable, modular code. 

When compared to the existing adodb tooling, doctrine/dbal has the following advantages:

- No reliance on global variables
- No need for manual, non-Composer inclusion
- Excellent support for middleware
- Improved support for newer DB types & their PHP equivalents (date-related, json, etc)  

Notably, the middleware support means that we can support things like:

- Query logging for easier development  
- Audit trails  
- Database health and performance measurements

Some of this has been built in-house; some is missing entirely. Centralizing this logic makes it easy to do the right thing and hard to do the wrong thing, which is critical for future development and security.

`doctrine/dbal` is the foundation for other widely-used tools in the ecosystem that could really benefit the project (most notably `doctrine/migrations` and `doctrine/orm`), though the inclusion of either is strictly out scope for this discussion.

## How?

Procedurally, this looks like:

- Install the library  
- Configure and make available a connection (using the same credentials)
- Update the existing wrappers (`sql.inc.php`, `QueryUtils`) to use the DBAL connection  
  - Inverting the "canonical" implementation from sql.inc.php to QueryUtils probably simplifies this  
- Soon: add a DBAL connection into the app config/container (related: #9867)

I've already done a proof of concept locally, and didn't encounter any real snags. The various functions/methods of the wrappers don't need to all be updated at once (nor should they be; I'd do them incrementally with appropriate test coverage), and this should be opaque to the rest of the application. The lbirary alone does no appreciable type conversion on SELECTed values, so there's no risk of e.g. a `DateTime` showing up somewhere that used to return a `string`; however, new code would have the ability to leverage those type conversions if desired. Any code using the `sqlQuery`(etc)/`QueryUtils` wrappers would behave identically to how they do now.

Assuming no opposition to the concept, I'll open child issues for the various steps and feel comfortable owning the majority of the migration. From there, we can discuss next steps.

## Transitional impact

- During the time where DBAL and ADOdb both exist, any given request may open an additional database connection. This could have performance implications for the database server. However, it seems that the main sticking point is the GACL tooling, which is *already* opening a unique connection.

## Challenges

- The interaction points with `PRIV_DB` will require some extra work to maintain feature parity.  
- The `GenID`\-based functions don't have a *direct* replacement, but the connection has a `lastInsertId()` method that can be used to implement an equivalent without much hassle.

## Scope

In scope:
- Add the library
- Update calls (primarily wrappers) to go through its connection
- MAYBE adding a proxy hook thing to detect and warn about code not using the db wrapper tooling

NOT in scope (yet?):
- Adopting doctrine/orm
- Adopting doctrine/migrations
- Removing existing db wrapper tooling
- Removing adodb entirely

I think on _some_ timeline, the not-in-scope items, including those that would be BC breaks, should get discussed - but they've got their own challenges and I I think incremental improvement is more important right now.

Firehed:
Fixes #9944

Adds a backwards-compatibility wrapper to interact with the new DBAL tooling from the existing sql functions. Mostly, it interacts with the existing DB config system, including connection certificates.

- Creates a new singleton-style class with utility methods for SELECTing one row, SELECTing many rows, and performing data changes
- Creates a callback-based wrapper to mimic the legacy-but-widespread `HelpfulDie` path[^hd]
- A couple basic unit tests

This does NOT integrate this path into the existing SQL tools (sql.inc.php, QueryUtils) - it's still dormant for now. Doing so will be a fast-follow PR; I think it'll be easier to actually do these all at once to avoid weird multi-connection issues (and possible performance loss) rather than iteratively like I'd originally planned. See #9980, which I'll repurpose into the actual integration.

> [!IMPORTANT]
> `QueryUtils` will remain, for now, the canonical and preferred way to interact with the database. This should be considered an implementation detail. A future change that improves support for interacting directly with the dbal or a library built on top of it may change this, but that's out of scope for now.

No

[^hd]: While unlikely, shifting from `die()` to an exception might lead to weird behavior if it's caught and handled. No new code should be written that relies on this, but there's so much that targets that API that testing such a change is effectively impossible.

### DIFF
--- a/src/BC/Database.php
+++ b/src/BC/Database.php
@@ -35,7 +35,7 @@ use PDOException;
  * `ADOdb` and `laminas-db`. For now, continue to use the existing wrappers
  * (e.g. QueryUtils) for database interactions.
  *
- * In the future, the DBAL `Connection` may be made avaible through a DI
+ * In the future, the DBAL `Connection` may be made available through a DI
  * container, and/or something like `doctrine/orm` for most interactions.
  *
  * @internal


src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
...
+     * @param literal-string $table The sequence table (usually `sequences`)
+     * @return int The next incremental ID
+     *
+     * Caution: this _may_ have data races in the current implementation,
+     * especially across parallel requests.
+     */
+    public function generateSequentialId(string $table): int
+    {
+        // Warning: table names cannot be parameterized.
+        $query = sprintf('UPDATE %s SET id=LAST_INSERT_ID(id+1)', $table);
```
Firehed, line 226: This implementation comes directly from the current ADODB internals. Good or bad, it's a known quantity.

src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
+<?php
+
```
adunsulag, line 2: Missing copyright header here.

src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
...
+     * e.g. SELECT a, b FROM foos would return something like this:
+     * [
+     *   ['a' => 'a1', 'b' => 'b1'],
+     *   ['a' => 'a2', 'b' => 'b2'],
+     * ]
+     *
+     * @param Bindings $bindings
+     * @return array<string, mixed>[]
+     */
+    public function fetchAll(string $sql, array $bindings): array
```
adunsulag, line 182: Should we add a mode or an another method to fetch as an array.  I ran into this with GACL the other day as the DB results there are fetched by array index order instead of by associative.  Very annoying but would let us migrate some of that code pathway as well.

tests/Tests/BC/DatabaseTest.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
```
adunsulag, line 2: Missing copyright header.

adunsulag (changes requested):
Minor change on the copyright.  Otherwise it looks fine to me.  Noted also one additional function that could be added to handle php GACL cases.

Also not sure if you want to rely on the filesystem but you could add a test case for the SSL options to verify that code is being exercised properly.

src/BC/Database.php
Firehed: Good note - I think we can cross that bridge when we get to it. There are a LOT of paths in right now and this won't cover everything,

src/BC/Database.php
kojiromike, line 1: Please add the standard docblock annotations.

src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
...
+            return $action();
+        } catch (DBALException $e) {
+            self::helpfulDieDbal($e);
+        }
+    }
+
+    // Down-convert an exception into a crash for extra backwards compat
+    private static function helpfulDieDbal(DBALException $e): never
+    {
+        $sql = $e->getQuery()->getSQL();
```
kojiromike: Might `getQuery` potentially return `null` here?

src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
...
+ * existing code without changes, replacing connections that are managed with
+ * `ADOdb` and `laminas-db`. For now, continue to use the existing wrappers
+ * (e.g. QueryUtils) for database interactions.
+ *
+ * In the future, the DBAL `Connection` may be made avaible through a DI
+ * container, and/or something like `doctrine/orm` for most interactions.
+ *
+ * @internal
+ *
+ * @phpstan-type Bindings array<string|int|float>
```
kojiromike:
```suggestion
 * @phpstan-type Bindings array<string|int|float|bool|null>
```
?

tests/Tests/BC/DatabaseTest.php
```diff
@@ -0,0 +1,73 @@
...
+declare(strict_types=1);
+
+namespace OpenEMR\BC;
```
kojiromike:
```suggestion
namespace OpenEMR\Test\BC;
```

src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
...
+ * `library/sql.inc.php`.
+ *
+ * DO NOT WRITE CODE THAT INTERACTS WITH THIS CLASS DIRECTLY!
+ *
+ * The aim of this class is to have a way to hook `doctrine/dbal` in to
+ * existing code without changes, replacing connections that are managed with
+ * `ADOdb` and `laminas-db`. For now, continue to use the existing wrappers
+ * (e.g. QueryUtils) for database interactions.
+ *
+ * In the future, the DBAL `Connection` may be made avaible through a DI
```
kojiromike:
```suggestion
 * In the future, the DBAL `Connection` may be made available through a DI
```

src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * @phpstan-type Bindings array<string|int|float>
+ */
+class Database
+{
+    /**
+     * Singleton instance. Future scope may have this removed in favor of some
+     * sort of DI/service container.
+     */
+    private static ?Database $instance = null;
```
kojiromike, line 51: Do we need a way to reset this for testing?

src/BC/Database.php
```diff
@@ -0,0 +1,264 @@
...
+
+        $certDir = sprintf('%s/documents/certificates', $siteDir);
+        $caFile = sprintf('%s/mysql-ca', $certDir);
+        $cert = sprintf('%s/mysql-cert', $certDir);
+        $key = sprintf('%s/mysql-key', $certDir);
+        if (file_exists($caFile)) {
+            $options[PDO::MYSQL_ATTR_SSL_CA] = $caFile;
+        }
+        if (file_exists($cert) || file_exists($key)) {
+            if (!file_exists($cert) || !file_exists($key)) {
```
kojiromike, line 141: Should we do a readability check as well as an existence check?

src/BC/Database.php
Firehed: We might, but I'd rather not add it proactively until we know it's needed. Same general issue as the globalsbag singleton problems. This class is just to get the sql.inc and queryutils on to dbal, and then asap start getting new code using dbal (or orm) directly which should bypass this need entirely.

src/BC/Database.php
Firehed: Goodness, you're absolutely right. It's worse actually - `getQuery()` is only on some of the specific exceptions, not the base. Need to get the PHPStan level bumped ASAP to catch this.

src/BC/Database.php
Firehed: Ordinarily I'd say yes, but in this case I think it's better not to:

1) This is consistent with existing behavior in the other places that read the files
2) If it's present but not readable, that's almost certainly a filesystems permission error and I think it's best to fail loudly rather than silently ignoring this and falling back to plaintext communications, leaving a false sense of security

tests/Tests/BC/DatabaseTest.php
Firehed: Whoops, my personal file generator utils make some assumptions for test cases that aren't true in OEMR!

src/BC/Database.php
Firehed: SOME DAY I'm going to figure out how to integrate spellcheck into vim, or at least in a way that actually works. It's been enabled for years (decades?) and never once worked.

adunsulag:
Thanks @Firehed for the change.  Looking forward to the next PRs.

bradymiller:

PR #10281, opened 2026-01-20
fix: Clinical Notes FHIR mismatched join #10168

### LINKED ISSUE #10168: FHIR DocumentReference endpoint returns clinical notes associated with the wrong patient
### Describe the bug

Hi folks, 

The FHIR DocumentReference endpoint returns clinical notes associated with the wrong patient. This occurs because `ClinicalNotesService.php` uses an incorrect SQL join that matches `forms.id` to `form_clinical_notes.form_id`, instead of the correct relationship `forms.form_id = form_clinical_notes.form_id`.

When clinical notes are created via `saveArray()`, both `forms.form_id` and `form_clinical_notes.form_id` are set to the same generated value. However, the FHIR query aliases `forms.id` (the auto-increment primary key) as `form_id` and joins on that, causing notes to be incorrectly associated with whichever patient happens to have a `forms.id` matching the note's `form_id`.

### To Reproduce

1. Create Patient A with an encounter
2. Add a clinical note to Patient A's encounter (this creates `forms.form_id = X` and `form_clinical_notes.form_id = X`)
3. Create Patient B with an encounter and add enough clinical notes so that a `forms.id` value equals X
4. Query the FHIR DocumentReference endpoint for Patient B: `GET /fhir/DocumentReference?patient=<Patient_B_UUID>`
5. Patient A's clinical note incorrectly appears in Patient B's results

### Expected behavior

The FHIR DocumentReference endpoint should only return clinical notes that belong to the queried patient. The join should use `forms.form_id = form_clinical_notes.form_id` (the intentional linkage created by `saveArray()`) rather than `forms.id = form_clinical_notes.form_id`.

### Suggested Fix

In `src/Services/ClinicalNotesService.php`, around line 96, change:

```php
JOIN (
    SELECT
        id AS form_id,
        encounter
        ,pid AS form_pid
        ,`date` AS date_created
    FROM
        forms
) forms ON forms.form_id = notes.form_id
```

To:

```php
JOIN (
    SELECT
        id,
        form_id,
        encounter
        ,pid AS form_pid
        ,`date` AS date_created
    FROM
        forms
    WHERE formdir = 'clinical_notes'
) forms ON forms.form_id = notes.form_id
```

This uses the actual `forms.form_id` column instead of aliasing `forms.id`, and adds a `formdir` filter for efficiency.

### Client configuration

- Browser: N/A (API issue)
- OpenEMR version: 7.x (confirmed in Docker image `openemr/openemr:7`)
- Operating system: Linux (Docker)

### Additional context

The creation logic in `saveArray()` correctly links records:
- Generates `form_id = MAX(form_clinical_notes.form_id) + 1`
- Calls `addForm()` which sets `forms.form_id = generated_value`
- Saves clinical note with `form_clinical_notes.form_id = generated_value`

The bug is isolated to the FHIR fetch query, not the creation logic.

All users were created manually through the UI. 

I actually updated the SQL like so to get it working: 

Updated form_clinical_notes.form_id to match the forms.id of the correct patient's forms entry, so that OpenEMR's FHIR join (forms.id = form_clinical_notes.form_id) would return each note under its actual patient instead of a random one.

But obviously the above won't solve the problem for new patients created.

adunsulag:
Fixed an issue where patients could get the wrong clinical notes based upon the form id joining on the FHIR query.  Made sure both the encounter form and the clinical note forms would not show up for the wrong patient by adding an additional safety check on the patient pid.

Also added in the author population for clinical notes which was missing.

Fixes #10168

### DIFF
--- a/src/Services/ClinicalNotesService.php
+++ b/src/Services/ClinicalNotesService.php
@@ -97,6 +97,7 @@ class ClinicalNotesService extends BaseService
                         ,`code`
                         ,codetext
                         ,`description`
+                        ,`pid` AS notes_pid
                         ,external_id
                         ,clinical_notes_type
                         ,note_related_to
@@ -109,21 +110,24 @@ class ClinicalNotesService extends BaseService
              ) notes
             JOIN (
                 SELECT
-                    id AS form_id,
-                    encounter
+                    id
+                    ,form_id
+                    ,encounter
                     ,pid AS form_pid
                     ,`date` AS date_created
                 FROM
                     forms
-            ) forms ON forms.form_id = notes.form_id
+                WHERE formdir = 'clinical_notes'
+            ) forms ON forms.form_id = notes.form_id AND forms.form_pid = notes.notes_pid
             LEFT JOIN (
                 select
                     encounter AS eid
                     ,uuid AS euuid
                     ,`date` AS encounter_date
+                    ,`pid` AS encounter_pid
                 FROM
                     form_encounter
-            ) encounters ON encounters.eid = forms.encounter
+            ) encounters ON encounters.eid = forms.encounter AND forms.form_pid = encounters.encounter_pid
             LEFT JOIN
             (
                 SELECT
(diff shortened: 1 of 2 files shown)


adunsulag:
We need to get this in, I'll merge this in tomorrow if I don't hear anything from another reviewer.

sjpadgett:
I don't have time to test so okay with me otherwise.

Issue #10249, opened 2026-01-17
refactor(security): Tighten types for SQL tooling


Firehed:
While in the long term we want to adopt broadly-safer tooling for interacting with the database, there are thousands of calls to the various DB wrappers that we'll have to maintain and eventually convert. Given that, I think we can adopt a few small changes to the existing database tools that can greatly increase safety:

1) Make bindings a required parameter

Not using any bound parameters is a strong hint that there's risk of a SQL injection attack. Far from a guarantee, but I don't think it's unreasonable to say that _every known instance of this should be audited_.

If you actually have a query that doesn't need parameters (`SELECT * FROM foos`), pass `[]` explicitly and it'll pass analysis.

2) Annotate the query parameter with the [`literal-string`](https://phpstan.org/writing-php-code/phpdoc-types#other-advanced-string-types) type. 

While there are both valid situations for dynamic strings and ways to handle them safely, they're _extremely high-risk_.

Note that for this to do anything useful, we need to run PHPStan at level 5 or greater. #10237 will take care of this.

Unlike most PHPStan refactors, I believe these _should not_ get baselined, but either fixed or explicitly `@phpstan-ignore`'d if we can determine it's not at-risk.

Given the scope of the change, this should probably be done one or two functions at a time, rather than all ~50 at once (yes, there should be fewer access paths... that's its own issue)

msummers42:
Strongly support this effort. Thanks @Firehed .

kojiromike:
I agree that they should not be baselined; however, I think we should merge the phpstan powerup first, then remove these explicitly from the baseline in a separate PR to highlight them specifically.

Firehed:
Yes, I'd like to see that land first prior to tackling this @kojiromike 👍

PR #10288, opened 2026-01-21
feat(security): PHPStan rule to require use of query bindings

Firehed:
Progress on #10249 (not a complete fix)

Requires bindings to be passed to various DB functions instead of inlined queries, de-risking SQL injection attacks.

- Adds a general PHPStan rule that soft-enforces function and method parameters that we _want_ to be required without actually changing the code and risking a BC break.
- Adds a couple calls into the, list, disabled for now.

As noted on #10249, these are probably NOT things we should add to the baseline; instead, the errors should actually be inspected and either fixed or explicitly ignored[^ign]. As such, I think we should turn them on one at a time, fix the problems, and commit the enablement along with the fixes. The surface area is huge, so trying to do this across dozens of sql functions that take bindings all at once is... unnecessarily ambitious.

Basically, this does nothing _yet_ but will enable rapid iterative progress towards the functions safer. That doesn't take away from migrating away from adodb+laminas+more but there's more foundation work to do first on that.

Yes: ChatGPT for some general research and AST tricks in the rule body, substantially edited by hand.

[^ign]: There's almost no case for ignoring them; pass an empty array in the handful of places it's legit.

### DIFF
--- a/.phpstan/SoftRequiredArgumentsRule.php
+++ b/.phpstan/SoftRequiredArgumentsRule.php
@@ -12,14 +12,14 @@ declare(strict_types=1);
 
 namespace OpenEMR\PHPStan\Rules;
 
+use OpenEMR\Common\Database\QueryUtils;
 use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\StaticCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleError;
 use PHPStan\Rules\RuleErrorBuilder;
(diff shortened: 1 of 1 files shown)


kojiromike:
I'm still confused by the baseline thing. I don't see a baseline in this PR, but I also don't see phpstan failing. Shouldn't there be one or the other?

Firehed:
Basically, the rule sets up a system where we can turn on analysis for each individual SQL function, but none of those functions are actually turned on yet so that we can fix them one at a time rather than put them into the baseline

kojiromike:
Please fix conflicts, then I will merge

bradymiller:

Issue #10294, opened 2026-01-21
bug(swagger): Try it out / Authorize not working


igormukhingmailcom:
Our currently available `openemr_auth (OAuth2, authorizationCode with PKCE)` flow described at OpenAPI annotations is not working.

Steps:
- Click `Authorize` button
- Fill fields with valid `client_id`, `client_secret`
- Select all scopes
- Click `Authorize`

It redirects to something like `http://localhost:8300/oauth2/default/authorize?response_type=code&client_id=0jNm3Sg_m8QoNh3By1mKiicCeRmAeEuHwkyMSOwKE58&redirect_uri=http%3A%2F%2Flocalhost%3A8300%2Fswagger%2Foauth2-redirect.html&scope=openid%20offline_access&state=V2VkIEphbiAyMSAyMDI2IDIxOjI3OjMyIEdNVCswMjAwIChFYXN0ZXJuIEV1cm9wZWFuIFN0YW5kYXJkIFRpbWUp&code_challenge=VLoBznV3UzcNoIivSSvJALL8rtWq5765noTYETU2O3M&code_challenge_method=S256`

Response will be:

```
{"error":"invalid_client","error_description":"Client authentication failed","message":"Client authentication failed"}
```

bradymiller:
hi @igormukhingmailcom .
Need to use https (use the 9300 port on the easy dev environment). I pretty much forget about this everytime i test swagger :)

igormukhingmailcom:
Hi, @bradymiller :)

Thanks for the idea, but it did not help:

<img width="727" height="88" alt="Image" src="https://github.com/user-attachments/assets/1348a234-ff51-4429-9ed5-2a0e0b98dd8f" />

When authorising at postman - I successfully using something like, for example:

```
URL: /oauth2/default/token

user_role:users
grant_type:password
client_id:{{client_id}}
client_secret:{{client_secret}}
password:{{ADMIN_PASSWORD}}
username:{{ADMIN_LOGIN}}
scope:{{ALL_SCOPES}}
```

bradymiller:
hi @igormukhingmailcom , that is odd. I literally just tested swagger api yesterday and it worked. (And made the same mistake always make when don't test it for awhile (used http insteaf of https).)

bradymiller:
i generally create a client via:
openemr-cmd roc

And then enable the client in OpenEMR.

PR #10295, opened 2026-01-21
feat(swagger): Added Bearer token to Swagger UI

igormukhingmailcom:
... to be able to use `Try it out` with manually generated token (via `Postman`, for example)

As currently available `openemr_auth (OAuth2, authorizationCode with PKCE)` flow is not working - https://github.com/openemr/openemr/issues/10294.

## Usage

```php
    /**
     *  @OA\Get(
     *      path="/api/user/me",
     *      security={{"bearer":{}}},                <------------------------------------------
     *  )
     */
    'GET /api/user/me' => static function (HttpRestRequest $request): ResponseInterface {},
```

### Entering token

image

### Entered token

image

### Successful response from endpoint using that bearer auth

image

### DIFF
--- a/swagger/openemr-api.yaml
+++ b/swagger/openemr-api.yaml
@@ -9885,6 +9885,11 @@ components:
         head_circ: '22.2'
         oxygen_saturation: '96'
   securitySchemes:
+    bearer:
+      type: http
+      description: 'Enter a previously generated Bearer token here to authenticate'
+      bearerFormat: JWT
+      scheme: bearer
     openemr_auth:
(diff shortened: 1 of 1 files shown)


adunsulag:
@igormukhingmailcom Thanks for the improvement.  On that other issue with authorize we need more details because the oauth2 is working for me and it appears for one of the other admin's as well.

bradymiller:
Hi, Do we have confirmation that the other authentication methods are working? If so, then will likely port this to rel-800.

PR #10296, opened 2026-01-21
chore(scopes): Multilined scopes at ApiTestClient

igormukhingmailcom:
... to have less conflicts / manage easier / have smaller diffs


adunsulag:
This looks good to me @igormukhingmailcom . Thanks for the contribution.

bradymiller:

PR #10297, opened 2026-01-21
chore(phpstan): Restrict new function additions to global namespace

### LINKED ISSUE #10293: chore(phpstan): Create rule around global function definitions
This could go one of several ways, but the root issues I want to address are:

- There are a lot of functions that have duplicate definitions; these only _don't_ crash since most are in directly-requested controller-type files
  - This is causing active problems in CI with PHPStan (most likely #10237 and more)
  - This is likely to disrupt work on a front controller system
- Some of the duplicated functions have conflicting signatures and/or implementations
- Where they're the same, it means we have completely copy-pasted code, which makes bug fixes less effective and refactors harder
- Functions in the global namespace are generally discouraged and can make upgrades more difficult (the global namespace is de-facto reserved for PHP itself these days)

A lot of them also rely on global state and other things we want to discourage, but that's a separate concern (however, it may make fixing them harder)

I made a rough scanner, and found:
- 2212 functions and classes in the global namespace
- 83 of those _names_ have more than one definition (anywhere from 2-6, so effectively a couple hundred problems)

No idea on active conflicts.

Broadly, I'd like a PHPStan rule that bans new functions being defined outside of the `autoload_files` path, and discourage any new global functions in general. Existing ones can be incrementally a) converted to a closure, b) converted into a class method under normal psr-4 autoload rules or c) moved into an autoloaded file (roughly in order of preference). The third option requires duplicate cleanup or it will result in a crash at runtime.

Unfortunately due to the runtime of PHPStan it can't reliably detect duplicate definitions, but flagging functions outside of a well-defined location should be easy enough. It or rector may be able to assist with a fix.

Firehed:
Fixes #10293

Prevents additional functions from being defined in the global namespace via PHPStan rules

New PHPStan rule to restrict defining new functions in the global baseline. It's not hardline forbidden, but restricts it to a few places which are in the autoloader path. This is intended to reduce the risk of duplicate (or substantially equivalent) definitions and encourage discovery of existing helper utilities.

All existing places added to the baseline.

Yes, small amounts from chatgpt building and refining the rule


.phpstan/baseline/loader.php
```diff
@@ -1,6 +1,6 @@
 <?php declare(strict_types = 1);
 
-// total 18982 errors
+// total 20673 errors
```
kojiromike: more

Firehed: There will be!

Firehed:
Yes. I did tweak the baseline script as part of this to avoid some future merge conflicts as well.

bradymiller:

PR #10302, opened 2026-01-22
feat(telemetry): Enhance telemetry

### LINKED ISSUE #10290: feat(telemetry): Report population capture and environment flag
- Start gathering instance patient and user population. I may break out users by their role, patient by portal or not etc.
- Add an environment variable to existing startup test to disable telemetry if a session flag is set in main.php. I currently use the session flag to disable in test because I could not find a way to set an environment flag that was in scope from main.php on test run. @kojiromike do you want to recommend the variable name for me to capture in telemetry init?

sjpadgett:
Add environment variable check to disable telemetry

Fixes #10290

### DIFF
--- a/tests/Tests/Isolated/Telemetry/TelemetryServiceTest.php
+++ b/tests/Tests/Isolated/Telemetry/TelemetryServiceTest.php
@@ -402,38 +402,6 @@ class TelemetryServiceTest extends TestCase
         $this->assertFalse($result);
     }
 
-    public function testReportUsageDataReturnsFalseWhenSiteUuidNotFound(): void
-    {
-        /** @var TelemetryRepository|MockObject $mockRepository */
-        $mockRepository = $this->createMock(TelemetryRepository::class);
-
-        /** @var VersionServiceInterface|MockObject $mockVersionService */
-        $mockVersionService = $this->createMock(VersionServiceInterface::class);
-
-        /** @var SystemLogger|MockObject $mockLogger */
-        $mockLogger = $this->createMock(SystemLogger::class);
-
-        // Create a partial mock to mock the isTelemetryEnabled method and getUniqueInstallationUuid
-        $telemetryService = $this->getMockBuilder(TelemetryService::class)
-            ->setConstructorArgs([$mockRepository, $mockVersionService, $mockLogger])
-            ->onlyMethods(['isTelemetryEnabled', 'getUniqueInstallationUuid'])
-            ->getMock();
-
-        // Mock isTelemetryEnabled to return true (enabled)
-        $telemetryService->expects($this->once())
-            ->method('isTelemetryEnabled')
-            ->willReturn(1);
-
-        // Mock getUniqueInstallationUuid to return empty string
-        $telemetryService->expects($this->once())
-            ->method('getUniqueInstallationUuid')
-            ->willReturn('');
-
-        $result = $telemetryService->reportUsageData();
-
-        $this->assertFalse($result);
-    }
-
     /**
      * Test reportUsageData with successful scenario.
      * Note: This method still has external dependencies (cURL) that would require


interface/main/tabs/main.php
```diff
@@ -41,6 +41,18 @@
 $allowRegisterDialog = $product_row['allowRegisterDialog'] ?? 0;
 $allowTelemetry = $product_row['allowTelemetry'] ?? null; // for dialog
 $allowEmail = $product_row['allowEmail'] ?? null; // for dialog
+
+// Check if telemetry is disabled via environment variable
+$env_name = 'OPENEMR_DISABLE_TELEMETRY';
```
kojiromike: If you like, this variable could either disable telemetry, or if a non-empty string (email address?), it could auto-register an openemr instance. I might also suggest this could be a const.

interface/main/tabs/main.php
```diff
@@ -41,6 +41,18 @@
...
+$val = getenv($env_name);
+if ($val === false || $val === '') {
+    $val = $_ENV[$env_name] ?? $_SERVER[$env_name] ?? null;
+}
+$disableTelemetry = $val !== null && in_array(strtolower(trim((string)$val)), ['1', 'true', 'yes', 'on', 'y'], true);
```
kojiromike:
```suggestion
$disableTelemetry = filter_var($val, FILTER_VALIDATE_BOOLEAN);
```

will replace most of this – unless recognizing 'y' is very important, in which case

```
$disableTelemetry = $val === 'y' || filter_var($val, FILTER_VALIDATE_BOOLEAN);
```

sjpadgett: Well I was unclear as to how you wanted to use the const so I tested for general values. I'm happy to implement however you need this to be set.
I'd think we'd need a different const if an auto register is desired e.g. OPENEMR_AUTO_REGISTER = {address}.
Maybe best if you just provide a code snippet of how you'd like to see this implemented. I don't have a problem with that at all.

sjpadgett:
Okay. I'll still make your suggested changes but I'm still working on adding the population additions unless this is holding you up.

sjpadgett:
Any suggestions?

```
array (
  'total_patients' => '3',
  'total_portal_patients' => '3',
  'total_users' => '7',
  'users_by_type' => 
  array (
    0 => 
    array (
      'abook_type' => '',
      'user_count' => '6',
    ),
    1 => 
    array (
      'abook_type' => 'external_provider',
      'user_count' => '1',
    ),
  ),
)
```

sjpadgett:
I need opinion on adding portal card click event counts. Working now on modules currently enabled. Any other static data we'd want collected?

kojiromike:
@sjpadgett is there any way we can distinguish via telemetry if the user is using one of the repo docker/*?

sjpadgett:
> repo docker

I'll look into.
I think maybe getenv('OPENEMR_DOCKER_ENV_TAG')
@kojiromike 
We already collect and here is an example snap of partial query from reported

image

kojiromike:
@sjpadgett so if it's docker/production/docker-compose.yml it'd be `production-docker` in telemetry?

sjpadgett:
@kojiromike I suppose. I collect the runtime environment variable  being reported of our dockers. You'd need to point me in a direction if there is additional docker runtime info wanted.

bradymiller:
@kojiromike ,
This is the variable that gets reported (is used on easydev, insanedev, and the demo dockers). Could also place it in the production docker if wish with a tag:
https://github.com/openemr/openemr/blob/master/docker/development-easy/docker-compose.yml#L68

bradymiller:
above link appears to not work (not highlighting the line). it is line 68. the 'OPENEMR_DOCKER_ENV_TAG' setting

kojiromike:
I think it'd be good to know for any of the official openemr deployment strategies, which ones are being used. Maybe each one, and then another one for one that openemr didn't make.

sjpadgett:
I forgot I put portal tracking in so yeah! I could add further tracking but later

image

sjpadgett:
First reporting interval 1203 so telemetry has been out a year in April

Geo tracking only reports for non-usa sites. Errors on IP lookup on US installs

sjpadgett:
Someone needs to look at [Isolated Tests / PHP 8.5 - Isolated Tests (pull_request)]
QueryUtils.php is trying to call the function sqlStatementThrowException(), which is not defined or not available in the current test context.
Location: src/Common/Database/QueryUtils.php:164 Called from: src/Telemetry/TelemetryService.php:145 During test: tests/Tests/Isolated/Telemetry/TelemetryServiceTest.php:432

I don't have a whitespace issue if I run ci and PhpStorm.  I have found that if another test fails for some reason whitespace sometimes fails. Maybe move it up in order.

Still I'm not trying to fix these fails.

src/Telemetry/TelemetryRepository.php
```diff
@@ -67,4 +67,68 @@ public function clearTelemetryData(): void
...
+        // Total active users
+        $sql = "SELECT COUNT(*) AS total_users FROM users WHERE active = 1 AND username IS NOT NULL AND fname IS NOT NULL";
+        $result = $this->fetchRecords($sql, [], noLog: true);
+        $population['total_users'] = $result[0]['total_users'] ?? 0;
+
+        // Active users grouped by abook_type
+        $sql = "SELECT abook_type, COUNT(*) AS user_count 
+            FROM users 
+            WHERE active = 1 AND username IS NOT NULL AND fname IS NOT NULL 
+            GROUP BY abook_type";
```
kojiromike: The trailing whitespace is in this sql. My suggestion would be to write multiline sql using now documents.

```suggestion
        $sql = <<<'SQL'
            SELECT abook_type, COUNT(*) AS user_count
              FROM users
             WHERE active = 1
               AND username IS NOT NULL
               AND fname IS NOT NULL
             GROUP BY abook_type
        SQL;
```

src/Telemetry/TelemetryService.php
```diff
@@ -130,7 +130,6 @@ public function reportUsageData(): int|bool
         $site_uuid = $this->getUniqueInstallationUuid();
         if (empty($site_uuid)) {
             error_log("Site UUID not found.");
-            return false;
```
kojiromike, line 133: Removing this return causes the isolated tests to end up in code that they can't run, so they fail.

src/Telemetry/TelemetryRepository.php
sjpadgett: I can agree with that. I usually don't do multi line but claude does. I like now document and use often but multi line is used all over core code.
My problem is if my local ide and ci test don't catch, is there any way to give a hint to check for this type of pattern in the fail result?
thanks btw

src/Telemetry/TelemetryService.php
sjpadgett: So do I need to fix the test or are you? I don't want to exit here.

kojiromike: Well, the test that's failing is asserting that this function returns `false` when the uuid is not found. So if that's not true anymore, then the test isn't useful anymore.

src/Telemetry/TelemetryRepository.php
kojiromike: What IDE do you use?

sjpadgett:
> I don't have a whitespace issue if I run ci and PhpStorm

src/Telemetry/TelemetryRepository.php
```diff
@@ -94,10 +94,13 @@ public function fetchSitePopulationData(): array
         $population['total_users'] = $result[0]['total_users'] ?? 0;
 
         // Active users grouped by abook_type
-        $sql = "SELECT abook_type, COUNT(*) AS user_count 
-            FROM users 
-            WHERE active = 1 AND username IS NOT NULL AND fname IS NOT NULL 
-            GROUP BY abook_type";
+        $sql = <<<SQL
```
kojiromike: If you put single quotes around this it won't allow variables to expand in the text, which means we can use static analysis to reduce the risk of sql injection later.

```suggestion
        $sql = <<<'SQL'
```

sjpadgett:
What do I need to run whitespace local? I finally got my phpstan matching actions run again.

kojiromike:
@sjpadgett install `pre-commit` (https://pre-commit.com/) and set it up to run on your local clone.

PR #10346, opened 2026-01-23
fix: #10332 uuid generation large dataset

### LINKED ISSUE #10332: bug: UUID generation slows to a crawl when generating 1 million uuids for a single table
### Describe the bug

The uuid generation slows to a crawl due to having hundreds of thousands of queries buffered inside a transaction.  Each 100K records updated began hitting exponential increases (10 seconds first batch, 20 seconds, then 40, etc).

Breaking the transactions up to perform commits every 10K rows allows UUID generation to finish all million records in a few minutes (checking for existance and insertion).

### To Reproduce
1. Generate 5K patients in synthea ```openemr-cmd irp 5000``` which currently skips generating uuid records for procedure_results table.  5K patients results in about ~1 million+ records for procedure results and 250K records for encounters
2. Run the https://localhost:9300/sql_upgrade.php upgrade
3. After 1 hour discover that process still has not completed (on mysql database show full processlist will show queries still executing)

### Client configuration

- Browser: Chrome
- OpenEMR version: 7.0.3+
- Operating system: Ubuntu 24 LTS

adunsulag:
When dealing with uuid generation for hundreds of thousands or millions of records in a single table, the transaction cost of all of those checks/updates slows the uuid generation process down immensely.  Changing the transaction commit to push updates out every 10K records speeds up the UUID generation process immensely.

Fixes #10332

### DIFF
--- a/.phpstan/baseline/openemr.deprecatedSqlFunction.php
+++ b/.phpstan/baseline/openemr.deprecatedSqlFunction.php
@@ -7062,21 +7062,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Use QueryUtils\\:\\:inTransaction\\(\\) instead of sqlBeginTrans\\(\\)\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Use QueryUtils\\:\\:inTransaction\\(\\) instead of sqlCommitTrans\\(\\)\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Use QueryUtils\\:\\:inTransaction\\(\\) instead of sqlRollbackTrans\\(\\)\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Use QueryUtils\\:\\:fetchRecords\\(\\) or QueryUtils\\:\\:fetchArrayFromResultSet\\(\\) instead of sqlFetchArray\\(\\)\\.$#',
     'count' => 6,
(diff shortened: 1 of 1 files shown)


adunsulag:
Will try clearing the phpstan cache locally and seeing if that reproduces these issues, but looks like this may be a github versus local problem.

adunsulag:
@kojiromike I can't reproduce the phpstan issue locally.  Do you have any idea why the .phpstan/baseline/openemr.deprecatedSqlFunction.php would not be triggered locally? 

I believe (after some hunting down) that is where these errors are originating from now instead of the consolidated phpstan baseline file we used before.

I've cleared the phpstan cache but that didn't seem to make a difference.

kojiromike:
@adunsulag that's interesting. I have definitely had phpstan behave differently in CI and locally, but in this case, I am able to reproduce the phpstan error locally in a fresh clone. But your question doesn't quite make sense to me:

> Do you have any idea why the .phpstan/baseline/openemr.deprecatedSqlFunction.php would not be triggered locally?

That isn't what's happening. The baseline has ignored errors in it. Either:

1. `deprecatedSqlFunction` is being triggered, and not finding anything, or
2. `deprecatedSqlFunction` is not being triggered (so still not finding anything)

Either way, the no-longer-needed baseline ignored error patterns would trigger this phpstan failure. There are only two ways you would not see this specific phpstan error locally. Either:

1. You have a config override that sets `parameters.reportUnmatchedIgnoredErrors` to `false`,
2. Phpstan is still looking at a UuidRegistry.php that has three `sqlRollbackTrans` occurring in it.

adunsulag:
Strange.  The reportUnmatchedIgnoredErrors is set to true in the config (haven't changed the value)
image

The ignored errors are still present in the deprecatedSqlFunction file
image

The functions have been removed though:
image

I must have some other kind of cache still hanging around.  I'll try a fresh clone and see what happens there.

adunsulag:
@kojiromike Fresh clone still doesn't trigger the error
image

Bizarre.  Well, I'll just remove the error lines in the baseline and push to resolve the issue but still strange.  Wonder if its a platform difference as I believe you're running this on a mac right?

kojiromike:
I am running on a mac, but the github action is linux.

Does it make any difference if you run phpstan on everything?

adunsulag:
@kojiromike looks like the problem doesn't crop up when running with a list of files.  It does happen when running over the entire project.  That means my current short-cuts of only running phpstan against files I've github staged won't work.  I'll have to figure out why its ignoring the baseline errors in that mode.

bradymiller:
nice!

PR #10362, opened 2026-01-23
fix(payments): Fix double-escaping and some raw SQL

### LINKED ISSUE #10361: bug(payments): double-escaped values following refactor
https://github.com/openemr/openemr/pull/10246#discussion_r2714067608

I mistakenly thought these calls were raw values, and it was pointed out they get escaped deeper down the call stack.

Firehed:
Fixes #10361

Values could be double-escaped following a refactor to use prepared statements

Changes most calls to `formData` to a new wrapper function that _does not_ escape, which is safe to use with the prepared statements.

For the few places not directly passed in, I verified that they either were on a different prepared statement path (in which case they were already double-escaped prior to the refactor) or converted them to a PS.

This eliminates _most_ calls to `formData`. There are a few left which will be more complex to update (see the dbal comment inline)

No

### DIFF
--- a/interface/billing/edit_payment.php
+++ b/interface/billing/edit_payment.php
@@ -107,17 +107,17 @@ if (isset($_POST["mode"])) {
             $updatedValues['global_amount'] = $globalReset;
         }
 
-        $updatedValues['user_id'] = trim($user_id);
+        $updatedValues['user_id'] = trim((string) $user_id);
         $updatedValues['closed'] = trim($closed);
         $updatedValues['reference'] = trimPost('check_number');
-        $updatedValues['check_date'] = trim($check_date);
-        $updatedValues['deposit_date'] = trim($deposit_date);
+        $updatedValues['check_date'] = trim((string) $check_date);
+        $updatedValues['deposit_date'] = trim((string) $deposit_date);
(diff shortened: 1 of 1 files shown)


interface/billing/edit_payment.php
```diff
@@ -101,27 +102,33 @@
...
+        $updatedValues['modified_time'] = trim($modified_time);
+        $updatedValues['payment_type'] = trimPost('type_name');
+        $updatedValues['description'] = trimPost('description');
+        $updatedValues['adjustment_code'] = trimPost('adjustment_code');
+        $updatedValues['post_to_date'] = trim($post_to_date);
+        $updatedValues['payment_method'] = trimPost('payment_method');
+
+        // This becomes MUCH more straightforward with actual dbal, but this is
+        // still safe from SQLI since the keys are all string literals.
+        $query = 'UPDATE ar_session SET ';
```
Firehed, line 125: ar_session interactions need a similar treatment to what was done in the refactor that prompted this cleanup.

bradymiller:
nice stuff!
(do not plan to port this to rel-800 (unless you think should go there))

Firehed:
@bradymiller This and #10246 should stay as a pair. I think both should stay out.

PR #10387, opened 2026-01-26
perf(ci): restore file timestamps to enable PHPStan caching

### LINKED ISSUE #10386: perf(ci): PHPStan cache not effective in GitHub Actions
## Summary

PHPStan runs take ~5-6 minutes in CI despite cache being restored. The cache is successfully downloaded but PHPStan still analyzes all ~4000 files from scratch.

## Evidence

From a recent workflow run ([#21345625828](https://github.com/openemr/openemr/actions/runs/21345625828)):

```
Cache restored from key: phpstan-result-cache-21345503684
```

But then:
```
    0/4000 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0%
  220/4000 [▓░░░░░░░░░░░░░░░░░░░░░░░░░░░]   5%
  ...
 4000/4000 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
```

## Root Cause

PHPStan has two cache layers:

1. **Result cache** - Uses content hashes for file changes ✓
2. **Container cache** (Nette DI) - Uses **file modification times** ✗

The container cache stores file mtimes in metadata files. When GitHub Actions does a fresh checkout, all files get current timestamps, which differ from the cached values, invalidating the container and triggering a full rebuild.

Additionally, `scanDirectories: ./vendor` in phpstan.neon.dist means vendor changes affect the cache. The `composer remove rector/rector` step regenerates the autoloader after composer install, which may invalidate cache entries.

## Proposed Solutions

1. **Restore file timestamps after checkout** using `git-restore-mtime`:
   ```yaml
   - run: pip install git-restore-mtime && git restore-mtime
   ```

2. **Add verbose output** (`-vvv`) to confirm cache invalidation reason

3. **Reorder steps** - Move rector removal before cache restore so vendor state is consistent

4. **Consider removing `scanDirectories: ./vendor`** if full vendor scanning isn't necessary (could use more targeted `scanFiles` instead)

## References

- [PHPStan Result Cache docs](https://phpstan.org/user-guide/result-cache)
- [phpstan/phpstan#2122: Cache should never rely on file modification time](https://github.com/phpstan/phpstan/issues/2122)
- [phpstan/phpstan#9301: Result cache on CI](https://github.com/phpstan/phpstan/discussions/9301)

kojiromike:
## Summary

Fixes #10386

PHPStan's container cache (Nette DI) validates against file modification times. Git doesn't preserve mtimes, so a freshly checked out file always looks newer than its cached metadata, invalidating the entire container and forcing a full re-analysis of all ~4000 files even when the result cache restored cleanly.

## Approach

The expensive parts of fixing this are (1) a full-history clone, which `git-restore-mtime` needs to walk, and (2) installing the tool itself. Both are pure waste on a cold cache miss. So this PR makes them conditional on whether a cache was actually restored:

- Replace `actions/cache@v5` with `actions/cache/restore@v5` + `actions/cache/save@v5` so the restore step exposes a `cache-hit` / `cache-matched-key` output.
- When (and only when) a cache was restored: install `git-restore-mtime` from apt, `git fetch --unshallow`, and walk the history to restore mtimes.
- Save the cache on `always()` so partial results from a failed analyze still seed the next run.

Cold misses pay zero new overhead vs. before this PR. Warm hits pay an unshallow + mtime walk to make the restored cache validate.

## Testing

The first run after merge is still cold-cache (no savings). Compare runtime on the second and third runs of the same branch to confirm the cache is now effective.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -48,14 +48,30 @@ jobs:
           | .version
         ' composer.lock)
         printf 'version=%s\n' "$version" >> "$GITHUB_OUTPUT"
-    - name: PHPStan Cache
+    - name: Restore PHPStan Cache
       # https://phpstan.org/user-guide/result-cache
-      uses: actions/cache@v5
+      id: phpstan-cache
+      uses: actions/cache/restore@v5
       with:
         path: tmp-phpstan # same as in phpstan.neon
         key: phpstan-result-cache-${{ steps.phpstan-version.outputs.version }}-${{ hashFiles('phpstan.neon.dist', '.phpstan/**', 'tests/PHPStan/**') }}-${{ github.run_id }}
         restore-keys: |
           phpstan-result-cache-${{ steps.phpstan-version.outputs.version }}-${{ hashFiles('phpstan.neon.dist', '.phpstan/**', 'tests/PHPStan/**') }}-
+    # PHPStan's container cache uses file mtimes for validation. Git doesn't
+    # preserve mtimes, so without restoration every cached file looks stale
+    # and the cache is effectively useless. Only worth doing when we actually
+    # restored a cache - on a cold miss this is pure overhead.
+    - name: Restore file timestamps
+      if: steps.phpstan-cache.outputs.cache-matched-key != ''
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get install -y --no-install-recommends git-restore-mtime
+        # Guard against running on a non-shallow checkout - --unshallow errors
+        # with "does not make sense" if history is already complete.
+        if [ -f .git/shallow ]; then
+          git fetch --unshallow --quiet
+        fi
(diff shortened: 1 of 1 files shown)


kojiromike:
## Known Limitation

This fix restores mtimes for **git-tracked files only** (phpstan.neon.dist, .phpstan/*.neon, etc.). However, the container cache metadata also includes **vendor extension configs**, which will still have current mtimes because:

- Composer cache stores downloaded packages (tarballs), not extracted files
- `composer install` extracts fresh with current timestamps

This means the container cache may still be invalidated by vendor file mtimes. Testing will determine if this partial fix provides meaningful improvement, or if additional measures are needed (e.g., touching vendor files to match cached mtimes, or finding a PHPStan setting to disable mtime validation).

PR #10400, opened 2026-01-26
fix(e2e): replace no-op wait() calls with proper WebDriver waits

### LINKED ISSUE #10399: fix(ci): E2E tests randomly fail with TimeoutException
## Summary

E2E tests in the `test-all.yml` workflow randomly fail on master but pass after retry. The failures are consistently `TimeoutException` errors when waiting for iframes or modal dialogs.

## Observed Failures

Recent examples from master branch:
- **Run 21360037444**: PHP 8.4/apache/mariadb, PHP 8.2/nginx/mariadb, PHP 8.4/apache/mysql
- **Run 21270782108**: PHP 8.2/nginx/mariadb, PHP 8.4/nginx/mariadb

Common failure patterns:
- `User add` test: TimeoutException at `UserAddTrait.php:101` waiting for modal to close
- `Encounter open` test: TimeoutException waiting for `fin` iframe within 30 seconds

## Root Causes

1. **Insufficient timeouts**: Default 30s implicit wait may not be enough under CI load
2. **Timing issues in test code**: Manual `sleep()` calls indicate known fragility
3. **Possible bug**: `UserAddTrait.php:145` has `$this->client->wait(1000000)` - WebDriver wait() takes milliseconds, so this would be ~16 minutes instead of the likely intended 1 second
4. **No retry mechanism**: CI has no retry logic for flaky E2E tests
5. **nginx configurations fail more frequently**: Different response times vs apache

## Proposed Fixes

- [ ] Fix the wait() call at line 145 (1000000 → 1000 if 1 second intended)
- [ ] Increase default E2E timeouts in CI (use coverage-enabled timeouts: 90s implicit, 180s page load)
- [ ] Add retry mechanism for E2E test step in CI workflow
- [ ] Replace implicit waits with explicit WebDriverWait where possible

kojiromike:
Fixes #10399

## Summary
- E2E tests randomly fail with TimeoutException because wait() calls without until() do nothing
- Replace no-op waits with explicit WebDriver conditions

## Changes
- Replace `$this->client->wait(1000000)` (no-op) with proper wait for Create User button to be clickable
- Replace manual retry loop in `setPasswordWithRetry()` with single wait until password field has expected value
- Remove unused `$userAddAttemptCounter` property and `ExpectationFailedException` import

## Test Plan
- [ ] E2E tests pass locally
- [ ] CI E2E tests pass (watch for fewer random failures over time)

## AI Disclosure
Yes

### DIFF
--- a/.phpstan/baseline/class.notFound.php
+++ b/.phpstan/baseline/class.notFound.php
@@ -91,15 +91,5 @@ $ignoreErrors[] = [
     'count' => 1,
     'path' => __DIR__ . '/../../src/Gacl/Gacl.php',
 ];
-$ignoreErrors[] = [
-    'message' => '#^Caught class OpenEMR\\\\Tests\\\\E2e\\\\User\\\\Exception not found\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../tests/Tests/E2e/BbCreateStaffTest.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Instantiated class OpenEMR\\\\Tests\\\\E2e\\\\User\\\\Exception not found\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../tests/Tests/E2e/BbCreateStaffTest.php',
-];
 
 return ['parameters' => ['ignoreErrors' => $ignoreErrors]];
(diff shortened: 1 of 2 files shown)


bradymiller:

PR #10404, opened 2026-01-26
fix(session): prevent session conflict for API requests with portal c…

### LINKED ISSUE #10416: fix(session): API requests with portal cookies cause session conflict error
## Bug Description

When an API request comes in with a portal App cookie, `SessionWrapperFactory` tries to start a portal session even though an API session is already active (started by `SiteSetupListener`), causing a "Failed to start the session: already started by PHP" error.

## Steps to Reproduce

1. Make an API request that includes a portal app cookie
2. Session conflict occurs because:
   - `SiteSetupListener` starts an API session
   - `SessionWrapperFactory::findSessionWrapper()` detects the portal cookie
   - It attempts to start a portal session, but PHP session is already active

## Expected Behavior

API requests should use their existing Symfony sessions regardless of any portal cookies present.

## Actual Behavior

Error: "Failed to start the session: already started by PHP"

## Related Files

- `src/Common/Session/PHPSessionWrapper.php`
- `src/Common/Session/SessionWrapperFactory.php`
- `src/RestControllers/Subscriber/SiteSetupListener.php`
- `interface/globals.php`

rollingventures:
Fixes #10416

When an API request comes in with a portal App cookie, `SessionWrapperFactory` tries to start a portal session even though an API session is already active (started by `SiteSetupListener`), causing a "Failed to start the session: already started by PHP" error.

### `src/Common/Session/PHPSessionWrapper.php`
- Added check in `__construct()` to detect API/OAuth requests before starting a core session
- If an API/OAuth request is detected, the constructor returns early without starting a session
- Prevents session conflicts when API sessions are already active

### `src/Common/Session/SessionWrapperFactory.php`
- Added check in `findSessionWrapper()` to detect API/OAuth requests before attempting to start a portal session
- If an API/OAuth request is detected, returns a `PHPSessionWrapper` instance without starting a new session
- Ensures API/OAuth requests use their existing Symfony sessions

## Technical Details

The fix uses `SessionUtil::API_WEBROOT` (`/apis/`) and `SessionUtil::OAUTH_WEBROOT` (`/oauth2/`) constants to detect API/OAuth requests by checking the `REQUEST_URI`. This prevents the session wrapper from attempting to start a new session when one is already active.

## Testing Considerations

- [x] API requests with portal cookies should no longer cause session conflicts
- [x] API requests without portal cookies should continue to work
- [x] Portal requests should continue to work normally
- [x] OAuth requests should continue to work normally
- [x] No session conflicts should occur in any scenario

## Related Files

- `src/Common/Session/SessionUtil.php` - Contains `API_WEBROOT` and `OAUTH_WEBROOT` constants
- `src/RestControllers/Subscriber/SiteSetupListener.php` - Starts API/OAuth sessions
- `interface/globals.php` - Calls `SessionWrapperFactory->getWrapper()`

### DIFF
--- a/src/Common/Session/PHPSessionWrapper.php
+++ b/src/Common/Session/PHPSessionWrapper.php
@@ -22,11 +22,15 @@ class PHPSessionWrapper implements SessionWrapperInterface
 {
     public function __construct()
     {
-        $globalsBag = OEGlobalsBag::getInstance();
-        $webroot = $globalsBag->get('webroot');
-        if ($webroot !== null && session_status() !== PHP_SESSION_ACTIVE) {
-            SessionUtil::coreSessionStart($webroot, false);
+        // Skip if session already active (e.g., API/OAuth context where Symfony session exists)
+        if (session_status() === PHP_SESSION_ACTIVE) {
+            return;
         }
+
+        $globalsBag = OEGlobalsBag::getInstance();
+        // Empty webroot is valid - means OpenEMR is at document root
+        $webroot = $globalsBag->get('webroot') ?? '';
+        SessionUtil::coreSessionStart($webroot, false);
     }
 
     public function getId(): string
--- a/src/Common/Session/SessionWrapperFactory.php
+++ b/src/Common/Session/SessionWrapperFactory.php
@@ -36,10 +36,13 @@ class SessionWrapperFactory
     private function findSessionWrapper(array $initData = []): SessionWrapperInterface
     {
         $app = SessionUtil::getAppCookie();
+
+        // Use PHPSessionWrapper for non-portal requests, or if a session is already active
+        // (e.g., API/OAuth requests where SiteSetupListener has already started a Symfony session)
         if (SessionUtil::isPredisSession()) {
             SessionUtil::portalPredisSessionStart();
             $session = new PHPSessionWrapper();
-        } else if ($app !== SessionUtil::PORTAL_SESSION_ID) {
+        } else if ($app !== SessionUtil::PORTAL_SESSION_ID || session_status() === PHP_SESSION_ACTIVE) {
             $session = new PHPSessionWrapper();
         } else {
             $session = new SymfonySessionWrapper(SessionUtil::portalSessionStart());
--- a/tests/Tests/Unit/Common/Session/SessionWrapperFactoryTest.php
+++ b/tests/Tests/Unit/Common/Session/SessionWrapperFactoryTest.php
@@ -1,0 +1,923 @@
+<?php
+
+/**
+ * SessionWrapperFactoryTest - Tests for session wrapper factory behavior
+ *
+ * Tests that the SessionWrapperFactory correctly handles session conflicts
+ * when sessions are already active (e.g., API/OAuth contexts) and various
+ * application cookie scenarios.
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    OpenCoreEMR <hello@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Tests\Unit\Common\Session;
+
+use OpenEMR\Common\Session\PHPSessionWrapper;
+use OpenEMR\Common\Session\SessionUtil;
+use OpenEMR\Common\Session\SessionWrapperFactory;
+use OpenEMR\Common\Session\SessionWrapperInterface;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+
+class SessionWrapperFactoryTest extends TestCase
+{
+    /**
+     * Store original superglobals to restore after each test
+     *
+     * @var array<mixed, mixed>
+     */
+    private array $originalCookie;
+
+    /**
+     * @var array<mixed, mixed>
+     */
+    private array $originalServer;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        // Store original superglobals
+        $this->originalCookie = $_COOKIE;
+        $this->originalServer = $_SERVER;
+
+        // Reset the singleton instance before each test
+        $this->resetSingleton();
+    }
+
+    protected function tearDown(): void
+    {
+        // Restore original superglobals
+        $_COOKIE = $this->originalCookie;
+        $_SERVER = $this->originalServer;
+
+        // Reset singleton after each test
+        $this->resetSingleton();
+
+        parent::tearDown();
+    }
+
+    /**
+     * Reset the SessionWrapperFactory singleton instance
+     */
+    private function resetSingleton(): void
+    {
+        $reflection = new ReflectionClass(SessionWrapperFactory::class);
+
+        // Reset the SingletonTrait's instances array
+        $instancesProperty = $reflection->getProperty('instances');
+        $instancesProperty->setValue(null, []);
+    }
+
+    /**
+     * Test that non-portal requests return PHPSessionWrapper
+     */
+    public function testNonPortalRequestReturnsPHPSessionWrapper(): void
+    {
+        // Set up non-portal cookie (core OpenEMR)
+        $_COOKIE[SessionUtil::APP_COOKIE_NAME] = SessionUtil::CORE_SESSION_ID;
+
+        $factory = SessionWrapperFactory::getInstance();
+
+        // Use reflection to call findSessionWrapper directly to avoid session issues
+        $reflection = new ReflectionClass($factory);
+        $method = $reflection->getMethod('findSessionWrapper');
+        $wrapper = $method->invoke($factory, []);
+
+        $this->assertInstanceOf(
+            PHPSessionWrapper::class,
+            $wrapper,
+            'Non-portal requests should return PHPSessionWrapper'
+        );
+    }
+
+    /**
+     * Test that when session is already active, PHPSessionWrapper is returned
+     * regardless of cookie value
+     *
+     * This is the key fix - prevents "session already started" errors when
+     * API/OAuth requests have a portal cookie present
+     */
+    public function testActiveSessionReturnsPHPSessionWrapper(): void
+    {
+        // Simulate portal cookie being present
+        $_COOKIE[SessionUtil::APP_COOKIE_NAME] = SessionUtil::PORTAL_SESSION_ID;
+
+        // Start a session to simulate API/OAuth having already started one
+        if (session_status() !== PHP_SESSION_ACTIVE) {
+            @session_start();
+        }
+
+        $this->assertEquals(
+            PHP_SESSION_ACTIVE,
+            session_status(),
+            'Session should be active for this test'
+        );
+
+        $factory = SessionWrapperFactory::getInstance();
+
+        // Use reflection to call findSessionWrapper directly
+        $reflection = new ReflectionClass($factory);
+        $method = $reflection->getMethod('findSessionWrapper');
+        $wrapper = $method->invoke($factory, []);
+
+        $this->assertInstanceOf(
+            PHPSessionWrapper::class,
+            $wrapper,
+            'When session is already active, should return PHPSessionWrapper to avoid conflicts'
+        );
+
+        // Clean up session
+        if (session_status() === PHP_SESSION_ACTIVE) {
+            session_write_close();
+        }
+    }
+
+    /**
+     * Test that API request URIs with subpath installations work correctly
+     *
+     * Addresses concern: Will this still work correctly if OpenEMR is installed
+     * at a subpath, like http://example.com/openemr/apis/foo
+     */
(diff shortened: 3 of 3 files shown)


kojiromike:
@zmilan for review

zmilan:
> @zmilan for review

This is good... I am sorry for not considering this when I was working on it. I didn't have experience with the API part or how to test it properly. But this is correct, it will start a regular wrapper around `$_SESSION`, and everything should work correctly.

zmilan:
@rollingventures Not related to these changes, but to my work on moving the Core session to Symfony Session...what would be the way to properly initiate an API session in some of these shared files? I see, based on your changes, how to check if it is an API request...

src/Common/Session/PHPSessionWrapper.php
```diff
@@ -22,6 +22,17 @@ class PHPSessionWrapper implements SessionWrapperInterface
 {
     public function __construct()
     {
+        // Check if we're in an API or OAuth request context
+        // If so, don't start a core session as API requests use Symfony sessions
+        $requestUri = $_SERVER['REQUEST_URI'] ?? '';
```
adunsulag: Wouldn't it make more sense just to remove the webroot check on line 38 here and just return false if the session is already started? We'll need to default webroot to empty string due to type safety further down the call chain, but we don't want sessions getting started if there's an already existing session anyways.

It also simplifies the logic here.

If there's a compelling reason why we don't want to do this, then this code looks fine.

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -35,6 +35,21 @@ public function getWrapper(array $initData = []): SessionWrapperInterface
...
+        // Check if we're already in an API/OAuth session context
+        $requestUri = $_SERVER['REQUEST_URI'] ?? '';
+        $isApiRequest = strpos($requestUri, SessionUtil::API_WEBROOT) !== false || 
+                        strpos($requestUri, SessionUtil::OAUTH_WEBROOT) !== false;
+        
+        if ($isApiRequest) {
+            // API/OAuth requests already have Symfony sessions active
+            // Return a PHPSessionWrapper that won't start a new session
+            $session = new PHPSessionWrapper();
+            foreach ($initData as $name => $value) {
```
adunsulag: This line is a duplicate of lines 48-50.  The if clause could be moved down to be part of the branching.

Line 39 could become:
if ($isApiRequest || $app !== SessionUtil::PORTAL_SESSION_ID) {

adunsulag: I'm working on the isPredisSession logic stuff anyways so will need to bring in your changes before my fixes go in.

adunsulag (changes requested):

src/Common/Session/PHPSessionWrapper.php
rollingventures: agreed...

- Removed the $webroot !== null check on line 38
- Added early return if a session is already active (prevents starting a new session)
- Defaulted $webroot to an empty string using the null coalescing operator (??) for type safety
- Simplified the logic: check API request → check if session active → start session with defaulted webroot

src/Common/Session/SessionWrapperFactory.php
rollingventures: removed, and doing actual testing with various use cases now.

- Consolidated duplicate code by combining the API request check with the non-portal check
- Changed line 45 to: if ($isApiRequest || $app !== SessionUtil::PORTAL_SESSION_ID)
- Moved the initData loop to after the branching logic, eliminating duplication between lines 47-49 and 63-65
- All three branches (API/non-portal, portal with Predis, portal with Symfony) now share the same initData setting logic

adunsulag: I'll be dropping the predis check with my changes once you've finished this.  Predis logic is being moved into SessionUtil to be part of the storage engine setup.  Current implementation breaks with predis.  I'm tempted to remove the predis capability anyways and deal with it in our subsequent session migration that zmilan is doing, but I'm close to a resolution so probably going to stick with this. Thanks for taking this on.

rollingventures:
@adunsulag @kojiromike - finishing many functional local session use case tests, php stan'sons and then this will be ready for final review.

rollingventures:
> > @zmilan for review
> 
> This is good... I am sorry for not considering this when I was working on it. I didn't have experience with the API part or how to test it properly. But this is correct, it will start a regular wrapper around `$_SESSION`, and everything should work correctly.

No problem - the wrapper abstraction you built made this an easy fix. Thanks for confirming!

rollingventures:
Included @adunsulag suggestions, and went with -> str_starts_with 
```
All 4 session types verified:
┌────────┬──────────────────────────────┬────────┐
│  Type  │          Cookie Set          │ Status │
├────────┼──────────────────────────────┼────────┤
│ API    │ apiOpenEMR only              │ ✅ OK  │
├────────┼──────────────────────────────┼────────┤
│ OAuth  │ authserverOpenEMR only       │ ✅ OK  │
├────────┼──────────────────────────────┼────────┤
│ Core   │ OpenEMR (+ App cookie)       │ ✅ OK  │
├────────┼──────────────────────────────┼────────┤
│ Portal │ PortalOpenEMR (+ App cookie) │ ✅ OK  │
└────────┴──────────────────────────────┴────────┘
 ```
 With some extra analysis:
 ```
  ┌──────────────┬───────────────────┬──────────┬─────────────────────────────────────────────────────────────────────┐
  │ Session Type │    Cookie Name    │   Path   │                          How It's Handled                           │
  ├──────────────┼───────────────────┼──────────┼─────────────────────────────────────────────────────────────────────┤
  │ Core UI      │ OpenEMR           │ /        │ PHPSessionWrapper starts session only if none active                │
  ├──────────────┼───────────────────┼──────────┼─────────────────────────────────────────────────────────────────────┤
  │ API          │ apiOpenEMR        │ /apis/   │ Detected by path → PHPSessionWrapper skips (session already active) │
  ├──────────────┼───────────────────┼──────────┼─────────────────────────────────────────────────────────────────────┤
  │ OAuth        │ authserverOpenEMR │ /oauth2/ │ Detected by path → PHPSessionWrapper skips (session already active) │
  ├──────────────┼───────────────────┼──────────┼─────────────────────────────────────────────────────────────────────┤
  │ Portal       │ PortalOpenEMR     │ /portal/ │ Detected by App cookie → SymfonySessionWrapper                      │
  ├──────────────┼───────────────────┼──────────┼─────────────────────────────────────────────────────────────────────┤
  │ Portal+Redis │ PortalOpenEMR     │ /portal/ │ Uses PHPSessionWrapper after Predis init                            │
  └──────────────┴───────────────────┴──────────┴─────────────────────────────────────────────────────────────────────┘
```

adunsulag (approved):
This looks fine to me with the changes.  If @kojiromike wants to take another look, but should be able to bring this in tomorrow morning.

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -35,11 +35,20 @@ public function getWrapper(array $initData = []): SessionWrapperInterface
 
     private function findSessionWrapper(array $initData = []): SessionWrapperInterface
     {
+        // Check if we're already in an API/OAuth session context
+        $requestUri = $_SERVER['REQUEST_URI'] ?? '';
+        $isApiRequest = str_starts_with((string) $requestUri, SessionUtil::API_WEBROOT) ||
```
kojiromike: Will this still work correctly if openemr is installed at a subpath, like `http://example.com/openemr/apis/foo`

src/Common/Session/PHPSessionWrapper.php
```diff
@@ -22,11 +22,14 @@ class PHPSessionWrapper implements SessionWrapperInterface
...
-        if ($webroot !== null && session_status() !== PHP_SESSION_ACTIVE) {
-            SessionUtil::coreSessionStart($webroot, false);
+        // Don't start a session if one is already active
+        if (session_status() === PHP_SESSION_ACTIVE) {
+            return;
         }
+
+        $globalsBag = OEGlobalsBag::getInstance();
+        $webroot = $globalsBag->get('webroot') ?? '';
+        SessionUtil::coreSessionStart($webroot, false);
```
kojiromike, line 33: Does coreSessionStart do the right thing if `$webroot === ''`/ if the global webroot was not properly initialized, and is null?

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -35,11 +35,20 @@ public function getWrapper(array $initData = []): SessionWrapperInterface
...
+                        str_starts_with((string) $requestUri, SessionUtil::OAUTH_WEBROOT);
+
         $app = SessionUtil::getAppCookie();
-        if (SessionUtil::isPredisSession()) {
-            SessionUtil::portalPredisSessionStart();
+
+        if ($isApiRequest || $app !== SessionUtil::PORTAL_SESSION_ID) {
```
kojiromike: This would be broken if API requests used predis. IIUC you did this because they don't, but please comment the code so it's clear why this order of precedence is ok.

src/Common/Session/SessionWrapperFactory.php
```diff
@@ -35,11 +35,20 @@ public function getWrapper(array $initData = []): SessionWrapperInterface
...
+            // API/OAuth requests already have Symfony sessions active
+            // Non-portal requests use PHPSessionWrapper
+            // Return a PHPSessionWrapper that won't start a new session
             $session = new PHPSessionWrapper();
-        } else if ($app !== SessionUtil::PORTAL_SESSION_ID) {
```
kojiromike, line 42: 🤔 could the entire change actually just boil down to adding `|| session_status() === PHP_SESSION_ACTIVE` to this `else if`?

rollingventures: Yes - much cleaner than path detection.  Modified, thank you.

src/Common/Session/SessionWrapperFactory.php
rollingventures: This is now moot - I removed the path detection entirely and went with your session_status() suggestion instead. The predis check remains first as before.

src/Common/Session/PHPSessionWrapper.php
rollingventures: Yes - empty webroot is valid and means OpenEMR is installed at the document root. The cookie path defaults to / in that case. Added a comment clarifying this.

src/Common/Session/SessionWrapperFactory.php
rollingventures: Removed the path detection entirely. The fix now relies solely on session_status() === PHP_SESSION_ACTIVE, which is path-agnostic. Added tests for root, subpath (/openemr/apis/), and deep subpath (/health/systems/openemr/apis/) installations - all pass.

rollingventures:
Also added tests, and will conduct some more custom edge cases before this is ready to merge pending ci agrees also.
```
Tests added (SessionWrapperFactoryTest.php - 20 tests):
┌─────────────────┬────────────────────────────────────────────────────────┐
│    Category     │                        Coverage                        │
├─────────────────┼────────────────────────────────────────────────────────┤
│ Session Types   │ Core, API, OAuth, Portal                               │
├─────────────────┼────────────────────────────────────────────────────────┤
│ Shared Files    │ Portal cookie + API/OAuth session, module entry points │
├─────────────────┼────────────────────────────────────────────────────────┤
│ Path Variations │ Root, subpath, deep subpath, empty/missing URI         │
├─────────────────┼────────────────────────────────────────────────────────┤
│ Edge Cases      │ Missing cookie, cached wrapper, init data              │
└─────────────────┴────────────────────────────────────────────────────────┘
OK (20 tests, 21 assertions)
Full suite: 425 tests ✅
```

rollingventures:
@adunsulag -> Predis edge case - If isPredisSession() is true AND a session is already active,portalPredisSessionStart() gets called before our check. Could conflict in Redis environments with the same portal-cookie-during-API scenario. Likely rare but worth noting.  I know you did mention you were working on this logic so this PR is not blocked.

Also a potential "Portal" behavior change that I am researching today - If session is already active with portal cookie, we now return PHPSessionWrapper instead of SymfonySessionWrapper.

Testing approach:

Test: 1. Portal smoke test

How: Log into patient portal, test messaging, signatures, payments - verify isSymfonySession() -> returns true

Test: 2. API + portal cookie

How: Open portal in tab 1, make API call in tab 2, verify no session conflict

Test: 3. Check entry points

How: Verify no portal files start sessions before calling SessionWrapperFactory

rollingventures:
Manual test confirmed:

- ✅ Portal login works normally
- ✅ API + portal cookie: returns 401 (expected), no session conflict
- ✅ Without fix: 500 - "Failed to start the session: already started by PHP."
- ✅ With fix: Clean 401, no session errors

29 unit tests covering all session types, shared file patterns, and path variations.

adunsulag:
@rollingventures thanks, there's a number of problems in the predis stuff I'm working on.  I may just disable the predis altogether if I don't have a solution in the next hour or two.  I appreciate you getting this in.

PR #10326, opened 2026-01-22
refactor(validation): consolidate phone number handling using PhoneNumberService

kojiromike:
## Summary
- Add `giggsey/libphonenumber-for-php` dependency for robust phone validation
- Add `ValidationUtils::isValidPhoneNumber()`, `toE164()`, and `toNationalDigits()` using libphonenumber
- Support strict mode (validates real area codes) and non-strict mode (for test data)
- Add `PhoneType` enum with translatable labels for phone number types
- Add `PhoneNumber` wrapper class combining libphonenumber with `PhoneType`
- Enhance `PhoneNumberService` with `parsePhone()`, `formatPhone()`, and `toHL7Phone()` methods
- Migrate `Pharmacy` and `InsuranceCompany` classes to use the service and enum
- Consolidate scattered phone validation/formatting across the codebase:
  - `ContactTelecomService` now uses `ValidationUtils::isValidPhoneNumber()`
  - `Claim.php` uses `ValidationUtils::toNationalDigits()` for billing
  - `CdaComponentParseHelpers` uses `ValidationUtils::toNationalDigits()`
  - `immunization_report.php` uses `PhoneNumberService::toHL7Phone()`
  - `ImmunizationController` uses `PhoneNumberService::toHL7Phone()`
  - `BillingExport.csv.php` and `DocumentTemplateRender` use `PhoneNumberService::formatPhone()`
  - `orders_results.php` uses `PhoneNumberService::formatPhone()`
- Remove legacy `PhoneNumber.class.php` (replaced by `PhoneNumberService`)
- Remove scattered local `format_phone()` and `formatPhone()` functions

Fixes #10309

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #10241, opened 2026-01-16
feat(ci): Add rule to forbid duplicate bare functions


Firehed:
If we ever end up in a path at runtime where a function gets redefined, it will trigger a fatal error. It can _also_ trip up CI (#10237, etc) but that's more of a secondary concern.

Since this WILL result in errors getting caught (I know of several just in a very limited surface area) I think this should entail getting any actual duplicates moved into an autoloaded `functions.php` or equivalent file. Anything with divergence should rename them before doing the same.

Firehed:
After a quick pass, this is trickier to deploy than I initially expected: PHPStan analyzes in parallel, so the collector is only reliable when run with a flag that disable analysis parallelism (`--debug`). Even without that I still get 75 errors. 148 with a flag passed (which counts the duplicates but not the originals; the total number of problems is higher in reality).

I think a more prudent approach would be to block naked function definitions outright outside of an allowed directory, and ensure the entire contents of that directory are on the autoload path.

A working implementation of the original fix is available [here](https://github.com/openemr/openemr/compare/master...Firehed:openemr:10241/dupe-fn-detect).

kojiromike:
@Firehed what's the feasibility of converting some of the existing cases into classes? Even if they're just static classes, that'd provide a degree of pseudo namespacing that'd probably make phpstan happy and inch us closer to eliminating the problem.

Firehed:
Probably tractable but ugly (esp. with rector), assuming they're not _overly_ reliant on globals (I know several that are). However I worry that approach will do more to hide the duplicated code, so I'm not sure if it actually ends up as a net win even in the short term. I'll stew on the idea for a bit to think through the trade-offs.

PR #10405, opened 2026-01-26
refactor: Consolidate more duplicate functions/symbols

Firehed:
Part of https://github.com/openemr/openemr/issues/10241, but does not complete it.

Consolidates duplicate function definitions from multiple files into library/global_functions.inc.php, reducing code duplication and ensuring consistent behavior across the codebase.

Note: hl7phone changes _not_ updated to the tooling in #10326, intentionally. Despite the function names, the existing code doesn't use the `^`-based formatting provided by the new `toHL7Phone` and related variants. That might mean there's a bug in the existing code; I have no idea. This is intended as a pure refactor without behavioral changes, so if the existing code was out of spec, it will remain as such for now.

  12 functions consolidated:
  - hl7Priority, hl7Text, hl7Zip, hl7Sex — identical implementations merged
  - hl7Time, hl7Phone, hl7SSN — added boolean parameters to handle lab-specific formatting differences (LabCorp uses digits-only format)
  - hl7RelationWord, hl7RelationCode — split into two functions to handle word vs numeric code output formats
  - tr, ucname, issueTypeIndex — utility functions consolidated

  Files cleaned up:
  - interface/orders/gen_hl7_order.inc.php
  - interface/procedure_tools/quest/gen_hl7_order.inc.php
  - interface/procedure_tools/labcorp/gen_hl7_order.inc.php
  - interface/procedure_tools/gen_universal_hl7/gen_hl7_order.inc.php
  - interface/forms/eye_mag/save.php
  - interface/forms/procedure_order/common.php
  - interface/orders/receive_hl7_results.inc.php
  - interface/patient_file/summary/add_edit_issue.php
  - interface/reports/immunization_report.php
  - interface/reports/non_reported.php

  Notes

  - New relation functions use PHP 8.2+ match expressions
  - Lab-specific behavior preserved via named parameters (e.g., hl7Phone($val, formatted: false) for LabCorp)
  - I guided the AI _away_ from a per-provider HL7 interface/implementation set in order to keep the change smaller and easier to review. In the future as we presumably push this more towards modules, adopting such an approach is preferable.

Yes

Most work done by Claude. Inspected the changeset for accuracy.

### DIFF
--- a/interface/orders/gen_hl7_order.inc.php
+++ b/interface/orders/gen_hl7_order.inc.php
@@ -34,8 +34,6 @@ use OpenEMR\Common\Logging\EventAuditLogger;
 use OpenEMR\Common\Orders\Hl7OrderGenerationException;
 use OpenEMR\Common\Orders\Hl7OrderResult;
 
-// hl7Time, hl7Phone, hl7SSN, hl7RelationWord, hl7RelationCode are defined in library/global_functions.inc.php
-
 /**
  * Get array of insurance payers for the specified patient as of the specified
  * date. If no date is passed then the current date is used.
--- a/interface/procedure_tools/gen_universal_hl7/gen_hl7_order.inc.php
+++ b/interface/procedure_tools/gen_universal_hl7/gen_hl7_order.inc.php
@@ -33,8 +33,6 @@ use OpenEMR\Common\Logging\EventAuditLogger;
 use OpenEMR\Common\Orders\Hl7OrderGenerationException;
 use OpenEMR\Common\Orders\Hl7OrderResult;
 
-// hl7Time, hl7Phone, hl7SSN, hl7RelationWord, hl7RelationCode are defined in library/global_functions.inc.php
-
 /**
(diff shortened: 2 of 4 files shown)


interface/reports/immunization_report.php
```diff
@@ -44,11 +44,6 @@
 $form_from_date = (isset($_POST['form_from_date'])) ? DateToYYYYMMDD($_POST['form_from_date']) : '';
 $form_to_date = (isset($_POST['form_to_date'])) ? DateToYYYYMMDD($_POST['form_to_date']) : '';
 
-function tr($a)
```
kojiromike, line 48: if only https://www.php.net/manual/en/function.strtr.php were a thing

Firehed: if only. It's somehow worse when you look how it's actually used.

kojiromike:
@Firehed regarding the phone thing see https://github.com/openemr/openemr/issues/10423

bradymiller:

Issue #10407, opened 2026-01-27
bug: Duplicate `sms` classes, both of which seem broken


Firehed:
While working on #10241, I ran across two different definitions of the `sms` class:

- modules/sms_email_reminder/sms_clickatell.php
- modules/sms_email_reminder/sms_tmb4.php

The way they're included means it shouldn't _crash_ at runtime, but the call site (`modules/sms_email_reminder/cron_sms_notification.php`, evidently used by `cron_SendSMS`) doesn't appear to use _either one_ correctly: both have differing `__construct` methods, neither of which match the actual instantiation of the class.

Given that I have to assume this path does not work and should be safe to remove entirely, but if anyone knows more I'd love to hear about it!

Firehed:
Looking further, the entire system within this directory seems to be on somewhat shaky footing, and despite the directory structure exists outside of the primary module system. Do we have any data as to whether current installations are actively using this at all?

sjpadgett:
I have considered removing batchcom which is where this is used. The current clickatell api is broken. We had a developer do  a PR to fix but I had him put in faxsms module which is where I want communications to be.

So do we get rid of or fix? I say get rid of but I may get pushback. @openemr/admins

Firehed:
Anything broken and unmaintained I'd be strongly in favor of removal, especially if there's a better replacement already in place which sounds like the case here.

PR #10408, opened 2026-01-27
refactor(sms): Split conflicting class name definitions

Firehed:
Progress towards #10241.

There were two `sms` classes, which can break analysis tooling.

- Renamed the two classes
- Updated the only instantiation point I could find to account for the rename.

The new `else` case is a _slightly different_ failure mode than before, but not a new one.

Note: I strongly suspect this is broken in practice (see #10407), but in the meantime, I'm just focused on making progress towards #10241 and this is the path of least resistance.

No

### DIFF
--- a/modules/sms_email_reminder/cron_sms_notification.php
+++ b/modules/sms_email_reminder/cron_sms_notification.php
@@ -41,7 +41,6 @@ $check_date = date("Y-m-d", mktime(date("h") + $SMS_NOTIFICATION_HOUR, 0, 0, dat
 // it's content - to do latter
 $db_email_msg = cron_getNotificationData($TYPE);
 
-// object for sms
 // get notification settings
 $vectNotificationSettings = cron_GetNotificationSettings();
 $SMS_GATEWAY_USENAME = $vectNotificationSettings['SMS_gateway_username'];


modules/sms_email_reminder/cron_sms_notification.php
```diff
@@ -60,7 +54,16 @@
 //echo "\nDEBUG :: user=".$vectNotificationSettings['SMS_gateway_username']."\n";
 
 // create sms object
-$mysms = new sms($SMS_GATEWAY_USENAME, $SMS_GATEWAY_PASSWORD, $SMS_GATEWAY_APIKEY);
+include_once("sms_clickatell.php");
+include_once("sms_tmb4.php");
+if ($db_email_msg['sms_gateway_type'] == 'CLICKATELL') {
```
kojiromike: switch/case?

Firehed: good call, fixed a global along the way.

modules/sms_email_reminder/cron_sms_notification.php
```diff
@@ -42,13 +42,6 @@
 $db_email_msg = cron_getNotificationData($TYPE);
 
 // object for sms
```
kojiromike:
```suggestion
```

bradymiller:

kojiromike:
@bradymiller apparently not backporting this was a mistake, although we didn't know it at the time. Before this change phpstan could not figure out whether the SMS constructor needed three parameters or zero. So right now even though it looks like phpstan is passing on rel-800 it's actually wrong. We need to bring this into the patch this so that we can have consistently passing checks in the various other things we need to bring in there.

Issue #9719, opened 2025-12-09
Move table of encounters/charges into Twig template


Firehed:
Part of #9705.

PR #10420, opened 2026-01-28
refactor(payments): Streamline table/form in UIs

Firehed:
Fixers #9719 (though not as originally intended)

Refactors the encounter and payment table shown in the main app and patient portal to avoid globals and solve a function name collision

- Removes the two (conflicting) `echoLine` functions
- Inlines the HTML used to render the table, better matching the rest of the file
- Removes a bunch of string concatenation which IMO increases readability substantially
- Adds a couple selectors and data to the HTML that will be used in some upcoming related refactors

Unlike #9719, I ended up _not_ moving this into a Twig template; there are some subtle differences in the UIs that would require making two different-but-similar templates, which negates a lot of the gain in moving to a template in the first place. In the long run I think this IS still the right direction, but for the sake of making incremental progress I'm going to hold off on that.

No

#### Test plan
I diffed the HTML (as displayed in devtools) before and after in both screens to ensure all existing values were preserved. There are a few additions, intentionally, so it's not byte-for-byte identical but no existing values changed.

### DIFF
--- a/portal/portal_payment.php
+++ b/portal/portal_payment.php
@@ -942,10 +942,9 @@ if (($_POST['form_save'] ?? null) || ($_REQUEST['receipt'] ?? null)) {
                     $duept = $brow['amount'] + $srow['amount'] - $drow['payments'] - $drow['adjustments'];
                 }
 
+                $balance = 0;
                 if ($duept == 0) {
                     $balance = $value['charges'] - (float)$dpayment_pat - $dpayment - $dadjustment;
-                } else {
-                    $balance = 0;
                 }
 
                 // Update running totals before rendering the table row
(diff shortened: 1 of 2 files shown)


portal/portal_payment.php
```diff
@@ -964,8 +942,47 @@ function getAuth() {
                     $duept = $brow['amount'] + $srow['amount'] - $drow['payments'] - $drow['adjustments'];
                 }
 
-                echoLine("form_upay[$enc]", $dispdate, $value['charges'], $dpayment_pat, ($dpayment + $dadjustment), $duept, ($enc . ': ' . $reason), $inscopay, $patcopay);
-            }
+                $balance = $value['charges'] - (float)$dpayment_pat - $dpayment - $dadjustment;
+                if ($duept != 0) {
+                    $balance = 0;
+                }
```
kojiromike, line 948: Why do the math if you don't have to? This pattern occurs a couple of times, too.

Firehed: Mostly a hold-over from the previous logic. Updated.

bradymiller:

stephenwaite:
Hi @Firehed, thanks for the improvements! Just curious why the functions weren't renamed or refactored/consolidated into someplace like `src/Billing` or a new namespace?

PR #10426, opened 2026-01-28
feat(phpstan): Disallow global keyword

### LINKED ISSUE #10424: feat(phpstan): Forbid global keyword
Similar to the rule that forbids access to `$GLOBALS`, we'd benefit from blocking the `global` keyword for basically the same reason.

Firehed:
Fixes #10424

Discourage use of `global` in new code through a PHPStan rule. Variables should be passed in to functions explicitly (by-reference if needed)

- Rule added
- Existing use of the `global` keyword added to a new baseline file

Claude generated most of the rule, reviewed and tweaked by me.


bradymiller:

PR #10432, opened 2026-01-29
Fix/calendar in office events

ophthal:
## Summary
Fixes two critical UX issues in the calendar appointment scheduler that prevented proper use of "In Office" and "Out of Office" provider availability blocks.

### 1. In Office events required "All Day" selection
**Problem**: Providers could not create In Office availability blocks for specific time ranges (e.g., 9am-5pm). The duration field validation forced users to either:
- Mark the event as "All Day" (1440 minutes), OR  
- Leave duration empty (causing validation errors)

**Root Cause**: The duration field was enabled for In Office events and required validation, even though In Office blocks don't need durations (they define availability windows, not appointment lengths).

**Fix**:
- Disabled the duration field when In Office category is selected (line 1060)
- Cleared duration value to prevent validation issues (line 1061)
- Updated validation logic to skip duration requirements for In Office events (line 1897)

### 2. AM/PM selector disabled for provider events
**Problem**: When creating In Office or Out of Office events with the time selector enabled, the AM/PM dropdown was incorrectly disabled. This forced users to enter times in 24-hour military format.

**Root Cause**: Line 1115 incorrectly disabled the AM/PM selector based on \`durationDisabled\` instead of \`timeDisabled\`.

**Fix**: Changed \`f.form_ampm.disabled = durationDisabled\` to \`f.form_ampm.disabled = timeDisabled\` so AM/PM is available whenever time selection is enabled.

## Technical Details

**Files Changed**: 
- \`interface/main/calendar/add_edit_event.php\`

**Changes**:
1. Added constants for category IDs (\`OUT_OF_OFFICE_CAT_ID\`, \`OFFICE_VISIT_CAT_ID\`) for code clarity
2. Modified \`set_display()\` function to disable duration field for In Office events (lines 1060-1061)
3. Modified \`set_allday()\` function to use correct disable flag for AM/PM selector (line 1117)
4. Updated \`validateform()\` to skip duration validation for In Office category (lines 1872, 1899)

## Testing
- ✅ In Office events can now be created with specific time ranges (e.g., 9:00 AM - 5:00 PM)
- ✅ AM/PM selector works correctly when creating In Office/Out of Office events
- ✅ Duration field properly disabled for In Office events
- ✅ Validation passes without requiring "All Day" selection
- ✅ Existing "All Day" In Office events still work correctly
- ✅ Regular appointment types unaffected

## User Impact
This fix enables the proper workflow for provider schedule management:
1. Providers can define their available hours using In Office blocks with specific times
2. Preferred categories can be assigned to time blocks (e.g., reserve 9-10am for New Patients)
3. Time entry uses familiar AM/PM format instead of requiring 24-hour time

## Related
This addresses similar issues to upstream commit f1fa98775, adapted for current codebase.

---

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/interface/main/calendar/add_edit_event.php
+++ b/interface/main/calendar/add_edit_event.php
@@ -934,6 +934,8 @@ if ($groupid) {
  var durations = new Array();
 
  const IN_OFFICE_CAT_ID = '2';
+ const OUT_OF_OFFICE_CAT_ID = '3';
+ const OFFICE_VISIT_CAT_ID = '5';
 <?php
  // Read the event categories, generate their options list, and get
  // the default event duration from them if this is a new event.
@@ -1112,7 +1114,7 @@ function set_allday() {
     f.form_hour.disabled = timeDisabled;
     f.form_minute.disabled = timeDisabled;
     <?php if ($GLOBALS['time_display_format'] == 1) { ?>
-        f.form_ampm.disabled = durationDisabled;
+        f.form_ampm.disabled = timeDisabled;
     <?php } ?>
     f.form_duration.disabled = durationDisabled;
 }


interface/main/calendar/add_edit_event.php
kojiromike, line 1: Please don't change the file mode. This file should not be made executable. It should remain `100644`.

interface/main/calendar/add_edit_event.php
```diff
@@ -934,6 +934,8 @@ function setEventDate($start_date, $recurrence)
  var durations = new Array();
 
  const IN_OFFICE_CAT_ID = '2';
+ const OUT_OF_OFFICE_CAT_ID = '3';
+ const OFFICE_VISIT_CAT_ID = '5';
```
kojiromike, line 938: These constants are not referenced anywhere in the file. Please remove them to avoid dead code.

interface/main/calendar/add_edit_event.php
```diff
@@ -1893,7 +1896,7 @@ function validateform(event,value){
         }
     };
 
-    if ( allDay == true) {
+    if ( allDay == true || isInOffice == true) {
```
kojiromike, line 1899: Nit: `== true` is redundant in JS. This could be `if (allDay || isInOffice)`. Not a blocker since the existing code already used this style.

kojiromike, line 1899:
```suggestion
    if (allDay || isInOffice) {
```

PR #10433, opened 2026-01-29
refactor: remove more duplicate function definitions

Firehed:
More progress towards #10241

Removes more duplicate function definitions

Collapses duplicate definitions of the following functions:

- `getListItem`
- `myCellText`

Based on the detection script I'm using (which will land in a final fix, it can't be baselined) this further drops the number of duplicate symbols to 46. We're about halfway there!

PHPStan baseline errors are mostly trades, not additions.

I'll also note that `myCellText` seems to be a flawed workaround for a styling issue; adding `&nbsp;` into `<td>` tags should not be required. I'll open a separate issue to investigate that further. It should be removed entirely, but I'm trying to avoid scope-creeping.

~No~ Yes, trivial refactor by Claude.

### DIFF
--- a/.phpstan/baseline/openemr.deprecatedSqlFunction.php
+++ b/.phpstan/baseline/openemr.deprecatedSqlFunction.php
@@ -5712,11 +5712,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../library/global_functions.inc.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Use QueryUtils\\:\\:querySingleRow\\(\\) or QueryUtils\\:\\:fetchRecords\\(\\) instead of sqlQuery\\(\\)\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../library/global_functions.inc.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Use QueryUtils\\:\\:sqlStatementThrowException\\(\\) or QueryUtils\\:\\:fetchRecords\\(\\) instead of sqlStatement\\(\\)\\.$#',
     'count' => 2,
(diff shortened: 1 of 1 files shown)


library/global_functions.inc.php
```diff
@@ -616,3 +616,41 @@ function cbcell($name, $desc, $colname): string
...
+ * Adapt text to be suitable as the contents of a table cell.
+ *
+ * @param  string $s Input text.
+ * @return string  Output text.
+ */
+function myCellText($s)
+{
+    $s = trim($s ?? '');
+    if ($s === '') {
+        return '&nbsp;';
```
kojiromike, line 632: I don't understand this. Why would we replace the empty string with a non-breaking space and call it suitable for a table cell?

Firehed: Your guess is as good as mine, see the note in the PR body.

library/global_functions.inc.php
```diff
@@ -616,3 +616,41 @@ function cbcell($name, $desc, $colname): string
...
+ *
+ * @param  string $listid List identifier.
+ * @param  string $value List item identifier.
+ * @return string  The item's title.
+ */
+function getListItem($listid, $value)
+{
+    $lrow = sqlQuery(
+        "SELECT title FROM list_options " .
+        "WHERE list_id = ? AND option_id = ? AND activity = 1",
```
kojiromike:
```suggestion
        "WHERE list_id = ? AND option_id = ? AND activity = 1 LIMIT 1",
```

?

Also, what's stopping you from using database utils here? I think there's even a method for returning a single row in there.

Firehed: Happy to change it; all of these initial passes are _purely_ cut and paste, warts and all.

bradymiller:

PR #10437, opened 2026-01-29
fix: #10436 redis session storage

### LINKED ISSUE #10436: bug: Fix sessions with redis cluster session handler
### Describe the bug

When running the system using a redis backed cluster using the predis storage engine, the system sessions were failing on the portal. 

### Client configuration

- OpenEMR version: OpenEMR 8.0.0

adunsulag:
Added a development-easy-redis option that has redis setup in a cluster environment that can be used for testing.  Debated on putting it in the devops repo but seems like this works for now.

Changed up the storage engine logic to make this work.

Fixes #10436 by moving the logic for determining a redis session into the SessionUtil class to make the session backend storage agnostic of what kind of session it is.

The following areas are being addressed in this PR
- [x] predis session in portal
- [x] predis session in oauth GUI
- [x] predis session in Local API (best case of seeing this is editing insurance from patient demographics which uses local api)
- [x] predis session in OAuth api
- [x] predis session on multi-site setup

### DIFF
--- a/src/Common/Session/SessionUtil.php
+++ b/src/Common/Session/SessionUtil.php
@@ -91,7 +91,7 @@ class SessionUtil
 
     private static array $SESSION_INSTANCES = [];
 
-    private static SessionHandlerInterface $sessionHandler;
+    private static ?SessionHandlerInterface $sessionHandler;
 
     // Following setting have been deprecated in PHP 8.4 and higher
     // (ie. will remove them when PHP 8.4 is the minimum requirement)


adunsulag:
With the introduction of the SessionWrapper singleton the local api does not have access to the PredisSentinel configuration and so set_save_handler is being called twice and failing (once by HttpSessionFactory) and once by SessionWrapperFactory.  Will need to make PredisSentinel a singleton or figure out the branching logic to avoid the duplicate call.

bradymiller:
2DV

adunsulag:
Just another progress update.  I've got the local api working and multi-site working.  However, I'm seeing issues on a webroot (such as openemr) now.  Need to verify if this is just the predis stuff or if webroot issues are on the standard portal.  Have to step away and will back in a few hours to this.

adunsulag:
Webroot resolved with PR: #10455 so will need to merge that one in before this.

Was hoping these changes would resolve the EHR launch w/o needing to login again as they are related, but it didn't resolve the issue.  Will work on it more tomorrow.

bradymiller:
nice stuff!
brought in the other PR: #10455

src/Common/Session/Predis/PredisSessionHandler.php
```diff
@@ -20,6 +20,9 @@
 use Predis\Client;
 use SessionHandlerInterface;
 
+/**
+ * @deprecated 8.0.0 No longer used by internal code use Symfony\Component\HttpFoundation\Session\Storage\Handler\RedisSessionHandler instead of this class
```
bradymiller, line 24: feel free to remove file if not being used :)

src/Common/Session/SessionUtil.php
```diff
@@ -89,14 +91,28 @@ class SessionUtil
 
     private static array $SESSION_INSTANCES = [];
 
+    private static SessionHandlerInterface $sessionHandler;
```
kojiromike: **Bug -- causes the 4 test failures.** This property is declared as non-nullable `SessionHandlerInterface`, but `getSessionHandler()` assigns `null` when not in predis mode (line 351 in the new code). This is the root cause of:

```
TypeError: Cannot assign null to property OpenEMR\Common\Session\SessionUtil::$sessionHandler of type SessionHandlerInterface
```

Should be:
```php
private static ?SessionHandlerInterface $sessionHandler = null;
```

Then the `isset()` check in `getSessionHandler()` needs adjustment since `isset()` on a null-initialized property returns false. You'd need a separate `$sessionHandlerInitialized` bool, or restructure to use a sentinel pattern.

src/Common/Session/Predis/SentinelUtil.php
```diff
@@ -143,7 +143,12 @@ public function __construct(private readonly int $ttl, private readonly ?LoggerI
         ]);
     }
 
-    public function configure(): \SessionHandlerInterface
+    public static function getRedisSessionStorage(): RedisSessionHandler
```
kojiromike: This method is added but never called anywhere in this PR. Dead code -- should be removed, or deferred to a PR that actually uses it.

Also, it doesn't accept a TTL parameter, so it would use Symfony's default (which may differ from the `DEFAULT_GC_MAXLIFETIME` of 14400 used everywhere else).

src/Common/Session/Predis/SentinelUtil.php
```diff
@@ -212,16 +217,11 @@ public function configure(): \SessionHandlerInterface
...
-        if (!$success) {
-            $this->logger->errorLogCaller("Failed to set session handler for Predis Sentinel.");
-            throw new \Exception("Failed to set session handler for Predis Sentinel.");
-        }
-        $this->logger->debug("Successfully set session handler for Predis Sentinel.");
-        return $handler;
+        return new Client($sentinelParameters, $options);
+    }
+
+    public function configure(int $ttl): \SessionHandlerInterface {
```
kojiromike: `self::configureClient()` on a non-static method is misleading. `configureClient()` accesses `$this->` properties so it can't actually be called statically. This works here because PHP resolves `self::` to the current instance in a non-static context, but should be `$this->configureClient()`.

src/Common/Session/Predis/SentinelUtil.php
```diff
@@ -212,16 +217,11 @@ public function configure(): \SessionHandlerInterface
...
-        $success = session_set_save_handler($handler, true);
-        if (!$success) {
-            $this->logger->errorLogCaller("Failed to set session handler for Predis Sentinel.");
-            throw new \Exception("Failed to set session handler for Predis Sentinel.");
-        }
-        $this->logger->debug("Successfully set session handler for Predis Sentinel.");
-        return $handler;
+        return new Client($sentinelParameters, $options);
+    }
+
```
kojiromike: Nit: opening `{` on same line doesn't match the rest of the file/codebase convention (opening brace on new line for methods). Same issue in `SessionUtil.php` on `getSessionStorage()` and `getSessionHandler()`.

src/Common/Session/Predis/PredisSessionHandler.php
kojiromike: Per Brady's inline comment -- since this class is no longer used (replaced by Symfony's `RedisSessionHandler`), consider deleting the file entirely rather than just deprecating.

bradymiller:
hi @adunsulag , Does this PR need a rebase to get it to pass testing?

adunsulag:
@bradymiller rebased this, feel free to pull it in if all the tests pass as I'm done for the day.

PR #10438, opened 2026-01-30
fix(reports): prevent inflated totals in Financial Summary by Service Code

### LINKED ISSUE #8651: Financial Summary by Service Code report incorrect totals

kojiromike:
Fixes #8651

The Financial Summary by Service Code report produces inflated totals (billed, paid, adjusted, balance) when the `codes` table contains multiple rows for the same code value with different modifiers.

**Primary fix — Cartesian product on codes JOIN:**

The `LEFT OUTER JOIN codes AS c ON c.code = b.code` matches every modifier variant, creating a Cartesian product that multiplies all SUM aggregates. For example, if code `99213` has 3 modifier rows in `codes`, every billing row for that code gets tripled.

Replace the direct join with a subquery that collapses modifier variants via `GROUP BY code, code_type`, and join on both `code` and `code_type` (through `code_types.ct_id`):

```sql
-- Before (buggy):
LEFT OUTER JOIN codes AS c ON c.code = b.code

-- After (fixed):
INNER JOIN code_types AS ct ON ct.ct_key = b.code_type AND ct.ct_fee = '1'
LEFT OUTER JOIN (
    SELECT code, code_type, MAX(financial_reporting) AS financial_reporting
    FROM codes
    GROUP BY code, code_type
) AS c ON c.code = b.code AND c.code_type = ct.ct_id
```

**Additional fixes to pre-existing query bugs:**

- `c.financial_reporting` was in SELECT without aggregation or GROUP BY — non-deterministic when a code maps to multiple code types. Wrapped in `MAX()`.
- `ORDER BY fe.date, fe.id` without those columns in GROUP BY — non-deterministic. Simplified to `ORDER BY b.code`.
- "Important Codes" filter used `WHERE c.financial_reporting = '1'` which excludes billing rows before aggregation. Moved to `HAVING MAX(c.financial_reporting) = '1'` so all billing amounts are included for codes that have any financial_reporting flag.
- `Balance` column redundantly recomputed all three SUMs in SQL. Now computed in the DTO from `billed - paid - adjusted`.

**Refactoring:**

- Extract query logic into `FinancialSummaryReportService` with a `ServiceCodeSummary` readonly DTO
- Service accepts `DateTimeInterface` for date params, `?int` for facility/provider (null = all)
- Report file handles only rendering; type conversion from POST strings happens at the boundary

**Tests:**

- 7 integration tests: single code, duplicate modifiers (core bug), financial reporting filter, MAX across modifiers, facility filter, provider filter, LEFT JOIN null behavior
- 4 e2e browser tests: page load, empty submit, correct totals, important codes filter, duplicate modifier non-inflation

**Note:** The `ar_activity` subquery groups by `(pid, encounter, code)` without `modifier`, which can cause a similar inflation when billing has multiple rows for the same `(pid, encounter, code)` with different modifiers. This is a separate pre-existing bug and should be filed as its own issue.

All new files (`ServiceCodeSummary.php`, `FinancialSummaryReportService.php`, `FinancialSummaryReportServiceTest.php`, `SvcCodeFinancialReportTest.php`) and the modifications to `svc_code_financial_report.php` were generated with Claude Code (Claude Opus 4.5).

### DIFF
--- a/src/Services/Reports/FinancialSummaryReportService.php
+++ b/src/Services/Reports/FinancialSummaryReportService.php
@@ -1,0 +1,105 @@
+<?php
+
+/**
+ * Query logic for the Financial Summary by Service Code report.
+ *
+ * Fixes a bug where the original LEFT JOIN on the codes table matched only
+ * on `code`, producing a Cartesian product when multiple rows shared the
+ * same code value (different modifiers or code types). The fix uses a
+ * subquery grouped by (code, code_type) and joins on both columns.
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Services\Reports;
+
+use OpenEMR\Common\Database\QueryUtils;
+
+class FinancialSummaryReportService
+{
+    /**
+     * Get financial summary data grouped by service code.
+     *
+     * @param ?int $facilityId Filter by facility (null = all)
+     * @param ?int $providerId Filter by provider (null = all)
+     * @param bool $financialReportingOnly Limit to codes flagged for financial reporting
+     * @return ServiceCodeSummary[]
+     */
+    public function getServiceCodeSummary(
+        \DateTimeInterface $fromDate,
+        \DateTimeInterface $toDate,
+        ?int $facilityId = null,
+        ?int $providerId = null,
+        bool $financialReportingOnly = false
+    ): array {
+        $query = <<<'SQL'
+            SELECT
+                b.code,
+                SUM(b.units) AS units,
+                SUM(b.fee) AS billed,
+                SUM(ar_act.paid) AS paid,
+                SUM(ar_act.adjusted) AS adjusted,
+                MAX(c.financial_reporting) AS financial_reporting
+            FROM form_encounter AS fe
+            JOIN billing AS b
+                ON b.pid = fe.pid AND b.encounter = fe.encounter
+            JOIN (
+                SELECT pid, encounter, code,
+                    SUM(pay_amount) AS paid, SUM(adj_amount) AS adjusted
+                FROM ar_activity
+                WHERE deleted IS NULL
+                GROUP BY pid, encounter, code
+            ) AS ar_act
+                ON ar_act.pid = b.pid
+                AND ar_act.encounter = b.encounter
+                AND ar_act.code = b.code
+            INNER JOIN code_types AS ct
+                ON ct.ct_key = b.code_type AND ct.ct_fee = '1'
+            LEFT OUTER JOIN (
+                SELECT code, code_type,
+                    MAX(financial_reporting) AS financial_reporting
+                FROM codes
+                GROUP BY code, code_type
+            ) AS c
+                ON c.code = b.code AND c.code_type = ct.ct_id
+            SQL;
+
+        $conditions = [
+            "b.code_type != 'COPAY'",
+            'b.activity = 1',
+            'fe.date BETWEEN ? AND ?',
+        ];
+        $binds = [
+            $fromDate->format('Y-m-d 00:00:00'),
+            $toDate->format('Y-m-d 23:59:59'),
+        ];
+
+        if ($facilityId !== null) {
+            $conditions[] = 'fe.facility_id = ?';
+            $binds[] = $facilityId;
+        }
+
+        if ($providerId !== null) {
+            $conditions[] = 'b.provider_id = ?';
+            $binds[] = $providerId;
+        }
+
+        $query .= ' WHERE ' . implode(' AND ', $conditions);
+        $query .= ' GROUP BY b.code';
+
+        if ($financialReportingOnly) {
+            $query .= " HAVING MAX(c.financial_reporting) = '1'";
+        }
+
+        $query .= ' ORDER BY b.code';
+
+        /** @var array<int, array{code: string, units: string, billed: string, paid: string, adjusted: string, financial_reporting: string|null}> $records */
+        $records = QueryUtils::fetchRecords($query, $binds);
+
+        return array_map(ServiceCodeSummary::fromArray(...), $records);
+    }
+}
--- a/src/Services/Reports/ServiceCodeSummary.php
+++ b/src/Services/Reports/ServiceCodeSummary.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * Data object representing one row of the Financial Summary by Service Code report.
+ *
+ * @package   OpenEMR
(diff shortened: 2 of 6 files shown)


src/Services/Reports/FinancialSummaryReportService.php
```diff
@@ -0,0 +1,122 @@
...
+            $binds[] = $providerId;
+        }
+
+        if ($financialReportingOnly) {
+            $query .= " AND c.financial_reporting = '1'";
+        }
+
+        $query .= ' GROUP BY b.code ORDER BY b.code, fe.date, fe.id';
+
+        $records = QueryUtils::fetchRecords($query, $binds);
```
kojiromike, line 101: Perhaps the DTO should have a `fromArray` method so we can get to sanity quicker.

src/Services/Reports/FinancialSummaryReportService.php
```diff
@@ -0,0 +1,122 @@
...
+        if (!is_array($records)) {
+            return [];
+        }
+
+        $results = [];
+        foreach ($records as $record) {
+            if (!is_array($record)) {
+                continue;
```
kojiromike: how could a single row not be an array?

src/Services/Reports/FinancialSummaryReportService.php
kojiromike: Added `ServiceCodeSummary::fromArray()` and removed the defensive guards. The `is_array` checks were scaffolding to satisfy PHPStan level 10 on `mixed` types from `fetchRecords()` — replaced with a `@var` annotation on the return value.

src/Services/Reports/ServiceCodeSummary.php
```diff
@@ -0,0 +1,63 @@
...
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Services\Reports;
+
+class ServiceCodeSummary
```
Firehed:
```suggestion
readonly class ServiceCodeSummary
```
and make the constructor args normal

src/Services/Reports/FinancialSummaryReportService.php
```diff
@@ -0,0 +1,108 @@
...
+            $query .= " AND c.financial_reporting = '1'";
+        }
+
+        $query .= ' GROUP BY b.code ORDER BY b.code, fe.date, fe.id';
+
+        /** @var array<int, array{code: string, units: string, billed: string, PaidAmount: string, AdjustAmount: string, Balance: string, financial_reporting: string|null}> $records */
+        $records = QueryUtils::fetchRecords($query, $binds);
+
+        $results = [];
+        foreach ($records as $record) {
```
Firehed: style: `array_map(ServiceCodeSummary::fromArray(...), $records)`?

src/Services/Reports/FinancialSummaryReportService.php
```diff
@@ -0,0 +1,108 @@
...
+namespace OpenEMR\Services\Reports;
+
+use OpenEMR\Common\Database\QueryUtils;
+
+class FinancialSummaryReportService
+{
+    /**
+     * Get financial summary data grouped by service code.
+     *
+     * @param string $fromDate Start date in YYYY-MM-DD format
```
Firehed: It's probably asking too much right now, but if this could take actual `DateTimeInterface` params, it'd be lovely.

src/Services/Reports/FinancialSummaryReportService.php
```diff
@@ -0,0 +1,108 @@
...
+     * @param string $toDate   End date in YYYY-MM-DD format
+     * @param int    $facilityId Filter by facility (0 = all)
+     * @param int    $providerId Filter by provider (0 = all)
+     * @param bool   $financialReportingOnly Limit to codes flagged for financial reporting
+     * @return ServiceCodeSummary[]
+     */
+    public function getServiceCodeSummary(
+        string $fromDate,
+        string $toDate,
+        int $facilityId = 0,
```
Firehed: Seeing the only (?) call site, I don't think there's any reason for these to be optional params with defaults.

I also am not a fan of the `0 == all` thing (and especially not the `$facilityId > 0` later on); `?int` and null checks would be a bit more semantically accurate.

src/Services/Reports/FinancialSummaryReportService.php
```diff
@@ -0,0 +1,108 @@
...
+            SQL;
+        $binds[] = $fromDate . ' 00:00:00';
+        $binds[] = $toDate . ' 23:59:59';
+
+        if ($facilityId > 0) {
+            $query .= ' AND fe.facility_id = ?';
+            $binds[] = $facilityId;
+        }
+
+        if ($providerId > 0) {
```
Firehed: See comment in constructor; applies to check above as well.

src/Services/Reports/ServiceCodeSummary.php
kojiromike: Done — switched to `readonly class` with normal constructor args.

src/Services/Reports/FinancialSummaryReportService.php
kojiromike: Done — `array_map(ServiceCodeSummary::fromArray(...), $records)`

src/Services/Reports/FinancialSummaryReportService.php
kojiromike: Done — changed to `?int` with `null` defaults and `\!== null` checks. The report file now converts the POST values to `?int` at extraction time (null = all, int = filter to that ID).

src/Services/Reports/FinancialSummaryReportService.php
kojiromike: Agreed that would be nicer. Leaving as-is for now since the report file hands us raw strings from POST, but it would be a good follow-up.

kojiromike: I changed Claude's mind and made it go with DateTimes anyway

kojiromike:
@Firehed ready for re-review

Firehed (approved):

stephenwaite:
Super fix @kojiromike ! Suppose you wanted a summary for every code/modifier pair in the codes table in case that was interesting or important to the A/R team. I guess that would be a new feature :)

kojiromike:
@stephenwaite sounds like it.

PR #10439, opened 2026-01-30
fix(ci): Adjust PHPStan to only scan the critical vendor paths

Firehed:
Towards #10236.

PHPStan is pre-scanning the _entire_ vendor directory. Doing so is, under normal circumstances, unnecessary (it knows how to read symbols from there). It can also result in conflicts - in particular, if it scans polyfills for native functions, it can throw off the signatures used during static analysis.

Since ADODB is not a good citizen regarding autoloading, we still need to do a limited amount of this, but not the whole shebang.

- Change PHPStan to only scan the necessary vendor paths to function as intended
- Bump the memory limit on the baseline splitter

No

### DIFF
--- a/.phpstan/phpstan.github.neon
+++ b/.phpstan/phpstan.github.neon
@@ -3,6 +3,8 @@ includes:
   - baseline/loader.php
 parameters:
   reportUnmatchedIgnoredErrors: true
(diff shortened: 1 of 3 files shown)


bradymiller:

PR #10443, opened 2026-01-30
fix(ci): run coverage conversion inside container for correct path resolution

### LINKED ISSUE #10442: ci: e2e/API coverage not reported to Codecov due to empty Filter in convert-coverage
## Summary

The `ci/convert-coverage` script produces an essentially empty Clover XML report (330 bytes) for e2e and API test coverage, despite collecting valid raw coverage data (286MB `.cov` file from 2293 raw files). This means Codecov receives no e2e or API line coverage data.

## Root cause

`ci/convert-coverage` line 148 creates a `Filter` with no source directories:

```php
$filter = new Filter();
$coverage = new CodeCoverage((new Selector())->forLineCoverage($filter), $filter);
```

The `Filter` controls which files appear in the Clover report. Without any directories added, the Clover output is empty.

By contrast, the native PHPUnit test suites (unit, services, etc.) receive filter directories via `--coverage-filter` CLI flags, which is why their Clover reports work correctly.

## Fix

Add the same source directories used by PHPUnit's `--coverage-filter` flags to the `Filter` object:

```php
$filter = new Filter();
foreach (['apis', 'gacl', 'interface', 'library', 'modules', 'oauth2', 'portal', 'sites', 'src', 'tests'] as $dir) {
    $filter->includeDirectory($projectRoot . '/' . $dir);
}
```

Or accept `--coverage-filter` arguments in the script to match the CI invocation.

## Impact

All e2e and API server-side coverage has been silently lost from Codecov reports. Files like `interface/reports/*.php` that are only exercised through browser requests show 0% despite being executed during e2e tests.

kojiromike:
Fixes #10442

E2E and API coverage reports produce empty/trivial Clover XML (330 bytes) because the `convert-coverage` script runs on the GitHub Actions host where container paths don't resolve.

- Add `--coverage-filter` option to `ci/convert-coverage` that populates the CodeCoverage `Filter` with PHP files from specified source directories using `FileIteratorFacade`
- Add `convert_coverage()` shell helper in `ci/ciLibrary.source` that runs the conversion **inside the Docker container** via `_exec`, passing the standard filter directories with a 4G memory limit
- Replace the copy-from-container + host-side conversion workflow steps with the new in-container `convert_coverage` helper in both `test.yml` (API and E2E) and `ci/inferno/run.sh`
- Skip `.cov` serialization (output to `/dev/null`) since Codecov handles server-side merging — only Clover XML is needed
- Remove the now-unused `COVERAGE_RAW_TMPDIR` env var and `dc cp` steps
- Update `ci/README-COVERAGE.md` to reflect the new approach

**Root cause:** The raw coverage arrays contain absolute container paths (e.g., `/var/www/localhost/htdocs/openemr/src/Foo.php`). On the host, `is_file()` returns false so no File nodes are added to the report tree, and `realpath()` returns false so files can't be added to the filter. Running inside the container where the source code lives fixes both issues.

**Result:** Coverage improved from 6.48% to 16.04% (+9.56pp) as e2e and API coverage data now reaches Codecov.

All changes in this PR were generated with the assistance of Claude Code (Claude Opus 4.5).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -382,7 +382,6 @@ jobs:
       run: |
         . ci/ciLibrary.source
         configure_coverage
-        echo "COVERAGE_RAW_TMPDIR=${RUNNER_TEMP}/coverage-e2e-raw" >> "${GITHUB_ENV}"
 
     ##
     # Setup auto_prepend.php for API and E2E test coverage collection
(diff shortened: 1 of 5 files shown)


bradymiller:
Did not port this over to rel-800.
(I did try, but it broke inferno ci testing)

Issue #10334, opened 2026-01-22
feat: Generic webhook handling infrastructure


Firehed:
Caveat: this should probably be a module eventually, but it may not make sense to implement it as such until we make more headway with the next-gen modules

## Background
A lot of external services deliver updates asynchronously via webhooks. While the security mechanisms and data formats are service-specific, there's a lot of commonality for pretty much all implementations:

- Webhooks should have a way of being processed exactly once, and tend to have unique identifiers enabling this
- Webhooks tend to have an event type, and any given application may have zero or more handlers for any given event type (and a handler may process multiple event types)
- Providers tend to resend several times if it got a non-2xx response
- It's useful to be able to process webhooks asynchronously
- It's useful to have a log of webhooks that were received or process ed(especially for auditing and debugging)
- It's useful to have tooling to force reprocessing[^reprocess]
- It's important to be able to test processing in unit tests that don't actually require a request coming in

I'm looking at this from a payments lens, but the same applies to numerous other services. Notifications is another big one.

## Goals

Rather than building all of this logic once per category (or, worse, once per provider!), it would be great to support this more generically.

At this point I'm not going to suggest any specific approaches, but I didn't see any tables in the system suggesting that this infrastructure exists. But very loosely, I expect:

- Some sort of structured serialization format
- Processing state tracking
- Some sort of "webhook arrived -> send to relevant processors" pipeline
- A couple CLI tools for debugging and replaying

[^reprocess]: This is mostly for when processing failed repeatedly due to a code issue, and after a patch has been applied to resolve said issue, you need to try again.

adunsulag:
@Firehed The foundation worked on a grant application last year to implement the FHIR subscriptions api to handle webhooks.  My ideal would be to build off of that implementation guide https://build.fhir.org/ig/HL7/fhir-subscription-backport-ig/  Sadly we didn't get the grant to fund it, but it'd be good to run with this.

I built a private module that sadly the client didn't contribute to core on webhooks so you're right the infrastructure is lacking there.

sjpadgett:
If it's meant to be a module then it should be a module from get. We still have to move core modules anyway and I feel it'd be easier.
Plus payments needs to be moved to a module as I planned before you took over. Your in this space now so please move payments to a module.
Since our mission is modules for the future it should be a primary focus

PR #10445, opened 2026-01-30
feat(payments): Add support for Rainforest as a payment gateway

Firehed:
Adds support for [Rainforest](https://www.rainforestpay.com) as a payment gateway. It works in both the main app and patient portal payment screens.

> [!IMPORTANT]
> I wanted to implement this as an external module, but due to the current implementation of payments logic in core, doing so is infeasible at the moment. We have plans to pull payments logic out to modules, but completing this is blocked on a few things. The implementation is designed to be as easy to extract as possible once doing so becomes practical. Doing this also helps inform the design of what a payments module should look like and how it needs to work!

- Adds Rainforest API integration
- Adds endpoints for getting data needed by their [component](https://docs.rainforestpay.com/docs/process-payins-via-component)
- Adds an endpoint for processing their webhooks, which triggers the table writes
- Adds support for configuring this path
- Broadens scope of the payment data recording service to also handle `ar_session`
- Adds a README explaining how to configure and use it
- A number of unit tests to validate functionality

Nearly all of the logic is contained in `src/PaymentProcessing/Rainforest`, and largely builds on the PSR standards to minimize core coupling and easing future extraction. The core touch-points are kept to a minimum:

- interface/patient_file/front_payment.php and portal/portal_payment.php have switching logic that can't be handled by modules yet. It follows the existing patterns within those files and their surrounding touch-points
- the new driver option has been added into globals (the driver-specific configs are easy enough to handle in a module, but the addition into `payment_gateway` is not)

The addition of `moneyphp/money` to composer.json is a formality; it's already a transient dependency.

Most of the tests were AI-generated, and reviewed by me prior to shipping. Nearly 100% of the source code is human-written (started at 100% then did a tiny amount of AI refactoring to improve testability)

### DIFF
--- a/interface/webhooks/payment/rainforest.php
+++ b/interface/webhooks/payment/rainforest.php
@@ -71,7 +71,7 @@ try {
         rethrowLastProcessingException: true,
     );
     $disp->dispatch($wh);
-} catch (Throwable $e) {
+} catch (Throwable) {
     // Already logged by dispatcher.
     header('HTTP/1.1 500 Internal Server Error');
     exit;
--- a/src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
+++ b/src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
@@ -47,7 +47,7 @@ readonly class Dispatcher
                     'Excepting during webhook {id} processing ({handler})',
                     [
                         'exception' => $e,
-                        'handler' => get_class($processor),
+                        'handler' => $processor::class,
                         'id' => $webhook->id,
                     ]
                 );


src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
```diff
@@ -0,0 +1,62 @@
...
+            } catch (Throwable $e) {
+                $processingError = true;
+                $this->logger->error('', [
+                    'exception' => $e,
+                ]);
+            }
+        }
+
+        if ($processingError) {
+        }
```
kojiromike: This block is a no-op. Either remove it or add a TODO explaining the intended behavior (e.g. returning a non-2xx status to trigger webhook retry).

.phpstan/baseline/variable.undefined.php
```diff
@@ -17273,7 +17273,7 @@
 ];
 $ignoreErrors[] = [
     'message' => '#^Variable \\$pid might not be defined\\.$#',
-    'count' => 1,
+    'count' => 23,
```
kojiromike: Jump from 1 to 23 seems disproportionate to the ~12 lines added to `portal_payment.php`. Was the baseline regenerated against the full file? Worth confirming this count is accurate and not masking a regression.

library/globals.inc.php
```diff
@@ -3406,6 +3407,31 @@ function gblTimeZones()
...
+            xl('Rainforest API Key'),
+            'encrypted',
+            '',
+            xl('Used when payment_gateway is Rainforest. Stored encrypted'),
+        ],
+        'rainforest_merchant_id' => [
+            xl('Rainforest Merchant ID'),
+            'text',
+            '',
+            xl('Used when payment_gateawy is Rainforest.'),
```
kojiromike: Typo: `payment_gateawy` should be `payment_gateway`. Same typo on the `rainforest_platform_id` entry a few lines down.

src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
kojiromike, line 59: Why the no-op block?

src/PaymentProcessing/Rainforest/Webhooks/Webhook.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    public function __construct(array $body)
+    {
+        $this->data = $body['data'];
+        $this->eventType = $body['event_type'];
+    }
+
+    public function getMerchantId(): ?string
+    {
+        return $this->data['merchant_id'];
```
kojiromike: The return type is `?string` but this will throw an undefined index notice if `merchant_id` is missing from `$data`, rather than returning `null`. Should this use `$this->data['merchant_id'] ?? null`?

src/PaymentProcessing/Rainforest/README.md
```diff
@@ -0,0 +1,100 @@
...
+- `Webhook`: A data structure representing a parsed, verified webhook body
+
+> [!NOTE]
+> Rainforest uses Svix for their webhooks; as such, the validation logic could potentially be made more general and compatible with other services.
+
+Other classes in the directory are generally implementations of `ProcessorInterface`.
+
+### Related code
+
+- `interface/webhooks/rainforest/webhook.php`: Endpoint to receive and handle the webhooks
```
kojiromike: Path is wrong — should be `interface/webhooks/payment/rainforest.php`.

templates/payments/rainforest.js
```diff
@@ -0,0 +1,56 @@
...
+        patientId,
+    }
+
+    const container = document.getElementById('payment-form');
+    container.innerHTML = '<p>Loading...</p>';
+
+    /**
+     * @param { session_key: string, payin_config_id: string } responseData
+     */
+    const createRainforstComponent = function (responseData) {
```
kojiromike: Typo: `createRainforstComponent` → `createRainforestComponent`

.phpstan/baseline/variable.undefined.php
kojiromike: If you rebase, this should not be needed anymore.

composer.json
```diff
@@ -81,6 +81,7 @@
         "league/csv": "^9.28.0",
         "league/oauth2-server": "^8.4",
         "league/omnipay": "^3.2.1",
+        "moneyphp/money": "^4.8",
```
kojiromike, line 84: Promotes `moneyphp/money` from a transitive dependency (via `league/omnipay`) to a direct one, and upgrades from v3 to v4. I checked: nothing else in the codebase uses `Money\` classes directly — `src/Billing/PaymentGateway.php` only uses `Omnipay\Omnipay` and `Omnipay\Common\CreditCard` with plain scalar amounts. Omnipay already permits v4 (`^3.1|^4.0.3`), and `ext-bcmath` (newly required by v4) is already a global requirement. Upgrade looks safe.

src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
Firehed: Left behind in error from previous error handling implementation.

composer.json
Firehed: https://github.com/moneyphp/money/blob/master/CHANGELOG.md#400---2021-05-17 calls out the few BC breaks across the major versions; as you note, they shouldn't be impactful based on current usage.

stephenwaite: https://github.com/openemr/openemr/pull/8396/changes

stephenwaite: wish oce was going to improve Sphere since they give a percentage to the foundation :100:

Firehed: dev-easy docker at least includes the extension; it should get flagged if missing anywhere else during `composer install` or `composer check-platform-reqs` unless explicitly bypassing it.

if there are specific sphere things you're referencing, let me know what tickets exist for them (or add some 😄). We may be able to pick some up as part of payments module-ization and next-steps.

interface/patient_file/front_payment.rainforest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+chdir(dirname(__DIR__, 2));
+require 'vendor/autoload.php';
```
stephenwaite: doesn't require globals below bring in vendor/autoload?

Firehed: _Technically_ yes. I wanted to avoid relying on that since when this does get ripped out to a module properly, the code should no longer depend on loading that file at all (making a lot of assumptions about how much progress we make with modules next steps, of course).

portal/portal_payment.rainforest.php
```diff
@@ -0,0 +1,22 @@
...
+chdir(dirname(__DIR__));
+require 'vendor/autoload.php';
+
+use Http\Discovery\Psr17Factory;
+use OpenEMR\Common\Session\SessionUtil;
+use OpenEMR\Core\OEGlobalsBag;
+use OpenEMR\PaymentProcessing\Rainforest\Apis\GetPayinComponentParameters;
+
+SessionUtil::portalSessionStart();
+$ignoreAuth_onsite_portal = true;
```
kojiromike: **Critical: No authentication verification.**

This sets `$ignoreAuth_onsite_portal = true` but never verifies the session belongs to an authenticated portal user. Other portal endpoints (e.g., `portal/messaging/secure_chat.php`) check `$session->get('pid')` and `$session->get('patient_portal_onsite_two')` *before* setting this flag. Without those checks, any request (with any session or no session) can reach the Rainforest API endpoint.

Also missing CSRF verification, same as `front_payment.rainforest.php`.

src/PaymentProcessing/Rainforest/Webhooks/RecordPayment.php
```diff
@@ -0,0 +1,127 @@
...
+    ) {
+    }
+
+    public function getEventTypes(): array
+    {
+        // For the moment, we're listening to authorized rather than succeeded
+        // for two main reasons:
+        // 1) We need CLI tooling to kick payments forward in the sandbox
+        //    environment, and it doesn't make sense to build that until the
+        //    new CLI modules are complete
```
kojiromike: **Warning: Race condition in idempotency check — potential duplicate payments.**

This is a SELECT-then-INSERT with no locking. If Svix delivers the same webhook concurrently (retries), both could pass this check before either inserts, creating duplicate payment records.

The comment on `Recorder::getSessionIdForReference()` acknowledges the `reference` column is neither unique nor nullable, so a unique constraint can't be added without a migration.

At minimum, `SELECT ... FOR UPDATE` would serialize concurrent checks within the transaction. A separate idempotency table or making `reference` nullable + unique would be the robust fix.

src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
```diff
@@ -0,0 +1,60 @@
...
+            );
+            return;
+        }
+
+        foreach ($this->getProcessorsFor($webhook->eventType) as $processor) {
+            try {
+                $processor->handle($webhook);
+            } catch (Throwable $e) {
+                $this->logger->error('', [
+                    'exception' => $e,
```
kojiromike: **Warning:** Empty string `''` as the log message makes these hard to find in logs. More importantly, catching all `Throwable` here means a database failure in `RecordPayment` is silently swallowed — combined with the webhook endpoint always returning 200, this means payment recording failures are invisible to the webhook sender and will never be retried.

Consider using a descriptive message like `'Webhook processor failed'` and/or allowing critical processors to propagate exceptions.

src/PaymentProcessing/Rainforest/Api.php
```diff
@@ -0,0 +1,177 @@
...
+         * (plus much more)
+         * @var array{
+         *   data: array{
+         *     session_key: string,
+         *   }
+         * }
+         */
+        $sessionResponse = $this->post('/v1/sessions', $sessionPayload);
+        $sessionKey = $sessionResponse['data']['session_key'];
+
```
kojiromike: **Warning: Idempotency key not persisted.**

As the TODO notes, this generates a fresh UUID every call. If the browser request fails after Rainforest creates the `payin_config` but before the response reaches the browser, a retry will send a different idempotency key and could create a second `payin_config`, potentially allowing a double charge.

templates/payments/rainforest.js
```diff
@@ -0,0 +1,56 @@
...
+        container.replaceChildren(component);
+
+        component.addEventListener('approved', function (data) {
+            // Processing happens async through the webhooks.
+            alert('Payment complete! It may take a few minutes to be reflected in the dashboard.');
+            window.location.reload();
+        })
+    }
+
+    $.ajax({
```
kojiromike: **Warning: No error handling.** There's no `error` callback on this AJAX call. If the server returns 500, 405, or a network error, the user sees the "Loading..." text indefinitely with no indication of what went wrong.

interface/patient_file/front_payment.rainforest.php
stephenwaite: oh okay, but shouldn't it be `require_once()`?

interface/webhooks/payment/rainforest.php
```diff
@@ -0,0 +1,63 @@
...
+            // Future: this should all get wired through DI, etc.
+            new RecordPayment(logger: $logger),
+        ],
+        merchantId: $mid,
+        logger: $logger,
+    );
+    $disp->dispatch($wh);
+} catch (Throwable $e) {
+    error_log((string)$e);
+    header('HTTP/1.1 400 Bad Request');
```
kojiromike: Does this `header()` call actually work here? Line 16 already sends `Content-type: text/plain`, and `globals.php` on line 19 likely starts the session (which sends headers). If headers are already sent by this point, this is silently ignored and Svix sees a 200 — meaning it won't retry.

Also, `Dispatcher` catches all `Throwable` from processors (line 43 in Dispatcher.php), so if `RecordPayment::handle()` throws (say, a database failure), this catch block is never reached either — the endpoint returns 200 regardless.

Would `http_response_code(400)` work better here? Or restructuring so no output is sent before the try/catch?

interface/patient_file/front_payment.rainforest.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Http\Discovery\Psr17Factory;
+use OpenEMR\Core\OEGlobalsBag;
+use OpenEMR\PaymentProcessing\Rainforest\Apis\GetPayinComponentParameters;
+
+require_once 'interface/globals.php';
```
kojiromike, line 22: Shouldn't this have the same CSRF check as `front_payment.php`? That file calls `CsrfUtils::verifyCsrfToken()` before processing. This endpoint accepts POSTs that create real payment sessions via the Rainforest API.

portal/portal_payment.rainforest.php
kojiromike, line 21: This sets `$ignoreAuth_onsite_portal = true` but doesn't verify the session actually belongs to an authenticated portal user. Other portal endpoints (e.g., `portal/messaging/secure_chat.php`) check `$session->get('pid')` and `$session->get('patient_portal_onsite_two')` before setting this flag — shouldn't this do the same?

Same CSRF question as `front_payment.rainforest.php` applies here too.

src/PaymentProcessing/Rainforest/Webhooks/Verifier.php
```diff
@@ -0,0 +1,123 @@
...
+ * them is the webhook body format.
+ */
+class Verifier
+{
+    private readonly string $secretBytes;
+
+    public function __construct(
+        #[SensitiveParameter] string $webhookSecret,
+    ) {
+        if (!str_contains($webhookSecret, '_')) {
```
kojiromike, line 55: The timestamp is used in the signature computation but never checked against the current time. The [Svix docs](https://docs.svix.com/receiving/verifying-payloads/how-manual#verify-timestamp) recommend comparing it to prevent replay attacks (their default tolerance is 5 minutes). Without this, a captured signed payload could be replayed indefinitely.

Something like `if (abs(time() - (int)$timestamp) > 300) throw ...` would close this.

src/PaymentProcessing/Rainforest/Webhooks/RecordPayment.php
kojiromike, line 43: This is a SELECT-then-INSERT without locking — if Svix delivers the same webhook concurrently (retries can overlap), both could pass this check before either inserts, creating duplicate payment records. Would `SELECT ... FOR UPDATE` help here? Or is the expectation that Svix won't send concurrent deliveries for the same event?

src/PaymentProcessing/Rainforest/Webhooks/RecordPayment.php
```diff
@@ -0,0 +1,127 @@
...
+                $this->logger->notice('ar_session exists for rainforest payin_id {id}, skipping further processing', [
+                    'id' => $reference,
+                ]);
+                return;
+            }
+
+            $metadata = Metadata::fromParsedJson($data['metadata']);
+            $patientId = $metadata->patientId;
+
+            $memo = sprintf('Rainforest transaction id %s', $data['payin_id']);
```
kojiromike, line 88: The Sphere integration writes to `payment_processing_audit` via `PaymentProcessing::saveAudit()` for every transaction. Is there a reason to skip that here, or is it just not wired up yet?

src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
kojiromike: Empty string as the log message will make these hard to find. More importantly, catching all `Throwable` here means a database failure in `RecordPayment` gets swallowed — combined with the webhook endpoint returning 200 (see comment there), payment recording failures would be invisible to the webhook sender and never retried.

src/PaymentProcessing/Rainforest/Api.php
kojiromike, line 82: The TODO covers this, but worth flagging: since this generates a fresh UUID every call and never persists it, a browser retry after a network failure could create a second `payin_config` with a different idempotency key. Is Rainforest's component smart enough to prevent the user from being charged twice in that scenario?

templates/payments/rainforest.js
kojiromike, line 53: No `error` callback here — if the server returns a 500 or there's a network issue, the user just sees "Loading..." forever with no indication of what happened.

interface/webhooks/payment/rainforest.php
Firehed: It does. As long as all headers are sent prior to (unbuffered) non-header content, it doesn't matter what order they're sent in - they get reordered appropriately by the mod_php/fastcgi handling. I've double-checked just now by forcing a `throw` and observed the correct response code.

Good call on the error handling though, I'll take another look at that.

src/PaymentProcessing/Rainforest/Webhooks/Verifier.php
Firehed: Ah, must have missed that bit. Good catch.

src/PaymentProcessing/Rainforest/Webhooks/Dispatcher.php
Firehed: I'm perhaps making an incorrect assumption that the selected logger is actually handling the [reserved key `exception`](https://www.php-fig.org/psr/psr-3/#13-context) correctly.

src/PaymentProcessing/Rainforest/Api.php
Firehed: Yes, this won't be an issue in practice. You could generate a thousand of these values and only the one actually passed in to the component when the user hits pay will get used. It's _unlike_ e.g. Stripe where the idempotency keys are used in the actual "charge the card" API call.

It's not how I'd implement it in a cleanroom environment, but given the current payment portal workflows and UI, it'll be fine. I would like to better account for this in an overhaul of the system, though.

interface/patient_file/front_payment.rainforest.php
Firehed: Sessions are throw-away temporary resources so I don't see any actual danger in leaving it open. If you think it's worth the complexity (on both ends - need to change the ajax as well) I can add it, but all else being equal I'd rather not.

src/PaymentProcessing/Rainforest/Webhooks/RecordPayment.php
Firehed: The actual data changes across the different payment processors is, at present, _very_ inconsistent in my findings. I based this implementation on snapshotting and diffing my database when using the Stripe APIs, which didn't touch that table (or, at least, not on the path I was on!)

So AFICT it's not necessary.

But, of course, these _should_ all work the same way and getting us there is the motivation behind the `Recorder` service. I very much want to get these all lined up properly but I think doing so is outside of the scope of this change.

src/PaymentProcessing/Rainforest/Webhooks/RecordPayment.php
Firehed: My understanding is that Svix guarantees you won't get retries in parallel (and it certainly wouldn't make any actual sense for them to send them that way). They won't retry unless they get a failure on a prior attempt or you explicitly request one.

The proper solutions here are twofold:
- #10334
- As part of a payment infra overhaul, add a unique constraint to the `ar_sessions.reference` (or something about equivalent to it), making the SELECT unnecessary.

I'm disinclined to add a whole lot of extra customization to the DB handling here since it'll make near-term DB updates harder given the way this will work once deployed, and this is already quite a bit more reliable than the other payment gateway paths.

interface/webhooks/payment/rainforest.php
Firehed: Update: the error handling design here _was_ intentional, but isn't really ideal in the current deployment context. I've updated it accordingly, with some future notes as to how it should work after #10334.

interface/patient_file/front_payment.rainforest.php
Firehed: https://github.com/openemr/openemr/pull/10445/commits/af2b01e17196f927acb4dc20cb0b8f1cd21fde87 adds csrf protection

interface/patient_file/front_payment.php
```diff
@@ -1852,6 +1854,19 @@ function posDialog() {
...
         }
+        if ($globalsBag->get('payment_gateway') === 'Rainforest') {
+            if ($globalsBag->getBoolean('gateway_mode_production')) {
+                echo '<script type="module" src="https://static.rainforestpay.com/payment.js"></script>';
+            } else {
+                echo '<script type="module" src="https://static.rainforestpay.com/sandbox.payment.js"></script>';
+            }
+            echo '<script type="text/javascript">';
+            echo $twig->render('payments/rainforest.js', [
+                'csrf' => CsrfUtils::collectCsrfToken('rainforest', $session->getSymfonySession()),
```
Firehed, line 1865: Note here and in portal_payment.php: the twig helper for csrf doesn't work in this use-case (presumably due to session changes); this has proven reliable so far.

kojiromike:
@Firehed the only trustworthy way to re-trigger the tests was to push an empty commit.

interface/patient_file/front_payment.rainforest.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
bradymiller, line 1: missing header

portal/portal_payment.rainforest.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
bradymiller, line 1: header

portal/portal_payment.rainforest.php
bradymiller: Above was not answered, but i am assuming this script needs to work for unauthorized users?
Which seems odd since the above session block relies on the app cookie to work (which require user to go the the patient portal login page)

src/PaymentProcessing/Rainforest/Api.php
```diff
@@ -0,0 +1,177 @@
...
+use Ramsey\Uuid\Uuid;
+use SensitiveParameter;
+
+/**
+ * Wrapper to interact with the Rainforest API
+ */
+readonly class Api
+{
+    private const SANDBOX_HOST = 'https://api.sandbox.rainforestpay.com';
+    private const PRODUCTION_HOST = 'https://api.rainforestpay.com';
```
bradymiller, line 29: seeing above lines duplicated in several places

bradymiller:

Issue #9233, opened 2025-10-23
`chmod 777` during ci test setup


kojiromike:
# CI: chmod 777 and Docker UID/GID mismatch

## Problem

The CI system currently uses `chmod 777` throughout the test workflow to work around file permission issues caused by Docker containers creating files with different UID/GID than the GitHub Actions runner.

### Current instances:
- `ciLibrary.source:85` - `chmod -R 0777 .` on entire repository
- `ciLibrary.source:121` - `chmod -R 0777 sites/default/documents`
- `ciLibrary.source:216` - `chmod 777 selenium-videos`
- Various other locations in test.yml workflow

### Issues:
1. **Security/cleanliness**: Files and directories get executable permissions unnecessarily
2. **Root cause**: Docker containers run as root/www-data but create files on volumes that are mounted from the host runner user
3. **TODO comment**: Line 76 in `ciLibrary.source` already acknowledges this: "TODO, figure out how not to require the below line (maybe chown or something like that)"

### Option 1: Run containers with host UID/GID
```yaml
services:
  openemr:
    user: "${UID}:${GID}"
```
**Challenge**: Services like Apache/PHP-FPM may need root privileges

### Option 2: Fix ownership instead of permissions
Use `chown` to change file ownership to runner user instead of opening permissions:
```bash
sudo chown -R $(id -u):$(id -g) coverage/
```

### Option 3: Use more specific permissions
Instead of 777, use:
- 644 for regular files (rw-r--r--)
- 755 for directories and executables (rwxr-xr-x)

### Option 4: Docker entrypoint script
Have containers fix permissions on mounted volumes in their entrypoint before starting services

### Option 5: Docker userns-remap
Use Docker's user namespace remapping feature to automatically map container root to unprivileged host user

PR #10450, opened 2026-01-30
fix(docker): Don't special-case vendor when mounting local filesystem

Firehed:
When using the `development-easy` Docker Compose setup, the host's version of `vendor` isn't mounted; instead, it's pinned to what's in the image. This can disrupt development and testing of package changes, including working on modules.

AFAICT the Compose file was always set up this way, it's unclear to me from the PRs in the git history why this is. I've never seen it done before - every containerized PHP app I've _ever_ worked on simply mounts the repo root straight-through (in fact, every interpreted language I've seen does that, full stop). I'd go so far as to say that _most_ of the surrounding volume mounts are likely to cause similar problems (the db ones are fine in the other services), but I haven't dug in there yet.

Removes the special-casing of `vendor` from the compose file, causing the host's `vendor/` directory to stay synced inside of the docker container at runtime.

No

### DIFF
--- a/docker/development-easy/docker-compose.yml
+++ b/docker/development-easy/docker-compose.yml
@@ -37,7 +37,6 @@ services:
     - themevolume:/var/www/localhost/htdocs/openemr/public/themes:rw
     - sitesvolume:/var/www/localhost/htdocs/openemr/sites:rw
     - nodemodules:/var/www/localhost/htdocs/openemr/node_modules:rw
-    - vendordir:/var/www/localhost/htdocs/openemr/vendor:rw
     - ccdanodemodules:/var/www/localhost/htdocs/openemr/ccdaservice/node_modules:rw
     - logvolume:/var/log
     - couchdbvolume:/couchdb/data
@@ -187,7 +186,6 @@ volumes:
   themevolume: {}
   sitesvolume: {}
   nodemodules: {}
-  vendordir: {}
   ccdanodemodules: {}
   logvolume: {}
   couchdbvolume: {}


Firehed:
If this is intentional/required, I'd love to know the motivation behind it and make sure the compose file(s) are commented accordingly. I've mostly been running development with this change uncommitted and haven't run into any weird behavior.

docker/development-easy/docker-compose.yml
```diff
@@ -37,7 +37,6 @@ services:
     - themevolume:/var/www/localhost/htdocs/openemr/public/themes:rw
     - sitesvolume:/var/www/localhost/htdocs/openemr/sites:rw
     - nodemodules:/var/www/localhost/htdocs/openemr/node_modules:rw
-    - vendordir:/var/www/localhost/htdocs/openemr/vendor:rw
```
kojiromike, line 40: I think the exclusion of vendor here is to protect the user's dev env from the chowns and chmods in [the entrypoint](https://github.com/openemr/openemr-devops/blob/3106d0efe47cd851ea9fcd1189ea14f63e2092e4/docker/openemr/flex/openemr.sh).

It's pretty much the same issue as https://github.com/openemr/openemr/issues/9233, but not just for the ci containers.

Firehed: Hmmm... ok. Maybe it makes sense to leave this one parked in draft for now. I'd like to take a more comprehensive pass at simplifying docker/containerization stuff (and, by extension, perhaps some deployment) but I also don't want to add yet another giant project to my plate until we make some headway on the ones in progress.

bradymiller:
Hi,
The challenge for the dev (I think the dev was toolbox) that created the
easy dev environment was to deal with a specific issue that devs were
having when they tested directly from their git repo. It's something I
don't do, but lots of younger devs were doing it and the PRs kept having
bunch of artifacts from the builds/testing and the devs seemed to be
getting frustrated with this. So, toolbox created easy dev so that
building/testing directly from a dev's git repo would not modify the actual
repo (ie. everything is hidden in the docker). So then the devs can code
and test from the same place.

PR #10452, opened 2026-01-31
fix(crypto): Tighten standard encryption test

### LINKED ISSUE #10451: refactor(crypto): Tighten cryptCheckStandard()
Current method can be unreliable.
intval() is very permissive:
"001" → 1 (maybe fine)
"1ab" → 1 (bad: non-numeric still becomes a number)
"abc" → 0 (bad: could accidentally pass if KeyVersion::from(0) exists)
Using exceptions for a “maybe” check is noisy/slow and can hide edge cases.
A 3-digit numeric prefix is a weak “magic” marker: plenty of plaintext can start with digits.

With intval(mb_substr(...)), strings like "007abc..." still pass the intval() part (intval('007') = 7), and you never verify the payload is actually an encrypted blob.

sjpadgett:
Prefix 007 (digits, valid KeyVersion::tryFrom(7) if you have it) Payload base64 regex
Strict base64_decode(..., true
Decoded length likely well above 16

Fixes #10451

### DIFF
--- a/src/Common/Crypto/KeyVersion.php
+++ b/src/Common/Crypto/KeyVersion.php
@@ -108,7 +108,7 @@ enum KeyVersion: int
      * Extract a KeyVersion from the prefix of a string.
      *
      * @param string $value The string to check (should be at least 3 bytes)
-     * @return KeyVersion|null the KeyVersion extracted from the first 3 bytes of the string
+     * @return KeyVersion the KeyVersion extracted from the first 3 bytes of the string
      */
     public static function fromPrefix(string $value): self
     {


src/Common/Crypto/KeyVersion.php
```diff
@@ -107,17 +107,21 @@ public static function fromString(string $version): self
     /**
      * Extract a KeyVersion from the prefix of a string.
      *
-     * @param  string $value The string to check (should be at least 3 bytes)
+     * @param string $value The string to check (should be at least 3 bytes)
      * @return self the KeyVersion extracted from the first 3 bytes of the string
      * @throws \ValueError If the prefix cannot be converted to a KeyVersion
      */
-    public static function fromPrefix(string $value): self
+    public static function fromPrefix(string $value): ?self
```
kojiromike: Changing from throwing `\ValueError` to returning `null` needs `decryptStandard()` updated too — it still catches `\ValueError` at CryptoGen.php:85-90:

```php
try {
    $encryptionVersion = KeyVersion::fromPrefix($value);
} catch (\ValueError) {
    error_log("Invalid encryption version prefix");
    return false;
}
```

With this change, `$encryptionVersion` becomes `null` on bad input and the catch never fires, leading to a fatal when `$encryptionVersion->usesLegacyDecryption()` is called on line 109.

One option: update `decryptStandard()` to handle the nullable return. Another: keep `fromPrefix()` throwing and add a `tryFromPrefix()` variant for `cryptCheckStandard()` to use.

src/Common/Crypto/KeyVersion.php
```diff
@@ -107,17 +107,21 @@ public static function fromString(string $version): self
     /**
      * Extract a KeyVersion from the prefix of a string.
      *
-     * @param  string $value The string to check (should be at least 3 bytes)
+     * @param string $value The string to check (should be at least 3 bytes)
      * @return self the KeyVersion extracted from the first 3 bytes of the string
```
kojiromike: Minor: the docblock should be updated to match — `@return ?self` and remove the `@throws \ValueError`.

src/Common/Crypto/CryptoGen.php
```diff
@@ -137,21 +137,51 @@ protected function legacyDecrypt(string $trimmedValue, ?string $customPassword,
...
+        // Validate the 3-digit KeyVersion prefix
+        $kv = KeyVersion::fromPrefix($value);
+        if ($kv === null) {
+            return false;
+        }
+        // Validate the remainder as strict base64
+        $payloadB64 = substr($value, 3);
+
+        // structural checks (fast fail)
+        if ($payloadB64 === '' || (strlen($payloadB64) % 4) !== 0) {
```
kojiromike: I traced all the call sites — every one immediately calls `decryptStandard()` if this returns true, so a false positive just results in a gracefully-failed decrypt. The `fromPrefix()` fix alone addresses #10451, and after that fix, a false positive requires a value starting with exactly `"001"`-`"007"` as literal digits — which doesn't happen with the actual inputs (file contents, database config values, URL tokens).

There's also a performance angle: `base64_decode()` and `preg_match()` both scan the full payload, making this O(n). Several call sites pass in `file_get_contents()` results that could be large.

Would it make sense to keep this as just the tightened prefix check?

sjpadgett: > Would it make sense to keep this as just the tightened prefix check?

okay...

sjpadgett: @kojiromike check me on this my brain isn't working today...

kojiromike: @sjpadgett sorry I missed it. All the runtime code is OK, but there's still a missing docblock change. 1 line. I'll make a PR

sjpadgett: no worries, I could have swore I fixed docbloc

PR #10455, opened 2026-01-31
fix: #10454 webroot session fix

### LINKED ISSUE #10454: bug: session with openemr installed in sub-directory (non-empty webroot) fails to login after logout
### Describe the bug

If an OpenEMR installation is installed in a webroot sub-directory (such as /openemr/) the user receives a site id missing error screen when they login, logout, and attempt to login again on the main openemr login screen.

Digging into this it appears the session setup logic in globals.php is relying on the globals **webroot** parameter which is not setup until after the session is instantiated instead the only available is the poorly named **web_root** which is not what is used in the SessionWrapperFactory class.

### To Reproduce

1. Setup the easy-dev docker environment ```cd docker/development-easy/ && docker compose up -d```
2. Use the openemr-cmd to setup openemr to run with the **openemr** webroot subdirecty ```openemr-cmd  cwo```
3. Open a browser to https://localhost:9300/
4. Login as default  admin user
5. Logout
6. Login again as default admin user
7. See error of site id not found

### Expected behavior
The user is able to login successfully.

### Client configuration

- Browser: Chrome
- OpenEMR version: 8.0.0

adunsulag:
Moved the order of the webroot variable to be higher up in the globals.php file so that it is available in the session instantiation logic.

Originally the session used the variable $web_root but the SymfonySessionWrapper relies on the global value of 'webroot' which didn't happen until further down the instantiation chain.

Fixes #10454

Didn't discover this until working through all of the redis session logic.

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -225,6 +225,8 @@ if (empty($globalsBag)) {
 }
 $globalsBag->set('webserver_root', $webserver_root);
 $globalsBag->set('web_root', $web_root);
+// Absolute path to the location of documentroot directory for use with include statements:
+$globalsBag->set('webroot', $web_root);
 $globalsBag->set('vendor_dir', $GLOBALS['vendor_dir'] ?? "$webserver_root/vendor");
 $globalsBag->set('restRequest', $restRequest);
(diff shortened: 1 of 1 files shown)


bradymiller (approved):

bradymiller:
Straightforward and bringing in to expedite session stuff for 8.0.0

PR #10462, opened 2026-01-31
fix(ci): reduce ~111 false positive Semgrep code scanning alerts

kojiromike:
Fixes #10461

Reduce ~111 false positive Semgrep code scanning alerts across three categories.

## Changes proposed in this pull request
- Add custom `printed-request` taint rule to `semgrep.yaml` mirroring the existing `echoed-request` override but with `print` as the sink (69 alerts)
- Exclude `Documentation/` directory — generated SchemaSpy HTML, not application code (6 `var-in-script-tag` alerts)
- Exclude `*.mustache` files — QRDA/CDA XML templates for clinical data interchange, not browser-rendered HTML (36 `unquoted-attribute-var` alerts)
- Fix `run-semgrep.sh` missing the `--exclude-rule` for `echoed-request` that the CI workflow already had

The mustache templates also have raw `{{{var}}}` in some XML attribute positions that should use escaped `{{var}}` instead — that will be addressed in a follow-up PR.

## AI Disclosure
Yes

PR #10464, opened 2026-01-31
fix(qrda): use escaped Mustache variables in XML attribute positions

### LINKED ISSUE #10463: fix(qrda): use escaped Mustache variables in XML attribute positions
## Summary

The QRDA/CDA Mustache templates under `src/Services/Qrda/` use raw `{{{var}}}` (unescaped) in ~20 XML attribute positions where `{{var}}` (escaped) would be correct. The values are timestamps, OIDs, and numeric data that don't contain XML-unsafe characters today, so the output is identical — but using escaped variables is strictly safer and would allow removing the `*.mustache` exclusion from the Semgrep workflow added in #10462.

## Context

`{{var}}` escapes `&`, `<`, `>`, `"`, `'` — exactly what XML attributes need. `{{{var}}}` skips escaping entirely. The `{{{var}}}` in text content positions (XML fragment injection like `{{{code_and_codesystem}}}`) should stay raw since they contain pre-built XML elements.

## AI Disclosure
Yes

kojiromike:
Fixes #10463

Switch raw `{{{var}}}` to escaped `{{var}}` in XML attribute values across QRDA/CDA Mustache templates.

- Change 20 instances of raw `{{{var}}}` (unescaped) to `{{var}}` (escaped) in XML attribute positions across 17 Mustache templates in `src/Services/Qrda/`
- Affected values are timestamps, OIDs, and numeric data that don't contain XML-unsafe characters, so output is identical today
- Escaped variables ensure proper XML escaping (`&`, `<`, `>`, `"`, `'`) if unexpected data ever appears
- Raw `{{{var}}}` in text content positions (XML fragment injection) is left unchanged since those contain pre-built XML elements

Once #10462 is merged, the `*.mustache` exclusion in the Semgrep workflow can potentially be removed since these were the source of the `unquoted-attribute-var` alerts.

## AI Disclosure
Yes

### DIFF
--- a/src/Services/Qrda/qrda-export/catI-r5/_header.mustache
+++ b/src/Services/Qrda/qrda-export/catI-r5/_header.mustache
@@ -15,7 +15,7 @@
 <code code="55182-0" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" displayName="Quality Measure Report"/>
 <title>QRDA Incidence Report</title>
 <!-- This is the document creation time -->
-<effectiveTime value="{{{current_time}}}"/>
+<effectiveTime value="{{current_time}}"/>
 <confidentialityCode code="N" codeSystem="2.16.840.1.113883.5.25"/>
 <languageCode code="en"/>
 <!-- reported patient -->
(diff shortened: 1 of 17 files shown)


kojiromike:
@sjpadgett @adunsulag in this PR I'm just changing these fields from raw to escaped mustache. This will (at least) make semantic analysis better and it seems safe to me, but did I miss anything?

PR #10466, opened 2026-01-31
fix(security): resolve 7 printed-request Semgrep alerts

kojiromike:
## Summary

- Validate `$_GET['sigline']` against allowed array keys in CAMOS `rx_print.php` instead of using unvalidated user input as an array index (4 alerts)
- Use `match` expression with IIFEs in eye_mag `view.php` to break Semgrep taint chain for `$refresh` (2 alerts)
- Replace `patientFilePostToGet()` with `http_build_query()` in `custom_report.php` and register `http_build_query` as a Semgrep sanitizer (1 alert)

## Test plan

- [ ] Verify CAMOS prescription printing still works with `sigline=plain`, `sigline=embossed`, and `sigline=signed`
- [ ] Verify CAMOS prescription printing falls back to `plain` with an invalid `sigline` value
- [ ] Verify eye_mag PMSFH refresh still loads correctly
- [ ] Verify custom report printable version link still works

## AI disclosure

Yes

### DIFF
--- a/interface/forms/CAMOS/rx_print.php
+++ b/interface/forms/CAMOS/rx_print.php
@@ -9,6 +9,8 @@
  * @author    Brady Miller <brady.g.miller@gmail.com>
  * @copyright Copyright (c) 2006-2009 Mark Leeds <drleeds@gmail.com>
  * @copyright Copyright (c) 2018 Brady Miller <brady.g.miller@gmail.com>
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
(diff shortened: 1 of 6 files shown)


interface/forms/CAMOS/rx_print.php
```diff
@@ -52,6 +54,10 @@
     "<div class='sig'>"
   . "<img src='./sig.jpg'>"
   . "</div>\n";
+$siglineType = $_GET['sigline'] ?? 'plain';
```
kojiromike: is $siglineType ever used except as an array key? Maybe this could be `$siglineValue = $sigline[$_GET['sigline'] ?? 'plain'];` and use the value from there on out.

bradymiller:
Was unable to port this to rel-800 (secondary lots of merge conflicts).

Issue #10456, opened 2026-01-31
bug: View Summary of Care gives 500 error if patient has no assigned PCP nor Carecoordination Default Primary Care Provider assigned


adunsulag:
### Describe the bug

If a patient attempts to view their Transfer Summary of Care document and there is no assigned PCP nor Carecoordination primary care provider the ccda throws a 500 error.

### To Reproduce

1. Generate a new patient w/ portal access and ensure there is no primary care provider 
2. Login as the patient.  
3. Go to Medical Reports -> View Summary of Care

### Expected behavior

CCDA / Summary of Care document displays

### Client configuration

- Browser: Chrome
- OpenEMR version: 8.0.0
- Operating system: Ubuntu 24.04 LTS / easy dev docker environment

PR #10470, opened 2026-01-31
fix: 10456 ccda author session fix

adunsulag:
If an author couldn't be found for the patient (missing PCP or care
coordination default PCP), then a 500 fatal error was thrown.

Changed the logic to fix that issue.  Also fixed a bunch of session
problems where the portal symfony session was not being used and was
using the $_SESSION variable which made the audit logs unusable and
filled the php error log with warnings.

Fixes https://github.com/openemr/openemr/issues/10456

### DIFF
--- a/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php
+++ b/interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php
@@ -236,19 +236,19 @@ class EncounterccdadispatchController extends AbstractActionController
 
                 // split content if unstructured is included from service.
                 $unstructured = "";
-                if (substr_count((string) $content, '</ClinicalDocument>') === 2) {
-                    $d = explode('</ClinicalDocument>', (string) $content);
+                if (substr_count($content, '</ClinicalDocument>') === 2) {
+                    $d = explode('</ClinicalDocument>', $content);
                     $content = $d[0] . '</ClinicalDocument>';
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php
```diff
@@ -384,8 +387,8 @@ public function getConsolidatedQrda3Content($pids = null, $measures = [])
             // Log the event (following your existing audit pattern)
             EventAuditLogger::getInstance()->newEvent(
                 "qrda3-consolidated-generation",
-                $_SESSION['authUser'],
-                $_SESSION['authProvider'],
+                $session->get('authUser')
+                , $session->get('authProvider'),
```
kojiromike:
```suggestion
                $session->get('authUser'),
                $session->get('authProvider'),
```

adunsulag: Bizarre.  rector both inside docker image and outside report the file as fine locally.

adunsulag: Had to fix the issue manually.

kojiromike: The comma thing wasn't a rector thing. Actually there's a whole other cbf thing we need to address.

bradymiller:
nice!

Issue #10235, opened 2026-01-16
Setup random fail with collate exception


sjpadgett:
### Describe the bug

Several issues posted on forum concerning collate not being passed a string value. in install setup
### Client configuration

- Browser:
- OpenEMR version:7.0.4
- Operating system: nginx apache

PR #10472, opened 2026-01-31
fix(setup): correct collate field id attribute and add default value

fzzio:
## Summary

This PR fixes two issues in `setup.php` that cause the installation wizard to fail when selecting "I have already created the database" option:

1. **Typo in HTML attribute**: `id=='collate'` should be `id='collate'` (double equals instead of single)
2. **Missing default value**: The collate field is only present in the "Have setup create the database" form path, but validation runs for both paths, causing `Undefined array key "collate"` and `TypeError`

## Changes

- Fix typo in collate select element: `id=='collate'` → `id='collate'`
- Add default value `utf8mb4_general_ci` for collate when not provided in the request

## Test plan

- [ ] Fresh install with "Have setup create the database" option
- [ ] Fresh install with "I have already created the database" option
- [ ] Verify collate validation works correctly in both paths

## Related Issues

Related to #10235 (Setup random fail with collate exception)

---

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -1127,7 +1127,7 @@ STP2TBLBOT;
                     }
 
                     // Default collate value when using pre-created database (collate field not in form)
-                    $_REQUEST['collate'] = $_REQUEST['collate'] ?? 'utf8mb4_general_ci';
+                    $_REQUEST['collate'] ??= 'utf8mb4_general_ci';
                     if (! $installer->collateNameIsValid($_REQUEST['collate'])) {
                         $pass_step2_validation = false;
                         $error_step2_message .= "$error - A collation name is required <br />\n";


setup.php
```diff
@@ -1126,6 +1126,8 @@ function cloneClicked() {
                         $error_step2_message .= "$error - A database name is required <br />\n";
                     }
 
+                    // Default collate value when using pre-created database (collate field not in form)
+                    $_REQUEST['collate'] = $_REQUEST['collate'] ?? 'utf8mb4_general_ci';
```
kojiromike:
```suggestion
                    $_REQUEST['collate'] ??= 'utf8mb4_general_ci';
```

fzzio: Thank you!

PR #10477, opened 2026-02-01
test(twig): add compilation test for all 268 Twig templates

### LINKED ISSUE #10475: Add Twig template compilation test to CI
OpenEMR has 268 Twig templates with autoescape globally disabled, but no CI step validates that templates compile successfully. Syntax errors and undefined filter/function usage are only caught at runtime when a page is visited.

**Proposed:** Add a PHPUnit test to the isolated test suite that compiles every `.twig` file against the Twig environment with all custom filters registered. This catches syntax errors, undefined filters (`|nonexistent`), and undefined functions at CI time.

This would run in the existing `isolated-tests.yml` workflow with no new dependencies.

**Follow-up work:**
- Snapshot tests for security-sensitive templates (render with fixture data, compare against stored HTML snapshots to catch escaping regressions)
- Convention for adding snapshot fixtures when modifying Twig templates

**AI Disclosure:** Yes

kojiromike:
Fixes #10475

Catch Twig template errors (syntax, undefined filters/functions/tests) at CI time rather than at runtime.

- Add `TwigTemplateCompilationTest` that compiles all 268 `.twig` templates against the real Twig environment using `compileSource()`
- Data provider discovers templates across `templates/`, `interface/forms/`, and `interface/modules/custom_modules/`
- Twig environment reuses `TwigContainer::getTwig()` with all custom extensions registered
- Registers a stub for `displayOptionClass` (added at runtime by `C_EncounterVisitForm`)
- No new dependencies, no CI config changes — file lives under `tests/Tests/Isolated/` which is already discovered by `phpunit-isolated.xml`

Claude Code (claude-opus-4-5-20251101) was used to generate the test file. The file header documents the author.

### DIFF
--- a/tests/Tests/Isolated/Common/Twig/TwigTemplateCompilationTest.php
+++ b/tests/Tests/Isolated/Common/Twig/TwigTemplateCompilationTest.php
@@ -1,0 +1,203 @@
+<?php
+
+/**
+ * Compile all .twig templates to verify syntax and filter/function/test references.
+ *
+ * compileSource() tokenizes, parses, and compiles a template to PHP without
+ * rendering it. The parse step validates that every referenced filter, function,
+ * and test is registered with the Twig environment. {% extends %}, {% include %},
+ * and {% import %} are recorded as nodes but not resolved until render time, so
+ * each template compiles independently.
+ *
+ * @package   OpenEMR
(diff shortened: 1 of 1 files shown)


bradymiller:

PR #10480, opened 2026-02-01
ci(codecov): use manual_trigger to prevent partial coverage reports

### LINKED ISSUE #10479: ci(codecov): replace static after_n_builds with wait_for_ci
## Problem

`codecov.notify.after_n_builds` is set to 14, a static constant that doesn't match any single CI context. This causes two problems:

1. **Partial coverage reports**: Fast tests (isolated tests) complete first and CodeCov immediately reports low coverage (~2%), alarming contributors. The full coverage number doesn't appear until slow integration tests finish ~30 minutes later.
2. **Inconsistent reporting**: The upload count varies by context (PRs, master pushes, nightly cron), so the static threshold is sometimes too high (CodeCov never reports) and sometimes too low (reports prematurely).

`wait_for_ci: true` does not reliably prevent this -- empirically, `after_n_builds` short-circuits it.

## Fix

Use `notify.manual_trigger: true` with a fan-in workflow that calls `send-notifications` only after all coverage-uploading workflows complete.

kojiromike:
Fixes #10479

## Description

CodeCov's `after_n_builds` and `wait_for_ci` settings don't reliably prevent partial coverage reporting on PRs. Contributors see a misleading low coverage number from fast tests while slow tests with higher coverage are still running.

Replace `after_n_builds` with `manual_trigger` and a fan-in workflow that calls `send-notifications` only after all coverage-uploading workflows have completed.

- Set `notify.manual_trigger: true` in `codecov.yml` so CodeCov sends no notifications until explicitly triggered
- Add `codecov-notify.yml` fan-in workflow that triggers on `workflow_run` completion, checks if all coverage workflows are done, and calls `send-notifications`

## AI disclosure

Yes

### DIFF
--- a/.github/workflows/codecov-notify.yml
+++ b/.github/workflows/codecov-notify.yml
@@ -1,0 +1,74 @@
+##
+# Fan-in workflow that sends CodeCov notifications after all
+# coverage-uploading workflows complete. This prevents CodeCov
+# from reporting partial coverage while slow tests are still running.
+#
+# codecov.yml sets notify.manual_trigger: true, so CodeCov will not
+# send any PR comments or status checks until this workflow calls
+# send-notifications.
+
+name: CodeCov Notify
+
+on:
+  workflow_run:
+    workflows:
+    - Test All Configurations
+    - Isolated Tests
+    - Inferno Certification Test
+    - Test (Scheduled)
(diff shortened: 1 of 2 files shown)


bradymiller:

PR #10482, opened 2026-02-01
fix: #10481 EHR launch skip login screen

### LINKED ISSUE #10481: bug: SMART on FHIR EHR launch with skip authorization enabled displays login screen
### Describe the bug

The SMART on FHIR EHR launch with a client enabled to skip the authorization screen (re-use the same login context) displays the login screen instead of using the existing user login session.  The session logic was changed in 7.0.4 and this bug was introduced as part of that release.

### To Reproduce

1. Register a private SMART on FHIR application with the 'launch' scope.  (oe-module-g9-ccda-exporter was used)
2. Open Admin -> System -> Api Applications
3. Find the registered application
4. Open a patient from within the EHR and navigate to their patient dashboard
5. Scroll down to SMART Enabled Apps and expand the applications
6. Press the launch button next to the registered application
7. See the login screen

### Expected behavior
User is brought immediately to the application.

### Client configuration

- Browser: Chrome
- OpenEMR version: 7.0.4+

adunsulag:
Fixes the issues with the session switching from oauth2 to core as the logic was broken as part of the api refactor.  This fix resolves those issues.  I had to handle the session_id uniquely as I couldn't just instantiate new session objects with symfony which was the error in the prior logic. 

Fixes #10481

### DIFF
--- a/src/Common/Http/HttpSessionFactory.php
+++ b/src/Common/Http/HttpSessionFactory.php
@@ -108,7 +108,7 @@ class HttpSessionFactory implements SessionFactoryInterface
         // while we migrate the sessions to testable objects.
         if (!empty($_SESSION)) {
             foreach ($_SESSION as $key => $value) {
-                if ($key !== $session->getName()) { // Avoid overwriting session name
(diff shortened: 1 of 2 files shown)


adunsulag:
I'm interested to see how github rector & phpstan treat this.  Had to clear my rector and phpstan caches and run them on the entire project instead of individually to catch things as the tools applied to single files just always return as OK.  Will need to investigate later why that is happening.

PR #10484, opened 2026-02-01
docs(api): add CLI testing tip for special characters in passwords

### LINKED ISSUE #10488: docs: add note about special characters in passwords when testing OAuth2 via CLI
When testing the OAuth2 Password Grant endpoint from the command line with `curl`, bash interprets special characters in passwords (`!`, `$`, `\`) before sending the request. This causes silent authentication failures — the API receives a mangled password with no indication of why.

For example, a password like `Test123!` gets modified by bash history expansion to `Test123\!`, producing a "Failed Authentication" error that's difficult to diagnose.

The AUTHENTICATION.md documentation should note this pitfall and suggest workarounds (single quotes, disabling history expansion, URL encoding, or using a scripting language).

**AI Disclosure:** No

fzzio:
Closes #10488

## Summary
- Adds a tip to the Password Grant section in `Documentation/api/AUTHENTICATION.md`
- Warns developers about bash shell interpretation causing authentication failures when passwords contain special characters (`!`, `$`, `\`)
- Provides practical solutions (disable history expansion, use single quotes, URL-encode, or use scripting languages)

## Context
While implementing OAuth2 password grant for a mobile healthcare app, I discovered that bash history expansion silently escapes the `!` character. This causes the API to receive an incorrect password (e.g., `Test123\!` instead of `Test123!`), resulting in a "Failed Authentication" error without any indication of the root cause.

This is a shell behavior issue, not an OpenEMR bug. Production apps using HTTP libraries will not experience this problem.

## Test plan
- [ ] Verify the markdown renders correctly in the documentation
- [ ] Confirm the tip appears after the Password Grant parameters section

---

### DIFF
--- a/Documentation/api/AUTHENTICATION.md
+++ b/Documentation/api/AUTHENTICATION.md
@@ -644,9 +644,11 @@ curl -X POST -k \
 > **Solutions for modified commands:**
 > - Disable bash history expansion (`!`): `set +H` before running curl
 > - Use single quotes around arguments containing special characters
-> - For passwords containing both `!` and `'`, use `printf -v`:
+> - For passwords containing both `!` and `'`, use a heredoc:
 >   ```bash
->   printf -v password "%s" 'my!complex'\''password'
+>   read -r password <<'EOF'
+>   my!complex'password
+>   EOF
 >   curl ... --data-urlencode "password=$password"
 >   ```
 >


Documentation/api/AUTHENTICATION.md
```diff
@@ -639,6 +639,16 @@ curl -X POST -k \
 - `password`: User's password
```
kojiromike, line 639: The curl examples just above this tip already use single-quoted `--data-urlencode 'password=...'` arguments, which prevent bash from interpreting `!`, `$`, and `\`. So the scenario described here would only happen if someone deviates from the documented examples — e.g., using double quotes or `-d` instead of `--data-urlencode`.

Would it be worth reframing the tip to say something like "if you modify these examples (e.g., switching to double quotes or `-d`), be aware that..."? That way it's clear the existing examples already handle this correctly.

Documentation/api/AUTHENTICATION.md
```diff
@@ -639,6 +639,16 @@ curl -X POST -k \
...
 - `email`: Required for patient role
 
+> **CLI Testing Tip**: If your password contains special characters like `!`, `$`, or `\`, you may encounter authentication failures when testing with bash/curl due to shell interpretation.
+>
+> **Solutions:**
+> - Disable bash history expansion: `set +H` before running curl
```
kojiromike: A couple of accuracy concerns with these bullets:

1. "Use single quotes around the entire curl command" — wrapping the *entire* command in single quotes would make it a string literal, not an executable command. The existing examples already use single quotes around each `--data-urlencode` *argument*, which is the correct approach. Could this say "use single quotes around arguments containing special characters" instead?

2. "URL-encode special characters manually" — since the examples use `--data-urlencode`, curl already handles encoding. Manually pre-encoding `!` as `%21` *and* using `--data-urlencode` would double-encode it. This advice would only apply with `-d` instead of `--data-urlencode`.

Documentation/api/AUTHENTICATION.md
```diff
@@ -639,6 +639,16 @@ curl -X POST -k \
...
+> - Use single quotes around the entire curl command
```
kojiromike: Single quotes handle `!` well, but what about passwords containing both `!` and `'`? Double quotes won't work (history expansion), and single quotes can't contain a literal single quote.

Might be worth showing `printf -v` as the escape hatch — you can split the tricky characters across differently-quoted arguments:

```bash
printf -v password "%s's %s" 'Robby' 'password!'
curl ... --data-urlencode "password=$password"
```

History expansion runs before parameter expansion, so `$password` is safe inside double quotes even when the value contains `!`.

fzzio:
Thanks for the thorough review @kojiromike! I've updated the tip based on your feedback:

- Clarified that the documented examples already use single quotes with `--data-urlencode`, which handles most cases
- Reframed the tip to apply only when deviating from the documented examples
- Fixed the inaccuracy about "entire curl command" → now says "arguments containing special characters"
- Removed the manual URL-encoding suggestion (since `--data-urlencode` handles it)
- Added the `printf -v` solution for passwords containing both `!` and `'`

Let me know if you'd like any further adjustments!

Documentation/api/AUTHENTICATION.md
```diff
@@ -639,6 +639,19 @@ curl -X POST -k \
 - `password`: User's password
 - `email`: Required for patient role
 
+> **CLI Testing Tip**: The examples above use single-quoted `--data-urlencode 'password=...'` arguments, which prevent bash from interpreting special characters like `!`, `$`, and `\`. If you modify these examples (e.g., switching to double quotes or using `-d` instead of `--data-urlencode`), you may encounter authentication failures due to shell interpretation.
+>
+> **Solutions for modified commands:**
+> - Disable bash history expansion: `set +H` before running curl
```
kojiromike: Nitpick: `set +H` only disables history expansion (the `!` character). It doesn't prevent `$` from triggering variable expansion or `\` from acting as an escape — both of which are listed in the opening sentence as problematic characters.

Might be worth scoping this to just `!`, e.g.:

```suggestion
> - Disable bash history expansion (`!`): `set +H` before running curl
```

Documentation/api/AUTHENTICATION.md
```diff
@@ -639,6 +639,19 @@ curl -X POST -k \
...
+> - Use single quotes around arguments containing special characters
```
kojiromike, line 646: This bullet is essentially restating what the examples already do (and what the opening sentence just explained). Since the tip's premise is "if you deviate from these examples," telling someone to go back to single quotes is circular — they'd just use the original examples.

Would it make sense to drop this bullet and keep just the two actionable alternatives (`set +H` and the heredoc) that actually help when single quotes aren't an option?

Documentation/api/AUTHENTICATION.md
```diff
@@ -639,6 +639,19 @@ curl -X POST -k \
...
+> - For passwords containing both `!` and `'`, use `printf -v`:
+>   ```bash
+>   printf -v password "%s" 'my!complex'\''password'
+>   curl ... --data-urlencode "password=$password"
+>   ```
```
kojiromike: Claude suggested using a heredoc here instead — `read -r` with a single-quoted heredoc delimiter is fully POSIX, whereas `printf -v` is bash-specific. The heredoc also lets the password appear exactly as-is with no quoting gymnastics:

```suggestion
> - For passwords containing both `!` and `'`, use a heredoc:
>   ```bash
>   read -r password <<'EOF'
>   my!complex'password
>   EOF
>   curl ... --data-urlencode "password=$password"
>   ```
```

kojiromike (commented):
The framing is much clearer now. I still have a few observations on the updated content, and Claude suggested a better way than printf to handle the really complex one.

PR #10486, opened 2026-02-01
fix(billing): prevent sending blank statement emails

### LINKED ISSUE #10485: bug: emailLogin() sends blank email when message is empty
### Describe the bug

`emailLogin()` in `interface/billing/sl_eob_search.php` does not validate its `$message` parameter. If called with an empty string, it sends a blank email to the patient with the subject "[Facility Name] Patient Statement Bill" and an empty body.

### Expected behavior

`emailLogin()` should return early (or throw) when `$message` is empty rather than sending a blank email to a patient.

kojiromike:
Fixes #10485

`emailLogin()` silently sent blank emails when called with an empty or markup-only message, and all failure paths returned `false` which callers ignored.

- `emailLogin()` throws instead of returning `false`: `InvalidArgumentException` for empty message (contract violation), `RuntimeException` for email validation, config, and send failures
- Call site (moved into the per-patient loop by #8327) catches `RuntimeException` and sets `$alertmsg`, so one patient's failure doesn't abort the batch
- Remove dead `upload_file_to_client_email()` function (sole call site removed by #8327)
- Add type annotations: native param/return types and PHPDoc `non-empty-string` constraint
- Replace `empty()` checks with explicit null/empty-string checks

Claude Code (Claude Opus 4.5) was used to implement this fix.

### DIFF
--- a/interface/billing/sl_eob_search.php
+++ b/interface/billing/sl_eob_search.php
@@ -180,19 +180,30 @@ function era_callback(&$out): void
     }
 }
 
-function emailLogin($patient_id, $message)
+/**
+ * Email a patient statement bill.
+ *
+ * @param non-empty-string $message
+ * @throws \InvalidArgumentException if message has no text content
+ * @throws \RuntimeException if the email cannot be sent
+ */
+function emailLogin(int $patient_id, string $message): void
 {
+    if (trim(strip_tags($message)) === '') {
+        throw new InvalidArgumentException('Statement message has no text content');
+    }
+
     $patientData = sqlQuery("SELECT * FROM `patient_data` WHERE `pid`=?", [$patient_id]);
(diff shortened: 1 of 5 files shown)


interface/billing/sl_eob_search.php
```diff
@@ -180,19 +180,30 @@ function era_callback(&$out): void
...
 
-function emailLogin($patient_id, $message)
+/**
+ * Email a patient statement bill.
+ *
+ * @param non-empty-string $message
+ * @throws \InvalidArgumentException if message has no text content
+ * @throws \RuntimeException if the email cannot be sent
+ */
+function emailLogin(int $patient_id, string $message): void
```
stephenwaite, line 190: thanks for fixing @kojiromike ! While here should this function name be improved and disambiguated from this [private function](https://github.com/openemr/openemr/blob/master/src/Services/PatientAccessOnsiteService.php#L254)?

kojiromike: @stephenwaite not part of this PR IMO. That's why I created #10487

PR #10497, opened 2026-02-02
feat(telemetry): Additional datapoint collections.

### LINKED ISSUE #10290: feat(telemetry): Report population capture and environment flag

sjpadgett:
collect enabled encounter forms by name, directory and category.

Fixes #10290

### DIFF
--- a/src/Telemetry/TelemetryRepository.php
+++ b/src/Telemetry/TelemetryRepository.php
@@ -134,4 +134,16 @@ class TelemetryRepository
 
         return $modulesData;
     }
+
+    public function fetchEnabledEncounterForms(): array
+    {
+        $sql = <<<'SQL'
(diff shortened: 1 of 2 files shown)


sjpadgett:
Encounter click already collected. My memory shot! Oh wait, I could collect how much memory site is using.:)
I was going to collect Dashboard nav bar clicks but decided against as common tasks. Merging and I need to get Telemetry out of my brain!

PR #10498, opened 2026-02-02
test: add @codeCoverageIgnore to data provider methods

### LINKED ISSUE #10496: Add @codeCoverageIgnore to data providers and other uncoverable test code
Data provider methods run before PHPUnit's coverage instrumentation starts, so they always appear as uncovered lines. This creates noise in patch coverage reports — every new test with a provider gets flagged for ~30-60 "uncovered" lines that can never be covered.

Roughly 900-1,000 lines across ~22 test files are affected, primarily data provider method bodies.

### What to annotate

- **Data provider methods** (`#[DataProvider]` / `@dataProvider`) — add `@codeCoverageIgnore` to the method docblock
- **Environment-gated branches** (e.g., `if (getenv('UPDATE_FIXTURES') === '1')`) — use `@codeCoverageIgnoreStart` / `@codeCoverageIgnoreEnd`

### What NOT to annotate

- `setUp()` / `tearDown()` — these run within instrumentation scope and are covered
- Helper methods called from test method bodies — also covered
- Legacy code with low coverage — that's a different problem

kojiromike:
## Description

Fixes #10496

Add `@codeCoverageIgnore` to data provider methods in test files. Data providers run before PHPUnit's coverage instrumentation starts, so they always appear as uncovered lines. This eliminates false negatives in patch coverage reports.

## Changes

- Annotated 35 data provider methods across 19 test files with `@codeCoverageIgnore`
- No functional changes to any test logic

## Testing

- `php -l` passes on all 19 modified files
- Isolated test suite passes (938 tests, 2525 assertions)
- All pre-commit hooks pass (PHPCS, PHPStan, Rector, etc.)

## AI Disclosure

Yes

### DIFF
--- a/tests/Tests/E2e/GgUserMenuLinksTest.php
+++ b/tests/Tests/E2e/GgUserMenuLinksTest.php
@@ -71,6 +71,7 @@ class GgUserMenuLinksTest extends PantherTestCase
         }
     }
 
+    /** @codeCoverageIgnore Data providers run before coverage instrumentation starts. */
     public static function menuLinkProvider()
     {
         return [
--- a/tests/Tests/E2e/HhMainMenuLinksTest.php
+++ b/tests/Tests/E2e/HhMainMenuLinksTest.php
@@ -76,6 +76,7 @@ class HhMainMenuLinksTest extends PantherTestCase
         }
     }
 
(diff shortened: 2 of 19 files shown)


bradymiller:

PR #10515, opened 2026-02-02
chore: reused SingletonTrait at OEGlobalsBag

igormukhingmailcom:


src/Core/OEGlobalsBag.php
```diff
@@ -11,31 +11,18 @@
...
-            self::$instance = new OEGlobalsBag($GLOBALS);
-        }
-
-        return self::$instance;
-    }
-
-    public function __construct(array $parameters = [])
-    {
-        parent::__construct($parameters);
+        return new self($GLOBALS);
```
kojiromike, line 25: Should this use `new static(...)` instead of `new self(...)`? The return type is `static`, so if `OEGlobalsBag` were ever subclassed, `new self` would return the wrong type. The other `SingletonTrait` users (`EventAuditLogger`, `SessionWrapperFactory`) both use `new static()` in their `createInstance()` overrides.

```suggestion
        return new static($GLOBALS);
```

igormukhingmailcom: Hi, @kojiromike 

`EventAuditLogger` uses `new self` as it has custom `createInstance`.

`SessionWrapperFactory` has no custom  `createInstance` so uses `new static` from `SingletonTrait::createInstance`.

I would say rules are next:
- When class constructor **has no arguments** - we don't need to have custom `createInstance`. And in that case `SingletonTrait::createInstance` is used with its `new static`.
- We should use `new self` only at custom `createInstance`; and we have custom `createInstance` only when class constructor **has arguments**;

For this particular case - we should not care if it will be subclassed until it actually is. If ever subclassed and constructors arguments match at both classes - then yes we should use `new static` and suppress phpstan warnings (`// @phpstan-ignore-line new.static`). If constructor arguments will be different - subclass will have its own `createInstance` with `new self`

kojiromike: TBH I'm not really satisfied with the "we shouldn't care if it will be subclassed until it actually is," argument, since it could violate the principle of least surprise at that time. I guess we can make the whole argument `moot` if we make it `final`, though.

igormukhingmailcom: Yeah, we can make it final.

Better this way - `/** @final */ class OEGlobalsBag {}` rather than `final class OEGlobalsBag {}`, to  be able to mock final classes at tests. 

PHPStan will report if we try to extend it `class OEGlobalsBag2 extends OEGlobalsBag {}`:
`Class OpenEMR\Core\OEGlobalsBag2 extends @final class OpenEMR\Core\OEGlobalsBag - class.extendsFinalByPhpDoc`

Issue #10149, opened 2026-01-09
refactor(db): Conversion to DBAL


Firehed:
Parent issue to track remaining conversion steps following #9944/#9980

`sql.inc.php`:

- [ ] `sqlStatement`
- [ ] `sqlStatementThrowException`
- [ ] `sqlGetLastInsertId`
- [ ] `sqlStatementNoLog`
- [ ] `sqlStatementCdrEngine`
- [ ] `sqlFetchArray`
- [ ] `sqlGetAssoc`
- [ ] `sqlInsert`
- [ ] `sqlQuery`
- [ ] `sqlQueryNoLogIgnoreError`
- [ ] `sqlQueryCdrEngine`
- [ ] `sqlInsertClean_audit`
- [ ] `sqlGetLastError`
- [ ] `sqlGetLastErrorNo`
- [ ] `sqlListFields`
- [ ] `sqlNumRows`
- [ ] `sqlQ`
- [ ] `sqlClose`
- [ ] `get_db`
- [ ] `generic_sql_select_db`
- [ ] `generic_sql_affected_rows`
- [ ] `generic_sql_insert_id`
- [ ] `sqlBeginTrans`, `sqlCommitTrans`, `sqlRollbackTrans`
- [ ] `getPrivDb`
- [ ] `privStatement`
- [ ] `privQuery`
- [ ] `edi_generate_id`

`QueryUtils`:

- [ ] `listTableFields`
- [ ] `escapeTableName`
- [ ] `escapeColumnName`
- [ ] `fetchRecordsNoLog`
- [ ] `fetchTableColumn`
- [ ] `fetchSingleValue`
- [ ] `fetchRecords`
- [ ] `fetchTableColumnAssoc`
- [ ] `fetchArrayFromResultSet`
- [ ] `sqlStatementThrowException`
- [ ] `existsTable`
- [ ] `sqlInsert`
- [ ] `selectHelper`
- [ ] `ediGenerateId`
- [ ] `startTransaction`/`commitTransaction`/`rollbackTransaction`
- [ ] `getLastInsertId`
- [ ] `querySingleRow`
- [ ] `escapeLimit`

Direct `ADOdb` touch-points outside of db wrappers:
- [ ] Numerous places within EventAuditLogger (#9881)
- [ ] ccr/transmitCCD
- [ ] interface/main/calendar/config.php ? (this seems to load the lib and then not use it)
- [ ] interface/patient_file/encounter/cach_receipt.php
[...]

PR #10516, opened 2026-02-02
refactor(db): Make QueryUtils the canonical implementation

Firehed:
In service of #10149

Inverts the wrapper relationship between `library/sql.inc.php` (and `library/formdata.inc.php`) and `src/Common/Database/QueryUtils.php` so that the namespaced class owns the actual database logic and the legacy global functions become thin wrappers that delegate to it.

Not _all_ functions have been updated, but it makes substantial progress towards having the logic in a class. This will make switching to DBAL quite a bit easier since it simplifies connection management. I've also left the GLOBALS stuff alone for now to keep this easy to review; I'll tackle that in a separate PR.

  - Move the implementation of 8 functions from `sql.inc.php`/`formdata.inc.php` into `QueryUtils`, with the legacy bare functions now delegating to `QueryUtils`:
  - `sqlBeginTrans()` → `QueryUtils::startTransaction()`
  - `sqlCommitTrans()` → `QueryUtils::commitTransaction()`
  - `sqlRollbackTrans()` → `QueryUtils::rollbackTransaction()`
  - `sqlFetchArray()` → `QueryUtils::fetchArrayFromResultSet()`
  - `generate_id()` → `QueryUtils::generateId()`
  - `edi_generate_id()` → `QueryUtils::ediGenerateId()`
  - `sqlGetLastInsertId()` → `QueryUtils::getLastInsertId()`
  - `sqlStatementThrowException()` → `QueryUtils::sqlStatementThrowException()`
  - `escape_table_name()` → `QueryUtils::escapeTableName()`
  - The throw-exception path of `sqlStatementNoLog()` also now delegates to `QueryUtils::sqlStatementThrowException(noLog: true)`; the `HelpfulDie` fallback path remains in `sql.inc.php`
  - Replace all internal bare function calls within `QueryUtils` (`sqlFetchArray`, `sqlStatementNoLog`, `escape_table_name`, etc.) with their `self::` equivalents
  - No behavioral changes; all external callers (both the global functions and `QueryUtils` static methods) continue to work identically

All changes in this PR were generated with the assistance of Claude (Anthropic). Each commit is co-authored accordingly. The changes are mechanical refactoring (moving existing logic between files and updating call sites) with no new logic introduced.

### DIFF
--- a/library/sql.inc.php
+++ b/library/sql.inc.php
@@ -213,11 +213,6 @@ function sqlGetLastInsertId()
 */
 function sqlStatementNoLog($statement, $binds = false, $throw_exception_on_error = false)
 {
-    // Below line is to avoid a nasty bug in windows.
-    if (empty($binds)) {
-        $binds = false;
-    }
-
     if ($throw_exception_on_error) {
         return QueryUtils::sqlStatementThrowException($statement, $binds, noLog: true);
     }


library/sql.inc.php
```diff
@@ -230,14 +218,14 @@ function sqlStatementNoLog($statement, $binds = false, $throw_exception_on_error
         $binds = false;
     }
 
+    if ($throw_exception_on_error) {
```
kojiromike, line 216: The early return to `QueryUtils::sqlStatementThrowException()` happens after the `empty($binds)` normalization on lines 229-231, so `$binds` gets normalized twice (once here, once inside `QueryUtils::sqlStatementThrowException`). Harmless, but the early return could move above the normalization since `QueryUtils` handles it internally anyway.

Firehed: Ah yes, I meant to smoke out some of those redundant "empty for windows" things. I think there are more on the follow-up PR. Will fix this one.

PR #10520, opened 2026-02-03
refactor(db): Move more logic from sql.inc to QueryUtils

Firehed:
Continuation of #10516; progress towards #10149.

Note that this doesn't attempt to fix any of the existing anti-patterns. Some PHPStan tests suggest to me there are _many thousands_ of places that aren't correctly handling query failures (e.g. the `return false` paths).

  The legacy `sql.inc.php` wrapper functions duplicated query execution logic that already exists in `QueryUtils`. This PR makes the legacy functions thin wrappers around `QueryUtils`, preserving the existing `HelpfulDie` error-handling behavior via catch blocks around `SqlQueryException`.

  - Delegate `sqlStatement()`, `sqlStatementNoLog()`, `sqlQuery()`, `sqlQueryNoLog()`, `sqlInsert()`, and `sqlInsertClean_audit()` to their `QueryUtils` counterparts, catching `SqlQueryException` to preserve `HelpfulDie` fallback behavior
  - Remove unused `sqlQueryNoLogIgnoreError()` and `getSqlLastErrorNo()`
  - Add `$log` parameter to `QueryUtils::querySingleRow()` to support the no-log path[^log]
  - Fix docblock `@param`/`@return` types across `sql.inc.php` and `QueryUtils.php` (correct return types to `ADORecordSet`/`array|false`, add missing `@throws`, remove phantom params)
  - Update PHPStan baselines to reflect the changes

  All commits were co-authored with Claude. The changes are mechanical delegations and docblock fixes rather than novel logic.

[^log]: I intentionally went against the grain with `bool $log = true` instead of `bool $noLog = false` since it is IMO _a lot_ easier to reason about

### DIFF
--- a/.phpstan/baseline/class.notFound.php
+++ b/.phpstan/baseline/class.notFound.php
@@ -2007,6 +2007,11 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../src/Common/ORDataObject/ORDataObject.php',
 ];
 $ignoreErrors[] = [
+    'message' => '#^Call to method RecordCount\\(\\) on an unknown class OpenEMR\\\\Common\\\\Database\\\\ADORecordSet\\.$#',
+    'count' => 2,
+    'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
+];
+$ignoreErrors[] = [
     'message' => '#^Parameter \\$uuid of method OpenEMR\\\\Common\\\\Uuid\\\\UuidRegistry\\:\\:getRegistryRecordForUuid\\(\\) has invalid type OpenEMR\\\\Common\\\\Uuid\\\\binary\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
--- a/.phpstan/baseline/openemr.deprecatedSqlFunction.php
+++ b/.phpstan/baseline/openemr.deprecatedSqlFunction.php
@@ -1543,7 +1543,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Use QueryUtils\\:\\:querySingleRow\\(\\) instead of sqlQueryNoLog\\(\\)\\.$#',
-    'count' => 5,
+    'count' => 4,
     'path' => __DIR__ . '/../../interface/globals.php',
 ];
 $ignoreErrors[] = [
@@ -6903,11 +6903,6 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Use QueryUtils\\:\\:fetchRecordsNoLog\\(\\) instead of sqlStatementNoLog\\(\\)\\.$#',
-    'count' => 6,
-    'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Use QueryUtils\\:\\:fetchRecords\\(\\) or QueryUtils\\:\\:fetchArrayFromResultSet\\(\\) instead of sqlFetchArray\\(\\)\\.$#',
     'count' => 2,
     'path' => __DIR__ . '/../../src/Common/Uuid/UuidRegistry.php',
 ];
--- a/.phpstan/baseline/openemr.forbiddenErrorLog.php
+++ b/.phpstan/baseline/openemr.forbiddenErrorLog.php
@@ -38,7 +38,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Use OpenEMR\\\\Common\\\\Logging\\\\SystemLogger instead of error_log\\(\\)\\.$#',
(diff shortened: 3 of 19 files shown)


library/sql.inc.php
```diff
@@ -209,21 +201,19 @@ function sqlGetLastInsertId()
 *
 * @param  string  $statement  query
 * @param  array   $binds      binded variables array (optional)
-* @return recordset
+* @param  bool    $throw_exception_on_error  if true throws SqlQueryException instead of calling HelpfulDie
+* @return ADORecordSet
```
kojiromike, line 205: does phpstan support throws annotations?

Firehed: Technically yes, but they're a little dubious with how PHP as a whole works. I've never found them to be useful in practice.

kojiromike:
@Firehed icymi I think these test failures are legit this time.

Firehed:
@kojiromike It looks like the regression may have been in the previous refactor (or latent from something else, I didn't dig TOO far) but switching all of the legacy utils to route through it made it active or more prominent. See cb2897b  for more details.

y'know, assuming I get green checkmarks this time.

Firehed:
kicking back to draft until I can fake-bisect this to track down the problems. First of several at #10621.

Firehed:
```
openemr-1  | NOTICE: PHP message: PHP Fatal error:  Uncaught TypeError: OpenEMR\Common\Database\QueryUtils::querySingleRow(): Argument #2 ($params) must be of type array, int given, called in /usr/share/nginx/html/openemr/library/sql.inc.php on line 315 and defined in /usr/share/nginx/html/openemr/src/Common/Database/QueryUtils.php:402
openemr-1  | Stack trace:
openemr-1  | #0 /usr/share/nginx/html/openemr/library/sql.inc.php(315): OpenEMR\Common\Database\QueryUtils::querySingleRow('SELECT patients...', 1)
openemr-1  | #1 /usr/share/nginx/html/openemr/src/Services/PatientService.php(927): sqlQuery('SELECT patients...', 1)
openemr-1  | #2 /usr/share/nginx/html/openemr/interface/patient_file/summary/demographics.php(88): OpenEMR\Services\PatientService->touchRecentPatientList(Array)
openemr-1  | #3 {main}
```
aha, found it. didn't realize logs got stuffed somewhere else. I think this revealed an existing logic bug somewhere through the accidental addition of a true typehint.

Firehed:
@kojiromike This ought to be good now if GH stays online. Turns out I'd accidentally introduced a new path to use correct typehints, and there are some oddball paths where `$params` is neither `false` nor an `array`; it seems ADODB has some internal handling for this case to wrap a single argument into an array. I removed that native hint for now and will file a ticket to resolve those areas (I swear I already did that on a different branch, but whatever)

Firehed:
The docblock signature corrections fixed a bunch of phpstan errors (yay) but also revealed a bunch of new issues (boo). Note hugely surprising; there's a LOT of code that doesn't handle DB errors well or at all. At least we know about them now.

src/Common/Database/QueryUtils.php
```diff
@@ -358,9 +360,25 @@ public static function getLastInsertId()
...
+     * @param  string  $sql     query
+     * @param  array   $params  binded variables array (optional)
+     * @param  bool    $log     if true the sql statement is logged, false bypasses the database logger
+     * @throws SqlQueryException Thrown if there is an error in the database executing the statement
+     * @return array|false
+     */
+    public static function querySingleRow(string $sql, $params = [], bool $log = true)
     {
-        $result = self::sqlStatementThrowException($sql, $params);
+        /** @var mixed @params */
```
kojiromike: Small typo in the PHPDoc annotation — should this be `$params` instead of `@params`?

```suggestion
        /** @var mixed $params */
```

src/Common/Database/QueryUtils.php
```diff
@@ -247,6 +248,7 @@ public static function existsTable($tableName)
      */
     public static function sqlInsert($statement, $binds = [])
     {
+        // execute/true
```
kojiromike: Is this comment intentional? It appears in a few places and I'm not sure what it's meant to convey.

library/sql.inc.php
```diff
@@ -177,10 +168,12 @@ function sqlStatement($statement, $binds = false)
  *
  * @param  string  $statement  query
  * @param  array   $binds      binded variables array (optional)
- * @return recordset
+ * @throws SqlQueryException Thrown if there is an error in the database executing the statement
+ * @return ADORecordSet
  */
 function sqlStatementThrowException($statement, $binds = false)
 {
+    // execute/true
```
kojiromike: is `// execute/true` a debug breadcrumb, or does it serve a documentation purpose I'm missing?

Firehed: whoops yes those were left in by mistake while debugging. will remove.

Firehed:
fixed issues, thx for catching them!

kojiromike:
@Firehed conflicts… 😢

Firehed:
Fixed (for a minute!)

PR #10530, opened 2026-02-03
fix(session): breakage when add new patient via new patient UI

### LINKED ISSUE #10529: Break when adding a new patient. (appears secondary to session breakage)
branch rel-800:
After add a new patient (when click submit) white screen of death and following error:
```
PHP Fatal error:  Uncaught RuntimeException: Failed to start the session: already started by PHP. in /var/www/localhost/htdocs/openemr/src/Common/Session/SessionUtil.php:105\nStack trace:\n#0 /var/www/localhost/htdocs/openemr/src/Common/Session/SessionUtil.php(131): OpenEMR\\Common\\Session\\SessionUtil::sessionStartWrapper()\n#1 /var/www/localhost/htdocs/openemr/src/Common/Session/SessionUtil.php(191): OpenEMR\\Common\\Session\\SessionUtil::coreSessionStart()\n#2 /var/www/localhost/htdocs/openemr/src/Common/Session/PatientSessionUtil.php(55): OpenEMR\\Common\\Session\\SessionUtil::setUnsetSession()\n#3 /var/www/localhost/htdocs/openemr/library/pid.inc.php(18): OpenEMR\\Common\\Session\\PatientSessionUtil::setPid()\n#4 /var/www/localhost/htdocs/openemr/interface/new/new_comprehensive_save.php(80): setpid()\n#5 {main}\n  thrown in /var/www/localhost/htdocs/openemr/src/Common/Session/SessionUtil.php on line 105, referer: http://localhost:8300/interface/new/new.php
```

bradymiller:
fixes #10529

To be honest, this fixes but not totally sure what I am doing here (seems to make sense to save the session before do start wrapper since that seems to fail if there is an active session). To repeat though, not sure my brain understands  how the session code is working.

Need more eyes on this.

### DIFF
--- a/src/Common/Session/SessionUtil.php
+++ b/src/Common/Session/SessionUtil.php
@@ -166,6 +166,12 @@ class SessionUtil
 
     public static function unsetSession($session_key_or_array): void
     {
+        // Since our default is read_and_close the session shouldn't be active here.
+        if (session_status() === PHP_SESSION_ACTIVE) {
+            // ensure the session file is written from a previous
+            // session open for write.
+            session_write_close();
+        }
         self::coreSessionStart($GLOBALS['webroot'], false);
         if (is_array($session_key_or_array)) {
             foreach ($session_key_or_array as $value) {
@@ -188,6 +194,12 @@ class SessionUtil
 
     public static function setUnsetSession($setArray, $unsetArray): void
     {
+        // Since our default is read_and_close the session shouldn't be active here.
+        if (session_status() === PHP_SESSION_ACTIVE) {
+            // ensure the session file is written from a previous
+            // session open for write.
+            session_write_close();
+        }
         self::coreSessionStart($GLOBALS['webroot'], false);
         foreach ($setArray as $key => $value) {
             $_SESSION[$key] = $value;


adunsulag:
@bradymiller This ensures nothing breaks / dies, but means we have some other code logic flaw elsewhere as we shouldn't be reaching the exception.  I'll look into this as I wasn't having problems creating users locally.

src/Common/Session/SessionUtil.php
```diff
@@ -166,6 +166,12 @@ public static function setSession($session_key_or_array, $session_value = null):
 
     public static function unsetSession($session_key_or_array): void
     {
+        // Since our default is read_and_close the session shouldn't be active here.
+        if (session_status() === PHP_SESSION_ACTIVE) {
+            // ensure the session file is written from a previous
+            // session open for write.
+            session_write_close();
+        }
         self::coreSessionStart($GLOBALS['webroot'], false);
```
kojiromike, line 175: Should `coreSessionStart()` itself handle this guard instead of relying on each caller to do it? All three methods (`setSession`, `unsetSession`, `setUnsetSession`) now have the identical block before calling `coreSessionStart()`. If `coreSessionStart()` handled the close-if-active check internally, it would be impossible for a future caller to forget it.

That said, `coreSessionStart()` is also called from `switchToCoreSession()` which does its own `session_write_close()` first, so moving the guard into `coreSessionStart()` would be a no-op there — still safe.

Not a blocker for this fix, just a thought for a follow-up to avoid the pattern spreading further.

adunsulag: I was coming to the same conclusion stepping through the code.  The fact we have to close the session feels like a pretty bad code smell, but at this point I think we do it to head off problems.  Will create more fs churn (or network churn w/ redis).

adunsulag: @bradymiller I'm going to test things out with the guard being in coreSessionStart.  Will also test the one time login tokens that Jerry mentioned as well as part of that.

bradymiller:
I think problem with putting it in core is some scripts (like login stuff)
do not use the save close mechanism.

adunsulag:
@bradymiller I'm not seeing problems in the login screen on both portal and user with closing the session in coreSessionStart.  I'm going through every link and GET/POST action I can trigger from the UX, so far not seeing an issue.

bradymiller:
@adunsulag , It is the scripts that have the `$sessionAllowWrite = true;` prior to the globals inlcude that we were concerned about in the past. Those unset/set functions are basically only used in scripts without this setting, so no risk in converting the session from allowwrite to not allow write. hope that makes sense.

adunsulag:
@adunsulag I'm not following.  Are you saying if we close the session it will change the behavior to where $allowWrite becomes true? 

I pushed out an alternative commit with the proposed core session change: https://github.com/openemr/openemr/pull/10535

I am concerned if you think there will be some session locking / timeout issues with my proposed fix as I know the allowWrite stuff was supposed to mitigate that.

PR #10537, opened 2026-02-03
fix(portal): one-time token for portal document notification broken

### LINKED ISSUE #10517: bug(portal): one-time token for portal document notification broken
### Describe the bug

The one-time for patient notification that a portal document needs attention is broken for both release v8 and 8.0.1-dev.
For dev version onetime is being redirected to openemr login
For v8.0.0 getting a fatal error
`
Fatal error: Uncaught OpenEMR\Common\Auth\Exception\OneTimeAuthException: Onetime decode failed Onetime auth: 007IzS92CV91uu1A1PtJc4iceosNtpQCiHgXYfcNh+2lVRVAakjvZpHJ0MtOX5ZFXtb7aJ9+pnfAHQdzsQRTZS3/C35XGNcZU2SRaa/O62+vBIAFylO5AwMtMvQ0AUMSadY in C:\xampp\htdocs\openemr\src\Common\Auth\OneTimeAuth.php:172 Stack trace: #0 C:\xampp\htdocs\openemr\portal\index.php(106): OpenEMR\Common\Auth\OneTimeAuth->decodePortalOneTime('007IzS92CV91uu1...', NULL, false) #1 {main} thrown in C:\xampp\htdocs\openemr\src\Common\Auth\OneTimeAuth.php on line 172
`
I'm pretty sure works in 7.0.4 but I'm not going to retest that.

### To Reproduce

- Ensure a valid email account is setup
- Ensure email is enabled in Choices and a valid test email address is enter as patient contact email
- Install and enable Email service in faxsms module
- Create or have a document template assigned and scheduled to a test patient.
- Be sure the Profile is active Profiles in Portal
- Select template from Patient Assigned Templates and click Notify.

### Client configuration

- Browser:
- OpenEMR version: 8.0.0
- Operating system:

rollingventures:
Encode one-time links with RFC3986 so + and / in tokens are sent as %2B and %2F. Query string parsing otherwise turns + into space and corrupts the token, causing decode failure and redirect to login.

- OneTimeAuth::encodeLink(): use PHP_QUERY_RFC3986 for all onetime query strings (register, reset_password, service_auth).
- portal/index.php: normalize service_auth and target by replacing space with + when reading from GET/POST so existing links still work.

Fixes #10517

### DIFF
--- a/portal/index.php
+++ b/portal/index.php
@@ -94,6 +94,7 @@ $redirectUrl = $_REQUEST['redirect'] ?? '';
  * The embedded pid in token is compared to the token looked up result pid.
  * Also verified as the portal account id is rebuilt from patient data
  * and compared to portal credential account id lookup.
+ * Restore + in token: query string parsing turns + into space (issue #10517).
  * */
 if (!empty($_REQUEST['service_auth'] ?? null)) {
     $oneTime = new OneTimeAuth();
@@ -102,15 +103,16 @@ if (!empty($_REQUEST['service_auth'] ?? null)) {
         // we also implement this mechanism in order to handle Same-Site cookie blocking when being referred by
         // an external site domain.  We used to auto process via GET but now we submit via the POST in order to make it
         // a same site cookie origin request. This is a workaround for the Same-Site cookie blocking.
-        $token = $_GET['service_auth'];
-        $ot = $oneTime->decodePortalOneTime($token, null, false);
+        $token = str_replace(' ', '+', (string) $_GET['service_auth']);
+        $target = isset($_GET['target']) ? str_replace(' ', '+', (string) $_GET['target']) : null;
+        $ot = $oneTime->decodePortalOneTime($token, $target, false);
         $pin_required = $ot['actions']['enforce_auth_pin'] ? 1 : 0;
         CsrfUtils::setupCsrfKey($session->getSymfonySession());
         $twig = new TwigContainer(null, $globalsBag->get('kernel'));
         echo $twig->getTwig()->render('portal/login/autologin.html.twig', [
             'action' => $globalsBag->getString('web_root') . '/portal/index.php',
-            'service_auth' => $_GET['service_auth'],
-            'target' => $_GET['target'] ?? null,
+            'service_auth' => $token,
+            'target' => $target,
             'csrf_token' => CsrfUtils::collectCsrfToken('autologin', $session->getSymfonySession()),
             'pagetitle' => xl("OpenEMR Patient Portal"),
             'images_static_relative' => $globalsBag->get('images_static_relative') ?? '',
@@ -118,8 +120,8 @@ if (!empty($_REQUEST['service_auth'] ?? null)) {
         ]);
         exit;
     } elseif (!empty($_POST['service_auth'] ?? null)) {
-        $token = $_POST['service_auth'];
-        $redirect_token = $_POST['target'] ?? null;
+        $token = str_replace(' ', '+', (string) $_POST['service_auth']);
+        $redirect_token = isset($_POST['target']) ? str_replace(' ', '+', (string) $_POST['target']) : null;
         $csrfToken = $_POST['csrf_token'] ?? null;
         try {
             if (!CsrfUtils::verifyCsrfToken($csrfToken, 'autologin', $session->getSymfonySession())) {
--- a/src/Common/Auth/OneTimeAuth.php
+++ b/src/Common/Auth/OneTimeAuth.php
@@ -232,28 +232,31 @@ class OneTimeAuth
         if (stripos((string) $site_addr, "?") === false) {
             $format = "%s?%s";
         }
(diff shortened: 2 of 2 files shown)


portal/index.php
```diff
@@ -102,24 +103,25 @@
         // we also implement this mechanism in order to handle Same-Site cookie blocking when being referred by
         // an external site domain.  We used to auto process via GET but now we submit via the POST in order to make it
         // a same site cookie origin request. This is a workaround for the Same-Site cookie blocking.
-        $token = $_GET['service_auth'];
-        $ot = $oneTime->decodePortalOneTime($token, null, false);
+        $token = str_replace(' ', '+', (string) $_GET['service_auth']);
+        $target = isset($_GET['target']) ? str_replace(' ', '+', (string) $_GET['target']) : null;
```
kojiromike, line 107: Is this just manual url encoding? Should we use a url encoding function?

rollingventures: We're not doing URL encoding.  We are correcting values that PHP has already decoded. When PHP parses the query string it turns + into space (per application/x-www-form-urlencoded). So $_GET['service_auth'] already has spaces where the original token had +. There's no standard function that undoes that; URL encoding functions (e.g. rawurlencode/rawurldecode) are for building or decoding percent-encoded segments, not for reversing the +→space conversion. We know the token is base64-like (can contain +, no literal spaces), so str_replace(' ', '+', ...) is "a" fix on the read path.

kojiromike: Hmm, but that still doesn't seem right. The thing providing the URL is broken, because it should be encoding the `+` character as `%2B`, so it gets properly decoded to `+` in that case.

kojiromike: (Or, it should be using http-safe base64.)

src/Common/Auth/OneTimeAuth.php
```diff
@@ -232,28 +232,31 @@ private function encodeLink($site_addr, $token_encrypt, $encrypted_redirect = nu
         if (stripos((string) $site_addr, "?") === false) {
             $format = "%s?%s";
         }
+        // Use RFC3986 so + and / in base64-like tokens are encoded as %2B and %2F,
```
kojiromike, line 235: `PHP_QUERY_RFC3986` has existed since PHP 5.4, and OpenEMR requires 8.2+. Could this drop the `defined()` guard and use the constant directly?

```suggestion
        $enc_type = PHP_QUERY_RFC3986;
```

portal/index.php
```diff
@@ -102,24 +103,25 @@
         // we also implement this mechanism in order to handle Same-Site cookie blocking when being referred by
         // an external site domain.  We used to auto process via GET but now we submit via the POST in order to make it
         // a same site cookie origin request. This is a workaround for the Same-Site cookie blocking.
-        $token = $_GET['service_auth'];
-        $ot = $oneTime->decodePortalOneTime($token, null, false);
+        $token = str_replace(' ', '+', (string) $_GET['service_auth']);
```
kojiromike, line 106: With the `encodeLink` fix using RFC3986, newly generated links will have `+` encoded as `%2B` and PHP will decode them correctly. The `str_replace` workaround here (and in the POST branch on line 123) shouldn't be needed — one-time tokens are short-lived and consumed on first use, so there's no backward-compat window to cover. Dropping the read-side changes would keep this a clean, single-layer fix in `encodeLink`.

portal/index.php
```diff
@@ -94,6 +94,7 @@
  * The embedded pid in token is compared to the token looked up result pid.
  * Also verified as the portal account id is rebuilt from patient data
  * and compared to portal credential account id lookup.
+ * Restore + in token: query string parsing turns + into space (issue #10517).
```
kojiromike, line 97: If the read-side changes are dropped, this comment should go too.

kojiromike (commented):
The write-side fix (`PHP_QUERY_RFC3986` in `encodeLink`) is the right approach — it percent-encodes `+` and `/` in the base64 tokens so PHP decodes them correctly on the read side. The read-side `str_replace` changes in `portal/index.php` should be dropped; one-time tokens are short-lived and consumed on first use, so there are no old links to worry about. That would make this a clean, minimal fix entirely in `OneTimeAuth::encodeLink()`.

sjpadgett:
@rollingventures I found issue. It's in my src/Common/Auth/OneTimeAuth.php processOnetime() where session var not set properly to match same schema used in portal so auth is invavid.
How do you want to handle? I can push changes to you or let you go and refactor all session needs and I'll test for you.

rollingventures:
> @rollingventures I found issue. It's in my src/Common/Auth/OneTimeAuth.php processOnetime() where session var not set properly to match same schema used in portal so auth is invavid. How do you want to handle? I can push changes to you or let you go and refactor all session needs and I'll test for you.

@sjpadgett Please push the changes on your PR and then we can review accordingly.  I will close this out as its still a DRAFT.

sjpadgett:
Will do. I first need to fix the v8.0.0 version which may have to be a separate PR.

PR #10541, opened 2026-02-03
refactor(db): Stop accepting boolean in $binds params

Firehed:
Stacked on #10516 and #10520. Progress towards #10149.

Avoids default `false` in various SQL $binds params, better aligning with intended usage.

- Changed default value in function signatures
- Found and fixed a few spots where the functions were being called with incorrect values

Yes

Claude used.


Firehed:
I still think this is worthwhile, but there's too much divergence to deal with this branch.

PR #10551, opened 2026-02-04
refactor(ci): simplify phpstan workflow using failure() condition

Firehed:

Simplifies the PHPStan CI workflow by replacing the workaround pattern of `continue-on-error` + manual failure check with the cleaner `if: failure()` condition.

- Remove `continue-on-error: true` from the PHPStan analyze step - let it fail the job naturally
- Use `if: failure()` on baseline generation and upload steps instead of checking step outcome
- Remove the `Check PHPStan Results` step that ran `false` to fail the job
- **Test commit included**: Adds a deliberate PHPStan error to verify the workflow works correctly (will be removed before merge)

Yes

Co-authored with Claude Opus 4.5. The changes are straightforward workflow simplifications.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -67,6 +67,8 @@ jobs:
     - name: PHPStan Baseline
       if: failure()
       run: composer run phpstan-baseline
+    # Upload the resulting baseline as an artifact. This is intended mostly for
+    # debugging and isn't actively used by GH workflows.
     - name: Upload PHPStan Baseline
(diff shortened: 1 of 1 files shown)


Firehed:
Two side notes:

- I found the annotations not reliably showing up previously; when I tested this on an intentionally-broken commit (reverted), I saw them
- The whole thing ought to run a touch faster since we eliminated some steps

Firehed:
Test failure is a flake unrelated to this change.

PR #8288, opened 2025-04-18
Fixing recalls workflow

bryanopenemr:
https://github.com/openemr/openemr/issues/8287

Fixes # 8287

#### Short description of what this resolves:
Overall improvement of the usability of the recalls workflow

#### Changes proposed in this pull request:
1. The filter button now filters with all of the user inputs: facility, provider, patient ID, and patient name
2. The recall board now has a better format that's more intuitive
3. The patient ID and patient name fields now retain their values after clicking the filter button
4. The "From" and "To" date fields now also update the search results onchange
5. There is now a "No Recalls Found" message when the search yields no recalls
6. The required fields on the add recall form now have red asterisks next to them, and there is now validation alerts when the user submits without filling these fields

#### Does your code include anything generated by an AI Engine? Yes / No
No (except function headers in JS)

PR #10557, opened 2026-02-04
fix(recalls): improve recall board filtering and form validation

kojiromike:
Supersedes #8288

The recall board had no server-side filtering beyond date range, client-side date filtering relied on parsing locale-formatted display text (broke for non-US date formats), form validation was missing for required fields, and filter selections were lost on page reload.

- **Server-side filtering** by facility, provider, patient ID, and patient name in `get_recalls()` SQL query
- **Locale-independent date filtering**: add `data-date` attribute (ISO `YYYY-MM-DD`) to each recall `<tr>`, and filter in JS by comparing ISO strings instead of parsing displayed date text
- **Form validation** with translated error messages for required fields (patient, date, provider, facility) and red asterisk indicators on required fields
- **Filter persistence** via `prevSetting` for patient ID and patient name fields
- **"No Recalls Found" message** displayed when filters match nothing, using a styled div with `#no_recalls_message` ID
- **Remove English-only fallback** for the `translations` object (always emitted by PHP)
- **Remove `parseDate()`/`extractDateFromText()`** functions that contained unnecessary regex escapes and locale-dependent parsing

Claude Code (Anthropic) was used to rework the date filtering approach and integrate the changes from #8288. AI-assisted sections are not marked inline as they are interleaved revisions of existing code rather than standalone generated blocks.

### DIFF
--- a/interface/main/messages/js/reminder_appts.js
+++ b/interface/main/messages/js/reminder_appts.js
@@ -4,7 +4,9 @@
  * @package MedEx
  * @link    http://www.MedExbank.com
  * @author  MedEx <support@MedExBank.com>
+ * @author  Michael A. Smith <michael@opencoreemr.com>
  * @copyright Copyright (c) 2017 MedEx <support@MedExBank.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
  * @license https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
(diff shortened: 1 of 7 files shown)


interface/main/messages/js/reminder_appts.js
```diff
@@ -324,29 +349,84 @@ function goMedEx() {
...
+        default: // 0 = YYYY-MM-DD (already ISO)
+            return val;
+    }
+}
+
 function show_this(colorish='') {
     var facV = $("#form_facility").val();
     var provV = $("#form_provider").val();
     var pidV = $("#form_patient_id").val();
     var pidRE = new RegExp(pidV, 'i');
```
kojiromike: The user input `pidV` goes directly into a `RegExp` constructor. If someone types `.*` or `^` in the Patient ID field, it'll match unintended rows. Would it make sense to escape regex metacharacters first?

Something like:
```suggestion
    var pidRE = new RegExp(pidV.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'i');
```

Same consideration applies to `pnameRE` on the next line.

library/MedEx/API.php
```diff
@@ -2064,22 +2078,46 @@ function SMS_bot(pid) {
...
+        }
+
+        if (!empty($patient_id)) {
+            $query .= " AND r_pid = ?";
+            $params[] = $patient_id;
+        }
+
+        if (!empty($patient_name)) {
+            $query .= " AND (CONCAT(pat.fname, ' ', pat.lname) LIKE ? OR CONCAT(pat.lname, ', ', pat.fname) LIKE ?)";
+            $params[] = "%$patient_name%";
```
kojiromike: Minor: SQL LIKE metacharacters (`%`, `_`, `\`) in `$patient_name` aren't escaped, so typing `%` would match anything. This isn't a security issue (parameterized queries handle injection), but it could produce unexpected results.

If you want strict substring matching, you could escape these:
```php
$escaped = addcslashes($patient_name, '%_\\');
$params[] = "%$escaped%";
```

Not blocking — the current behavior is arguably more flexible for power users.

library/MedEx/API.php
```diff
@@ -2005,9 +2016,12 @@
                 <div class="tab-content">
                    <div class="tab-pane active" id="tab-all">
                         <?php
-                            $this->recall_board_top();
-                            echo $processed['ALL'] ?? '';
+                        $has_recall = !empty($processed['ALL']);
+                        $this->recall_board_top($has_recall);
+                        if ($has_recall) {
+                            echo $processed['ALL'];
```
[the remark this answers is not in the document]
kojiromike: Not a real issue here — $processed["ALL"] is built server-side in recall_board_process() using text() and attr() throughout. But the semgrep rule should recognize these as sanitizers so it stops flagging this pattern. Tracked in #10843.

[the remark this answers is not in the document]
kojiromike: Not a real issue here — the output is built server-side in recall_board_process() using text() and attr() throughout. But the semgrep rule should recognize these as sanitizers so it stops flagging this pattern. Tracked in #10843.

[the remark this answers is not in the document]
kojiromike: Not a real issue here — the output is built server-side in recall_board_process() using text() and attr() throughout. But the semgrep rule should recognize these as sanitizers so it stops flagging this pattern. Tracked in #10843.

interface/main/messages/js/reminder_appts.js
kojiromike: Fixed — now escaping regex metacharacters with `.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")` before passing to `RegExp` constructor, for both `pidRE` and `pnameRE`.

library/MedEx/API.php
kojiromike: Fixed — now escaping LIKE metacharacters with `addcslashes($patient_name, '%_\\')` before wrapping in `%...%`.

PR #10567, opened 2026-02-04
fix(security): set HTTP 403 in csrfNotVerified()

### LINKED ISSUE #10566: fix: CsrfUtils::csrfNotVerified() should set HTTP response code
## Problem

`CsrfUtils::csrfNotVerified()` handles echoing the error message, logging, and optionally dying — but it never sets the HTTP response code. Callers must remember to call `http_response_code(403)` separately before invoking it, and some don't.

A CSRF verification failure is always a 403. There's no scenario where you'd want to fail CSRF but return HTTP 200.

kojiromike:
## Summary

Fixes #10566

`CsrfUtils::csrfNotVerified()` handles echoing the error, logging, and dying — but never sets the HTTP response code. Callers must remember to call `http_response_code(403)` separately, and most don't.

### Changes

- Split into `csrfViolation()` (log only, no exit) and `csrfNotVerified()` (log + exit)
- `csrfViolation()` sets `http_response_code(403)`, echoes the error, and logs
- `csrfNotVerified()` calls `csrfViolation()` then exits, with an optional `beforeExit` callback for pre-exit work (e.g. rendering a Twig template)
- `csrfNotVerified()` returns `never` for static analysis
- Switch `AuthorizationController` to `csrfViolation()` (needs log-only, no exit)
- Switch `account.php` and `validate_messages_document_ajax.php` to use `beforeExit` callbacks
- Remove redundant `http_response_code(403)` from callers that had it

## AI Disclosure

Yes

## Test plan

- [x] Isolated PHPUnit test: `CsrfUtilsTest::testCsrfViolationSetsHttp403`
- [ ] Verify CSRF failure responses now include HTTP 403 status code
- [ ] Verify `AuthorizationController` CSRF handling still returns proper responses

### DIFF
--- a/src/Common/Csrf/CsrfUtils.php
+++ b/src/Common/Csrf/CsrfUtils.php
@@ -15,7 +15,9 @@
  * @package   OpenEMR
  * @link      https://www.open-emr.org
  * @author    Brady Miller <brady.g.miller@gmail.com>
+ * @author    Michael A. Smith <michael@opencoreemr.com>
  * @copyright Copyright (c) 2018-2019 Brady Miller <brady.g.miller@gmail.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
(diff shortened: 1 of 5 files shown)


src/Common/Csrf/CsrfUtils.php
```diff
@@ -78,20 +80,38 @@ public static function verifyCsrfToken($token, $subject = 'default', ?SessionInt
...
             error_log("OpenEMR CSRF token authentication error");
         }
-        if ($die) {
-            die;
+    }
+
+    /**
+     * Record a CSRF violation and terminate the request.
+     *
+     * @param ?callable $beforeExit Optional callback to run before exiting
```
Firehed: Suggestion: proactively tighten up the callable's signature so functions requiring args can't be passed
```suggestion
     * @param ?callable(): mixed $beforeExit Optional callback to run before exiting
```

kojiromike: @Firehed should it be allowed to return anything? the only reason this exists is to support rendering a twig access-denied template. Could be void (or string, and we echo it...)

PR #10568, opened 2026-02-04
feat(security): add AccessDeniedHelper for standardized denial responses

### LINKED ISSUE #10562: refactor: no standard pattern for access denial responses in legacy scripts
## Problem

When fixing security vulnerabilities in legacy scripts (`interface/`, `library/`, `portal/`), we repeatedly write the same boilerplate sequence:

1. Log the denial to SystemLogger
2. Record an audit event via EventAuditLogger
3. Set an HTTP response code (403, 401)
4. Output a message and exit

There is no shared function for this. Every security fix assembles these steps ad-hoc, leading to inconsistency — some files log but don't audit, some audit but don't set the response code, some do all four steps but with different event names or formats.

The modern controller layer handles this well via `AccessDeniedException` and framework exception handlers, but legacy scripts are standalone entry points with no exception handler to catch and format the response. Each one must handle the full deny-and-exit sequence inline.

`CsrfUtils::csrfNotVerified()` is the closest existing precedent — it combines output + log + die for CSRF failures specifically — but it doesn't set the HTTP response code and only covers CSRF, not general access denials.

The denial condition itself varies (ACL checks, patient ID validation, session checks, etc.), so the abstraction should only cover the *response* to a denial, not the check that triggers it.

kojiromike:
## Summary

Fixes #10562

Add `AccessDeniedHelper::deny()` — a single method that handles the full deny-and-exit sequence for legacy scripts: SystemLogger warning, EventAuditLogger audit event, HTTP response code, and exit.

Every security fix in legacy code currently assembles this sequence ad-hoc, leading to inconsistency (some log but don't audit, some audit but don't set the response code, etc.). This provides a standard one-liner:

```php
AccessDeniedHelper::deny('Unauthorized access to patient notes');
```

Modern controllers should continue using `throw new AccessDeniedException(...)`.

### Changes

- New `AccessDeniedHelper::deny()` with SystemLogger + EventAuditLogger + HTTP 403 + exit
- New `AccessDeniedResponseFormat` enum (Text, Json)
- Uses Symfony `Response::HTTP_*` constants for HTTP status, constrained by PHPStan type annotation
- Uses `SessionWrapper` instead of `$_SESSION`
- Optional `beforeExit` callback for pre-exit work (e.g. session cleanup)
- Convert 60+ access denial sites across 40 files to use `AccessDeniedHelper::deny()`:
  - Document retrieval (C_Document.class.php: raw file, patient_picture, document access, IDOR check)
  - Patient demographics and insurance (demographics_full, demographics_save, demographics_print, insurance_edit)
  - Patient history (history_full, history_save, history_sdoh, etc.)
  - Deleter (patient, encounter, form, issue, document, payment, billing, transaction)
  - Fee sheet AJAX endpoints
  - Various other forms and scripts

### Skipped

- Custom modules (oe-module-faxsms, oe-module-comlink-telehealth, oe-module-dashboard-context)
- CSRF checks (handled by `csrfNotVerified` in #10567)
- Twig-rendered unauthorized pages
- Feature gates and non-access-denial error conditions

## AI Disclosure

Yes

## Test plan

- [ ] Verify the class autoloads correctly (PSR-4 in `src/Common/Acl/`)
- [ ] Verify `deny()` sets HTTP 403, logs to SystemLogger and EventAuditLogger, and exits
- [ ] Verify `format: AccessDeniedResponseFormat::Json` returns JSON with `Content-Type: application/json`
- [ ] Verify converted callers still deny access correctly

### DIFF
--- a/controllers/C_Document.class.php
+++ b/controllers/C_Document.class.php
@@ -17,6 +17,7 @@
 require_once(__DIR__ . "/../library/forms.inc.php");
 require_once(__DIR__ . "/../library/patient.inc.php");
 
+use OpenEMR\Common\Acl\AccessDeniedHelper;
 use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\Common\Crypto\CryptoGen;
 use OpenEMR\Common\Csrf\CsrfUtils;
@@ -648,16 +649,8 @@ class C_Document extends Controller
         //  which could introduce xss vulnerabilities.
         if ($disable_exit == true) {
             if (!$this->isReturnRetrieveKey()) {
-                // Access to return the raw file has not been granted. Very likely bad actor, so die.
-                EventAuditLogger::getInstance()->newEvent(
-                    "security-access",
-                    $session->get('authUser') ?? '',
-                    $session->get('authProvider') ?? '',
-                    0,
-                    "Unauthorized attempt to return raw document file"
-                );
-                http_response_code(403);
-                die(xlt("Not authorized to return raw file."));
+                // Access to return the raw file has not been granted. Very likely bad actor, so deny.
+                AccessDeniedHelper::deny('Attempt to return raw file without retrieve key');
             }
         }
 
@@ -672,19 +665,7 @@ class C_Document extends Controller
             if (!($session->has('patient_portal_onsite_two') && $session->has('pid'))) {
                 $allowed_pid = $GLOBALS['pid'] ?? 0;
                 if ($allowed_pid === 0 || $patient_id === null || $patient_id !== (string)$allowed_pid) {
-                    (new SystemLogger())->warning(
-                        "An attempt was made to retrieve a patient picture for an unauthorized patient",
-                        ['user-id' => $session->get('authUserID'), 'requested-patient-id' => $patient_id, 'session-pid' => $allowed_pid]
-                    );
-                    EventAuditLogger::getInstance()->newEvent(
-                        "security-access",
-                        $session->get('authUser') ?? '',
-                        $session->get('authProvider') ?? '',
-                        0,
-                        "Unauthorized attempt to retrieve patient picture for pid " . $patient_id
-                    );
-                    http_response_code(403);
-                    die(xlt("Not authorized to view requested file"));
+                    AccessDeniedHelper::deny("Attempt to retrieve patient picture for pid $patient_id");
                 }
             }
         }
@@ -695,48 +676,18 @@ class C_Document extends Controller
         if ($session->has('patient_portal_onsite_two') && $session->has('pid')) {
             // ensure patient has access (called from patient portal)
             if (!$d->can_patient_access($session->get('pid'))) {
-                (new SystemLogger())->warning("An attempt was made by a patient to download a document from an unauthorized category", ['patient-id' => $session->get('pid'), 'document-id' => $document_id]);
-                EventAuditLogger::getInstance()->newEvent(
-                    "security-access",
-                    $session->get('pid') ?? '',
-                    '',
-                    0,
-                    "Patient unauthorized to access document " . $document_id . " category"
-                );
-                http_response_code(403);
-                die(xlt("Not authorized to view requested file"));
+                AccessDeniedHelper::deny("Patient unauthorized to access document $document_id category");
             }
         } else {
             // ensure user has access
             if (!$d->can_access()) {
-                (new SystemLogger())->warning("An attempt was made by a user to download a document from an unauthorized category", ['user-id' => $session->get('authUserID'), 'patient-id' => $patient_id, 'document-id' => $document_id]);
-                EventAuditLogger::getInstance()->newEvent(
-                    "security-access",
-                    $session->get('authUser') ?? '',
-                    $session->get('authProvider') ?? '',
-                    0,
-                    "Unauthorized attempt to access document " . $document_id . " from restricted category"
-                );
-                http_response_code(403);
-                die(xlt("Not authorized to view requested file"));
+                AccessDeniedHelper::deny("Unauthorized attempt to access document $document_id from restricted category");
             }
 
             // Verify the document belongs to the requested patient to prevent IDOR.
             $doc_pid = $d->get_foreign_id();
             if ($patient_id !== null && (int)$doc_pid !== (int)$patient_id) {
-                (new SystemLogger())->warning(
-                    "An attempt was made to retrieve a document belonging to a different patient",
-                    ['user-id' => $session->get('authUserID'), 'requested-patient-id' => $patient_id, 'document-patient-id' => $doc_pid, 'document-id' => $document_id]
-                );
-                EventAuditLogger::getInstance()->newEvent(
-                    "security-access",
-                    $session->get('authUser') ?? '',
-                    $session->get('authProvider') ?? '',
-                    0,
-                    "Unauthorized attempt to retrieve document " . $document_id . " belonging to pid " . $doc_pid
-                );
-                http_response_code(403);
-                die(xlt("Not authorized to view requested file"));
+                AccessDeniedHelper::deny("Unauthorized attempt to retrieve document $document_id belonging to pid $doc_pid");
             }
         }
 
--- a/src/Common/Acl/AccessDeniedHelper.php
+++ b/src/Common/Acl/AccessDeniedHelper.php
@@ -1,0 +1,80 @@
+<?php
+
+/**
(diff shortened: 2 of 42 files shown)


src/Common/Acl/AccessDeniedHelper.php
```diff
@@ -0,0 +1,71 @@
...
+     * @param string $comment     Audit log comment describing what was attempted
+     * @param string $auditEvent  Event name for EventAuditLogger
+     * @param int    $httpCode    HTTP response code
+     * @param string $format      Response body format: 'text' or 'json'
+     * @return never
+     */
+    public static function deny(
+        string $comment,
+        string $auditEvent = 'security-access-denied',
+        int $httpCode = 403,
```
kojiromike: can we do better than an int here? Is there an enum of http status codes we can use?

src/Common/Acl/AccessDeniedHelper.php
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param string $comment     Audit log comment describing what was attempted
+     * @param string $auditEvent  Event name for EventAuditLogger
+     * @param int    $httpCode    HTTP response code
+     * @param string $format      Response body format: 'text' or 'json'
+     * @return never
+     */
+    public static function deny(
+        string $comment,
+        string $auditEvent = 'security-access-denied',
```
kojiromike, line 41: can we do better than a string here? Is there an enum of auditEvents we can validate against?

src/Common/Acl/AccessDeniedHelper.php
```diff
@@ -0,0 +1,71 @@
...
+        int $httpCode = 403,
+        string $format = 'text',
```
kojiromike: This could be an enum too -- we only support text and json right now after all.

src/Common/Acl/AccessDeniedHelper.php
kojiromike: Done — added `AccessDeniedHttpStatus` int-backed enum with `Unauthorized` (401) and `Forbidden` (403).

src/Common/Acl/AccessDeniedHelper.php
kojiromike: Leaving this as a string for now. There are ~30 different event names passed to `newEvent()` across the codebase (`login`, `backup`, `delete`, `authorize`, etc.). Typing them all into an enum is valuable but it's a broader effort — this helper would introduce a narrow enum that only covers its own use cases. The default (`security-access-denied`) covers the common case here.

src/Common/Acl/AccessDeniedHelper.php
kojiromike: Done — added `AccessDeniedResponseFormat` enum with `Text` and `Json` cases.

src/Common/Acl/AccessDeniedHttpStatus.php
```diff
@@ -0,0 +1,20 @@
...
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Common\Acl;
+
+enum AccessDeniedHttpStatus: int
```
Firehed: OEMR\Common\Http\StatusCode already exists (I bet there's several in vendor as well). While I prefer the enum I'd rather avoid having duplicative implementations until we can convert the existing stuff.

If need be, you can PHPStan-type the int where it's used and have about the same effect

```php
/**
 * @param StatusCode::* $code
 */
function useCode(int $code) {}
```

adunsulag: Since I introduced a heavy symfony refactor in the API's I've been using Symfony\Component\HttpFoundation\Response which has all status code's in the HTTP specification there as consts.  The enum is nice but I'd almost prefer to get rid of the StatusCode and use the symfony one.

src/Common/Acl/AccessDeniedHelper.php
```diff
@@ -0,0 +1,77 @@
...
+     *                                                (e.g. render a template, clean up session)
+     */
+    public static function deny(
+        string $comment,
+        string $auditEvent = 'security-access-denied',
+        AccessDeniedHttpStatus $httpStatus = AccessDeniedHttpStatus::Forbidden,
+        AccessDeniedResponseFormat $format = AccessDeniedResponseFormat::Text,
+        ?callable $beforeExit = null,
+    ): never {
+        $user = $_SESSION['authUser'] ?? 'unknown';
```
adunsulag: Should probably use the SessionWrapper piece here instead of $_SESSION

kojiromike:
@adunsulag I believe I've addressed all your requested changes.

controllers/C_Document.class.php
```diff
@@ -695,48 +676,18 @@ public function retrieve_action(?string $patient_id, $document_id, $as_file = tr
...
-                (new SystemLogger())->warning(
-                    "An attempt was made to retrieve a document belonging to a different patient",
-                    ['user-id' => $session->get('authUserID'), 'requested-patient-id' => $patient_id, 'document-patient-id' => $doc_pid, 'document-id' => $document_id]
-                );
-                EventAuditLogger::getInstance()->newEvent(
-                    "security-access",
-                    $session->get('authUser') ?? '',
-                    $session->get('authProvider') ?? '',
-                    0,
-                    "Unauthorized attempt to retrieve document " . $document_id . " belonging to pid " . $doc_pid
```
adunsulag, line 736: I think you need to add the document_id and doc_pid back here into the deny message so the event audit logger has the information of what data was being attempted to be accessed.

controllers/C_Document.class.php
```diff
@@ -695,48 +676,18 @@ public function retrieve_action(?string $patient_id, $document_id, $as_file = tr
...
-                (new SystemLogger())->warning("An attempt was made by a user to download a document from an unauthorized category", ['user-id' => $session->get('authUserID'), 'patient-id' => $patient_id, 'document-id' => $document_id]);
-                EventAuditLogger::getInstance()->newEvent(
-                    "security-access",
-                    $session->get('authUser') ?? '',
-                    $session->get('authProvider') ?? '',
-                    0,
-                    "Unauthorized attempt to access document " . $document_id . " from restricted category"
-                );
-                http_response_code(403);
-                die(xlt("Not authorized to view requested file"));
```
adunsulag, line 721: I think you need to add the document_id back here into the deny message so the event audit logger has the information of what data was being attempted to be accessed.

controllers/C_Document.class.php
```diff
@@ -695,48 +676,18 @@ public function retrieve_action(?string $patient_id, $document_id, $as_file = tr
         if ($session->has('patient_portal_onsite_two') && $session->has('pid')) {
             // ensure patient has access (called from patient portal)
             if (!$d->can_patient_access($session->get('pid'))) {
-                (new SystemLogger())->warning("An attempt was made by a patient to download a document from an unauthorized category", ['patient-id' => $session->get('pid'), 'document-id' => $document_id]);
-                EventAuditLogger::getInstance()->newEvent(
```
adunsulag, line 699: Same thing as below, I'd have to review our audit trail requirements (which I'm pretty sure are followed inconsistently, but we should keep track of the patient pid that attempted this.

controllers/C_Document.class.php
```diff
@@ -672,19 +665,7 @@ public function retrieve_action(?string $patient_id, $document_id, $as_file = tr
             if (!($session->has('patient_portal_onsite_two') && $session->has('pid'))) {
                 $allowed_pid = $GLOBALS['pid'] ?? 0;
                 if ($allowed_pid === 0 || $patient_id === null || $patient_id !== (string)$allowed_pid) {
-                    (new SystemLogger())->warning(
-                        "An attempt was made to retrieve a patient picture for an unauthorized patient",
```
adunsulag, line 676: See comments above on putting info back into event audit log.

controllers/C_Document.class.php
kojiromike: @adunsulag maybe I don't understand your phrasing. Would a _patient_ have attempted this? Or could it have been another user with that patient in their session?

bradymiller: @kojiromike , This is catching a patient (in portal) trying to see another patient data.

PR #10573, opened 2026-02-04
refactor(sql): remove QuotedOrNull and convert to prepared statements

Firehed:
Towards #10241 and #10149 (ish)

Removes the duplicate `QuotedOrNull` function definitions (5 total) and converts the affected SQL queries to use proper prepared statements, fixing a possible SQL injection vulnerability in the process.

- Remove dead code: `QuotedOrNull` in `add_edit_issue.php` and `types_edit.php` (defined but never called)
- Convert `orders_results.php` from SQL string concatenation to prepared statements (fixes SQL injection - the QuotedOrNull there did not escape input)
- Remove `QuotedOrNull` from `procedure_order/common.php` by inlining the empty-to-null conversion
- Convert `eye_mag/save.php` lists queries to prepared statements
- Use `QueryUtils` methods for better error handling via exceptions

Yes

All commits include `Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>` trailer. The refactored SQL queries using HEREDOC syntax and QueryUtils were generated with Claude Code assistance.


interface/forms/eye_mag/save.php
```diff
@@ -686,49 +687,100 @@
...
+                            `erx_uploaded` = '0',
+                            `modifydate` = NOW(),
+                            `subtype` = ?
+                        WHERE `id` = ?
+                        SQL,
+                        [
+                            $form_type,
+                            $_REQUEST['form_title'],
+                            $_REQUEST['form_comments'],
+                            empty($form_begin) ? null : $form_begin,
```
kojiromike: Any way we can eliminate the use of `empty` here?

Firehed: I was able to address two of the three since the variables definitely exist. I have no idea where `form_return` comes from, if anywhere, so I don't want to touch that one (I suspect it's never set and has been emitting silent warnings for quite a while).

interface/forms/eye_mag/save.php
```diff
@@ -686,49 +687,100 @@
...
+                            `subtype` = ?
+                        WHERE `id` = ?
+                        SQL,
+                        [
+                            $form_type,
+                            $_REQUEST['form_title'],
+                            $_REQUEST['form_comments'],
+                            $form_begin ?: null,
+                            $form_end ?: null,
+                            empty($form_return) ? null : $form_return,
```
kojiromike, line 721: This `$form_return` variable appears to never be set anywhere in this file. The `$form_begin` and `$form_end` are set from `$_REQUEST` via `DateToYYYYMMDD()` on lines 648-649, but there's no corresponding assignment for `$form_return`.

This means `returndate` has effectively always been NULL in the database. If that's intentional, this refactor correctly preserves that behavior. If not, it might be worth adding:
```php
$form_return = DateToYYYYMMDD($_REQUEST['form_return'] ?? '');
```

Call it outside scope if you want, but if you don't address it here, please create a ticket.

Issue #10084, opened 2026-01-04
refactor(test): better tests placement


igormukhingmailcom:
I propose moving different kinds of tests to separate special directories. Now they a bit mixed.

New structure (avoiding `tests/Tests`):

```
tests
- integration
- isolated
- ...
- smoke (future)
- load (future)
- security / penetration (future)
```

Also:
- Move fixtures from `tests/` to `src/`
- Move `tests/bootstrap*` files to related sub-directories (`tests/isolated/bootstrap.php`, `tests/integration/bootstrap.php`, ``tests/.../bootstrap.php`)
- Update `tests/README.md` and define naming/placement rules:
 - Which one goes to which folder / what should go to `unit` and what to `isolated` directory
 - What postfix should have, like isolated tests should benamed like `*IsolatedTest.php`)
 - Move docs from `tests/certification` to better place
- Move `eventdispatcher/*-example` dirs to some `./examples` or `./docs` root directory or as they are not tests but examples
- Add composer scripts to run all tests or particular suite or group quicker (its a bit pain to copy-paste correct one-liner to run isolated tests, for example)
- Suite `unit-isolated` gets files from not-obvious path, so should be fixed somehow (moved?) as well I believe
- Probably, add naming/placement rules check with https://github.com/phparkitect/arkitect
- Probably, remove api/services/validators/etc suites as they are same kind; don't rely on suites so we have simpler `phpunit.xml`, should not update it when new sub-directory added; we can use groups during development as they are more flexible
- Can't find where `DISABLE_DATABASE` is used, probably needs removing from `phpunit-isolated.xml`

Firehed:
I like the thought process here. A couple other things to consider:

- A lot of the tests in `unit` are not textbook-definition unit tests, and require additional boostrapping (global state, etc) to function correctly. How far you want to take the terminology is up for debate[^mock], but we can probably all agree that unit tests should not require any bootstrapping to pass
- By extension, running unit tests should be fast and portable. My rough guideline from other projects is executing roughly 500 tests per second. They should not require docker, network, or any appreciable setup, so long as composer dependencies (including platform/extension) are met
- If we're smart about namespacing, authoring tests becomes easier. Notably, the TestCase class and the class under test are in _the same namespace_[^autoloading] so there's no need to import the test subject.
- The PHPUnit test suite configs shouldn't need to be regularly updated as new tests are authored. Ideally there's a single config file fill stop, but that's probably not feasible (at least not for a while). You already touched on this :)

[...]

PR #10575, opened 2026-02-04
refactor(tests): improve isolated test suite and remove error_log output

Firehed:
Makes some progress on #10084.

Improves the isolated test suite by removing the bootstrap file entirely and refactoring classes to use PSR-3 loggers instead of `error_log()`, eliminating noisy output during test runs.

There are still a couple things (at least on PHP 8.4) that emit noise due to some issues in an upstream validation library, but only at the end so it's quite a bit quieter.

- Remove `tests/bootstrap-isolated.php` — isolated tests now use only the Composer autoloader
- Refactor `TelemetryService` to use its injected PSR-3 logger instead of `error_log()`
- Refactor `Installer` to accept a `LoggerInterface` constructor parameter and use it instead of `error_log()`
- Update all `Installer` call sites to provide appropriate loggers (SystemLogger for production, NullLogger for tests)
- Update `README-Isolated-Testing.md` with clearer guidance on writing isolated tests, including a PHPUnit mock example

All commits include `Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>` to indicate AI assistance. The changes are refactoring and documentation improvements guided by AI suggestions.

🤖 Generated (mostly) with [Claude Code](https://claude.ai/code)

### DIFF
--- a/library/classes/Installer.class.php
+++ b/library/classes/Installer.class.php
@@ -18,7 +18,6 @@
 
 use OpenEMR\Gacl\GaclApi;
 use Psr\Log\LoggerInterface;
-use Psr\Log\NullLogger;
 
 class Installer
 {


library/classes/Installer.class.php
```diff
@@ -17,6 +17,8 @@
  */
 
 use OpenEMR\Gacl\GaclApi;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
```
kojiromike: It looks like `NullLogger` isn't used anywhere in this file — only `LoggerInterface` is referenced (in the constructor). Was this left over from an earlier draft that defaulted the parameter to `new NullLogger()`?

```suggestion
```

library/classes/Installer.class.php
```diff
@@ -60,8 +62,9 @@ class Installer
      * Initialize the Installer with configuration variables.
      *
      * @param array $cgi_variables Configuration array containing installation parameters
+     * @param LoggerInterface $logger Logger instance for error reporting
      */
-    public function __construct(array $cgi_variables)
+    public function __construct(array $cgi_variables, private readonly LoggerInterface $logger)
```
kojiromike, line 66: Was a default of `new NullLogger()` considered here? Since `Installer` is a public class (even if narrowly used), a default would keep existing call sites working without changes and make the logger opt-in rather than required. On the other hand, making it required forces callers to be explicit — which has its own advantages.

All current call sites are updated so it's not a functional issue, just wondering about the intent.

Firehed: I did consider that (and had to steer Claude away). I felt requiring it was more appropriate since a default of logging to /dev/null (vs the previous behavior of error_log) would be likely to suppress issues that are critical to address during installation, and that would likely lead to goofy support issues.

PR #8726, opened 2025-08-01
test(isolated): tests that don't require initialization

kojiromike:
Fixes #8725

#### Short description of what this resolves:

Introduces a new workflow and tests that can run primarily with just phpunit, without initialization. For some kinds of work this dramatically reduces cycle time, and enables AI to cycle on problems more rapidly as well.

#### Changes proposed in this pull request:

- style(whitespace): fix whitespace in tests
- tests(Validators): tests without database
- tests(USPS): tests without database
- tests(Tools\OAuth2) tests without database
- ci(isolated-tests): workflow for fast tests with no install
- doc(tests): differentiate between isolated and initialized tests

#### Does your code include anything generated by an AI Engine? Yes

Several of the tests themselves are Claude generated.

PR #10576, opened 2026-02-04
refactor(ci): extract common PHP/Composer setup into reusable action

### LINKED ISSUE #10300: refactor(ci): extract common PHP setup steps for reuse
I'm looking at adding another CI job and noticed there's a lot of common setup code repeated across workflows and jobs. It would be convenient if most of these common things were extracted into a common reusable workflow.

Specifically, many jobs do all of the following things in a copy-pasted way:

- run setup-php
- composer cache stuff (which is multiple steps)
- composer install

There's probably other opportunities for similar setup re-use, dunno.

Firehed:
Fixes #10300

Extracts the repeated PHP setup, Composer cache configuration, and Composer install steps into a reusable composite action to reduce duplication across CI workflows.

- Create new composite action at `.github/actions/setup-php-composer/action.yml` with inputs:
  - `php-version` (required): PHP version to install
  - `coverage` (optional, default: `none`): Coverage driver
  - `install` (optional, default: `true`): Whether to run composer install
- Update 7 workflows to use the new action:
  - api-docs.yml
  - composer.yml
  - conventional-commits.yml
  - isolated-tests.yml
  - phpstan.yml (with separate rector removal step)
  - rector.yml
  - styling.yml (PHP job only)

Net reduction of ~130 lines of duplicated YAML.

The composite action and workflow modifications were generated with assistance from Claude (Anthropic). The changes are straightforward YAML refactoring that consolidates existing patterns.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -32,6 +32,7 @@ jobs:
       with:
         php-version: ${{ matrix.php-version }}
     # Hack: temporarily uninstall rector, which interferes with PHPStan
+    # (it bundles some type-incomplete polyfills which can throw off results)
     - name: Remove Rector
       run: composer remove --dev --optimize-autoloader rector/rector
     - name: PHPStan Cache


.github/actions/setup-php-composer/action.yml
```diff
@@ -0,0 +1,38 @@
...
+        php-version: ${{ inputs.php-version }}
+        coverage: ${{ inputs.coverage }}
+
+    - name: Get composer cache directory
+      id: composer-cache
+      shell: bash
+      run: echo "dir=$(composer config cache-files-dir)" >> "$GITHUB_OUTPUT"
+
+    - name: Cache Composer dependencies
+      uses: actions/cache@v5
```
kojiromike, line 44: The composite action uses `actions/cache@v5` (restore + save), but `isolated-tests.yml` was deliberately using `actions/cache/restore@v5` (restore-only, no save). It's the only workflow in the converted set that did this.

The restore-only pattern was introduced in [#8726](https://github.com/openemr/openemr/pull/8726) (`0499a36`) — the isolated-tests workflow was designed as a "fast tests with no install" workflow. The intent was that these jobs should be pure consumers of the cache populated by other workflows, not writers — avoiding cache churn from the 5-PHP-version matrix.

Switching to `actions/cache@v5` means isolated-test runs will now save back to the cache on misses. That's probably fine in practice (the cache key includes the PHP version so they won't stomp each other), but it is a behavioral change from the original design. Worth a quick sanity check — maybe adding a `save-cache` boolean input (default `true`) to the composite action so callers can opt out?

.github/actions/setup-php-composer/action.yml
```diff
@@ -0,0 +1,38 @@
+name: Setup PHP and Composer
+description: Sets up PHP, configures Composer caching, and installs dependencies
+
+inputs:
```
kojiromike, line 4: Minor: would it be worth adding `extensions` and `ini-values` inputs (defaulting to empty) as pass-throughs to `shivammathur/setup-php`? That would let `composer-require-checker.yml` (which specifies a long `extensions` list) adopt this action too. Not blocking — just thinking about future reuse.

.github/actions/setup-php-composer/action.yml
Firehed: Useful context for sure - thanks!

Looking closer, I think the previous behavior was incorrect; a complete cache miss would either make the job fail, and an exact cache value miss could lead to running the checks against the wrong dependency versions which is arguably worse.

.github/actions/setup-php-composer/action.yml
Firehed: Yes, next couple commits have taken this further and enabled more consolidation. CRC is down to almost nothing after this change.

I'll look at doing more in the two heaviest workflows (test and inferno) in a separate PR

Firehed:
Note: I cross-checked the runtime of the changed workflows against master, and everything other than PHPStan is either faster or within the margin of error. 

PHPStan's runtime has a very high degree of variability depending on the changeset; the delta here was an uncached run vs a cached one (of the PHPStan result cache, which was NOT changed in this PR). The setup code remains a cache hit and was comparable to master (and, overall, cache-miss runs on master) so I don't think it's a regression.

PR #10591, opened 2026-02-05
chore(cleanup): Remove completely empty files

Firehed:
Empty files should not remain in the codebase. There might be more but these came up while looking at unrelated stuff.

Removes two completely empty files.

No


kojiromike:
Would you mind reconciling the other files in https://github.com/openemr/openemr/pull/9771/changes and see what else got re-added by mistake?

Firehed:
@kojiromike It looks like the two files in this change are the only two that were re-introduced. I diffed the tree and didn't spot anything else.

PR #10592, opened 2026-02-06
feat(about): add event hooks for module content injection

### LINKED ISSUE #10587: Add event hooks to About page for module content injection
## Summary

The About page (`interface/main/about_page.php`) renders directly via Twig without dispatching any events, making it impossible for modules to inject content (like certification badges or additional information).

## Use Case

Modules that provide certification status (e.g., ONC certification) would benefit from displaying badges/logos on the About page, which is a natural place for users to see certification and compliance information.

## Proposed Solution

Add a `TemplatePageEvent` dispatch (or similar) to `about_page.php` that allows modules to:
1. Add additional Twig variables
2. Inject HTML content at designated points (e.g., after version info, before/after the buttons section)

Example pattern from other pages:
```php
$event = new TemplatePageEvent('about_page');
$event->setTwigVariables($viewArgs);
$GLOBALS['kernel']->getEventDispatcher()->dispatch($event, TemplatePageEvent::RENDER_EVENT);
$viewArgs = $event->getTwigVariables();
```

And corresponding Twig blocks:
```twig
{% if additionalContent is defined %}
    {{ additionalContent|raw }}
{% endif %}
```

## Alternatives Considered

- JavaScript injection via main tabs events (hacky, timing issues)
- Overriding the template entirely (breaks with updates)

kojiromike:
## Summary

Fixes #10587

Dispatch `TemplatePageEvent` from the About page to allow modules to inject content via Twig variables.

- Dispatch `TemplatePageEvent::RENDER_EVENT` with `pageName='about_page'` before rendering
- Add two content injection points in the Twig template:
  - `additionalInfoContent`: after the version/support info section
  - `additionalButtonsContent`: after the buttons section

### Usage example

Modules can listen for the event and inject content:

```php
$eventDispatcher->addListener(TemplatePageEvent::RENDER_EVENT, function (TemplatePageEvent $event) {
    if ($event->getPageName() !== 'about_page') {
        return;
    }
    $event->setTwigVariables([
        'additionalInfoContent' => '<div class="certification-badge">...</div>',
    ]);
});
```

## AI disclosure

- [x] Yes, I have used an AI to help craft this PR

### DIFF
--- a/templates/core/about.html.twig
+++ b/templates/core/about.html.twig
@@ -41,6 +41,12 @@
                 </div>
                 {% endif %}
 
+                {% if additionalInfoContent is defined and additionalInfoContent is not empty %}
+                <div class="additional-info-content mt-3">
+                    {{ additionalInfoContent|raw }}
+                </div>
+                {% endif %}
+
                 {% if userManualHref %}
                 <div class="user-manual mt-3">
                     <a href="{{ userManualHref|javascriptStringRemove|attr }}" target="_blank" rel="opener" class="btn text-left btn-block btn-outline-secondary btn-lg">
@@ -70,6 +76,12 @@
                     <a href="https://www.open-emr.org/donate" target="_blank" rel="opener" class="btn text-left btn-block btn-outline-danger btn-lg">
                         <i class="fa fa-fw fa-heart fa-lg"></i>&nbsp;{{ "Donate"|xlt }}
                     </a>
+                </div>
+                {% endif %}
+
+                {% if additionalButtonsContent is defined and additionalButtonsContent is not empty %}
+                <div class="additional-buttons-content mt-3">
+                    {{ additionalButtonsContent|raw }}
                 </div>
                 {% endif %}
             </div>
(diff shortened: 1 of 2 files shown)


templates/core/about.html.twig
```diff
@@ -41,6 +41,12 @@
                 </div>
                 {% endif %}
 
+                {% if additionalInfoContent is defined and additionalInfoContent is not empty %}
+                <div class="additional-info-content mt-3">
+                    {{ additionalInfoContent|raw }}
```
bradymiller, line 46: note `raw` not useful here since auto escaping is off in twig
Recommend removing it and commenting this code that it is escaped previously (I am assuming that is what is happening)

Otherwise, a dev may see this in future and remove it by mistake :)

templates/core/about.html.twig
```diff
@@ -72,6 +78,12 @@
                     </a>
                 </div>
                 {% endif %}
+
+                {% if additionalButtonsContent is defined and additionalButtonsContent is not empty %}
+                <div class="additional-buttons-content mt-3">
+                    {{ additionalButtonsContent|raw }}
```
bradymiller, line 84: note `raw` not useful here since auto escaping is off in twig
Recommend removing it and commenting this code that it is escaped previously (I am assuming that is what is happening)

Otherwise, a dev may see this in future and remove it by mistake :)

PR #10593, opened 2026-02-06
refactor(ci): consolidate PHPCS config into phpcs.xml.dist

Firehed:

CI and local PHPCS invocations had divergent flags (extensions, report-width, warnings, progress). Configuration lived at `ci/phpcs.xml`, a non-standard location requiring `--standard=` everywhere. Two other configs (`ci/phpcs_src.xml`, `ci/phpcs_strict.xml`) were unreferenced. `tmp-phpstan` was being erroneously analyzed.

- Move PHPCS configuration from `ci/phpcs.xml` to `phpcs.xml.dist` at the repo root for automatic discovery
- Embed all runtime flags (extensions, memory_limit, report-width, warning-severity, parallel, progress) in the config so every invocation is identical
- Simplify `composer phpcs`/`phpcbf` scripts by removing `--standard=` flag
- Replace raw `phpcs` command in CI workflow with `composer phpcs`
- Add a problem matcher so PHPCS errors produce inline PR annotations
- Add `phpcs.xml` to `.gitignore` for local overrides
- Delete unused `ci/phpcs_src.xml` and `ci/phpcs_strict.xml`

Mostl commits in this PR were co-authored with Claude Code (Claude Opus 4.6).

---

Locally, this change allows `composer phpcs` to actually work since it has a default directory, and not crash since it now correctly excludes the phpstan temp directory.

Runtime improved from ~72s to ~15s, a just under 6x improvement ⚡ Improvement in CI isn't as pronounced, it's about twice as fast on the actual analysis step. This makes sense since IIRC the default github runners have 2 cores, and my local system is an M2 Pro with 6 performance and 4 efficiency cores.


kojiromike:
Better check if there's any tooling in openemr-devops impacted. (Claude can probably tell you.)

Firehed:
tl;dr yes there are some references in some utilities. Mostly in `obsolete/`; one set in the current (?) `flex/`. Each set is about the same. In relation to the changed config files:

- `phpcs_strict.xml` is unused
- `phpcs_src.xml` is used by a "PSR-12 report" shell script
- `phpcs.xml` is used by everything else.

The setup in devops  is installing and using a separate copy of PHPCS, and looks like it won't be completely reliable in the first place (it hardcodes v3; v4 is in the repo and ci).

My vote is to clean up CONTRIBUTING.md, strip the old paths out of the devops repo entirely, and echo a "use composer phpcs" type message. The commands as documents don't seem to work. Even if they did, the one to generate the PSR-12 report was pointed at the wrong file.

stephenwaite:
hi @Firehed, seems like you aren't using the easy dev environment and the openemr-cmd [devtools stuff](https://github.com/openemr/openemr-devops/blob/master/docker/openemr/flex/utilities/devtools#L144). I think we talked about this a little bit a couple weeks ago on the admin call. Looks like the proposed changes you have here would break the `openemr-cmd pr` and `openemr-cmd pf` commands.

Firehed:
@stephenwaite I'm using the `development-easy` docker setup, but run most tools locally instead of through docker wrappers or other sidecar tooling. I think tweaking the dev tools script accordingly will be straightforward change, though obviously not on this branch since it's in the other repo.

kojiromike:
@Firehed this one's in conflict now. Plus I think I'd be more comfortable if there were already a PR in the other repo linked to this, ready to go.

Firehed:
https://github.com/openemr/openemr-devops/pull/551 should take care of it.

PR #10595, opened 2026-02-06
fix(e2e): add fresh session retry when JS fails to initialize after login

### LINKED ISSUE #10594: fix(e2e): add fresh session retry when JavaScript fails to initialize after login
## Problem

E2E tests have a **58% retry rate** on master. Analysis of the last 50 `Test All Configurations` runs shows:

- 29 of 50 runs needed at least one retry
- Some runs required up to 6 attempts to pass
- Failures are randomly distributed across all PHP/webserver/database configurations

### Observed Failure Pattern

Every sampled failure (9/9) shows identical diagnostic state:

```json
{
  "koAvailable": false,
  "mainMenuChildren": 0,
  "readyState": "complete",
  "bodyLength": 6552
}
```

**What we know for certain:**
- Page HTML loads completely (`readyState: complete`)
- Knockout.js is not defined (`koAvailable: false`) in all observed failures
- Menu has no children (`mainMenuChildren: 0`) in all observed failures
- Body size is consistent across failures (6552 bytes)

**What we can infer:**
- When `mainMenuChildren > 0` (test passes), `ko.applyBindings()` must have run, implying knockout loaded
- We never observe `koAvailable: true` with `mainMenuChildren: 0` (would indicate ko loaded but binding failed)

**What we cannot directly verify:**
- The `koAvailable` value when tests pass (diagnostics only log on failure)

**Likely cause:** The random distribution across all configurations suggests CI environment resource pressure (script loading timeouts, network issues between containers) rather than a deterministic code bug.

### Why existing waits don't help

When scripts fail to load, `ko.applyBindings()` never executes. No amount of waiting will make the menu appear - the only recovery is to start a fresh session.

A page refresh won't work because `prevent_browser_refresh` defaults to 2, which causes `main.php` to consume the session token after first validation. A refresh would fail token validation and destroy the session.

## Proposed Solution

Add a **fresh session retry** mechanism to `login()`:

1. **Fast detection (5s)** - Use a short initial timeout to detect JS initialization failure quickly
2. **Fresh session** - If JS doesn't initialize, quit the browser entirely and create a new session with a fresh login
3. **Single retry only** - Prevents masking persistent bugs while handling transient CI issues
4. **Preserve diagnostics** - If retry also fails, throw exception with full page state

### Why this is different from removed retry patterns

PR #10500 removed 3x retry loops from test methods because they masked issues. This approach is different:

| Removed pattern | New approach |
|----------------|--------------|
| Retry in test methods | Retry at login() - earliest possible point |
| Catch any exception | Only when JS fails to initialize |
| 3x retry loop | Single retry |
| Page refresh (doesn't work) | Fresh browser session with new token |

## Analysis Data

Failed configurations from recent runs (showing random distribution):
```
   8 PHP 8.5 - nginx - mariadb 11.8 / run
   8 PHP 8.2 - nginx - mariadb 11.8 / run
   7 PHP 8.4 - nginx - mariadb 11.8 / run
   7 PHP 8.3 - nginx - mariadb 11.8 / run
   5 PHP 8.4 - apache - mariadb 10.11 / run
   4 PHP 8.6 - nginx - mariadb 11.8 / run
   ...
```

Common failure patterns:
```
✘ Main menu link with data set "Admin -> System -> API Clients menu link"
✘ Main menu link with data set "Modules -> Manage Modules menu link"  
✘ Main menu link with data set "Admin -> Practice -> Rules menu link"
✘ Login authorized
✘ User add
```

## AI Disclosure

Yes - Claude Code analyzed CI logs and proposed the solution.

kojiromike:
Fixes #10594

## Short description of what this resolves

E2E tests have a 58% retry rate on master. Every sampled failure (9/9) shows identical symptoms: `koAvailable: false`, `mainMenuChildren: 0`. This indicates Knockout.js never loads in these cases, likely due to CI environment resource pressure.

Add a fresh session retry mechanism to `login()`:

1. **Fast detection (5s)** - Short initial timeout detects failure quickly instead of waiting 30s
2. **Fresh session** - If JS doesn't initialize, quit browser entirely and create new session with fresh login
3. **Single retry only** - Prevents masking persistent bugs while handling transient CI issues
4. **Diagnostic logging** - Log `koAvailable` state on success to verify hypothesis; log when retry triggers

### Why this works

A page refresh won't work because `prevent_browser_refresh` consumes the session token after first validation. But quitting the browser entirely and creating a fresh session with `base()` + `performLogin()` gets a new token, sidestepping the issue.

### Why this is different from removed retry patterns

PR #10500 removed 3x retry loops from test methods because they masked issues. This approach is different:

| Removed pattern | This approach |
|----------------|---------------|
| Retry in test methods | Retry at login() - earliest possible point |
| Catch any exception | Only when JS fails to initialize |
| 3x retry loop | Single retry |
| Page refresh (doesn't work) | Fresh browser session with new token |

### Evidence and limitations

**What we know:**
- 100% of sampled failures show `koAvailable: false` (knockout never loaded)
- We never see `koAvailable: true` with `mainMenuChildren: 0` (would indicate ko loaded but binding failed)
- Failures are randomly distributed across all PHP/DB configurations

**What we will verify via diagnostic logging:**
- Every successful `waitForAppReady()` logs: `[E2E] waitForAppReady succeeded: {"koAvailable":...,"mainMenuChildren":...}`
- Every retry logs: `[E2E] JS failed to initialize after login, retrying with fresh session...`

## AI Disclosure

Yes

### DIFF
--- a/tests/Tests/E2e/Base/BaseTrait.php
+++ b/tests/Tests/E2e/Base/BaseTrait.php
@@ -85,40 +85,56 @@ trait BaseTrait
      * fail because the page HTML loaded but the JS framework hasn't
      * finished rendering.
      *
-     * Note: a refresh-and-retry strategy won't work here because
-     * prevent_browser_refresh defaults to 2, which causes main.php
-     * to consume the session token after first validation. A refresh
-     * would fail token validation and destroy the session.
+     * @param int $timeout Seconds to wait before giving up
+     * @return bool True if app initialized, false if timeout
      */
-    private function waitForAppReady(): void
+    private function waitForAppReady(int $timeout = 30): bool
     {
         try {
-            $this->client->wait(30)->until(fn($driver) => $driver->executeScript(
+            $this->client->wait($timeout)->until(fn($driver) => $driver->executeScript(
                 'return document.getElementById("mainMenu")?.children.length > 0'
             ));
+            // Log state on success to verify hypothesis that koAvailable
+            // is always true when the menu renders successfully
+            $state = $this->client->executeScript(<<<'JS_WRAP'
+                return JSON.stringify({
+                    koAvailable: typeof ko !== 'undefined',
+                    mainMenuChildren: document.getElementById('mainMenu')?.children.length ?? 0
+                });
+            JS_WRAP);
+            fwrite(STDERR, "[E2E] waitForAppReady succeeded: {$state}\n");
+            return true;
         } catch (TimeoutException) {
-            // Gather page state to help diagnose timeout failures.
-            // Knowing whether the page loaded, Knockout.js initialized,
-            // and whether #mainMenu exists narrows down root causes.
-            try {
-                $diagnostics = (string) $this->client->executeScript(<<<'JS_WRAP'
-                    return JSON.stringify({
-                        url: location.href,
-                        readyState: document.readyState,
-                        title: document.title,
-                        koAvailable: typeof ko !== 'undefined',
-                        mainMenuExists: document.getElementById('mainMenu') !== null,
-                        mainMenuChildren: document.getElementById('mainMenu')?.children.length ?? 0,
-                        bodyLength: document.body?.innerHTML?.length ?? 0
-                    });
-                JS_WRAP);
-            } catch (\Throwable) {
-                $diagnostics = 'unable to gather diagnostics (executeScript failed)';
-            }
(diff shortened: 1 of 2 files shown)


kojiromike:
0% coverage is expected here, because the e2e test files aren't currently themselves covered.

PR #6775, opened 2023-08-20
feat: Enable a force new password on next login feature

robertdown:
Fixes #6774

PR #10610, opened 2026-02-06
feat(auth): allow admins to force password reset on next login

### LINKED ISSUE #6774: feat: Allow admins to force a password reset upon next login of user

kojiromike:
Fixes #6774

Adds an administrator-settable flag that prompts a user to change their password at their next login.

- Add a `force_new_password` column to `users_secure`, with a matching upgrade step in `sql/8_3_0-to-8_4_0_upgrade.sql` and a `$v_database` bump so existing installs pick it up
- Add a "Force Password Change" checkbox to the user edit form (Administration > Users), hidden for LDAP users
- Open the Change Password tab first at login when the flag is set for that user
- Clear the flag automatically when the user's password is successfully changed
- Add an end-to-end test covering the login-side flow

#### Scope note

The Change Password tab is opened first, but it does not block navigation — a user can switch to another tab and continue working without changing their password, and will be prompted again at their next login. This is the same behavior as the existing `password_expiration_days` prompt. Hard enforcement would require gating the main screen itself, which is a larger change and not part of this PR.

LDAP users are excluded, matching the password expiration feature.

This takes over from #6775 (WIP by @robertdown).

### DIFF
--- a/interface/main/main_screen.php
+++ b/interface/main/main_screen.php
@@ -439,30 +439,49 @@ if (OEGlobalsBag::getInstance()->getBoolean('login_into_facility')) {
     }
 }
 
-// Fetch the password expiration date (note LDAP skips this)
+// Fetch the password expiration date and the forced-password-change flag (note LDAP skips this)
 $is_expired = false;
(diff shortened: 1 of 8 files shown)


bradymiller (approved):

bradymiller:
hi @kojiromike , code looks great! (I am assuming is testing well for you)

Recommend incrementing $v_database in version.php

Then good to go.

kojiromike:
Not working as expected yet. Adding tests.

PR #10616, opened 2026-02-07
feat(phpstan): add rule to forbid empty() in favor of explicit checks

### LINKED ISSUE #10613: Add PHPStan rule to forbid empty() in favor of explicit checks
## Problem

The `empty()` function has surprising behavior that leads to bugs:
- `empty("0")` returns `true` (string zero is considered empty)
- `empty(0)` returns `true` (integer zero is considered empty)
- `empty(0.0)` returns `true` (float zero is considered empty)
- `empty([])` returns `true` (empty array)
- `empty(null)` returns `true`
- `empty(false)` returns `true`
- `empty("")` returns `true`

The `empty("0")` case is particularly egregious — a non-empty string containing a digit should not be considered "empty".

## Proposed Solution

Add a custom PHPStan rule that forbids calling `empty()` and suggests explicit type-appropriate checks instead:
- `$value === null` for null checks
- `$value === ''` for empty string checks
- `$value === null || $value === ''` for nullable strings
- `count($array) === 0` or `$array === []` for array checks

## Acceptance Criteria

- PHPStan reports an error when `empty()` is used
- Error message suggests using explicit checks based on the expected type

kojiromike:
Fixes #10613

## Summary

- Add new `ForbiddenEmptyRule` that catches all `empty()` usage
- The rule explains why `empty()` is problematic (especially `empty("0")` returning `true`)
- Suggests explicit type-appropriate checks instead

## Changes proposed

- `.phpstan/ForbiddenEmptyRule.php` - New rule targeting `PhpParser\Node\Expr\Empty_`
- `.phpstan/extension.neon` - Register the new rule
- `.phpstan/README.md` - Document the rule with before/after examples
- `.phpstan/baseline/openemr.forbiddenEmpty.php` - Baseline for 8967 existing usages
- `.phpstan/baseline/loader.php` - Include the new baseline

## AI Disclosure

Yes


Firehed:
I'm pretty sure we get this automatically in the higher PHPStan levels, or perhaps in the strict rules package. I think in any case we can leverage that even if we don't turn the whole thing on.

See also #10622 which will clash with this.

Firehed (changes requested):
Adding Request Changes review per previous comment, but I am in favor of what this change is doing.

Confirmed it's in the strict rules package: https://github.com/phpstan/phpstan-strict-rules, which documents how to toggle specific rules.

kojiromike:
@firehed re-implemented using phpstan-strict-rules

.phpstan/baseline/empty.variable.php
```diff
@@ -3,406 +3,487 @@
 $ignoreErrors = [];
 $ignoreErrors[] = [
     'message' => '#^Variable \\$code_list in empty\\(\\) always exists and is not falsy\\.$#',
+    'identifier' => 'empty.variable',
```
Firehed: do you generate baselines differently than I do? I've seen this get added several times, which I find rather perplexing.

.phpstan/README.md
```diff
@@ -167,6 +167,46 @@ class SomeServiceTest extends TestCase
...
+
+**After (✅ Recommended):**
+```php
+// Be explicit about what you're checking
+if ($value === null) {              // Check for null
+if ($value === '') {                // Check for empty string
+if ($value === null || $value === '') {  // Check for null or empty string
+if (count($array) === 0) {          // Check for empty array
+if (!$array) {                      // Boolean check on array (empty array is falsy)
+```
```
Firehed, line 204: Might want to add `isset()` for dealing with the leaky globals (probably unavoidable today) and `array_key_exists` (typically input validation)

Firehed (approved):

.phpstan/baseline/empty.variable.php
kojiromike: @Firehed It's probably Claude, but IDK. I don't 100% trust phpstan itself, tbh. I've many times bitten by the weird difference between ci and local, so when the change is small I've let Claude do it from time to time. But tbqh, I don't know.

Firehed: I'm pretty sure that _should_ be stable now (from what we've discussed, 100% of the issues were related to the varying signatures on duplicate function definitions; afiact all of the problem ones are fixed). For the sake of dogfooding, ideally we stick strictly with the documented tools and fix any other issues that crop up if they remain.

But yeah, I get where you're coming from. I specifically put instructions to use `phpstan-baseline` in my system claude.md file, if only to save time and tokens.

PR #10620, opened 2026-02-07
refactor: replace catch (Exception) with catch (Throwable)

### LINKED ISSUE #10618: refactor: replace catch (Exception) with catch (Throwable)
## Summary

The codebase has ~247 instances of `catch (Exception)` or `catch (\Exception)` that should be replaced with `catch (Throwable)` or `catch (\Throwable)`.

## Rationale

`\Throwable` is the base interface for all throwable values in PHP (since PHP 7.0). Using `\Exception` misses `\Error` subclasses, which means:

- **Type errors** (e.g., passing wrong argument types)
- **Parse errors** in evaluated code
- **Assertion errors**
- **Division by zero errors**

...are not caught when using `catch (Exception)`, potentially causing unexpected crashes.

When the intent is "catch everything that can go wrong", `\Throwable` is correct. If specific exception handling is needed, catching the specific exception type is preferable.

## Affected Files

All PHP files using the generic `catch (Exception)` pattern — approximately 247 instances across the codebase.

kojiromike:
## Summary

Fixes #10618

Replace all instances of generic `catch (Exception)` with `catch (Throwable)` across the codebase, and add a PHPStan rule to prevent future additions.

- Replace ~247 instances of `catch (Exception)` or `catch (\Exception)` with `catch (\Throwable)`
- Simplify one instance of `catch (\Exception | ExceptionInterface)` to `catch (\Throwable)`
- Add `ForbiddenCatchExceptionRule` PHPStan rule to flag any future `catch (Exception)` additions

## Rationale

`\Throwable` is the base interface for all throwable values in PHP (since PHP 7.0). Using `\Exception` misses `\Error` subclasses:

- Type errors (e.g., passing wrong argument types)
- Parse errors in evaluated code
- Assertion errors
- Division by zero errors

When the intent is "catch everything that can go wrong", `\Throwable` is correct.

## Testing

This is a safe refactoring that expands the scope of what's caught — no existing behavior is broken. Any code that previously caught `Exception` will now also catch `Error` subclasses.

## AI Disclosure

Yes - Claude Code

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,6 +2,10 @@ codecov:
   notify:
     manual_trigger: true
 
+ignore:
+- tests/PHPStan/         # Static analysis rules — not exercised by PHPUnit
+- tests/Rector/          # Rector rules — run outside PHPUnit coverage
+
 coverage:
   status:
     project:


interface/modules/zend_modules/module/Application/src/Application/Model/ApplicationTable.php
```diff
@@ -1,439 +0,0 @@
...
-use Exception;
-use Laminas\Db\Adapter\Adapter;
-use Laminas\Db\Adapter\Exception\ExceptionInterface;
-use Laminas\Db\TableGateway\AbstractTableGateway;
-use Laminas\Db\ResultSet\ResultSet;
-use Laminas\Db\TableGateway\Feature\GlobalAdapterFeature;
-use OpenEMR\Common\Logging\EventAuditLogger;
-use OpenEMR\Common\Database\QueryUtils;
-
-class ApplicationTable extends AbstractTableGateway
```
Firehed, line 25: uh, is this removal intentional?

interface/patient_file/report/custom_report.php
```diff
@@ -582,18 +582,22 @@ function getContent()
                                             $itpl = $pdf->importPage($i + 1);
                                             $pdf->useTemplate($itpl);
                                         }
-                                    } catch (Exception) {
+                                    } catch (\Throwable) {
                                         // chances are PDF is > v1.4 and compression level not supported.
                                         // regardless, we're here so lets dispose in different way.
                                         //
-                                        unlink($from_file_tmp_name);
+                                        if (isset($from_file_tmp_name)) {
```
kojiromike: why do this here if it's going to happen again on line 598?

interface/modules/zend_modules/module/Application/src/Application/Model/ApplicationTable.php
kojiromike: oops. Fixed

interface/patient_file/report/custom_report.php
kojiromike: why do this here and again on line 598?

kojiromike: Good catch - removed the redundant unlink from the catch block. The finally always runs and handles the cleanup.

interface/modules/zend_modules/module/Application/src/Application/Model/ApplicationTable.php
Firehed: I don't blame you though.

Firehed (approved):

PR #10636, opened 2026-02-08
fix(zend): include port in PDO DSN for non-default database ports

### LINKED ISSUE #10635: bug: Zend modules ignore non-default database port
### Describe the bug

The Zend modules configuration (`interface/modules/zend_modules/config/autoload/global.php`) does not consistently use `$GLOBALS['port']` in PDO DSN strings. This causes database connection failures for sites using a non-standard MySQL port.

### Affected areas

1. The main database connection DSN
2. PDO connection for multiple database support
3. Additional adapter DSNs for multiple databases

### Expected behavior

All PDO DSN strings should include the port from `$GLOBALS['port']` to support non-default database ports.

### To Reproduce

1. Configure OpenEMR to use a non-default MySQL port (e.g., 3307)
2. Attempt to use Zend module functionality
3. Connection fails because the DSN uses default port (3306)

### OpenEMR Version

Current master

### Additional context

Reported via PR #7431 by @wjassell

kojiromike:
Fixes #10635

Database connections fail for sites using non-standard MySQL ports because the Zend modules configuration doesn't include the port in PDO DSN strings.

- Add port parameter to main database connection DSN
- Add port parameter to multiple database support PDO connection
- Add port parameter to per-database adapter DSNs (remove redundant separate `port` key since PDO uses DSN)
- Update PHPStan baseline for new `$GLOBALS` access

#### AI Disclosure

Yes

---

This is a fresh implementation of the fix from #7431, which became too stale to merge cleanly.

Credit to @wjassell for originally identifying and reporting this issue.

### DIFF
--- a/src/Common/Utils/ValidationUtils.php
+++ b/src/Common/Utils/ValidationUtils.php
@@ -222,18 +222,14 @@ class ValidationUtils
     /**
      * Validates a TCP/UDP port number.
      *
-     * Accepts empty string (no port specified) or valid port numbers (1-65535).
-     * Port 0 is reserved and not considered valid for connection purposes.
+     * Valid port numbers are 1-65535. Port 0 is reserved and not considered
+     * valid for connection purposes.
      *
      * @param string $port The port to validate
-     * @return bool True if empty string or valid port number, false otherwise
+     * @return bool True if valid port number, false otherwise
      */
     public static function isValidPort(string $port): bool
     {
-        if ($port === '') {
-            return true;
-        }
-
         return self::validateInt($port, 1, 65535) !== false;
     }
 
--- a/src/Common/Database/DbUtils.php
+++ b/src/Common/Database/DbUtils.php
@@ -28,7 +28,7 @@ class DbUtils
      */
     public static function buildMysqlDsn(string $dbname, string $host, string $port = ''): string
     {
-        if (!ValidationUtils::isValidPort($port)) {
+        if ($port !== '' && !ValidationUtils::isValidPort($port)) {
             throw new InvalidArgumentException("Invalid port: '$port'. Must be empty or a number between 1 and 65535.");
         }
 


interface/modules/zend_modules/config/autoload/global.php
```diff
@@ -96,10 +95,9 @@
 return [
     'db' => [
         'driver'         => 'Pdo',
-        'dsn'            => 'mysql:dbname=' . ($sqlConf['dbase'] ?? '') . ';host=' . ($sqlConf['host'] ?? ''),
+        'dsn'            => 'mysql:dbname=' . ($sqlConf['dbase'] ?? '') . ';host=' . ($sqlConf['host'] ?? '') . ';port=' . ($sqlConf['port'] ?? ''),
```
kojiromike: Will this work correctly with `;port=` if $sqlConf['port'] is blank? Would it be better to build the dsn and only provide the entries that are set?

```php
dsnParts = array_filter([
    'dbname' => $sqlConf['dbase'] ?? '',
    'host' => $sqlConf['host'] ?? '',
    'port' => $sqlConf['port'] ?? '',
]);
```

then use implode and array map to turn dsnParts into the necessary string?

kojiromike: Good catch. Added a `$buildMysqlDsn` closure that filters out empty values before building the DSN string.

src/Common/Utils/ValidationUtils.php
```diff
@@ -219,6 +219,24 @@ public static function isValidUuid(string $uuid): bool
...
+     *
+     * @param string $port The port to validate
+     * @return bool True if empty string or valid port number, false otherwise
+     */
+    public static function isValidPort(string $port): bool
+    {
+        if ($port === '') {
+            return true;
+        }
+
```
kojiromike:
```suggestion
```

kojiromike: `isValidPort` should only return a valid port. An empty string can be allowed by buildMysqlDsn, but not here.

kojiromike:
Addressed in 2abd0f1:

- `isValidPort()` now strictly validates port numbers (1-65535), returning `false` for empty string
- Empty port handling moved to `buildMysqlDsn()`, which checks `$port !== ''` before calling the validator

PR #10664, opened 2026-02-09
fix(sql): wrap scalar values in arrays for SQL function params

Firehed:
Part of #10149.

Several SQL function calls were passing scalar values (int, string) or null directly as the `$binds` parameter instead of wrapping them in arrays. While the functions handle this gracefully at runtime due to loose typing, it violates the expected type signature and was flagged by PHPStan at level 5.

- `controllers/C_Document.class.php`: wrap `$document_id` in array for `sqlStatement()`
- `src/Services/DocumentService.php`: wrap `str_replace()` results in arrays for `sqlQuery()`
- `src/Services/PhoneNumberService.php`: wrap `$this->foreignId` in array for `sqlQuery()`
- `src/Services/Utils/TranslationService.php`: wrap `$languageId` in array for `sqlStatement()`
- `src/ClinicalDecisionRules/Interface/RulesPlanMappingEventHandlers.php`: remove unnecessary `null` second arguments from `sqlStatement()` calls (queries have no placeholders)

Yes

The changes are minimal single-line fixes (wrapping values in `[]` or removing `, null`). These were identified using PHPStan static analysis and fixed with assistance from Claude. The fixes are straightforward type corrections that do not warrant inline AI attribution comments.

👨 I actually had an equivalent change in a different branch that's gotten lost in my local pile of stuff.

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -21,6 +21,7 @@ if ($response !== true) {
 }
 
 use Dotenv\Dotenv;
+use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Core\Kernel;
 use OpenEMR\Core\ModulesApplication;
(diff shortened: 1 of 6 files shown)


Firehed:
Found a few more issues following PHPStan upgrade.

interface/globals.php
```diff
@@ -745,11 +746,7 @@ function GetCallingScriptName()
...
 // module configurations
 // upgrade fails for versions prior to 4.2.0 since no modules table
-try {
-    $checkModulesTableExists = sqlQueryNoLog('SELECT 1 FROM `modules`', false, true);
-} catch (\Throwable $ex) {
-    error_log(errorLogEscape($ex->getMessage() . $ex->getTraceAsString()));
-}
+$checkModulesTableExists = QueryUtils::existsTable('modules');
 
 if (!empty($checkModulesTableExists)) {
```
igormukhingmailcom, line 751: Should `!empty` be removed or whole condition replaced to `if (QueryUtils::existsTable('modules')) {`?

PR #10674, opened 2026-02-10
feat(internals): add ServiceContainer for centralized service access

### LINKED ISSUE #8929: Introduce Service Locator with module overrides

Firehed:
Fixes #8929, alternative to #8944 (well, some of it)

Introduces a ServiceContainer class that provides centralized access to a selection of common services via static methods returning interfaces. This is a stepping stone toward full dependency injection, allowing code to more easily reference interfaces rather than concrete implementations. I consider it two steps forward, one step back.

Modules can override default implementations during bootstrap using `ServiceContainer::override()`, addressing the core use case from #8929 without the complexity of dealing with the event system. Due to order of operations this may not be perfect, but should handle several common cases from that issue.

Notably, this doesn't really aim to lay _any_ groundwork for next-gen modules; it's intended strictly to ease the transition on existing code (hence being in the BC namespace). It's similar in nature to OEGlobalsBag vs direct Globals. Next-gen modules should a) use a proper PSR-11 container and b) get hooked _so early_ in the request lifecycle that most of the overrides aren't even needed.

Logical next-steps for this, if we want to go with it, would be to add PHPStan rules that disallow direct instantiation of (some of) the classes this provides (systemlogger and cryptogen, for two) and guides towards referencing this. And do the migrations. Once we have a true DI container (with autowiring) set up, we can both migrate the internals of this to use it and deprecate it in favor of the container.

There's probably other classes/interfaces that this should provide!

- Add `ServiceContainer` class in `src/BC/` with static accessors for:
  - PSR-3: `getLogger()` → `LoggerInterface`
  - PSR-17: `getRequestFactory()`, `getResponseFactory()`, `getServerRequestFactory()`, `getStreamFactory()`, `getUploadedFileFactory()`, `getUriFactory()`
  - PSR-20: `getClock()` → `ClockInterface`
  - `getCrypto()` → `CryptoInterface`
  - `getTwig()` → Twig `Environment`
- Add `override(string $interface, object $instance)` method for module overrides
  - Uses PHPStan generics for type safety
  - Validates instance implements the interface
  - Last override wins
  - Lazy instantiation: defaults only created when no override exists
- Add isolated unit tests for almost all methods

#### If you answered yes:

Most code in this PR was generated with assistance from Claude (Anthropic). The code was reviewed and approved by a human before inclusion. Files affected:
- `src/BC/ServiceContainer.php` - AI-assisted
- `tests/Tests/Isolated/BC/ServiceContainerTest.php` - AI-assisted

---

### DIFF
--- a/src/BC/ServiceContainer.php
+++ b/src/BC/ServiceContainer.php
@@ -18,7 +18,6 @@ use OpenEMR\Common\Crypto;
 use OpenEMR\Common\Logging;
 use Lcobucci\Clock\SystemClock;
 use OpenEMR\Common\Http\Psr17Factory;
-use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Core\OEGlobalsBag;
 use Psr\Clock\ClockInterface;
 use Psr\Http\Message\{
@@ -29,7 +28,6 @@ use Psr\Http\Message\{
     UploadedFileFactoryInterface,
     UriFactoryInterface,
 };
-use Twig\Environment;
 
 /**
  * Utility class for accessing common system services.
@@ -125,21 +123,6 @@ class ServiceContainer
     public static function getStreamFactory(): StreamFactoryInterface
     {
         return self::resolve(StreamFactoryInterface::class) ?? new Psr17Factory();
-    }
-
-    /**
-     * Get a configured Twig environment.
-     *
-     * @param string|null $path Additional template path to include alongside
-     *                          the default /templates directory
-     */
-    public static function getTwig(?string $path = null): Environment
-    {
-        if (($env = self::resolve(Environment::class)) !== null) {
-            return $env;
-        }
-        $kernel = OEGlobalsBag::getInstance()->getKernel();
-        return (new TwigContainer($path, $kernel))->getTwig();
     }
 
     public static function getUploadedFileFactory(): UploadedFileFactoryInterface
--- a/tests/Tests/Isolated/BC/ServiceContainerTest.php
+++ b/tests/Tests/Isolated/BC/ServiceContainerTest.php
@@ -4,7 +4,6 @@
  * ServiceContainer Smoke Tests
  *
  * Tests that ServiceContainer methods return the expected interface types.
- * getTwig() is excluded as it requires global state.
  *
  * @package   OpenEMR
  * @link      http://www.open-emr.org


src/BC/ServiceContainer.php
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Get a configured Twig environment.
+     *
+     * @param string|null $path Additional template path to include alongside
+     *                          the default /templates directory
+     */
+    public static function getTwig(?string $path = null): Environment
+    {
+        /** @var Kernel $kernel */
+        $kernel = OEGlobalsBag::getInstance()->get('kernel');
```
kojiromike: I still think OEGlobalsBag should have a typed getKernel method. This isn't supposed to be a slippery slope towards having a bunch of typed getters in there. This is one thing that OpenEMR uses _all the time_ that should be typed. I really don't understand the resistance.

Firehed: I'm fine/indifferent on this one, at least for the medium term. Didn't feel addressing it made sense for the scope of this change since it was basically a ten-minute experiment.

Firehed: I see you added #10679 so I'll adjust this accordingly.

src/BC/ServiceContainer.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    public static function getServerRequestFactory(): ServerRequestFactoryInterface
+    {
+        return self::resolve(ServerRequestFactoryInterface::class) ?? new Psr17Factory();
+    }
+
+    public static function getStreamFactory(): StreamFactoryInterface
+    {
+        return self::resolve(StreamFactoryInterface::class) ?? new Psr17Factory();
```
kojiromike, line 125: Is it intentional that `$path` is ignored when an override is registered? If someone does:

```php
ServiceContainer::override(Environment::class, $customTwig);
// Later...
$twig = ServiceContainer::getTwig('/some/module/templates');
```

The `$path` parameter is silently ignored and the override is returned. This might be the right behavior (overrides are absolute), but could be surprising. Worth a brief note in the docblock?

Firehed: I'm going to remove Twig for now, since it has some special-casing that I don't want to address yet. I think it's worth adding in later (including a solve for this), but I don't want to rush it through since it's the only method that accepts an argument at all.

PR #10679, opened 2026-02-10
feat(core): add typed getKernel() method to OEGlobalsBag

### LINKED ISSUE #10678: Add typed getKernel() method to OEGlobalsBag
## Problem

`$GLOBALS['kernel']` access is untyped throughout the codebase. There are 100+ usages like:

```php
$GLOBALS['kernel']->getEventDispatcher()->dispatch($event, ...);
```

Since `OEGlobalsBag::get()` returns `mixed`, PHPStan can't verify these calls and IDEs can't provide autocomplete.

Some code already adds defensive checks:
```php
// sqlconf.php
if ($GLOBALS['kernel'] instanceof Kernel) { ... }

// AuthorizationController
if (!$oeKernel instanceof Kernel) {
    throw new RuntimeException("...");
}
```

## Proposed Solution

Add a typed getter to `OEGlobalsBag`:

```php
public function getKernel(): Kernel
{
    $kernel = $this->get('kernel');
    if (!$kernel instanceof Kernel) {
        throw new \RuntimeException('Kernel not initialized');
    }
    return $kernel;
}
```

Then migrate call sites from `$GLOBALS['kernel']->...` to `OEGlobalsBag::getInstance()->getKernel()->...`.

## Notes

- All `$GLOBALS['kernel']` usages only need `Kernel` methods (primarily `getEventDispatcher()`)
- `OEHttpKernel` (used in REST API) is a separate class hierarchy and accessed via Symfony events, not `$GLOBALS['kernel']`
- This is incremental progress toward replacing global state with typed accessors

kojiromike:
Fixes #10678

## Description

Add a typed `getKernel()` method to `OEGlobalsBag` for type-safe access to the OpenEMR Kernel instance.

## Changes

- Add `getKernel(): Kernel` method that returns the kernel or throws `RuntimeException` if not initialized

## Test Plan

- [ ] Verify PHPStan passes
- [ ] Existing tests pass (kernel is set up in globals.php before any code runs)

## AI Disclosure

Yes

### DIFF
--- a/.phpstan/baseline/return.type.php
+++ b/.phpstan/baseline/return.type.php
@@ -3292,7 +3292,7 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../src/BC/Database.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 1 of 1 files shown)


kojiromike:
The semgrep findings are false positives, but do indicate #10682, which I'll follow up on.

PR #10683, opened 2026-02-10
refactor(security): use AccessDeniedHelper for ACL denial patterns

### LINKED ISSUE #10682: Refactor ACL denial patterns to use AccessDeniedHelper
## Summary

Many files manually render the `core/unauthorized.html.twig` template and exit when ACL checks fail. This bypasses proper logging and auditing that `AccessDeniedHelper::deny()` provides.

## Files affected

Found via `git grep "TwigContainer.*unauthorized"`:

- `ccr/createCCR.php`
- `ccr/display.php` (uses `error/400.html.twig` with 401 status)
- `contrib/util/dupecheck/index.php`
- `contrib/util/dupecheck/mergerecords.php`
- `controllers/C_Document.class.php` (2 locations)
- `controllers/C_PracticeSettings.class.php`
- `custom/import_xml.php`
- `gacl/admin/*.php` (10+ files)
- `interface/reports/patient_list_creation.php`

Plus several files recently converted in #10679.

## Current pattern

```php
if (!AclMain::aclCheckCore(...)) {
    echo (new TwigContainer(null, $kernel))->getTwig()
        ->render('core/unauthorized.html.twig', ['pageTitle' => $title]);
    exit;
}
```

## Proposed pattern

```php
if (!AclMain::aclCheckCore(...)) {
    AccessDeniedHelper::deny(
        "ACL check failed for: " . $title,
        beforeExit: function () use ($title): void {
            echo (new TwigContainer(null, OEGlobalsBag::getInstance()->getKernel()))
                ->getTwig()
                ->render('core/unauthorized.html.twig', ['pageTitle' => $title]);
        }
    );
}
```

## Benefits

1. Proper audit logging of access denials
2. Consistent HTTP status codes
3. SystemLogger warning for monitoring
4. Standardized pattern across codebase

kojiromike:
Fixes #10682

## Summary

Refactor ~155 files from manual template rendering to use `AccessDeniedHelper`, which provides proper audit logging, SystemLogger warnings, and consistent HTTP 403 status codes.

## Changes

- Add `AccessDeniedResponseFormat::None` to suppress default output when using custom `beforeExit` callbacks
- Add `denyWithTemplate()` method for rendering the unauthorized template
- Add `createDeniedResponse()` for controller patterns that need Response objects instead of `exit()`
- Fix bug in `C_EncounterVisitForm`: was rendering non-existent template with no echo statement
- Add TODO comment to `linked_documents.php` (issue #10686) for parameter validation incorrectly using unauthorized template

## Baseline changes (all reductions)

- `patient_list_creation.php`: 28 → 27 occurrences (extracted array values)
- `C_EncounterVisitForm.class.php` aclCheckCore: 2 → 1 (type narrowing)
- Removed 3 stale baseline entries for files no longer using TwigContainer

## Test plan

- [x] All files pass PHP syntax check
- [x] All files pass PHPStan at level 10
- [x] Pre-commit hooks pass
- [ ] Manual testing of ACL denial on sample pages

## AI Disclosure

Yes

### DIFF
--- a/contrib/util/dupecheck/mergerecords.php
+++ b/contrib/util/dupecheck/mergerecords.php
@@ -20,11 +20,10 @@ class DupeCheckMergeRecordsIsDeprecated
 require_once("../../../interface/globals.php");
 require_once("../../../library/pnotes.inc.php");
 
+use OpenEMR\Common\Acl\AccessDeniedHelper;
 use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Logging\EventAuditLogger;
-use OpenEMR\Common\Twig\TwigContainer;
-use OpenEMR\Core\OEGlobalsBag;
 
 if (!empty($_POST)) {
     if (!CsrfUtils::verifyCsrfToken($_POST["csrf_token_form"])) {
@@ -45,8 +44,7 @@ if (!empty($_GET)) {
 }
 
 if (!AclMain::aclCheckCore('admin', 'super')) {
-    echo (new TwigContainer(null, OEGlobalsBag::getInstance()->getKernel()))->getTwig()->render('core/unauthorized.html.twig', ['pageTitle' => xl("Merge Records")]);
-    exit;
+    AccessDeniedHelper::denyWithTemplate("ACL check failed for admin/super: Merge Records", xl("Merge Records"));
 }
 
 ?>
--- a/interface/reports/cdr_log.php
+++ b/interface/reports/cdr_log.php
@@ -16,14 +16,15 @@ require_once "$srcdir/options.inc.php";
 require_once "$srcdir/clinical_rules.php";
 
 use OpenEMR\ClinicalDecisionRules\Interface\ControllerRouter;
+use OpenEMR\Common\Acl\AccessDeniedException;
+use OpenEMR\Common\Acl\AccessDeniedHelper;
+use OpenEMR\Common\Csrf\CsrfInvalidException;
+use OpenEMR\Common\Logging\SystemLogger;
 use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Core\OEGlobalsBag;
 use Symfony\Component\HttpFoundation\Request;
-use OpenEMR\Common\Csrf\CsrfInvalidException;
-use OpenEMR\Common\Acl\AccessDeniedException;
-use OpenEMR\Common\Logging\SystemLogger;
-use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 try {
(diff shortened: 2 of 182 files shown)


contrib/util/dupecheck/mergerecords.php
```diff
@@ -45,8 +44,7 @@ public function __construct()
 }
 
 if (!AclMain::aclCheckCore('admin', 'super')) {
```
Firehed, line 46: Consideration (future?): `AclMain::aclEnforceCore(...)` that auto-denies?

kojiromike: I thought about that, but I like that `deny*` has a `never` return type, and `aclCheckCore` is a boolean. If I implemented a `maybeDeny` then we would have a little less clarity about dead code.

src/Common/Acl/AccessDeniedHelper.php
```diff
@@ -77,4 +80,73 @@ public static function deny(
...
+     * Convenience method for the common pattern of rendering the unauthorized
+     * template when an ACL check fails. Handles logging, auditing, and exits.
+     *
+     * @param string $comment   Audit log comment describing what was attempted
+     * @param string $pageTitle Title to display on the unauthorized page
+     * @param string $auditEvent Event name for EventAuditLogger
+     */
+    public static function denyWithTemplate(
+        string $comment,
+        string $pageTitle,
```
Firehed, line 96: Seeing that most paths are calling xl() ahead of this, it might be worth moving that into here. Fine as-is though.

interface/reports/cdr_log.php
```diff
@@ -16,14 +16,15 @@
 require_once "$srcdir/clinical_rules.php";
 
 use OpenEMR\ClinicalDecisionRules\Interface\ControllerRouter;
+use OpenEMR\Common\Acl\AccessDeniedException;
```
Firehed, line 19: are these left over or resolving other issues in the file? most don't seem used

kojiromike: Hmm, I can't actually find what you're referring to. `AccessDeniedException` still seems to be used where I've imported it.

src/Common/Acl/AccessDeniedHelper.php
bradymiller: fyi, for the translation pipeline that collect the new constants from the codebase, generally best to keep the xl with the actual constant (rather than a variable), if possible. So good decision to not bring it in :)

Firehed: Ah, makes sense. Thanks for the tip!

interface/reports/cdr_log.php
Firehed: Oh, this was a sort not an add. Ignore me.

src/Common/Acl/AccessDeniedHelper.php
kojiromike: That reminds me. Now that we're on phpstan level 10, we can assert that xl calls should have a literal.

Issue #8756, opened 2025-08-08
Cases where OpenEMR code exits without appropriate context or handling


kojiromike:
(This report is partly written by Claude, but guided by me. The main issue is exiting in a web context, whether failure or not, creates missing context issues.)

# PHP Exit Status Analysis Report

## Overview

This report identifies PHP files in the OpenEMR codebase that use `exit;` or `exit();` statements without proper context-aware exit handling. The main issues are:

1. **CLI contexts**: Error conditions should use `exit(1)` for proper shell script integration
2. **Web contexts**: Exit behavior is often appropriate but should differentiate success/error
3. **Testing contexts**: Should throw exceptions instead of exiting
4. **Library contexts**: Should return status or throw exceptions instead of exiting

## Statistics

- **Total PHP files analyzed**: 3,762
- **Files with `exit;` statements**: 327
- **Files with `exit();` statements**: 113
- **Files already using context detection**: 19 (via `php_sapi_name()`)

### Existing Context-Aware Patterns
The codebase already uses context detection in several places:
- `contrib/util/billing/load_fee_schedule.php`: CLI-only script
- `contrib/util/ccda_import/import_ccda.php`: CLI-only script  
- `library/allow_cronjobs.php`: Returns early if not CLI
- And 16 other files using `php_sapi_name()` checks

## Key Findings by Context

The following locations need context-aware exit handling:

#### _rest_config.php
- **Line 346**: OAuth scope validation failure (HTTP 401) - *should use `exit(1)` for error*
- **Line 351**: Empty global scope array error (HTTP 401) - *should use `exit(1)` for error*
- **Line 323**: Site validation failure (HTTP 400) - *should use `exit(1)` for error*
- **Line 391**: Site validation failure (HTTP 400) - *should use `exit(1)` for error*

#### apis/dispatch.php
These are REST API handlers where exit is appropriate, but should differentiate success/error:
- **Line 80**: Token verification failure - *should use `exit(1)`*
- **Line 165**: Token verification failure - *should use `exit(1)`*
- **Line 210**: Token verification failure - *should use `exit(1)`*
- **Line 106**: API site error (HTTP 400) - *should use `exit(1)`*
- **Line 124**: Missing userid/tokenid (HTTP 400) - *should use `exit(1)`*
- **Line 193**: CSRF validation failure (HTTP 401) - *should use `exit(1)`*
- **Line 220**: Invalid token (HTTP 401) - *should use `exit(1)`*
[...]

PR #10690, opened 2026-02-11
refactor(globals): Switch from error_log to standard system logger, fix status codes

Firehed:
The `SystemLogger` class (or, more accurately, any PSR-3 logger, which it is) is preferred in favor of raw `error_log` calls. This fixes several. There were also a number of early-exits that didn't send an appropriate status code so I fixed those too (part of #8756)

- Sets up the autoloader as early as possible, negating the need for a manual require (the surrounding comments were obsolete)
- Converts `error_log` to use the `SystemLogger` throughout
- Corrects an inconsistency I noticed in an error handling path

Yes, basic mechanical work only.

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -19,6 +19,7 @@ require_once dirname(__DIR__) . '/vendor/autoload.php';
 // Checks if the server's PHP version is compatible with OpenEMR:
 $response = OpenEMR\Common\Compatibility\Checker::checkPhpVersion();
 if ($response !== true) {
+    http_response_code(500);
     die(htmlspecialchars($response));
 }
 
@@ -37,11 +38,13 @@ $logger = new SystemLogger();
 
 // Throw error if the php openssl module is not installed.
 if (!(extension_loaded('openssl'))) {
+    http_response_code(500);
     $logger->critical('OpenEMR is not working since the php openssl module is not installed');
     die("OpenEMR Error : OpenEMR is not working since the php openssl module is not installed.");
 }
 // Throw error if the openssl aes-256-cbc cipher is not available.
 if (!(in_array('aes-256-cbc', openssl_get_cipher_methods()))) {
+    http_response_code(500);
     $logger->critical('OpenEMR is not working since the openssl aes-256-cbc cipher is not available');
     die("OpenEMR Error : OpenEMR is not working since the openssl aes-256-cbc cipher is not available.");
 }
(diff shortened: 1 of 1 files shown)


interface/globals.php
```diff
@@ -13,15 +13,19 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+// Set up autoloader as early as possible
+require_once dirname(__DIR__) . '/vendor/autoload.php';
+
 // Checks if the server's PHP version is compatible with OpenEMR:
-require_once(__DIR__ . "/../src/Common/Compatibility/Checker.php");
```
bradymiller, line 17: As I recall, this actually breaks in older versions of PHP when run autoloader first.

bradymiller: @kojiromike , @Firehed  (if there is php code being run during autoloader that is not compatible with a legacy php version, then will error out instead of getting a simply message that php version is not compatible)

Firehed: I'll do a quick follow-up here to check if there's anything that looks problematic. Thanks for the heads-up.

kojiromike: @bradymiller what legacy versions of php do we want to support here?

Firehed: #10699 has details on what versions would cause problems. In practice all this does is swap an error like this:

```
Parse error: syntax error, unexpected ')', expecting variable (T_VARIABLE) in /in/jITmh on line 55

Process exited with code 255.
```

with a slightly more pleasant one. OpenEMR still won't run on PHP < 8.2.

bradymiller: hi, Does this codeblock need to remain under the purview of autoloader. Would like any php version below what we set to get that message since is a common problem.

kojiromike: @bradymiller I think you're saying that we regularly get bug reports and requests for help from people who are using a too-old version of php, is that right?

bradymiller: This has been a common problem by diy users in the past. Just seems like having a straightforward message here makes sense unless there is more benefit derived by putting this below the autoloader.

kojiromike: @bradymiller would it make sense to resolve this in setup.php and the installer? AFAIK those are the only two places you can _start_ running openemr with an old version of php.

bradymiller: hi @kojiromike , It already runs there and in all other upgrade points of entry (main path for this issue to crop up is folks using old php versions with upgrading) and the admin.php script. Not showstopper and ok with keeping it where it is.

PR #10698, opened 2026-02-11
fix: rename 'uncomplete' mode to 'incomplete' in AMC tracking

### LINKED ISSUE #10382: fix: rename 'uncomplete' mode to 'incomplete' in AMC tracking
## Summary

The AMC (Automated Measure Calculation) tracking system uses `"uncomplete"` as a mode value, which should be `"incomplete"`.

This affects:
- `interface/patient_file/encounter/forms.php:241` - sends mode value
- `interface/reports/amc_tracking.php:115` - uses `"uncomplete_safe"`
- `library/amc.php:64-66` - handles `"uncomplete"` and `"uncomplete_safe"` modes

## Why this wasn't fixed in the spellcheck PR

Changing the string value in forms.php without updating the backend handler in amc.php would break the functionality. This requires a coordinated change across all files.

## Suggested fix

1. Update `library/amc.php` to accept both `"incomplete"` and `"uncomplete"` (for backwards compatibility during transition)
2. Update all callers to use `"incomplete"`
3. Optionally deprecate `"uncomplete"` after a release cycle

Found during spellcheck PR #10370.

evolvomind:
Fixes #10382 

### Short description of what this resolves:

The AMC (Automated Measure Calculation) tracking system used "uncomplete" as a mode value in several files, which is not a valid English word. This PR corrects it to "incomplete" across all callers and the backend handler, while preserving backward compatibility for any third-party code or modules that may still send the old value.

### Changes proposed in this pull request:

- library/amc.php: Updated processAmcCall() to accept "incomplete" and "incomplete_safe" as the primary mode values, while still accepting the deprecated "uncomplete" and "uncomplete_safe" for backward compatibility. Renamed internal functions amcUnComplete() / amcUnCompleteSafe() to amcInComplete() / amcInCompleteSafe(). Updated the $date_created parameter comment to reference the corrected mode name.
- interface/patient_file/encounter/forms.php: Changed the JavaScript mode value sent when unchecking the "Medication Reconciliation Performed?" checkbox from "uncomplete" to "incomplete".
- interface/reports/amc_tracking.php: Changed the JavaScript mode value sent when unchecking the "Medical Records Sent" checkbox in the provide_rec_pat handler from "uncomplete_safe" to "incomplete_safe".
- .phpstan/baseline/openemr.noGlobalNsFunctions.php: Updated 2 baseline entries to reflect the renamed functions.
- .phpstan/baseline/missingType.parameter.php: Updated 10 baseline entries to reflect the renamed functions.

### Does your code include anything generated by an AI Engine? Yes / No

No.

### DIFF
--- a/.codespell-ignore-words.txt
+++ b/.codespell-ignore-words.txt
@@ -117,5 +117,4 @@ esy
 hightlight
 occurance
 seperator
-uncomplete
 vlaue
--- a/library/amc.php
+++ b/library/amc.php
@@ -61,11 +61,9 @@ function processAmcCall($amc_id, $complete, $mode, $patient_id, $object_category
         amcComplete($amc_id, $patient_id, $object_category, $object_id);
     } elseif ($mode == "complete_safe") {
         amcCompleteSafe($amc_id, $patient_id, $object_category, $object_id, $date_created);
-    } elseif ($mode == "incomplete" || $mode == "uncomplete") {
-        // "uncomplete" is accepted for backward compatibility but deprecated; use "incomplete" instead
+    } elseif ($mode == "incomplete") {
         amcInComplete($amc_id, $patient_id, $object_category, $object_id);
-    } elseif ($mode == "incomplete_safe" || $mode == "uncomplete_safe") {
-        // "uncomplete_safe" is accepted for backward compatibility but deprecated; use "incomplete_safe" instead
+    } elseif ($mode == "incomplete_safe") {
         amcInCompleteSafe($amc_id, $patient_id, $object_category, $object_id, $date_created);
     } elseif ($mode == "soc_provided") {
         amcSoCProvided($amc_id, $patient_id, $object_category, $object_id);


kojiromike:
Thanks for tackling this! The change looks good overall.

One thought: is the backward compatibility for the old `"uncomplete"` / `"uncomplete_safe"` values necessary? This is an internal AJAX endpoint called by OpenEMR's own JavaScript, not a documented public API. I'm not aware of third-party modules that would be calling `processAmcCall()` directly with hardcoded mode values.

If we drop the backward compatibility, you could simplify this to just:

```php
} elseif ($mode == "incomplete") {
    amcInComplete($amc_id, $patient_id, $object_category, $object_id);
} elseif ($mode == "incomplete_safe") {
    amcInCompleteSafe($amc_id, $patient_id, $object_category, $object_id, $date_created);
}
```

That would also let you remove `uncomplete` from `.codespell-ignore-words.txt` (line 120), so the spellchecker catches any future uses of the non-word.

What do you think?

evolvomind:
> Thanks for tackling this! The change looks good overall.
> 
> One thought: is the backward compatibility for the old `"uncomplete"` / `"uncomplete_safe"` values necessary? This is an internal AJAX endpoint called by OpenEMR's own JavaScript, not a documented public API. I'm not aware of third-party modules that would be calling `processAmcCall()` directly with hardcoded mode values.
> 
> If we drop the backward compatibility, you could simplify this to just:
> 
> ```
> } elseif ($mode == "incomplete") {
>     amcInComplete($amc_id, $patient_id, $object_category, $object_id);
> } elseif ($mode == "incomplete_safe") {
>     amcInCompleteSafe($amc_id, $patient_id, $object_category, $object_id, $date_created);
> }
> ```
> 
> That would also let you remove `uncomplete` from `.codespell-ignore-words.txt` (line 120), so the spellchecker catches any future uses of the non-word.
> 
> What do you think?

I totally agree with you. I will make an update right now.

PR #10703, opened 2026-02-11
fix(phpstan): Fail if baseline doesn't match generator

Firehed:
If the PHPStan files are manually edited (where "manual" means _literally anything other than `composer phpstan-baseline`_), minor changes can crop up down the line which lead to weird PRs and merge conflicts. You can see an example triggered by https://github.com/openemr/openemr/pull/10698; this PR includes a regenerated baseline (6c64ef3d27490d3544dd7797d9460cb0a0d72897) which fixes it (not to call anyone out here! it's happened a few times, and it's _very easy_ to do with broad find-and-replace tooling)

This adds a step after analysis to regenerate the baseline and fail if the files are _in any way_ changed. This most commonly manifests as a change in sort order. Note this specific failure:

https://github.com/openemr/openemr/actions/runs/21925386064/job/63316450704?pr=10703

No

### DIFF
--- a/.phpstan/baseline/missingType.parameter.php
+++ b/.phpstan/baseline/missingType.parameter.php
@@ -17682,6 +17682,51 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../library/amc.php',
 ];
 $ignoreErrors[] = [
+    'message' => '#^Function amcInCompleteSafe\\(\\) has parameter \\$amc_id with no type specified\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/amc.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Function amcInCompleteSafe\\(\\) has parameter \\$date_created with no type specified\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/amc.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Function amcInCompleteSafe\\(\\) has parameter \\$object_category with no type specified\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/amc.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Function amcInCompleteSafe\\(\\) has parameter \\$object_id with no type specified\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/amc.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Function amcInCompleteSafe\\(\\) has parameter \\$patient_id with no type specified\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/amc.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Function amcInComplete\\(\\) has parameter \\$amc_id with no type specified\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/amc.php',
+];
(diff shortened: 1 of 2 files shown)


.github/workflows/phpstan.yml
```diff
@@ -47,6 +47,12 @@ jobs:
       run: vendor/bin/phpstan --memory-limit=8G diagnose
     - name: PHPStan Analyze
       run: vendor/bin/phpstan --memory-limit=8G analyze --error-format=github
+    # Changes to baseline through anything other than the native generator tend
+    # to lead to merge conflicts and weird diffs, typically from sort order.
+    - name: Ensure baseline is stable
+      run: |
+        composer phpstan-baseline
+        git diff --exit-code .phpstan/baseline/
```
kojiromike: Would it help contributors to echo guidance before the diff check? When this fails, they'll see the diff but might not immediately know what action to take.

For example, adding an echo before the commands:
```yaml
run: |
  echo "If this step fails, run 'composer phpstan-baseline' locally and commit the results."
  composer phpstan-baseline
  git diff --exit-code .phpstan/baseline/
```

Not critical - just thinking about the contributor experience when they hit this for the first time.

kojiromike: Since the baseline artifact is already uploaded on failure, could we tell contributors how to fetch it?

```yaml
      run: |
        composer phpstan-baseline
        if ! git diff --exit-code .phpstan/baseline/; then
          echo ""
          echo "Baseline files are out of sync. Run this to fix:"
          printf '  gh run download %s --name phpstan-baseline-php8.4 -D .phpstan/baseline/\n' "$GITHUB_RUN_ID"
          exit 1
        fi
```

Tested locally - the artifact extracts directly to the target path, so contributors just run that one command and commit.

kojiromike, line 55: Could we add a workflow summary to guide contributors? This would appear prominently on the Actions run page:

```yaml
- name: Ensure baseline is stable
  run: |
    composer phpstan-baseline
    if ! git diff --exit-code .phpstan/baseline/; then
      {
        echo '## ❌ PHPStan Baseline Out of Sync'
        echo
        echo "The baseline files don't match what the generator produces."
        echo
        echo '### Recommended: Run PHPStan locally'
        echo
        echo 'Running PHPStan as part of your workflow catches type errors early:'
        echo
        echo '| Method | Command |'
        echo '|--------|---------|'
        # shellcheck disable=SC2016
        echo '| Pre-commit hook | `prek install` or `pre-commit install` (runs automatically on commit) |'
        # shellcheck disable=SC2016
        echo '| Composer | `composer phpstan` |'
        # shellcheck disable=SC2016
        echo '| Docker | `openemr-cmd pst` |'
        echo
        echo '### Quick fix'
        echo
        echo "If you're confident your PR doesn't introduce new errors, you may apply the baseline from this workflow:"
        echo
        echo '```bash'
        echo "gh run download $GITHUB_RUN_ID --name phpstan-baseline-php8.4 -D .phpstan/baseline/"
        echo '```'
        echo
        echo 'Then commit and push the changes.'
      } >> "$GITHUB_STEP_SUMMARY"
      exit 1
    fi
```

Firehed: Can we do this as a fast-follow? I like the gist of it but this change is blocking/merge-conflicting a bunch of PRs and I want to consider this in a bit more detail before adding it into the workflow.

Firehed: #10721

PR #10704, opened 2026-02-11
feat(db): Support Doctrine migrations for new schema migrations

### LINKED ISSUE #10706: Database: add and configure doctrine/migrations library

### LINKED ISSUE #10707: Database: add migrations tooling into CI

### LINKED ISSUE #10769: CLI: Define standard way to specify site
Today, there does not appear to be a well-defined way to specify the site id (and the downstream settings derived from it) on the command line - the current implementation is tied to a combination of session data and HTTP request components. This means any sort of CLI tooling can only work on the `default` site and, by extension, database.

Firehed:
Fixes #10706, fixes #10707, fixes #10769.

Adds support for [Doctrine Migrations](https://www.doctrine-project.org/projects/doctrine-migrations/en/3.9/reference/introduction.html#introduction).

This supports vendor-agnostic[^ish] DB schema changes, and running them _bidirectionally_. The bidirectional nature will ease module management significantly, especially for uninstallation.

Further, this tooling should greatly simplify end-to-end testing in and outside of CI, since incremental database changes can be more easily applied. One can trivially preview the DB migration path when upgrading their installation(s) of OpenEMR.

- Adds library
- Adds config so the library works
- Adds README with usage info
- Adds placeholder starter migration so that the tooling can generate and use the `migrations` table
- Adds a CI workflow that executes all migrations in a new database to ensure they're valid

No

[^ish]: within the context of different RDBMSs having slightly different features and requirements. It's not 100%, but it's largely not tied to a single platform.

### DIFF
--- a/db/migration-config.php
+++ b/db/migration-config.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 /**
  * Doctrine Migrations config file
  *
+ * This file is read by config/cli-config.php.
+ *
  * @link https://www.doctrine-project.org/projects/doctrine-migrations/en/3.9/reference/configuration.html#configuration
  *
  * @package   openemr


bradymiller:
hi @Firehed , Look fine to me since not impacting anything at this time. Is the plan to have upgrades go through this or is this just for modules?

Firehed:
Core and modules eventually. The main issue with about a dozen sub-issues has more details on the complete plan. 

Plus of course moving a lot of core _into_ modules eventually, but that's more changing how things are organized rather than how they run since it will go through the same system.

adunsulag:
This looks fine to me.  I'd like to see another PR to bring in the symfony bridge layer so we could have a single tool (php bin/console) to run the migration commands instead of another tool in the vendor directory https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html

But that's not necessary for this work.

Firehed:
Yep @adunsulag - fully agree! See #10712.

PR #10720, opened 2026-02-12
feat(phpstan): add generator for database table type aliases

### LINKED ISSUE #10718: feat(phpstan): Add generated type aliases for database table rows
## Problem

Over 10,000 PHPStan errors stem from database queries returning `array<mixed>`. Code like `$patient['fname']` can't be validated because PHPStan doesn't know what keys exist in the result arrays.

Manual entity classes would solve this but create a maintenance burden—the `patient_data` table alone has 132 columns that change over time.

## Solution

Create a code generator that reads table schemas from `database.sql` and outputs PHPStan type aliases:

```php
namespace OpenEMR\Common\Database;

/**
 * @phpstan-type PatientDataRow array{
 *   id: int,
 *   pid: int,
 *   fname: string,
 *   lname: string,
 *   DOB: ?string,
 *   ...
 * }
 */
class TableTypes {}
```

Services can then import and use these types:

```php
use OpenEMR\Common\Database\TableTypes;

/**
 * @phpstan-import-type PatientDataRow from TableTypes
 */
class PatientService
{
    /** @return PatientDataRow|false */
    public function findByPid(int $pid): array|false
```

## Prototype

A working prototype exists at https://github.com/kojiromike/openemr/tree/phpstan-type-generator

- `bin/generate-phpstan-types.php` - 268-line generator script
- Parses CREATE TABLE statements from `database.sql`
- Maps SQL types to PHP types (int, string, ?string, etc.)
- Generates valid PHP that PHPStan can analyze

Tested: PHPStan correctly validates array key access and field types.

## Scope (MVP)

1. Add `bin/generate-phpstan-types.php` generator script
2. Add `composer generate-phpstan-types` script
3. Generate `src/Common/Database/TableTypes.php` for high-impact tables:
   - `patient_data` (132 columns)
   - `users` (66 columns)
   - `form_encounter` (35 columns)
   - `insurance_data`
   - `billing`
4. Update `PatientService` to use `PatientDataRow` type
5. Add to CI: verify generated types match schema (detect drift)

## Future Work

- Generate types for all 281 tables
- Add composer script to regenerate on schema changes
- Update more services to use typed returns

## Impact

- Estimated 2,000-3,000 baseline errors resolved for these core tables
- Pattern established for incrementally covering more tables
- No runtime changes, no breaking changes to callers

---

**AI Disclosure:** Yes

kojiromike:
Fixes #10718

## Summary

Add infrastructure for generating PHPStan `@phpstan-type` aliases from `database.sql` CREATE TABLE statements, enabling static analysis of array key access on database query results.

**Components:**
- `bin/generate-phpstan-types.php` - CLI entry point
- `src/Common/Command/GeneratePhpstanTypesCommand.php` - Symfony Console command
- `src/Common/Database/TableTypes.php` - Generated type aliases interface
- `.github/workflows/phpstan-types.yml` - CI job to keep types in sync

## Usage

```bash
# Generate types for specific tables
php bin/generate-phpstan-types.php sql/database.sql patient_data users

# List all tables
php bin/generate-phpstan-types.php sql/database.sql --list

# Generate types for all tables
php bin/generate-phpstan-types.php sql/database.sql --all
```

## Type Application Pattern

1. Import type from `TableTypes` interface
2. Add `@return` annotations to service methods (e.g., `UserService`, `PatientService`)
3. Narrow types at call sites before property access

Type narrowing examples throw `RuntimeException` when user lookups fail unexpectedly, preserving fail-loudly behavior while satisfying PHPStan.

## Test Plan

- [x] Generator parses `patient_data` (132 cols), `users` (66 cols), `form_encounter`, `insurance_data`, `billing`
- [x] PHPStan recognizes imported types and validates array key access
- [x] CI workflow detects when `TableTypes.php` is out of sync with `database.sql`
- [x] Pre-commit hooks pass at PHPStan level 10

---

**AI Disclosure:** Yes


Firehed:
this will be a huge win for future doctrine support to generate entities, too! (although I think it may have a built-in schema importer)

bin/generate-phpstan-types.php
```diff
@@ -0,0 +1,36 @@
...
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+require_once __DIR__ . '/../vendor/autoload.php';
+
+// Check for required dev dependency
+if (!class_exists(\iamcal\SQLParser::class)) {
```
kojiromike: is there a more symfony-friendly way to write this?

custom/chart_tracker.php
```diff
@@ -116,7 +116,9 @@ function userSelect() {
     $current_location = xlt('Unassigned');
     if ($ct_userid) {
         $user = $userService->getUser($ct_userid);
-        $current_location = text($user['lname'] . ", " . $user['fname'] . " " . $user['mname'] . " " . oeFormatDateTime($row['ct_when'], "global", true));
+        if ($user !== false) {
```
kojiromike: what would have happened in the previous implementation, had $user been false?

interface/forms/newpatient/report.php
```diff
@@ -38,12 +38,12 @@ function newpatient_report($pid, $encounter, $cols, $id): void
...
+        $provider = (!$hasAccess || $rawProvider === false) ? false : $rawProvider['fname'] .
             (($rawProvider['mname'] ?? '') ? " " . $rawProvider['mname'] . " " : " ") .
             $rawProvider['lname'] .
             ($rawProvider['suffix'] ? ", " . $rawProvider['suffix'] : '') .
             ($rawProvider['valedictory'] ? ", " . $rawProvider['valedictory'] : '');
-        $referringProvider = (!$hasAccess || !$rawRefProvider) ? false : $rawRefProvider['fname'] . " " . $rawRefProvider['lname'];
+        $referringProvider = (!$hasAccess || $rawRefProvider === false) ? false : $rawRefProvider['fname'] . " " . $rawRefProvider['lname'];
         $posCode = (!$hasAccess) ? false : sprintf('%02d', trim($result['pos_code'] ?? false));
         $posCode = ($posCode && $posCode != '00') ? $posCode : false;
         $facility_name = (!$hasAccess) ? false : $result['facility_name'];
```
kojiromike: rewrite this with an `if/else` around $hasAccess.

src/Common/Command/GeneratePhpstanTypesCommand.php
```diff
@@ -0,0 +1,236 @@
...
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Common\Command;
+
+use iamcal\SQLParser;
```
kojiromike: Add a guard around this.

bin/generate-phpstan-types.php
kojiromike: Moved the check into the command's `execute()` method and simplified the entry point.

custom/chart_tracker.php
kojiromike: Previously it would error (PHP 8: accessing `false['lname']` throws TypeError). Added a comment explaining the behavior change: deleted users now fall back to "Unassigned" instead of showing broken output.

interface/forms/newpatient/report.php
kojiromike: Rewritten with explicit if/else.

src/Common/Command/GeneratePhpstanTypesCommand.php
kojiromike: Added `class_exists` check at the start of `execute()` using a string literal to avoid triggering autoload before the check. Added inline comment on the `use` statement pointing to the guard.

src/Common/Command/GeneratePhpstanTypesCommand.php
```diff
@@ -0,0 +1,246 @@
...
+        $output .= " * Type definitions for database row arrays\n";
+        $output .= " *\n";
+        $output .= " * Import this interface and reference types with @see annotations:\n";
+        $output .= " *   @return TableTypes::PatientDataRow|false\n";
+        $output .= " *\n";
+        $output .= implode("\n *\n", $typeAliases);
+        $output .= "\n */\ninterface TableTypes\n{\n";
+        $output .= "    // This interface exists only to hold PHPStan type definitions.\n";
+        $output .= "}\n";
+
```
kojiromike: use a heredoc for this

custom/chart_tracker.php
kojiromike: Updated: falling back to "Unassigned" is acceptable here - the original behavior was a crash (PHP 8 TypeError when accessing `false['lname']`). Showing the deleted user's ID would require expanding the PHPStan baseline, which defeats the purpose. Added a comment explaining the behavior change.

src/Common/Command/GeneratePhpstanTypesCommand.php
kojiromike: Done - switched to flexible heredoc (PHP 7.3+).

custom/chart_tracker.php
kojiromike: Updated: now throws `RuntimeException` if the user lookup fails unexpectedly. Preserves the original fail-loudly behavior but makes it intentional.

bin/generate-phpstan-types.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * CLI entry point for generating PHPStan type aliases
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
```
bradymiller, line 12: @kojiromike , Ok for this script to be public and run by anybody?
(note above not time sensitive and can be answered next week when online)

kojiromike: @bradymiller yes, definitely. It should be run by anyone who changes database.sql in a way that changes the types.

bradymiller: @kojiromike , sorry, meant to ask if ok to have this script open on a production server (appears it can be run directly via web browser without any guardrail or auth)
(again, not time sensitive and can be answered next week)

Firehed: The CLI application path shouldn't do anything if it's requested through the web, but also the entire bin directory shouldn't be reachable. Generally a part of the front controller stuff.

Issue #10666, opened 2026-02-09
bug(phpstan): phpstan-baseline easy dev local run does not match github CI


stephenwaite:
### Describe the bug

https://github.com/openemr/openemr/pull/10661#issuecomment-3873934914

```openemr-cmd pst
PHPStan diagnose
Note: Using configuration file /var/www/localhost/htdocs/openemr/phpstan.neon.dist.
PHP runtime version: 8.4.16
PHP composer.json required version: 8.2-8.5.99
PHP version for analysis: 8.2 (from config.platform.php in composer.json)

PHPStan version: 2.1.38
PHPStan running from:
/var/www/localhost/htdocs/openemr/vendor/phpstan/phpstan

Extension installer:
composer/composer: 2.9.3
composer/pcre: 3.3.2
phpstan/phpstan-phpunit: 2.0.12
shipmonk/phpstan-baseline-per-identifier: 2.3.0

Discovered Composer project root:
/var/www/localhost/htdocs/openemr

Note: Using configuration file /var/www/localhost/htdocs/openemr/phpstan.neon.dist.
 4014/4034 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓░]  99%

 4034/4034 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ----------------------------------------------------------------------- 
  Line   interface/batchcom/batch_phone_notification.php                        
 ------ ----------------------------------------------------------------------- 
         Ignored error pattern #^Parameter \#1 \$type of function               
         cron_InsertNotificationLogEntry expects string, mixed given\.$# in     
         path                                                                   
         /var/www/localhost/htdocs/openemr/interface/batchcom/batch_phone_noti  
         fication.php was not matched in reported errors.                       
         Ignored error pattern #^Parameter \#2 \$prow of function               
         cron_InsertNotificationLogEntry expects array, mixed given\.$# in      
         path                                                                   
         /var/www/localhost/htdocs/openemr/interface/batchcom/batch_phone_noti  
         fication.php was not matched in reported errors.                       
         Ignored error pattern #^Parameter \#3 \$db_sms_msg of function         
         cron_InsertNotificationLogEntry expects array, mixed given\.$# in      
         path                                                                   
         /var/www/localhost/htdocs/openemr/interface/batchcom/batch_phone_noti  
[...]

PR #10729, opened 2026-02-12
refactor: Deduplicate more global functions

Firehed:
Towards #10241.

Duplicate function names in the global namespace which can throw off static analysis (see #10666/#10713) and could lead to a fatal error if the include path goes awry.

Renames a bunch of functions detected as duplicates by adding a localized prefix and updating call sites. There's relatively little logic changes

Yes, Claude.

### DIFF
--- a/.phpstan/baseline/assign.propertyType.php
+++ b/.phpstan/baseline/assign.propertyType.php
@@ -1937,31 +1937,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../src/FHIR/SMART/SMARTLaunchToken.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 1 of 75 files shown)


interface/forms/procedure_order/common.php
```diff
@@ -342,8 +342,13 @@ function normalizeDirectoryName(string $input): string
                 }
 
                 try {
-                    // Generate the HL7 order
-                    $result = gen_hl7_order($formid);
+                    // Generate the HL7 order using lab-specific function
+                    $result = match ($gbl_lab) {
+                        'labcorp' => labcorp_gen_hl7_order($formid),
```
kojiromike, line 347: mumble mumble strategy pattern

Firehed:
Ohhh yes, this whole thing needs a proper refactor. This is very much a "stop the bleeding" fix.

PR #10752, opened 2026-02-12
fix(claimrev): throw exceptions on API failures instead of returning error values

### LINKED ISSUE #10737: ClaimRev: Null pointer errors when API calls fail
## Summary

Multiple methods in `ClaimRevApi.php` access object properties without null checks. When `curl_exec()` fails or `json_decode()` returns null, these cause fatal errors in PHP 8+.

### ClaimRevApi.php:54-61
```php
$result = curl_exec($ch);
$data = json_decode($result);
if (property_exists($data, 'access_token')) {  // $data could be null!
```

`property_exists(null, ...)` throws TypeError in PHP 8+.

### ClaimRevApi.php:99
```php
if ($data->isError) {  // No null check on $data
```

### EligibilityTransfer.php:86
```php
if ($result->isFatalError) {  // $result could be boolean false
```

## Impact

- API failures cause fatal PHP errors instead of graceful handling
- No error logging when curl fails
- Difficult to debug connectivity issues

## Suggested Fix

Add null/false checks before accessing properties and log curl errors.

kojiromike:
## Summary
Fixes #10737

Add proper error handling for `curl_exec()` and `json_decode()` failures that would otherwise cause fatal PHP 8+ errors when accessing properties on `false`/`null` values.

- Add explicit TLS verification (`CURLOPT_SSL_VERIFYPEER`, `CURLOPT_SSL_VERIFYHOST`) to all curl calls
- Check `curl_exec()` result before `json_decode()` and log errors via SystemLogger
- Check `json_decode()` result before `property_exists()` calls
- Add `property_exists()` check for `isError` in `uploadClaimFile()`
- Add `property_exists()` check for `isFatalError` in `saveEligibility()`
- Update PHPStan baseline for new error handling code

## AI Disclosure
Yes

### DIFF
--- a/interface/modules/custom_modules/oe-module-claimrev-connect/src/ClaimRevApi.php
+++ b/interface/modules/custom_modules/oe-module-claimrev-connect/src/ClaimRevApi.php
@@ -1,32 +1,40 @@
 <?php
 
 /**
+ * ClaimRev API client
  *
- * @package OpenEMR
- * @link    http://www.open-emr.org
- *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
  * @author    Brad Sharp <brad.sharp@claimrev.com>
+ * @author    Michael A. Smith <michael@opencoreemr.com>
  * @copyright Copyright (c) 2022 Brad Sharp <brad.sharp@claimrev.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
 namespace OpenEMR\Modules\ClaimRevConnector;
 
 use OpenEMR\Common\Http\HttpRestRequest;
+use OpenEMR\Common\Logging\SystemLogger;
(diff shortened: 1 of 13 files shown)


interface/modules/custom_modules/oe-module-claimrev-connect/src/ClaimRevApi.php
```diff
@@ -122,20 +147,34 @@ public static function getReportFiles($reportType, $token)
...
+        curl_setopt($ch, CURLOPT_SSL_VERIFYHOST, 2);
         curl_setopt($ch, CURLOPT_URL, $url);
         curl_setopt($ch, CURLOPT_HTTPHEADER, $headers);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+
         $result = curl_exec($ch);
+        if ($result === false) {
+            (new SystemLogger())->error('ClaimRev: cURL error in ' . __FUNCTION__, ['error' => curl_error($ch)]);
+            curl_close($ch);
+            return '';
```
kojiromike: If we return '' here, does that mean the http server returns a 200? Shouldn't we return a 500 here?

PR #10754, opened 2026-02-12
fix(claimrev): prevent HTML escaping of ERA file downloads

### LINKED ISSUE #10738: ClaimRev: ERA file download corrupted by HTML escaping
## Summary

The ERA file download in `EraDownload.php` uses `text()` to escape file content, which corrupts EDI files by HTML-encoding special characters.

## Location

`interface/modules/custom_modules/oe-module-claimrev-connect/public/EraDownload.php` line 35

## Current Code

```php
header("Content-Type: application/force-download");
// ...
echo text($fileViewModel->fileText);
```

## Problem

The `text()` function converts `&` to `&amp;`, `<` to `&lt;`, etc. EDI X12 files (835 ERA files) use these characters as delimiters. Escaping them corrupts the file structure, making it unparseable by billing systems.

## Expected Behavior

For file downloads with `Content-Type: application/force-download`, the content should be output raw:

```php
echo $fileViewModel->fileText;
```

The data comes from the ClaimRev API response which is trusted.

kojiromike:
Fixes #10738

## Description

ERA file downloads from the ClaimRev module were being corrupted because `text()` was HTML-encoding special characters (`&` → `&amp;`, `<` → `&lt;`, etc.). EDI X12 files use these characters as delimiters, so the encoding broke file parsing.

## Changes

- Remove `text()` wrapper from file output (the actual fix)
- Change Content-Type to `application/edi-x12` (IANA-registered MIME type)
- Quote filename in Content-Disposition header (RFC 6266 compliant)

## Why it's safe

The `fileText` is only used for download, never rendered in HTML. The endpoint sets `Content-Disposition: attachment`, forcing browser download rather than rendering. This matches other download endpoints in OpenEMR (e.g., `sl_eob_search.php`).

## Testing

1. In OpenEMR with ClaimRev module enabled, navigate to ERA processing
2. Download an ERA file
3. Verify the downloaded file contains raw EDI content (delimiters like `~`, `*`, `&` should not be HTML-encoded)

## AI Disclosure

Yes

### DIFF
--- a/.phpstan/baseline/cast.string.php
+++ b/.phpstan/baseline/cast.string.php
@@ -748,7 +748,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Cannot cast mixed to string\\.$#',
-    'count' => 1,
+    'count' => 2,
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-claimrev-connect/public/EraDownload.php',
 ];
 $ignoreErrors[] = [
--- a/interface/modules/custom_modules/oe-module-claimrev-connect/public/EraDownload.php
+++ b/interface/modules/custom_modules/oe-module-claimrev-connect/public/EraDownload.php
@@ -1,12 +1,15 @@
 <?php
 
 /**
+ * ERA file download handler
  *
- * @package OpenEMR
- * @link    http://www.open-emr.org
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
  *
  * @author    Brad Sharp <brad.sharp@claimrev.com>
+ * @author    Michael A. Smith <michael@opencoreemr.com>
  * @copyright Copyright (c) 2022 Brad Sharp <brad.sharp@claimrev.com>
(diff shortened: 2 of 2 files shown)


kojiromike:
## Note on Semgrep OSS check failure

The "Semgrep OSS" check (GitHub Advanced Security integration) is flagging line 38 as an XSS risk. This is a **false positive** because:

1. **This is a file download endpoint, not HTML output** - `Content-Disposition: attachment` forces the browser to download the file rather than render it
2. **Content-Type is `application/edi-x12`** (IANA-registered MIME type for X12 EDI files), not `text/html`
3. **The data comes from the ClaimRev API**, not directly from user input - the `$_GET['eraId']` is just an identifier used to fetch the file

The local "Semgrep Security Scan" workflow (which uses the repo's `semgrep.yaml` config) passes because the `// nosemgrep: echoed-request` comment is respected. However, the GHAS "Semgrep OSS" check doesn't honor inline suppression comments.

A maintainer can dismiss this alert through the [Security tab](https://github.com/openemr/openemr/security/code-scanning?query=pr%3A10754) after review.

PR #10761, opened 2026-02-13
refactor(gacl): Narrow scope and apply correct type to Gacl properties

Firehed:
This changes a lot of `Gacl` properties to have proper variable scoping (i.e. not relying on the very old convention of `_` prefixes for private) and types. It simplifies some upcoming DB work to help verify what properties are used outside of the Gacl tooling (unsurprisingly, convention-based scoping means there are some inappropriate reads). It also knocks out a decent number of baselined errors.

The "new" baselined errors are those revealed due to the changes, not due to logic changes. Pretty common stuff like not checking whether or not the query succeeded before trying to read the result.

Done while keeping PHPStan green:

- Most properties changed to private
- Some left as protected since they're used by a subclass
- One left public since it's used (inappropriuately) in what looks like some UI code

Further cleanup reducing baseline errors:
- Sets a couple native types where it's unambiguous
- Provides more specific type info (object -> ADOConnection made a big difference)
- Fixed some mixed-casing method calls

No


bradymiller:
nice!

PR #10762, opened 2026-02-13
fix(db): rename misspelled 'declne_to_specfy' value in race

phantom930:
Fixes #10385 

# Summary
The race list option contains a misspelled value `declne_to_specfy` that should be `decline_to_specify`.

# Changes proposed in this pull request
- Added SQL migration to rename the value in `list_options`
- Updated any `patient_data` records that reference this value
- Removed the hardcoded workaround in `FhirPatientService.php`
- Ensured backwards compatibility for existing patient records.

PR #10766, opened 2026-02-13
chore: bump v_database

stephenwaite:


kojiromike:
In case we need reference later, this should have been done in #10762.

Issue #10774, opened 2026-02-14
Database: encapsulate setup of DB connections


Firehed:
These are done in about a half-dozen places in the codebase. The logic is _relatively_ consistent across them, but it's still dupliated and not without some variance.

This issue is only about centralizing and unifying the logic for _existing_ DB connections; once that's in place, updating them to return DBAL (#10149) becomes a more straightforward task of "change the type and chase the PHPStan errors"

ADODB:
- [x] library/sql.inc.php (#10777) 
- [x] interface/main/calendar/includes/pnAPI.php (#10773)
- [x] GACL (#10778)

mysqli:
- [x] portal/patient/fwk/libs/verysimple/DB/DataDriver/MySQLi.php (#10802)
- [x] library/classes/Installer.class.php (#10805)

Laminas:
- [x] interface/modules/zend_modules/config/autoload/global.php

mysql CLI (!):
- [ ] interface/main/backup.php

Issue #10549, opened 2026-02-03
chore(db): Remove legacy support for non-utf8 installations


Firehed:
There's a `$disable_utf8_flag` variable in the DB connection handling that crops up in a few places. Perhaps counterintuitively, it puts MySQL _into_ `utf8` mode - but that's because MySQL's `utf8` (a.k.a. `utf8mb3`) mode is both deprecated and broken, and `utf8mb4` is what you want (which is what you get when the flag is off, which seems to be the default from the installer).

See https://dev.mysql.com/doc/refman/8.4/en/charset-unicode-utf8.html

I vote we scrap it entirely unless there's a strong need to keep it around. Any installation relying on it is skating on thin ice; MySQL's documentation states:

> Expect utf8mb3 to be removed in a future major release of MySQL.

Perhaps this is a two-part thing where we detect and flag it so admins can see in let's say 8.0.1 and remove it in a version following that, but documentation indicates this has been the default since OpenEMR 6 so I think the direct approach should be fine.

I'll take this on unless someone strongly feels we shouldn't, since it ties in to other DB improvement work I'm handling.

PR #10776, opened 2026-02-14
feat(db): Prepare central ADODB setup

Firehed:
Towards #10774

Once integrated, this will eliminate a bunch of repeated database setup code. That will unlock moving the connection to DBAL in a _far_ more direct way.

Adds a new DB factory method that does all of the ADODB setup that's repeated in several places.

No

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,22 @@
+self-hosted-runner:
+  # Labels of self-hosted runner in array of strings.
+  labels: []
+
+# Configuration variables in array of strings defined in your repository or
+# organization. `null` means disabling configuration variables check.
+# Empty array means no configuration variable is allowed.
+config-variables:
+
+# Configuration for file paths. The keys are glob patterns to match to file
+# paths relative to the repository root. The values are the configurations for
+# the file paths. Note that the path separator is always '/'.
+# The following configurations are available.
+#
+# "ignore" is an array of regular expression patterns. Matched error messages
+# are ignored. This is similar to the "-ignore" command line option.
+paths:
+  .github/workflows/database-version.yml:
+    # SC2016 warns about $ and backticks in single quotes not expanding.
+    # We intentionally use single quotes for literal markdown output.
+    ignore:
+    - SC2016
--- a/.github/workflows/database-version.yml
+++ b/.github/workflows/database-version.yml
@@ -1,0 +1,102 @@
+# Ensures v_database is aligned between version.php and sql/database.sql
+#
+# The v_database value must be:
+# 1. Declared in version.php as: $v_database = NNN;
+# 2. Declared in sql/database.sql header as: -- v_database: NNN
+# 3. Both values must match
+#
+# This prevents database migrations from being silently skipped on upgrades.
+name: Database Version Check
+
+on:
+  pull_request:
+    branches:
+    - master
+    - rel-*
+    paths:
+    - version.php
+    - sql/database.sql
+    - sql/patch.sql
+    - sql/*-to-*_upgrade.sql
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  check-v-database:
+    name: Check v_database alignment
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v6
+
+    - name: Setup PHP
+      uses: ./.github/actions/setup-php-composer
+      with:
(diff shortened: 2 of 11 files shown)


Firehed:
See #10777 and #10778 for how this works in practice.

src/BC/DatabaseConnectionFactory.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        // Configure the charset. This doesn't precisely match the previous
+        // behavior, but should handle any current installation and not disrupt
+        // upgrades.
+        $charset = strtolower($config->charset);
+        if ($charset !== 'utf8mb4') {
+            // You're running something pretty weird and probably broken. Emit
+            // a clear warning.
+            if ($charset !== 'utf8') {
```
bradymiller: hi @Firehed , Note sites/default/sqlconf.php has the $disable_utf8_flag flag in it, which is set to false. This allows legacy users on dreaded latin (i forgot the full name :) ) encoding to still work (they add that variable with a true setting to override and not use utf8*). Can see it here:
https://github.com/openemr/openemr/blob/master/library/sql.inc.php#L104

(the override basically means that a SET NAMES command is not sent)

(and basically ignores $db_encoding setting)

Firehed: Thanks for calling this out. See #10549 for more info here. I'm not fully confident that settings works as expected across the board; the installer in particular seems to support only utf8mb4 (which is ideal).

If that's not the case in the wild, I thing we ought to prioritize getting installations updated. Using pretty much anything else these days tends to lead to bugs and data loss. All supported versions of MySQL have supported this encoding for a very long time.

bradymiller: hi @Firehed ,

They are only pertinent for upgraders and new installations basically forced to use utf8mb4.

Back story is openemr's current upgrade process should work since version 3.1 (year 2009) (prior to that required special steps to bring in phpgacl, so can go back even many years further with that) and that is back when was using latin1 encoding (note even until several years ago, this was even happening with diy users that were creating their own database). And upgrading from latin1 to utf8/utf8mb4 can be a bear. So, made it flexible for users that are stuck on latin1 encoding. The sqlconf.php file generally is persistent during upgrades, so that is why have those settings like they are in there. If a upgrade from a legacy openemr database wants to use latin1 then they can add that override variable. All those settings are simply doing is dealing with SET NAMES stuff. So, if possible, would maintain the same logic for them for now as is at:
https://github.com/openemr/openemr/blob/master/library/sql.inc.php#L104-L116

Regarding dropping utf8, again, wouldn't do that at this point. If mysql really goes through with it (sounds like mariabd has no plans to do this btw), then can deal with it at that time. Would be better to get a break anyways (guessing mysql will break on the set names utf8) than just  silently do utf8mb4 instead.

adunsulag: I know this was more a concern 5 years ago, but at this point if people haven't migrated from latin1? Will they ever?

Firehed: I'm investigating what it might look like to safely migrate all installations. I believe this can be done the same as any schema migration. Even if upgrading from latin1 was once "a bear", from what I've seen this isn't really the case in the versions of mysql OpenEMR supports. But I'm researching further, so I may eat my words here :)

IMO it's unsafe for user data to have inconsistent table encodings across installations, especially in the context of modules. Most code is written in a way that will not correctly handle more than one storage encoding, and since the default for new installs is utf8 (i.e. mysql utf8mb4) I'd wager most code is written for that.

FWIW, the `SET NAMES` and the actual table encodings are loosely coupled at most. It influence the over-the-wire encoding but isn't inherently incompatible with older table encodings. Everything I've found so far STRONGLY indicates that `set names = 'utf8mb4'` will not cause problems if tables are latin1 or something else. MySQL will convert the wire format in and out of the storage format.

Firehed: Based on the admin call discussion and the work in #10789 I'm going to consider this resolved. Landing that change will require me to update this regardless, but it should be simpler.

bradymiller: agreed :)

src/BC/DatabaseConnectionFactory.php
```diff
@@ -0,0 +1,76 @@
...
+        } else {
+            $conn->Connect(
+                argHostname: $config->host,
+                argUsername: $config->user,
+                argPassword: $config->password,
+                argDatabaseName: $config->dbname,
+            );
+        }
+
+        $conn->ExecuteNoLog("SET NAMES '$config->charset'");
```
bradymiller, line 58: hi @Firehed ,
This variable no longer exists. Should be hardcoding the SET NAME.

PR #10777, opened 2026-02-14
feat(db): Migrate ADODB setup in sql.inc to new factory method

Firehed:
Towards #10774.

Removes a bunch of now-redundant inline setup code.

Nearly all ADODB setup deleted from sql.inc.php (scary, I know!) in favor of the new factory added in #10776.

This should have zero runtime impact, and not require any changes by people running the software.

No

### DIFF
--- a/.phpstan/baseline/empty.notAllowed.php
+++ b/.phpstan/baseline/empty.notAllowed.php
@@ -3263,7 +3263,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#',
-    'count' => 4,
+    'count' => 3,
     'path' => __DIR__ . '/../../library/sql.inc.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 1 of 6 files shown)


library/sql.inc.php
```diff
@@ -55,38 +59,12 @@
...
-                $database->ssl_ca = "{$GLOBALS['OE_SITE_DIR']}/documents/certificates/mysql-ca";
-            } else {
-                // without client side certificate/key
-                $database->ssl_ca = "{$GLOBALS['OE_SITE_DIR']}/documents/certificates/mysql-ca";
-            }
-            $database->clientFlags = MYSQLI_CLIENT_SSL;
-        }
-    }
-    $database->port = $port;
+    $config = DatabaseConnectionOptions::forSite($GLOBALS['OE_SITE_DIR']);
```
Firehed, line 58: This runs too early to use GlobalsBag reliably.

kojiromike: I think this should be a code comment rather than a PR comment. 🙏

kojiromike: I think this should be a code comment.

PR #10800, opened 2026-02-18
refactor(tests): Prototype session sharing for HhMainMenuLinksTest

Firehed:
Refactor HhMainMenuLinksTest to use a shared browser session across all test cases instead of creating a fresh session for each test.

Key changes:
- Use setUpBeforeClass() to create client and login once
- Store client in static property shared across all tests
- Remove per-test login/quit cycle
- Inline the navigation methods to use shared client

This should significantly reduce test execution time by eliminating 59 redundant login operations (one per menu link test case).

Trade-offs:
- Less test isolation (state accumulates between tests)
- Cascading failures possible if one test leaves bad state
- Login flow exercised once instead of 60 times

Fixes #

### DIFF
--- a/tests/Tests/E2e/HhMainMenuLinksTest.php
+++ b/tests/Tests/E2e/HhMainMenuLinksTest.php
@@ -16,20 +16,116 @@ declare(strict_types=1);
 
 namespace OpenEMR\Tests\E2e;
 
-use OpenEMR\Tests\E2e\Base\BaseTrait;
+use Facebook\WebDriver\Exception\TimeoutException;
+use Facebook\WebDriver\Remote\DesiredCapabilities;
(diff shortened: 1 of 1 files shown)


kojiromike:
There's a long history of `fix(e2e)` commits that I have never gotten to the bottom of. Also, if you're changing e2e tests, please use the `Test-Mode: full` commit trailer and suffer through the extra tests to make sure they don't fall over. Time and time again the e2e tests pass in PRs and fail in the nightlies.

Firehed:
Noted. This doesn't seem to be having an appreciable effect regardless so I don't really expect it to go anywhere, but I'll keep that in mind!

Firehed:
this is just hanging the e2e tests now, and I don't want a five-minute AI experiment to turn into a huge project.

Issue #8890, opened 2025-09-16
Disabling connection pooling should not require a database connection first


kojiromike:
To disable database connection pooling in openemr, you must set a global. Setting a global means setting a value in the database. So to disable connection pooling, you must first make a connection to the database. This can lead to circular problems where the user is unable to disable the thing preventing them from making a healthy database connection.

Even with the webUI setup, sqlconf.php is written to the filesystem. It must be done before any database connection can be made. Should connection pooling be configured in that file as well, instead of being configured in a global?

bradymiller:
Hi @kojiromike ,
Would migrate the setting to simple global at top of sites/default/config.php file. Would remove the session variable and globals.inc.php stuff and migrate the require_once of sites/default/config.php in globals.php to be right above the require_once of sql.inc.php.
(sqlconf.php is an option, but is problematic since users will generally keep their old sqlconf.php on upgrades)

bradymiller:
Also note this will require unwinding of that more_secure global crud at top of config.php (no idea why i did that; will blame it on being overcaffeinated at the time :) ). Would convert those globals to simple globals (ie. keep them in config.php and remove the encrypt crud) and remove the decrypt of them where they are used in codebase (just 2 times).

Firehed:
Since I've been doing some recent DB work has been adjacent to this, I'll plan on taking over here.

I think a setting in the db config file(s) is the way to go[^f], with the general need to bypass it for the installer. Persisting this through sessions is pretty shaky and opens up a door to malicious requests (more DOS territory than anything else, no data leaks or anything).

That does present a challenge for upgraders; ideally, some sort of upgrade process detects the in-database setting and presents a note that they'll need to make a change to the file. This seems like another place where the banner for admins would be really useful (this could be trivially logged too, but that's less visible than a banner).

[^f]: or whatever future supplemental db config tooling we support, such as envvars. Not related to this change yet.

PR #10802, opened 2026-02-18
chore(db): Convert legacy Phreeze to updated connection handling

Firehed:
Towards #10774.

I went through several portal screens (with some added logging which isn't in the PR) to verify there were no hiccups in paths using the new connection and that it was being used, and didn't run into anything.

Removes manual credentials and certificate handling from `DataDriverMySQLi`

- Add "create MySQLI" path in DB factory using standard credentials handling
- Replace the portal vendored code to use it

Note: as commented in the code, this ignores the parameters passed to `Open()` entirely; they're (if you go deep enough) sourced from the same place.

Limited; Claude. Mostly for QA.

### DIFF
--- a/.phpstan/baseline/class.notFound.php
+++ b/.phpstan/baseline/class.notFound.php
@@ -1382,11 +1382,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/DB/DataDriver/MySQLi.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Method DataDriverMySQLi\\:\\:Open\\(\\) has invalid return type connection\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/DB/DataDriver/MySQLi.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Method DataDriverMySQLi\\:\\:Query\\(\\) has invalid return type resultset\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/DB/DataDriver/MySQLi.php',
--- a/.phpstan/baseline/empty.notAllowed.php
+++ b/.phpstan/baseline/empty.notAllowed.php
@@ -3469,11 +3469,6 @@ $ignoreErrors[] = [
 $ignoreErrors[] = [
     'message' => '#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#',
     'count' => 1,
-    'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/DB/DataDriver/MySQLi.php',
-];
-$ignoreErrors[] = [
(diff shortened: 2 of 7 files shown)


src/BC/DatabaseConnectionFactory.php
```diff
@@ -64,6 +65,52 @@ public static function createAdodb(
...
+                certificate: $config->sslClientCert['cert'] ?? null,
+                ca_certificate: $config->sslCaPath,
+                ca_path: null,
+                cipher_algos: null,
+            );
+        }
+
+        // TODO: Sockets support (do all paths at once)
+
+        $success = $mysqli->real_connect(
```
kojiromike, line 97: Is connection pooling intentionally dropped here? The old code in `MySQLi.php` had logic to prepend `p:` to the hostname when `enable_database_connection_pooling` was set:

```php
if ((!empty($globalsBag->get("enable_database_connection_pooling")) || !empty($session->get("enable_database_connection_pooling"))) && empty($globalsBag->get('connection_pooling_off'))) {
    $host = "p:" . $host;
}
```

I don't see equivalent handling in `DatabaseConnectionOptions` or here. The `createAdodb` method has a `$persistent` parameter that uses `PConnect()`, but `createMysqli()` doesn't have an equivalent. Should there be a `$persistent` parameter here that prepends `p:` to the hostname when true?

Firehed: Somewhat of an oversight, though I'm increasingly inclined to believe that the connection persistence (called pooling in config) doesn't work anywhere. As you've noted elsewhere, you need to connect to the DB to know if the setting is enabled, and it doesn't do anything to re-establish the connection.

I guess I'll restore it for consistency's sake for now. On the offchance it _does_ work correctly, I don't want that to regress.

kojiromike: https://github.com/openemr/openemr/issues/8890 in case you didn't already know the ticket number… ;)

Firehed: Figured you had that one memorized.

Firehed: n.b. I'm going to centralize that nasty detection logic first in a separate pr and merge it in

bradymiller: hi, the pooling worked when brought in many years and should still work. the first hit doesn't use the setting, but since it is then set via a session variable, it will then impact all subsequent hits in the same session.

Firehed: Ah, good to know! The setup has been restored on to this branch consistent with the previous logic.

src/BC/DatabaseConnectionFactory.php
```diff
@@ -66,6 +67,55 @@ public static function createAdodb(
...
+
+        if (!$success) {
+            throw new RuntimeException(
+                sprintf('Could not connect to the database (%s)',  $mysqli->connect_error),
+                $mysqli->connect_errno,
+            );
+        }
+
+        // This is preferred over SET NAMES since it also influences escaping
+        $mysqli->set_charset($config->charset);
```
bradymiller, line 114: nice!

PR #10803, opened 2026-02-19
refactor(db): centralize connection persistence detection

Firehed:
Towards #10774.

Consolidates duplicated connection pooling/persistence detection logic into a single location, improving maintainability and reducing code duplication.

- Add `DatabaseConnectionFactory::detectConnectionPersistence()` method that accepts explicit `ParameterBag` and `SessionWrapperInterface` parameters
- Add deprecated `DatabaseConnectionFactory::detectConnectionPersistenceFromGlobalState()` helper for callers that still rely on global state
- Migrate all existing callers to use the centralized methods:
  - `src/Gacl/Gacl.php`
  - `library/sql.inc.php` (2 locations)
  - `interface/modules/zend_modules/config/autoload/global.php`
  - `portal/patient/fwk/libs/verysimple/DB/DataDriver/MySQLi.php`

The logic was previously duplicated across these files with slight variations in how globals/session were accessed. Now there is a single source of truth.

Yes

All commits are co-authored by Claude Opus 4.5 as indicated in the commit messages.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.phpstan/baseline/empty.notAllowed.php
+++ b/.phpstan/baseline/empty.notAllowed.php
@@ -1443,7 +1443,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#',
-    'count' => 4,
+    'count' => 1,
     'path' => __DIR__ . '/../../interface/modules/zend_modules/config/autoload/global.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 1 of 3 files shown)


src/BC/DatabaseConnectionFactory.php
```diff
@@ -64,6 +66,43 @@ public static function createAdodb(
...
+        ParameterBag $globals,
+        SessionWrapperInterface $session,
+    ): bool {
+        if ($globals->getBoolean('connection_pooling_off')) {
+            return false;
+        }
+        if ($globals->getBoolean('enable_database_connection_pooling')) {
+            return true;
+        }
+        if (!empty($session->get('enable_database_connection_pooling'))) {
```
Firehed, line 79: I don't want to make assumptions about how `empty` vs `isset` might apply in the various internals here and decided to keep these as they were. It's still a net reduction in errors so I'll call it a win.

Firehed:
This will clobber some of my open DB PRs but I'll deal with that.

bradymiller:
nice!

PR #10805, opened 2026-02-19
refactor(installer): use centralized database connection factory

Firehed:
Towards #10774.

Consolidates database connection creation in the Installer to use the new `DatabaseConnectionFactory`, reducing code duplication and aligning with the ongoing database layer unification.

- Refactors `Installer::connect_to_database()` to use `DatabaseConnectionFactory::createMysqli()` instead of direct mysqli calls
- Makes `DatabaseConnectionOptions::inferSslPaths()` public so the Installer can reuse SSL certificate detection logic
- Removes unused mysqli wrapper methods (`mysqliInit`, `mysqliRealConnect`, `mysqliSslSet`) that were only needed for the old implementation
- Removes implementation-specific tests that tested mysqli internals (now handled by the factory); existing tests for `root_database_connection` and `user_database_connection` already cover success/failure paths

Yes

All commits include `Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>`. The changes are minimal refactoring (delegating to the factory) and code removal, with no substantial new logic added.


tests/Tests/Isolated/library/classes/InstallerTest.php
```diff
@@ -3019,362 +3013,6 @@ public function testExecuteSqlWithShowErrorFalse(): void
...
-        ];
-
-        return $this->getMockBuilder(Installer::class)
-            ->setConstructorArgs([$config, new NullLogger()])
-            ->onlyMethods($mockMethods)
-            ->getMock();
-    }
-
-    // Test connect_to_database through root_database_connection() - Success without SSL
-    public function testConnectToDatabaseSuccessNoSSL(): void
```
Firehed, line 3079: All of these removed tests were covering connection logic, which is now obsolete.

library/classes/Installer.class.php
```diff
@@ -1602,56 +1604,25 @@ protected function execute_sql(string $sql, bool $showError = true): mysqli_resu
...
-                    null,
-                    null,
-                    $pathToCerts . "mysql-ca",
-                    null,
-                    null
-                );
-            }
-        }
+        $siteDir = __DIR__ . "/../../sites/" . $this->site;
+        $ssl = DatabaseConnectionOptions::inferSslPaths($siteDir);
```
Firehed, line 1608: the whole implication that you'd have client certs present but you're still running the installer seems rather odd to me. It's not objectively incorrect, but seems so unlikely that the need to support it feels dubious.

bradymiller: hi @Firehed , docker install will have them on certain workflows (including the aws packages as I recall).

Firehed: That makes sense! I've been getting over a bad cold so I'm not firing on all cylinders at the moment 🙃

PR #10806, opened 2026-02-19
feat(patients): optionally generate csv file of duplicate patients

### LINKED ISSUE #6187: output csv file of reported duplicate patients
In admin/patient/manage duplicates  output the table into a csv file. 

## describe the problem :
Our situation is one where there are separate facilities run by separate staff and under different management structures.
(a) Where they have their own review process for determining duplicate patients and want to merge patients, having the ability to output  the table from Duplicate Manager into a CSV file enables staff to filter the list and produce separate overview lists for each team based on the facility noted in the system.
(b). having duplicate data in a spreadsheet makes it much easier to count and do other statistics on the duplicates lists enabling us to keep track of the number of duplicates, which facilities generate most, etc etc. 
So the export to CSV is a critical production tool. 
(c)  the export to csv makes it easier to use the merge feature directly instead of going through the duplicates manager, which is described as an unstable feature. 

In the Duplicate Management Patient window add a button "export to csv". If clicked this generates a csv file with name 'duplicate_patients_<openemr_name>_<date>.csv'. (note openemr_name is set in admin/config/branding/application title).
The content is the table as displayed with an added column 'group' which indicates how rows are grouped together, and thus without a blank row between each group.

ruthkonyn:
Fixes #6187

enables users to generate a csv file of a table of duplicate patients

displays a button which when pressed downloads a csv file to the client's system

no

### DIFF
--- a/interface/patient_file/manage_dup_patients.php
+++ b/interface/patient_file/manage_dup_patients.php
@@ -8,12 +8,12 @@
  * @author    Rod Roark <rod@sunsetsystems.com>
  * @author    Jerry Padgett <sjpadgett@gmail.com>
  * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @author    Ruth Moulton <ruth@muswell.me.uk>
  * @copyright Copyright (c) 2017-2021 Rod Roark <rod@sunsetsystems.com>
  * @copyright Copyright (c) 2025 Jerry Padgett <sjpadgett@gmail.com>
  * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @copyright Copyright (c) 2026 Ruth Moulton <ruth@muswell.me.uk>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
- *
- * Ruth Moulton optionally output a csv file of the list of dubplicate patients
  */
 
 require_once("../globals.php");
@@ -26,8 +26,6 @@ use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Core\Header;
 use OpenEMR\Core\OEGlobalsBag;
-//use OpenEMR\Services\FacilityService;
-
 
 $first_time = true;
 $group = 1;
@@ -35,27 +33,22 @@ $group = 1;
 $session = SessionWrapperFactory::getInstance()->getWrapper();
 
 /**
- * @param &$first_time
- * @param &$group
- * @param $row
- * @param $pid
+ * @param bool $first_time
+ * @param int $group
+ * @param array<string, string> $row
+ * @param string $pid
  * @return void
- *
- * rm added refs to  first_time and group as parameters, as PHPstan doesn't like global variables
  */
-function displayRow(bool & $first_time, int &$group, $row, $pid = ''): void
+function displayRow(bool &$first_time, int &$group, $row, string $pid = ''): void
 {
- //   global $first_time, $group;
-/** @var array<string, string> $row */
-/** @var string $pid */
+    /** @var array<string, string> $row */
+    $is_csv = ($_POST['form_csvexport'] ?? '') === 'CSV';
 
-
-    if (empty($pid)) {
+    if ($pid === '') {
         $pid = $row['pid'];
     }
 
     if (isset($row['myscore'])) {
-/** @var string $myscore */
         $myscore = $row['myscore'];
         $options = "<option value=''></option>" .
             "<option value='MK'>" . xlt('Merge and Keep') . "</option>" .
@@ -67,7 +60,7 @@ function displayRow(bool & $first_time, int &$group, $row, $pid = ''): void
             "<option value='R'>" . xlt('Recompute Score') . "</option>";
         if (!$first_time) {
             $group += 1;
-            if ($_POST['form_csvexport'] != "CSV") {     //rm - don't put the next line into the csv file
+            if (!$is_csv) {
                 echo " <tr><td class='detail' colspan='12'>&nbsp;</td></tr>\n";
             }
         }
@@ -75,7 +68,6 @@ function displayRow(bool & $first_time, int &$group, $row, $pid = ''): void
 
     $first_time = false;
     $ptname = $row['lname'] . ', ' . $row['fname'] . ' ' . $row['mname'];
-     /** @var string[]  $phones */
     $phones = [];
     if (trim((string) $row['phone_home'])) {
         $phones[] = trim((string) $row['phone_home']);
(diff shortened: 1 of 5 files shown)


kojiromike:
@ruthkonyn if you run `composer rector-fix` locally it will clear up the rector issues for you.

ruthkonyn:
thanks @kojiromike .  i was overwhelmed by the error reports so started a brand new PR with the latest base code, i.e. this one.  It hasn't helped! 
there is a clash between rector which suggested using += and PHPstan which doesn't like that. I'm new to PHPstan, and not understanding it's error reports! As an example can't global be used ? - it's used in most other php files...  any way, i've found out how to run that locally too, so hopefully will be able to sort out issues before the next pr - 

running rector locally was good. It's current complaints are my attempts to satisfy PHPstan, so don't worry, i'll try to get this sorted...

interface/patient_file/manage_dup_patients.php
```diff
@@ -26,6 +28,7 @@
 use OpenEMR\Services\FacilityService;
 
 $first_time = true;
+$group = 1;
```
kojiromike, line 32: Add `@var` annotations after the `global` statement to help PHPStan understand the types:

```suggestion
    global $first_time, $group;
    /** @var bool $first_time */
    /** @var int $group */
```

This tells PHPStan that `$group` is an int, eliminating several "mixed" type errors.

interface/patient_file/manage_dup_patients.php
```diff
@@ -53,7 +56,10 @@ function displayRow($row, $pid = ''): void
             "<option value='U'>" . xlt('Mark as Unique') . "</option>" .
             "<option value='R'>" . xlt('Recompute Score') . "</option>";
         if (!$first_time) {
-            echo " <tr><td class='detail' colspan='12'>&nbsp;</td></tr>\n";
+            $group += 1;
+            if ($_POST['form_csvexport'] != "CSV") {     //rm - don't put the next line into the csv file
+                echo " <tr><td class='detail' colspan='12'>&nbsp;</td></tr>\n";
```
kojiromike, line 64: The `(string)` cast on `','` is redundant (a string literal is already a string). This is what Rector flags.

Instead of:
```php
echo csvEscape(text($group)) . (string) ',';
```

Cast the *value*, not the comma, and use array+implode to avoid concatenation issues:
```php
$csvFields = [
    csvEscape(text((string) $group)),
    // ... other fields
];
echo implode(',', $csvFields) . "\n";
```

interface/patient_file/manage_dup_patients.php
```diff
@@ -126,7 +150,8 @@ function displayRow($row, $pid = ''): void
         <?php echo text($row['street']); ?>
     </td>
```
kojiromike: The date construction has concatenation issues. Consider:

```suggestion
    $dateInfo = getdate();
    $today = text((string) $dateInfo['year'] . $dateInfo['mon'] . $dateInfo['mday']);
    $openemrName = (string) ($GLOBALS['openemr_name'] ?? 'openemr');
    $filename = "duplicate_patients_{$openemrName}_{$today}.csv";
```

This avoids repeated `getdate()` calls and handles the `mixed` type from `$GLOBALS`.

interface/patient_file/manage_dup_patients.php
```diff
@@ -253,7 +289,29 @@ function selectChange(select, toppid, rowpid) {
             <div class="btn-sm-group mb-1 text-center">
                 <button class="btn btn-sm btn-primary btn-refresh" type='submit' name='form_refresh' value="Refresh"><?php echo xla('ReCalculate Scores') ?></button>
                 <button class="btn btn-sm btn-primary btn-print" type='button' value='Print' onclick='window.print()'><?php echo xla('Print'); ?></button>
+                <button class="btn btn-sm btn-primary btn-file" type='submit'  name='form_csvexport' value='CSV' ><?php echo xla('Generate a spreadsheet'); ?></button>
```
kojiromike: Same pattern here for the CSV headers - use array+implode:

```php
$csvHeaders = [
    csvEscape(xl('Group')),
    csvEscape(xl('Score')),
    csvEscape(xl('PID')),
    // ... etc
];
echo implode(',', $csvHeaders) . "\n";
```

interface/patient_file/manage_dup_patients.php
ruthkonyn: yes, it was PHPstan that seemed to be complaining about 

         Binary operation "." between mixed and ',' results in an error.                                   
                     🪪  binaryOp.invalid  
 - so i'd been trying to cast both sides as string in some way. 

but now i've got phpstan running locally i'm trying things out... implode looks useful!

interface/patient_file/manage_dup_patients.php
ruthkonyn: thanks, that all looks really useful, i'll take a look tomorrow.

ruthkonyn:
@kojiromike - still one PHPstan error being reported - it's not in a file that i modified for this PR but is in one of the libraries used by my code. It's in src/Services/SpreadSheetService.php
i can look at fixing that, but it will have to wait until sunday or monday now...

meanwhile one feature that i'd like your opinion on: the filename for the csv code is generated using the name of the openemr instance and the current date. This assumes no more than one report would be generated each day on a system - would it be better to include something like time of day (hour, minute) if multiple reports are likely to be needed in a given day?

my firefox on ubuntu gives me the option to change the file name if a file of the same name already exists, but i don't know what other systems do, and it's clunky for less techie staff or automatic generation...

ruthkonyn:
i can see the problem in src/Services/SpreadSheetService.php, lines 66 and 73, in fact there is already a comment there that highlights the problem and says it should be fixed!! i'm going to be busy all monday now, but hope to have a fix on tuesday.

interestingly i did work a fix for this today, but then found phpstan was no longer complaining, so i didn't submit it.

ruthkonyn:
hi @kojiromike 
i'm afraid i don't really understand the current PHPstan errors - they all seem to have been added to the $ignoreErrors array - so does this mean they can be ignored in any case?

when i run the test locally the errors do not show, and the report here doesn't give line numbers, so it's more difficult to see what they might be referring to. There is reference to the use of the 'text' function, i'll go through all of those and make sure they have @var statements

one of the errors is in a file that i've not been modifying, nor use. history/encounters.php

one of the errors is an alleged use of the keyword 'global' - but i've already corrected that, and can't see it occurring again, in manage_dup_patients.php -  the word does occur twice - once in a comment and the other time in a file name globals.php. 

But the global keyword is used in one or two of the included files (require_once) - do i need to modify them as well - these are files in the library directory.

i'll resubmit tomorrow.

kojiromike:
Hi, @ruthkonyn, the PHPStan errors are all coming from the `@var` annotations. Three specific issues:

**1. Missing `$` on variable names** — PHPStan requires `@var` to specify the variable name with `$`:

```php
// ❌ "does not specify variable name"
/** @var string pubpid */
/** @var string highlight_text */
/** @var string ptname */

// ✅ correct
/** @var string $pubpid */
/** @var string $highlight_text */
/** @var string $ptname */
```

**2. `string` vs `non-falsy-string`** — For `$instance_name` from `OEGlobalsBag::getInstance()->get()`, PHPStan infers it returns `non-falsy-string` (a string that's guaranteed non-empty). Annotating it as `string` widens the type, which PHPStan rejects. You can just drop that `@var` — the type is already known.

**3. You don't need `@var` on most of these** — When you're assigning from `$row['field']`, the variable type is already determined by the assignment. You only need `@var` for variables whose type PHPStan can't infer (like globals). For the `$row` array, a single `@var` at the top of the loop where `$row` is assigned would be cleaner than annotating every field access.

**About errors in other files** — You don't need to fix errors in `history/encounters.php` or any file you didn't modify. The baseline system should cover pre-existing errors. If PHPStan is complaining about files you didn't touch, it's likely because the baseline needs regenerating. Running `composer phpstan-baseline` in your branch should fix that — it regenerates the baseline to match the current state of the code.

**About `global`** — You don't need to change library files. The `global` keyword is fine in OpenEMR's legacy code. My earlier suggestion to add `@var` after `global` was just to help PHPStan understand the types of the global variables in *your* file.

**About the filename** — Including hours/minutes is a good idea. Something like `$dateInfo['hours'] . $dateInfo['minutes']` appended to the date would handle multiple reports per day without overwriting.

ruthkonyn:
> Hi, @ruthkonyn, the PHPStan errors are all coming from the `@var` annotations. Three specific issues:
> 
> **1. Missing `$` on variable names** — PHPStan requires `@var` to specify the variable name with `$`:
> 
> ```
> // ❌ "does not specify variable name"
> /** @var string pubpid */
> /** @var string highlight_text */
> /** @var string ptname */
> 
> // ✅ correct
> /** @var string $pubpid */
> /** @var string $highlight_text */
> /** @var string $ptname */
> ```
thanks, yes i'd seen this and thought i'd fixed it everywhere but clearly not - fix's in version i'm working on, not yet submitted.
> 
> **2. `string` vs `non-falsy-string`** — For `$instance_name` from `OEGlobalsBag::getInstance()->get()`, PHPStan infers it returns `non-falsy-string` (a string that's guaranteed non-empty). Annotating it as `string` widens the type, which PHPStan rejects. You can just drop that `@var` — the type is already known.
> 
 i put it in 'cause of PHPstan complaints, but i'll try taking it out again...

> **3. You don't need `@var` on most of these** — When you're assigning from `$row['field']`, the variable type is already determined by the assignment. You only need `@var` for variables whose type PHPStan can't infer (like globals). For the `$row` array, a single `@var` at the top of the loop where `$row` is assigned would be cleaner than annotating every field access.
> 

i put them in because PHPstan complained about them - i'll try again to take out the @var statements?

> **About errors in other files** — You don't need to fix errors in `history/encounters.php` or any file you didn't modify. The baseline system should cover pre-existing errors. If PHPStan is complaining about files you didn't touch, it's likely because the baseline needs regenerating. Running `composer phpstan-baseline` in your branch should fix that — it regenerates the baseline to match the current state of the code.
> 
thanks

> **About `global`** — You don't need to change library files. The `global` keyword is fine in OpenEMR's legacy code. My earlier suggestion to add `@var` after `global` was just to help PHPStan understand the types of the global variables in _your_ file.
> 

thanks

> **About the filename** — Including hours/minutes is a good idea. Something like `$dateInfo['hours'] . $dateInfo['minutes']` appended to the date would handle multiple reports per day without overwriting.

excellent, i'll add that in.

I have a meeting this evening, but should have some time tomorrow to continue with this...

ruthkonyn:
@kojiromike - i guess i don't understand PHPstan enough - but it seems to be inconsistent - for example 
  
 ```
if ($_POST['form_csvexport'] == "CSV" ) {   // rm out put the line to csv file
            echo csvEscape(text(strval($group))) . ',' ;
 /** @var string $myscore */
          echo  csvEscape(text($myscore)) . ','  ;
/**  @var string $pat_pid */
            $pat_pid = $row['pid'];
            echo  csvEscape($pat_pid) . ',';
/** @var string $ptname_pid */
            $ptname_pid = $row['ptnamepid'];
            echo csvEscape($ptname_pid) . ',';
```
generates the error (when run locally) on the line     $ptname_pid = $row['ptnamepid'];
  
```
108    Cannot access offset 'ptnamepid' on mixed.  
         🪪  offsetAccess.nonOffsetAccessible
```  

note that it is quite happy with the pat_pid value, which is similarly extracted from the $row array, just above. Anyway, i added in /** @var array<string> $row */, so the code is now

```
**  @var string $pat_pid */
            $pat_pid = $row['pid'];
            echo  csvEscape($pat_pid) . ',';
/** @var string $ptname_pid */
/** @var array<string> $row */
            $ptname_pid = $row['ptnamepid'];
            echo csvEscape($ptname_pid) . ',';
```
and phpstan reporting no errors. We'll see what happens when i submit as a PR :-)

**Comment by ruthkonyn** [comment]
hi mike (@kojiromike ),
the phpstan run here is still reporting a fair few errors in the file i've been working on, manage_dup_patients.php

i followed your suggestion - gave a single @var tag for the row[] array - i put it at the start of the function for which it's an argument - and took out all the other tags for the individual members of the array. on my system phpstan gives zero errors. So it's problematic trying to see what's needed when the only way to test is via a PR.

from experience there may be just one thing here that's triggering all the errors, but i don't think i can understand what it is.

do you have more suggestions? thanks

**Comment by kojiromike** [comment]
Hi @ruthkonyn, I dug into the PHPStan CI failure and here's what's going on and how to fix it.

## Why PHPStan CI Fails

The OpenEMR PHPStan CI does **not** check for zero errors — the codebase has ~162,000 pre-existing errors tracked in baseline files under `.phpstan/baseline/`. The CI check works like this:

1. Regenerates the baseline from scratch
2. Compares the regenerated baseline to the committed baseline files
3. If there's a diff, it fails

Your code changes the error count (fixing some old errors, introducing some new ones), so the committed baseline no longer matches. **You need to regenerate the baseline after fixing the code issues.**

## Code Issues to Fix

### 1. `$myscore` is never assigned

`$myscore` is referenced but never defined in `displayRow()`. It comes from the SQL result but isn't extracted from `$row`:

```php
// You have:
echo csvEscape(text($myscore)) . ',';

// But $myscore is never set — you need:
$myscore = $row['myscore'];
```

### 2. `$phones` is an array — can't pass to `text()` or `csvEscape()` directly

`$phones` is built as an array (`$phones[] = ...`). Passing an array to `text()` or `csvEscape()` (which expect strings) is a type error:

```php
// ❌ array passed to string function
echo csvEscape(text($phones)) . ',';

// ✅ join the array first
echo csvEscape(text(implode(', ', $phones))) . ',';
```

### 3. Remove most `@var` annotations

Your `@var array<string> $row` at the top of `displayRow()` actually *works* — it tells PHPStan that `$row` is an array of strings, which eliminates all the "Cannot access offset on mixed" errors. That's great! But after that, **you don't need `@var` on individual fields** pulled from `$row`. PHPStan already knows they're strings.

Remove these — they're unnecessary and some cause new errors:
```php
// Remove all of these:
/** @var string $myscore */
/** @var string $pid */
/** @var string $date_str */
/** @var string $regdate_str */
/** @var string $sex */
/** @var string $highlight_text */
/** @var string $ptname */
/** @var array<string> $phones */  // (this one is correct but could be just: @var string[] $phones)
```

The only `@var` you need is the one on `$row` at the top.

### 4. `selchange` vs `selectChange`

You renamed the JavaScript callback from `selectChange` to `selchange` in the HTML `onchange` handler, but the JavaScript function is still defined as `selectChange`. This will break the dropdown:

```php
// Your HTML has:
onchange='selchange(this, ...)'

// But the JS function is:
function selectChange(select, toppid, rowpid) {
```

Change it back to `selectChange` in the HTML.

### 5. Proper `$row` type annotation

Instead of `@var array<string> $row`, a more accurate type would be:

```php
/** @var array<string, string> $row */
```

This says "an array with string keys and string values" — which is what `sqlFetchArray()` returns.

## Regenerating the Baseline

After fixing the code issues above, regenerate the baseline:

```bash
composer phpstan-baseline
```

Then commit the updated baseline files along with your code changes. This should make the CI pass.

## Summary of Changes Needed

1. Set `$myscore = $row['myscore'];` before using it
2. Use `implode(', ', $phones)` before passing to `text()`/`csvEscape()`
3. Remove unnecessary `@var` annotations (keep only the one on `$row`)
4. Fix `selchange` → `selectChange` in the HTML
5. Run `composer phpstan-baseline` and commit the result

Let me know if you have questions!

ruthkonyn:
@kojiromike - i hope you got my reply to your most useful message above. I sent it by email and can't see it here (yet?)
There was one question at the end:

i've made the code changes - (but see comments in the email), and then
run phpstan-baseline - it generate a large number of errors, as an example here are just two:

Writing baseline file /home/ruth/software/OpenEMR/openemr/.phpstan/baseline/ternary.resultUnused.php with 1 errors
Writing baseline file /home/ruth/software/OpenEMR/openemr/.phpstan/baseline/throws.notThrowable.php with 4 errors

and on my branch about 17 changed files in .phpstan/baseline/ which i can commit.

also a couple of untracked - 
../../.phpstan/baseline/attribute.notFound.php
    ../../.phpstan/baseline/doWhile.alwaysFalse.php
which i'll not commit.

my question is - should i go ahead and commit and submit the changed baseline files, despite all the error reports? From what you've written about why phpstan CI fails, i think the answer is 'yes' so i'm going to go ahead and do that before it gets too late here (i may not have more time on this till friday afternoon or next week)

thanks for all your help

ruthkonyn:
looks like those two extra files were needed after all - have submitted them now

kojiromike:
Hi @ruthkonyn, thanks for sticking with this — I know the PHPStan baseline has been frustrating!

The remaining CI failure isn't actually a problem with your code. The issue is that master has had ~20 commits touching `.phpstan/` since you branched off (including a PHPStan version bump and a swagger-php 6.x upgrade). When you ran `composer phpstan-baseline` on your branch, it regenerated against the old base, so the baseline doesn't match what CI expects from current master.

The fix is to rebase onto current master and then regenerate the baseline — but don't worry about doing that yourself. I'll take care of the rebase and baseline regeneration and get this merged. Thanks for the contribution!

ruthkonyn:
thanks @kojiromike - on the point about rebasing - i had done a rebase on master earlier yesterday, didn't think to do another one before redoing the baseline and final submission. it has been a bit frustrating yes, but glad to be at the end of it. out of interest how long have the phpstan and rector tests been in place?

anyway i'll make the final fix you suggested to the PatientService.php code when i have time - hopefully before the end of the week, but depends on some other work i'm doing

PR #10820, opened 2026-02-20
refactor(db): Port migrations to doctrine WIP

Firehed:
After sitting on this for a short while, I think the best migration path all things considered looks something like this:

- wait for next release to get branched based on current schema
- create a patch on that release branch that creates the migrations table and seeds it with the files/entries that doctrine will use, using the existing migration system
- the upgrade process today (sql_upgrade.php) should either be migrated to call doctrine or deprecated entirely; either way the old custom sql system will be removed
- 

In terms of what's happening on this branch, I'm 50/50 between continuing on this direction (file per table + some future seed files) OR trying to consolidate so it's closer to the one-per-version schema. I don't think it matters much in practice.

---

Fixes #

Automated reworking of database.sql into Doctrine migrations, using their DSL

To do:
- [ ] MANUALLY verify schema consistency
- [ ] Sort out collation stuff (this isn't really related but is definitely a task)
- [ ] Figure out a handful of column types where there's not a "true" 1:1 mapping in the DSL
- [ ] Some index names changed
- [ ] Some table comments have disappeared?

This is nowhere near ready for prime-time but I want to get the branch somewhere for at least brainstorming.

In terms of integration, I want to have the installer run this and add a "hook" migration that determines if it's a new install or an existing one, that way nothing runs twice and it's safe for everyone.

Yes

Whole lotta Claude.


Firehed:
Making progress here.

https://github.com/doctrine/dbal/issues/7320 is in response to the `TINYINT` thing. `DATETIME` vs `TIMESTAMP` I'm going to open a separate issue about. The current CI failing is basically saying that the current _schema_ is noncompliant with SQL strict mode, which is a bit more worrying than the runtime.

Take Claude's summary below with a grain of salt, but it's directionally accurate.
---
## Schema Migration Status Update

Ran schema comparison between source (legacy SQL) and Doctrine migrations. Current diff summary:

| Category | Count | Status |
|----------|-------|--------|
| Total diff lines | ~1456 | |
| TINYINT(1) vs TINYINT(4) | ~194 | **Cosmetic** - Doctrine BOOLEAN outputs display width (4) |
| TINYINT vs SMALLINT | ~100 | **Doctrine limitation** - no native TINYINT type |
| TIMESTAMP vs DATETIME | 89 | **Doctrine limitation** - no TIMESTAMP type |
| ON UPDATE current_timestamp() | ~20 | **Doctrine limitation** - not supported |
| INT display widths | ~50 | **Cosmetic** - e.g. int(5) vs int(11) |
| Missing tables | 4 | **Intentional** - test/temp tables not migrated |
| Index naming | 0 | **Fixed** |
| Column collation | 0 | **Fixed** - utf8mb4_bin for lang_constants.constant_name |

### Recent Fixes
- Added `utf8mb4_bin` collation to `lang_constants.constant_name`
- Fixed `codes_history.active` and `codes_history.financial_reporting` to use `DEFAULT NULL`

### Remaining Differences (Unavoidable)
All remaining differences are due to Doctrine DBAL limitations:
1. **No TINYINT type** - Must use SMALLINT for non-boolean tiny integers
2. **No TIMESTAMP type** - Must use DATETIME (functionally equivalent for storage)
3. **BOOLEAN display width** - Doctrine outputs TINYINT(4) instead of TINYINT(1)
4. **No ON UPDATE clause** - Auto-update timestamps must be handled in application code
5. **INT display widths** - Cosmetic only, no storage/behavior impact

The schema is functionally equivalent; remaining differences are cosmetic or require application-level handling.

kojiromike:
@aisle-analyzer

PR #10839, opened 2026-02-20
fix(semgrep): add generate_select_list and json_encode as sanitizers

kojiromike:
## Summary

- Add `generate_select_list()` as a sanitizer - this function (library/options.inc.php) uses `attr()` internally to escape all output
- Add `json_encode()` as a sanitizer - PHP's JSON encoding escapes special characters, making output safe for HTML context

## Impact

Reduces `echoed-request` false positives from ~211 to ~72 alerts. The remaining alerts need individual review to distinguish real XSS issues from functions that internally sanitize their output.

## Test plan

- [x] Run `semgrep --config semgrep.yaml --include='*.php' .` locally to verify reduction
- [ ] Verify CI semgrep workflow passes

### DIFF
--- a/semgrep.yaml
+++ b/semgrep.yaml
@@ -55,8 +55,11 @@ rules:
   - pattern: xmlEscape(...)
   - pattern: csvEscape(...)
   - pattern: errorLogEscape(...)
-      # OpenEMR display function (library/options.inc.php) - uses htmlspecialchars internally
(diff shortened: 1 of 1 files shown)


bradymiller:
hi @kojiromike , Wouldn't consider json_encode safe for escaping in html context. Only safe for javascript context.

kojiromike:
@bradymiller I agree with you. I need to look into if semgrep can understand context, just the same as the twig auto-escaping conversation.

PR #10850, opened 2026-02-22
refactor(billing): return booleans instead of string values in connectivity checks

### LINKED ISSUE #10819: refactor: Return booleans instead of "Yes"/"No" strings in connectivity checks
Functions like `ConnectorApi::canConnectToClaimRev()` return the strings `"Yes"` or `"No"` instead of boolean `true`/`false`.

This pattern appears in multiple modules and makes the code harder to use correctly - callers must compare against magic strings rather than using native boolean logic.

**Example current code:**
```php
public static function canConnectToClaimRev()
{
    try {
        ClaimRevApi::makeFromGlobals();
        return "Yes";
    } catch (ClaimRevAuthenticationException) {
        return "No";
    }
}
```

**Should be:**
```php
public static function canConnectToClaimRev(): bool
{
    try {
        ClaimRevApi::makeFromGlobals();
        return true;
    } catch (ClaimRevAuthenticationException) {
        return false;
    }
}
```

Callers displaying this to users can convert to localized strings at the presentation layer.

**Files to check:**
- `interface/modules/custom_modules/oe-module-dorn/src/ConnectorApi.php`
- Other modules with similar connectivity check patterns

jbaiad:
## Summary

- Replace `"Yes"`/`"No"` string literals with native PHP `bool` in `ConnectorApi::canConnectToClaimRev()` and `ConnectivityInfo::$hasToken`
- Move Yes/No string rendering to the presentation layer in `debug-info.php` using `xlt()` on a boolean ternary
- Add explicit `: bool` return type annotation to `canConnectToClaimRev()`
- Remove now-resolved PHPStan baseline suppression for missing return type

## Changes

- `interface/modules/custom_modules/oe-module-dorn/src/ConnectorApi.php`: Changed return type from `string` (`"Yes"`/`"No"`) to `bool` (`true`/`false`); added `: bool` return type declaration
- `interface/modules/custom_modules/oe-module-claimrev-connect/src/ConnectivityInfo.php`: Changed `$hasToken` property type from `string` to `bool`; updated all assignments to boolean values
- `interface/modules/custom_modules/oe-module-claimrev-connect/public/debug-info.php`: Updated token display to use `$connectivityInfo->hasToken ? xlt("Yes") : xlt("No")` — localized string conversion now happens at the view layer
- `.phpstan/baseline/missingType.return.php`: Removed suppression for `canConnectToClaimRev()` return type (now explicitly typed)

## Testing

- Verify the ClaimRev Connect debug page still displays "Yes" or "No" for the Token field
- `composer phpstan` passes with no errors
- `composer phpcs` passes with no violations
- `codespell` passes on all changed files

Fixes #10819

> Code generated with the assistance of Claude (Anthropic)

### DIFF
--- a/.phpstan/baseline/missingType.return.php
+++ b/.phpstan/baseline/missingType.return.php
@@ -4822,11 +4822,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-dorn/src/ConnectorApi.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Method OpenEMR\\\\Modules\\\\Dorn\\\\ConnectorApi\\:\\:canConnectToClaimRev\\(\\) has no return type specified\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-dorn/src/ConnectorApi.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Method OpenEMR\\\\Modules\\\\Dorn\\\\ConnectorApi\\:\\:createRoute\\(\\) has no return type specified\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-dorn/src/ConnectorApi.php',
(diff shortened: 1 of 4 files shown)


kojiromike (approved):
LGTM - verified no other callers of canConnectToClaimRev() and ConnectivityInfo->hasToken is only used in debug-info.php which this PR updates.

PR #10865, opened 2026-02-22
fix(eyeform): pass encounter/pid in navbar form URLs to prevent stale session

### LINKED ISSUE #10844: bug(eyeform): creating new eye form fails with form not found
### Describe the bug

Image

>[Sat Feb 21 17:10:22.682815 2026] [php:notice] [pid 63:tid 63] [client 172.18.0.1:34976] [2026-02-21T17:10:22.681778+00:00] OpenEMR.WARNING: An attempt was made to view an eye form belonging to a different patient or encounter {"user-id":"1","requested-form-id":"2","session-pid":"1","session-encounter":"0"} [], referer: http://localhost:8300/interface/patient_file/encounter/load_form.php?formname=eye_mag

### To Reproduce
can go to a dev demo or use easy dev and try creating an eye form in an encounter

### Expected behavior
eye form is created and then opened for editing

**note the eye form is actually created but not displayed/opened so have to go back to the encounter screen and edit it:

Image

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

jbaiad:
## Summary

- Fixes a 404 "Form not found" error when creating a new Eye Exam form
- Passes `encounter` and `pid` as URL parameters in the encounter navbar
  form links, making them self-contained and immune to session staleness
- Removes the buggy `date("Ymd")` fallback from `eye_mag/new.php`
- Adds isolated Twig render tests and an E2E browser test verifying
  the fix

## Root Cause

The encounter navbar rendered form links as:

```
load_form.php?formname=eye_mag
```

without `pid` or `encounter` in the URL. When a user clicked a form
link, `new.php` read `$_SESSION['encounter']`, which could be stale
(`0`). The old code then fell back to `date("Ymd")` (e.g. `20260222`)
as the encounter ID, created a form with that bogus value, and
redirected to `view.php` — whose IDOR guard compared the stored
encounter (`20260222`) against the session encounter (still `0`), found
a mismatch, and returned HTTP 404 "Form not found".

## Changes

- **`interface/patient_file/encounter/forms.php`** — passes `encounter`
  and `pid` to the Twig render array (both already in scope from
  `globals.php`)
- **`templates/encounter/forms/navbar.html.twig`** — appends
  `&pid=X&encounter=Y` to every `load_form.php` URL so form links
  carry their own context
- **`interface/forms/eye_mag/new.php`** — removes the
  `!$_SESSION['encounter'] ? date("Ymd") : …` fallback; uses
  `$encounter` from scope (set by `load_form.php` from URL params);
  adds an `empty($encounter)` guard with a user-facing error
- **`.phpstan/baseline/`** — bumps baseline counts for the new
  `empty()` call and `(int)` cast
- **`tests/Tests/Isolated/Forms/EyeMag/NavbarFormUrlTest.php`** —
  renders `navbar.html.twig` with known parameters and asserts form
  URLs contain `pid=` and `encounter=` (3 test cases)
- **`tests/Tests/E2e/KkEncounterFormNavbarUrlTest.php`** — E2E browser
  test that navigates to an encounter, inspects navbar dropdown links
  via JavaScript, and asserts every `load_form.php` URL has `pid=` and
  `encounter=` with non-zero values

## How the fix flows

```
forms.php renders navbar with encounter=123&pid=1 in URL
  → User clicks "Eye Exam"
  → Browser requests load_form.php?formname=eye_mag&pid=1&encounter=123
  → load_form.php sets $encounter = $_GET['encounter']
  → load_form.php requires new.php (same scope)
  → new.php sees $encounter = 123 (correct, not from stale session)
  → Form created with encounter=123
  → Session synced via SessionUtil::setSession('encounter', 123)
  → Redirect to view.php — IDOR guard passes (stored=123, session=123)
```

## Testing

1. `composer phpstan` — static analysis passes
2. `composer phpcs` — code style passes
3. `composer codespell` — spelling passes
4. `composer phpunit-isolated -- --filter NavbarFormUrlTest` — 3/3 pass
5. E2E test (`KkEncounterFormNavbarUrlTest`) — verified against a live
   OpenEMR instance with Selenium: all navbar form URLs include
   `pid=` and `encounter=` params with correct values

Fixes #10844

> Code generated with the assistance of Claude (Anthropic)

### DIFF
--- a/tests/Tests/E2e/KkEncounterFormNavbarUrlTest.php
+++ b/tests/Tests/E2e/KkEncounterFormNavbarUrlTest.php
@@ -98,16 +98,14 @@ class KkEncounterFormNavbarUrlTest extends PantherTestCase
                 );
                 // The encounter value must not be 0 (the stale-session scenario)
                 $this->assertDoesNotMatchRegularExpression(
-                    '/encounter=0[^0-9]/',
+                    '/encounter=0(?:[^0-9]|$)/',
                     $onclickStr,
                     "Form link has encounter=0 (stale session bug): {$onclickStr}"
                 );
             }
-        } catch (\Throwable $e) {
+        } finally {
             $this->client->quit();
-            throw $e;
         }
-        $this->client->quit();
     }
 
     private function initClient(): void


interface/forms/eye_mag/new.php
```diff
@@ -40,7 +40,16 @@
     $group = $_SESSION['authProvider'];
 }
 
-$encounter = !$_SESSION['encounter'] ? date("Ymd") : $_SESSION['encounter'];
+// $encounter is already in scope from globals.php / load_form.php.
+// Guard against missing encounter (the scenario that caused #10844):
+if (empty($encounter)) {
```
kojiromike: Could we use an explicit check here instead of `empty()`? `empty()` is type-ambiguous — it treats `"0"`, `0`, `null`, `""`, and `false` all as empty, which obscures what we're actually guarding against.

```suggestion
if ($encounter === null || $encounter === '' || $encounter === 0 || $encounter === '0') {
```

Or if `$encounter` is always expected to be an integer at this point:

```php
if ((int) $encounter < 1) {
```

tests/Tests/E2e/KkEncounterFormNavbarUrlTest.php
```diff
@@ -0,0 +1,129 @@
...
+use OpenEMR\Tests\E2e\Encounter\EncounterOpenTrait;
+use OpenEMR\Tests\E2e\Login\LoginTestData;
+use OpenEMR\Tests\E2e\Login\LoginTrait;
+use OpenEMR\Tests\E2e\Patient\PatientTestData;
+use OpenEMR\Tests\E2e\Xpaths\XpathsConstants;
+use PHPUnit\Framework\Attributes\Depends;
+use PHPUnit\Framework\Attributes\Test;
+use Symfony\Component\Panther\PantherTestCase;
+
+class KkEncounterFormNavbarUrlTest extends PantherTestCase
```
kojiromike, line 35: This E2E test adds ~241 PHPStan baseline entries across 8 files. New code should pass PHPStan cleanly rather than expanding the baseline.

The errors are all fixable:
- **`mixed` type errors** — add `@var` annotations or type assertions to variables like `$this->crawler`, `$this->client`, etc.
- **`empty()` not allowed** (28) — replace with explicit checks
- **`?? is not nullable`** (42) — remove unnecessary null coalescing
- **`sqlQuery()` deprecated** (42) — use `QueryUtils::querySingleRow()` or `QueryUtils::fetchRecords()`
- **`encapsedStringPart.nonString`** (14) — cast or annotate variables used in string interpolation

interface/forms/eye_mag/new.php
```diff
@@ -72,6 +81,11 @@
     $sql = "insert into forms (date, encounter, form_name, form_id, pid, " .
             "user, groupname, authorized, formdir) values (NOW(),?,?,?,?,?,?,?,?)";
     $answer = sqlInsert($sql, [$encounter,$form_name,$newid,$pid,$user,$group,$providerid,$form_folder]);
+    // Keep the session encounter in sync with the value just written to the
+    // forms table.  view.php reads $encounter exclusively from the session
+    // (via globals.php) and compares it against the stored encounter for IDOR
+    // protection; if they diverge the guard fires a 404 "Form not found".
```
kojiromike, line 89: Good catch — this `SessionUtil::setSession` call closes the gap between the URL-passed encounter and what `view.php` reads from the session. Nice comment explaining the IDOR guard interaction.

kojiromike (changes requested):
Nice root-cause analysis and well-structured fix. The core approach — passing encounter/pid in the URL rather than relying on session state — is correct, and the `SessionUtil::setSession` sync is a good catch.

Two items to address before this can merge:

1. **`empty()` in new code** — `empty()` is type-ambiguous and we're trying to move away from it. Please use an explicit check instead.
2. **PHPStan in the E2E test** — New code should pass PHPStan cleanly rather than expanding the baseline by ~241 entries. The errors are all standard type-safety issues (mixed types, `empty()`, deprecated `sqlQuery()`, unnecessary `??`) that are straightforward to fix.

stephenwaite:
hi @jbaiad, thanks for your PR! I'm not sure it's fixing the issue because the encounter already does exist. I tested on your branch and still getting the error.

jbaiad:
> hi @jbaiad, thanks for your PR! I'm not sure it's fixing the issue because the encounter already does exist. I tested on your branch and still getting the error.

@stephenwaite thanks for flagging! I think my latest commits should address the issue fully now. Looks like the additional URL params were getting lost in some redirect hops.

tests/Tests/E2e/KkEncounterFormNavbarUrlTest.php
```diff
@@ -0,0 +1,208 @@
...
+                    "Form link missing encounter param: {$onclickStr}"
+                );
+                // The encounter value must not be 0 (the stale-session scenario)
+                $this->assertDoesNotMatchRegularExpression(
+                    '/encounter=0[^0-9]/',
+                    $onclickStr,
+                    "Form link has encounter=0 (stale session bug): {$onclickStr}"
+                );
+            }
+        } catch (\Throwable $e) {
```
kojiromike: Minor: the `try/catch` with `$this->client->quit()` duplicated in both the catch block and after the try could be simplified with a `finally` block:

```php
$this->initClient();
try {
    // ... test logic ...
} finally {
    $this->client->quit();
}
```

This avoids the duplicated `quit()` call and is the idiomatic PHP pattern for resource cleanup.

tests/Tests/E2e/KkEncounterFormNavbarUrlTest.php
```diff
@@ -0,0 +1,208 @@
...
+                    "Form link missing pid param: {$onclickStr}"
+                );
+                $this->assertStringContainsString(
+                    'encounter=',
+                    $onclickStr,
+                    "Form link missing encounter param: {$onclickStr}"
+                );
+                // The encounter value must not be 0 (the stale-session scenario)
+                $this->assertDoesNotMatchRegularExpression(
+                    '/encounter=0[^0-9]/',
```
kojiromike: Nit: the regex `/encounter=0[^0-9]/` won't match if `encounter=0` happens to be at the very end of the string (no character after `0` for `[^0-9]` to match). Not a real problem in practice since these onclick strings have trailing characters, but `/encounter=0(?:[^0-9]|$)/` would be technically correct.

PR #10883, opened 2026-02-23
refactor(db): Remove support for secure_sqlconf.php

Firehed:
Towards #10774

https://community.open-emr.org/t/adodb-mysql-log-inc-php/10251/22

While well-intentioned, the "private database" connection has been rarely used and, upon looking at the tables it was designed to support, was't really reliable. While I haven't verified every touchpoint, some AI research indicates that trying to configure it today (and actually denying access to the "secure" tables) would break your installation, full stop.

I like what this was trying to do, but I don't think it makes sense to maintain it as-is. Future improves to database access controls could allow this to come back more reliably and consistently, but I'm going to call that out of scope for now.

- No-ops all of the previous internals of `getPrivDB()`, just return the existing connection
- Removes some duplicated logic from one of the calling functions
- Marks `getPrivDB()`, `privQuery()`, `privStatement()` as deprecated
- Fixes the return type of `get_db()`

No


bradymiller:
agree with removing this :+1:

PR #10891, opened 2026-02-24
fix(search): wrap DATETIME empty-string comparisons with CAST

### LINKED ISSUE #10834: bug(prescriptions): Unable to edit/add prescriptions after 7.0.4 upgrade
### Describe the bug

Received via [forum post](https://community.open-emr.org/t/unable-to-get-prescriptions-after-7-0-4-upgrade/26569).
Appears to be only a MySQL strict mode conflict `MySQL 5.7.5+` (MariaDB diverged which explains why it's not reproducible in easy dev or online demos)
>The query contains enddate = '' to catch legacy empty-string dates stored in the lists.enddate DATETIME column, but MySQL strict mode treats '' as an invalid DATETIME value and throws an error before it can even evaluate the comparison.

### To Reproduce
see above forum post

### Expected behavior

able to add/edit prescription

### Disclaimer
I ❤️ openemr.

kojiromike:
## Summary
- MySQL 8.4 rejects `DATETIME = ''` during `PREPARE` statement validation, regardless of `sql_mode` settings
- The FHIR search layer generates `enddate = ''` for the `MISSING` modifier in `SearchFieldStatementResolver::resolveTokenField()`, which fails on MySQL 8.4
- Wrap empty-string comparisons with `CAST(field AS CHAR)` so `PREPARE` validation succeeds on all MySQL versions
- Remove dead `enddate = ''` checks from `eRx_xml.php` — DATETIME columns cannot store empty strings, so these conditions never match

## Detail

The call chain is: `C_Prescription::getDiagnosisCodesList()` → `PatientIssuesService::getActiveIssues()` → `search()` → `FhirSearchWhereClauseBuilder` → `SearchFieldStatementResolver::resolveTokenField()`, which generates `(enddate IS NULL OR enddate = '')`.

MySQL 8.4 validates column types during `PREPARE` and rejects comparing a DATETIME column to an empty string, even with `sql_mode = ''`. The `CAST(field AS CHAR)` wrapper is safe for all column types: for DATETIME it yields a date string (never `''`), and for VARCHAR it's a no-op.

Closes #10834

## Test plan
- [ ] New isolated unit tests pass: `composer phpunit-isolated`
- [ ] PHPStan passes
- [ ] Test prescription add/edit on MySQL 8.4 to confirm no error
- [ ] Test prescription add/edit on MariaDB to confirm no regression

### DIFF
--- a/interface/eRx_xml.php
+++ b/interface/eRx_xml.php
@@ -788,7 +788,7 @@ function PatientMedication($doc, $r, $pid, $med_limit)
     global $msg;
     $active = '';
     if ($GLOBALS['erx_upload_active'] == 1) {
-        $active = " and (enddate is null or enddate = '' or enddate = '0000-00-00' )";
+        $active = " and (enddate is null or enddate = '0000-00-00' )";
     }
 
     $res_med = sqlStatement("select * from lists where type='medication' and pid=? and title<>''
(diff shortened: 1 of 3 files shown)


Firehed (approved):
Good find.

stephenwaite (approved):

stephenwaite:
:man_dancing:

PR #10894, opened 2026-02-24
fix(eye): add state license number to prescription popup

### LINKED ISSUE #10893: feat(eye): add state license number to prescription
### Describe the problem

no state license number to prescription which can be a requirement for some ophthalmologists 

add it

stephenwaite:
Fixes #10893


bradymiller:
Is the phpstan fix?
```
<?php echo text((string)$prov_data['state_license_number']); ?>
```

kojiromike:
> What's the project's thinking on adding to the baseline versus making the necessary changes to resolve the underlying PHPStan errors for existing files like this?
> 
> Do we eventually want the repo to have no PHPStan errors ignored in the baseline?

All of the phpstan findings indicate a real problem. Either legitimately incorrect code, or an incorrect annotation giving phpstan the wrong idea. So we should be aiming to monotonically decrease the findings, even if we can't make them all go away at once. A good rule of thumb is: don't introduce _new_ problems in a change, and _do_ fix problems on lines that you touch.

PR #10895, opened 2026-02-24
fix(security): protect bin/ directory from web access

### LINKED ISSUE #10797: fix: protect bin/ directory from web access
## Problem

The `bin/` directory contains CLI scripts that should not be web-accessible:
- `bin/console`
- `bin/command-runner`
- `bin/generate-ehi-documentation.sh`
- `bin/generate-phpstan-types.php`
- `bin/get-v-database.php`

Currently there's no protection to block web access. While these scripts are generally harmless, exposing CLI tools to the web is poor security practice.

### 1. Add `bin/.htaccess` (Apache)

```apache
# CLI scripts only - deny web access
Require all denied
```

### 2. Update nginx documentation

The wiki or other nginx setup docs should include a deny rule for `bin/`:

```nginx
location /bin/ {
    deny all;
    return 404;
}
```

## Context

Discovered while adding `bin/get-v-database.php` in #10768.

craigrallen:
Fixes #10797

The `bin/` directory contains CLI-only tools (`console`, `command-runner`, PHP scripts) that are not intended to be accessible via the web server. Unlike other sensitive directories (e.g., `sites/*/documents`, `portal/patient/fwk/libs`), the `bin/` directory had no web access restrictions.

- Add `.htaccess` file to `bin/` with `Deny From All` to block access on Apache deployments
- Add `location ^~ /bin/` deny block to the CI nginx configuration (`ci/nginx/nginx.conf`)
- Add the same deny block to the dev nginx configuration (`docker/library/dockers/dev-nginx/nginx.conf`)

#### If you answered yes:
Code was generated with assistance from Claude Code (Anthropic). All files in this PR were AI-assisted. The changes have been reviewed and tested by the contributor.

### DIFF
--- a/ci/nginx/nginx.conf
+++ b/ci/nginx/nginx.conf
@@ -113,7 +113,6 @@ http {
         # deny web access to bin/ directory (CLI-only tools)
         location ^~ /bin/ {
             deny all;
-            return 404;
         }
 
         # Pick one of the following two blockc, but not both:
--- a/docker/library/dockers/dev-nginx/nginx.conf
+++ b/docker/library/dockers/dev-nginx/nginx.conf
@@ -147,7 +147,6 @@ http {
         # deny web access to bin/ directory (CLI-only tools)
         location ^~ /bin/ {
             deny all;
-            return 404;
         }
 
         # Pick one of the following two blockc, but not both:
--- a/setup.php
+++ b/setup.php
@@ -1669,7 +1669,6 @@ STP4TOP;
                             . text("}") . "\n\n"
                             . text("location ^~ /bin/ {") . "\n"
                             . text("    deny all;") . "\n"
-                            . text("    return 404;") . "\n"
                             . text("}")
                         . "</code></pre>
 


bradymiller:
hi @craigrallen , Also need to address the configuration instructions for apache and nginx in the setup.php script.

craigrallen:
Thanks @bradymiller — added `bin/` directory deny instructions to both the Apache and nginx/FPM configuration sections in `setup.php`. Pushed in e183324.

kojiromike (commented):
Good security improvement — layered protection across Apache, nginx, and setup instructions is the right approach.

One nit on the nginx blocks: `deny all` followed by `return 404` is redundant — `return 404` takes precedence and `deny all` never fires. The existing deny blocks in the same file (e.g., the `documents` location) use only `deny all` without `return 404`. I'd suggest picking one approach for consistency:

- `deny all` alone (matches existing pattern, returns 403)
- `return 404` alone (hides directory existence, but different from existing blocks)

Same applies to the setup.php nginx instructions.

Minor: the `.htaccess` uses Apache 2.2 syntax (`order deny,allow` / `Deny From All`) while setup.php uses 2.4 syntax (`Require all denied`). Both work (2.2 syntax is still supported via mod_access_compat), and the `.htaccess` matches existing files like `vendor/.htaccess`, so this is fine.

kojiromike (approved):
Previous feedback addressed: redundant `return 404` removed from nginx blocks, and `setup.php` instructions added for both Apache and nginx/FPM. LGTM.

PR #10897, opened 2026-02-24
fix(security): add safe_href URL scheme validation for Twig href attributes

### LINKED ISSUE #10509: Twig href/action attributes lack URL scheme validation
The `|attr` and `|attr_url` Twig filters prevent HTML attribute breakout (quote injection) but do not validate URL schemes. A server-side value containing `javascript:`, `data:`, or `vbscript:` would pass through both filters and execute in the browser when used in `href` or `action` attributes.

The existing `|javascriptStringRemove` filter only strips the literal word "javascript" and does not cover `data:`, `vbscript:`, `blob:`, or other dangerous schemes.

OpenEMR already ships HTMLPurifier (`ezyang/htmlpurifier`), which includes `HTMLPurifier_AttrDef_URI` with a proper scheme allowlist (`http`, `https`, `mailto`, `ftp`, `tel`, etc.). A lightweight Twig filter wrapping this — or a simpler allowlist check using `parse_url()` + `attr()` — would close the gap.

Current exposure is low since the URLs in these attributes generally come from server-side PHP rather than raw user input, but defense-in-depth calls for scheme validation at the template layer.

craigrallen:
## Summary

Adds a `safe_href()` PHP function and `|safe_href` Twig filter that validates URL schemes against an allowlist before rendering in `href`/`action` attributes. Dangerous schemes (`javascript:`, `data:`, `vbscript:`, `blob:`, etc.) are replaced with `#`.

**Problem:** The existing `|attr` filter prevents HTML attribute breakout (quote injection) but does not validate URL schemes. The existing `|javascriptStringRemove` filter only strips the literal word "javascript" and does not cover `data:`, `vbscript:`, `blob:`, or other dangerous schemes. A server-side value containing any of these schemes would pass through and execute in the browser.

**Solution:** 
- New `safe_href()` function in `library/htmlspecialchars.inc.php` using `parse_url()` + scheme allowlist (`http`, `https`, `mailto`, `tel`, `ftp`, `ftps`) + `attr()` escaping
- Registered as `|safe_href` Twig filter in `TwigExtension.php`
- Applied to 5 highest-risk template locations:
  - `card_base.html.twig` — 2 completely unescaped dynamic href attributes
  - `rx.html.twig` — 1 completely unescaped dynamic href attribute
  - `about.html.twig` — 2 hrefs using `|javascriptStringRemove|attr` (incomplete protection)
  - `telecom_display.html.twig` — 1 user-input URL using only `|attr`
- Updated semgrep rule messages to reference `|safe_href`
- Added isolated PHPUnit tests covering safe URLs, dangerous schemes, HTML escaping, null input, and whitespace

Fixes #10509

## Testing

- Isolated PHPUnit test: `tests/Tests/Isolated/library/SafeHrefTest.php`
  - Tests all 6 allowed schemes pass through with `attr()` escaping
  - Tests relative URLs (`/path`, `?query`, `#fragment`) pass through  
  - Tests `javascript:`, `data:`, `vbscript:`, `blob:` all return `#`
  - Tests case-insensitive scheme matching
  - Tests null and whitespace handling
  - Tests HTML entity escaping in allowed URLs
- Manual verification: Patient card buttons, Rx card, About page, and Telecom display should render links correctly for valid URLs and block dangerous schemes

## AI Disclosure

Yes — Claude Code (Anthropic) used for implementation. Code has been reviewed by the contributor.

## Checklist

- [x] New function follows existing escaping function patterns in `htmlspecialchars.inc.php`
- [x] Filter registered alongside existing filters in `TwigExtension.php`  
- [x] Template changes are minimal — only the filter name changes, not the template structure
- [x] Semgrep rule messages updated to document `|safe_href` as safe pattern
- [x] Isolated tests included (no Docker required)

### DIFF
--- a/library/htmlspecialchars.inc.php
+++ b/library/htmlspecialchars.inc.php
@@ -98,8 +98,6 @@ function safe_href(?string $url): string
     if ($scheme === false) {
         return '#';
     }
-        return attr($url);
-    }
 
     // Allowlist of safe URL schemes
     $allowedSchemes = ['http', 'https', 'mailto', 'tel', 'ftp', 'ftps'];


library/htmlspecialchars.inc.php
```diff
@@ -58,6 +58,53 @@ function attr_url($text): string
...
+
+    // Allowlist of safe URL schemes
+    $allowedSchemes = ['http', 'https', 'mailto', 'tel', 'ftp', 'ftps'];
+
+    if (in_array(strtolower($scheme), $allowedSchemes, true)) {
+        return attr($url);
+    }
+
+    // Disallowed scheme — return safe fallback
+    return '#';
```
bradymiller, line 114: hi @craigrallen , would it make sense to log this at debug or error level?

craigrallen: Good catch — added an `error_log()` call at error level since a disallowed scheme could indicate an XSS attempt. Uses `errorLogEscape()` for safe output. Pushed in 1e3ac55.

craigrallen: > hi @craigrallen , would it make sense to log this at debug or error level?

I've updated to log at error level. Good call, sorry I missed that

bradymiller: sorry @craigrallen , I should of directed your use of OpenEMR's SystemLogger class for doing the error logging. In this case was more thinking of debug mode case (for example, for some reason a valid string got caught up in the filtering) but error also ok.

craigrallen: Thanks for the guidance @bradymiller — replaced `error_log()` with `SystemLogger::debug()` since a blocked scheme might need investigation but isn't necessarily an attack. Pushed in b33bc35.

kojiromike: Hmm, I don't see that commit in this PR. But also, I don't agree that it should be debug. Even if it's not _necessarily_ an attack, there are basically three scenarios to consider:

1. it's an attack. Log a warning.
2. Someone is developing on openemr and made a mistake. Why not log a warning?
3. We have some way a user can inject a url via some (possibly ill-advised) configuration/database thing, and they made a mistake. Log a warning (at least).

in two of these cases, I'd expect debug to not be visible at all. Running openemr in production, I definitely want to see this log message.

bradymiller: hi @kojiromike and @craigrallen  , warning ok with me

craigrallen: Updated to `warning` level per @kojiromike's and @bradymiller's consensus. Pushed in 65a04b4.

library/htmlspecialchars.inc.php
```diff
@@ -58,6 +58,57 @@ function attr_url($text): string
...
+    // Relative URLs (path-only or query-only) are safe
+    if ($url[0] === '/' || $url[0] === '?') {
+        return attr($url);
+    }
+
+    // Determine the URL scheme
+    $scheme = parse_url($url, PHP_URL_SCHEME);
+
+    // No explicit scheme means a relative URL — safe
+    if ($scheme === null || $scheme === false) {
```
kojiromike: Minor suggestion: when `parse_url` returns `false` (seriously malformed URL), the safer default would be to block rather than allow. Browsers are more forgiving than PHP's parser, so a URL that PHP can't parse might still be executable in a browser.

```suggestion
    if ($scheme === null) {
        return attr($url);
    }

    // Block malformed URLs that parse_url can't handle
    if ($scheme === false) {
        return '#';
    }
```

This is a very unlikely edge case, so not blocking on it — just something to consider.

PR #10898, opened 2026-02-24
fix(core): read software version from globals instead of database

### LINKED ISSUE #10821: bug: version always requires a database migration
OpenEMR's version semantics are too tightly coupled to the database. For example, the VersionService, and the About OpenEMR tab, will only state what version is in the `version` table in the database. The `version.php` file is only used, practically, for setting the version in that table.

But that doesn't really make sense, because version.php also has a `v_database` value, and that's the only value the database actually cares about. There really oughtn't be a table for version at all, _except_ for tracking `v_database`. The rest of the version is the version of the software, and that can be gotten entirely from version.php without looking at the database.

craigrallen:
## Summary

Changes `VersionService::asString()` to read the software version from `version.php` globals instead of querying the `version` database table.

**Problem:** The About page and all other version display callsites use `VersionService::asString()`, which queries `SELECT * FROM version`. This means every release requires a database migration just to update the displayed version — even when there are no actual schema changes.

**Solution:** Read `$GLOBALS['v_major']`, `$GLOBALS['v_minor']`, `$GLOBALS['v_patch']`, `$GLOBALS['v_tag']`, and `$GLOBALS['v_realpatch']` directly, which are already populated from `version.php` via `interface/globals.php` (lines 399-403).

**What's unchanged:**
- `fetch()` still queries the database — needed by `admin.php` for upgrade detection (comparing `v_database` from code vs DB)
- `update()` still writes to the database — needed by upgrade scripts
- The method signature and return format are identical

Fixes #10821

## Testing

- The About page (`/interface/main/about_page.php`) should display the version from `version.php` without needing a DB migration
- All 11 callsites of `asString()` continue to work identically since `version.php` globals match what was previously in the DB after migration
- `admin.php` upgrade detection is unaffected (it queries the DB directly, not via `asString()`)

## AI Disclosure

Yes — Claude Code (Anthropic) used for implementation. Code has been reviewed by the contributor.


src/Services/VersionService.php
```diff
@@ -31,19 +31,29 @@ public function fetch(): array
...
+     * @return string Dot separated major, minor, patch version string (tag at end, if included)
      */
     public function asString(bool $includeTag = true, bool $includeRealpatch = true): string
     {
-        $v = $this->fetch();
-        $string = "{$v['v_major']}.{$v['v_minor']}.{$v['v_patch']}";
-        $string = ($includeTag == true) ? $string . "{$v['v_tag']}" : $string;
-        if ($includeRealpatch && (!empty($v['v_realpatch']))) {
-            $string .= " (" . $v['v_realpatch'] . ")";
+        $globals = \OpenEMR\Core\OEGlobalsBag::getInstance();
```
kojiromike: Please add a `use` statement so you don't have to fully qualify this.

kojiromike (approved):

src/Services/VersionService.php
```diff
@@ -31,19 +31,29 @@ public function fetch(): array
...
+        $major = (string) $globals->get('v_major', '0');
+        $minor = (string) $globals->get('v_minor', '0');
+        $patch = (string) $globals->get('v_patch', '0');
+        $tag = (string) $globals->get('v_tag', '');
+        $realpatch = (string) $globals->get('v_realpatch', '');
```
bradymiller: @kojiromike , shouldn't the explicit cast work here for phpstan?

src/Services/VersionService.php
craigrallen: Good catch — added a `use OpenEMR\Core\OEGlobalsBag;` statement and switched to the short name. Pushed in 363b9ad.

src/Services/VersionService.php
```diff
@@ -31,19 +33,29 @@ public function fetch(): array
...
      */
     public function asString(bool $includeTag = true, bool $includeRealpatch = true): string
     {
-        $v = $this->fetch();
-        $string = "{$v['v_major']}.{$v['v_minor']}.{$v['v_patch']}";
-        $string = ($includeTag == true) ? $string . "{$v['v_tag']}" : $string;
-        if ($includeRealpatch && (!empty($v['v_realpatch']))) {
-            $string .= " (" . $v['v_realpatch'] . ")";
+        $globals = OEGlobalsBag::getInstance();
+        $major = (string) $globals->get('v_major', '0');
```
kojiromike: `ParameterBag::get()` returns `mixed`, so PHPStan flags `(string) $globals->get(...)` as "Cannot cast mixed to string" (this is what's failing CI).

`ParameterBag` has a `getString()` method that returns `string` directly — it handles the type check internally and throws if the value isn't scalar. Switching to that fixes PHPStan and removes the need for the explicit casts:

```php
$major = $globals->getString('v_major', '0');
$minor = $globals->getString('v_minor', '0');
$patch = $globals->getString('v_patch', '0');
$tag = $globals->getString('v_tag', '');
$realpatch = $globals->getString('v_realpatch', '');
```

src/Services/VersionService.php
```diff
@@ -31,19 +33,29 @@ public function fetch(): array
...
+        $minor = (string) $globals->get('v_minor', '0');
+        $patch = (string) $globals->get('v_patch', '0');
+        $tag = (string) $globals->get('v_tag', '');
+        $realpatch = (string) $globals->get('v_realpatch', '');
+
+        $string = "{$major}.{$minor}.{$patch}";
+        $string = ($includeTag == true) ? $string . $tag : $string;
```
kojiromike: Nit: `$includeTag == true` uses loose comparison (carried over from the old code). Since `$includeTag` is already typed `bool`, this could just be `$includeTag`:

```suggestion
        $string = $includeTag ? $string . $tag : $string;
```

src/Services/VersionService.php
kojiromike: @bradymiller yes, I'll post a review including that, thanks.

kojiromike (commented):
Thanks for picking this up! The direction is right — reading the displayed version from globals instead of the DB is exactly what #10821 asks for.

A couple of things to address:

1. **PHPStan failures** — 5 "Cannot cast mixed to string" errors on the `(string) $globals->get(...)` lines. `ParameterBag::get()` returns `mixed`, and PHPStan won't allow casting `mixed` to `string`. The parent class has a `getString()` method that returns `string` directly — switching to that fixes PHPStan and removes the need for the casts.

2. **Partial fix for #10821** — The issue analysis ([comment](https://github.com/openemr/openemr/issues/10821#issuecomment-3934014958)) called for introducing distinct `getSoftwareVersion()` and `getSchemaVersion()` methods so callers have a clear API for "what version is the code?" vs "what version is the database schema?". Right now `fetch()` still serves both purposes, which makes the boundary fuzzy. Would you be up for adding those methods in this PR, then having `asString()` delegate to `getSoftwareVersion()`? That way `fetch()` can become `getSchemaVersion()` (or wrap it), and the intent at each callsite is unambiguous.

kojiromike:
Thanks @craigrallen for picking this issue up first — your PR helped surface the right direction (reading from globals instead of the DB).

This has been addressed in #11428, which took a slightly different approach: instead of modifying `asString()` in place, it introduces a `SoftwareVersion` DTO (reading from globals) and a `SchemaVersion` DTO (reading from the DB) and removes `asString()` entirely. The 14 callsites were migrated to use the typed DTO directly.

Closing this in favor of the merged PR. Thanks again for the original push on #10821!

Issue #2010, opened 2018-12-03
replace eval() for js files


ghost:
I believe this is the `eval()` described here: http://php.net/manual/en/function.eval.php

> The eval() language construct is very dangerous 

It is also risky for JS and I'll find a way to replace some of the instances I uncovered as well.

danehrlich1:
`interface/forms/fee_sheet/review/js`  (1 usage found)
                fee_sheet_core.js  (1 usage found)
                    69 eval(diags_matches[i]);
            `interface/main/calendar/modules/PostCalendar/pnincludes`  (16 usages found)
                `overlib.js`  (8 usages found)
                    931 winoffset = (ie4) ? eval('o3_frame.'+docRoot+'.scrollLeft') : o3_frame.pageXOffset;
                    932 if (ie4) iwidth = eval('o3_frame.'+docRoot+'.clientWidth');
                    937 if ( (o3_x - winoffset) > ((eval(iwidth)) / 2)) {
                    951 if ( (eval(placeX) + eval(o3_width)) > (winoffset + iwidth) ) {
                    981 scrolloffset = (ie4) ? eval('o3_frame.'+docRoot+'.scrollTop') : o3_frame.pageYOffset;
                    985 if (ie4) iheight = eval('o3_frame.'+docRoot+'.clientHeight');
                    988 iheight = (eval(iheight)) / 2;
                    1035 if (ie5) {o3_x=eval('event.x+o3_frame.'+docRoot+'.scrollLeft'); o3_y=eval('event.y+o3_frame.'+docRoot+'.scrollTop');}
                `overlib_mini.js` (8 usages found)
                    575 winoffset=(ie4)? eval('oframe.'+docRoot+'.scrollLeft'): oframe.pageXOffset;
                    576 if(ie4)iwidth=eval('oframe.'+docRoot+'.clientWidth');
                    579 if((ox - winoffset)>((eval(iwidth))/ 2)){
                    591 if((eval(placeX)+ eval(owidth))>(winoffset + iwidth)){
                    613 scrolloffset=(ie4)? eval('oframe.'+docRoot+'.scrollTop'): oframe.pageYOffset;
                    615 if(ie4)iheight=eval('oframe.'+docRoot+'.clientHeight');
                    617 iheight=(eval(iheight))/ 2;
                    649 if(ie5){ox=eval('event.x+oframe.'+docRoot+'.scrollLeft');oy=eval('event.y+oframe.'+docRoot+'.scrollTop');}
[...]

PR #10902, opened 2026-02-24
fix(security): remove eval() from application JavaScript files (#2010)

craigrallen:
## Summary
- Replaced `eval()` with safer alternatives in 4 first-party JavaScript files
- Eliminates 8 `eval()` calls from application code (tree menus, fee sheet, file uploader)
- Third-party/vendor files left unchanged (should be updated upstream)

## Problem
`eval()` is a well-known security risk (arbitrary code execution) and a CSP violation. Issue #2010 tracks removing all `eval()` usage from the codebase. This PR addresses the first-party application JavaScript files.

## Changes

**`library/js/DocumentTreeMenu.js` & `library/js/CategoryTreeMenu.js`** (6 eval calls):
- Event handlers (`ontoggle`, `oncollapse`, `onexpand`) were executed via `eval(handler)`
- Replaced with type-checked approach: if handler is a function, call it directly; if it's a string, use `Function(handler)()` which creates a new function scope without access to local variables

**`interface/forms/fee_sheet/review/js/fee_sheet_core.js`** (1 eval call):
- `eval(diags_matches[i])` was used to execute `diags.push(...)` strings extracted from AJAX response
- Replaced with regex capture group extraction + `JSON.parse()` for safe argument parsing

**`interface/modules/zend_modules/public/js/scripts/file_uploader.js`** (1 eval call):
- `eval(success_function+"("+data+")")` built and executed code to call a callback by name
- Replaced with `window[success_function](data)` for safe global function lookup

## Not Changed (vendor/third-party)
- `overlib_mini.js` — legacy calendar overlay library
- `jquery.easyui.min.js` — third-party jQuery UI library
- `jquery-ui.min.js` / `jquery.ui.datepicker.js` — jQuery UI
- `telehealth.js` — webpack bundle (eval in devtool source maps)

These should be addressed by updating the upstream packages.

Refs #2010

## Testing
- Document tree and category tree menus: expand/collapse/toggle behavior preserved
- Fee sheet diagnosis refresh: `diags.push()` parsing works with JSON.parse fallback
- File uploader: callback invocation via `window[fn]()` lookup

## AI Disclosure
- **Files touched:** 4 JS files — AI-assisted

### DIFF
--- a/interface/modules/zend_modules/public/js/scripts/file_uploader.js
+++ b/interface/modules/zend_modules/public/js/scripts/file_uploader.js
@@ -241,7 +241,13 @@ function sendFileToServer(formData,status,progress_bar,success_function)
 				status.setProgress(100);
 			}
 			if (typeof window[success_function] === 'function') {
-				window[success_function](typeof data === 'string' ? JSON.parse(data) : data);
+				try {
+					window[success_function](typeof data === 'string' ? JSON.parse(data) : data);
+				} catch (error) {
+					console.warn('file_uploader: callback failed:', success_function, error);
+				}
+			} else {
+				console.warn('file_uploader: callback not found on window:', success_function);
 			}
 		}
 	});
--- a/library/js/CategoryTreeMenu.js
+++ b/library/js/CategoryTreeMenu.js
@@ -53,6 +53,45 @@ function arrayCopy(input)
 	return output;
 }
 
+function treeMenuInvokeEventHandler(handler, eventName)
+{
+	if (typeof handler === 'function') {
+		handler();
+		return;
+	}
+
+	if (typeof handler !== 'string') {
+		return;
+	}
+
+	var handlerName = handler.replace(/^\s+|\s+$/g, '');
+
+	if (handlerName.slice(-2) === '()') {
+		handlerName = handlerName.slice(0, -2);
+	}
+
+	if (!/^[A-Za-z_$][\w$]*(\.[A-Za-z_$][\w$]*)*$/.test(handlerName)) {
+		console.warn('TreeMenu: unsupported ' + eventName + ' handler format:', handler);
+		return;
+	}
+
+	var context = window;
+	var parts = handlerName.split('.');
+
+	for (var i = 0; i < parts.length; i++) {
+		if (context == null) {
+			break;
+		}
+		context = context[parts[i]];
+	}
+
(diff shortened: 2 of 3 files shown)


library/js/CategoryTreeMenu.js
```diff
@@ -314,13 +314,25 @@ function arrayCopy(input)
 				nodeObject = this.nodeRefs[layerID];
 	
 				if (nodeObject.ontoggle != null) {
-					eval(nodeObject.ontoggle);
```
kojiromike, line 317: The `Function()` constructor is an improvement over raw `eval()` (no access to local scope), but most CSP policies that block `eval()` also block `Function()` — both require `unsafe-eval`. The PR description mentions CSP as a motivation, so wanted to flag that this doesn't fully get us there for these files.

A complete CSP fix would require refactoring how these handlers are assigned — e.g., storing actual function references instead of strings. That's a bigger change and probably out of scope here, but worth noting for the issue tracker.

interface/modules/zend_modules/public/js/scripts/file_uploader.js
```diff
@@ -240,7 +240,9 @@ function sendFileToServer(formData,status,progress_bar,success_function)
 			if(progress_bar) {
 				status.setProgress(100);
 			}
-			eval(success_function+"("+data+")");
+			if (typeof window[success_function] === 'function') {
```
kojiromike, line 243: If `success_function` doesn't resolve to a function on `window`, this silently does nothing. The original `eval()` would have thrown. Should this log a warning so callers notice when their callback isn't being invoked?

```suggestion
			if (typeof window[success_function] === 'function') {
				window[success_function](typeof data === 'string' ? JSON.parse(data) : data);
			} else {
				console.warn('file_uploader: callback not found on window:', success_function);
			}
```

craigrallen:
Thanks for the review — I pushed a follow-up commit to address both observations:

- Replaced the `Function(...)` fallback in both `CategoryTreeMenu.js` and `DocumentTreeMenu.js` with a safe named-handler resolver (no `eval`/`Function`, so this aligns better with CSP goals).
- Updated `file_uploader.js` to log when `success_function` is missing, and wrapped callback invocation/parsing in `try/catch` with a warning on failure.

I kept the two tree files separate for this PR to minimize risk in legacy include paths, but they now have matching behavior.

PR #10903, opened 2026-02-24
fix(security): add CSV formula escaping to League\Csv exports (#10415)

### LINKED ISSUE #10415: feat: Enable CSV formula escaping in League\Csv exports
## Summary

When exporting data to CSV, spreadsheet applications like Excel may interpret cells starting with `=`, `+`, `-`, `@`, `\t`, or `\r` as formulas. While this is a client-side behavior rather than a server vulnerability, it's good practice to sanitize these values as a defense-in-depth measure.

## Suggestion

OpenEMR already uses `league/csv` for CSV handling. This library has built-in support for formula injection prevention via the `EscapeFormula` class:

```php
use League\Csv\EscapeFormula;
use League\Csv\Writer;

$writer = Writer::createFromString();
$writer->addFormatter(new EscapeFormula());
```

This prefixes potentially dangerous cell values with a tab character, which prevents formula interpretation while remaining invisible to users.

## References

- [League\Csv EscapeFormula documentation](https://csv.thephpleague.com/9.0/interoperability/escape-formula/)
- [OWASP CSV Injection](https://owasp.org/www-community/attacks/CSV_Injection)

## Scope

This would apply to any CSV export functionality, particularly in reports and data exports accessible to administrators.

craigrallen:
## Summary
- Added `EscapeFormula` formatter to the CDR log League\Csv Writer
- Patched `csvEscape()` to also strip leading tab (`\t`) and carriage return (`\r`) characters
- Added comprehensive PHPUnit tests for `csvEscape()` formula injection prevention

## Problem
CSV formula injection (DDE injection) allows spreadsheet applications to interpret cells starting with `=`, `+`, `-`, `@`, `\t`, or `\r` as formulas. The CDR log CSV download using League\Csv had no formula injection protection. The existing `csvEscape()` function also missed tab and carriage return as injection vectors.

## Solution
1. **ControllerLog.php**: Add `$writer->addFormatter(new EscapeFormula())` to the League\Csv Writer, which prefixes dangerous cell values with a tab character to prevent formula interpretation
2. **csvEscape()**: Extend the leading-character stripping regex from `/^[\-@]+/` to `/^[\-@\t\r]+/` to include tab and carriage return

Fixes #10415

## Testing
- Added `CsvEscapeTest.php` with data providers covering:
  - Safe values (plain text, numbers, dates)
  - Formula injection vectors (`=CMD()`, `+CMD()`, leading `-`, `@`, `\t`, `\r`, `|`, `"`)
  - Null input handling

## AI Disclosure
- **Files touched:** `ControllerLog.php`, `htmlspecialchars.inc.php`, test file — AI-assisted

### DIFF
--- a/tests/Tests/Isolated/library/CsvEscapeTest.php
+++ b/tests/Tests/Isolated/library/CsvEscapeTest.php
@@ -45,7 +45,7 @@ class CsvEscapeTest extends TestCase
             'plain text' => ['Hello World', '"Hello World"'],
             'numeric value' => ['12345', '"12345"'],
             'date with hyphen' => ['2024-01-15', '"2024-01-15"'],
-            'email in middle' => ['user@example.com', '"userexample.com"'], // leading @ stripped
+            'email in middle' => ['user@example.com', '"user@example.com"'], // @ preserved when not leading
             'empty string' => ['', '""'],
         ];
     }
@@ -95,7 +95,8 @@ class CsvEscapeTest extends TestCase
 
     public function testNullInputReturnsEmptyQuoted(): void
     {
-        // csvEscape uses null coalescing on the input
+        // csvEscape uses null coalescing on the input; intentionally testing null handling
+        /** @phpstan-ignore argument.type */
         $this->assertSame('""', csvEscape(null));
     }
 


tests/Tests/Isolated/library/CsvEscapeTest.php
```diff
@@ -0,0 +1,124 @@
...
+            'numeric value' => ['12345', '"12345"'],
+            'date with hyphen' => ['2024-01-15', '"2024-01-15"'],
+            'email in middle' => ['user@example.com', '"user@example.com"'], // @ preserved when not leading
+            'empty string' => ['', '""'],
+        ];
+    }
+
+    /**
+     * @return array<string, array{string, string}>
+     */
```
kojiromike, line 55: The `pipe removed` case expects `'"echo "test""'` — but step 1 of `csvEscape()` strips `"` globally with `preg_replace('/[=+"|]/', '', ...)`. So the inner double quotes around `test` would also be removed, producing `'"echo test"'` rather than `'"echo "test""'`.

Worth double-checking this expectation — CI is running now and should tell us.

PR #10906, opened 2026-02-24
fix(prescriptions): remove empty string DATETIME comparison for MySQL strict mode (#10834)

### LINKED ISSUE #10834: bug(prescriptions): Unable to edit/add prescriptions after 7.0.4 upgrade

craigrallen:
## Summary
- Remove `enddate = ''` from two SQL WHERE clauses in `eRx_xml.php` that compare a DATETIME column to an empty string, which is invalid under MySQL 5.7+ strict mode
- Preserve `IS NULL` and `= '0000-00-00'` checks for legacy date handling

## Problem
The `PatientMedication()` and `PatientFreeformAllergy()` functions in `eRx_xml.php` construct SQL queries with `enddate = ''` to catch legacy empty-string dates stored in the `lists.enddate` DATETIME column. Under MySQL 5.7+ strict mode, the empty string is rejected as an invalid DATETIME value before the comparison can evaluate, causing the query to error.

MariaDB diverges on this behavior and tolerates the comparison, which is why the bug wasn't caught in the standard dev environment or online demos.

## Changes
- `interface/eRx_xml.php` line 791: Remove `or enddate = ''` from `$active` filter
- `interface/eRx_xml.php` line 851: Remove `or enddate = ''` from allergy WHERE clause
- Add `tests/Tests/Isolated/eRx/ErxSqlStrictModeTest.php` — regression test

## Test plan
- [ ] Verify prescription add/edit works on MySQL 5.7+ with strict mode enabled
- [ ] Verify allergy queries still return correct results
- [ ] Run isolated tests: `composer phpunit-isolated`

## AI Disclosure
Yes — Claude Code (Anthropic) used for implementation. All AI-touched files reviewed.

Fixes #10834

### DIFF
--- a/interface/eRx_xml.php
+++ b/interface/eRx_xml.php
@@ -788,7 +788,7 @@ function PatientMedication($doc, $r, $pid, $med_limit)
     global $msg;
     $active = '';
     if ($GLOBALS['erx_upload_active'] == 1) {
-        $active = " and (enddate is null or enddate = '' or enddate = '0000-00-00' )";
+        $active = " and (enddate is null or enddate = '0000-00-00' )";
     }
 
     $res_med = sqlStatement("select * from lists where type='medication' and pid=? and title<>''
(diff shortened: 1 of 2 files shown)


stephenwaite:
thanks @craigrallen, these are most likely needed fixes for those running mysql as well but it doesn't address the issue you tagged. Mind opening another issue and updating this PR please? Thank you

stephenwaite:
oh looks like @kojiromike beat you to it [here](https://github.com/openemr/openemr/pull/10891/changes) :)

craigrallen:
Closing — this overlaps with #10891 by @kojiromike which addresses the root issue more thoroughly. Thank you @stephenwaite for pointing that out!

craigrallen:
Thanks @stephenwaite — you're right, this overlaps with @kojiromike's #10891 which addresses the root cause more thoroughly. I've closed this PR. Will open a separate issue if I find additional MySQL strict mode issues beyond what #10891 covers.

Issue #10705, opened 2026-02-11
Database: introduce and migrate to doctrine/migrations


Firehed:
This is the overarching issue to track support of doctrine/migrations tooling.

Simple version: doctrine/migrations is a well-maintained library for running and managing schema changes. It supports running migrations incrementally, tracks the status of what is and is not run, and allows you to _undo_ changes if needed. While I don't have actual data to support this, I strongly suspect it's _by far_ the most widely used database schema management tooling in the PHP ecosystem.

Importantly, adopting this makes reviewing schema changes much easier, and greatly simplifies upgrading (or potentially downgrading) active OpenEMR instances.

As a bonus it's database-agnostic, though running OpenEMR on anything that's not a MySQL variant is a distant pie-in-the-sky dream. Being able to use sqlite during automated testing would be a nice performance improvement though.

PR #10908, opened 2026-02-24
fix(ci): detect database.sql schema changes missing upgrade SQL (#10854)

### LINKED ISSUE #10854: Database Version Check CI doesn't detect database.sql changes missing upgrade SQL
The **Database Version Check** CI job (`database-version.yml`) currently only verifies that the `v_database` value matches between `version.php` and `sql/database.sql`. It does not detect when `database.sql` is modified without a corresponding upgrade SQL entry.

For example, PR #10849 changes default values in `database.sql` INSERT statements but has no upgrade SQL. In that specific case it's fine (the upgrade was already handled in 7.0.3→7.0.4), but the check would also silently pass if someone added a new column or table to `database.sql` without a matching ALTER/CREATE in the upgrade path.

This creates a risk of fresh installs and upgraded installs diverging silently.

craigrallen:
## Summary

Extends the Database Version Check CI workflow to detect when `database.sql` has schema-affecting changes (CREATE/ALTER/DROP TABLE) without a corresponding `v_database` bump and upgrade SQL file.

**Problem:** The existing CI only checks that `v_database` matches between `version.php` and `database.sql`. It does not detect when a contributor modifies `database.sql` with new columns/tables but forgets to include the upgrade SQL — causing fresh installs and upgraded installs to diverge silently.

**Solution:** Adds a second CI step that uses `git diff` against the base branch to:
1. Detect schema-affecting DDL changes in `database.sql`
2. Check if `v_database` was bumped compared to the base branch
3. Verify that at least one upgrade SQL file (`sql/*-to-*_upgrade.sql` or `sql/patch.sql`) was also modified

Three detection scenarios with clear guidance in the GitHub Actions summary:
- ⚠️ Schema changes without `v_database` bump → fail with instructions
- ⚠️ `v_database` bumped but no upgrade SQL → fail with instructions
- ✅ Both present → pass with list of upgrade files

Uses `fetch-depth: 0` for full git history needed by `git diff`.

Fixes #10854

## Testing

- [x] YAML syntax validated (pyyaml)
- [x] Actionlint config already covers this workflow (SC2016 suppression)
- [x] Checked that existing v_database alignment step is unchanged

## AI Disclosure

Yes — Claude Code (Anthropic) used for implementation. All code reviewed and tested.

### DIFF
--- a/.github/workflows/database-version.yml
+++ b/.github/workflows/database-version.yml
@@ -1,9 +1,13 @@
 # Ensures v_database is aligned between version.php and sql/database.sql
+# and that database.sql schema changes include corresponding upgrade SQL.
 #
 # The v_database value must be:
 # 1. Declared in version.php as: $v_database = NNN;
 # 2. Declared in sql/database.sql header as: -- v_database: NNN
 # 3. Both values must match
+#
+# Additionally, if database.sql has schema changes and v_database was bumped,
+# at least one upgrade SQL file should also be modified.
 #
 # This prevents database migrations from being silently skipped on upgrades.
 name: Database Version Check
@@ -33,6 +37,8 @@ jobs:
     steps:
     - name: Checkout Code
       uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
 
     - name: Setup PHP
       uses: ./.github/actions/setup-php-composer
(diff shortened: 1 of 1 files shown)


Firehed:
FYI, we're in the midst of trying to shift from the existing DB tooling to Doctrine Migrations, which will probably render this obsolete. I'm not opposed, but it may be _very_ short-lived. See #10705.

PR #10915, opened 2026-02-25
test(services): add AppointmentService test coverage

craigrallen:
## Summary

Adds a comprehensive 26-test suite for `AppointmentService`, one of the largest untested service files in the codebase (696 lines). This is part of ongoing work to improve test coverage from ~27% toward a meaningful floor.

**Tests cover:**
- **Validation** (5 tests): Success, missing required fields, title length, date format, start time format
- **Insert** (3 tests): Success with data verification, start/end time calculation, optional website field
- **Read queries** (5 tests): getAppointment, getAppointmentsForPatient (with data / empty), includes UUID/patient/facility joins
- **Status helpers** (4 tests): isCheckInStatus, isCheckOutStatus, isPendingStatus, isValidAppointmentStatus
- **Status updates** (2 tests): Success and throws for non-existent appointment
- **Calendar categories** (4 tests): getCalendarCategories, getOneCalendarCategory, contains office_visit, empty for invalid
- **Delete** (1 test): deleteAppointmentRecord
- **Data joins** (2 tests): UUID presence, patient data, facility data in results

Also adds `AppointmentFixtureManager` that manages test appointment lifecycle and installs patient/facility dependencies.

## Files Changed
- `tests/Tests/Services/AppointmentServiceTest.php` (new) — 26 tests
- `tests/Tests/Fixtures/AppointmentFixtureManager.php` (new) — Fixture manager

## Testing
- All 26 new tests pass: `docker compose exec openemr /root/devtools services-test`
- No regressions — existing test suite passes (530 tests, 2400 assertions)
- Pre-existing failure in `DuplicatePatientDetectionTest` is unrelated

## AI Disclosure
- **Was AI used?** Yes — Claude Code (Anthropic) used for implementation
- **AI-touched files:** Both new files contain AI disclosure header comments

### DIFF
--- a/tests/Tests/Fixtures/AppointmentFixtureManager.php
+++ b/tests/Tests/Fixtures/AppointmentFixtureManager.php
@@ -55,7 +55,9 @@ class AppointmentFixtureManager
             [self::FIXTURE_PREFIX . "%"]
         );
         if ($patientRow === false || !isset($patientRow['pid']) || !is_numeric($patientRow['pid'])) {
+            // @codeCoverageIgnoreStart Defensive check — only fires if test infrastructure is broken.
             throw new \RuntimeException('Failed to find test patient fixture — did installPatientFixtures() succeed?');
+            // @codeCoverageIgnoreEnd
         }
         $pid = (int) $patientRow['pid'];
 
@@ -65,7 +67,9 @@ class AppointmentFixtureManager
             [self::FIXTURE_PREFIX . "%"]
         );
         if ($facilityRow === false || !isset($facilityRow['id']) || !is_numeric($facilityRow['id'])) {
+            // @codeCoverageIgnoreStart Defensive check — only fires if test infrastructure is broken.
             throw new \RuntimeException('Failed to find test facility fixture — did installFacilityFixtures() succeed?');
+            // @codeCoverageIgnoreEnd
         }
         $facilityId = (int) $facilityRow['id'];
 
@@ -140,12 +144,14 @@ class AppointmentFixtureManager
                 "DELETE FROM openemr_postcalendar_events WHERE pc_title LIKE ?",
                 [$bindVariable]
             );
+        // @codeCoverageIgnoreStart Defensive catch — only fires on unexpected DB errors during cleanup.
         } catch (SqlQueryException $exception) {
             ServiceContainer::getLogger()->error(
                 "Failed to delete appointment fixture data",
                 ['message' => $exception->getMessage(), 'trace' => $exception->getTraceAsString()]
             );
         }
+        // @codeCoverageIgnoreEnd
 
         // Remove dependency fixtures
         if ($this->hasInstalledDependencies) {


aishanisingh5a-a11y (approved):

aishanisingh5a-a11y (approved):

tests/Tests/Fixtures/AppointmentFixtureManager.php
```diff
@@ -0,0 +1,161 @@
...
+
+namespace OpenEMR\Tests\Fixtures;
+
+use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Common\Database\SqlQueryException;
+use OpenEMR\Common\Logging\SystemLogger;
+
+class AppointmentFixtureManager
+{
+    const FIXTURE_TITLE_PREFIX = "test-fixture";
```
kojiromike: Nit: the other fixture managers use `FIXTURE_PREFIX` for this constant name (see `BaseFixtureManager::FIXTURE_PREFIX`). Would it make sense to match that naming for consistency?

```suggestion
    const FIXTURE_PREFIX = "test-fixture";
```

tests/Tests/Fixtures/AppointmentFixtureManager.php
```diff
@@ -0,0 +1,161 @@
...
+        $this->patientFixtureManager->installPatientFixtures();
+        $this->facilityFixtureManager->installFacilityFixtures();
+        $this->hasInstalledDependencies = true;
+
+        // Get the first test patient's pid
+        $patientRow = sqlQuery(
+            "SELECT pid FROM patient_data WHERE pubpid LIKE ? LIMIT 1",
+            [self::FIXTURE_TITLE_PREFIX . "%"]
+        );
+        $pid = intval($patientRow['pid']);
```
kojiromike: If `installPatientFixtures()` or `installFacilityFixtures()` fails silently (e.g., the JSON fixture file is missing or the insert is a no-op), `sqlQuery` would return `false` here, and `intval(false['pid'])` would give `0`. That would cause confusing failures downstream rather than a clear error at the source.

Would it be worth adding a guard here? Something like:

```php
if ($patientRow === false || !isset($patientRow['pid'])) {
    throw new \RuntimeException('Failed to find test patient fixture — did installPatientFixtures() succeed?');
}
```

Same thought applies to the facility query below.

tests/Tests/Fixtures/AppointmentFixtureManager.php
```diff
@@ -0,0 +1,161 @@
...
+        $this->facilityFixtureManager = $facilityFixtureManager ?? new FacilityFixtureManager();
+    }
+
+    /**
+     * Installs patient and facility fixtures that appointments depend on.
+     * Returns the pid and facility id for use in test data.
+     *
+     * @return array{pid: int, facility_id: int}
+     */
+    public function installDependencies(): array
```
kojiromike, line 44: Could this method benefit from an idempotency guard? If a test (or a subclass) calls `installDependencies()` twice, it would re-install patient and facility fixtures on top of existing ones. The `hasInstalledDependencies` flag is only checked in `removeFixtures()` but not here.

A simple early return would do:

```php
public function installDependencies(): array
{
    if ($this->hasInstalledDependencies) {
        // already installed — just re-query the IDs
    }
```

tests/Tests/Fixtures/AppointmentFixtureManager.php
craigrallen: Good call — renamed to `FIXTURE_PREFIX` for consistency. Pushed in a40694a.

tests/Tests/Fixtures/AppointmentFixtureManager.php
craigrallen: Great suggestion. Added `RuntimeException` guards for both the patient and facility queries. Pushed in a40694a.

tests/Tests/Fixtures/AppointmentFixtureManager.php
craigrallen: Done — added an early-return idempotency guard that checks `hasInstalledDependencies` before re-installing. The ID re-query still runs so the return value is always correct. Pushed in a40694a.

bradymiller:
Still some phpstan things to do. After addressed, can bring this in.

craigrallen:
Hi @bradymiller — the PHPStan and Rector issues were addressed in commit 8f3a5df (pushed Feb 25). CI is now fully green with all checks passing, including PHPStan (8.4). Let me know if there's anything else needed!

bradymiller:
hi @craigrallen , this PR was already brought in.

PR #10919, opened 2026-02-25
test(care-team): add CareTeamService unit tests

craigrallen:
## Description

Adds comprehensive PHPUnit test coverage for `CareTeamService` with **24 tests** covering all public methods.

### Tests Added

**CareTeamServiceTest** (`tests/Tests/Services/CareTeamServiceTest.php`):
- `getUuidFields` — returns expected UUID field names
- `hasActiveCareTeam` — true for active teams, false for inactive/no teams
- `saveCareTeam` — creates new teams with UUID, adds provider members with roles/facilities, updates existing teams, adds new members, marks removed members as inactive
- `getCareTeamData` — returns empty structure when no team, returns team with member details, prefers active over inactive, excludes inactive members
- `getOne` — valid UUID returns data, invalid UUID returns validation messages, respects patient UUID binding
- `getAll` — returns results, filters by patient UUID binding, validates invalid patient UUID
- `search` — returns results with providers and facilities, empty for non-matching criteria
- `createResultRecordFromDatabaseResult` — sets correct default values

**CareTeamFixtureManager** (`tests/Tests/Fixtures/CareTeamFixtureManager.php`):
- Manages patient, facility, and practitioner dependencies
- Provides test data IDs for care team operations
- Cleans up care_team_member, care_teams, and uuid_registry entries

### How Has This Been Tested?

All 24 tests pass via `docker compose exec openemr /root/devtools services-test`.
No pre-existing tests were affected.

### AI Disclosure

Yes — Claude Code (Anthropic) used for implementation. All AI-generated files include the standard disclosure comment header.

### DIFF
--- a/tests/Tests/Fixtures/CareTeamFixtureManager.php
+++ b/tests/Tests/Fixtures/CareTeamFixtureManager.php
@@ -22,7 +22,7 @@ use OpenEMR\Common\Logging\SystemLogger;
 
 class CareTeamFixtureManager
 {
-    const FIXTURE_TEAM_NAME_PREFIX = "test-fixture";
+    const FIXTURE_PREFIX = "test-fixture";
 
     /**
      * @var FixtureManager
@@ -61,30 +61,41 @@ class CareTeamFixtureManager
      */
     public function installDependencies(): array
     {
-        $this->patientFixtureManager->installPatientFixtures();
-        $this->facilityFixtureManager->installFacilityFixtures();
-        $this->practitionerFixtureManager->installPractitionerFixtures();
-        $this->hasInstalledDependencies = true;
+        if (!$this->hasInstalledDependencies) {
+            $this->patientFixtureManager->installPatientFixtures();
+            $this->facilityFixtureManager->installFacilityFixtures();
+            $this->practitionerFixtureManager->installPractitionerFixtures();
+            $this->hasInstalledDependencies = true;
+        }
 
         // Get the first test patient's pid
         $patientRow = sqlQuery(
             "SELECT pid FROM patient_data WHERE pubpid LIKE ? LIMIT 1",
-            [self::FIXTURE_TEAM_NAME_PREFIX . "%"]
+            [self::FIXTURE_PREFIX . "%"]
         );
+        if ($patientRow === false || !isset($patientRow['pid'])) {
+            throw new \RuntimeException('Failed to find test patient fixture — did installPatientFixtures() succeed?');
+        }
         $pid = intval($patientRow['pid']);
 
         // Get the first test facility's id
         $facilityRow = sqlQuery(
             "SELECT id FROM facility WHERE name LIKE ? LIMIT 1",
-            [self::FIXTURE_TEAM_NAME_PREFIX . "%"]
+            [self::FIXTURE_PREFIX . "%"]
         );
+        if ($facilityRow === false || !isset($facilityRow['id'])) {
+            throw new \RuntimeException('Failed to find test facility fixture — did installFacilityFixtures() succeed?');
+        }
         $facilityId = intval($facilityRow['id']);
 
         // Get the first test practitioner's id
         $providerRow = sqlQuery(
             "SELECT id FROM users WHERE fname LIKE ? LIMIT 1",
-            [self::FIXTURE_TEAM_NAME_PREFIX . "%"]
+            [self::FIXTURE_PREFIX . "%"]
         );
+        if ($providerRow === false || !isset($providerRow['id'])) {
+            throw new \RuntimeException('Failed to find test practitioner fixture — did installPractitionerFixtures() succeed?');
+        }
         $providerId = intval($providerRow['id']);
 
         return ['pid' => $pid, 'facility_id' => $facilityId, 'provider_id' => $providerId];
@@ -95,7 +106,7 @@ class CareTeamFixtureManager
      */
     public function removeFixtures(): void
     {
-        $bindVariable = self::FIXTURE_TEAM_NAME_PREFIX . "%";
+        $bindVariable = self::FIXTURE_PREFIX . "%";
 
         try {
             // Get care team IDs for our test teams
--- a/tests/Tests/Fixtures/CareTeamFixtureManager.php
+++ b/tests/Tests/Fixtures/CareTeamFixtureManager.php
@@ -25,33 +25,15 @@ class CareTeamFixtureManager
     const FIXTURE_PREFIX = "test-fixture";
 
     /**
-     * @var FixtureManager
-     */
-    private $patientFixtureManager;
-
-    /**
-     * @var FacilityFixtureManager
-     */
-    private $facilityFixtureManager;
-
-    /**
-     * @var PractitionerFixtureManager
-     */
-    private $practitionerFixtureManager;
-
-    /**
      * @var bool
      */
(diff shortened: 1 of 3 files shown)


tests/Tests/Fixtures/CareTeamFixtureManager.php
```diff
@@ -0,0 +1,139 @@
...
+use OpenEMR\Common\Logging\SystemLogger;
+
+class CareTeamFixtureManager
+{
+    const FIXTURE_PREFIX = "test-fixture";
+
+    /**
+     * @var bool
+     */
+    private $hasInstalledDependencies = false;
```
kojiromike: New code should use native types. Please add the `bool` type declaration here.

```suggestion
    private bool $hasInstalledDependencies = false;
```

tests/Tests/Fixtures/CareTeamFixtureManager.php
```diff
@@ -0,0 +1,139 @@
...
+     * @return array{pid: int, facility_id: int, provider_id: int}
+     */
+    public function installDependencies(): array
+    {
+        if (!$this->hasInstalledDependencies) {
+            $this->patientFixtureManager->installPatientFixtures();
+            $this->facilityFixtureManager->installFacilityFixtures();
+            $this->practitionerFixtureManager->installPractitionerFixtures();
+            $this->hasInstalledDependencies = true;
+        }
```
kojiromike, line 53: The idempotency guard wraps the install calls but not the queries below, so `installDependencies()` re-queries the database on every call even when `hasInstalledDependencies` is already true. Worth caching the return value:

```php
private array $cachedDependencies = [];

public function installDependencies(): array
{
    if ($this->hasInstalledDependencies) {
        return $this->cachedDependencies;
    }
    // ... install fixtures ...
    $this->hasInstalledDependencies = true;
    // ... query IDs ...
    $this->cachedDependencies = ['pid' => $pid, ...];
    return $this->cachedDependencies;
}
```

tests/Tests/Services/CareTeamServiceTest.php
```diff
@@ -0,0 +1,801 @@
...
+use OpenEMR\Tests\Fixtures\CareTeamFixtureManager;
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Attributes\Test;
+
+class CareTeamServiceTest extends TestCase
+{
+    /**
+     * @var CareTeamService
+     */
+    private $service;
```
kojiromike: New code should use native type declarations, not docblock-only types. Please add types to `$service` and `$fixtureManager` to match the other properties in this class.

```suggestion
    private CareTeamService $service;
```

tests/Tests/Services/CareTeamServiceTest.php
```diff
@@ -0,0 +1,801 @@
...
+    {
+        $this->service = new CareTeamService();
+        $this->fixtureManager = new CareTeamFixtureManager();
+        $deps = $this->fixtureManager->installDependencies();
+        $this->testPid = $deps['pid'];
+        $this->testFacilityId = $deps['facility_id'];
+        $this->testProviderId = $deps['provider_id'];
+
+        // saveCareTeam reads $_SESSION['authUserID']
+        if (!isset($_SESSION['authUserID'])) {
```
kojiromike: Two things here:

1. **Session abstraction:** OpenEMR has a `SessionWrapperFactory` / `SessionWrapperInterface` layer for session access. New code should use `SessionWrapperFactory::getInstance()->getWrapper()->get('authUserID')` instead of raw `$_SESSION`. See `PatientService::databaseInsert()` for the pattern. (The underlying `CareTeamService` itself uses raw `$_SESSION` too — that's preexisting, but the test shouldn't further entrench the legacy pattern.)

2. **Test isolation:** Other tests in the codebase (e.g., `FhirObservationEmployerServiceTest`) back up and restore `$_SESSION` in setUp/tearDown. The conditional-set here means if another test class runs first and sets a different `authUserID`, this test silently inherits that value. Please back up the session in setUp and restore it in tearDown:

```php
$this->backupSession = $_SESSION;
// ... in tearDown:
$_SESSION = $this->backupSession;
```

tests/Tests/Services/CareTeamServiceTest.php
```diff
@@ -0,0 +1,801 @@
...
+
+        // Verify initial state: one active member
+        $initialCount = QueryUtils::querySingleRow(
+            "SELECT COUNT(*) as cnt FROM care_team_member WHERE care_team_id = ? AND status = 'active'",
+            [$teamId]
+        );
+        $this->assertEquals(1, (int) $initialCount['cnt']);
+
+        // Get a second provider id — the admin user (id=1) should always exist
+        $secondProvider = QueryUtils::querySingleRow(
```
kojiromike: This queries for a real non-fixture user (`id != ? AND id > 0 AND username IS NOT NULL`), which couples the test to the database having at least two qualifying users. Please use the practitioner fixture manager to create a second test provider instead, so the test stays self-contained regardless of database state.

tests/Tests/Services/CareTeamServiceTest.php
```diff
@@ -0,0 +1,801 @@
...
+        $this->assertEquals('active', $result['care_team_status']);
+        $this->assertNull($result['date']);
+        $this->assertIsArray($result['providers']);
+        $this->assertEmpty($result['providers']);
+        $this->assertIsArray($result['facilities']);
+        $this->assertEmpty($result['facilities']);
+        $this->assertIsArray($result['contacts']);
+        $this->assertEmpty($result['contacts']);
+    }
+}
```
kojiromike, line 939: The test suite covers user-type members thoroughly but doesn't exercise contact-type members at all. `CareTeamService::saveCareTeam()` has a separate code path for `contact_id` members (lines 421–433 of the service), and `getCareTeamData()` has distinct logic for contact members too. Is that intentional, or something you're planning for a follow-up? Even one test that saves a team with a contact member would cover a meaningful gap.

kojiromike (changes requested):
Solid test suite for CareTeamService — 24 tests covering the key operations is a nice contribution. A few things below, mostly around types, session access, test isolation, and coverage completeness.

craigrallen:
@kojiromike I rebased this onto current master and addressed the review threads: native types, cached fixture dependencies, typed properties, SessionWrapperFactory usage with session restoration, fixture-only second provider selection, and contact-member coverage. Targeted PHPStan and PHPCS pass; the database-backed suite is running in CI. Re-review when convenient.

tests/Tests/Fixtures/CareTeamFixtureManager.php
```diff
@@ -0,0 +1,149 @@
...
+ * and tested by the contributor.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Craig Allen <craigrallen@gmail.com>
+ * @copyright Copyright (c) 2026 Craig Allen <craigrallen@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Tests\Fixtures;
```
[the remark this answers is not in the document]
craigrallen: Fixed in fbce3bf0e: both new PHP files now declare strict_types=1 after their file headers and before the namespace. Targeted PHPStan and PHPCS pass.

PR #10923, opened 2026-02-26
fix(lab): parse compound HL7 placer order numbers instead of truncating with intval

### LINKED ISSUE #7762: Lab results being marked as orphan when valid control_id is returned
### Describe the bug

Lab results are being displayed as Orphan files when there is a correct control_id in the result set.
image

### To Reproduce
Changing the code fixes the orphan file issue
image

### Client configuration

- Browser: Any
- OpenEMR version: 7.0.2(1)
- Operating system: ubuntu 20.04

### LINKED ISSUE #8130: Lab results order ID issue
### Describe the bug

When lab results are being parsed. The order ID is in the format of 11545596-0175. The number after the dash is the actual order ID in the procedure order table. The intval() is cutting off the information after the dash. See image below.
 

Image

### To Reproduce
Import a lab result. No all labs but the information in the ORC[2] segment. Quest does. 

### Expected behavior
not to cut off the dash 

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

jbaiad:
## Summary

Fixes #8130
Fixes #7762

When a lab vendor returns a compound placer order number like `310137493-L700107` in OBR-2, the existing code uses `intval()` to extract the numeric procedure order ID. `intval()` silently truncates at the first non-numeric character (the dash), returning just `310137493`. This causes result-to-order matching failures when the numeric prefix alone doesn't match any `procedure_order_id` in the database.

## Changes

- Replace `intval($in)` with a custom parser `rhl7OrderId()` that:
  - Returns `intval($in)` for plain numeric strings (backward compatible)
  - Extracts the numeric portion before the first dash for compound IDs like `310137493-L700107`
  - Returns `0` for non-parseable inputs
- Add 9 isolated unit tests covering numeric, zero-padded, compound, multi-dash, and edge cases
- Add 5 real HL7 ORU^R01 fixture files from the OpenEMR wiki (Hl7_sample.zip) for future integration testing

## Test Plan

- [x] Run `composer phpunit-isolated` — all isolated tests pass, no regressions
- [ ] Integration test with real HL7 messages containing compound placer numbers


interface/orders/receive_hl7_results.inc.php
```diff
@@ -973,7 +973,11 @@ function receive_hl7_results(&$hl7, &$matchreq, $lab_id = 0, $direction = 'B', $
             $porow = false;
             $pcrow = false;
             if ($direction != 'R' && $a[2]) {
-                $in_orderid = intval($a[2]);
+                // Extract order ID from compound placer numbers like "11545596-0175"
+                // or "ACME-0042" where the ID follows the last dash (#8130, #7762).
+                $in_orderid = str_contains($a[2], '-')
```
kojiromike, line 978: This expression is duplicated in the OBR handler below (around line 992). Would it make sense to extract it to a named function? The PR description mentions `rhl7OrderId()` as the intended name, and the test already has a `extractOrderId()` that reimplements this logic — if there's a real function, the test could call it directly instead of duplicating the expression.

Something like:

```suggestion
                $in_orderid = rhl7OrderId($a[2]);
```

With a function defined earlier in the file:

```php
function rhl7OrderId(string $s): int
{
    return str_contains($s, '-')
        ? intval(substr($s, strrpos($s, '-') + 1))
        : intval($s);
}
```

tests/Tests/Isolated/Orders/Rhl7OrderIdTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+use PHPUnit\Framework\TestCase;
+
+class Rhl7OrderIdTest extends TestCase
+{
+    /**
+     * Replicate the inline expression used in receive_hl7_results.inc.php
+     * for extracting procedure_order_id from compound placer order numbers.
+     */
+    private function extractOrderId(string $s): int
```
kojiromike, line 36: If the logic gets extracted to a `rhl7OrderId()` function as suggested on the production code, this test could call the real function instead of reimplementing the expression. That way the test validates the actual code path rather than a copy of it.

kojiromike (commented):
The fix correctly addresses the compound placer order number format generated by `gen_hl7_order.inc.php` (line 130: `{send_fac_id}-{zero_padded_order_id}`). The approach of taking the part after the last dash and passing it through `intval()` matches the outbound format.

The main issue is code duplication — the same expression appears in two places in the production code, and the test reimplements it a third time. The PR description mentions a `rhl7OrderId()` function that was planned but never created. Extracting the logic to a named function would eliminate all three copies and let the test call the real code.

The HL7 fixture files are a nice addition for future integration testing.

A few notes on the PR description:
- It mentions `rhl7OrderId()` as a custom parser function, but the diff inlines the expression instead
- Minor: the description says "9 isolated unit tests" but the test file has 8 test methods

Issue #10910, opened 2026-02-24
refactor(orders): improve lab ordering testability and fix HL7 parsing gaps


jbaiad:
## Summary

The lab ordering system has a modern, FHIR-compliant service layer and database schema, but the HL7 generation/parsing and UI layers remain monolithic include files with raw SQL, making them effectively untestable. There is **zero unit test coverage** for the core procedure ordering business logic, HL7 generation, and HL7 result parsing.

This issue proposes an incremental refactoring path to make the system testable, fix known bugs, and enable FHIR write operations — without a big-bang rewrite.

### What's good

- **Database schema**: Well-normalized, 10 tables with UUID support, recent FHIR additions (procedure_order_type, order_intent, specimen tracking, relationship linking) align with US Core 8.0 / USCDI v5
- **Service layer**: `ProcedureService` (1,479 lines), `ProcedureOrderRelationshipService`, `ProcedureProviderService` use modern patterns (`BaseService`, `UuidRegistry`, `ProcessingResult`)
- **FHIR mappings**: ServiceRequest, DiagnosticReport, Observation resources are implemented (read-only)

### What needs work

| Area | Issue |
|------|-------|
| **HL7 generation** | `gen_hl7_order.inc.php` (~2000 lines) — monolithic include file with raw SQL and globals |
| **HL7 parsing** | `receive_hl7_results.inc.php` (1,946 lines) — same pattern, known bugs (#8130, #7762) |
| **Validation** | No `ProcedureOrderValidator` — data goes straight from form POST to SQL |
| **Test coverage** | Zero unit/isolated tests for `ProcedureService`, HL7 gen/parse, order CRUD |
| **FHIR writes** | ServiceRequest is read-only — no programmatic order creation |
| **UI layer** | `common.php` (~1000+ lines) mixes init, rendering, and JavaScript; 107 raw SQL calls and 48 `global $GLOBALS` references in `interface/orders/` |

### Related open issues

- #8130 — `intval()` truncates control_id containing dashes (e.g., `11545596-0175` → `11545596`), breaking result matching
- #7762 — Valid lab results incorrectly flagged as orphaned due to control ID validation logic
- #10453 — DORN AOE questions not persisting during order creation

### 1. Add `ProcedureOrderValidator`

A validator following the existing `PatientValidator` pattern:
- Required fields: `patient_id`, `provider_id`, `encounter_id`, `lab_id`
- At least one `procedure_order_code` entry
[...]

PR #10924, opened 2026-02-26
refactor(lab): extract HL7 order and result processing into strategy patterns

jbaiad:
## Summary

Refs #10910

Extracts HL7 order generation and result parsing from monolithic procedural functions into clean OOP strategy patterns, making the code testable, extensible, and easier to maintain.

## Changes

**Order Generation (Strategy Pattern):**
- Add `Hl7OrderGeneratorInterface` with `generateOrder()` contract
- Add `Hl7OrderGeneratorFactory` to select generators by lab type
- Extract `DefaultHl7OrderGenerator`, `LabCorpHl7OrderGenerator`, `QuestHl7OrderGenerator`, `UniversalHl7OrderGenerator` from inline switch/case logic in `common.php`
- Update `common.php` to delegate to factory-created generators

**Result Parsing (Wrapper Pattern):**
- Add `Hl7ResultParserInterface` with `receiveResults()` contract
- Add `DefaultHl7ResultParser` wrapping the procedural `receive_hl7_results()` function
- Add `Hl7ResultParseResult` value object and `Hl7ResultParseException`
- Update `list_reports.php` to use the new parser interface

**Tests:**
- 7 isolated unit tests for factory, interface contracts, and value objects
- 5 order generator integration tests (default, universal, compound placer, demographics, error handling)
- 18 result parser integration tests (dryrun, full write-path, malformed messages, NTE comments, multi-OBR, status/flag mapping, specimen, facility, data types)
- 5 HL7 fixture files from OpenEMR wiki

## Test Plan

- [x] Run `composer phpunit-isolated` — all tests pass, no regressions


src/Common/Orders/Hl7OrderGeneratorFactory.php
```diff
@@ -0,0 +1,80 @@
...
+    ];
+
+    /**
+     * Create the appropriate HL7 order generator for the given lab type.
+     *
+     * @param string $labType       Lab identifier (e.g., 'labcorp', 'quest', 'ammon', 'clarity')
+     * @param string $interfaceDir  Absolute path to the interface/ directory
+     * @return Hl7OrderGeneratorInterface
+     */
+    public static function create(string $labType, string $interfaceDir): Hl7OrderGeneratorInterface
```
kojiromike, line 60: Should `create()` be an instance method on a `Hl7OrderGeneratorFactory` instance (or a method on a service registered in the container)?

A static method here means the call sites in `common.php` and `list_reports.php` can't substitute the factory in tests or in a future plugin/extension scenario. CLAUDE.md's "Authorization Modeling / Service Wiring" sections recommend constructor injection over static access for service-layer objects — once a static is in the call graph, the dependency is invisible from the constructor signature.

Happy to keep the static if there's a reason (e.g., this needs to be callable from a context where DI isn't available yet) — would be useful to note that in the docblock.

src/Common/Orders/DefaultHl7ResultParser.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function receiveResults(
+        string &$hl7,
+        array &$matchReq,
+        int $labId = 0,
+        string $direction = 'B',
+        bool $dryRun = false,
+        ?array $matchResp = null
+    ): Hl7ResultParseResult {
+        // @phpstan-ignore argument.type, argument.type, parameterByRef.type (procedural function PHPDoc is mistyped)
```
kojiromike, line 28: The triple `@phpstan-ignore` is pointing at a real defect in the underlying PHPDoc, not a quirk of this adapter. Could you fix the docblock on `receive_hl7_results()` (and the `parameterByRef` annotation) at the source instead?

Fix-at-the-source means future adapters / direct callers of `receive_hl7_results()` get correct types automatically, and this class doesn't have to carry a permanent ignore. If the procedural function is genuinely doing something the type system can't express, it'd be useful to spell that out in the comment so reviewers don't have to guess.

interface/orders/list_reports.php
```diff
@@ -244,7 +246,9 @@ function doWait(e){
     }
     // Attempt to post any incoming results.
     if (!empty($_REQUEST['form_process_labs']) || (!empty($info['orphaned_order']) && $info['orphaned_order'] == "R")) {
-        $errmsg = poll_hl7_results($info, $processing_lab);
+        $parser = new DefaultHl7ResultParser();
```
kojiromike, line 249: `new DefaultHl7ResultParser()` instantiates the service directly inside the page handler. Same DI concern as the factory — once this is `new`'d here, swapping in a fake or alternate parser requires editing the file rather than wiring a dependency.

For procedural top-of-page code this is admittedly the path of least resistance, but it's worth at least making this an injectable point (e.g., resolve from a container, or accept a parameter) so the parser interface earns its keep.

interface/orders/list_reports.php
```diff
@@ -244,7 +246,9 @@ function doWait(e){
...
+        $labId = is_numeric($processing_lab) ? (int) $processing_lab : 0;
```
kojiromike, line 250: Minor: `(int) (is_numeric($processing_lab) ? $processing_lab : 0)` and the similar `(int) (is_numeric($formid) ? $formid : 0)` upstream are unusual. PHP's `(int)` cast on a non-numeric string already yields `0`, so `(int) $processing_lab` is equivalent unless you specifically want to short-circuit the cast on, e.g., float-shaped strings.

If the goal is to reject non-integer-shaped values, `ctype_digit((string) $processing_lab) ? (int) $processing_lab : 0` is more explicit. If the goal is just "coerce to int, default 0," `(int) $processing_lab` does it.

interface/forms/procedure_order/common.php
```diff
@@ -384,12 +360,7 @@ function normalizeDirectoryName(string $input): string
                             }
                         }
                     } else {
-                        $alertmsg = match ($gbl_lab) {
-                            'labcorp' => labcorp_send_hl7_order($ppid, $hl7),
-                            'quest' => quest_send_hl7_order($ppid, $hl7),
-                            'ammon', 'clarity' => universal_send_hl7_order($ppid, $hl7),
-                            default => default_send_hl7_order($ppid, $hl7),
-                        };
+                        $alertmsg = $generator->send($ppid, (string) $hl7);
```
kojiromike, line 363: The `(string) $hl7` cast here suggests `Hl7OrderResult::hl7` is nullable (or untyped). If so, narrowing the field once on the value object — `public readonly string $hl7` with the null-coalesce handled at construction — beats every caller doing a defensive cast. This was already a baseline entry, so this PR is the right moment to push the narrowing upstream rather than leaving the cast at the call site.

kojiromike:
@aisle-analyzer

kojiromike (commented):
Solid refactor — replacing the `match($gbl_lab)` dispatch and the `$labConfigs` require_once block with a strategy/factory pair is a clear win, and the PHPStan baseline shrinks rather than growing (5 entries removed, none added). Tests and fixtures look thorough. A few questions before merge:

**Static factory vs. constructor injection.** `Hl7OrderGeneratorFactory::create()` is static, and `list_reports.php` uses `new DefaultHl7ResultParser()` directly. CLAUDE.md's "Inject all dependencies through the constructor. Never use `new` for service-layer objects inside business logic" applies here — once the factory is static, callers and tests can't substitute. Given this is a transitional refactor over procedural includes, an instance factory (or a service registered in the container) would be a small additional step and would let the result-parser branch be exercised in integration tests without bootstrapping the whole include chain.

**Leaky abstraction on `receiveResults()`.** The interface returns `Hl7ResultParseResult` *and* keeps `string &$hl7` and `array &$matchReq` as by-reference. State now flows through both channels, which is harder to reason about than the legacy array. Worth a follow-up to migrate the by-ref state into the value object (or a dedicated context object) so the contract is single-channel.

**`@phpstan-ignore` in the parser.** `DefaultHl7ResultParser::receiveResults` carries a triple-ignore because the procedural function's PHPDoc is wrong. Per CLAUDE.md "fix-at-the-source typing," the cleaner move is to correct the PHPDoc on `receive_hl7_results()` itself rather than ignoring the violation at every adapter.

**`(string) $hl7` cast at the call site (common.php).** The cast suggests `Hl7OrderResult::hl7` is `?string` (or untyped). If the source field is nullable, narrow it once on the value object — every caller doing the same defensive cast is a smell.

Specific inline comments below.

PR #10926, opened 2026-02-26
feat(fhir): add FHIR ServiceRequest write operations (POST/PUT)

jbaiad:
## Summary

Refs #10910
Depends on #10925

Implements FHIR R4 ServiceRequest write support for procedure orders, enabling external systems to create and update lab/imaging orders via the FHIR API.

**Note:** This PR depends on #10925 (ProcedureOrderValidator + ProcedureService write ops) and should be merged after it.

## Changes

**FHIR ServiceRequest Service:**
- Add `parseFhirResource()` with reverse mappings (FHIR status/category/priority → OpenEMR enums)
- Add `insertOpenEMRRecord()` and `updateOpenEMRRecord()` delegating to `ProcedureService`
- Fix `date_collected` undefined key in `parseOpenEMRRecord()`

**Serialization:**
- Add `FhirServiceRequestSerializer` for FHIR → OpenEMR deserialization
- Support subject, requester, performer, category, priority, status, notes, reason codes

**REST Controller & Routes:**
- Register POST/PUT routes in FHIR R4 US Core routing
- Add `setSystemLogger()` for test observability
- Add `readonly` properties (Rector compliance)

**UtilsService:**
- Fix `getUuidFromReference()` to accept nullable parameter

**Tests:**
- 4 isolated test files: serializer, reverse mapping, contract, UtilsService (32 tests)
- 1 REST controller test with response handler traits
- 1 service CRUD integration test with round-trip verification

**CI Fixes:**
- Update PHPStan baselines (remove 17 stale entries)
- Add procedural HL7 function names to composer-require-checker allowlist

## Test Plan

- [x] Run `composer phpunit-isolated` — all 1690 tests pass, no regressions

### DIFF
--- a/.composer-require-checker.json
+++ b/.composer-require-checker.json
@@ -92,8 +92,22 @@
     "sqlStatement",
     "sqlStatementNoLog",
     "sqlStatementThrowException",
+    "default_gen_hl7_order",
+    "default_loadPayerInfo",
+    "default_send_hl7_order",
+    "labcorp_gen_hl7_order",
+    "labcorp_loadPayerInfo",
+    "labcorp_send_hl7_order",
+    "poll_hl7_results",
+    "quest_gen_hl7_order",
+    "quest_loadPayerInfo",
+    "quest_send_hl7_order",
+    "receive_hl7_results",
     "ub04Dispose",
-    "ub04_dispose"
+    "ub04_dispose",
+    "universal_gen_hl7_order",
+    "universal_loadPayerInfo",
+    "universal_send_hl7_order"
   ],
   "php-core-extensions": [
     "Core",
--- a/.phpstan/baseline/binaryOp.invalid.php
+++ b/.phpstan/baseline/binaryOp.invalid.php
@@ -19902,11 +19902,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../src/RestControllers/FHIR/FhirServiceRequestRestController.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Binary operation "\\." between non\\-falsy\\-string and mixed results in an error\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../src/RestControllers/FHIR/FhirServiceRequestRestController.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Binary operation "\\." between mixed and mixed results in an error\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../src/RestControllers/FHIR/FhirSpecimenRestController.php',
--- a/.phpstan/baseline/booleanAnd.rightAlwaysTrue.php
+++ b/.phpstan/baseline/booleanAnd.rightAlwaysTrue.php
@@ -4,6 +4,11 @@ $ignoreErrors = [];
 $ignoreErrors[] = [
     'message' => '#^Right side of && is always true\\.$#',
     'count' => 1,
+    'path' => __DIR__ . '/../../custom/code_types.inc.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Right side of && is always true\\.$#',
+    'count' => 1,
     'path' => __DIR__ . '/../../interface/forms/vitals/C_FormVitals.class.php',
 ];
 $ignoreErrors[] = [
--- a/.phpstan/baseline/booleanNot.alwaysFalse.php
+++ b/.phpstan/baseline/booleanNot.alwaysFalse.php
@@ -123,8 +123,13 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Negated boolean expression is always false\\.$#',
-    'count' => 1,
+    'count' => 2,
     'path' => __DIR__ . '/../../library/sql.inc.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Negated boolean expression is always false\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/sqlconf.php',
 ];
 $ignoreErrors[] = [
     'message' => '#^Negated boolean expression is always false\\.$#',
--- a/.phpstan/baseline/class.notFound.php
+++ b/.phpstan/baseline/class.notFound.php
@@ -2052,11 +2052,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../src/RestControllers/FHIR/FhirProvenanceRestController.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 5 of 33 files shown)


kojiromike:
@aisle-analyzer

kojiromike (commented):
Thanks for the work here — the test coverage on the new code is strong (94%+ patch), the validator + serializer split is clean, and the reverse-mapping table in the test fixtures is the right shape. A few concerns before merge. (The aisle bot has separately flagged cross-patient IDOR, log forging via UUID, FHIR Reference parsing, and exception-message disclosure — not duplicating those here.)

### Scope

This PR contains all of #10925's `ProcedureService` insert/update + `ProcedureOrderValidator` content **plus** a FHIR write layer on top. Reviewing the FHIR layer in isolation is hard when half the diff is the same code under review on the prior PR. Could you either rebase this on #10925's branch (so the diff shows only the FHIR delta), or drop the duplicated content and merge #10925 first? Same questions raised on #10925 (audit logging, transaction boundary, authorization gate) apply to the copies here.

Issue #10910 also calls for "create order + codes + specimens in a single transaction" — the insert path here only writes `procedure_order` and never touches `procedure_order_code` or `procedure_specimen`. Is that follow-up work or did the description over-promise?

### Missing audit logging (HIPAA)

Neither `ProcedureService::insert/update` nor `FhirServiceRequestService::insertOpenEMRRecord/updateOpenEMRRecord` emit audit events. Clinical orders are PHI; HIPAA §164.312(b) requires recording who created/modified them. ONC certification 170.315(d)(2) is more specific. The existing `EventAuditLogger` is the in-repo convention — see how `EncounterService::insert` and `PatientService::insert` instrument audit. Please thread through here too.

### REST semantics

A few REST/FHIR-spec gaps in the controller and route layer:

- **`patch()` is named for PUT semantics.** FHIR R4 §3.1.0.5 distinguishes PUT (full replace) from PATCH (partial). The route `PUT /fhir/ServiceRequest/:uuid` is wired to `patch()`, but the body of the method does a full replace. Either rename to `put()`/`update()` (preferred — matches what it does), or implement true PATCH semantics.
- **No `If-Match` / ETag on PUT.** Concurrent updates to the same order will silently last-write-wins. FHIR §3.1.0.5.1 expects `If-Match` for optimistic concurrency on PUT. Without it, two clinicians editing the same order race.
- **POST-with-id is silently accepted.** A client POSTing a body with an `id` field gets that id ignored (or worse, used) without an OperationOutcome. FHIR servers should reject this with 400.
- **PUT `:uuid` vs body `id` mismatch is unchecked.** If they disagree, FHIR requires 400 — we just use the path param.
- **Malformed JSON becomes `[]`, not 400.** The route uses `(array)(json_decode((string)file_get_contents("php://input"), true))`. Invalid JSON silently becomes `[]` and validation runs against an empty doc. Use `json_decode(..., flags: JSON_THROW_ON_ERROR)` and return 400 on `JsonException`.

### Reverse-mapping silent defaults

`FhirServiceRequestService` reverse-maps `status`, `category`, `priority`, and `intent` from FHIR codes to OpenEMR codes. Unknown values silently fall through to a default (e.g., unknown priority → `routine`). For a clinical order, an unknown `stat` mistyped as `staat` should return 400, not be quietly downgraded to `routine`. Same for status: a typo'd status shouldn't silently become `pending`. Make the reverse map exhaustive and reject unknowns.

Related: collapsing `entered-in-error` to `canceled` loses information (FHIR distinguishes these for audit/error correction). At minimum document the policy; better, model the OpenEMR side to preserve it.

### Constructor injection

CLAUDE.md "Inject all dependencies through the constructor. Never use `new` for service-layer objects inside business logic":

- `FhirServiceRequestRestController::setSystemLogger()` is setter injection — should be a constructor parameter so callers can't forget it.
- `parseFhirResource()` calls `new CodeTypesService()` (and similar) inline. A `CodeTypesService` instance per parsed resource is wasteful, but more importantly it makes the parser untestable without a real codes table. Pass it in via the constructor.
- `FhirValidationService` is `new`'d in the controller constructor — fine for now but worth threading through if we ever want to swap the validator in tests.

### New file convention slip

`src/Services/FHIR/Serialization/FhirServiceRequestSerializer.php` is missing `declare(strict_types=1);`. Other new files in the FHIR namespace declare it (see `FhirEncounterSerializer`, etc.). Without strict types, the `(string)` casts inside `parseFhirResource` will silently coerce `null`/objects.

### Narrow, don't cast

`parseFhirResource()` in `FhirServiceRequestService.php` is full of `(string) $resource->getStatus()` and similar. PHPStan level 10 flags these as `cast.string` (an object's `__toString` may or may not exist, and casting `null` to string gives `''`). The fhir-php library returns nullable typed objects — narrow with type guards instead:

```php
$statusObj = $resource->getStatus();
if ($statusObj === null) {
    throw new InvalidArgumentException('ServiceRequest.status is required');
}
$status = $statusObj->getValue();  // typed string
```

This also catches the silent-default class above.

### Validator details

`src/Validators/ProcedureOrderValidator.php`:

- `if ($existingUuid != null)` — use `!== null`. Loose comparison treats `0`, `''`, and `'0'` as null-equivalent.
- `clinical_hx` validated as `lengthBetween(1, 255)` — FHIR `note.text` is unbounded `markdown`. Round-tripping a clinical note over 255 chars (which is short — a paragraph) will fail validation. Either widen or document the truncation policy.
- The `ORDER_STATUSES` constant is a hand-maintained mirror of the FHIR `request-status` value set. When FHIR R5 ships or we add a status, this drifts silently. Consider deriving from the FHIR enum.

### Authorization

The route layer only checks `request_authorization_check($request, "patients", "med")`. There's no SMART-on-FHIR scope check (`user/ServiceRequest.write` vs `patient/ServiceRequest.write`) and no facility-scope filtering on update. Aisle's IDOR finding above is the most acute manifestation; the broader pattern is that write operations need typed authorization, not a single ACL string. CLAUDE.md "Authorization Modeling" recommends typing the principal — at minimum, document where the SMART scope check is expected to live.

### Smaller things

- `(new SystemLogger())->errorLogCaller(...)` instantiates the logger per call. Inject it.
- `catch (\Throwable $e)` in `resolveUuidToId()` is broad enough to swallow programmer errors. Narrow to `InvalidArgumentException` or `RuntimeException`.
- Double-quoted strings without interpolation throughout (`"patient_data"`, `"pid"`, etc.) — single-quote per PER-CS.
- `!empty(...)` shows up on edited lines; CLAUDE.md "Avoid `empty()`" — use explicit checks.
- `FhirServiceRequestRestControllerTest` instantiates the controller directly and calls methods on it, bypassing the route layer. Since the route file (`apis/routes/_rest_routes_fhir_r4_us_core_3_1_0.inc.php`) has 35% patch coverage, the route handler — including the JSON decode and auth check — is the part that most needs an integration test.

### Suggested next steps

Given the size of the diff and the dependency on #10925, I'd suggest:

1. Land #10925 first.
2. Rebase this PR; re-scope to the FHIR write layer only.
3. Add audit logging and SMART scope check (or document where they'll live).
4. Address the REST semantics (PUT method name, malformed-JSON handling, `If-Match`).
5. Tighten reverse-map to reject unknowns.

Happy to re-review once the scope is narrowed.

PR #10927, opened 2026-02-26
chore(db): Remove multipledb module

Firehed:
Towards #10149 and #10774.

This removes some high-complexity code for running multiple database connections, limited solely to Zend modules. This code blocks the removal of Laminas, and seems to be completely unused by any path in reality. When I tried to install+enable+configure it anyways, the actual UI to continue setup crashed, so I have strong reason to believe nobody on a current version of OpenEMR is relying on this.

It also has our usual fun problem of needing the database to configure the database, further solidifying the likelihood that it really _can't_ function today.

The most I can find is some vague commentary [from 2019](https://community.open-emr.org/t/what-is-the-intended-purpose-of-allow-multiple-databases/12537) in the forums; the only newer update was also [2.5 years ago](https://community.open-emr.org/t/second-database-connection/21388) and the implication of what that user did was to ignore it and use their own connection management system.

In the event that I'm totally wrong here and a modern deployment somehow still is relying on this, the latter forum post has a very good workaround for its removal: move the custom connection info into the module that depends on it. This is in general a better approach: modules that are intended to expose shared resources should have the connection logic for those resources self-contained. Anyone that has developed a custom module that would have relied on this should be _more_ than capable of making such an adjustment without hand-holding.

- Removes MultipleDB zend module and its various configuration & setup tendrils
- Removes some ACL setup code that's now irrelevant
- Tidies up some DB stuff on the main schema
- Re-dump error baseline

### DIFF
--- a/interface/modules/zend_modules/config/autoload/global.php
+++ b/interface/modules/zend_modules/config/autoload/global.php
@@ -60,6 +60,7 @@ return [
         'username'       => $sqlConf['login'] ?? '',
         'password'       => $sqlConf['pass'] ?? '',
         'driver_options' => $utf8,
+        'adapters'       => [],
     ],
     'service_manager' => [
         'factories' => $factories


src/Common/Command/AclModify.php
```diff
@@ -25,7 +25,7 @@ protected function configure(): void
         $this
             ->setName('openemr:acl-modify')
             ->setDescription('Modify ACL group of module (enable, disable)')
-            ->addUsage('--site=default --modname=Multipledb --aclgroup=admin --aclaction=enable')
+            ->addUsage('--site=default --modname=Carecoordination --aclgroup=admin --aclaction=enable')
```
Firehed, line 28: these are just usage examples and don't do anything at runtime

bradymiller (approved):

bradymiller:
fare thee well :)

Issue #9308, opened 2025-10-30
laminas-db 2.20.0 not compatible with php 8.4


kojiromike:
OpenEMR's composer.json claims compatibility with php >=8.2.0, and openemr is supposed to work with php 8.4, but openemr has a dependency on laminas/laminas-db 2.20.0, and that only supports up to 8.3.

https://github.com/laminas/laminas-db/blob/207b9ee70a8b518913c1fad688d7a64fe89a8b91/composer.json#L34

bradymiller:
Goodness, what a mess:
https://github.com/laminas/laminas-db/pull/312

Looks like 2 issues going on.

One is microsoft driver sql server for php 8.4 has not come to fruition: https://github.com/microsoft/msphpsql/issues/1523

and looks like it now only considered for security updates: https://getlaminas.org/packages-maintenance-status/

Best case scenario is that things are still working fine with php 8.4/8.5 for OpenEMR use, but won't be able to build when minimum version (ie. platform version) is php 8.4 (will be in 2 years)

Worse case scenario is that things are broken with php 8.4/8/5 for OpenEMR use...

I wonder if we can jump ship on this dependency?

kojiromike:
@bradymiller hmm, IIUC, laminas-db could continue de facto support for newer php versions in their own codebase, but feels forced to declare 8.3 as the ceiling because sql server support is stuck. That explains why we haven't had any real problems with the code itself, because we don't use sql server. I wonder if we can figure out a workaround that will satisfy composer.

We could try to drop the dependency, though.

BTW, this is why these kinds of dependencies (for laminas-db, not for us) should be marked as "suggest" instead of hard dependencies.

kojiromike:
laminas-db is used across many zend modules:

| Module                | Files Using Laminas-DB |
|-----------------------|------------------------|
| Carecoordination      | 10 files               |
| Installer             | 4 files                |
| Patientvalidation     | 3 files                |
| Ccr                   | 3 files                |
| Application           | 2 files                |
| Immunization          | 2 files                |
| Multipledb            | 2 files                |
| Syndromicsurveillance | 2 files                |
| Documents             | 1 file                 |
| Acl                   | 1 file                 |
| Config                | 1 file                 |

PR #10930, opened 2026-02-27
chore(db): Remove most of ApplicationTable & Laminas DB coupling

Firehed:
Towards #10774 and #9308

This is a just about fully AI-based migration. I've tried to self-review it as much as possible, and had the bots review the work several times. By and large, it's changing `zQuery` to `QU::fetchRecords`, `QU::sqlInsert`, or `QU::sqlStatementThrowException` depending on select vs insert vs update.

Moves (nearly?) all queries from Laminas DB to QueryUtils, making substantial progress on cleaning up all of the Laminas stuff, and in particular its DB connection.

- Mechanically migrated hundreds of `new AppliationTable()->zQuery(...)` type calls to use QueryUtils
- Remove the Laminas touch-points from ApplicationTable so it's just holding some legacy date formatting stuff (scope was too large to deal with that in this PR)
- Various dependency-related cleanup
- Update phpstan baseline, removing over 1k errors

Yes

So much claude.

### DIFF
--- a/.phpstan/baseline/openemr.forbiddenGlobalsAccess.php
+++ b/.phpstan/baseline/openemr.forbiddenGlobalsAccess.php
@@ -1308,6 +1308,11 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Direct access to \\$GLOBALS is forbidden\\. Use OEGlobalsBag\\:\\:getInstance\\(\\)\\-\\>get\\(\\) instead\\.$#',
+    'count' => 4,
+    'path' => __DIR__ . '/../../interface/modules/zend_modules/module/Ccr/src/Ccr/Model/CcrTable.php',
+];
+$ignoreErrors[] = [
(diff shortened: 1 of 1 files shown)


interface/modules/zend_modules/module/Application/src/Application/Controller/IndexController.php
```diff
@@ -107,4 +93,73 @@ public function autoSuggestAction()
                                     ]);
         return $index;
     }
+
+    /**
+     * @return array<string, mixed>
+     */
+    private function listAutoSuggest($post, $limit)
```
Firehed, line 100: This was moved from ApplicationTable.

interface/modules/zend_modules/module/Application/src/Application/Plugin/CommonPlugin.php
```diff
@@ -13,45 +13,20 @@
...
-    /**
-     * Function checkACL
-     * Plugin functions are easily access from any where in the project
-     * Call the ACL Check function zAclCheck from ApplicationTable
-     *
-     * @param int    $useID
-     * @param string $sectionID
-     * @return type
-     */
-    public function checkACL($useID, $sectionID)
```
Firehed, line 50: this was completely unused so it enabled a bunch of cleanup

Issue #10637, opened 2026-02-08
feat: add support for MySQL socket connections


kojiromike:
### Describe the problem

OpenEMR currently only supports TCP/IP connections to MySQL via host:port configuration. Some deployments prefer or require Unix socket connections for:
- Performance (sockets are faster for local connections)
- Security (no network exposure)
- Containerized environments where socket mounting is preferred
- Systems where MySQL is configured to only accept socket connections

Currently, socket connections only work implicitly when `$host = 'localhost'` and rely on PHP's default socket path, which may not match the actual MySQL socket location.

### Describe the solution you'd like

Add explicit socket configuration support:

1. **sqlconf.php** - Add `$socket` variable:
   ```php
   $socket = '';  // e.g., '/var/run/mysqld/mysqld.sock'
   $sqlconf["socket"] = $socket;
   ```

2. **library/sql.inc.php** - Pass socket to ADODB:
   ```php
   if (!empty($socket)) {
       $database->socket = $socket;
   }
   ```

3. **Zend modules config** - Add socket to PDO DSN:
   ```php
   // When socket is set, use it instead of host:port
   $dsn = !empty($sqlConf['socket']) 
       ? "mysql:dbname={$db};unix_socket={$socket}"
       : "mysql:dbname={$db};host={$host};port={$port}";
   ```

4. **Setup wizard** - Add optional socket field during installation

### Connection pooling interaction

No special handling needed. PHP persistent connections (`PConnect`) are keyed by all connection parameters including socket path. Socket and TCP connections will naturally be separate pool entries.

### Additional context

This came up while working on #10635 / #10636 (port configuration). Socket support follows the same pattern of ensuring all connection parameters are properly passed through the stack.

mdsupport:
We have been using [this connection logic](https://github.com/mdsupport/mdpub/commit/61facfb4f38fce7adbfb8b3021f7890820c66649) for a few years. Do make this default for production systems. Apart from port and socket, the connection function also accepts other parameters that can simplify PHP code.

On a related note, Mariadb is shedding its MySQL interoperable commands. AUTH_SOCKET vs UNIX_SOCKET is one of those differences relevant for this setup.

mdsupport:
Here is [our socket code](https://github.com/openemr/openemr/compare/master...mdsupport:openemr:feature/adodb-socket) in 8.0.

Issue #10890, opened 2026-02-24
chore(db): Migrate mysql timestamp columns to datetime


Firehed:
OpenEMR currently has ~40 `TIMESTAMP` columns alongside ~220 `DATETIME`. I propose we standardize on the latter.

Motivation:

- Fewer changes go in this direction
- `TIMESTAMP` has 32-bit integer limitations; it'll overflow in 2038. `DATETIME` is good until the year 9999.
- Most library tooling prefers DT over TS

Timezone handling is _different_ across the two, I wouldn't really argue that one is better than the other. But having a mix will almost certainly lead to weird bugs, especially around backups and restoration..

@kojiromike mentioned there may already be an issue for this, but I couldn't find one.

kojiromike:
I wonder what the implications of #10891 might be for this. Apparently OpenEMR's attempts to turn off MySQL strict mode don't work in this case -- mysql refuses to compare a datetime to '' anyway.

Firehed:
There are also columns with invalid dates in their default value. Suffice to say, there's a lot of room for unintended consequences with any change... but I'd rather know about the issues than have people silently losing or corrupting data.

Firehed:
This isn't strictly a subtask of #11307, but it's related for sure and will ease doctrine/orm if not unblock it outright.

Firehed:
## Discovery: TIMESTAMP columns in OpenEMR

Ran an audit to identify all TIMESTAMP columns and their behaviors. Server timezone is UTC (`NOW() == UTC_TIMESTAMP()`), which simplifies migration - no timezone conversion issues.

### Summary: 40 TIMESTAMP columns across 33 tables

#### By behavior:

| Behavior | Count | Tables |
|----------|-------|--------|
| `ON UPDATE CURRENT_TIMESTAMP` | 9 | ccda, ccda_table_mapping, documents_legal_detail, dsi_source_attributes, form_clinical_instructions, form_encounter, form_eye_locking, lists, medex_recalls, users_secure |
| `DEFAULT CURRENT_TIMESTAMP` only | 14 | ar_session, audit_master, background_services, clinical_notes_*, direct_message_log, form_eye_mag_dispense, immunization_observation, list_options, medex_outgoing, medex_prefs, onsite_documents, procedure_order_relationships |
[...]

PR #10944, opened 2026-02-28
feat(db): ORM scaffolding

Firehed:
Prototypes a superficial `doctrine/orm` (`^3.0`) integration:

- Loads DB connection from existing infrastructure
- Sets up automatic column mapping (`camelCase` in PHP is auto-translated to `snake_case` in the DB)
- Determines "dev mode" from existing `OPENEMR__ENVIRONMENT` envvar
- Property types are type-safe, including those that hold serialized json[^json]

[^json]: With the limitation that PHP has no native generics, so you can only go so far with this

Where this will go:
- [x] Setup code to move into a proper PSR-11 container (need that set up first, should not be in scope)
- [ ] Entity Mapping to occur later, probably as part of making other improvements (the ApiToken entity is just for testing)
- [ ] Probably setting up some repositories as well
- [ ] Need to figure out a path for the query logging (this is probably more of a DBAL concern than ORM, but still general blocker)
- [ ] Improve caching setup
- [x] We'll probably want a bunch of custom type mapping support[^types]
- [x] Automatic timestamp (and other generated) column types
- [ ] As the module overhaul continues, some way for them to vend additional paths for `ORMSetup`

[^types]: TBD what, if anything, we need here. I've used this extensively in the past for custom data structures, money types, binary data wrappers, and plenty more. I mostly want to point out that it exists.

Adjacent but out of scope:
- [ ] #10637 
- [ ] Configure DB connection w/ envvars or something other than sqlconf.php
- [ ] Test tooling
- [ ] A couple utility functions [^util] for performance and type safety
- [ ] Additional phpstan integration as needed

[^util]: I frequently have three go-tos: `getByIds($em, $className, $ids)`, `errorIfMissing($maybeEntity)` and `logicallyRequired($maybeEntity)`. The latter two are basically the same except for semantics (client error vs server error)

What I've tested so far:
- [x] Inserting a mapped entity
- [x] Reading existing records created from non-ORM system
- [x] CLI tools 

- Adds doctrine/orm and symfony/cache libraries
- Adds dotenv load to the new bootstrap script
- Sets up basic config concept, though this is not where it will actually go yet
- Adds integration into CLI wrapper (mostly schema validation)

No


src/Entities/Facility.php
```diff
@@ -0,0 +1,140 @@
...
+    #[Mapping\Column]
+    public bool $inactive = false;
+
+    #[Mapping\Column(name: 'date_created', options: ['default' => 'CURRENT_TIMESTAMP'])]
+    #[Attributes\CreatedAt]
+    public readonly DateTimeImmutable $createdAt;
+
+    #[Mapping\Column(name: 'last_updated')]
+    #[Attributes\UpdatedAt]
+    public DateTimeImmutable $updatedAt;
```
Firehed, line 136: Note: in the future if we're targeting 8.4+, this can and should use aviz: `public private(set) DateTimeImmutable $updatedAt`

src/Entities/Facility.php
```diff
@@ -0,0 +1,140 @@
...
+
+    #[Mapping\Column(length: 10, nullable: true)]
+    public ?string $wenoId;
+
+    #[Mapping\Column]
+    public bool $inactive = false;
+
+    #[Mapping\Column(name: 'date_created', options: ['default' => 'CURRENT_TIMESTAMP'])]
+    #[Attributes\CreatedAt]
+    public readonly DateTimeImmutable $createdAt;
```
Firehed, line 132: See side note in #10890 - we can at least use consistent PHP-side variable names even if the column names are not made consistent yet.

Firehed: Related caveat: this specific approach means the property remains uninitialized until after `$em->flush()`. In quite rare circumstances, this could lead to bugs, but they'd be obvious during development and are easy to work around (just don't read the value until after flush)

src/Entities/Facility.php
```diff
@@ -0,0 +1,140 @@
...
+namespace OpenEMR\Entities;
+
+use DateTimeImmutable;
+use Doctrine\DBAL\Types\Types;
+use Doctrine\ORM\Mapping;
+use Ramsey\Uuid\UuidInterface;
+
+#[Mapping\Entity]
+#[Mapping\Table(name: 'facility')]
+class Facility
```
Firehed, line 14: For whatever reason, Claude suggested this would be a good first entity to support/migrate. It has some self-contained-ish logic in FacilityService which is a good place to demonstrate migrating things, though there are a LOT of touch points... so TBD. It's certainly been enough to POC things with.

config/cli.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+use OpenEMR\Console\Command;
+
+return [
+    Command\DebugLoadFacilityCommand::class,
```
Firehed: Debugging code, will be removed.

config/services.php
```diff
@@ -21,4 +22,5 @@
 
     SystemLogger::class => fn (TC $c) => new SystemLogger($c->get(Level::class)),
 
+    TimestampSubscriber::class,
```
Firehed: This is generally more representative of service config under the container model - basically do nothing other than tell the compiler to pre-handle it.

src/Console/Command/DebugLoadFacilityCommand.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
Firehed, line 1: Debugging code, will be removed.

cli
```diff
@@ -23,7 +27,15 @@ file_put_contents(
 );
 
 $application = new Application();
+$application->add($container->get(DebugLoadFacilityCommand::class));
```
Firehed: Debugging code, will be removed.

src/Entities/Facility.php
```diff
@@ -0,0 +1,140 @@
+<?php
```
Firehed, line 1: This is part of the debugging path and doesn't need to be in the initial setup. Probably should land separately.

PR #10955, opened 2026-03-02
refactor(layout): Improve add layout form

### LINKED ISSUE #10954: refactor(layout): Improve add layout form
Eliminate need for manually entering LBF/LBT when creating layout

mdsupport:
Fixes #10954 

Eliminates need for manually entering LBF/LBT when creating layout

### DIFF
--- a/interface/super/edit_layout_props.php
+++ b/interface/super/edit_layout_props.php
@@ -259,7 +259,7 @@ if ($layout_id) {
       <option value='LBT'><?php echo xlt('Patient'); ?></option>
     </select>
 <?php } else { ?>
-    <?php echo text(substr($layout_id, 0, 3) === 'LBT' ? xlt('Patient') : xlt('Encounter')); ?>
+    <?php echo (substr($layout_id, 0, 3) === 'LBT' ? xlt('Patient') : xlt('Encounter')); ?>
 <?php } ?>
   </td>
  </tr>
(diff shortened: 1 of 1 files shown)


interface/super/edit_layout_props.php
```diff
@@ -247,15 +247,33 @@ function get_related() {
...
+  </td>
+  <td>
+<?php if (empty($layout_id)) { ?>
+    <select name='layout_type' id='layout_type' class='form-control' onchange='updateLayoutId()'>
+      <option value='LBF' selected><?php echo xlt('Encounter'); ?></option>
+      <option value='LBT'><?php echo xlt('Patient'); ?></option>
+    </select>
+<?php } else { ?>
+    <?php echo text(substr($layout_id, 0, 3) === 'LBT' ? xlt('Patient') : xlt('Encounter')); ?>
+<?php } ?>
```
kojiromike, line 263: `xlt()` already wraps its result in `text()` (see `library/htmlspecialchars.inc.php:258`), so this double-wraps the output through `htmlspecialchars`. It would turn `&` into `&amp;amp;` in translated strings.

Would something like this work?

```suggestion
    <?php echo (substr($layout_id, 0, 3) === 'LBT') ? xlt('Patient') : xlt('Encounter'); ?>
```

interface/super/edit_layout_props.php
```diff
@@ -530,6 +553,70 @@ function get_related() {
...
+        }
+    }
+
+    formIdEl.value = layoutName !== '' ? layoutType + layoutName : '';
+
+    if (mappingField) {
+        if (layoutType === 'LBT') {
+            mappingField.value = 'Transactions';
+            mappingField.disabled = true;
+        } else {
```
kojiromike, line 589: Disabled form elements aren't included in the POST data — so when "Patient" is selected, `$_POST['form_mapping']` (read at line 138) will be missing, and the category ends up empty instead of "Transactions".

Would using `readOnly` instead of `disabled` work here? Or alternatively, could you add a hidden input that mirrors the value?

```suggestion
            mappingField.value = 'Transactions';
            mappingField.readOnly = true;
```

(and correspondingly for the `else` branch: `mappingField.readOnly = false;`)

PR #303, opened 2016-10-05
Encounter Forms Display Order

mdsupport:
Provides an optional way to control the order in display of forms in encounter summary or reports.  Current display order is vitals + order of entry which creates ever changing / random display.

PR #10969, opened 2026-03-03
feat(forms): Set forms display order

### LINKED ISSUE #10968: feat(forms): Set forms display order
Currently encounter forms are presented in inconsistent order which causes audit issues. Need a mechanism that ensures the forms are presented in an order per local requirements.

mdsupport:
Fixes #10968
Updated version of #303 

Gives optional control over the sequence of forms for an encounter as specified in the list "Encounter Forms Seq and Access".

### DIFF
--- a/src/Services/FormService.php
+++ b/src/Services/FormService.php
@@ -20,6 +20,26 @@ use OpenEMR\Common\Session\SessionWrapperFactory;
 
 class FormService
 {
+    private $listFormsSeqAcl = 'formSeqAcl';
+    
+    private function setupLocalSeqAcl() {
+        $localSeqAcl = [];
+        
+        // Check formSeqAcl list
+        $loRecs = sqlQuery(
+            'select * from list_options where list_id=? and option_id=?',
+            ['lists', $this->listFormsSeqAcl]
+            );
+        // Create formSeqAcl if needed
+        if (!isset($loRecs['option_id'])) {
+            sqlStatement(
+                'INSERT INTO list_options (list_id, option_id, title) VALUES (?,?,?)',
+                ['lists', $this->listFormsSeqAcl, 'Encounter Forms Sequence and Access']
+            );
+        }
+        return true;
+    }
+    
     public function getFormByEncounter(
         $attendant_id,
         $encounter,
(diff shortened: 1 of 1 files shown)


src/Services/FormService.php
```diff
@@ -20,6 +20,26 @@
 
 class FormService
 {
+    private $listFormsSeqAcl = 'formSeqAcl';
```
kojiromike, line 23: Missing type declaration:

```suggestion
    private string $listFormsSeqAcl = 'formSeqAcl';
```

src/Services/FormService.php
```diff
@@ -20,6 +20,26 @@
...
+    
+    private function setupLocalSeqAcl() {
```
kojiromike, line 25: Opening brace should be on its own line (Allman style, per OpenEMR coding standards), and the return type should be declared.

Also, as noted in the review summary, this entire method should be removed — create the list entry in `database.sql` and the upgrade SQL instead of checking at runtime.

```suggestion
    private function setupLocalSeqAcl(): bool
    {
```

src/Services/FormService.php
```diff
@@ -20,6 +20,26 @@
...
+        $localSeqAcl = [];
```
kojiromike, line 26: `$localSeqAcl` is assigned but never used — please remove it.

src/Services/FormService.php
```diff
@@ -46,15 +71,15 @@ public function getFormByEncounter(
         array_push($arraySqlBind, $attendant_id);
 
         // Default $orderby puts vitals first in the list, and newpatient last:
-        $sql .= "ORDER BY $orderby";
+        $sql .= "ORDER BY IFNULL(lo.seq, -1), $orderby";
```
kojiromike, line 74: Using `-1` as the fallback means unconfigured forms sort *first* (before any form with seq >= 0). Should unconfigured forms sort last instead?

```suggestion
        $sql .= "ORDER BY IFNULL(lo.seq, 2147483647), $orderby";
```

src/Services/FormService.php
```diff
@@ -46,15 +71,15 @@ public function getFormByEncounter(
...
 
         $res = sqlStatement($sql, $arraySqlBind);
 
         $all = [];
         for ($iter = 0; $row = sqlFetchArray($res); $iter++) {
             $all[$iter] = $row;
         }
-
+
```
kojiromike, line 82: Unrelated whitespace-only change (removes the blank line). Please revert to avoid noise in the diff.

kojiromike (changes requested):
Thanks for working on this — configurable form ordering is a reasonable feature. A few things need attention before this can move forward.

### Runtime list auto-creation

`setupLocalSeqAcl()` runs a SELECT (and potentially an INSERT) on every call to `getFormByEncounter()`. This is a hot path — it fires every time an encounter is viewed. The `formSeqAcl` list should be created via `database.sql` + upgrade SQL (like every other `list_options` entry), not at runtime.

### Naming

The name `formSeqAcl` and the column alias `local_acl` imply access control, but this PR only implements ordering. If ACL is planned for later, the current PR should use names that match what it actually does (e.g., `formDisplayOrder`). The title "Encounter Forms Sequence and Access" also overpromises.

### Sort direction for unconfigured forms

`IFNULL(lo.seq, -1)` sorts forms without a seq entry *before* everything else (since -1 < any positive seq value). This seems backwards — unconfigured forms should probably sort *after* configured ones (e.g., use `IFNULL(lo.seq, 999)` or `IFNULL(lo.seq, 2147483647)`).

### Relationship to #10955

You also have PR #10955 open ("Improve add layout form"). How do these relate? Is #10955 the UI for managing the list that #10969 consumes?

PR #10982, opened 2026-03-04
refactor(crypto): Migrate CryptoGen to CryptoInterface

Firehed:
Crypto operations have largely relied on a specific implementation; this updates code to rely on the interface methods instead and get its instance from the service locator. This change will ease refactoring of cryptographic wrappers and, potentially, simplify a path towards upgrading data using a legacy crypto version.

This doesn't solve DI, but is a tiny step in that direction.

- Bulk change of `new CryptoGen()` to `ServiceContainer::getCrypto()`
- Fix some typehints in equivalent ways
- Resolves an inconsistency between the interface and implementation

Yes

Claude wrote the rector rule to perform the conversion.


Firehed:
Styling: some of the autofixes are still putting imports in dumb places. I'm going to leave it for now, this is easily fixable through phpcs/phpcbf. I'm going to set up and apply those rules in a separate PR, which will catch these and likely a whole lot of others.

PR #10992, opened 2026-03-04
feat(mobile): add native-feeling responsive mobile shell (MVP)

craigrallen:
## Summary
Introduces a mobile-first, app-like responsive shell for OpenEMR as a first step toward a native-feeling mobile interface.

## What this adds
- `interface/mobile/index.php` app shell with:
  - top app bar
  - fullscreen OpenEMR viewport (`iframe`)
  - bottom tab navigation for key workflows
  - settings sheet for configuring installation URL
- `interface/mobile/mobile.css`
  - touch-first layout, safe-area support, responsive behavior
- `interface/mobile/manifest.webmanifest`
  - PWA metadata
- `interface/mobile/sw.js`
  - shell asset caching for fast loads
- `interface/mobile/README.md`
  - rollout notes and next iteration roadmap

## Why
This keeps the full existing web functionality while providing a mobile UX baseline that feels closer to a native app and can be iterated quickly.

## Next loop items
1. Role-based tabs and dynamic routing by context
2. Mobile overrides for top-used screens
3. Native packaging via Capacitor
4. Deep links to patient, encounter, and scheduling flows

### DIFF
--- a/Documentation/api/SMART_ON_FHIR.md
+++ b/Documentation/api/SMART_ON_FHIR.md
@@ -343,7 +343,7 @@ const response = await fetch('https://localhost:9300/apis/default/fhir/Patient/1
 
 #### Characteristics
 - ✅ Launched from OpenEMR UI
-- ✅ Patient context pre-selected
+- ✅ Patient context preselected
 - ✅ Encounter context available (SMART v2.2.0)
 - ✅ User already authenticated
 - ✅ Faster workflow for clinicians
@@ -530,7 +530,7 @@ SMART apps can receive contextual information about the launch environment.
 #### When Available
 - ✅ EHR launch from patient chart
 - ✅ Patient standalone apps (patient is authorized user)
-- ❌ Standalone provider apps (no pre-selected patient)
+- ❌ Standalone provider apps (no preselected patient)
 
 #### Requesting Patient Context
 
--- a/interface/main/holidays/Holidays_Storage.php
+++ b/interface/main/holidays/Holidays_Storage.php
@@ -104,7 +104,7 @@ class Holidays_Storage
     }
 
     /**
-     * This function opend the $file(csv) and parses it to insert the values in the calendar_external so later they can be imported as events
+     * This function opens the $file(csv) and parses it to insert the values in the calendar_external so later they can be imported as events
      * csv format -> date,description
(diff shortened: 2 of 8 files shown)


Firehed:
Paging @rollingventures who may have thoughts here.

I like the general motivation here, but I do want to be aware of other UI stuff in flight so we can avoid having too many conflicts. Could be fine, dunno yet :)

rollingventures:
@Firehed - do not merge.

1. No auth check - interface/mobile/index.php lines 1-10
2. Open redirect - index.php lines 56-78 (the navigate() and getBaseUrl() functions)
3. SW caches PHP - sw.js line 3
4. Admin tab visible - index.php line 38

This adds tons of security issues specific to the PR. Is this a bot? :)

Firehed:
The base URL is, AFAIK, an existing (anti-)pattern, but it sounds like there's quite a bit of work that needs to be done first either way.

craigrallen:
I am trying to start implementing a mobile interface in a iterative way. Build a PWA first then move to capacitor.

Happy to abandon this PR attempt. If you want to give me a clear direction the project wants to move in, I'll create it.

Been using since v5.0.1, but have some push back that it is hard to do things on mobile devices from users and want to give something back to the project.

Firehed:
Fully agreed that mobile is far from the best experience right now.

As of yet, I don't think we have any specific directions regarding mobile; however, we want to start moving things in the direction of API-based data fetches rather than the php-html-js entrypoint files that exist today, which are messy to write and even harder to maintain. I can say pretty confidently we don't want to have a dedicated set of mobile-specific pages; rather, we'd prefer to make the existing UI more responsive.

Given the long-term size and scope of a mobile interface (or even apps, which is where I assume you're going re: capacitor), I think it'd be beneficial to share your ideas in the forums or the [weekly conference call](https://www.open-emr.org/wiki/index.php/OpenEMR_Once_Weekly_Conference_Call_Details) if you're able to do so. For smaller changes there's certainly no need, but something this broad would really benefit from community input first.

If possible (and this does need dedicated pages rather than just style/layout changes to existing UIs), it's probably best for this to be in a module rather than core - but that's also something where community input would be valuable.

(I've moved this into draft state pending further discussion)

PR #10995, opened 2026-03-05
refactor(crypto): Pass enums instead of strings

Firehed:
Hoists up the KeySource information in the CryptoInterface so that callers pass an enum instead of strings. This improves the ergonomics since you'll get development-time feedback from tooling instead of waiting for a failure at runtime or test-time.

Note: I want to do the same with the keyversion stuff, separately.

- Changes `$keySource` param from `string` to `KeySource`
- Updates callers accordingly

Yes, mechanical updates of callers done by claude.

### DIFF
--- a/portal/index.php
+++ b/portal/index.php
@@ -26,7 +26,6 @@ use OpenEMR\BC\ServiceContainer;
 use OpenEMR\Common\Auth\Exception\OneTimeAuthException;
 use OpenEMR\Common\Auth\Exception\OneTimeAuthExpiredException;
 use OpenEMR\Common\Auth\OneTimeAuth;
-use OpenEMR\Common\Crypto\KeySource;
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Logging\EventAuditLogger;
 use OpenEMR\Common\Logging\SystemLogger;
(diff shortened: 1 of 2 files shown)


portal/index.php
```diff
@@ -174,7 +175,7 @@
         exit();
     }
 
-    $token_one_time = $crypto->decryptStandard($_GET['forward_email_verify'], null, 'drive', 6);
+    $token_one_time = $crypto->decryptStandard($_GET['forward_email_verify'], null, KeySource::Drive, 6);
```
Firehed: Aside: anything explicitly passing `Drive` is kind of suspect, though I suspect this really is "I should be using named arguments". Thoughts?

src/Common/Auth/OneTimeAuth.php
```diff
@@ -159,7 +160,7 @@ public function decodePortalOneTime($onetime_token, $redirect_token = null, $log
 
         if (strlen((string)$onetime_token) >= 64) {
             if ($this->cryptoGen->cryptCheckStandard($onetime_token)) {
-                $one_time = $this->cryptoGen->decryptStandard($onetime_token, null, 'drive', 6);
+                $one_time = $this->cryptoGen->decryptStandard($onetime_token, null, KeySource::Drive, 6);
```
Firehed: (same)

portal/index.php
bradymiller: @Firehed , This is because need that 4th parameter, so a bad actor doesn't try to use a lower potentially less secure encryption version (ie. sets the minimum version to 6).

src/Common/Auth/OneTimeAuth.php
bradymiller: see above post

portal/index.php
Firehed: Thanks for confirming!

Firehed:
I'm ignoring the spellcheck thing right now, it's unrelated to this change.

PR #11001, opened 2026-03-06
feat(config): Start to set up PSR-11 DI container

Firehed:
Adds a PSR-11 container and gets it into a minimally-usable state for upcoming broader integration.

Noteworthy: the database connection stuff tends to be some of the most complex setup for any application; that's not really illustrative of day-to-day service configuration. Nearly all services will be autowired one-liners (just listing the class).

- Add library
- Perform setup process in some upcoming (but not accessible through normal means) paths
- Rework DBAL setup to leverage it
- Rework migrations tooling to leverage it
- Sets up several PSR services for future use (logger already lightly used here)

No

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -37,7 +37,7 @@ Services added here **must** support constructor-based dependency injection. Thi
 
 ### On Singletons
 
-By default, the container itself will re-use services as they are requested.
+By default, the container itself will reuse services as they are requested.
 Like (just about) all PHP code, there is still a fresh instance per request.
 
 In the rare case that this behavior is inappropriate for the service, the definition can be wrapped in a `factory()` call.


config/psr.php
```diff
@@ -0,0 +1,46 @@
...
+    ResponseFactoryInterface,
+    ServerRequestFactoryInterface,
+    StreamFactoryInterface,
+    UploadedFileFactoryInterface,
+    UriFactoryInterface,
+};
+use Psr\Log\LoggerInterface;
+use OpenEMR\Common\Http\Psr17Factory;
+
+return [
```
Firehed, line 29: Implementation note: there's zero technical need for these files to all be separated. I've found it helpful to group adjacent things together, both to keep files a manageable size and to avoid name conflicts. I tend to end up with about a dozen files that group related services in large projects

config/env.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare(strict_types=1);
+
+use function Firehed\Container\env;
+
+return [
+    // The unspecified default today is dev=debug, other=warning
+    'LOG_LEVEL' => env('LOG_LEVEL', 'warning'),
+
+    // FIXME: this works only for the CLI path, not actual multi-site.
```
Firehed, line 21: This will need to be resolved prior to collapsing BC\Database to use this, but it's not a problem today.

config/cli-config.php
```diff
@@ -1,48 +0,0 @@
-<?php
```
Firehed, line 1: The need for this file has been obsoleted by reusable config in the container and adjusted CLI integration.

cli
```diff
@@ -5,9 +5,11 @@ declare(strict_types=1);
 
 use Doctrine\Migrations\DependencyFactory;
 use Doctrine\Migrations\Tools\Console\ConsoleRunner;
+use Firehed\Container\TypedContainerInterface;
 use Symfony\Component\Console\Application;
 
-require_once __DIR__ . '/bootstrap.php';
+$container = require_once __DIR__ . '/bootstrap.php';
```
Firehed, line 11: At least for the moment, I want to avoid making it too easy to grab the container (e.g. a wrapper function) to keep this as "clean" as possible to start, and to avoid replicating existing anti-patterns. It really only should be read in a very few select locations for long-term maintainability - namely, front-controllers and related setup paths.

.phpstan/baseline/argument.type.php
```diff
@@ -416,6 +416,11 @@
     'count' => 4,
     'path' => __DIR__ . '/../../ccr/transmitCCD.php',
 ];
+$ignoreErrors[] = [
```
Firehed, line 419: Not much to be done here since it's not using the standard enum path; Monolog is actually not a great citizen as far as interop goes.

.phpstan/baseline/openemr.deprecatedSqlFunction.php
```diff
@@ -51,6 +51,11 @@
     'count' => 1,
     'path' => __DIR__ . '/../../ccr/transmitCCD.php',
 ];
+$ignoreErrors[] = [
```
Firehed, line 54: nit: the error code here doesn't line up with the actual error. nbd.

.phpstan/baseline/openemr.forbiddenInstantiation.php
```diff
@@ -16,6 +16,11 @@
     'count' => 1,
     'path' => __DIR__ . '/../../ccr/display.php',
 ];
+$ignoreErrors[] = [
```
Firehed, line 19: This and above are expected here; the container is the one place it _should_ be instantiated.

phpstan.neon.dist
```diff
@@ -16,6 +16,7 @@ parameters:
   - apis
   - ccdaservice
   - ccr
+  - config
```
Firehed, line 19: At some point I'd like to simplify this whole list to `.`.

Firehed:
Docs added, we can and will make adjustments over time as integration changes.

PR #11053, opened 2026-03-09
fix: remove OEGlobalsBag dependency from version.php

### LINKED ISSUE #11052: fix: admin.php crashes with OEGlobalsBag not found after globals refactor
## Bug

`admin.php` returns a 500 error:

```
PHP Fatal error: Uncaught Error: Class "OpenEMR\Core\OEGlobalsBag" not found
in /var/www/localhost/htdocs/openemr/version.php:59
```

## Cause

PR #11017 (commit be636987b7) added `OEGlobalsBag::getInstance()->set('v_js_includes', ...)` to `version.php`. However, `admin.php` includes `version.php` without first loading the Composer autoloader, so the `OEGlobalsBag` class is unavailable.

## Fix

Move the `OEGlobalsBag::getInstance()->set('v_js_includes', ...)` call from `version.php` to `interface/globals.php` (which already loads the autoloader before including `version.php`). `version.php` should remain a lightweight file that only sets plain variables.

## Regression test

Add an isolated test that verifies `version.php` can be included standalone without the autoloader (i.e., it must not reference any classes).

kojiromike:
## Summary

- Remove `OEGlobalsBag` `use` statement and `::getInstance()->set()` call from `version.php`, which crashes `admin.php` (no autoloader)
- The call is redundant — `interface/globals.php:406` already sets `v_js_includes` into the bag
- Add isolated regression test (`VersionFileTest`) that verifies standalone inclusion and prevents future class dependencies

Closes #11052

## Test plan

- [x] `composer phpunit-isolated` — new `VersionFileTest` passes (standalone inclusion + no class tokens)
- [x] `composer phpstan` — no errors
- [x] All pre-commit hooks pass
- [x] Manual: visit `/admin.php` and confirm it loads without fatal error

### DIFF
--- a/tests/Tests/E2e/AaLoginTest.php
+++ b/tests/Tests/E2e/AaLoginTest.php
@@ -81,6 +81,21 @@ class AaLoginTest extends PantherTestCase
         $this->client->quit();
     }
 
+    #[Test]
+    public function testAdminPageLoads(): void
+    {
+        $this->base();
+        try {
+            $this->crawler = $this->client->request('GET', '/admin.php');
+            $title = $this->client->getTitle();
+            $this->assertSame('OpenEMR Site Administration', $title, 'FAILED to load admin.php');
+        } catch (\Throwable $e) {
+            $this->client->quit();
+            throw $e;
+        }
+        $this->client->quit();
+    }
+
     private function loginPage(): void
     {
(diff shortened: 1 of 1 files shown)


kojiromike:
Not a candidate for backport to rel-800 — depends on the OEGlobalsBag refactoring that is only on master.

PR #11056, opened 2026-03-09
feat(users): add email field to user admin UI

### LINKED ISSUE #11054: Add Users email property in the UI
### Describe the problem

Would like to include email as a property for User accounts. This field could be useful for sending password resets, notifications, an extra check for user uniqueness, among other use cases. The field exists in the database, but I don't see it in the UI (Config > Users).

Add Email field to Users, so that I can add, update, and delete emails associated with a specific user account.

Image

rollingventures:
Add email field to the Users admin page (Config > Users) to allow administrators to manage user email addresses. The email column already exists in the users database table but was not exposed in UI.

This enables future features and future migrations:
(but not limited to...)
- Password reset functionality
- User notifications
- User uniqueness verification

Fixes #11054

The email column exists in the users database table but was not accessible through the admin UI. Administrators had no way to view or edit user email addresses without direct database access.

See above description.

### DIFF
--- a/.phpstan/baseline/binaryOp.invalid.php
+++ b/.phpstan/baseline/binaryOp.invalid.php
@@ -11643,7 +11643,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Binary operation "\\." between non\\-falsy\\-string and mixed results in an error\\.$#',
-    'count' => 31,
+    'count' => 29,
     'path' => __DIR__ . '/../../interface/usergroup/usergroup_admin.php',
 ];
 $ignoreErrors[] = [


rollingventures:
See modification:
image

rollingventures:
Added JavaScript email validation to both forms using the existing isValidEmail() function (same as Google Email field).

rollingventures:
Added Email Column to the Admin -> Users / Groups
Screenshot 2026-03-09 at 3 58 13 PM

interface/usergroup/usergroup_admin_add.php
```diff
@@ -401,6 +407,9 @@ function authorized_clicked() {
 <td><span class="text"><?php echo xlt('Weno Provider ID'); ?>: </span></td><td><input type="text" name="erxprid" style="width:120px;" class="form-control" value="<?php echo attr($iter["weno_prov_id"] ?? ''); ?>"></td>
 <td><span class="text"><?php echo xlt('Google Email for Login'); ?>: </span></td><td><input type="text" name="google_signin_email" style="width:150px;" class="form-control" value="<?php echo attr($iter["google_signin_email"] ?? ''); ?>"></td>
 </tr>
+<tr>
+<td><span class="text"><?php echo xlt('Email'); ?>: </span></td><td><input type="text" name="email" style="width:120px;" class="form-control"></td>
```
Firehed:
```suggestion
<td><span class="text"><?php echo xlt('Email'); ?>: </span></td><td><input type="email" name="email" style="width:120px;" class="form-control"></td>
```

get some free validation here plus better support for on-screen keyboards

interface/usergroup/usergroup_admin.php
```diff
@@ -357,6 +362,7 @@
             "', mname = '"         . add_escape_custom(trim(($_POST['mname'] ?? ''))) .
             "', lname = '"         . add_escape_custom(trim(($_POST['lname'] ?? ''))) .
             "', suffix = '"         . add_escape_custom(trim(($_POST['suffix'] ?? ''))) .
+            "', email = '"         . add_escape_custom(trim(($_POST['email'] ?? ''))) .
```
Firehed, line 365: Normally I'd say this needs server-side validation (and it does), but seeing the status quo, perhaps it's not a blocker.

interface/usergroup/usergroup_admin.php
```diff
@@ -306,6 +306,11 @@
             sqlStatement("update users set google_signin_email = ? where id = ? ", [$googleSigninEmail, $_POST["id"]]);
         }
 
+        if (isset($_POST["email"])) {
+            $email = empty($_POST["email"]) ? null : trim((string) $_POST["email"]);
```
Firehed: This is a bit funky, though I assume it matches most of the surrounding code.

I think it can be simplified to `$email = trim($_POST['email'] ?? ''); if ($email !== '') { update }` or thereabouts.

adunsulag:
FYI, these fields are accessible via the Admin -> Address Book. Its pretty annoying how these two GUIs are kept separate, but that is how people have historically edited the email fields.   Note there is email and trusted email (used for the Direct Trust email messaging through EMRDirect.

I think its good to add this here, but be aware that is where editing has historically been done in the past.

image

rollingventures:
> FYI, these fields are accessible via the Admin -> Address Book. Its pretty annoying how these two GUIs are kept separate, but that is how people have historically edited the email fields. Note there is email and trusted email (used for the Direct Trust email messaging through EMRDirect.
> 
> I think its good to add this here, but be aware that is where editing has historically been done in the past.

@adunsulag Thanks for pointing that out! As I was reversing - it appears the Address Book manages external contacts that don't have login credentials, while User Admin manages actual system users with logins. They both use the users table but for different entry types. If I am not mistaken, the Address Book has always had the email field for external contacts, but system users with login accounts couldn't have their email edited. 

This PR attempts to fill that gap.

adunsulag:
@rollingventures Address Book will manage both internal and external users and in the original OpenEMR core code (almost 20 years ago when I traced this back a while ago) was intended to be the complete contact management solution whereas the user management screen was strictly for permission and credential management.  Why embedding the address book or launching it from within the user edit screen was never done I don't know.

kojiromike (commented):
I can't tell if this posted visibly or not…

LGTM. One thing for a followup: the UPDATE path guards on $email !== '', so once set, admins can't clear the field. The INSERT path always writes the value, so the two are inconsistent. Worth a small followup to remove the guard and run the update unconditionally.

interface/usergroup/usergroup_admin.php
```diff
@@ -306,6 +306,11 @@
             sqlStatement("update users set google_signin_email = ? where id = ? ", [$googleSigninEmail, $_POST["id"]]);
         }
 
+        $email = trim($_POST['email'] ?? '');
+        if ($email !== '') {
+            sqlStatement("update users set email = ? where id = ? ", [$email, $_POST["id"]]);
+        }
```
bradymiller, line 312: hi @rollingventures ,
If a user sets an email. And then later comes in and removes the email (ie. changes field to empty), will this capture that?

PR #11066, opened 2026-03-10
refactor(audit): Rework internals of audit logger

Firehed:
> [!IMPORTANT]
> Not working yet! It'll take a lot of tuning to integrate properly, but I think the high-level direction is about right.

Fixes #

PR #11069, opened 2026-03-10
refactor(audit): Extract ATNA logging

Firehed:
Towards #11073.

ATNA auditing, and its transport mechanism, are hard-coupled to the audit logging service. This makes unit testing difficult, along with additional refactors to the auditing service.

Where this is going (over several PRs):
- Each destination for audit logs gets turned into a "sink" class
- The events (either provided directly or through the query tools) will get decoupled and restructured, then passed as structured data to the "sink"s
- The events will become increasingly type-safe
- The aim is to have a foundation that we can build on for DBAL integration as that work continues.
- ATNA specifically will probably get reworked to use `SimpleXML` instead of a giant string with sprintf

See also: https://github.com/openemr/openemr/pull/11066#pullrequestreview-3924816868

- Creates `AtnaSink` class, `Atna\TcpWriter` class, and `Atna\WriterInterface` interface
- Delegates (effecively) all of the ATNA-related logic in `EventAuditLogger` to the new class
- The transport mechanism is pluggable, though the existing EventAuditLogger path continues to hard-wire TCP
- Direct DateTime moved to ClockInterface to further ease testing

Tests were AI written (Claude), most of the rest was me.


src/Common/Logging/Audit/AtnaSink.php
```diff
@@ -0,0 +1,181 @@
...
+        private Atna\WriterInterface $writer,
+        private bool $enabled,
+        private string $host,
+        private string $serverName,
+        private string $serverAddress,
+    ) {
+    }
+
+    // Future: handle a better-typed DTO
+    public function record(string $user, string $group, string $event, int $patientId, int $outcome, string $comments): void
```
Firehed, line 70: I'm not going to fight the oddity of `int $outcome` over `bool $outcome`... at least not yet. Bigger fish to fry.

PR #11072, opened 2026-03-11
fix(session): respect sessionAllowWrite global in PHPSessionWrapper (#10931)

### LINKED ISSUE #10931: PHPSessionWrapper constructor bypasses read_and_close session mode, causing lock contention
## Problem

`PHPSessionWrapper::__construct()` hardcodes `SessionUtil::coreSessionStart($webroot, false)` (write mode), which starts the session before `globals.php` can start it in read-only mode. This means the `$sessionAllowWrite` / `read_and_close` mechanism documented in `globals.php` is dead code — every request acquires an exclusive write lock on the session file regardless.

### How it happens

In `globals.php`:

```php
$read_only = empty($sessionAllowWrite);  // true by default
$session = SessionWrapperFactory::getInstance()->getWrapper();  // ← starts session in WRITE mode
if (session_status() === PHP_SESSION_NONE && !$session->isSymfonySession()) {
    SessionUtil::coreSessionStart($web_root, $read_only);  // never reached
}
```

`getWrapper()` creates `PHPSessionWrapper`, whose constructor calls `coreSessionStart($webroot, false)`. By the time `globals.php` checks `session_status()`, the session is already active in write mode, so the read-only start is skipped.

### Impact

Pages that fire concurrent AJAX requests (e.g., patient summary with ~7 fragment endpoints) serialize on the session file lock. Each request holds the lock for its full duration instead of releasing it immediately via `read_and_close`. This causes cascading slow loads (3s+ per request in production).

### Suggested fix

Remove the `session_start` call from `PHPSessionWrapper::__construct()`. The constructor should not start the session — that's already handled by `globals.php`. The wrapper should be passive, operating on whatever session state exists.

abishop1990:
Fixes #10931

## Problem
PHPSessionWrapper::__construct() hardcodes the second parameter to SessionUtil::coreSessionStart() as false (write mode), which bypasses the sessionAllowWrite mechanism documented in globals.php. This means every request acquires an exclusive write lock on the session file, preventing the read_and_close functionality from working.

## Solution
Check the sessionAllowWrite global setting and respect it when starting sessions:
- If sessionAllowWrite is empty/false (default), open in read-only mode
- If sessionAllowWrite is true, open in write mode

This allows concurrent read-only sessions to proceed without waiting for a write lock, significantly improving performance on high-concurrency deployments.

## Changes
- Modified PHPSessionWrapper::__construct() to retrieve sessionAllowWrite from OEGlobalsBag
- Calculate read-only mode based on sessionAllowWrite (following globals.php pattern)
- Pass calculated value to SessionUtil::coreSessionStart()

## Testing
- Change is minimal and focused (8 lines added)
- Follows existing pattern documented in globals.php and used throughout the codebase
- SessionWrapperFactory tests should pass with this change

### DIFF
--- a/src/Common/Session/PHPSessionWrapper.php
+++ b/src/Common/Session/PHPSessionWrapper.php
@@ -29,15 +29,14 @@ class PHPSessionWrapper implements SessionWrapperInterface
 
         $globalsBag = OEGlobalsBag::getInstance();
         // Empty webroot is valid - means OpenEMR is at document root
-        $webroot = $globalsBag->get('webroot') ?? '';
+        $webroot = $globalsBag->getString('webroot');
 
         // Respect sessionAllowWrite global: if not set/empty, open in read-only mode.
         // This follows the pattern in globals.php: $read_only = empty($sessionAllowWrite)
         // Allows read_and_close mechanism to reduce session lock contention on concurrent requests.
-        $sessionAllowWrite = $globalsBag->get('sessionAllowWrite');
-        $readOnly = empty($sessionAllowWrite);
+        $sessionAllowWrite = $globalsBag->getBoolean('sessionAllowWrite');
 
-        SessionUtil::coreSessionStart($webroot, $readOnly);
+        SessionUtil::coreSessionStart($webroot, !$sessionAllowWrite);
     }
 
     public function getId(): string


src/Common/Session/PHPSessionWrapper.php
```diff
@@ -30,7 +30,14 @@ public function __construct()
         $globalsBag = OEGlobalsBag::getInstance();
         // Empty webroot is valid - means OpenEMR is at document root
         $webroot = $globalsBag->get('webroot') ?? '';
-        SessionUtil::coreSessionStart($webroot, false);
+
+        // Respect sessionAllowWrite global: if not set/empty, open in read-only mode.
+        // This follows the pattern in globals.php: $read_only = empty($sessionAllowWrite)
+        // Allows read_and_close mechanism to reduce session lock contention on concurrent requests.
+        $sessionAllowWrite = $globalsBag->get('sessionAllowWrite');
```
zmilan: This will not work in 100% of cases, since on some pages, the session is created before `globals.php` is required. But it could improve most of the cases.
I am sorry for creating this issue, but at that moment, I didn't fully understand what and why we are doing that way... it is handled properly in https://github.com/openemr/openemr/pull/10244, but this can help until then.

PR #11077, opened 2026-03-11
refactor(calendar): extract inline JS from add_edit_event.php

### LINKED ISSUE #8057: Refactor add_edit_event.php to Separate JavaScript from PHP for Better Readability

haoyu-haoyu:
## Summary

- Extracts ~300 lines of inline JavaScript from `add_edit_event.php` into a new `add_edit_event.js` file
- PHP-dependent values (durations, date/time formats, translations, user ID, etc.) are passed to JS via a `window.addEditEventConfig` configuration object
- The PHP file retains only the config object setup and the event dispatcher call (which requires PHP)

This is an incremental step toward the broader JS/PHP separation effort described in #8057. Starting with one of the larger inline script blocks to demonstrate the pattern.

### `interface/main/calendar/add_edit_event.js` (new)
- Contains all extracted JavaScript functions: `setpatient`, `sel_patient`, `setgroup`, `sel_group`, `set_display`, `set_category`, `set_allday`, `set_repeat`, `set_days_every_week`, `dateChanged`, `setappt`, `find_available`
- References PHP values through `window.addEditEventConfig` instead of inline PHP interpolation

### `interface/main/calendar/add_edit_event.php` (modified)
- Builds `$durationsJson` array from DB query results, then passes it via `json_encode()` instead of echoing JS array assignments
- Adds `window.addEditEventConfig` object that bridges PHP → JS
- Replaces inline `<script>` block with `<script src="add_edit_event.js">`
- Keeps `AppointmentRenderEvent::RENDER_JAVASCRIPT` dispatch inline (requires PHP event system)

## Test plan

- [ ] Verify add/edit event dialog loads correctly
- [ ] Test patient search popup callback
- [ ] Test group search popup callback  
- [ ] Test category selection updates duration and display
- [ ] Test all-day vs timed event toggle
- [ ] Test repeat/days-every-week checkbox behavior
- [ ] Test date change updates repeat type options
- [ ] Test find-available popup with various date/time formats
- [ ] Verify `AppointmentRenderEvent` JS hooks still fire

Closes #8057 (partial)


interface/main/calendar/add_edit_event.php
```diff
@@ -993,11 +993,11 @@ function setEventDate($start_date, $recurrence)
...
     durations: <?php echo json_encode($durationsJson); ?>,
-    timeDisplayFormat: <?php echo js_escape(OEGlobalsBag::getInstance()->get('time_display_format')); ?>,
-    dateDisplayFormat: <?php echo js_escape(OEGlobalsBag::getInstance()->get('date_display_format')); ?>,
-    webRoot: <?php echo js_escape(OEGlobalsBag::getInstance()->get('web_root')); ?>,
-    eid: <?php echo js_escape((int)$eid); ?>,
-    userId: <?php echo js_escape($userid); ?>,
+    timeDisplayFormat: <?php echo js_escape(OEGlobalsBag::getInstance()->getString('time_display_format')); ?>,
+    dateDisplayFormat: <?php echo js_escape(OEGlobalsBag::getInstance()->getString('date_display_format')); ?>,
+    webRoot: <?php echo js_escape(OEGlobalsBag::getInstance()->getString('web_root')); ?>,
+    eid: <?php echo js_escape((string)(int)$eid); ?>,
```
kojiromike: It isn't necessary to js_escape an integer. That is, `(int)` is sufficient sanitization here.

interface/main/calendar/add_edit_event.js
```diff
@@ -0,0 +1,286 @@
...
+    //Infeg Save button should become active once an appointment is selected.
+    $('#form_save').attr('disabled', false);
+    top.restoreSession();
+    // (CHEMED) Conditional value selection, because there is no <select> element
+    // when making an appointment for a specific provider
+    var s = document.forms[0].form_provider;
+    var f = document.forms[0].facility;
+    if (addEditEventConfig.userId != 0) {
+        s = document.forms[0].form_provider.value;
+        f = document.forms[0].facility.value;
```
rollingventures, line 264: **Fragile type coercion: `userId != 0` compares a string to a number**

  The original PHP was a compile-time branch (`<?php if ($userid != 0) { ?>`) — the JS for the wrong branch was never emitted. Now it's a runtime check where `userId` is a JSON-encoded string (from
  `js_escape((string)$userid)`).

  This *happens* to work because JS loose equality coerces `"0" == 0` → true and `"" == 0` → true, matching PHP's behavior. But it's non-obvious and fragile.

  Building on @kojiromike's point about `eid` — the same applies here. `js_escape()` on an integer value is unnecessary and actually *introduces* the type mismatch by turning a number into a quoted
  string. Both `eid` and `userId` should be emitted as bare integers:

  ```php
  eid: <?php echo (int)$eid; ?>,
  userId: <?php echo (int)$userid; ?>,
  ```

  This produces clean JS number literals (`userId: 0` instead of `userId: "0"`), making strict equality work naturally on the JS side:

  ```js
  if (addEditEventConfig.userId !== 0) {
  ```

  The `(int)` cast is sufficient sanitization since it can only produce a number — no XSS vector possible.

interface/main/calendar/add_edit_event.js
```diff
@@ -0,0 +1,286 @@
...
+        } else {
+            durationDisabled = false;
+        }
+    }
+    document.getElementById('tdallday1').style.color = color1;
+    document.getElementById('tdallday2').style.color = color2;
+    document.getElementById('tdallday5').style.color = color2;
+    f.form_hour.disabled = timeDisabled;
+    f.form_minute.disabled = timeDisabled;
+    if (addEditEventConfig.timeDisplayFormat == 1) {
```
rollingventures: **Minor: Missing null guard for `form_ampm` element**

  The original PHP completely omitted this line when `time_display_format != 1` (24-hour mode),
  meaning the code never tried to access `f.form_ampm` when it didn't exist in the DOM. The new
  runtime check is functionally equivalent but slightly less safe.

  Consider adding a null check:

  ```js
  if (addEditEventConfig.timeDisplayFormat == 1 && f.form_ampm) {
  ```

  This prevents a potential `TypeError` if the DOM element is absent when the format is not 12-hour.

haoyu-haoyu:
Thanks @kojiromike and @rollingventures for the thorough review! All three points addressed in the latest commit:

1. **`eid`**: Removed `js_escape()` wrapping — `(int)$eid` is output directly. `(int)` cast is sufficient sanitization.

2. **`userId`**: Changed from `js_escape((string)$userid)` to `(int)$userid` so the JS `userId != 0` comparison uses numeric equality instead of fragile string coercion.

3. **`form_ampm` null guard**: Added `&& f.form_ampm` check in both `set_allday()` (line 128) and `setappt()` (line 252) to handle 24-hour mode where the AM/PM select element doesn't exist in the DOM.

rollingventures:
The other two items worth addressing... 

1. **Cache-busting**: `<script src="add_edit_event.js">` needs a versioned query string (`?v=<?php echo OEGlobalsBag::getInstance()->get('v_js_includes'); ?>`) or browsers will serve stale files after updates.

2. **RENDER_JAVASCRIPT ordering**: The event dispatch now runs *before* the external JS loads. If any module's injected JS immediately calls extracted functions (`setgroup()`, `set_category()`, etc.), it'll throw a `ReferenceError`. Consider moving the dispatch after the `<script src="add_edit_event.js">` tag.

haoyu-haoyu:
Thanks @rollingventures, both addressed:

1. **Cache-busting**: Added `?v=<?php echo OEGlobalsBag::getInstance()->getString('v_js_includes'); ?>` to the script tag.

2. **Script ordering**: Moved the external `<script src="add_edit_event.js">` **before** the `RENDER_JAVASCRIPT` event dispatch, so module listeners can call the extracted functions immediately.

(Apologies for the accidental close — reopened.)

interface/main/calendar/add_edit_event.js
bradymiller: hi @rollingventures ,
To clarify, agree js_escape() is not needed on a known (int), but it should not be harmful. It is basically a wrapper for json_encode, which will not quote int (at least from my understanding and testing it at https://onlinephp.io/json-encode). This is why json_encode works so nicely when the type is not known.

bradymiller: interestingly though json_encode return type is a string, so if doing it to a variable in php code then would have typing issues. But since echoing it in a js variable, that js variable will be an int without quotes around it.

interface/main/calendar/add_edit_event.php
```diff
@@ -992,300 +985,53 @@ function setEventDate($start_date, $recurrence)
     $catoptions .= ">" . text(xl_appt_category($crow['pc_catname'])) . "</option>\n";
 }
 ?>
+<!-- Configuration object: pass PHP values to external JS (Issue #8057) -->
+<script>
+<?php require OEGlobalsBag::getInstance()->get('srcdir') . "/formatting_DateToYYYYMMDD_js.js.php" ?>
 
+var mypcc = <?php echo OEGlobalsBag::getInstance()->getInt('phone_country_code'); ?>;
+
+window.addEditEventConfig = {
```
kojiromike: Would it be worth building this entire config as a PHP associative array and emitting it with a single `json_encode()`? Since `js_escape` *is* `json_encode` and `xlj` is `json_encode(xl(...))`, the output is identical — but the PHP side becomes a clean data structure instead of ~30 lines of interleaved PHP/JS.

Something like:

```php
$addEditEventConfig = [
    'durations' => $durationsJson,
    'timeDisplayFormat' => (int) OEGlobalsBag::getInstance()->getString('time_display_format'),
    'dateDisplayFormat' => (int) OEGlobalsBag::getInstance()->getString('date_display_format'),
    'webRoot' => OEGlobalsBag::getInstance()->getString('web_root'),
    'eid' => (int) $eid,
    'userId' => (int) $userid,
    'translations' => [
        'patientSearch' => xl('Patient Search'),
        'groupSearch' => xl('Group Search'),
        'availableAppointments' => xl('Available Appointments Calendar'),
        'last' => xl('Last'),
        'occurNames' => [
            xl('1st{{nth}}'),
            xl('2nd{{nth}}'),
            xl('3rd{{nth}}'),
            xl('4th{{nth}}'),
        ],
        'weekDays' => [
            xl('Sunday'),
            xl('Monday'),
            xl('Tuesday'),
            xl('Wednesday'),
            xl('Thursday'),
            xl('Friday'),
            xl('Saturday'),
        ],
    ],
];
```

Then in the `<script>` block:

```php
window.addEditEventConfig = <?php echo json_encode($addEditEventConfig); ?>;
```

This collapses ~30 lines of interleaved PHP/JS into a pure-PHP array + a one-liner. It also naturally produces integer types for `timeDisplayFormat` and `dateDisplayFormat` (via the `(int)` casts) instead of the quoted strings that `js_escape(getString(...))` currently emits — which avoids the loose-equality subtlety on the JS side.

interface/main/calendar/add_edit_event.js
```diff
@@ -0,0 +1,286 @@
...
+    } // Added by epsdky 2016 (details in commit)
+}
+
+// This is for callback by the find-available popup.
+function setappt(year, mon, mday, hours, minutes) {
+    //Infeg Save button should become active once an appointment is selected.
+    $('#form_save').attr('disabled', false);
+    var f = document.forms[0];
+    var dateFormat = addEditEventConfig.dateDisplayFormat;
+    if (dateFormat == 0) {
```
kojiromike: Would it make sense to replace this `if/else if` chain with an object lookup? The three branches only differ in component ordering and separator, and the padding expressions are duplicated across all three.

Something like:

```js
var mon2 = ('' + (mon + 100)).substring(1);
var mday2 = ('' + (mday + 100)).substring(1);
var dateFormats = {
    0: year + '-' + mon2 + '-' + mday2,
    1: mon2 + '/' + mday2 + '/' + year,
    2: mday2 + '/' + mon2 + '/' + year,
};
f.form_date.value = dateFormats[dateFormat];
```

6 lines instead of 13, the format patterns are visible side by side, and the duplicated padding is extracted. Bracket notation does string coercion on the key, so it works whether `dateFormat` is a number or string.

interface/main/calendar/add_edit_event.js
```diff
@@ -0,0 +1,286 @@
...
+        s = document.forms[0].form_provider.value;
+        f = document.forms[0].facility.value;
+    } else {
+        s = document.forms[0].form_provider.options[s.selectedIndex].value;
+        f = document.forms[0].facility.options[f.selectedIndex].value;
+    }
+    var c = document.forms[0].form_category;
+    var formDate = document.forms[0].form_date;
+    var title = addEditEventConfig.translations.availableAppointments;
+    dlgopen(addEditEventConfig.webRoot + '/interface/main/calendar/find_appt_popup.php' +
```
kojiromike: Would it make sense to use `URL` / `URLSearchParams` here instead of string concatenation? The current approach doesn't URL-encode parameter values, so a value with special characters could break the URL.

Something like:

```js
var url = new URL(
    addEditEventConfig.webRoot + '/interface/main/calendar/find_appt_popup.php',
    window.location.origin
);
url.searchParams.set('providerid', s);
url.searchParams.set('catid', c.options[c.selectedIndex].value);
url.searchParams.set('facility', f);
url.searchParams.set('startdate', formDate.value);
url.searchParams.set('evdur', document.forms[0].form_duration.value);
url.searchParams.set('eid', addEditEventConfig.eid);
dlgopen(url.pathname + url.search + extra, '', 725, 200, '', title);
```

The `extra` parameter complicates things a bit since it's a pre-formatted query string fragment, but the core parameters would at least be properly encoded. (Not a blocker — the original code had the same issue — but since we're touching this code anyway it could be a nice incremental improvement.)

haoyu-haoyu:
Thanks @kojiromike — all three suggestions adopted in the latest commit:

1. **PHP array + `json_encode`**: The config is now built as a pure PHP associative array with `xl()` (not `xlj()`) and emitted via a single `json_encode()`. `(int)` casts on `timeDisplayFormat` and `dateDisplayFormat` ensure numeric types.

2. **Object lookup for date formats**: The if/else chain in `setappt()` is replaced with a `dateFormats` lookup table (6 lines vs 13). Added `|| dateFormats[0]` as fallback for safety.

3. **URL/URLSearchParams in `find_available()`**: Core parameters are now properly encoded via the URL API. The `extra` parameter is appended as-is since it's a pre-formatted query string fragment.

kojiromike:
The PHPStan failure is because the two `(int)` casts on `$eid` and `$userid` in the config array push the baseline count from 11 to 13.

Rather than just regenerating the baseline, you can fix the root cause by typing those variables at the point of assignment (line 69/71):

```php
$eid           = filter_input(INPUT_GET, 'eid', FILTER_VALIDATE_INT) ?: 0; // only for existing events
$date          = $_GET['date'] ?? null;        // this and below only for new events
$userid        = filter_input(INPUT_GET, 'userid', FILTER_VALIDATE_INT) ?: 0;
```

`filter_input` with `FILTER_VALIDATE_INT` returns `int|false|null`, and `?: 0` collapses that to `int`. PHPStan knows this signature, so it stops flagging downstream casts.

With `$eid` and `$userid` already typed as `int`, you can drop the `(int)` casts in the config array:

```php
    'eid' => $eid,
    'userId' => $userid,
```

Then `composer phpstan-baseline` to update the counts — the ripple effect actually *reduces* 3 other baseline entries too (`attr()` and `echo` calls that previously received `mixed`), so it's a net improvement beyond just fixing the CI failure.

haoyu-haoyu:
Thanks @kojiromike — applied `filter_input(INPUT_GET, ..., FILTER_VALIDATE_INT) ?: 0` for both `$eid` and `$userid` at lines 69/71, and removed the now-redundant `(int)` casts in the config array.

Note: during review I noticed a couple of pre-existing items (not introduced by this PR) that could be follow-up improvements:
- Line 722 uses `add_escape_custom($eid)` string concatenation instead of parameterized query like the other 12+ SQL statements in this file
- Line 1067 re-assigns `$eid` from `$_REQUEST` without `filter_input`

Happy to address those in a separate PR if useful.

kojiromike:
> Thanks @kojiromike — applied `filter_input(INPUT_GET, ..., FILTER_VALIDATE_INT) ?: 0` for both `$eid` and `$userid` at lines 69/71, and removed the now-redundant `(int)` casts in the config array.
> 
> Note: during review I noticed a couple of pre-existing items (not introduced by this PR) that could be follow-up improvements:
> 
>     * Line 722 uses `add_escape_custom($eid)` string concatenation instead of parameterized query like the other 12+ SQL statements in this file
> 
>     * Line 1067 re-assigns `$eid` from `$_REQUEST` without `filter_input`
> 
> 
> Happy to address those in a separate PR if useful.

Please do open a ticket about those so we can address them.

haoyu-haoyu:
PHPStan baseline regenerated — downloaded the CI artifact and committed.

Also opened #11184 for the two pre-existing items (line 722 SQL concatenation + line 1067 unfiltered `$_REQUEST`).

haoyu-haoyu:
All review feedback has been addressed in the latest commits. Could the reviewers please resolve the conversation threads so the PR can proceed?

Summary of addressed items:
- @kojiromike: ✅ PHP array + json_encode, ✅ date format object lookup, ✅ URL/URLSearchParams, ✅ filter_input for $eid/$userid, ✅ getInt() typed getters, ✅ PHPStan baseline regenerated
- @rollingventures: ✅ userId as int (not string), ✅ form_ampm null guard, ✅ cache-busting, ✅ script load order

The two Semgrep `echoed-request` findings appear to be pre-existing (from the `$_GET['date']` and form action patterns in the original file) — happy to address those in the follow-up issue #11184 if needed.

Issue #11073, opened 2026-03-11
refactor(audit): Update EventAuditLogger to support future integrations


Firehed:
The EventAuditLogger (EAL) as it exists today has several challenges that make upcoming refactors and changes difficult:

- It's coupled to many inline reads of application config (OEGlobalsBag/etc)
- There's almost no type safety
- The logic of what logging systems run and when is hard to follow (lots of `if (!$foo || $bar && $baz)` stuff) coupled with dynamic config reads
- While past efforts have improved this, it's still pretty coupled to the DB internals
- Test coverage is ok, but the tests are fairly brittle and cannot be run in "isolated" mode
- The break glass user subsystem creates

Long story short, EAL is one of the main blockers of next-steps for the DBAL migration (#10149). Making it more modular will improve maintenance and ease of integration:

- Split the log writers out, having them implement a common interface
- Add an intermediate data type for the actual events with type safety
- Improve the structure of the query parsing
- Rework the inline config reads into constructor-provided settings; prep it so it'll work with DI and service autowiring

Possibly other items, but those should move the needle significantly and be enough to unblock further DBAL progress.

PR #11078, opened 2026-03-11
refactor(audit): Extract DB loggers from EventAuditLogger

Firehed:
Towards #11073

This decouples the "log event" from the rest of the event parsing, increasing testability of the actual logging mechanics.

It also uses a second connection specifically for the audit logs. This is necessary to avoid tramping the autoincrement ids since a logged insert in the main code path causes secondary inserts before the original caller returns. Since that's a connection-level property inherent to how the database works, there's no practical workaround in user space other than a second connection. 

However this is beneficial regardless - ideally, the audit trail would be going to a different database with different credentials and limited access. This change doesn't actually include the full set of mechanics to do that, but gets it a whole lot closer.

Note: the previous system seemed to have (or mask?) a quite odd bug with the global state for insert IDs:

> ⏺ Here's the flow:
> 
>   BEFORE (audit logging via ADODB):
>   1. AddressService::insert() calls QueryUtils::sqlInsert() with explicit ID
>   2. ADODB_mysqli_log::Execute() runs the INSERT
>   3. Insert_ID() returns 0 (explicit ID, no auto-increment)
>   4. lastidado is set to 0
>   5. Audit logging runs via ADODB - inserts into log table
>   6. Those inserts go through ADODB_mysqli_log::Execute() too
>   7. The log table uses auto-increment, so Insert_ID() now returns the log's ID (e.g., 778)
>   8. Back in QueryUtils::sqlInsert(), it returns: lastidado > 0 ? lastidado : Insert_ID() → returns 778
>   9. AddressService checks 778 === 0 → false, so it returns $freshId
> 
>   AFTER (audit logging via DBAL):
>   1. Same steps 1-4
>   2. Audit logging runs via separate DBAL connection
>   3. DBAL inserts don't touch ADODB's state
>   4. ADODB's Insert_ID() is still 0
>   5. QueryUtils::sqlInsert() returns: lastidado > 0 ? lastidado : Insert_ID() → returns 0
>   6. AddressService checks 0 === 0 → true, returns false
> 
>   The old code was "working" by accident - the audit logging pollution made sqlInsert return a non-zero value (the wrong ID, but non-zero). The === 0 check was always incorrect for explicit ID inserts; it just never triggered because of this side effect.
> 
> ❯ ok, but I want to be double sure about the logic change for this insert thing. I thought the previous code for the query logger was explicitly bypasing the mechanism where the global insert id was set... or is that not the case? I thought you'd mentioned that earlier.
> 
> ⏺ You're right to double-check. Let me trace through the OLD code path more carefully.
> 
>   The old audit logging used sqlInsertClean_audit() which calls ExecuteNoLog():
> 
>   function ExecuteNoLog($sql, $inputarr = false)
>   {
>       return parent::Execute($sql, $inputarr);  // bypasses ADODB_mysqli_log::Execute
>   }
> 
>   This did bypass:
>   - lastidado being set (that only happens in ADODB_mysqli_log::Execute)
>   - Recursive audit logging
> 
>   But it did NOT bypass:
>   - Insert_ID() on the ADODB connection - because MySQL's LAST_INSERT_ID() is connection-level state that gets updated by ANY insert on that connection
> 
>   So looking at sqlInsert's return:
>   return lastidado > 0 ? lastidado : $db->Insert_ID();
> 
>   OLD flow:
>   - lastidado = 0 (explicit ID insert)
>   - Audit logging via ExecuteNoLog() → MySQL updates LAST_INSERT_ID() on the ADODB connection
>   - Insert_ID() = log table's auto-increment ID
>   - Returns the log table's ID (wrong, but non-zero)
> 
>   NEW flow:
>   - lastidado = 0 (explicit ID insert)
>   - Audit logging via DBAL → completely separate PDO connection
>   - ADODB's Insert_ID() = still 0
>   - Returns 0
> 
>   So the pollution was at the Insert_ID() level, not lastidado. The old code bypassed lastidado but not the underlying MySQL connection state.
> 

- Adds an Audit\Event DTO class to pass data around in a more structured manner
- Adds Audit\LogTablesSink which records the data to the DB
- Moves the stored data encryption logic into same path
- Work around issue (above) in two services that this refactor exposed
- 

Unit tests generated by claude


tests/Tests/Unit/Common/Logging/EventAuditLoggerTest.php
```diff
@@ -1529,7 +1544,8 @@ public function testAuditSQLEventDisabledSpecificEventType(): void
      */
     public function testNullPatientIdHandling(): void
     {
-        // Test with string "NULL"
+        // Test with string "NULL" (there's an internal workaround for getting
```
Firehed, line 1547: This absolutely should not be necessary, but there is/was some odd historical necessity for it. As we continue to improve types, this will go away.

tests/Tests/Unit/Common/Logging/EventAuditLoggerTest.php
```diff
@@ -15,6 +15,7 @@
 
 namespace OpenEMR\Tests\Unit\Common\Logging;
 
+use Doctrine\DBAL\Connection;
```
Firehed, line 18: This test case is closing in on being safe to move to "isolated". There's still a bunch of dependence on globals and sessions, but it's no longer hard-coupled to an active DB connection

Firehed:
What a slog. Got there eventually.

src/Common/Logging/EventAuditLogger.php
```diff
@@ -645,45 +671,19 @@ public function recordLogItem($success, $event, $user, $group, $comments, $patie
...
-        }
+
+        $logTableSink = new Audit\LogTablesSink(
+            conn: $this->connection,
+            crypto: $this->cryptoGen,
+            shouldEncrypt: $shouldEncrypt,
+        );
+        $logTableSink->record($auditEvent);
+
+        // TODO: convert ATNA to same style/interface
```
adunsulag, line 698: Are you planning on doing this as part of this PR? Or in a different PR?  Just wondering if this is still outstanding.  Note at this point $comments no longer has the encoding/encryptoin going into sendAtaAuditMsg with your earlier changes.

Firehed: Oh, good catch. I was planing to convert the ATNA stuff in a fast-follow PR but I don't want to lose that.

Firehed: I've updated the PR so encryption still happens in the main logger path, so the encrypted data should fall through correctly to the ATNA stuff. The rest of the conversion will happen separately (but soon).

PR #11105, opened 2026-03-12
refactor(audit): Make audit destinations ("sinks") configurable

### LINKED ISSUE #11073: refactor(audit): Update EventAuditLogger to support future integrations

Firehed:
Fixes #11073 (I think this makes it "done enough" to unblock more DBAL work, at least)

Makes EventAuditLogger extensible and configurable by injecting sinks rather than building at call-time. Should also yield a performance boost, though not a major one.

- Sinks now configured and injected in singleton setup rather than inline
- In main `recordLogItem` path, there are no longer reads from globals/bag
- ATNA "enabled" is no longer passed in; instead, the sink is added or not based on the value
- There are now isolated tests covering EventAuditLogger in addition to those requiring services

Yes, Claude updated/wrote some of the tests.

### DIFF
--- a/.phpstan/README.md
+++ b/.phpstan/README.md
@@ -49,7 +49,7 @@ $apiKey = $cryptoGen->decryptStandard($globals->get('gateway_api_key'));
 **Purpose:** Prevents use of legacy functions:
 - Legacy `sql.inc.php` functions in the `src/` directory
 - `call_user_func()` and `call_user_func_array()` functions (use modern PHP syntax instead)
-- `error_log()` function (use `SystemLogger` instead)
+- `error_log()` function (use `Psr\Log\LoggerInterface` instead, through OpenEMR\BC\ServiceContainer::getLogger() if needed)
 
 **Rationale for SQL functions:** Contributors should use `QueryUtils` or `DatabaseQueryTrait` instead for modern database patterns.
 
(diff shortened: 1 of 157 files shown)


.phpstan/baseline/argument.type.php
```diff
@@ -75581,6 +75581,11 @@
     'count' => 2,
     'path' => __DIR__ . '/../../tests/Tests/Isolated/Billing/BillingLoggerTest.php',
 ];
+$ignoreErrors[] = [
```
Firehed, line 75484: Intentional addition to cover some weird backwards-compatibility hack in the existing code. Bleh.

PR #11113, opened 2026-03-13
refactor(audit): Get most global state out of the instance path

Firehed:
The EventAuditLogger has depended on intertwined global state. This has made standalone setup for integration into DBAL impossible, since it would either a) have circular dependencies or b) require the previous system to remain in place forever to function.

Reworking that to use DI breaks the cycle, and paves the way for the next set of progress on integration.

- Extract BreakGlass user detection into service
- Extract EventAuditLogger's runtime config into DTO
- Take SessionWrapperInterface in constructor, along with above two items
- Change inline global reads to use injected values
- Update singleton setup so there are no outside changes
- Whole bunch of tests

Yes, Claude

### DIFF
--- a/tests/Tests/Isolated/Common/Logging/BreakglassCheckerTest.php
+++ b/tests/Tests/Isolated/Common/Logging/BreakglassCheckerTest.php
@@ -33,29 +33,29 @@ class BreakglassCheckerTest extends TestCase
         $this->conn->expects($this->never())->method('fetchOne');
 
         $checker = new BreakglassChecker($this->conn);
-        $this->assertFalse($checker->isBreakglassUser(''));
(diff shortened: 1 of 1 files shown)


src/Common/Logging/BreakglassChecker.php
```diff
@@ -0,0 +1,63 @@
...
+        if ($user === '') {
+            return false;
+        }
+
+        if (array_key_exists($user, $this->cache)) {
+            return $this->cache[$user];
+        }
+
+        $sql = <<<'SQL'
+            SELECT 1
```
Firehed, line 50: This has changed very slightly from the previous query to use traditional join syntax

PR #11127, opened 2026-03-16
refactor(issues): move ACL check before database fetch in add_edit_issue.php

### LINKED ISSUE #11126: refactor(issues): move ACL check before database fetch in add_edit_issue.php
In `interface/patient_file/summary/add_edit_issue.php`, the issue row is fetched from the database (line 343) before the ACL check runs (line 344). The ACL check correctly blocks unauthorized access and no data leaks to the response, but as a defense-in-depth measure the fetch should happen after the ACL check so unauthorized requests never load PHI into memory.

Olauge-Secret:
Fixes #11126

In `add_edit_issue.php`, the full issue row (containing PHI) was fetched from the database before the ACL check ran. While the ACL check correctly blocked unauthorized access and no data leaked to the response, as a defense-in-depth measure the fetch should happen after authorization is confirmed.

- Added a lightweight query to fetch only the issue `type` column before the ACL check
- Moved the full row fetch (`PatientIssuesService::getOneById`) to after the ACL check passes
- This ensures unauthorized requests never load PHI into memory

Yes

### DIFF
--- a/interface/patient_file/summary/add_edit_issue.php
+++ b/interface/patient_file/summary/add_edit_issue.php
@@ -309,11 +309,14 @@ if (!empty($_POST['form_save'])) {
 
 $irow = [];
 if ($issue) {
-    $patientIssuesService = new PatientIssuesService();
-    $irow = $patientIssuesService->getOneById($issue);
-    if (!AclMain::aclCheckIssue($irow['type'], '', 'write')) {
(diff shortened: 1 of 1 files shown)


kojiromike:
Closing per the review — after investigating, the ACL check requires the issue type from the database, so there's no way to avoid a query before authorization. The current code is already the right approach.

PR #11130, opened 2026-03-16
fix(logging): SystemLogger PSR-3 compliance and simplification

### LINKED ISSUE #11109: bug(logging): SystemLogger doesn't appear to actually interpolate strings per spec
While overhauling some of the logging infra recently, I discovered that the system logger doesn't seem to actually interpolate log message placeholders per the PSR-3 spec. While in _some_ cases this is reasonable for _some_ loggers (metadata for structured logging, which should generally be additive rather than in replacement), it's pretty much always the wrong behavior for normal text output.

Actual:
```
$ LOG_LEVEL=debug php y.php
[13-Mar-2026 18:19:35 UTC] [2026-03-13T18:19:35.831484+00:00] OpenEMR.DEBUG: connect [] []
[13-Mar-2026 18:19:35 UTC] [2026-03-13T18:19:35.836542+00:00] OpenEMR.DEBUG: {ms}ms: {sql} {"ms":"1.43575","sql":"SELECT * FROM globals","error":"no"} []
```

Expected (give or take):
```
$ LOG_LEVEL=debug php y.php
[13-Mar-2026 18:19:35 UTC] [2026-03-13T18:19:35.831484+00:00] OpenEMR.DEBUG: connect [] []
[13-Mar-2026 18:19:35 UTC] [2026-03-13T18:19:35.836542+00:00] OpenEMR.DEBUG: 1.43575ms: SELECT * FROM globals {"error":"no"} []
```

Since this uses Monolog internally the behavior is pretty surprising, but there's some customization in front of it that may be causing errors.

Firehed:
Fixes #11109

SystemLogger was not interpolating PSR-3 placeholders in log messages. For example, `{ms}ms: {sql}` would appear literally instead of being replaced with context values.

- Add `PsrLogMessageProcessor` to interpolate `{placeholder}` values in log messages per PSR-3 spec
- ~Remove unnecessary HTML escaping of context values (was using `htmlspecialchars` on plain text logs, which had no security value and made logs harder to read)~ Left intentionally for now in the "legacy" path to guard against LFI attacks if a server is configured poorly
- Update DI container (usage pending, needs FC) to use monolog directly, with same configuration
- Use `LoggerTrait` to eliminate boilerplate delegation methods
- Configure `LineFormatter` to include full exception stack traces

Yes

Changes were pair-programmed with Claude Code. Individual commits are marked with `Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>`.


Firehed:
Given basic logger script:

```php
 $ cat x.php
<?php
$c = require __DIR__ . '/bootstrap.php';
$l = $c->get(Psr\Log\LoggerInterface::class);

function level1($l) {
    level2($l);
}

function level2($l) {
    level3($l);
}

function level3($l) {
    $l->critical('asdf {foo} {bar}', [
        'foo' => 'foo',
        'bar' => 'bar',
        'baz' => 'baz',
        'exception' => new RuntimeException('I broke it'),
    ]);
}

level1($l);
```

before:
```
[16-Mar-2026 19:00:58 UTC] [2026-03-16T19:00:58.560603+00:00] OpenEMR.CRITICAL: asdf {foo} {bar} {"foo":"foo","bar":"bar","baz":"baz","exception":"{}"} []
```

after:
```
 $ php x.php
[16-Mar-2026 19:00:21 UTC] [2026-03-16T19:00:21.545525+00:00] OpenEMR.CRITICAL: asdf foo bar {"baz":"baz","exception":"[object] (RuntimeException(code: 0): I broke it at /Users/firehed/dev/github/firehed/openemr-docker-ccda/x.php:18)
[stacktrace]
#0 /Users/firehed/dev/github/firehed/openemr-docker-ccda/x.php(10): level3(Object(OpenEMR\\Common\\Logging\\SystemLogger))
#1 /Users/firehed/dev/github/firehed/openemr-docker-ccda/x.php(6): level2(Object(OpenEMR\\Common\\Logging\\SystemLogger))
#2 /Users/firehed/dev/github/firehed/openemr-docker-ccda/x.php(22): level1(Object(OpenEMR\\Common\\Logging\\SystemLogger))
#3 {main}
"} []
```

adunsulag:
```
Remove unnecessary HTML escaping of context values (was using htmlspecialchars on plain text logs, which had no security value and made logs harder to read)
```
We specifically did this to mitigate any kind of Log File Inclusion (LFI) attack so I would argue that there is indeed security value here.  You could argue that OS hardening should prevent the web server from reading the log (only allowing write) but people don't typically have OpenEMR setup that way.

kojiromike (approved):
In order to make progress on this and not get stuck behind front controller/webapp config, what do you say to making a conditional html sanitizer that we can disable for sensible configurations?

adunsulag:
@kojiromike I'd be fine with that.

Firehed:
I can work with that. 

I think I'll just restore the previous sanitizer logic; the wrapped/FC path ought to be able to bypass SystemLogger entirely in favor of wiring a configured Monolog _directly_ to PSR-3. That will give you the existing behavior for non-FC requests as a safety net, and since the FC is known to go through a sensible server config it won't need it.

kojiromike:
Not a candidate for backport to rel-800 — depends on the logging/SystemLogger refactoring that is only on master.

PR #11143, opened 2026-03-17
fix: Fixes #11142 employer data not saving

### LINKED ISSUE #11142: bug: Saving existing demographics data erases employer data fields
### Describe the bug

The employer demographics data is being erased in employer data fields when you save an existing patient.

### To Reproduce

1. Add a new patient with employer demographic data populated (employer name as a bare minimum)

Image

2. Click on the edit demographics field and see the data populated in the employer tab

Image

3. Hit the save button
4. See the employer data missing on the demographics page, if you edit the demographics and look at the employer tab you can see the data missing as well.

Image

### Expected behavior
The employer data should stay and not be modified.  Tracing this down the function signature was changed in the demographics_save.php for an existing patient and the wrong arguments are being sent in for the change.

### Client configuration

- Browser: Chrome
- OpenEMR version: OpenEMR 8.0.0+ (also in mainline master branch)
- Operating system: Easy Dev Docker environment.

adunsulag:
This fixes the employer data not saving in the demographics save. Replaced the deprecated call with the direct service call.  Also documented the method parameters.

Fixes #11142

### DIFF
--- a/src/Services/EmployerService.php
+++ b/src/Services/EmployerService.php
@@ -113,7 +113,7 @@ class EmployerService extends BaseService
 
     /**
      * @param string $pid The pid of the patient whose employer data we want to update
-     * @param array $new the new employer data we want to save. This should be an associative array where the keys are the column names in the employer_data table and the values are the values we want to save for those columns.
+     * @param array<string,mixed> $new the new employer data we want to save. This should be an associative array where the keys are the column names in the employer_data table and the values are the values we want to save for those columns.
      * @param boolean $create boolean flag to indicate if we are creating a new record or updating an existing record. If true, we will not compare the new data to the old data and will just save it as is. If false, we will compare the new data to the most recent employer data for the patient and only save the values that have changed. This allows us to keep a history of changes for the patient without saving duplicate data.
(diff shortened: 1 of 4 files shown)


src/Services/EmployerService.php
```diff
@@ -111,6 +111,13 @@ public function search($openEMRSearchParameters, $isAndCondition = true): Proces
         return $processingResult;
     }
 
+    /**
+     * @param $pid The pid of the patient whose employer data we want to update
+     * @param $new the new employer data we want to save. This should be an associative array where the keys are the column names in the employer_data table and the values are the values we want to save for those columns.
+     * @param $create boolean flag to indicate if we are creating a new record or updating an existing record. If true, we will not compare the new data to the old data and will just save it as is. If false, we will compare the new data to the most recent employer data for the patient and only save the values that have changed. This allows us to keep a history of changes for the patient without saving duplicate data.
```
kojiromike:
```suggestion
     * @param string $pid The pid of the patient whose employer data we want to update
     * @param array $new the new employer data we want to save. This should be an associative array where the keys are the column names in the employer_data table and the values are the values we want to save for those columns.
     * @param boolean $create flag to indicate if we are creating a new record or updating an existing record. If true, we will not compare the new data to the old data and will just save it as is. If false, we will compare the new data to the most recent employer data for the patient and only save the values that have changed. This allows us to keep a history of changes for the patient without saving duplicate data.
```

PR #11144, opened 2026-03-17
fix(encounter): hide dispensed medications section when inhouse_pharmacy is disabled

### LINKED ISSUE #11098: Dispensed Medications section renders in encounter even when inhouse_pharmacy is disabled
The Dispensed Medications section (heading, table, and "No medications dispensed during this encounter" message) always renders in the encounter summary and patient report, even when the `inhouse_pharmacy` global setting is `0` (disabled).

## Steps to reproduce

1. Ensure `inhouse_pharmacy` global is set to `0` (Admin > Globals > Features > Drugs and Products)
2. Open any encounter
3. Observe the "Dispensed Medications" section with an empty table

## Expected behavior

When `inhouse_pharmacy` is `0`, the Dispensed Medications section should not render at all.

## Actual behavior

The section always renders with "No medications dispensed during this encounter." The data fetch is correctly gated by the setting, but the Twig template partial is unconditionally included.

## Versions

Reproduced on the demo server (8.0.0). Also reported by a community member on 7.0.4 and 8.0.0.

## Forum reference

https://community.open-emr.org/t/missing-custom-layouts-report-and-dispensed-medications/26608

pradhankukiran:
Fixes #11098

The Dispensed Medications section (heading, table, and "No medications dispensed" message) renders in the encounter report even when the `inhouse_pharmacy` global setting is disabled

- Removed the default empty `dispensedMedications` array from the encounter record in `report.php`, so the key is only set when `inhouse_pharmacy` is enabled
- Wrapped the Twig template include with `{% if e.dispensedMedications is defined %}` to skip rendering when the key is absent

Changes were generated with assistance from Claude Code. The modifications are minimal (one line removed, one conditional added) and have been manually verified.

### DIFF
--- a/interface/forms/newpatient/templates/newpatient/report.html.twig
+++ b/interface/forms/newpatient/templates/newpatient/report.html.twig
@@ -10,6 +10,8 @@
     {% if e.referringProvider %}
         <br />{{ "Referring Provider"|xlt }} - {{ e.referringProvider|text }}
     {% endif %}
-    {% include "templates/newpatient/partials/report/_dispensed-medications.html.twig" %}
+    {% if e.dispensedMedications is defined %}
+        {% include "templates/newpatient/partials/report/_dispensed-medications.html.twig" %}
+    {% endif %}
 {% endfor %}
 </div>
--- a/interface/forms/newpatient/report.php
+++ b/interface/forms/newpatient/report.php
@@ -68,7 +68,6 @@ function newpatient_report($pid, $encounter, $cols, $id): void
             'referringProvider' => $referringProvider,
             'posCode' => $posCode,
             'facility' => $facility_name,
-            'dispensedMedications' => []
         ];
         /**
          * @var OEGlobalsBag $globalsBag


interface/forms/newpatient/templates/newpatient/report.html.twig
```diff
@@ -10,6 +10,10 @@
     {% if e.referringProvider %}
         <br />{{ "Referring Provider"|xlt }} - {{ e.referringProvider|text }}
     {% endif %}
-    {% include "templates/newpatient/partials/report/_dispensed-medications.html.twig" %}
+    {# AI-generated code (Claude Code) - start #}
```
kojiromike: The AI comment markers seem like overkill wrapping a 3-line `if`/`include`/`endif`. The OpenEMR policy says "Sections of code that are entirely or mostly generated by AI should be marked" — would you consider this section substantial enough to warrant the markers? The disclosure in the PR body already covers it.

Not a blocker, just a style thought.

```suggestion
    {% if e.dispensedMedications is defined %}
        {% include "templates/newpatient/partials/report/_dispensed-medications.html.twig" %}
    {% endif %}
```

PR #11146, opened 2026-03-17
fix(reports): include custom LBF layout forms in patient report

### LINKED ISSUE #11141: fix(reports): Custom LBF layout forms missing from patient report
### Describe the bug

Custom layout forms (LBF) created via Admin > Forms > Layouts do not appear in the Patient Report (Reports > Patient Report). The forms display correctly within the encounter view, but are absent from report output.

This was introduced by PR #8949 (refactor: rector apply str_starts_with, merged Oct 2025), which converted `strpos($form_name, $var) == 0` to `str_starts_with($form_name, $var)` in `interface/patient_file/report/patient_report.php` at line 436. The conversion is semantically correct, but it exposed a pre-existing bug: custom LBF form titles (e.g. "Custom Assessment Form") are never in the `registry` table, so they can't match any registry name. The old code accidentally worked because `strpos()` returns `false` when not found, and `false == 0` evaluates to `true` in PHP's loose comparison, causing the form name to be remapped to a registry entry.

The real issue is that the print loop (lines 453-461) only iterates over registry names, so any form whose name isn't in the registry is silently dropped.

### To Reproduce

1. Create a custom layout form via Admin > Forms > Layouts (e.g. "Custom Assessment Form")
2. Open a patient encounter and fill in the custom form
3. Go to Reports > Patient Report for that patient
4. Observe that the custom layout form is missing from the encounter's form list

### Expected behavior

Custom LBF layout forms should appear in the Patient Report alongside standard forms like Vitals, SOAP, etc.

### Client configuration

- Browser: Any
- OpenEMR version: 7.0.4+, 8.0.0 (regression from PR #8949)
- Operating system: Any

### Forum reference

https://community.open-emr.org/t/missing-custom-layouts-report-and-dispensed-medications/26608

Also related to #11098, which covers a separate bug from the same forum thread.

pradhankukiran:
Fixes #11141

Custom layout forms (LBF) created via Admin > Forms > Layouts are missing from the Patient Report. This regression was introduced by PR #8949 which converted `strpos($form_name, $var) == 0` to `str_starts_with($form_name, $var)`, exposing a pre-existing bug where the report print loop only iterates over registry table names, silently dropping any form not in the registry.

- Added a second pass after each registry print loop in `patient_report.php` to print forms whose names are not in the registry (e.g. custom LBF layouts)
- Both print loops are patched: the between-encounters loop and the final loop

Changes were generated with assistance from Claude Code. The modifications are minimal (two identical loop blocks added) and have been manually verified.

### DIFF
--- a/interface/patient_file/report/patient_report.php
+++ b/interface/patient_file/report/patient_report.php
@@ -393,6 +393,16 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                                                             }
                                                         }
                                                     }
+                                                    // AI-generated code (Claude Code) - start
+                                                    // Print any forms not in the registry (e.g. custom LBF layouts)
+                                                    foreach ($html_strings as $key => $toprint) {
+                                                        if (!in_array($key, $registry_form_name, true)) {
+                                                            foreach ($toprint as $item) {
+                                                                print $item;
+                                                            }
+                                                        }
+                                                    }
+                                                    // AI-generated code (Claude Code) - end
                                                     $html_strings = [];
                                                     echo "</div>\n"; // end DIV encounter_forms
                                                     echo "</div>\n\n";  //end DIV encounter_data
@@ -459,6 +469,16 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
                                                 }
                                             }
                                         }
+                                        // AI-generated code (Claude Code) - start
+                                        // Print any forms not in the registry (e.g. custom LBF layouts)
+                                        foreach ($html_strings as $key => $toprint) {
+                                            if (!in_array($key, $registry_form_name, true)) {
+                                                foreach ($toprint as $item) {
+                                                    print $item;
+                                                }
+                                            }
+                                        }
+                                        // AI-generated code (Claude Code) - end
                                         ?>
 
                                         <?php


interface/patient_file/report/patient_report.php
```diff
@@ -393,6 +393,16 @@ function show_date_fun(){
                                                             }
                                                         }
                                                     }
+                                                    // AI-generated code (Claude Code) - start
+                                                    // Print any forms not in the registry (e.g. custom LBF layouts)
+                                                    foreach ($html_strings as $key => $toprint) {
+                                                        if (!in_array($key, $registry_form_name)) {
```
kojiromike: Use strict comparison to avoid loose type coercion.

```suggestion
                                                        if (!in_array($key, $registry_form_name, true)) {
```

interface/patient_file/report/patient_report.php
```diff
@@ -459,6 +469,16 @@ function show_date_fun(){
                                                 }
                                             }
                                         }
+                                        // AI-generated code (Claude Code) - start
+                                        // Print any forms not in the registry (e.g. custom LBF layouts)
+                                        foreach ($html_strings as $key => $toprint) {
+                                            if (!in_array($key, $registry_form_name)) {
```
kojiromike:
```suggestion
                                            if (!in_array($key, $registry_form_name, true)) {
```

PR #11147, opened 2026-03-17
refactor(crypto): Decouple key management, versioning, etc [wip]

Firehed:
Fixes #


Firehed:
This is an incomplete subset of #11267, closing.

PR #11152, opened 2026-03-17
fix(documents): view document from clinical view past encounters

### LINKED ISSUE #11151: bug(documents): Unable to load controller DocId, rel-800
### Describe the bug

[refactor with url search params](https://github.com/openemr/openemr/commit/6a4e18c5ec73e0c755f6f65b28a9652aded1a58b#diff-9e3f8573b3dc55b398e78939f05055f5f1360c19f823f28b00614b556fa3970aL224-R231) prevents getting to document from clinical view 

Image

### To Reproduce
after clicking on Document: results in error

### Expected behavior
goes to the Document in the document viewer

Image

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

stephenwaite:
Fixes #11151


interface/patient_file/history/encounters.php
```diff
@@ -227,12 +227,10 @@ function toencounter(rawdata) {
...
 function todocument(docid) {
   const params = new URLSearchParams({
-    doc_id: docid,
-    document: '',
     patient_id: <?php echo js_escape($pid); ?>,
-    view: ''
+    doc_id: docid
   });
-  h = '<?php echo OEGlobalsBag::getInstance()->get('webroot') ?>/controller.php?' + params;
+  h = '<?php echo OEGlobalsBag::getInstance()->get('webroot') ?>/controller.php?document&view&' + params;
```
bradymiller: @kojiromike , I wonder if this bug also crept into some of the other paths that were converted (appears that when a variable is meant to be blank, it does not work as intended).

stephenwaite: good call, I'll fix the issues in the eye form here too

kojiromike: I don't understand this change. Should it be `document=view`?

kojiromike: @bradymiller you mean `$bag->get('webroot')` is different from `$GLOBALS['webroot']` when 'webroot' isn't defined in globals? (Good chance we're hiding a phpstan finding here.)

I've been following up on that PR by slowly adding types, so for example `$bag->getString('webroot')` would definitely be `''` in that case, whereas it's probably `null` here.

kojiromike: Oh, you aren't talking about the GlobalsBag change, are you? You're referring to `URLSearchParams`. I get it now. I still think we should use URLSearchParams as it's _much_ safer than manual concatenation. Let me look into it.

kojiromike: Wait wait, are you telling me that changing the _order_ of url parameters breaks controller.php?

adunsulag: controller.php uses the very first argument as the controller class and the second argument as the method call... yes very stupid but that is the functionality.

adunsulag: See how the controller class (Controller.class.php) functions, yes it is sad.
```
$args = array_reverse(array_keys($qarray));
        $c_name = preg_replace("/[^A-Za-z0-9_]/", "", (string) array_pop($args));
        $parts = explode("_", (string) $c_name);
        $name = "";

        foreach ($parts as $p) {
            $name .= ucfirst($p);
        }

            $c_name = $name;
            $c_action = preg_replace("/[^A-Za-z0-9_]/", "", (string) array_pop($args));
            $args = array_reverse($args);
            ```

adunsulag (approved):

interface/patient_file/history/encounters.php
kojiromike: I'm going to fix this.

stephenwaite: would you rather fix another way or is this okay to bring in @kojiromike ?

kojiromike: @stephenwaite if you are OK with me changing it back after I have the less-sensitive controller dispatcher done, then feel free to merge this.

stephenwaite: Okay, I'm fine leaving this open and then closing without merging if it's not a big overhaul.

adunsulag: I have to do a production hot fix, I'd prefer we bring this in as I need to deploy it today.

adunsulag: Apparently we talked over each other.  @stephenwaite, I'll just take your file change and put it in my own code branch and deploy it.  Once the system I'm working with deploys 8.0.2 or 8.0.3 whatever the end solution is will have the updated code, so its fine if you want to wait.

adunsulag:
Talking with Stephen Waite we won't be able to backport this into rel-800 until the ServiceContainer is brought in. @Firehed is that class file ready to be brought into rel-800 in one of the patches? This updated file now depends on it.

stephenwaite:
It's testing okay in `rel-800` if I add `use OpenEMR\Core\OEGlobalsBag;` to the top of `encounters.php`.

Firehed:
@adunsulag I see no reason it shouldn't be ok to pull in, and I suspect it'll make other patches easier to port across too.

stephenwaite:
closing in favor of https://github.com/openemr/openemr/pull/11485 for rel-800

PR #11157, opened 2026-03-18
ci(test): use static build matrix for faster fan-out

Firehed:

The `collect` job in the test workflow adds latency before matrix jobs can start, as it must dynamically parse docker directories and compute configurations at runtime.

**test-all.yml:**
- Remove dynamic `collect` job entirely
- Replace commit trailer (`Test-Mode: full`) with `test-full` label for triggering full matrix on PRs
- Add `labeled` PR event type so adding the label triggers a re-run
- Split into two jobs that fan out instantly:
  - `test-pr`: PHP 8.2 only (default for PRs)
  - `test-full`: All configurations (push to master, or PRs with `test-full` label)

**test-scheduled.yml:**
- Remove dynamic `collect` job
- Use same static matrix structure as `test-full`

**Behavior:**
- PRs: Instant fan-out of PHP 8.2 tests (apache + nginx)
- PRs with `test-full` label: Instant fan-out of all 16 configurations
- Push to master/rel-*: Instant fan-out of all 16 configurations
- Scheduled: Same as push, with coverage enabled

**Tradeoff:** The matrix is duplicated between test-all.yml and test-scheduled.yml. Comments in both files indicate they should be kept in sync. This is necessary because GitHub Actions does not allow runtime filtering of matrices for reusable workflow calls.

#### If you answered yes:

The workflow changes were generated with assistance from Claude. The commits are co-authored accordingly.

---

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -6,13 +6,10 @@
 #
 # Matrix strategy:
 # - Pull requests: Only PHP 8.2 configurations (oldest supported) for faster feedback
-# - Pull requests with "Test-Mode: full" trailer: All configurations (like nightly)
+# - Pull requests with "test-full" label: All configurations
 # - Push to master/rel-*: All configurations
 #
-# To run all configurations on a PR (useful for debugging flaky tests that only
-# appear in specific configurations), add this trailer to any commit message:
-#
-#   Test-Mode: full
+# To run all configurations on a PR, add the "test-full" label.
 #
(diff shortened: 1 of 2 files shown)


Firehed:
This is, of course, proving quite persnickety. There's a lot of setup code that would negate the benefit of the earliest-possible fanout.

Firehed:
I think #11178 is probably the better direction to go rather than trying to untangle the existing system.

PR #11163, opened 2026-03-18
fix(controller): add order-independent routing via dispatch() method

### LINKED ISSUE #11162: refactor(controller): make routing independent of query parameter order
The legacy `Controller::act()` method determines the controller and action from the **order** of query parameters, not their names. This is fragile and surprising.

For example, the URL `?document&view&patient_id=123` works, but `?patient_id=123&document&view` routes to the wrong controller because `patient_id` appears first.

This caused #11151 where using `URLSearchParams` in JavaScript produced parameters in a different order than expected, breaking document viewing with the error "Unable to load controller DocId".

**Current behavior:**
- First query key → controller name
- Second query key → action name
- Relies on `array_keys()` order from `$_GET`

**Proposed fix:**
Add explicit `controller` and `action` parameters:
```
controller.php?controller=document&action=view&patient_id=123
```

A `dispatch()` method has been drafted that supports this pattern. The existing `act()` method can remain for backwards compatibility while call sites are migrated incrementally.

kojiromike:
## Summary

The legacy `act()` method determines controller and action from query parameter **order**, which is fragile and breaks when JavaScript's `URLSearchParams` produces a different order.

**Changes:**
- Add `dispatch()` method using explicit `controller` and `action` params
- Add `VALID_CONTROLLERS` whitelist instead of regex sanitization
- Add `ControllerInterface` marker interface for future PSR-4 migration
- Use `is_callable()` for action validation instead of regex
- Migrate all internal callers from `act()` to `dispatch()`
- Add proper HTTP exceptions (400/404) instead of `exit()`
- Add error reference IDs for troubleshooting 500 errors

`controller.php` detects URL style and routes to appropriate method, maintaining backwards compatibility with existing URLs.

## Test plan

- [ ] Verify existing URLs still work (e.g., `controller.php?document&view&patient_id=123`)
- [ ] Verify new explicit URLs work (e.g., `controller.php?controller=document&action=view&patient_id=123`)
- [ ] Test prescription, practice_settings, and document controllers
- [ ] Verify 400 response for missing/invalid controller
- [ ] Verify 404 response for non-existent action

Fixes #11162

### DIFF
--- a/src/Core/ControllerInterface.php
+++ b/src/Core/ControllerInterface.php
@@ -1,0 +1,19 @@
+<?php
+
+/**
+ * Marker interface for legacy controllers.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc. <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Core;
+
+interface ControllerInterface
+{
+}
--- a/.phpstan/baseline/array.invalidKey.php
+++ b/.phpstan/baseline/array.invalidKey.php
@@ -3,11 +3,6 @@
 $ignoreErrors = [];
 $ignoreErrors[] = [
     'message' => '#^Possibly invalid array key type mixed\\.$#',
-    'count' => 7,
-    'path' => __DIR__ . '/../../controllers/C_PracticeSettings.class.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Possibly invalid array key type mixed\\.$#',
     'count' => 2,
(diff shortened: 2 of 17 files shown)


src/Core/ControllerInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc. <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Core;
+
+interface ControllerInterface
```
Firehed, line 17: This is intentionally empty?

kojiromike: Yeah, for now it's just a marker interface. Most of the downstream implementations are pretty dynamic.

kojiromike:
Not a candidate for backport to rel-800 — depends on refactoring that is only on master.

PR #11168, opened 2026-03-18
fix(session): prevent double session start with read_and_close

zmilan:
Improves #10931

- **Session:** add coreSessionStarted flag to prevent `globals.php` from re-starting a `read_and_close` session that `PHPSessionWrapper` already started. Fix `sessionAllowWrite` scope in SiteSetupListener.
- **Crypto:** remove stale `null` argument from 8 `decryptStandard` calls in
  `C_Document` that broke after the signature change in 7f33297a5.

### DIFF
--- a/controllers/C_Document.class.php
+++ b/controllers/C_Document.class.php
@@ -719,7 +719,7 @@ class C_Document extends Controller
                 die(xlt("File retrieval from CouchDB failed"));
             }
             if ($d->get_encrypted() == 1) {
-                $filetext = $this->cryptoGen->decryptStandard($content, null, KeySource::Database);
+                $filetext = $this->cryptoGen->decryptStandard($content, KeySource::Database);
             } else {
                 $filetext = base64_decode((string) $content);
             }
@@ -758,7 +758,7 @@ class C_Document extends Controller
                 $couchM = new CouchDB();
                 $respM = $couchM->retrieve_doc($couch_docid);
                 if ($d->get_encrypted() == 1) {
-                    $contentM = $this->cryptoGen->decryptStandard($respM->data, null, KeySource::Database);
+                    $contentM = $this->cryptoGen->decryptStandard($respM->data, KeySource::Database);
                 } else {
                     $contentM = base64_decode((string) $respM->data);
                 }
@@ -801,14 +801,14 @@ class C_Document extends Controller
                 $couchF = new CouchDB();
                 $respF = $couchF->retrieve_doc("converted_" . $couch_docid);
                 if ($d->get_encrypted() == 1) {
-                    $content = $this->cryptoGen->decryptStandard($respF->data, null, KeySource::Database);
+                    $content = $this->cryptoGen->decryptStandard($respF->data, KeySource::Database);
                 } else {
                     $content = base64_decode((string) $respF->data);
                 }
             } else {
                 // decrypt/decode when converted jpg already exists
                 if ($d->get_encrypted() == 1) {
-                    $content = $this->cryptoGen->decryptStandard($resp->data, null, KeySource::Database);
+                    $content = $this->cryptoGen->decryptStandard($resp->data, KeySource::Database);
                 } else {
                     $content = base64_decode((string) $resp->data);
                 }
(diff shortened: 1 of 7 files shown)


interface/globals.php
```diff
@@ -235,10 +235,14 @@ function GetCallingScriptName()
...
 $read_only = empty($sessionAllowWrite);
 $session = SessionWrapperFactory::getInstance()->getWrapper();
-if (session_status() === PHP_SESSION_NONE && !$session->isSymfonySession()) {
-    //error_log("1. LOCK ".GetCallingScriptName()); // debug start lock
+$factory = SessionWrapperFactory::getInstance();
+// PHPSessionWrapper starts the session in its constructor (respecting sessionAllowWrite via
+// OEGlobalsBag). Only start here if no wrapper has started a session yet — this prevents the
+// double session-file read that occurred when read_and_close closed the session and
+// session_status() returned PHP_SESSION_NONE.
+if (!$factory->isCoreSessionStarted() && session_status() === PHP_SESSION_NONE && !$session->isSymfonySession()) {
```
kojiromike, line 252: This `coreSessionStarted` guard prevents the fallback session start, but many core scripts set `$sessionAllowWrite = true` before including `globals.php` and rely on this fallback to get a writable session. For example:

- `interface/main/main_screen.php` sets `$sessionAllowWrite = true` and then writes directly to `$_SESSION` at lines 398, 408, 472, 475, 479
- `interface/login/login.php`, `interface/logout.php`, `interface/usergroup/mfa_totp.php`, `sql_upgrade.php`, and many others do the same

The problem: `PHPSessionWrapper` reads `sessionAllowWrite` from `OEGlobalsBag`, where it's **not set** for core requests — only the `SiteSetupListener` (REST/API path) sets it there. So `PHPSessionWrapper` always starts with `read_and_close = true` for core requests, the session closes immediately, and now this guard prevents `globals.php` from re-starting with write access. Direct `$_SESSION` writes after that would silently not persist.

Would it work to propagate the caller's `$sessionAllowWrite` to `OEGlobalsBag` **before** creating the wrapper? Something like:

```suggestion
$factory = SessionWrapperFactory::getInstance();
// Propagate the caller's sessionAllowWrite into OEGlobalsBag so PHPSessionWrapper
// starts the session with the correct read/write mode in the first place.
if (!empty($sessionAllowWrite)) {
    $globalsBag->set('sessionAllowWrite', true);
}
// PHPSessionWrapper starts the session in its constructor (respecting sessionAllowWrite via
// OEGlobalsBag). Only start here if no wrapper has started a session yet — this prevents the
// double session-file read that occurred when read_and_close closed the session and
// session_status() returned PHP_SESSION_NONE.
if (!$factory->isCoreSessionStarted() && session_status() === PHP_SESSION_NONE && !$session->isSymfonySession()) {
```

That way `PHPSessionWrapper` starts with the correct mode, there's still only one session start, and callers that need write access get it.

kojiromike (commented):
The crypto changes are correct — decryptStandard's second parameter changed from a nullable $passphrase string to a non-nullable KeySource enum in https://github.com/openemr/openemr/commit/7f33297a53d0475a8d25083c7b5ee5e4bf9f7dc1, so removing the stale null argument is the right fix. No remaining stale calls on master.

The session changes have a potential regression, though — see inline comment.

interface/globals.php
zmilan: @kojiromike thanks for discovering this. I implemented your suggestion and added tests for such a scenario in commit https://github.com/openemr/openemr/pull/11168/commits/f2377e4d6c306ae5da064e75210a53538d725518

kojiromike:
@zmilan if you like, maybe you could make the crypto changes in one pr and the session in another?

zmilan:
> @zmilan if you like, maybe you could make the crypto changes in one pr and the session in another?

I will do it later today... this should be a separate PR for sure...

controllers/C_Document.class.php
stephenwaite, line 722: odd because this was already fixed in https://github.com/openemr/openemr/pull/11155/changes/a9fd4f761cd3906fde35498d00531ff9aec93692

kojiromike: GitHub's diffing seems really janky now. Look at https://github.com/openemr/openemr/pull/11171, and then look at what's actually in that file in master.

PR #11171, opened 2026-03-18
chore: Updating Changelog

adunsulag:
Updating changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,20 +5,21 @@
 
 #### Security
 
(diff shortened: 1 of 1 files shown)


kojiromike:
@adunsulag those links should go to the advisory, but they look like they go to the private fork. Do you want to cherry-pick the CL from rel-800 instead?

adunsulag:
@kojiromike yea, probably be simpler at this point to just cherry-pick it.

Firehed:
I think we did this at the same time; it should already be sorted through #11176 unless there's something different here that I'm missing?

adunsulag:
Closing this as I believe the change was already incorporated into changelog and the fix is now redundant.

PR #11178, opened 2026-03-19
refactor(test): another attempt to improve test speed+parallelism

Firehed:
Bypasses most docker and bootstrapping on tests that support it


Firehed:
Latest version (7478e8d) has this showing promising results. It looks like some of the data seeding is incomplete (ACLs, etc), but I got test runs with feedback before the previous matrix even started to fan out. Runtimes of the actual "do work" jobs are comparable or marginally faster than the existing versions, without the earlier setup overhead.

kojiromike:
heh, I tried this [once](https://github.com/kojiromike/openemr/pull/29/changes). Great minds!

Firehed:
This is more or less done through #11701. There's more room for improvement in other test suites but I think that serves as a clear path forward. Closing.

PR #11181, opened 2026-03-19
chore(testing): show PHPUnit deprecation details in test output

### LINKED ISSUE #11116: chore(testing): Avoid suppression of warnings, notices, etc.
We should be running tests with the `--show-all-issues` flag to avoid suppressing valuable feedback that could result in future breakage. It won't automatically mean warnings cause failures (that's a separate discussion, though also valid) but we're less likely to be caught off-guard during updates to test tools.

See #11074

Desel72:
Fixes #11116

PHPUnit's own deprecation warnings are suppressed by default, hiding valuable feedback that could result in future breakage during test tool updates.

- Add `displayDetailsOnPhpunitDeprecations="true"` to both `phpunit.xml` and `phpunit-isolated.xml`, completing the set of `displayDetailsOn*` attributes already enabled for errors, warnings, notices, and PHP deprecations.

### DIFF
--- a/phpunit-isolated.xml
+++ b/phpunit-isolated.xml
@@ -23,6 +23,7 @@ This file configures the kind that does not require initialization.
          displayDetailsOnTestsThatTriggerErrors="true"
          displayDetailsOnTestsThatTriggerWarnings="true"
          displayDetailsOnTestsThatTriggerNotices="true"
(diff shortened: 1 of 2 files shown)


Daedalus-Icarus:
@Firehed Could you please review this PR?

Daedalus-Icarus:
@Firehed Thank you for merging.

PR #11197, opened 2026-03-19
feat: remove MedEx from core, preserve recall and flow board functionality

### LINKED ISSUE #11182: Time to Remove MedEx
It is time to remove the embedded MedEx code from the codebase.  It is going to move into a custom module.

ophthal:
Reviewed auto failed things and removed the incomplete module files which will be added as a third party...

- Extract recall board UI into PSR-4 services (DisplayService, AddRecallForm, RecallService)
- Add CSRF protection to all AJAX endpoints (save.php, save_recall.php, get_patient_data_recall.php)
- Fix HTML escaping (attr, text, xlt/xl, xla, xlj, js_escape) throughout recall board
- Fix PHP 8.2 type safety issues (int comparisons, null coalescing, string casts)
- Fix SQL zero-date defaults (0000-00-00 -> CURRENT_TIMESTAMP)
- Remove MedEx library files (MedEx.php, MedEx_background.php) - replaced by stub API.php
- Update phpstan baselines for all modified files
- Remove MedEx references from calendar, patient tracker, demographics, globals

Fixes #11182


ophthal:
There were some Recall Board fixes that needed to be completed to finish the dissection...

.phpstan/baseline/argument.byRef.php
```diff
@@ -1,5 +1,7 @@
 <?php declare(strict_types = 1);
 
+// total 10 errors
```
Firehed: The addition of these is indicating that the baselines are being generated incorrectly. Make sure to use the `composer phpstan-baseline` command when updating.

Note: we made this change since these counters were generating a _lot_ of merge conflicts.

interface/main/messages/ajax/save_recall.php
```diff
@@ -0,0 +1,112 @@
...
+}
+
+if (!empty($missing)) {
+    (new SystemLogger())->warning("Missing required fields", ['fields' => $missing]);
+    http_response_code(400);
+    echo json_encode([
+        'success' => false,
+        'message' => xl('Missing required fields') . ": " . implode(', ', $missing),
+        'received_fields' => array_keys($input)
+    ]);
```
kojiromike: Exposing `received_fields` (the POST field names) in error responses leaks implementation details. Could this be removed from the production response?

```suggestion
            'error' => 'Missing required fields: pid, r_eventDate, r_provider',
```

interface/main/messages/ajax/save_recall.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+/**
+ * AJAX endpoint to save a new recall
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Ray Magauran <magauran@medfetch.com>
+ * @copyright Copyright (c) 2026 OpenEMR Foundation
```
kojiromike: Minor: `Ray Magauranuman` looks like a copy-paste artifact — should this be `Ray Magauran`?

```suggestion
 * @author    Ray Magauran
```

interface/main/messages/save.php
```diff
@@ -202,60 +81,96 @@
...
-    echo text(json_encode($pidList));
+    echo json_encode(['success' => true, 'item' => $item, 'pids' => $pidList]);
     exit;
 }
-if ($_REQUEST['go'] == "Messages") {
-    if ($_REQUEST['msg_id']) {
-        $result = updateMessage($_REQUEST['msg_id']);
-        echo json_encode($result);
-        exit;
+if (($_REQUEST['action'] ?? '') == 'save_postcard_template') {
```
kojiromike: test - should land on save_postcard_template

interface/main/messages/ajax/get_patient_data_recall.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/**
+ * AJAX endpoint to get patient data for recall form
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Ray Magauranuman <magauran@medfetch.com>
```
kojiromike: test - should land on Magauranuman

kojiromike: Minor: `Ray Magauranuman` looks like a copy-paste artifact — should this be `Ray Magauran`?

```suggestion
 * @author    Ray Magauran <magauran@medfetch.com>
```

Firehed (commented):
I'll get in with a more substantive review soon enough, but I wanted to point out a couple things that quickly stuck out.

interface/main/messages/save.php
```diff
@@ -202,60 +81,96 @@
...
+                [$r_eventDate, $r_provider, $r_facility, $r_reason, $r_pid]
+            );
+        } else {
+            sqlStatement(
+                "INSERT INTO patient_recalls (r_pid, r_eventDate, r_provider, r_facility, r_reason, r_created) VALUES (?, ?, ?, ?, ?, NOW())",
+                [$r_pid, $r_eventDate, $r_provider, $r_facility, $r_reason]
+            );
+        }
+
+        // Update patient demographics (phone, email, address, HIPAA preferences)
```
kojiromike: Should this have an ACL check before updating `patient_data`? The new `ajax/save_recall.php` properly checks `AclMain::aclCheckCore('patients', 'appt')` but this endpoint doesn't check any ACL at all. Writing patient demographics (phone, email, address, HIPAA preferences) typically requires `patients/demo/write`.

Also — `save_recall.php` delegates to `RecallService::createRecall()` which has its own patient_data update logic. Having two code paths that both update demographics creates a maintenance risk. Could this delegate to the same `RecallService` method?

Finally, please use QueryUtils instead of `sqlStatement` here.

interface/main/messages/save.php
kojiromike, line 171: Writing directly to the `globals` table without an admin-level ACL check seems like it could be a privilege escalation — any user who can reach this endpoint could modify global settings. Should this require `AclMain::aclCheckCore('admin', 'super')` or similar?

src/Views/RecallBoard/DisplayService.php
```diff
@@ -0,0 +1,880 @@
...
+    }
+
+    public function display_recalls()
+    {
+        $rcb_selectors = $this->rcb_selectors;
+        $rcb_facility = $this->rcb_facility;
+        $rcb_provider = $this->rcb_provider;
+
+        // Initialize form variables
+        $form_patient_id = $_REQUEST['form_patient_id'] ?? '';
```
kojiromike, line 37: A PSR-4 service class reading `$_REQUEST` and `$_SESSION` directly couples it to the HTTP layer. Would it make sense to pass these values in as method parameters instead? That way the service is testable and reusable outside a web request context.

Same applies to the `$_SESSION` access nearby — new code should use `SessionWrapperFactory::getInstance()->getWrapper()->get('key')` per project conventions.

src/Views/RecallBoard/DisplayService.php
```diff
@@ -0,0 +1,880 @@
...
+ * @copyright Copyright (c) 2024-2026 OpenEMR <dev@open-emr.org>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Services\RecallBoard;
+
+use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Core\OEGlobalsBag;
+
+class DisplayService
```
kojiromike, line 22: This class and `AddRecallForm` render HTML directly (via `echo` and `ob_start()`/`ob_get_clean()`), which makes them view/presenter classes rather than services. Having them under `src/Services/RecallBoard/` is misleading — could they move to something like `src/Views/RecallBoard/` or be converted to Twig templates? Mixing SQL queries, business logic, and HTML output in one class makes it very hard to test or maintain.

src/Services/RecallBoard/DisplayService.php
```diff
@@ -0,0 +1,880 @@
...
+
+        // Allow any module to inject assets into the recall board by providing a 'recall_buttons_loader.php'
+        $scan_dirs = [
+            $globals->get('srcdir') . '/../modules',
+            $globals->get('srcdir') . '/../interface/modules/custom_modules'
+        ];
+        
+        foreach ($scan_dirs as $modules_dir) {
+            if (is_dir($modules_dir)) {
+                $modules = scandir($modules_dir);
```
kojiromike: Using `scandir()` + `include_once` to load module files is a file inclusion pattern that bypasses the established module/event system. Is there a reason this can't use the standard module hooks instead? The silent `catch (\Throwable $e)` also means any errors from included files are swallowed silently.

src/Services/RecallBoard/RecallService.php
```diff
@@ -0,0 +1,218 @@
...
+     * Delete a patient recall
+     *
+     * @param int|null $pid Patient ID
+     * @param int|null $r_ID Recall ID
+     * @return void
+     */
+    public static function deleteRecall(?int $pid = null, ?int $r_ID = null): void
+    {
+        // Get from POST if not provided (for backward compatibility)
+        $pid = $pid ?? ($_POST['pid'] ?? null);
```
kojiromike: A service-layer method falling back to `$_POST` for its parameters is surprising — callers should pass validated values explicitly. If `$pid` and `$r_ID` are both null, the early return handles it, but the superglobal access here couples this to HTTP context unnecessarily. Consider removing the `$_POST` fallback and requiring callers to pass the values.

src/Services/RecallBoard/RecallService.php
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Calculate patient age from date of birth
+     *
+     * @param string $dob Date of birth (YYYY-MM-DD)
+     * @param string $asof Calculate age as of this date (default: today)
+     * @return int Age in years
+     */
+    public static function getAge(string $dob, string $asof = ''): int
```
kojiromike: All methods in this class are static, which makes it untestable (can't mock/stub) and incompatible with dependency injection. Meanwhile `src/Services/RecallService.php` is a proper instance class extending `BaseService` that does similar CRUD on the same `patient_recalls` table.

Having two `RecallService` classes with the same short name in related namespaces will cause confusion. Could these be consolidated into one?

src/Services/RecallService.php
```diff
@@ -0,0 +1,208 @@
...
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+namespace OpenEMR\Services;
+
+use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Common\Logging\SystemLogger;
+use OpenEMR\Services\BaseService;
+
+class RecallService extends BaseService
```
kojiromike, line 23: This looks like the right approach (instance class, extends `BaseService`, proper DI). But its existence alongside `RecallBoard\RecallService` with overlapping functionality is confusing. Could the static methods from `RecallBoard\RecallService` be migrated here as instance methods?

interface/main/messages/ajax/save_recall.php
```diff
@@ -0,0 +1,112 @@
...
+
+use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Logging\SystemLogger;
+use OpenEMR\Services\RecallService;
+
+header('Content-Type: application/json');
+
+// Enable error logging
+ini_set('display_errors', 0);
```
kojiromike: These `ini_set` calls override the global error handling configuration. Was this left over from debugging? The application handles error display/logging centrally.

interface/main/messages/ajax/save_recall.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+if (!empty($missing)) {
+    (new SystemLogger())->warning("Missing required fields", ['fields' => $missing]);
+    http_response_code(400);
+    echo json_encode([
+        'success' => false,
+        'message' => xl('Missing required fields') . ": " . implode(', ', $missing),
+        'received_fields' => array_keys($input)
```
kojiromike: Exposing `received_fields` (the POST field names) in error responses leaks implementation details. Could this be removed from the production response?

kojiromike (changes requested):
Thanks for taking on MedEx removal — this is needed and appreciated. A few concerns below, mostly around architecture and a couple of security items that should be addressed before merge.

Also noting that CI has several failures (PHPStan, Rector, Styling PHP/JS, Whitespace, Database Version Check, Semgrep OSS) that will need to be resolved.

**Architecture:** The new `RecallBoard\DisplayService` and `AddRecallForm` classes live under `src/Services/` but they render HTML directly — they're view/presenter classes, not services. And there are two `RecallService` classes with overlapping responsibilities. Worth consolidating before this merges, since it'll be harder to untangle later.

**Duplicate code paths:** `save.php` and `ajax/save_recall.php` both implement recall creation with different validation and ACL enforcement. One should delegate to the other (or both to `RecallService`).

**Scope:** This PR would benefit from a rebase onto current master — there have been several recent security fixes that will need to be reconciled during merge. The baseline regeneration will be cleaner after a rebase too.

src/Views/RecallBoard/DisplayService.php
```diff
@@ -0,0 +1,866 @@
...
+                <i class="far fa-square fa-stack-2x"></i>
+                <i id="print_caret" class='fas fa-caret-<?php echo $caret = ($rcb_selectors === 'none') ? 'down' : 'up'; ?> fa-stack-1x'></i>
+            </span>
+
+            <div class="tab-content">
+               <div class="tab-pane active" id="tab-all">
+                    <?php
+                        $this->recall_board_top();
+                    if (is_array($processed) && isset($processed['ALL'])) {
+                        echo $processed['ALL'];
```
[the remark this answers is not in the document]
ophthal: Ok This is  the rebase on to the master.  Steep learning curve for me.  It has 2  parts.  Tests pending.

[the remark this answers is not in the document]
ophthal: Test results processed and code updated.

ChatGPT - Building a Scheduling App_files/ansi-1f6vhsjh(1).css
kojiromike: Did you intend to commit this entire subdirectory?

kojiromike:
@ophthal taking https://github.com/openemr/openemr/pull/11265 as an example, I recommend you consider shipping a oe-module-medex as a separate composer installable package, listed on packagist. Then reduce this PR to

1. Removing MedEx from core.
2. Adding oe-module-medex to composer.json

This will provide the fastest path to merging this and several other related benefits you can find under the heading _Why this is better than the original submission_ in the linked PR.

PR #11198, opened 2026-03-20
fix(users): allow clearing email field in user admin update

rollingventures:
## Summary

Followup to #11056 — addresses review feedback from @bradymiller and @kojiromike.

- Removes the `$email !== ''` guard on the UPDATE path so admins can clear a previously-set email field
- The INSERT path (new user creation) already writes the email value unconditionally — this makes the two paths consistent

## Test plan

- [x] Edit a user, set an email address, save — email should persist
- [x] Edit the same user, clear the email field, save — email should be cleared
- [x] Create a new user with and without email — both should work as before


rollingventures:
@bradymiller - done. please merge when you feel like it.

bradymiller:
looks great!

Issue #11179, opened 2026-03-19
chore(db): Remove 0000-00-00 date defaults for MySQL strict mode compatibility


Firehed:
## Summary

The database schema uses `'0000-00-00'` and `'0000-00-00 00:00:00'` as default values for date/datetime columns. This is incompatible with MySQL's default strict SQL mode (`NO_ZERO_DATE`), which has been enabled by default since MySQL 5.7.

Currently, OpenEMR works around this by configuring MySQL with relaxed SQL modes (e.g., `NO_ENGINE_SUBSTITUTION` only). This prevents running against a stock MySQL 8+ installation without custom configuration.

## Affected Schema

**CREATE TABLE defaults:**
- `openemr_postcalendar_events.pc_eventDate` - `date NOT NULL default '0000-00-00'`
- `openemr_postcalendar_events.pc_endDate` - `date NOT NULL default '0000-00-00'`
- `medex_recalls.r_created` - `timestamp NOT NULL DEFAULT '0000-00-00 00:00:00'`

**INSERT statements with zero dates:**
- `document_templates` seed data uses `'0000-00-00 00:00:00'` for `send_date` and `end_date`

## Code Impact

The value `'0000-00-00'` is used semantically throughout the codebase to mean "no date" or "ongoing/open-ended". A grep finds ~50+ locations checking for this value:

```php
// Examples:
"(pc_endDate = '0000-00-00' AND pc_eventDate >= ? ...)"  // portal/find_appt_popup_user.php
$event['pc_endDate'] = '0000-00-00';                      // library/MedEx/API.php
if ($value == '0000-00-00') ...                           // many form reports
```

Key files with zero-date logic:
- `library/sanitize.inc.php` - helper functions explicitly check for `'0000-00-00'`
- `library/patient.inc.php` - `fixDate()` defaults to `'0000-00-00'`
- `src/Services/FHIR/FhirAppointmentService.php`
- `portal/find_appt_popup_user.php`

## Proposed Solution

1. **Schema migration**: Change affected columns to `DEFAULT NULL` (requires making them nullable) or remove the default entirely
2. **Data migration**: Convert existing `'0000-00-00'` values to `NULL`
3. **Code updates**: Change comparisons from `== '0000-00-00'` to `IS NULL` or use a helper function
4. **Seed data**: Update INSERT statements to use `NULL` instead of zero dates

## Related Issues

- #10890 - Migrate TIMESTAMP columns to DATETIME (related cleanup)
[...]

PR #11210, opened 2026-03-20
fix(schema): reduce onetime_auth index prefix length

Firehed:
Towards the broad goal of #11179

With utf8mb4 (4 bytes/char), a 255 character prefix on the `onetime_token` column in the `pid` index exceeds InnoDB's maximum key length in MySQL strict mode.

- Reduce the index prefix from 255 to 32 characters in `sql/database.sql`
- Add upgrade migration in `sql/8_0_0-to-8_0_1_upgrade.sql` to fix existing installations

#### If you answered yes:

This PR was generated with assistance from Claude (Anthropic). The changes are minimal schema modifications.


Firehed:
Note: this should address the straight-shot import of database.sql (for the affected section) but upgraders might still experience weird behavior. I'm not terribly worried since the overarching goal is to enable a clean-from-day-one schema for doctrine/migrations, which as a starting point won't care about the patches.

PR #11211, opened 2026-03-20
fix(schema): correct empty string in list_options integer columns

Firehed:
Towards the broad goal of #11179

The English nationality row in `list_options` had empty strings for `is_default` and `option_value` columns. MySQL strict mode rejects empty strings for integer columns.

Note: anyone who had run this change already has the correct data in place; mysql in looser modes interpreted the empty string as `0`. It doesn't require a forward migration, this just enables a small part of the existing scripts to run in strict mode.

- Fix the INSERT in `sql/database.sql` to use `'0'` instead of `''`
- Fix the same INSERT in `sql/7_0_1-to-7_0_2_upgrade.sql`

No data migration needed since existing databases would have cast the empty string to 0 on insert.

#### If you answered yes:

This PR was generated with assistance from Claude (Anthropic). The changes are minimal data corrections.


Firehed:
Other stuff about this entry is weird compared to the surrounding:
- `option_id` is `English` instead of `EN`
- `codes` is empy

But those a) I'm not sure what the implications are and b) require an actual migration to fix the data, if anything needs fixing. This is just allowing the existing SQL to run properly.

sql/7_0_1-to-7_0_2_upgrade.sql
```diff
@@ -413,7 +413,7 @@ INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'EC', 'Ecuadorean', '680', '0', '0', '', 'Ecuador', 'ECU:218');
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'EG', 'Egyptian', '690', '0', '0', '', 'Egypt', 'EGY:818');
 INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'AE', 'Emirati', '700', '0', '0', '', 'United Arab Emirates', 'ARE:784');
-INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'English', 'English', '710', '', '', '', 'England', '');
+INSERT INTO `list_options` (`list_id`, `option_id`, `title`, `seq`, `is_default`, `option_value`, `mapping`, `notes`, `codes`) VALUES('nationality_with_country', 'English', 'English', '710', '0', '0', '', 'England', '');
```
kojiromike, line 416: It's too late to change 701-702, isn't it? This should be considered frozen, no?

bradymiller: ok to fix here if this causes issues on upgrade.
Will upgraders from 8.0.0 to 8.0.1 need adjustment to their database though? (or have they already been set to 0 even though blank was used?)

bradymiller: btw, answered my own questions above by reading the PR notes, so can disregard them.

bradymiller (approved):

PR #11212, opened 2026-03-20
fix(schema): use CURRENT_TIMESTAMP default for medex_recalls.r_created

Firehed:
Ref: #11179

The `medex_recalls.r_created` column uses `'0000-00-00 00:00:00'` as a default value, which is incompatible with MySQL strict mode (NO_ZERO_DATE).

- Change default from `'0000-00-00 00:00:00'` to `CURRENT_TIMESTAMP` in `sql/database.sql`
- Fix the same in `sql/5_0_0-to-5_0_1_upgrade.sql` (where the table was introduced)
- Add upgrade migration in `sql/8_0_0-to-8_0_1_upgrade.sql` for existing installations

Note: there's nothing in the code that I could find that reads this data. Or writes it, for that matter.

#### If you answered yes:

This PR was generated with assistance from Claude (Anthropic). The changes are schema modifications only.

### DIFF
--- a/sql/5_0_0-to-5_0_1_upgrade.sql
+++ b/sql/5_0_0-to-5_0_1_upgrade.sql
@@ -772,7 +772,7 @@ CREATE TABLE `medex_recalls` (
   `r_facility` int(11) NOT NULL,
   `r_provider` int(11) NOT NULL,
   `r_reason` varchar(255) DEFAULT NULL,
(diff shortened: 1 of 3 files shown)


kojiromike:
Are you following #11197? (not just 111**79**)

Firehed:
> Are you following #11197? (not just 111**79**)

I'm aware of it, but haven't gone through it with a fine-toothed comb.

Issue #11202, opened 2026-03-20
Chart notes display portal_username instead of patient name for recipient


kojiromike:
When a clinician sends a chart note to a patient via the Messages interface, the recipient dropdown shows the patient's full name. But after sending, the message summary displays the internal `portal_username` (e.g., "admin to zhportal") instead of the patient's name.

This makes it hard to reconcile who the message was sent to, since the identifier shown after sending doesn't match what was selected.

PR #7556, opened 2024-07-12
Portal Updates for Usability study, UI/UX and bug fixes.

sjpadgett:
- bug prevent version check in ckeditor.

Fixes #

#### Short description of what this resolves:
Major overhaul of UI/UX
Finish SPA refactor

#### Changes proposed in this pull request:
See issue once I do one!

PR #11215, opened 2026-03-21
fix(portal): portal messaging display fix for #11202

### LINKED ISSUE #11206: chore(session): clean leftovers from the 10244
# Description:
Since #10244 has become too big to keep in review for too long, we extracted some small leftovers to address separately.

- [x] globals.php manual reopen bypass (lines 278-284) — Manually checks instanceof `ReadAndCloseNativeSessionStorage` and calls reopenForWriting()/save() instead of using `SessionUtil::clearSession()`, which exists for exactly this purpose.
- [x] Dead `SessionUtil::sessionStartWrapper()` — No longer called anywhere. All session creation goes through `HttpSessionFactory` now. - **The fix:** It is removed together with `SessionUtil::oauthSessionStart()`.
- [x] Portal messaging username display — `getAuthPortalUsers()` now correctly uses `portal_username`, but the message summary shows that internal identifier instead of the patient's name. Filed as Chart notes display `portal_username` instead of patient name for recipient #11202 — this is pre-existing UX, not a regression. - **The fix:** `handle_note.php:117` now passes `$sn` (`sender_name`, the patient's real name like "John Smith") instead of `$sid` (`sender_id`, the `portal_username` like "patient1") as the `$assigned_to` parameter to `addPnote()`. This means the `pnote` body will show (admin to John Smith) instead of (admin to patient1).
- [x] /meta/health/index.php reopening for writes — The Docker health check triggers a session reopen-for-write on every poll (~1/min). Not harmful, but noisy in the logs. - **The fix:** added `$sessionAllowWrite = true` in `meta/health/index.php` before loading `globals.php`. Each health probe starts a fresh session (no cookie persistence), so `globals.php` always writes `site_id` to the session. Without the writable flag, this triggered a `read_and_close` reopen + warning log on every poll (~1/min). Now the session starts writable, so the write is normal and silent.
- [x] Remaining `$_SESSION` references — ~15 application files still use `$_SESSION` directly (e.g., `eye_mag_functions.php`, `calendar/index.php`, `date_functions.php`, `FormService.php`). Some may be intentionally left, but are worth confirming. - **The fix:** Actually, some were left in comments/doc comments, which are now updated

- `CsrfUtils` breaking signature change — Parameter order flipped, and `$session` became required. Correct change, but module developers calling `collectCsrfToken()` or `verifyCsrfToken()` will break. Needs a note in the upgrade docs. - This is true, but the good thing about this refactoring of parameters is that now PHPStan raises an error if you use them in the old way, so it will be visible in PR.
- Dual enforcement strategy — Both `WriteThroughSession` (runtime safety net) and `ForbidDirectSessionWriteRule` (`PHPStan` rule with ~20 whitelisted paths) exist to prevent the same problem. Worth clarifying which is primary. The large whitelist on the PHPStan rule undermines its value.
These aren't really redundant — they serve different roles (added brief explanation in code too):
  - `WriteThroughSession` is the actual guardrail. It prevents data loss at runtime, no matter what. It's always on for factory-created sessions.
  - `ForbidDirectSessionWriteRule` is developer guidance. It steers people toward `SessionUtil` methods, which are the canonical API. The whitelist isn't really undermining the rule — nearly all whitelisted paths are legitimately writable contexts where direct `$session->set()` is correct and efficient (no reopen needed). The rule's real value is catching direct writes in ~95% of code that runs with `read_and_close=true`.

zmilan:
Fixes #11206

- [x] Portal messaging username display — `getAuthPortalUsers()` now correctly uses `portal_username`, but the message summary shows that internal identifier instead of the patient's name. Filed as Chart notes display `portal_username` instead of patient name for recipient #11202 — this is pre-existing UX, not a regression. - **The fix:** `handle_note.php:117` now passes `$sn` (`sender_name`, the patient's real name like "John Smith") instead of `$sid` (`sender_id`, the `portal_username` like "patient1") as the `$assigned_to` parameter to `addPnote()`. This means the `pnote` body will show (admin to John Smith) instead of (admin to patient1).
- [x] Replace `$_SESSION` with `$session/SessionUtil` in some commented-out example of code

### DIFF
--- a/controllers/C_Document.class.php
+++ b/controllers/C_Document.class.php
@@ -355,18 +355,6 @@ class C_Document extends Controller
                         $error = $tmp;
                     }
                 }
-                // Following is just an example of code in such a plugin file.
-                /*****************************************************
-                public function documentUploadPostProcess($filename, &$d) {
-                  $userid = $session->get('authUserID');
-                  $row = sqlQuery("SELECT username FROM users WHERE id = ?", array($userid));
-                  $owner = strtolower($row['username']);
-                  $dn = '1_' . ucfirst($owner);
-                  $filepath = "/shared_network_directory/$dn/$filename";
-                  if (@unlink($filepath)) return '';
-                  return "Failed to delete '$filepath'.";
-                }
-                *****************************************************/
             }
         }
 
--- a/interface/forms/eye_mag/report.php
+++ b/interface/forms/eye_mag/report.php
@@ -147,32 +147,7 @@ function eye_mag_report($pid, $encounter, $cols, $id, $formname = 'eye_mag'): vo
    * @return string => returns the HTML of the report selected
    */
 
-    if ($choice == 'DRAW') {
-        /*
-      $side="OU";
-      $zone = array("HPI","PMH","VISION","NEURO","EXT","ANTSEG","RETINA","IMPPLAN");
-        //  for ($i = 0; $i < count($zone); ++$i) {
-        //  show only 2 for now in the encounter page
-      ($choice =='drawing') ? ($count = count($zone)) : ($count ='2');
-      for ($i = 0; $i < $count; ++$i) {
-        $file_location = $GLOBALS["OE_SITES_BASE"]."/".$session->get('site_id')."/documents/".$pid."/".$form_folder."/".$encounter."/".$side."_".$zone[$i]."_VIEW.png";
-        $sql = "SELECT * from documents where url='file://".$file_location."'";
-        $doc = sqlQuery($sql);
-        if (file_exists($file_location) && ($doc['id'] > '0')) {
-        $filetoshow = $GLOBALS['web_root']."/controller.php?document&retrieve&patient_id=$pid&document_id=$doc[id]&as_file=false";
-        ?><div style='position:relative;float:left;width:100px;height:75px;'>
-        <img src='<?php echo attr($filetoshow); ?>' width=100 heght=75>
-        </div> <?
-        } else {
-             // $filetoshow = "../../forms/".$form_folder."/images/".$side."_".$zone[$i]."_BASE.png?".rand();
-        }
-        ?>
-
-        <?php
-      }
-      } else if ($choice == "drawing") {
-        */
-        ?>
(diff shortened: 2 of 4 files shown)


portal/home.php
```diff
@@ -102,8 +102,9 @@
 }
 
 // force to message page if new messages.
```
kojiromike: This entire block is dead code (commented out since PR #7556, July 2024). Could you delete it rather than updating the session references? The `landOn` query parameter system replaced this functionality.

```suggestion
```

controllers/C_Document.class.php
```diff
@@ -358,7 +358,7 @@ public function upload_action_process()
                 // Following is just an example of code in such a plugin file.
                 /*****************************************************
```
kojiromike: This line is inside a `/* */` comment block (example plugin function). Could you delete the entire commented-out block instead of updating the `$_SESSION` reference?

interface/forms/eye_mag/report.php
```diff
@@ -155,7 +155,7 @@ function eye_mag_report($pid, $encounter, $cols, $id, $formname = 'eye_mag'): vo
         //  show only 2 for now in the encounter page
       ($choice =='drawing') ? ($count = count($zone)) : ($count ='2');
```
kojiromike: This line is inside a `/* */` comment block (lines ~151-174). Could you delete the entire commented-out block instead of updating the `$_SESSION` reference?

interface/main/calendar/index.php
```diff
@@ -51,7 +51,7 @@
 $sessionSetArray['pc_facility'] = 0;
```
kojiromike, line 52: This line is inside a `/* */` comment block. Could you delete the block instead of updating the `$_SESSION` reference?

zmilan:
@kojiromike I removed all of them. I am sorry for the confusion. I know that they are inside commented-out code, which is why I didn't change them in the main core session PR. But when you mentioned them in the previous PR, I was thinking that it was left for some reason, like an example of code or something. I fully agree about git history and not leaving commented-out code inside production code.

PR #11216, opened 2026-03-21
fix(security): parameterize SQL, remove dead code, and migrate to QueryUtils in pnotes.inc.php

haoyu-haoyu:
## Summary

Complete security hardening and modernization of `library/pnotes.inc.php` — the messaging/notes system core.

## Security fixes (CRITICAL)

| Function | Issue | Fix |
|----------|-------|-----|
| `getPnotesByDate()` | `add_escape_custom()` + `str_replace()` builds `IN(...)` clause via string concat | Parameterized `IN (?, ?, ...)` via `array_fill()` + `explode()` |
| `getSentPnotesByDate()` | Same pattern | Same fix |
| `getPnotesByUser()` | `$user_plug` SQL fragment interpolated into query | Conditional `OR pnotes.assigned_to = ?` with bind param |

## Dead code removal (HIGH — contained SQL injection)

Removed 3 functions with **zero external callers** that each contained unparameterized `$search` variable interpolated directly into SQL:
- `getPatientNotes()` 
- `getPatientNotifications()`
- `getPatientSentNotes()`

Verified no callers via `grep -rn` across the entire codebase.

## QueryUtils migration (complete file)

All legacy SQL functions migrated to `QueryUtils`:

| Before | After | Count |
|--------|-------|-------|
| `sqlQuery()` | `QueryUtils::querySingleRow()` | 6 |
| `sqlStatement()` (UPDATE/DELETE) | `QueryUtils::sqlStatementThrowException()` | 8 |
| `sqlStatement()` + `sqlFetchArray()` loop | `QueryUtils::fetchRecords()` | 3 |
| `sqlInsert()` | `QueryUtils::sqlInsert()` | 2 |
| `sqlNumRows()` | `count(QueryUtils::fetchRecords())` | 1 |
| `die()` | `throw \RuntimeException()` | 3 |

## PHPStan baseline

All 4 deprecated SQL baseline entries for this file eliminated (26 violations removed). The file now has **zero** deprecated SQL function warnings.

## Net effect

`-293 lines` / `+46 lines` — significant reduction from dead code removal and replacing verbose `sqlStatement + sqlFetchArray` loops with single `fetchRecords()` calls.

## Test plan

- [ ] `composer phpstan-baseline` — baseline regenerated
- [ ] Verify message list loads (`pnotes_full.php`)
- [ ] Verify message creation and status filtering
- [ ] Verify portal-user message visibility

### DIFF
--- a/interface/main/messages/messages.php
+++ b/interface/main/messages/messages.php
@@ -341,11 +341,11 @@ if (!empty($_REQUEST['go'])) { ?>
                             // Update alert.
                             $noteid = $_POST['noteid'];
                             $form_message_status = $_POST['form_message_status'];
-                            $reply_to = filter_input(INPUT_POST, 'reply_to', FILTER_VALIDATE_INT) ?: 0;
+                            $reply_to_pid = filter_input(INPUT_POST, 'reply_to', FILTER_VALIDATE_INT) ?: 0;
                             if ($task == "save") {
                                 updatePnoteMessageStatus($noteid, $form_message_status);
                             } else {
-                                updatePnotePatient($noteid, $reply_to);
+                                updatePnotePatient($noteid, $reply_to_pid);
                             }
                             $task = "edit";
                             $note = $_POST['note'];


library/pnotes.inc.php
```diff
@@ -119,22 +121,25 @@ function getPnotesByUser($activity = "1", $show_all = "no", $user = '', $count =
...
         $sql .= " order by " . escape_sql_column_name($sortby, ['users','patient_data','pnotes'], true) .
             " " . escape_sort_order($sortorder) .
             " limit " . escape_limit($begin) . ", " . escape_limit($listnumber);
     }
 
-    $result = sqlStatement($sql, [$usrvar]);
-
   // return the results
     if ($count) {
-        $total = sqlNumRows($result) != 0 ? sqlNumRows($result) : 0;
```
kojiromike, line 133: The original used `sqlNumRows()` which counts rows from the result set metadata without fetching them into memory. `QueryUtils::fetchRecords()` + `count()` materializes every row into a PHP array just to count it — that's a memory regression if the result set is large.

Since you're already refactoring this function, would you consider extracting the shared FROM/WHERE into a variable so the count path can use `SELECT COUNT(*)`? Something like:

```php
$fromWhere = "FROM ((pnotes LEFT JOIN users ON pnotes.user = users.username)
      LEFT JOIN patient_data ON pnotes.pid = patient_data.pid) WHERE $activity_query
      pnotes.deleted != '1' AND (pnotes.assigned_to LIKE ?";
$sqlBindArray[] = $usrvar;
if ($includePortalUser) {
    $fromWhere .= " OR pnotes.assigned_to = ?";
    $sqlBindArray[] = 'portal-user';
}
$fromWhere .= ")";

if ($count) {
    $row = QueryUtils::querySingleRow("SELECT COUNT(*) AS cnt $fromWhere", $sqlBindArray);
    return (int) $row['cnt'];
}

$sql = "SELECT pnotes.id, ... $fromWhere";
// ... ORDER BY / LIMIT ...
return QueryUtils::sqlStatementThrowException($sql, $sqlBindArray);
```

That lets the database do the counting and avoids fetching all rows into PHP.

haoyu-haoyu: Good catch — replaced with `SELECT COUNT(*) AS cnt` via `querySingleRow()` in 031e422. Extracted the shared `FROM/WHERE` into `$fromWhere` so the count path lets the database do the counting without materializing rows into PHP.

kojiromike:
@haoyu-haoyu I'm confused. It looks like you pushed a little while ago, but I still see conflicts. (On all of your most recent PRs.)

haoyu-haoyu:
Rebased all four branches onto latest master — conflicts were from the session migration (getWrapper → getActiveSession). All resolved and force-pushed. Should be clean now.

PR #11199, opened 2026-03-20
fix(security): parameterize SQL and filter request inputs in add_edit_event

haoyu-haoyu:
## Summary

Follow-up from #11077 review, per @kojiromike's request.

### 1. Parameterize SQL UPDATE (line 701)

The only non-parameterized SQL statement in `add_edit_event.php` — all other 12+ queries already use `?` placeholders.

```diff
-sqlStatement("UPDATE openemr_postcalendar_events SET " .
-"pc_catid = '" . add_escape_custom($_POST['form_category']) . "', " .
-...
-"WHERE pc_eid = '" . add_escape_custom($eid) . "'");

+sqlStatement(
+    "UPDATE openemr_postcalendar_events SET " .
+    "pc_catid = ?, pc_aid = ?, pc_pid = ?, ..." .
+    "WHERE pc_eid = ?",
+    [$_POST['form_category'], $prov, ..., $eid]
+);
```

### 2. Replace `$_REQUEST` with `$_GET` + `filter_input` (line 1091)

The tab navigation variables used `$_REQUEST` which includes POST data unnecessarily. Changed to `$_GET` for string params and `filter_input(INPUT_GET, ..., FILTER_VALIDATE_INT)` for `$eid` and `$uid`, matching the pattern from line 69.

## Test plan

- [ ] Create/edit calendar events in all modes (single, recurring, provider, group)
- [ ] Verify tab navigation (Patient/Provider/Group) preserves URL parameters
- [ ] Confirm no SQL errors in PHP error log

Fixes #11184

PR #11218, opened 2026-03-21
fix(security): parameterize SQL, add CSRF, and fix serialization injection in portal appointment

haoyu-haoyu:
## Summary

Comprehensive security hardening of `portal/add_edit_event_user.php` — the patient-facing portal appointment create/edit/delete handler.

### 1. SQL Injection — 81 `add_escape_custom()` → parameterized queries (CRITICAL)

All 5 SQL blocks (INSERT/UPDATE for multi-provider and single-provider paths) converted from string concatenation with `add_escape_custom()` to `QueryUtils::sqlStatementThrowException()` with `?` bind placeholders.

This is the portal-facing mirror of `interface/main/calendar/add_edit_event.php` (fixed in PR #11199). Same pattern, same fix.

### 2. Missing CSRF Protection (CRITICAL)

The appointment form had **zero** CSRF token verification. Any malicious website could forge POST requests to create, modify, or delete appointments for authenticated portal patients.

Added `CsrfUtils::verifyCsrfToken()` at form action entry point and hidden CSRF token field in the form HTML.

### 3. PHP Serialization Injection (HIGH)

`$_POST['form_repeat_freq']` and `$_POST['form_repeat_type']` were interpolated **raw** (no escaping) into a hand-crafted PHP serialized string (`a:5:{s:17:...}`). An attacker could corrupt the serialized structure by injecting `";}` sequences.

Replaced with `serialize()` on proper PHP arrays, matching the pattern used in `interface/main/calendar/add_edit_event.php:409-416`.

### 4. Input Type Validation (MEDIUM)

`$_GET` parameters (`userid`, `catid`, `patid`) now use `filter_input(INPUT_GET, ..., FILTER_VALIDATE_INT)` at point of assignment.

## Test plan

- [ ] Portal appointment creation (single + multi-provider)
- [ ] Portal appointment editing
- [ ] Portal appointment deletion
- [ ] CSRF token present in form HTML and verified on submit
- [ ] Recurring appointment serialization works correctly


portal/add_edit_event_user.php
```diff
@@ -54,10 +56,10 @@
 // Things that might be passed by our opener.
 //
 $eid = intval($_GET['eid'] ?? 0);         // only for existing events
-$date = $_GET['date'] ?? null;        // this and below only for new events
-$userid = $_GET['userid'] ?? null;
-$default_catid = ($_GET['catid'] ?? null) ? $_GET['catid'] : '5';
-$patientid = $_GET['patid'] ?? null;
+$date = filter_input(INPUT_GET, 'date') ?: '';
+$userid = filter_input(INPUT_GET, 'userid', FILTER_VALIDATE_INT) ?: 0;
+$default_catid = filter_input(INPUT_GET, 'catid', FILTER_VALIDATE_INT) ?: 0;
```
kojiromike: The original fallback was `'5'` (the "Office Visit" category ID). Changing it to `0` means no category will be pre-selected in the dropdown for new appointments. Was this intentional, or should the fallback remain `5`?

```suggestion
$default_catid = filter_input(INPUT_GET, 'catid', FILTER_VALIDATE_INT) ?: 5;
```

portal/add_edit_event_user.php
```diff
@@ -151,6 +153,13 @@
 
 // If we are saving, then save and close the window.
 //
+// Verify CSRF token for all form actions
+if (!empty($_POST['form_action'])) {
```
kojiromike: `empty()` has surprising behavior — `empty("0")` returns `true`. If `form_action` could legitimately be `"0"`, this would skip CSRF verification. An explicit check would be safer:

```suggestion
if (($_POST['form_action'] ?? '') !== '') {
```

portal/add_edit_event_user.php
```diff
@@ -709,6 +666,7 @@
 <body class="skin-blue">
     <div class="container-fluid">
         <form method='post' name='theaddform' id='theaddform' action='add_edit_event_user.php?eid=<?php echo attr_url($eid); ?>'>
+<input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken('portal-appointment', $session->getSymfonySession())); ?>" />
```
kojiromike: no `attr` needed here. See #11224

portal/add_edit_event_user.php
haoyu-haoyu: Restored to `?: 5` in 627f6a4. Good catch — I didn't realize the original default was the Office Visit category ID.

portal/add_edit_event_user.php
haoyu-haoyu: Fixed in 627f6a4 — changed to `($_POST['form_action'] ?? '') \!== ''`. Also applied the same fix to the second occurrence at line 501.

portal/add_edit_event_user.php
haoyu-haoyu: Removed `attr()` wrapper in 627f6a4.

PR #11219, opened 2026-03-21
fix(security): add CSRF protection and parameterize SQL in new_payment.php

haoyu-haoyu:
## Summary

Security hardening of `interface/billing/new_payment.php` — the payment entry form handling insurance and patient payments.

### 1. Missing CSRF Protection (CRITICAL)

The file had **zero** CSRF token verification. All POST-based operations (creating payments, posting distributions, finishing payments) could be triggered by cross-site request forgery.

Added `CsrfUtils::verifyCsrfToken()` at POST entry point and hidden token field in the form.

### 2. SQL Injection via string concatenation (CRITICAL)

The `ar_session` INSERT (lines 68-82) was built entirely from string concatenation mixing `add_escape_custom()` and the deprecated `formData()` function — 17 fields total, none parameterized. 

Replaced with `QueryUtils::sqlInsert()` and 13 bind parameters.

### 3. Fragile `$QueryPart` pattern (HIGH)

The `$QueryPart` variable used unterminated SQL quotes (`patient_id = '0` with no closing `'`), relying on the next line to close it. Replaced with clean `(int)` type-casting of `$payer_id` / `$patient_id` at assignment.

### 4. Deprecated `formData()` calls (MEDIUM)

9 `formData()` calls (which internally call `add_escape_custom()`) replaced with direct `$_POST` access through bind parameters.

## Test plan

- [ ] Create insurance payment → verify ar_session record
- [ ] Create patient payment → verify ar_session record
- [ ] Verify CSRF token present in form and validated on submit

### DIFF
--- a/interface/billing/new_payment.php
+++ b/interface/billing/new_payment.php
@@ -42,16 +42,18 @@ if (!AclMain::aclCheckCore('acct', 'bill', '', 'write') && !AclMain::aclCheckCor
     $screen = 'new_payment';
 //===============================================================================
 
+// Initialisations
+$mode = filter_input(INPUT_POST, 'mode') ?: '';
+$payment_id = filter_input(INPUT_GET, 'payment_id', FILTER_VALIDATE_INT)
+    ?: filter_input(INPUT_POST, 'payment_id', FILTER_VALIDATE_INT)
+    ?: 0;
+
 // Verify CSRF token for all POST actions
-if (($_POST['mode'] ?? '') !== '') {
+if ($mode !== '') {
     if (!CsrfUtils::verifyCsrfToken($_POST["csrf_token_form"] ?? '', $session)) {
         CsrfUtils::csrfNotVerified();
     }
 }
-
-// Initialisations
-$mode                    = $_POST['mode'] ?? '';
-$payment_id              = isset($_REQUEST['payment_id'])          ? $_REQUEST['payment_id'] + 0      : 0;
 $request_payment_id      = $payment_id ;
 $hidden_patient_code     = $_REQUEST['hidden_patient_code'] ?? '';
 $default_search_patient  = $_POST['default_search_patient'] ?? '';
@@ -60,7 +62,7 @@ $hidden_type_code        = $_REQUEST['hidden_type_code'] ?? '';
 //ar_session addition code
 //===============================================================================
 
-if ($mode == "new_payment" || $mode == "distribute") {
+if ($mode === "new_payment" || $mode === "distribute") {
     // Validate and extract POST inputs at the source with filter_input
     $type_name = trim(filter_input(INPUT_POST, 'type_name') ?: '');
     $hidden_type_code_int = filter_input(INPUT_POST, 'hidden_type_code', FILTER_VALIDATE_INT) ?: 0;
@@ -133,7 +135,7 @@ if ($mode == "PostPayments" || $mode == "FinishPayments") {
     }
     if ($mode == "FinishPayments") {
         // @todo This is not useful. Gonna let fall through to form init.
-        header("Location: edit_payment.php?payment_id=" . urlencode($payment_id) . "&ParentPage=new_payment");
+        header("Location: edit_payment.php?payment_id=" . $payment_id . "&ParentPage=new_payment");
         die();
     }
     $mode = "search";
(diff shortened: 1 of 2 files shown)


interface/billing/new_payment.php
```diff
@@ -36,6 +38,14 @@
 //===============================================================================
     $screen = 'new_payment';
 //===============================================================================
+
+// Verify CSRF token for all POST actions
+if (!empty($_POST['mode'])) {
```
kojiromike: `empty()` has surprising behavior — `empty("0")` returns `true`. Should this use an explicit check instead?

```suggestion
if (($_POST['mode'] ?? '') != '') {
```

interface/billing/new_payment.php
```diff
@@ -49,37 +59,48 @@
...
+    $user_id = $_SESSION['authUserID'];
+    $modified_time = date('Y-m-d H:i:s');
+    $check_date = DateToYYYYMMDD(trim($_POST['check_date'] ?? ''));
+    $deposit_date = DateToYYYYMMDD(trim($_POST['deposit_date'] ?? ''));
+    $post_to_date = DateToYYYYMMDD(trim($_POST['post_to_date'] ?? ''));
     if ($post_to_date == '') {
         $post_to_date = date('Y-m-d');
     }
-    if ($_POST['deposit_date'] == '') {
+    if (($_POST['deposit_date'] ?? '') == '') {
```
kojiromike: Loose `==` here means values like `0`, `false`, or `"0"` would also match. Should this be a strict comparison?

```suggestion
    if (($_POST['deposit_date'] ?? '') === '') {
```

interface/billing/new_payment.php
```diff
@@ -49,37 +59,48 @@
 
 if ($mode == "new_payment" || $mode == "distribute") {
     if (trim((string) $_POST['type_name']) == 'insurance') {
```
kojiromike: All these `$_POST` accesses return `mixed`, which will cause PHPStan to complain at every call site (`trim()`, `DateToYYYYMMDD()`, bind parameters, etc.). Consider using `filter_input(INPUT_POST, 'key')` for string values and `filter_input(INPUT_POST, 'key', FILTER_VALIDATE_INT)` for integer values — that gives PHPStan a known return type and avoids a pile of baseline entries.

For example, at the top of this block:
```php
$type_name = trim(filter_input(INPUT_POST, 'type_name') ?? '');
$hidden_type_code = filter_input(INPUT_POST, 'hidden_type_code', FILTER_VALIDATE_INT) ?: 0;
$check_number = trim(filter_input(INPUT_POST, 'check_number') ?? '');
// ... etc.
```

Then use the local variables throughout instead of raw `$_POST` access.

interface/billing/new_payment.php
```diff
@@ -49,37 +59,48 @@
...
+            description = ?, adjustment_code = ?, post_to_date = ?,
+            payment_method = ?",
+        [
+            $payer_id,
+            $patient_id,
+            $user_id,
+            trim($_POST['check_number'] ?? ''),
+            $check_date,
+            $deposit_date,
+            trim($_POST['payment_amount'] ?? ''),
```
kojiromike: `payment_amount` is bound to a numeric column (`pay_total`). Should this be validated as a number? Non-numeric strings would silently become `0` in MySQL.

```php
$pay_total = filter_input(INPUT_POST, 'payment_amount', FILTER_VALIDATE_FLOAT) ?: 0.0;
```

interface/billing/new_payment.php
```diff
@@ -339,6 +357,7 @@ function FillUnappliedAmount(){
                 } else {
                     echo 'return false;';
                 }?>" style="display:inline">
+                    <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken()); ?>" />
```
kojiromike: `attr()` isn't needed here. See #11224

interface/billing/new_payment.php
haoyu-haoyu: Fixed in df0f257 — changed to `($_POST['mode'] ?? '') !== ''`.

interface/billing/new_payment.php
haoyu-haoyu: Fixed in df0f257 — strict `===` comparison.

interface/billing/new_payment.php
haoyu-haoyu: Adopted in df0f257. Extracted all POST values via `filter_input()` at the top of the block:

```php
$type_name = trim(filter_input(INPUT_POST, 'type_name') ?? '');
$hidden_type_code_int = filter_input(INPUT_POST, 'hidden_type_code', FILTER_VALIDATE_INT) ?: 0;
$check_number = trim(filter_input(INPUT_POST, 'check_number') ?? '');
$payment_amount = filter_input(INPUT_POST, 'payment_amount', FILTER_VALIDATE_FLOAT) ?: 0.0;
// ... etc.
```

Then used the local variables throughout the INSERT bind array.

interface/billing/new_payment.php
haoyu-haoyu: Added `FILTER_VALIDATE_FLOAT ?: 0.0` in df0f257.

interface/billing/new_payment.php
haoyu-haoyu: Removed `attr()` wrapper in df0f257.

Issue #8965, opened 2025-09-30
Proposal: Configuration Management APIs for OpenEMR


kojiromike:
# Proposal: Configuration Management APIs for OpenEMR

## Overview

This proposal outlines the introduction of comprehensive REST APIs for secure configuration management in OpenEMR, enabling programmatic administration of users, groups, access control, and global settings.

## Motivation

Currently, OpenEMR configuration management requires manual interaction through the web interface. Modern deployment practices, including Infrastructure as Code (IaC), containerization, and automated provisioning, would benefit from programmatic access to configuration management capabilities.

### Benefits

- **Automation**: Enable automated setup and configuration of OpenEMR instances
- **Consistency**: Ensure consistent configuration across multiple environments
- **DevOps Integration**: Support CI/CD pipelines and infrastructure automation tools
- **Scalability**: Facilitate large-scale deployments and multi-tenant scenarios
- **Auditability**: Programmatic changes can be version-controlled and audited

#### Create User
```
POST /api/admin/users
```
Create a new user with specified roles and permissions.

#### Update User
```
PUT /api/admin/users/{userId}
```
Modify user properties, roles, and status.

#### Delete/Deactivate User
```
DELETE /api/admin/users/{userId}
```
Deactivate or remove a user from the system.

#### List Users
```
GET /api/admin/users
```
Retrieve users with filtering and pagination.

#### Get User Details
```
GET /api/admin/users/{userId}
```
Retrieve detailed information about a specific user.

#### Create Group
```
POST /api/admin/groups
```
Create a new user group with specified permissions.

#### Update Group
```
PUT /api/admin/groups/{groupId}
```
Modify group properties and permissions.

#### Delete Group
```
DELETE /api/admin/groups/{groupId}
```
Remove a group from the system.

#### List Groups
```
GET /api/admin/groups
```
Retrieve all groups with filtering options.

#### Manage Group Members
```
POST /api/admin/groups/{groupId}/members
DELETE /api/admin/groups/{groupId}/members/{userId}
```
Add or remove users from groups.

#### List ACL Objects
```
GET /api/admin/acl/objects
```
Retrieve all available ACL objects and sections.

#### Get ACL Permissions
```
GET /api/admin/acl/permissions
```
Retrieve current ACL/GACL configuration.

#### Update ACL Permissions
```
PUT /api/admin/acl/{aclId}/permissions
```
Modify ACL permissions for users or groups.

#### Grant ACL Access
```
POST /api/admin/acl/grant
[...]

PR #11222, opened 2026-03-21
feat(api): add admin user management endpoints (GET/POST /api/admin/users)

zmilan:
Implements phase 1 for #8965

Adds **Phase 1** of the Configuration Management API (issue #8965): three admin user management endpoints under `/api/admin/users` for programmatic user provisioning. This establishes the foundational pattern for subsequent admin API endpoints.

- Add `GET` `/api/admin/users` — list users with admin-level detail (username, authorized, ACL groups)
- Add `GET` `/api/admin/users/{uuid}` — get single user with enriched fields
- Add `POST` `/api/admin/users` — create user (password, UUID, groups, ACL, audit log, event dispatch)
- Create `AdminUserService` extends `UserService` with `getSelectColumns()` override for field whitelisting
- Refactor `UserService` to extract g`etSelectColumns()` protected method (deduplicates hardcoded column lists in `search()` and `getAll()`)
- Create `UserValidator` for input validation (username format, required fields, access_group)
- Create `UserManagementRestController` following `FacilityRestController` pattern
- Register `user/users.read` and `user/users.write` OAuth scopes
- Add `user/user.read` and `user/users.*` scopes to `ApiTestClient::ALL_SCOPES`
- Add standard-admin OpenAPI tag and regenerate swagger spec
- Add 16 PHPUnit API tests (happy path, validation errors, 401/403 auth, regression)

#### Assisted-by: Claude Code


src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,288 @@
+<?php
+
+/**
+ * Admin User Management Service — read and create operations for admin user endpoints.
+ *
+ * @package   OpenEMR
+ * @author    Milan Zivkovic <zivkovic.milan@gmail.com>
+ * @link      https://www.open-emr.org
```
kojiromike, line 6: All new and changed files in this PR need the OpenCoreEMR copyright header. Same applies to `UserManagementRestController.php`, `UserValidator.php`, and `UserManagementApiTest.php`.

```suggestion
 * @link      https://www.open-emr.org
 * @author    Milan Zivkovic <zivkovic.milan@gmail.com>
 * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
```

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,288 @@
...
+            "', billing_facility_id = '" . self::escape($billingFacilityId) .
+            "', specialty = '" . self::escape($specialty) .
+            "', calendar = '" . self::escape((string) $calendar) .
+            "', portal_user = '" . self::escape((string) $portalUser) .
+            "'";
+
+        // AuthUtils::updatePassword() verifies the admin password and executes the INSERT
+        $session = SessionWrapperFactory::getInstance()->getActiveSession();
+        $adminPass = self::strVal($data['admin_password'] ?? '');
+        $authUtils = new AuthUtils();
```
kojiromike, line 234: `admin_password` is functionally required — `AuthUtils::updatePassword()` rejects an empty `$currentPwd` at line 520 with a generic "Password update error" message. But it's not listed as required in the OpenAPI spec, and `UserValidator` doesn't check for it.

A caller following the API docs won't know to include it. Should this be documented as required in the schema and validated upfront with a clear error message?

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,288 @@
...
+        if (is_array($existing) && isset($existing['username']) && $existing['username'] !== '') {
+            $processingResult->setValidationMessages(['username' => ['Username already exists']]);
+            return $processingResult;
+        }
+
+        // Build INSERT SQL — AuthUtils::updatePassword() requires raw SQL string
+        // (it executes via privStatement($insert_sql, []) with no bind params)
+        $insertUserSQL =
+            "INSERT INTO users SET " .
+            "username = '" . self::escape($username) .
```
kojiromike: Building the INSERT via string concatenation is necessary because `AuthUtils::updatePassword()` takes a raw SQL string — tracked in #11227 for future refactoring.

For now this works, but worth noting that `add_escape_custom()` wraps `mysqli_real_escape_string()` which is charset-dependent. It's fine with OpenEMR's UTF-8, but another reason to move to parameterized queries when the AuthUtils API changes.

src/Services/UserService.php
```diff
@@ -66,6 +66,26 @@ public function getUuidFields()
...
+     */
+    protected function getSelectColumns(): string
+    {
+        $columns = "id, uuid, users.title as title, fname, lname, mname,
+                    federaltaxid, federaldrugid, upin, facility_id, facility,
+                    npi, email, active, specialty, billname, url, assistant,
+                    organization, valedictory, street, streetb, city, state,
+                    zip, phone, fax, phonew1, phonecell, users.notes,
+                    state_license_number, abook.title as abook_title, last_updated";
+        if ($this->_includeUsername) {
```
kojiromike, line 87: The new shared `getSelectColumns()` includes `last_updated`, but the old `getAll()` method did not. This changes the behavior of `getAll()` — it will now return `last_updated` where it didn't before.

Probably fine (additive change), but is this intentional?

apis/routes/_rest_routes_standard.inc.php
```diff
@@ -563,113 +564,131 @@
...
+    },
+    "GET /api/admin/users/:uuid" => static function ($uuid, HttpRestRequest $request) {
+        RestConfig::request_authorization_check($request, "admin", "users");
+        return (new UserManagementRestController())->getOne(is_string($uuid) ? $uuid : '', $request);
+    },
+    "POST /api/admin/users" => static function (HttpRestRequest $request) {
+        RestConfig::request_authorization_check($request, "admin", "super");
+        /** @var array<string, mixed> $data */
+        $data = (array) (json_decode((string) file_get_contents("php://input")));
+        return (new UserManagementRestController())->post($data, $request);
```
kojiromike, line 747: Some routes in this file use `$request->query->all()` for query parameters (patient, practitioner). Would it be more consistent to use that here?

```suggestion
        $searchParams = $request->query->all();
```

(See also #11226 for cleaning up the inconsistency across all routes.)

src/RestControllers/Admin/UserManagementRestController.php
```diff
@@ -0,0 +1,136 @@
...
+namespace OpenEMR\RestControllers\Admin;
+
+use OpenApi\Attributes as OA;
+use OpenEMR\Common\Http\HttpRestRequest;
+use OpenEMR\RestControllers\RestControllerHelper;
+use OpenEMR\Services\Admin\UserManagementService;
+use Psr\Http\Message\ResponseInterface;
+
+class UserManagementRestController
+{
```
kojiromike, line 23: New PHP files should include `declare(strict_types=1)`. Same applies to `UserManagementService.php`, `UserValidator.php`, and `UserManagementApiTest.php`.

kojiromike (commented):
Nice foundational work for the admin API — the controller/service/validator structure follows existing patterns well, and the test coverage is solid with good happy-path and sad-path cases.

The mass `function` → `static function` change across all existing routes makes the diff hard to review. Would you consider splitting that into a separate PR? It's a valid improvement but orthogonal to the admin user endpoints.

src/Services/UserService.php
zmilan: No, it seems that I didn't notice it... should I remove it?

src/Services/Admin/UserManagementService.php
zmilan: Added as required and properly handled.

src/Services/UserService.php
kojiromike: I doubt it matters, but it's good to know for testing

zmilan: Ok, I fixed it, so it now preserve original behavior in non-admin routes

kojiromike:
@claude review this pr

(IDK if this will work.)

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,289 @@
+<?php
+
+/**
+ * Admin User Management Service — read and create operations for admin user endpoints.
+ *
+ * @link      https://www.open-emr.org
+ * @author    Milan Zivkovic <zivkovic.milan@gmail.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ */
```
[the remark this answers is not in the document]
kojiromike: @copilot that instruction is outdated. We want a commit trailer, not comments in the code.

kojiromike (commented):
Follow-up review — skipping points already covered in the first round and by Copilot (duplicate last_updated, error mapping, transaction wrapping). One additional finding on field consistency between the list and detail endpoints.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,289 @@
...
+    /**
+     * Get a single user by UUID with ACL group enrichment.
+     *
+     * @param string $uuid UUID string
+     * @return ProcessingResult
+     */
+    public function getOneByUuid(string $uuid): ProcessingResult
+    {
+        $processingResult = new ProcessingResult();
+        $user = $this->getUserByUUID($uuid);
```
kojiromike: `getUserByUUID()` in the parent `UserService` runs `SELECT * FROM users WHERE uuid = ?` — it does not go through `search()` or `getSelectColumns()`. So the list endpoint returns a curated column set (via `searchUsers()` → `search()` → `getSelectColumns()`), while the detail endpoint returns every column in the `users` table.

This means:
1. The two GET endpoints return different field shapes for the same resource
2. `SELECT *` could expose columns that the list endpoint deliberately omits

Would it make sense to route this through `searchUsers()` with a UUID filter instead, so both endpoints use the same column list and enrichment pipeline?

zmilan: @kojiromike thanks for the comment. It is fixed now.

kojiromike (commented):
Authorization model seems correct, SQL is parameterized, input validation is explicit, and tests cover both happy and sad paths including authz scenarios.

One item to address:

`access_group` values are not validated against real ACL groups. The validator checks that `access_group` is a non-empty array of non-empty strings, but doesn't verify the group names actually exist. `AclExtended::setUserAro()` silently ignores non-matching titles, so `{"access_group": ["NonExistentGroup"]}` would create a user with zero ACL memberships while the API reports 201 Created. A pre-creation check should verify that at least one provided group title matches a real ACL group.

zmilan:
> Authorization model seems correct, SQL is parameterized, input validation is explicit, and tests cover both happy and sad paths including authz scenarios.
> 
> One item to address:
> 
> `access_group` values are not validated against real ACL groups. The validator checks that `access_group` is a non-empty array of non-empty strings, but doesn't verify the group names actually exist. `AclExtended::setUserAro()` silently ignores non-matching titles, so `{"access_group": ["NonExistentGroup"]}` would create a user with zero ACL memberships while the API reports 201 Created. A pre-creation check should verify that at least one provided group title matches a real ACL group.

@kojiromike I saw the first comment about `access_group` values not validated in PR for phase 2 and fixed it there. Is it ok if it stays as that, or should I cherry-pick it to this branch (it is not a single change on that commit)?

tests/Tests/Api/Admin/UserManagementApiTest.php
```diff
@@ -0,0 +1,585 @@
...
+        $this->assertEquals(Response::HTTP_NOT_FOUND, $response->getStatusCode());
+    }
+
+    #[Test]
+    public function testGetOneMalformedUuidReturnsError(): void
+    {
+        $response = $this->testClient->getOne(self::API_ENDPOINT, "not-a-uuid");
+
+        // Invalid UUID may return 400, 404, or 500 depending on how deep the parsing goes
+        $this->assertNotEquals(Response::HTTP_OK, $response->getStatusCode());
```
kojiromike: Now that `getOneByUuid()` calls `UuidRegistry::isValidStringUUID()` and returns a `uuid` validation message for a malformed value, a malformed UUID has a deterministic outcome: the controller maps validation messages to HTTP 400. This assertion (and the `assertNotEquals(HTTP_CREATED)` on the next line) would still pass if that guard regressed to letting `uuidToBytes()` throw a 500 — so it doesn't actually pin the new behavior. Could this assert the specific contract, e.g. `assertEquals(Response::HTTP_BAD_REQUEST, ...)` and check `validationErrors` contains a `uuid` key? The stale comment about "400, 404, or 500 depending on how deep the parsing goes" can go too.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,313 @@
...
+    public function searchUsers(array $search = [], bool $isAndCondition = true): ProcessingResult
+    {
+        /** @var ProcessingResult $processingResult */
+        $processingResult = $this->search($search, $isAndCondition);
+        /** @var list<array<string, mixed>> $enrichedData */
+        $enrichedData = [];
+        /** @var list<array<string, mixed>> $currentData */
+        $currentData = $processingResult->getData();
+        foreach ($currentData as $record) {
+            $this->enrichWithAclGroups($record);
```
kojiromike: `enrichWithAclGroups()` runs one `AclExtended::aclGetGroupTitles()` query per record, and the GET list endpoint applies no limit/pagination — so an install with many users does N+1 ACL lookups and returns an unbounded payload on every list call. Worth a question on two fronts: (1) should the list endpoint take a paging/limit parameter? and (2) could the ACL titles be fetched in one batched query keyed by username rather than per-row? Not blocking for a phase-1 foundation, but the cost scales with user count.

kojiromike (commented):
note: this is an old pending review I apparently never submitted. It doesn't look like any code was pushed after I generated it, but still, it might be good to take it with a grain of salt…

-----

Re-reviewed after the latest changes. The big items from earlier rounds look genuinely resolved: the detail endpoint now routes through `searchUsers()` so both GET shapes share the curated column whitelist (no `SELECT *`, no credential columns leaked), JSON parse errors return 400, facility IDs are validated pre-create, UUID format is validated, and the create flow is wrapped in a transaction with rollback. Create input is explicitly field-by-field copied into a fixed array rather than mass-assigned, and POST is gated on the `super` ACL — so privileged-user creation is intended and access-controlled.

A few smaller follow-ups below, none of which block the structural approach. I deliberately skipped points already raised by others (the `access_group`-not-validated-against-real-groups item is still open in the thread and being handled in phase 2 per your note; the duplicate-`last_updated`, error-mapping, and transaction items from the earlier Copilot/round-1 reviews are addressed).

tests/Tests/Api/Admin/UserManagementApiTest.php
zmilan: Fixed in f66c0b3.

The assertions are now `assertEquals(Response::HTTP_BAD_REQUEST, ...)` plus `assertArrayHasKey('uuid', $validationErrors)`, and the stale "400, 404, or 500" comment is gone. As you noted, `getOneByUuid()` validates the string and returns a `uuid` validation message which the controller maps to 400, so the outcome is deterministic and worth pinning rather than merely excluding 200/201.

src/Services/Admin/UserManagementService.php
zmilan: Both parts addressed in aa3eb61.

**Batched ACL lookup.** `enrichWithAclGroups()` is gone. The enrichment now issues a single join over `gacl_aro` / `gacl_groups_aro_map` / `gacl_aro_groups` keyed by username, following the pattern already used in `BreakglassChecker` and `DatabaseLocalUserDirectory`. I kept the semantics of the helper it replaces rather than tidying them up: direct (non-recursive) memberships in the `users` section, `gacl_aro_groups.name` as the title (that is index `[3]` of `get_group_data()`), sorted per user, and matching stays case-insensitive as the gacl lookup was — so the batched form returns the same titles, just in one query.

**Paging.** `_limit` / `_offset` are now supported through the existing `QueryPagination` stack, which already feeds the `links` block the controller emits (it was passing `isMultipleResultResponse: true` all along, there was simply never a limit set). `UserService::search()` takes the pagination as an optional third argument, so every existing caller gets byte-identical SQL.

Two decisions worth flagging:

- `_count` and `_maxresults` are accepted as aliases for `_limit`. `QueryPagination::getLinks()` advertises `_count` in the `first`/`next` links, so without the alias a client following a `next` link would have silently got an unpaged response.
- I did **not** impose a default cap. `QueryPagination::DEFAULT_LIMIT` is 0 and carries an explicit TODO asking whether to break backwards compatibility by imposing one; that reads like a repo-wide contract decision rather than something to settle unilaterally in one endpoint. Happy to add a default here if you would rather this endpoint lead on it.

Covered by two new tests (`testGetAllRespectsLimit`, `testGetAllPagesWithLimitAndOffset`); the second also asserts the `next` link appears. I checked they actually pin the behaviour — disabling the pagination wiring makes both fail (9 records instead of 1) rather than passing vacuously.

zmilan:
Rebased onto current master and pushed (force-update, since the branch was ~1300 commits behind and history was rewritten). The PR now shows as mergeable again.

Two things worth calling out beyond the two review threads:

- The mass `function` → `static function` change is **gone**. It did not survive the rebase, since master never adopted it — which happens to be what you asked for in the first round. `apis/routes/_rest_routes_standard.inc.php` is now a 36-line diff instead of a whole-file one, and only the three new admin routes are `static`.
- Regenerating `swagger/openemr-api.yaml` picked up POST body fields that had been added to the controller attributes but never written into the spec (`suffix`, `billing_facility_id`, `npi`, `taxonomy`, `specialty`, `calendar`, `portal_user`, `federaltaxid`, `state_license_number`). That closes the earlier Copilot note about the documented schema being narrower than the accepted payload.

One rebase consequence needed a fix of its own (c7501e3): `UserValidator` carried a `@phpstan-ignore method.nonObject` justified by `BaseValidator::$validator` being untyped. Master types it now, so with `reportUnmatchedIgnoredErrors` the stale ignore became an analysis failure.

Verification: admin API suite 23/23, services suite 740 tests / 0 failures, phpcs clean, PHPStan level 10 clean across the whole codebase.

Still open and not addressed here: validating `access_group` values against real ACL groups. That is still slated for phase 2 as discussed, so `{"access_group": ["NonExistentGroup"]}` continues to return 201 with no memberships on this branch.

src/Services/UserService.php
```diff
@@ -285,9 +271,23 @@ public function search(array $search, $isAndCondition = true)
...
+        // statement is unchanged, so existing callers keep their current behavior.
+        if ($pagination !== null && $pagination->getLimit() > 0) {
+            $processingResult->setPagination($pagination);
+            $offset = $pagination->getCurrentOffsetId();
+            // Both values are integers, so they interpolate safely; LIMIT/OFFSET cannot be bound
+            // as parameters here. One row past the limit lets ProcessingResult::addData() flag
+            // that more data exists.
+            $limitClause = $pagination->getLimit() + 1;
+            $offsetClause = is_numeric($offset) ? (int)$offset : 0;
+            $sql .= " LIMIT " . $limitClause . " OFFSET " . $offsetClause;
```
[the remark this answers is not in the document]
zmilan: Good catch — fixed in f47f843.

This was a real defect in the paging I had just added. `ORDER BY users.id` now precedes `LIMIT`/`OFFSET`. Worth noting the existing offset test passed only because InnoDB happened to return rows in primary-key order, so it was not actually pinning anything here.

The sort is applied inside the pagination branch only, so callers that pass no pagination still get byte-identical SQL to before.

src/Validators/Admin/UserValidator.php
```diff
@@ -0,0 +1,65 @@
...
+                $context->optional('npi')->lengthBetween(0, 15);
+                $context->optional('taxonomy')->lengthBetween(0, 255);
+                $context->optional('specialty')->lengthBetween(0, 255);
+                $context->optional('calendar')->inArray([0, 1, '0', '1']);
+                $context->optional('portal_user')->inArray([0, 1, '0', '1']);
+                $context->optional('federaltaxid')->lengthBetween(0, 255);
+                $context->optional('state_license_number')->lengthBetween(0, 255);
+                $context->optional('federaldrugid')->lengthBetween(0, 255);
+                $context->optional('upin')->lengthBetween(0, 255);
+                $context->required('access_group')->isArray()->callback(function ($value): bool {
```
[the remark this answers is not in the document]
zmilan: Fixed in f47f843 — this matches the finding @kojiromike raised earlier in the thread, so it is now addressed here rather than deferred to phase 2.

Two deliberate differences from the suggestion:

- **Placed in the service, not the validator callback.** `UserValidator` is a Particle schema validator; the group check needs a gacl lookup, and putting a DB query inside a schema callback mixes concerns. The service sets `setValidationMessages(['access_group' => ...])`, which the controller maps to the same 400 response with the same `access_group` key, so the API contract is what you described either way.
- **Rejects if *any* title is unknown**, rather than accepting the payload when at least one resolves. Accepting `["Physicians", "Nonsense"]` and silently dropping `Nonsense` is the same class of problem the finding is about. This is stricter than the original wording, so flagging it explicitly in case that is not wanted.

The check runs before `startTransaction()`, so nothing is written and no rollback is needed. Covered by `testPostUnknownAccessGroupReturns400`, which also asserts no user is left behind; I verified it fails (201 instead of 400) with the guard removed.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,364 @@
...
+use OpenEMR\Common\Session\SessionWrapperFactory;
+use OpenEMR\Common\Uuid\UuidRegistry;
+use OpenEMR\Core\OEGlobalsBag;
+use OpenEMR\Events\User\UserCreatedEvent;
+use OpenEMR\Services\UserService;
+use OpenEMR\Validators\Admin\UserValidator;
+use OpenEMR\Validators\BaseValidator;
+use OpenEMR\Validators\ProcessingResult;
+
+class UserManagementService extends UserService
```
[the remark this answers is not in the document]
zmilan: Pushing back on this one, though the guideline citation is accurate.

`UserService` does not itself extend `BaseService` — it is a standalone class with its own `search()`, `getSelectColumns()` and the `abook` join. `UserManagementService` extends it deliberately, to reuse exactly those: that inheritance is what makes the list and detail endpoints share one curated column set, which was the fix for an earlier review finding about `getUserByUUID()` doing `SELECT *` and returning a different field shape than the list endpoint.

Re-parenting to `BaseService` would mean either duplicating the users query in this service or introducing a composition seam through `UserService` — a refactor of a shared legacy service, in a PR the human reviewer has already asked once to keep narrowly scoped. It is also not a correctness issue, as your own "Heavy lift / refactor suggestion" labelling suggests.

Happy to be overruled by a maintainer, but I would rather see `UserService` itself moved onto `BaseService` as its own change, which would bring this service along with it, than special-case the child here.

src/Services/Admin/UserManagementService.php
zmilan: Follow-up: CodeRabbit caught a real defect in the paging I described above — no deterministic sort before `LIMIT`/`OFFSET`, so consecutive pages could repeat or skip users. Fixed in f47f843 with `ORDER BY users.id`. The offset test I mentioned had been passing on InnoDB's incidental primary-key ordering rather than on any guarantee.

zmilan:
Update on f47f843 — this supersedes the last line of my previous comment.

**`access_group` is now validated on this branch**, so it is no longer deferred to phase 2. `POST /api/admin/users` with an unknown group title returns 400 with an `access_group` validation error instead of 201 with no memberships. The check runs before the transaction opens, so nothing is written. It rejects if *any* supplied title is unknown rather than accepting the payload when one happens to resolve — stricter than the original wording, so say the word if you want it relaxed.

Also fixed in the same commit: the `_limit`/`_offset` paging I added earlier had **no deterministic sort**, so a client walking the pagination links could have seen a user twice or missed one. Now ordered by `users.id`. Credit to CodeRabbit for spotting it; my own offset test had been passing on incidental InnoDB ordering rather than on any guarantee, which is a good reminder that a green test is not the same as a pinned one.

I have pushed back on the third CodeRabbit item (re-parenting `UserManagementService` onto `BaseService`) in its thread — reasoning there, and happy to be overruled.

Previous run was 70 checks green with only `codecov/patch` short of its threshold; the two new tests here should help that.

tests/Tests/Api/Admin/UserManagementApiTest.php
```diff
@@ -0,0 +1,674 @@
...
+        // QueryPagination::getLinks() advertises _count in the first/next links, so a client
+        // following those links must get the same paging it asked for via _limit.
+        $response = $this->testClient->get(self::API_ENDPOINT, ["_count" => "1"]);
+        $this->assertEquals(Response::HTTP_OK, $response->getStatusCode());
+
+        $body = $this->decodeResponse($response);
+        /** @var list<array<string, mixed>> $data */
+        $data = $body["data"] ?? [];
+        $this->assertCount(1, $data);
+    }
```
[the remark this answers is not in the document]
zmilan: Fixed in da7f3cd. Valid point — `assertCount(1, ...)` alone would also have passed against an endpoint that ignored `_count` completely.

Added the same skip guard the offset test already uses (skip unless the unpaged list returns at least two users), so the assertion only runs when a single result actually demonstrates that the alias was applied.

tests/Tests/Api/Admin/UserManagementApiTest.php
```diff
@@ -0,0 +1,674 @@
...
+        $response = $this->testClient->getOne(self::API_ENDPOINT, $fakeUuid);
+        $this->assertEquals(Response::HTTP_NOT_FOUND, $response->getStatusCode());
+    }
+
+    #[Test]
+    public function testPostUnknownAccessGroupReturns400(): void
+    {
+        $adminPass = getenv("OE_PASS", true) ?: "pass";
+        $username = "phpunit_badacl_" . bin2hex(random_bytes(4));
+        $response = $this->testClient->post(self::API_ENDPOINT, [
```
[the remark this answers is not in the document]
zmilan: Fixed in da7f3cd — and this one is not hypothetical.

While confirming the new guard actually fails the test when removed, I temporarily disabled it. The create then succeeded, and because the username had never been registered, `tearDownAfterClass()` had no idea it existed and left it in the database. Exactly the scenario you describe.

`self::$createdUsernames[] = $username;` now runs before the request. Registering a username that is never created is harmless: teardown does a `DELETE` and an `UPDATE` that simply match no rows.

tests/Tests/Api/Admin/UserManagementApiTest.php
```diff
@@ -0,0 +1,674 @@
...
+        /** @var array<string, mixed> $validationErrors */
+        $validationErrors = $body["validationErrors"] ?? [];
+        $this->assertArrayHasKey('access_group', $validationErrors);
+
+        // The check runs before any write, so the user must not exist afterwards.
+        $listResponse = $this->testClient->get(self::API_ENDPOINT, ["username" => $username]);
+        $listBody = $this->decodeResponse($listResponse);
+        /** @var list<array<string, mixed>> $listData */
+        $listData = $listBody["data"] ?? [];
+        $this->assertCount(0, $listData);
```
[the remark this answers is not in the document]
zmilan: Fixed in da7f3cd. The follow-up lookup now asserts HTTP 200 with empty `validationErrors` and `internalErrors` before the `assertCount(0, ...)`, so a failed or malformed response can no longer masquerade as proof that the user is absent.

src/Common/Auth/OpenIDConnect/Entities/ServerScopeListEntity.php
```diff
@@ -251,6 +251,8 @@ public function apiScopes(): array
                 "user/transaction.read",
                 "user/transaction.write",
                 "user/user.read",
+                "user/users.read",
```
kojiromike, line 254: `users` is the only standard-API resource in this list with no counterpart in `getV2ApiScopes()` — `facility`, `patient`, `user` and the rest all appear in both. A v2-syntax client therefore can't request these endpoints at all.

Based on how `checkSecurity()` derives permissions (list → `s`, `/{uuid}` → `r`, POST → `c`), adding `,'users' => ['c','r','s']` to `$userResources` would close the gap.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,400 @@
...
+            }
+
+            // Always assign UUID to the newly created user
+            $uuid = UuidRegistry::getRegistryForTable('users')->createUuid();
+            QueryUtils::sqlStatementThrowException("UPDATE users SET uuid = ? WHERE BINARY username = ?", [$uuid, $username]);
+
+            // Update facility name fields (IDs were validated above)
+            if ($facilityId !== '') {
+                QueryUtils::sqlStatementThrowException(
+                    "UPDATE users, facility SET users.facility = facility.name WHERE facility.id = ? AND users.username = ?",
```
kojiromike: This matches on `users.username` without `BINARY`, while the uniqueness check (L148) and the UUID update three lines up both use `BINARY`. Usernames are case-sensitive throughout (`AuthUtils`, `UserService::getIdByUsername`) and `users.username` carries no unique index — so `Bob` and `bob` can coexist, and creating `Bob` with a `facility_id` rewrites `bob`'s `facility` too. Same on the `billing_facility` statement below.

```suggestion
                    "UPDATE users, facility SET users.facility = facility.name WHERE facility.id = ? AND BINARY users.username = ?",
```

src/Services/UserService.php
```diff
@@ -285,9 +281,25 @@ public function search(array $search, $isAndCondition = true)
...
         $processingResult = new ProcessingResult();
+        // Callers that pass a pagination object get a bounded query; without one the
+        // statement is unchanged, so existing callers keep their current behavior.
+        if ($pagination !== null && $pagination->getLimit() > 0) {
+            $processingResult->setPagination($pagination);
+            $offset = $pagination->getCurrentOffsetId();
+            // Both values are integers, so they interpolate safely; LIMIT/OFFSET cannot be bound
+            // as parameters here. One row past the limit lets ProcessingResult::addData() flag
+            // that more data exists.
+            $limitClause = $pagination->getLimit() + 1;
```
kojiromike, line 292: `LIMIT`/`OFFSET` do bind in this codebase — `PersonService::search()` does `" ORDER BY last_name, first_name LIMIT ? OFFSET ?"` with the values pushed onto the params array (src/Services/PersonService.php:285); `ContactService` and `PersonPatientLinkService` do the same. ADODB's emulated binds inline int values raw, so the int cast is what keeps them unquoted.

Could these be `?` placeholders appended to `$sqlBindArray`, with the "cannot be bound" sentence dropped?

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,400 @@
...
+                "New user created via API: " . $username
+            );
+
+            // Dispatch event
+            $eventData = $data;
+            $eventData['uuid'] = UuidRegistry::uuidToString($uuid);
+            $eventData['username'] = $username;
+            unset($eventData['password'], $eventData['admin_password']);
+            $userCreatedEvent = new UserCreatedEvent($eventData);
+            OEGlobalsBag::getInstance()->getKernel()->getEventDispatcher()->dispatch(
```
kojiromike, line 343: `UserCreatedEvent` fires before `commitTransaction()`. A listener that does outside work — welcome email, downstream provisioning — runs even if the commit then fails, and a listener that throws rolls back an otherwise-successful creation with no way for the caller to tell the two apart.

Would moving the dispatch below the commit be safer? The audit entry arguably belongs inside, since it should vanish with a rollback.

kojiromike (approved):
The rebase did shrink the route diff, and the two open threads read as genuinely closed. Four items below.

The scope registration is the one I would fix before merge: `users` is the only standard-API resource that lands in `apiScopes()` without a matching entry in `getV2ApiScopes()`. A client using v2 scope syntax cannot request these endpoints. The tests authenticate with the v1 strings, so CI stays green either way.

The facility `UPDATE` is the other substantive one — it matches on a non-`BINARY` username while everything around it is `BINARY`.

src/Common/Auth/OpenIDConnect/Entities/ServerScopeListEntity.php
zmilan: Added `,'users' => ['c','r','s']` in 38b0f9a54b. Confirmed the derivation before taking the values: `parseRouteParams()` resolves the resource to `users` for all three routes (the `:uuid` form pops the identifier and lands on `users` too), and `checkSecurity()` derives `s` for the list, `r` for the instance and `c` for the POST — so no `u` or `d`, since no route resolves to them.

Also added `user/users.crs` to the OpenAPI security scheme in `OpenApiDefinitions.php` and regenerated the spec. That list is hand-maintained and carries the v2 scopes for `api:oemr`, so without it the scope works but is not discoverable from the docs.

On the "CI stays green either way" point — that is the part that bothered me more than the missing entry, so there is now a `ServerScopeListEntityTest` pinning the invariant that every v1 API resource has a v2 counterpart. It fails with `missing: ['user/users']` without the fix.

src/Services/Admin/UserManagementService.php
zmilan: Applied to both statements in 36e4a494e7. Confirmed the premise: `users.username` is `varchar(255)` with no unique index (only `PRIMARY (id)` and `UNIQUE KEY uuid`), and every other username lookup — `AuthUtils` login, the lockout counters, `getIdByUsername`, and this method's own uniqueness check — uses `BINARY`.

An independent security pass on the branch rated this higher than a consistency nit: under the default collation the non-BINARY form could overwrite a *different* existing user's `facility`/`billing_facility`, so it was a cross-account write, not just a stale value.

src/Services/UserService.php
zmilan: You are right, and the sentence is gone — bound as `?` placeholders appended to `$sqlBindArray` in 0e3dd8863b. I checked the three call sites you named plus `PersonPatientLinkService:340` and `VitalsCalculatedService:395`; the pattern is well established.

One correction to my own framing rather than to yours: the interpolated form was not actually unsafe, since both values are already ints and the limit is clamped to `QueryPagination::MAX_LIMIT`. So this is a consistency fix, and the value of it is mostly that the misleading comment stops telling the next person something untrue.

src/Services/Admin/UserManagementService.php
zmilan: Moved below the commit in 36e4a494e7, with the audit entry left inside the transaction as you suggested. It also matches the event's own docblock ("triggered after a user has been created") and the legacy `usergroup_admin.php` call site, which dispatches with no transaction at all.

A follow-up review caught that moving it out was only half the fix: unguarded after the commit, a throwing listener returns a 5xx for a user that *was* created, and the retry then fails with "Username already exists" — unrecoverable for the caller. So operational listener failures are now logged and absorbed.

The catch is narrowed rather than blanket, which is worth flagging since it reads oddly next to the house `catch (\Throwable)` guidance: `Core\ErrorHandler` promotes PHP warnings to `ErrorException`, so `catch (\Exception)` alone would swallow real defects. `ErrorException` is re-thrown and `\Error` subclasses never match `\Exception`, so genuine errors still propagate — which is also what `openemr.forbiddenCatchType` is there to enforce.

zmilan:
Pushed the four review items above, plus a few things an independent code + security pass over the branch turned up before I committed. Flagging them here since they are not tied to any thread.

**Also fixed**

- **ACL group enrichment matched case-insensitively** (`36e4a494e7`). `fetchAclGroupTitles()` keyed on `strtolower($username)`, and since `gacl_aro.value` has no unique index, `Smith` and `smith` could both hold AROs — each record then reported the *union* of both users' groups. The docblock claimed this preserved the semantics of the per-record `aclGetGroupTitles()` it replaced, but that is not what the legacy path did: `GaclApi::get_object_id()` returns `false` when more than one row matches, so it reported *no* groups on a collision. Neither behaviour is right; matching is now `BINARY` and keyed on the exact username.
- **Error classification broke on non-English installs** (`36e4a494e7`). The `match (true)` substring-matched English literals against messages `AuthUtils` builds with `xl()`, so every branch missed and everything fell through to `admin_password` — a rejected-because-too-short password was reported as an admin password error. Now compares against `xl()` of the same source strings.
- **Validator lengths exceeded the column widths** (`0f13065bf6`). `taxonomy` was validated at 255 against `varchar(30)`, `state_license_number` at 255 against `VARCHAR(25)`. Since `sql_mode = ''`, MySQL truncates silently: POST returns 201, GET returns something different.
- **Corrected a false claim in the transaction comment** (`36e4a494e7`). It said the ACL step rolls back. It does not — `AclExtended::setUserAro()` goes through `GaclApi`, whose constructor opens its own ADODB connection, so `gacl_aro` / `gacl_groups_aro_map` rows survive a rollback and are left orphaned against a username that no longer exists. Enrolling gacl in the caller's transaction is a real architectural change and did not belong in this PR, so the comment now states the limitation rather than denying it.

**Known, not fixed here**

- Two concurrent POSTs for the same username can both pass the `SELECT ... WHERE BINARY username` pre-check and both INSERT, since `users` has no unique index on `username`; the `UPDATE users SET uuid = ?` then matches both rows. Fixing it properly means having `AuthUtils::updatePassword()` return the insert id, which is shared legacy API used well beyond these endpoints. Happy to open a follow-up issue.
- A failed `admin_password` does not increment `users_secure.login_fail_counter`. Pre-existing — the legacy UI behaves identically — and reaching it already requires a valid `admin/super` token.

**Verification:** PHPStan level 10 full-codebase clean (no new baseline entries), `UserManagementApiTest` + `AuthorizationGrantFlowTest` 28/28, unit tests 7/7, phpcs clean. The security pass found no HIGH or MEDIUM issues; it did independently rate the missing `BINARY` as a cross-account write rather than a cosmetic inconsistency.

src/Services/Admin/UserManagementService.php
zmilan: Correction to my reply above — the narrowed `catch (\Exception)` I described did not survive CI, and it turns out the project already has a settled position on this that I had been working against.

`CatchExceptionToThrowableRector` rewrites any `catch (\Exception)` to `catch (\Throwable)`, and `openemr.forbiddenCatchType` then rejects a `catch (\Throwable)` that does not re-throw. Between them the two rules say `\Error` and `\ErrorException` must not be swallowed anywhere — so absorbing the listener failure would have meant suppressing both custom rules, which is not a good reason to keep a design.

Now logged with the username and uuid and re-thrown (`bcca43b4c0`), matching `interface/forms/questionnaire_assessments/native_save.php`. The post-commit ordering you asked for is unaffected.

That does leave the wart it was covering, and it is documented at the call site rather than papered over: a failing listener still leaves the caller with a 5xx for a user that exists, and the retry fails with "Username already exists". Deciding what the endpoint should actually return when the record is created but post-processing fails felt like a maintainer call rather than something to settle in a review fix — happy to follow whatever you prefer.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,437 @@
...
+        $authUtils = new AuthUtils();
+
+        // Wrap the DB writes on this connection in a transaction so a failure in any step
+        // (user row, UUID, facility, groups) rolls back the entire user creation.
+        //
+        // Caveat: the ACL write below does NOT participate. AclExtended::setUserAro() goes
+        // through GaclApi, whose constructor opens its own ADODB connection, so gacl_aro and
+        // gacl_groups_aro_map rows it creates survive a rollback here and are left orphaned
+        // against a username that no longer exists. Enrolling gacl in this transaction means
+        // giving it the caller's connection, which is beyond the scope of these endpoints.
```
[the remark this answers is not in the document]
zmilan: Real problem, fixed in 2002a36703 — though by removing the window rather than by either route you suggested, both of which turn out not to be available here.

Sharing the transaction connection means handing gacl the caller's connection, which reaches well past these endpoints. Compensating cleanup is not available either: `removeUserAros()` routes back through `setUserAro()` and reads `SELECT * FROM users WHERE username = ?`, which returns nothing once the row has been rolled back, and `setUserAro()` has no path that deletes the ARO object itself — it only drops group associations, so the ARO would survive any cleanup built from the existing helpers.

So the ACL write now runs after the commit. That cannot produce the orphaned state at all: if the audit step or the commit fails, `setUserAro()` has not run yet and there is nothing to strand. The failure mode becomes a user that exists with missing or partial ACL groups — visible in the user admin UI, and failing closed rather than granting anything.

Worth noting the consequence was sharper than "leaves rows behind": `setUserAro()` *edits* a matching ARO rather than creating one, so a later create of the same username would silently inherit the orphaned grants. That is the part that made this worth fixing now rather than deferring.

Covered by the existing round-trip test, which creates a user with `access_group: ["Physicians"]` and asserts the group comes back on the following GET.

zmilan:
On the second CodeRabbit finding — the binary-safe unique constraint on `users.username` (flagged outside the diff at `UserManagementService.php:153`).

The diagnosis is correct, and it matches what an earlier review pass found independently: the availability check runs before the transaction while `AuthUtils::updatePassword()` inserts later, so two concurrent POSTs for the same username can both pass the check and both insert. `users` carries only `PRIMARY (id)`, `UNIQUE KEY uuid` and `KEY abook_type` — nothing on `username`.

I am not going to add the constraint in this PR, and I want to be explicit about why rather than quietly skipping it.

A `UNIQUE` index on `users.username` is a schema change to a core table shared by the whole application — the login path, the user admin UI, LDAP/external sync, and every integration that has ever written a user. The table has permitted duplicates for the entire life of the schema, so the migration cannot simply add the index: it has to detect and resolve pre-existing duplicates on live installations first, and "resolve" there means merging or renaming real user accounts, which is not a decision an API PR should be making on an administrator's behalf. It also needs a deliberate collation choice, since binary-safe uniqueness and the case-insensitive default disagree about what counts as a duplicate.

That is a standalone change with its own migration, its own upgrade testing across supported MySQL and MariaDB versions, and maintainer buy-in on the duplicate-resolution policy. Phase 1 of the configuration management API is the wrong vehicle for it.

What this PR does do is avoid making the situation worse: the pre-check is `BINARY`-safe, and the uniqueness question is now documented rather than assumed. Happy to open a separate issue for the constraint and the duplicate-resolution migration if that is useful — say the word and I will write it up with the findings from this thread.

Also fixed from this review: the ACL atomicity finding, in `2002a36703` — details in the inline thread.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,443 @@
...
+        if ($usernames === []) {
+            return [];
+        }
+
+        $placeholders = implode(', ', array_fill(0, count($usernames), '?'));
+        $sql = "SELECT aro.value AS username, grp.name AS group_title
+                FROM gacl_aro aro
+                INNER JOIN gacl_groups_aro_map map ON map.aro_id = aro.id
+                INNER JOIN gacl_aro_groups grp ON grp.id = map.group_id
+                WHERE aro.section_value = 'users' AND BINARY aro.value IN (" . $placeholders . ")";
```
kojiromike: `gacl_aro` carries `UNIQUE KEY gacl_section_value_value_aro (section_value, value)`, so under the install default `utf8mb4_general_ci` (`Installer.class.php:355`) `smith` and `Smith` cannot both hold a `users` ARO — the union case the docblock guards against cannot arise.

What does happen: `acl_get_groups()` (`Gacl.php`) compares `o.value` with plain `Quote()`, no `BINARY`. So `Smith` resolves to `smith`'s single ARO and is *enforced* with `smith`'s groups, while this now reports none — an admin privilege view that under-reports. `get_object_id()` matches the same way.

Dropping `BINARY` keeps display and enforcement in agreement, and keeps that unique index usable (`BINARY aro.value` is non-sargable, so each list page scans `gacl_aro`).

```suggestion
                WHERE aro.section_value = 'users' AND aro.value IN (" . $placeholders . ")";
```

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,443 @@
...
+                $errorMsg = is_string($rawError) && $rawError !== '' ? $rawError : 'User creation failed';
+                // AuthUtils builds these messages with xl(), so they arrive translated. Comparing
+                // against English literals would misclassify every failure on a non-English
+                // install; running the same source strings through xl() here matches in any locale.
+                $field = match (true) {
+                    $errorMsg === xl("Incorrect password!") => 'admin_password',
+                    $errorMsg === xl("Password not long enough"),
+                    $errorMsg === xl("Empty Password Not Allowed") => 'password',
+                    $errorMsg === xl("Trying to create user with existing username!") => 'username',
+                    default => 'admin_password',
```
kojiromike, line 271: Two more create-path messages land in this default arm: `xl("Password too long")` (`AuthUtils.php:657`) and `xl("Password not strong enough")` (`:665`). Both reject the *new* user's password, but get reported against `admin_password` — so the caller is told their admin credential is wrong.

Not a regression (the old substring form missed them too), but this is the moment to close it — two more arms mapping both to `password`.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,443 @@
...
+            QueryUtils::rollbackTransaction(); // @phpstan-ignore openemr.deprecatedSqlFunction
+            throw $e;
+        }
+
+        // Moved out of the transaction deliberately. AclExtended::setUserAro() writes through
+        // GaclApi, which holds its own ADODB connection, so its rows were never covered by the
+        // rollback above: a failure in the audit step or the commit left gacl_aro and
+        // gacl_groups_aro_map rows behind, keyed to a username whose users row had just been
+        // rolled away. setUserAro() then *edits* a matching ARO rather than creating one, so the
+        // next create of that username would silently inherit the stale grants.
```
kojiromike: This sentence overstates it. `setUserAro()`'s edit branch walks `get_object_groups(..., 'NO_RECURSE')` and calls `del_group_object()` on every current association before adding the selected ones — so the next create of that username *overwrites* the stale grants rather than inheriting them.

The orphaned-row problem the rest of the comment describes is real and does justify the move. Worth trimming just this claim so the next reader does not take it as load-bearing.

tests/Tests/Unit/Common/Auth/OpenIDConnect/Entities/ServerScopeListEntityTest.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/**
+ * Tests the consistency of the standard API scope registry.
+ *
+ * @link      https://www.open-emr.org
+ * @author    Milan Zivkovic <zivkovic.milan@gmail.com>
+ * @copyright Copyright (c) 2026 Milan Zivkovic
```
kojiromike: Every other file in this PR uses `Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>`. Worth matching here.

```suggestion
 * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
```

kojiromike (approved):
Approved again…

---

All four items from the last round are resolved, and the reasoning in each reply holds up under checking: `Gacl::__construct()` does build its own ADODB connection via `DatabaseConnectionFactory::createAdodb()`, so the ACL write genuinely was outside the rollback; the validator widths match `sql/database.sql` (`taxonomy varchar(30)`, `state_license_number VARCHAR(25)`); and the v2 scope derivation is right.

Three notes plus a header nit, none blocking.

The one I would act on is the `BINARY` in `fetchAclGroupTitles()`. `gacl_aro` does carry a unique key, and gacl's own lookups are not binary — so the reported groups now diverge from the enforced ones in the opposite direction from the docblock's reasoning.

src/Services/Admin/UserManagementService.php
zmilan: Fixed in 70371f846b — you were right about the direction, though the one-line change turned out to be half of it.

Confirmed the premises before acting: `gacl_aro` carries `UNIQUE KEY gacl_section_value_value_aro (section_value, value)` (`sql/database.sql:2644`), the installer defaults to `utf8mb4_general_ci` (`library/classes/Installer.class.php:355`), and both `Gacl::acl_get_groups()` (`Gacl.php:528`) and `GaclApi::get_object_id()` (`GaclApi.php:2747`) compare with a plain `Quote()`. So the union case my docblock guarded against cannot arise, and the docblock's "gacl_aro.value has no unique index" was the opposite of the schema — it is rewritten rather than patched.

The part the suggestion does not reach: the map is keyed by `aro.value` but looked up by `users.username` with an exact PHP string comparison. Dropping `BINARY` alone fetches the case-differing row and then discards it at the array lookup — the same under-reporting, one line later. So the fix is the plain `IN()` **plus** folding both sides.

The two sides still fold by different rules (collation vs `mb_strtolower`). They agree over the ASCII set the validator admits, and the collation folds at least as much, so the query returns a superset: a row the collation matched but `mb_strtolower` would not is keyed under a fold no user matches and is dropped. That can only under-report, never attribute one user's groups to another. The docblock now records this rather than leaving it to be rediscovered.

Pinned by `testGetReportsAclGroupsWhenAroCaseDiffers`, which flips a created user's stored ARO case directly. I checked each half in isolation — restoring `BINARY` fails it, and keeping the SQL fix while removing the folding fails it too.

src/Services/Admin/UserManagementService.php
zmilan: Both arms added in 7a558ab7fb, mapping to `password`.

Confirmed all four messages come from the same block validating the *new* user's password — `Empty Password Not Allowed` (`AuthUtils.php:641`), `Password not long enough` (`:649`), `Password too long` (`:657`), `Password not strong enough` (`:665`) — and that they are raised after the admin credential has already been verified at `:616`. So attributing them to `admin_password` was telling the caller their own credential was wrong.

One detail worth recording for the next reader: `testMaximumPasswordLength()` sets a longer message of its own at `:1048`, but the caller at `:657` overwrites it, so matching the literal `xl("Password too long")` is the correct thing to key on.

Covered by `testPostTooLongPasswordReportsPasswordField` and `testPostWeakPasswordReportsPasswordField`, both asserting `password` is present and `admin_password` absent. Reverting the two arms fails both.

src/Services/Admin/UserManagementService.php
zmilan: You are right — trimmed in cefdb1d52f.

`setUserAro()`'s edit branch does walk `get_object_groups(..., 'NO_RECURSE')` and call `del_group_object()` on every current association before adding the selected ones (`AclExtended.php:282-285`), so the next create overwrites the stale grants rather than inheriting them. The comment now claims only the orphaned rows, which is the part that actually justifies the move.

The same wrong claim was in my reply on the CodeRabbit thread above, which it recorded as a learning — corrected there too.

tests/Tests/Unit/Common/Auth/OpenIDConnect/Entities/ServerScopeListEntityTest.php
zmilan: Matched in cefdb1d52f.

src/Services/Admin/UserManagementService.php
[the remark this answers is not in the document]
zmilan: Correction to my reply above, caught in review by @kojiromike.

I wrote that because `setUserAro()` edits a matching ARO rather than creating one, "a later create of the same username would silently inherit the orphaned grants." That is wrong. The edit branch walks `get_object_groups(..., 'NO_RECURSE')` and calls `del_group_object()` on every current association before adding the selected ones (`AclExtended.php:282-285`), so a later create **overwrites** the stale memberships rather than inheriting them.

The orphaned-row problem itself is real and still justifies running the ACL write after the commit — only that one consequence was overstated. The code comment is corrected in cefdb1d52f.

Flagging it because it was recorded as a learning: the sentence about inheriting stale grants should not be carried forward.

zmilan:
## Re-review needed — this push goes beyond the four review notes

@kojiromike thanks for the approval. Pushing five commits on top of it, and **three of them are changes you have not seen**, one of which alters the endpoint's contract. Flagging that up front rather than letting an approval carry over something it did not cover.

### ⚠️ Behavior change

**`POST /api/admin/users` now rejects a username that differs from an existing user only by case** (`d3cf62d9cc`). Previously `SMITH` was accepted alongside an existing `smith`.

This is a security fix, not a tidy-up. The uniqueness check matched with `BINARY`, but the ARO namespace the create writes into is case-insensitive — `GaclApi::get_object_id()` and `Gacl::acl_get_groups()` both use a plain equality compare, and `gacl_aro` carries `UNIQUE KEY (section_value, value)`, so a username has exactly one ARO regardless of case. `AclExtended::setUserAro()` therefore resolved *`smith`'s*  ARO, renamed it, deleted every group association it held, and installed the ones from the request. An endpoint scoped `user/users.c` — create only — could rewrite another account's ACL memberships, leaving only a user-creation entry in the audit log.

Both guards that should catch this miss for the same reason: `getIdByUsername()` looks up with `BINARY`, so `gacl_protect` is read for the wrong user (and returns `false` for a user with no settings row anyway), and the literal `$user_name == "admin"` does not match `"Admin"`. `users.username` has no unique index and `AuthUtils::updatePassword()` is called with `$targetUser = 0`, so its own duplicate check never fires — this was the only guard.

I reproduced it before fixing: with `BINARY` restored, the case-variant create returns `201` and the existing user's groups are replaced. `usergroup_admin.php:343` matches case-insensitively and refuses the same create, so this restores parity with the UI rather than inventing a rule — but it is still a stricter API contract than what you approved.

### Also new, and stricter than before (`8b0cc2707e`)

Two more ways a create could return `201` for a user that is not actually provisioned:

- **`username` is now capped at 150, not 255.** `setUserAro()` registers it as a `gacl_aro.value`, which is `varchar(150)`, and OpenEMR sets `sql_mode = ''` — so a longer name truncated in the ARO while the `users` row kept the full string, the lookup then matched nothing, and every group mapping was silently dropped. `add_object()` independently refuses a value ≥ 230, so 150–229 was exactly the silent window.
- **The assembled `fname [mname] lname` is now bounded.** It becomes `gacl_aro.name`, and `add_object()` returns `false` outright for a name ≥ 255 (`GaclApi.php:2963`). Each part was bounded at 255 individually, so the combination could reach 767 — no ARO created at all, and the user got none of its requested groups. It is a cross-field callback on `lname`; `testPostNameAtCombinedBoundSucceeds` pins the other side at 254 bytes so the bound is not over-tight.
- **`taxonomy` is only sent when supplied**, so an omitted value no longer writes `''` over `varchar(30) NOT NULL DEFAULT '207Q00000X'`. A provider created here was carrying a blank taxonomy into 837 claim generation, unlike one created in the UI.

### Your four notes

All four are addressed and answered inline. Two corrections worth surfacing here:

- On the `BINARY` in `fetchAclGroupTitles()` — your reasoning was right, but the suggested diff is half the fix. The map is keyed by `aro.value` and looked up by `users.username` with an exact PHP comparison, so dropping `BINARY` alone fetches the case-differing row and discards it at the array lookup. Both sides needed folding.
- On the overstated comment — you were right, and the same wrong claim was in my earlier reply on the CodeRabbit thread, which it stored as a learning. Corrected in both places.

### How these were found

Two independent review passes over the branch after your notes were addressed — one general code review, one security review. The case-variant hole was found by both independently. Every fix here was reproduced as failing first: I reverted each guard and confirmed the bad outcome (`201` where a `400` belongs, `''` where the default belongs) rather than assuming the reasoning held.

### Verification

34 API tests / 108 assertions, PHPStan level 10 clean on the full codebase, phpcs and rector clean. Seven new tests, each confirmed to fail with its own fix reverted.

### Deliberately not done

- **`GET` does not apply the `iHavePermissionsOf()` filter** that the legacy list uses at `usergroup_admin.php:694`. Rated below the bar because no such principal exists in a stock install — only `Administrators` and `Emergency Login` hold `admin|users`, and both also hold `admin|super` — and the pre-existing `GET /api/user` already exposes the same PII columns under identical ACL. Worth a follow-up if operators are expected to split those ACOs; your call whether it blocks.
- Check-then-insert race on username (wants a unique index + migration), `_count`/`_maxresults` duplicated in pagination links, and `BINARY uuid = ?` from `SearchModifier::EXACT` bypassing the unique index on detail lookups. Happy to file these as issues.

src/Validators/Admin/UserValidator.php
```diff
@@ -0,0 +1,96 @@
...
+                            $raw = $values[$key] ?? null;
+                            $part = is_string($raw) ? trim($raw) : '';
+                            if ($part !== '') {
+                                $parts[] = $part;
+                            }
+                        }
+                        $parts[] = is_string($value) ? trim($value) : '';
+
+                        // strlen(), not mb_strlen(): GaclApi::add_object() bounds bytes.
+                        return strlen(implode(' ', $parts)) < 255;
```
kojiromike: A caller sending `fname` 130 + `lname` 130 — both inside the per-field bounds, and the spec documents no other constraint — gets back `lname is invalid`, with no hint that the combined length is what failed. `CoverageValidator` throws `InvalidValueException` from its callbacks to name the reason; the same shape would work here.

src/Validators/Admin/UserValidator.php
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                );
+                $context->optional('mname')->lengthBetween(0, 255);
+                $context->optional('suffix')->lengthBetween(0, 255);
+                $context->optional('email')->email();
```
kojiromike: `email` is the one varchar column in this list without a width bound. `users.email` is `varchar(255)`, and under `sql_mode = ''` a longer address truncates silently — the same failure mode the username and combined-name bounds were just added to prevent.

```suggestion
                $context->optional('email')->email()->lengthBetween(0, 255);
```

kojiromike (approved):
Both items from the last round hold up. The create-time uniqueness check now matches the ARO namespace's case-insensitivity, which is also what the legacy UI does (`usergroup_admin.php:343`), and the enrichment fold keeps the reported groups in step with what gacl enforces.

The new bounds line up with the code they guard — `add_object()` rejects `strlen($name) >= 255` (`GaclApi.php:2963`) and `testPostNameAtCombinedBoundSucceeds` pins 254 — and omitting `taxonomy` is safe because `updatePassword()` builds its INSERT from `array_keys($userData)`.

Two small gaps in the width sweep, neither blocking.

src/Validators/Admin/UserValidator.php
zmilan: Fixed in 442367aa86.

The callback now throws `InvalidValueException`, which `Particle\Validator\Rule\Callback::validate()` catches and substitutes for its `'{{ name }} is invalid'` template, keyed by `getIdentifier()` — the same mechanism `CoverageValidator` and `PatientValidator` use, as you pointed at. The response now carries `Record::INVALID_COMBINED_NAME_LENGTH` with the reason and the actual byte count, rather than blaming a field that is within its own bound.

`testPostOverlongCombinedNameReturns400` asserts the reason key and message text; restoring the bare `return false` fails it on the missing key.

One thing your comment surfaced that I have not fixed here: you noted the spec documents no other constraint, and that is literally true — the `requestBody` at `UserManagementRestController.php:157` declares every property as a bare `type: 'string'` with no `maxLength` on any of them, while the validator enforces six length rules (`username` 150, the combined name, `email` 255, `npi` 15, `taxonomy` 30, `state_license_number` 25). So a caller still cannot learn any limit from the spec up front. Happy to add the annotations in this PR or file it as a follow-up — say which you prefer.

src/Validators/Admin/UserValidator.php
zmilan: Added in 442367aa86, as suggested.

Worth recording what I found while checking it, because it cuts against the stated failure mode but not against the fix.

Particle's `email()` rule is nothing but `filter_var($value, FILTER_VALIDATE_EMAIL)`, and PHP caps a valid address at **254 bytes** (RFC 5321). I swept ~317k generated addresses — local part 1–70, 1–8 domain labels of 1–63 bytes, TLD 2–10 — and 254 is the longest accepted, identical on PHP 8.3.33, 8.4.25 and 8.5.6, the range `composer.json` supports. So an address that could truncate in `varchar(255)` cannot reach the column today: `email()` already rejects it.

My first instinct was therefore to leave the bound out as unreachable. That was wrong, and the bound is in. The 254 cap is an artifact of the regex in `ext/filter` — the manual documents RFC 822 syntax "with some limitations" and says nothing about length. Using it to guarantee a constraint that belongs to our schema is exactly the kind of implicit coupling that breaks quietly, and "unreachable because another component happens to behave this way" is where a defensive bound earns its keep rather than where it is redundant.

One related detail: `users.email` is utf8mb4, so it is 255 *characters* / 1020 bytes, while Particle's `lengthBetween` measures *bytes* via `strlen()`. Those coincide only because `filter_var` rejects non-ASCII without `FILTER_FLAG_EMAIL_UNICODE` — a second implicit dependency on the same undocumented behaviour. The bound errs safe either way, since bytes are never fewer than characters.

`testPostOverlongEmailReturns400` sends a 256-byte address so it is over both bounds. Removing `email()` turns it from 400 back into 201 without the width bound, and leaves it at 400 with it — which is the layering doing its job. Note no test can isolate the bound on its own, since `filter_var` will not accept a valid address long enough to reach it.

PR #10119, opened 2026-01-06
refactor(bootstrap): migrate from Bootstrap 4 to 5.3.3

rollingventures:
# refactor(bootstrap): migrate from Bootstrap 4 to 5.3.3

Closes #7932

## Summary

This PR upgrades OpenEMR from Bootstrap 4.6.2 to Bootstrap 5.3.3, following the [official Bootstrap 5 Migration Guide](https://getbootstrap.com/docs/5.3/migration/).

**671 files changed** across 27 commits covering CSS class migrations, data attribute updates, JavaScript API modernization, theme fixes, and portal cleanup.

### Dependencies & Build System

- **Bootstrap**: 4.6.2 → 5.3.3
- **Bootswatch**: Removed (only used in Solar theme; `$primary` color defined directly)
- **DataTables**: Updated to Bootstrap 5 theme
- **Select2**: Updated to Bootstrap 5 theme
- **Popper.js**: Added as explicit dependency (no longer bundled)
- **Sass compiler**: `gulp-dart-sass` → `gulp-sass`
- **RTL**: Removed `bootstrap-rtl` package (Bootstrap 5 has native RTL support via `dir="rtl"`)
- **Gulp build**: Renamed SCSS injection markers from `bs4import` → `bs5import`

### CSS Class Migrations

| Category | Bootstrap 4 | Bootstrap 5 |
|----------|-------------|-------------|
| **Directional** | `float-left`, `float-right` | `float-start`, `float-end` |
| **Margins** | `ml-*`, `mr-*` | `ms-*`, `me-*` |
| **Padding** | `pl-*`, `pr-*` | `ps-*`, `pe-*` |
| **Text alignment** | `text-left`, `text-right` | `text-start`, `text-end` |
| **Borders** | `border-left`, `border-right` | `border-start`, `border-end` |
| **Forms** | `form-group`, `form-row` | `mb-3`, `row gx-2` |
| **Buttons** | `btn-block` | `d-grid` wrapper or `w-100` |
| **Close button** | `close` + `&times;` | `btn-close` |
| **Screen readers** | `sr-only` | `visually-hidden` |
| **Font weight** | `font-weight-bold` | `fw-bold` |
| **Font style** | `font-italic` | `fst-italic` |
| **Monospace** | `text-monospace` | `font-monospace` |
| **Muted text** | `text-muted` | `text-body-secondary` |
| **Badges** | `badge-*` | `bg-*` + `text-bg-*` |
[...]

Issue #10109, opened 2026-01-05
Move dygraphs to npm package management


kojiromike:
## Summary

The `public/assets/modified/dygraphs-2-0-0/` directory contains a vendored copy of the Dygraphs charting library.

## Current State

- Located in `public/assets/modified/` suggesting possible local modifications
- Used in ~8 files
- Package available as `dygraphs` on npm

## Proposed Solution

1. Verify if local modifications exist (compare with upstream 2.0.0)
2. If unmodified: add `dygraphs` to `package.json` and update build process
3. If modified: document modifications or upstream them

## Benefits

- Easier updates
- Smaller git repo
- Clear tracking of any local modifications

bradymiller:
hi @kojiromike , Check out the README for the modification:
https://github.com/openemr/openemr/blob/master/public/assets/modified/dygraphs-2-0-0/README.md

Issue #10108, opened 2026-01-05
Move lforms/ to npm package management


kojiromike:
## Summary

The `interface/forms/questionnaire_assessments/lforms/` directory contains a vendored copy of LHC-Forms library. This could be managed via npm instead.

## Current State

- 30 files vendored
- Webpack bundles and source maps
- Package available as `lhc-forms` on npm

## Proposed Solution

1. Add `lhc-forms` to `package.json`
2. Update build process to copy/reference dist files
3. Remove vendored files from git

## Benefits

- Easier updates via `npm update`
- Smaller git repo
- Security patches via dependabot

adunsulag:
@sjpadgett you had a modified version of lhc-forms due to some kind of high cpu usage parsing problem right?  I can't recall but didn't you mention that was now resolved?  It would be nice to use package manager instead of the vendored library.

sjpadgett:
@adunsulag I fixed having to build library and am now using vendor library but don't recall which version.

Issue #10107, opened 2026-01-05
Move swagger/ to npm package management


kojiromike:
## Summary

The `swagger/` directory contains a vendored copy of Swagger UI distribution. This could be managed via npm instead.

## Current State

- 19 files vendored in `swagger/`
- Includes non-minified webpack bundles like `swagger-ui-bundle.js`
- Last updated in PR #6240

## Proposed Solution

1. Add `swagger-ui-dist` to `package.json`
2. Update build process to copy dist files to `swagger/` (or serve from `node_modules`)
3. Remove vendored files from git

## Benefits

- Easier updates via `npm update`
- Smaller git repo
- Security patches via dependabot

PR #11231, opened 2026-03-23
build(webpack): replace gulp with webpack for theme compilation

### LINKED ISSUE #7933: Remove gulp-dart-sass and replace with gulp-sass
As part of the Bootstrap 5 upgrade, the current gulp-dart-sass has not been updated for quite some time. To make sure sass stays current, I will be replacing gulp-dart-sass with gulp-sass and sass.

rollingventures:
Fixes #7933
Related: #10107, #10108, #10109

Replaces the unmaintained Gulp build pipeline with Webpack for SCSS theme compilation. Gulp 4 and `gulp-dart-sass` have no successors — this migration eliminates 15 dead
devDependencies and delivers 160x faster cached rebuilds (131ms vs ~20s) while producing identical CSS output at the same paths.

Replace the entire Gulp build pipeline with Webpack and plain Node scripts.

**Why:**
- `gulp-dart-sass` is unmaintained, Gulp 4 has no successor — security advisories accumulate with no upstream fixes
- Webpack is already a project dependency (module React builds) — running both doubled the Node tooling surface
- Webpack filesystem cache: 131ms cached vs ~20s Gulp (160x faster). Docker BuildKit cache mounts enable ~100ms rebuilds in CI

**What changed:**
- `webpack.themes.js`: compiles all 41 SCSS theme variants via resource queries (`?variant=compact`, `?variant=rtl`, etc.)
- `webpack/loaders/sass-bsimport-loader.js`: variant transforms replacing `gulp-append-prepend` and `gulp-inject-string`
- `webpack/loaders/strip-bom-loader.js`: BOM fix for FontAwesome SCSS
- `scripts/install-assets.js`: replaces gulp's asset copy task
- `scripts/sync-css.js`: replaces gulp's static CSS sync task
- `gulpfile.js`: deleted (593 lines)
- `package.json`: -15 gulp deps, +13 webpack deps, updated scripts
- `ajax_calendar_sass.scss`: added `!optional` to `@extend` directives
- `CLAUDE.md`: updated build tool references (Gulp 4 → Webpack 5) and build commands

**Files:**

- **Added** `webpack.themes.js` — single config compiling all 41 SCSS theme variants using `?variant=` resource queries for compact, RTL, RTL-compact, RTL-tabs, and RTL-misc transforms
- **Added** `webpack/loaders/sass-bsimport-loader.js` — custom pre-loader replacing `gulp-append-prepend` and `gulp-inject-string` plugin transforms
- **Added** `webpack/loaders/strip-bom-loader.js` — strips UTF-8 BOM from FontAwesome SCSS that causes postcss parse errors
- **Added** `scripts/install-assets.js` — replaces `gulp -i` asset installation task, runs as npm `postinstall` hook
- **Added** `scripts/sync-css.js` — replaces `gulp sync` task, copies 3 static CSS files with autoprefixer
- **Deleted** `gulpfile.js` (593 lines)
- **Modified** `package.json` — removed 15 gulp devDependencies, added 13 webpack devDependencies, updated all build scripts
- **Modified** `ajax_calendar_sass.scss` — added `!optional` to 10 `@extend` directives required for webpack's sass-loader compilation
- **Modified** `.gitignore` — added `.webpack-cache/`
- **Modified** `.github/dependabot.yml` — updated build-tools group from `gulp*` to `webpack*`
- **Modified** `interface/README.md` — updated build documentation
- **Modified** `CLAUDE.md` — updated technology stack and build commands
- **Modified** `jest.config.js` — removed deleted `gulpfile.js` from coverage ignore

**Verification:**
- 44 CSS files produced (41 webpack + 3 sync-css) — identical file list to Gulp
- All 105 Jest tests pass
- Tested in `development-easy` Docker environment — confirmed webpack cache present, gulp removed, themes rendering correctly
- ESLint: 0 errors on changed files (27 warnings — all pre-existing Node.js global references)
- PHPStan: no errors in changed files (26 pre-existing errors in unrelated `interface/react/entry.php`)
- PHPCS: no errors in changed files (1 pre-existing error in unrelated `interface/react/entry.php`)
- Cold build: 16.7s | Cached rebuild: 131ms
- Rebased on latest upstream/master (2026-03-25)

**Looking ahead:**

Aside from resolving the immediate Gulp deprecation, webpack positions us to modernize the frontend build infrastructure incrementally — CSS minification via `css-minimizer-webpack-plugin`, elimination of the `napa` dependency by handling non-npm packages directly, and a `buildModuleConfig()` helper already in `webpack.themes.js` that lets future custom modules opt into root-level bundling without maintaining their own webpack configs. It also unblocks the Bootstrap 5.3.3 migration (#10119) by providing a build pipeline that can actually evolve with the framework.

I conducted manual testing with local docker compose and k8s. Manually checked all output CSS and Theme functionality.


package.json
```diff
@@ -4,12 +4,12 @@
...
-    "build": "gulp -b",
-    "dev-build": "npm run gulp-build",
-    "dev": "gulp --dev && gulp watch",
-    "gulp-build": "gulp -b --dev",
-    "gulp-watch": "gulp watch",
+    "postinstall": "napa && node scripts/install-assets.js",
+    "build": "npm run build:webpack && npm run build:sync",
+    "build:webpack": "webpack --config webpack.themes.js --mode production",
+    "build:sync": "node scripts/sync-css.js",
+    "dev": "npm run build:webpack -- --mode development && npm run watch",
```
kojiromike: Should `dev` also run `build:sync` before starting the watcher? The old `gulp --dev` ran the full default task (which included `sync`), so static CSS files in `interface/themes/*.css` were copied to `public/themes/` before watching started. Currently a developer running `npm run dev` on a fresh checkout would be missing `ajax_calendar_ie.css`, `jquery.autocomplete.css`, and `rtl_style_pdf.css` until they separately run `npm run build:sync`.

```suggestion
    "dev": "npm run build:webpack -- --mode development && npm run build:sync && npm run watch",
```

webpack.themes.js
```diff
@@ -0,0 +1,313 @@
...
+/**
+ * Webpack build for OpenEMR SCSS themes and custom module React assets.
+ *
+ * Invoked by `npm run build` and `npm run build:webpack`.
+ *
+ * Usage:
+ *   npm run build             # Webpack (all themes + React)
+ *   npm run build:webpack     # Webpack only (dev/CI iteration)
+ *   npm run watch:webpack     # Webpack dev build with file watching
+ *
```
kojiromike, line 15: Minor: the comment references `npm run watch:webpack` but the package.json script is named `watch`. Would it make sense to update the comment to match?

```suggestion
 *   npm run watch             # Webpack dev build with file watching
```

package.json
```diff
@@ -40,34 +40,31 @@
     "bb >= 10"
   ],
   "devDependencies": {
+    "@babel/core": "^7.24.0",
```
kojiromike: Are the babel-related deps (`@babel/core`, `@babel/preset-env`, `@babel/preset-react`, `@babel/preset-typescript`, `babel-loader`) needed right now? They're only referenced by the unused `buildModuleConfig` scaffold function in `webpack.themes.js`. If the intent is forward-looking for custom module React builds, that makes sense — just want to confirm it's intentional since it adds several packages to every `npm install`.

webpack.themes.js
```diff
@@ -0,0 +1,313 @@
...
+            includePaths: [
+              path.resolve(__dirname, "node_modules"),
+              path.resolve(__dirname, "interface/themes"),
+              path.resolve(__dirname, "public/assets"),
+              path.resolve(__dirname, "public/assets/modified"),
+            ],
+            importer: publicAssetsImporter,
+          },
+          // webpackImporter: true ensures every transitively imported SCSS file
+          // also passes through the sass-bsimport-loader pre-loader chain.
```
kojiromike: Nit: this comment is slightly misleading. `webpackImporter: true` tells sass to use webpack's module resolution for `@import` statements, but it doesn't cause transitively imported SCSS files to pass through the webpack loader chain (loaders only run on entry files). The transitive `public/assets/` path resolution is actually handled by the `publicAssetsImporter` custom sass importer above — which works well. Maybe update the comment to reflect that?

Something like: "webpackImporter: true uses webpack's module resolution for @import, complementing the custom publicAssetsImporter for public/assets paths."

webpack.themes.js
```diff
@@ -0,0 +1,313 @@
...
+// The rsync in the Dockerfile picks up *.js / *.css from anywhere in the tree,
+// so these bundles land in the nginx image at the correct sub-path.
+//
+// To add a new module, copy the block below, uncomment, and update paths:
+//
+// To add a new module, copy this block and update paths.
+// ---------------------------------------------------------------------------
+// eslint-disable-next-line no-unused-vars -- scaffold helper for new module configs
+function buildModuleConfig(name, entryFile, outputDir) {
+  return {
```
kojiromike: Would it make sense to defer this scaffold (and the babel deps above) to a follow-up PR when a custom module actually needs it? Keeping it here is fine for readability as a template, but the eslint-disable + unused deps add a small maintenance surface.

kojiromike:
If you rebase the spellcheck error will go away.

webpack.themes.js
rollingventures: Updated the file header to use npm run watch so it matches package.json.

package.json
rollingventures: Done. npm run dev now runs build:webpack (development), then build:sync, then watch, so a fresh clone gets the three static CSS files under public/themes/ before the watcher starts—same idea as the old Gulp default task. Thanks for catching that.

package.json
rollingventures: They aren’t needed for this PR anymore. We removed the unused buildModuleConfig scaffold, so @babel/* and babel-loader are dropped from devDependencies. We can add a dedicated module bundling setup in a follow-up when a module actually needs it.

webpack.themes.js
rollingventures: Good catch. Replaced that comment: webpackImporter: true is documented as webpack’s resolution for @import, and we spell out that public/assets/... handling is the custom publicAssetsImporter, not “loaders on every transitive file.”

webpack.themes.js
rollingventures: Agreed—we removed the scaffold and the Babel deps from this PR to keep the change focused. A future PR can introduce module webpack config when there’s a real entry point to wire up.

kojiromike (commented):
Good progress on the Gulp→Webpack migration. The earlier feedback about babel deps, scaffold removal, dev script ordering, and comments has been addressed. A few items remain from earlier threads (still open), plus some new findings:

### `--mode` duplication is still broken (existing thread, reminder)

`dev` calls `build:webpack` (hardcodes `--mode production`) then appends `--mode development`. Webpack CLI uses the **last** `--mode`, but the config's `isProduction` detection at line 33 reads the **first** `--mode` from `process.argv`. Net effect: **dev builds silently compile in production mode** (wrong `mode`, wrong `devtool`).

The cleanest fix: export the webpack config as a function `(env, argv) => { ... }` and use `argv.mode` instead of parsing `process.argv`. Then remove the hardcoded `--mode` from `build:webpack` and pass it from each npm script.

### PHP file sync may be dropped

The old Gulp `sync()` task copied both `*.css` and `*.php` from `interface/themes/` to `public/themes/`. The new `sync-css.js` only copies `*.css` files. If `ajax_calendar_ie.css`, `jquery.autocomplete.css`, and `rtl_style_pdf.css` are the only files that matter, this is fine. But if any PHP files were being synced (the old Gulp code explicitly included `*.php`), they'll be silently missing. Worth confirming.

### `output.clean: false` leaves stale artifacts

The old Gulp pipeline ran `del.sync()` before each build. With `clean: false`, renamed or removed theme entries leave stale CSS/JS in `public/themes/`. This affects CI and Docker builds where the output directory persists. Consider `clean: true` on production builds.

### `!optional` on `@extend` directives in `ajax_calendar_sass.scss`

The diff adds `!optional` to many `@extend` statements. This silences errors when the extended selector doesn't exist, which prevents build failures — but it also means if Bootstrap isn't properly imported, the extends silently produce no CSS and the calendar layout breaks without any error. The old Gulp build didn't need `!optional`. Is this papering over a resolve-path difference in the webpack config?

package.json
```diff
@@ -4,12 +4,12 @@
...
-    "build": "gulp -b",
-    "dev-build": "npm run gulp-build",
-    "dev": "gulp --dev && gulp watch",
-    "gulp-build": "gulp -b --dev",
-    "gulp-watch": "gulp watch",
+    "postinstall": "napa && node scripts/install-assets.js",
+    "build": "npm run build:webpack && npm run build:sync",
+    "build:webpack": "webpack --config webpack.themes.js --mode production",
+    "build:sync": "node scripts/sync-css.js",
+    "dev": "npm run build:webpack -- --mode development && npm run build:sync && npm run watch",
```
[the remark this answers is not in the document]
kojiromike: This is still an issue — dev builds silently run in production mode because `isProduction` reads the first `--mode` flag from `process.argv`. Please either fix with a code change (e.g., export the config as a function and use `argv.mode`) or explain why the current behavior is acceptable.

webpack.themes.js
```diff
@@ -0,0 +1,258 @@
...
+
+const path = require("path");
+const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+
+// webpack --mode production sets NODE_ENV=production via DefinePlugin for
+// bundled code; also check argv so the config itself sees the right mode.
+const isProduction =
+  process.env.NODE_ENV === "production" ||
+  (process.argv.includes("--mode") &&
+    process.argv[process.argv.indexOf("--mode") + 1] === "production");
```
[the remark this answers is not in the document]
kojiromike: Still open — `isProduction` detection via `process.argv` scanning is fragile with the current npm scripts. Please address with a code change or explain why this doesn't need fixing.

package.json
```diff
@@ -42,32 +42,24 @@
   "devDependencies": {
     "@types/jest": "29.5.14",
     "autoprefixer": "10.4.27",
-    "colors": "1.4.0",
-    "del": "6.1.1",
+    "css-loader": "^6.8.1",
     "eslint": "9.39.4",
     "eslint-plugin-import": "2.32.0",
```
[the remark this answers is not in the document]
kojiromike: Still open — new devDependencies use caret ranges while existing ones are pinned. Please either pin these versions to match the existing convention or explain why ranges are preferred here.

webpack.themes.js
```diff
@@ -0,0 +1,259 @@
...
+// Helper to build an absolute entry path with an optional resource query.
+function entry(relPath, variant) {
+  const abs = path.resolve(themesDir, relPath);
+  return variant ? abs + "?variant=" + variant : abs;
+}
+
+const themesConfig = {
+  name: "themes",
+  mode: isProduction ? "production" : "development",
+  entry: {
```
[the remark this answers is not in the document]
kojiromike: Still open — production CSS output isn't minified, which is a regression from the Gulp pipeline's `csso()` step. Please add a CSS minimizer (e.g., `css-minimizer-webpack-plugin`) or explain why unminified production CSS is acceptable.

rollingventures:
All feedback has been addressed in the latest push:

1. **`--mode` handling** — Fixed. Converted `webpack.themes.js` to export a function `(env, argv) => {...}` so webpack CLI passes the resolved mode via `argv.mode`. No more `process.argv` parsing.

2. **CSS minification** — Fixed. Added `css-minimizer-webpack-plugin` to the optimization config.

3. **Caret ranges** — Fixed. All devDependencies are now pinned (no `^` prefixes).

4. **`.dockerignore` comment** — Removed the misleading comment about `openemr-internal/.webpack-cache`.

5. **`!optional` flags** — Removed. These were masking a local issue (stale BS5 files in `public/assets/`), not a real resolve-path problem. The `@extend` directives work correctly with Bootstrap 4.6.2.

kojiromike:
@aisle-analyzer

stephenwaite:
hi @rollingventures, we've got an [up for grabs demo](https://www.open-emr.org/wiki/index.php/Development_Demo#Alpha_-_Up_For_Grabs_Demo) but it didn't fire up b/c of a [php fatal error](https://four.openemr.io/log/logPhp.txt).

rollingventures:
Rebased onto current master (resolves the prior spellcheck/conflict feedback). The `Built-in webserver 8.2 / api` job failed on a transient Docker registry pull (`Docker pull failed with exit code 1` after the built-in backoff retries) — unrelated to this build-tooling change; 8.3/8.4/8.5 of the same job passed. Reopening to re-trigger CI.

kojiromike (approved):
All prior review threads are addressed in the current HEAD — confirmed by reading the files at `ab75213`, not just the resolution comments.

Earlier feedback (mine + Copilot):
- `dev` now runs `build:webpack:dev && build:sync && watch`
- config exports `(env, argv) => …` and reads `argv.mode`; `build:webpack:prod/dev` pass a single `--mode`, so the dev-builds-in-prod-mode bug is gone
- new devDependencies pinned to exact versions
- `CssMinimizerPlugin` added for production CSS
- `output.clean` now keeps the three synced static CSS files
- empty per-entry JS shims stripped via `RemoveEmptyJsPlugin`
- `strip-bom-loader` strips only a leading BOM
- dropping the `*.php` sync glob is a non-issue: `interface/themes/` has no top-level `.php` files

Security findings:
- symlink traversal — `lstatSync` + skip symlinks + `realpathSync` root-confinement
- Sass importer path traversal — `startsWith(baseDir)` guard
- production source maps disabled (`devtool: false` in prod)
- dotfiles / non-asset dirs excluded from the `public/assets` copy

Approving the code. Holding the merge until we complete manual testing of the rendered themes (LTR/RTL/compact variants + the three static CSS files) in a running environment.

kojiromike:
tested and works great here.

.gitignore
```diff
@@ -49,3 +49,4 @@ docker/development-easy-light/docker-compose.override.yml
 docker/development-easy-redis/docker-compose.override.yml
 compose.override.yml
 tmp/
+.webpack-cache/
```
bradymiller, line 52: hi @kojiromike , does this contain build assets that linger (jsut seeing if it needs to be walled off in the easy dev environment like other directories such as the phpstan cache did)

CLAUDE.md
```diff
@@ -251,9 +251,12 @@ npm run stylelint         # CSS/SCSS lint
...
 ```bash
-npm run build        # Production build
-npm run dev          # Development with file watching
-npm run gulp-build   # Build only (no watch)
+npm run build             # Production build (webpack + CSS sync)
+npm run build:webpack     # Webpack theme compilation (caller provides --mode)
+npm run build:webpack:prod  # Webpack production build
+npm run build:webpack:dev   # Webpack development build
+npm run build:sync        # Sync static CSS to public/themes/
+npm run dev               # Dev theme build, static CSS sync, then webpack watch
```
bradymiller, line 259: will need to check if these impact any devtools/openemr-cmd stuff

package.json
```diff
@@ -4,12 +4,14 @@
...
-    "gulp-build": "gulp -b --dev",
-    "gulp-watch": "gulp watch",
+    "postinstall": "napa && node scripts/install-assets.js",
+    "build": "npm run build:webpack:prod && npm run build:sync",
+    "build:webpack": "webpack --config webpack.themes.js",
+    "build:webpack:prod": "npm run build:webpack -- --mode production",
+    "build:webpack:dev": "npm run build:webpack -- --mode development",
+    "build:sync": "node scripts/sync-css.js",
+    "dev": "npm run build:webpack:dev && npm run build:sync && npm run watch",
+    "watch": "webpack --config webpack.themes.js --mode development --watch",
```
bradymiller, line 14: will need to check if these impact any devtools/openemr-cmd stuff

bradymiller:
Claude analysis of my above comments:

## Comment 1 — `.gitignore` `.webpack-cache/` wall-off — **valid, needs followup**

Same shape as the old phpstan-cache problem. Webpack writes its filesystem cache to `<repo-root>/.webpack-cache/` (`webpack.themes.js` line 115 — `cacheDirectory: path.resolve(__dirname, ".webpack-cache")`). In `docker/development-easy/docker-compose.yml` the repo is bind-mounted rw to `/var/www/localhost/htdocs/openemr`, so container builds will pollute the host repo with cache files — exactly what `phpstanvolume` was added to prevent.

The fix mirrors the phpstan pattern in the dev-easy compose file:
- under the openemr service volumes, add `- webpackcachevolume:/var/www/localhost/htdocs/openemr/.webpack-cache:rw`
- under the top-level `volumes:` block, add `webpackcachevolume: {}`

Also worth replicating in `development-easy-light/` and `development-easy-redis/` — they share the same bind-mount shape. (`development-insane/` is structured differently and out of scope.)

## Comments 2 & 3 — `CLAUDE.md` / `package.json` script changes vs devtools/openemr-cmd — **no breakage**

Walked the actual call chain:
- `openemr-cmd build-themes` -> `run_devtools_in_docker ... build-themes` -> `/root/devtools build-themes` inside the container -> `openemr-devops/docker/openemr/flex/utilities/devtools` runs `npm run build`.
- `npm run build` is preserved in the new `package.json` (it just now resolves to `build:webpack:prod && build:sync` internally instead of `gulp -b`). The contract holds.

Searched openemr-devops + openemr/docker/ + openemr/.github/workflows/ for the removed names (`gulp-build`, `gulp-watch`, `dev-build`, bare `gulp`, `gulpfile`) and got zero hits. The only npm scripts referenced externally are `npm run build`, `npm run test:js`, `npm run stylelint`, `npm run lint:js` and the lint-fix variants — all still present.

So these two comments can be resolved without code changes; only the cache wall-off needs a followup PR.

PR #11244, opened 2026-03-23
refactor(acl): throw exceptions instead of exit() in controller ACL checks

### LINKED ISSUE #11200: refactor(acl): throw exceptions instead of exit() in controller ACL checks
## Summary

`AccessDeniedHelper::denyWithTemplate()` calls `exit()`, which bypasses any `try/catch` error handling in the caller. In `controller.php`, we now have structured exception handling via `HttpExceptionInterface`, but ACL denials never reach it because they exit directly.

## Current behavior

- `Controller::checkControllerAcl()` calls `AccessDeniedHelper::denyWithTemplate()`, which exits
- Individual controllers (e.g., `C_PracticeSettings::__construct()`) also call `denyWithTemplate()` directly
- The `try/catch` in `controller.php` never sees these denials

## Desired behavior

- ACL failures in the controller routing path should throw `AccessDeniedHttpException` (Symfony, 403)
- `controller.php`'s existing `HttpExceptionInterface` catch handles the response
- Audit logging should still occur (before the throw or in a dedicated catch)

## Scope

- `Controller::checkControllerAcl()` — replace `denyWithTemplate()` with exception
- Individual controller constructors that call `denyWithTemplate()` — same treatment
- `AccessDeniedHelper` itself remains unchanged (still needed for legacy standalone scripts)

phantom930:
Fixes #11200 

# Summary
- Controller-level ACL denials were handled via a template helper that terminates execution, which bypasses normal exception-driven request handling.
- This change routes ACL denials through a shared controller method that logs/audits the denial and throws an AccessDeniedHttpException (HTTP 403), keeping behavior consistent with Symfony flow.

# Changes proposed in this pull request
- Added throwAccessDenied() to library/classes/Controller.class.php to centralize ACL-denial handling.
- Replaced direct AccessDeniedHelper::denyWithTemplate(...) calls with $this->throwAccessDenied(...) in:
    `controllers/C_Document.class.php`
    `controllers/C_PracticeSettings.class.php`
- Kept security observability intact by logging warnings and writing audit events before throwing.
- Removed now-unused AccessDeniedHelper imports in updated controllers.

### DIFF
--- a/src/Common/Acl/AccessDeniedHelper.php
+++ b/src/Common/Acl/AccessDeniedHelper.php
@@ -42,7 +42,8 @@ class AccessDeniedHelper
         $user = $session->get('authUser', 'unknown');
         $group = $session->get('authProvider', '');
 
-        ServiceContainer::getLogger()->warning("Access denied: $comment", [
+        ServiceContainer::getLogger()->warning("Access denied: {comment}", [
+            'comment' => $comment,
             'user' => $user,
         ]);
 


library/classes/Controller.class.php
```diff
@@ -117,13 +120,40 @@ private function checkControllerAcl(string $controllerName): void
...
+    protected function throwAccessDenied(
+        string $comment,
+        string $message,
+        string $auditEvent = 'security-access-denied'
+    ): never {
+        $session = SessionWrapperFactory::getInstance()->getActiveSession();
+        $user = $session->get('authUser', 'unknown');
+        $group = $session->get('authProvider', '');
+
+        ServiceContainer::getLogger()->warning("Access denied: $comment", [
```
Firehed: Minor: PSR-3 logging interpolation format would be better here:


```suggestion
        ServiceContainer::getLogger()->warning("Access denied: {comment}", [
        'comment' => $comment,
```

phantom930: @Firehed , I addressed this comment.

phantom930:
@Firehed , could you review it again and help me get it merged?
cc: @kojiromike

library/classes/Controller.class.php
```diff
@@ -117,13 +120,41 @@ private function checkControllerAcl(string $controllerName): void
...
 
+    /**
+     * Log/audit an ACL denial and throw an HTTP 403 exception for controller flow.
+     */
+    protected function throwAccessDenied(
+        string $comment,
+        string $message,
+        string $auditEvent = 'security-access-denied'
+    ): never {
+        $session = SessionWrapperFactory::getInstance()->getActiveSession();
```
kojiromike: The session lookup, logger warning, and audit event here duplicate `AccessDeniedHelper::deny()` (lines 48–62). Would it make sense to extract the shared log-and-audit logic into a static method on `AccessDeniedHelper` — something like `logDenial(string $comment, string $auditEvent)` — so both `deny()` and this method can call it?

That way the auditing stays in one place if the format ever changes (e.g., adding request context to the log entry).

controllers/C_PracticeSettings.class.php
```diff
@@ -17,7 +16,7 @@ function __construct(public $template_mod = "general")
         $this->direction = (OEGlobalsBag::getInstance()->get('_SESSION')['language_direction'] == 'rtl') ? 'right' : 'left';
 
         if (!AclMain::aclCheckCore('admin', 'practice')) {
```
kojiromike: Since `practice_settings` is already in `CONTROLLER_ACL_MAP`, `dispatch()` calls `checkControllerAcl()` before the constructor runs — so this check is redundant when reached through `controller.php`.

If someone instantiates `C_PracticeSettings` from a legacy script (no `HttpExceptionInterface` catch), this constructor throw would surface as a 500 instead of a 403. The old `denyWithTemplate()` was self-contained so it handled that safely.

Would it make sense to remove this constructor check entirely and rely on `CONTROLLER_ACL_MAP`? That eliminates the regression risk for this controller.

For the general case (preventing future controllers from throwing HTTP exceptions in constructors), I opened #11260.

library/classes/Controller.class.php
```diff
@@ -117,13 +118,25 @@ private function checkControllerAcl(string $controllerName): void
...
         if (!AclMain::aclCheckCore($section, $value)) {
-            AccessDeniedHelper::denyWithTemplate(
+            $this->throwAccessDenied(
                 "ACL check failed for $section/$value: $displayName",
                 xl($displayName)
             );
         }
     }
 
+    /**
```
kojiromike, line 128: Nit: `logDenial()` still uses PHP string interpolation for the logger message (`"Access denied: $comment"`). @Firehed's earlier suggestion to use PSR-3 placeholder format applies here — since the logging moved into `logDenial()`, that's where the fix should land:

```php
ServiceContainer::getLogger()->warning('Access denied: {comment}', [
    'comment' => $comment,
    'user' => $user,
]);
```

This lets log backends (Monolog, structured logging) handle interpolation correctly and avoids injecting arbitrary strings into the message template.

phantom930: Resolved this issue

phantom930:
@kojiromike I also removed the constructor check from `C_PracticeSettings`

PR #11245, opened 2026-03-23
fix(calendar): remove 0000-00-00 date defaults from postcalendar events

### LINKED ISSUE #11179: chore(db): Remove 0000-00-00 date defaults for MySQL strict mode compatibility

Firehed:
Fixes #11179 (partial - postcalendar events table only)

MySQL strict mode (`TRADITIONAL`, `NO_ZERO_DATE`) rejects `'0000-00-00'` as an invalid date value. This PR removes zero-date defaults from the `openemr_postcalendar_events` table and updates all code that reads/writes `pc_endDate` in that table to use `NULL` instead.

**Schema changes:**
- `sql/database.sql`: Change `pc_eventDate` and `pc_endDate` from `DEFAULT '0000-00-00'` to `DEFAULT NULL`
- `sql/8_1_0-to-8_1_1_upgrade.sql`: Add migration to update column defaults and convert existing zero-date values to NULL

**Code changes to handle NULL instead of '0000-00-00':**
- `portal/find_appt_popup_user.php`: Use `IS NULL` check
- `interface/main/calendar/find_appt_popup.php`: Use `IS NULL` check
- `interface/main/calendar/modules/PostCalendar/pnuserapi.php`: Use `IS NULL` checks
- `interface/main/calendar/add_edit_event.php`: Use null coalescing for pc_endDate checks and write NULL for non-recurring events
- `library/appointments.inc.php`: Write NULL instead of '0000-00-00'
- `library/MedEx/API.php`: Write NULL instead of '0000-00-00'
- `library/encounter_events.inc.php`: Handle null form_enddate in InsertEvent()
- `portal/add_edit_event_user.php`: Convert fixDate() zero-date returns to NULL before insert

**Note:** `src/Services/FHIR/FhirAppointmentService.php` already uses `Utilities::isDateEmpty()` which handles NULL correctly, so no changes were needed.

#### If you answered yes:
All changes in this PR were generated with assistance from Claude (Anthropic). Each commit represents a logical change to a specific file or component.


Firehed:
The scope of this should pull back slightly with #11320.

sql/8_0_0-to-8_1_0_upgrade.sql
```diff
@@ -260,3 +260,26 @@ UPDATE `document_templates` SET `send_date` = CURRENT_TIMESTAMP WHERE `send_date
 UPDATE `document_templates` SET `end_date` = NULL WHERE `end_date` = '0000-00-00 00:00:00';
 SET sql_mode = @currentSQLMode;
 #EndIf
+
```
Firehed: Since we already cut 8.1.0, this should either move into the 8.1.0->next or get ported into 8.1.0. Honestly I'd prefer the former but don't feel strongly.

interface/main/calendar/add_edit_event.php
```diff
@@ -903,7 +903,7 @@ function setEventDate($start_date, $recurrence)
         $repeattype = $rspecs['event_repeat_on_num'] < 5 ? 5 : 6;
     }
 
-    $recurrence_end_date = ($row['pc_endDate'] && !Utilities::isDateEmpty($row['pc_endDate'])) ? $row['pc_endDate'] : null;
+    $recurrence_end_date = $row['pc_endDate'] ?: null;
```
Firehed: is the key ever missing here? if not, remove the ternary

portal/add_edit_event_user.php
```diff
@@ -350,14 +350,25 @@
...
-                            $locationspec, $facility,
+                            $_POST['form_category'],
+                            $row['pc_multiple'],
+                            $to_be_inserted,
+                            $pid,
+                            $_POST['form_title'],
+                            $_POST['form_comments'],
+                            $session->get('providerId'),
+                            $event_date,
+                            (($d = fixDate($_POST['form_enddate'] ?? '')) !== '0000-00-00' ? $d : null),
```
Firehed: this looks weird to me

portal/add_edit_event_user.php
```diff
@@ -376,13 +387,24 @@
...
-                        $recurrspec, $starttime, $endtime, $_POST['form_allday'],
-                        $_POST['form_apptstatus'], $_POST['form_prefcat'],
-                        $facility, $provider, $row['pc_multiple'],
+                        $_POST['form_category'],
+                        $pid,
+                        $_POST['form_title'],
+                        $_POST['form_comments'],
+                        $session->get('providerId'),
+                        $event_date,
+                        (($d = fixDate($_POST['form_enddate'] ?? '')) !== '0000-00-00' ? $d : null),
```
Firehed: same

portal/add_edit_event_user.php
```diff
@@ -403,13 +425,24 @@
...
-                    $_POST['form_apptstatus'], $_POST['form_prefcat'] ?? '',
-                    $facility, $eid,
+                    $_POST['form_category'],
+                    $prov,
+                    $pid,
+                    $_POST['form_title'],
+                    $_POST['form_comments'],
+                    $session->get('providerId'),
+                    $event_date,
+                    (($d = fixDate($_POST['form_enddate'] ?? '')) !== '0000-00-00' ? $d : null),
```
Firehed: ^

portal/add_edit_event_user.php
```diff
@@ -446,14 +479,25 @@
...
-                        $locationspec, $facility,
+                        $_POST['form_category'],
+                        $new_multiple_value,
+                        $provider,
+                        $pid,
+                        $_POST['form_title'],
+                        $_POST['form_comments'],
+                        $session->get('providerId'),
+                        $event_date,
+                        (($d = fixDate($_POST['form_enddate'] ?? '')) !== '0000-00-00' ? $d : null),
```
Firehed: ^

portal/add_edit_event_user.php
```diff
@@ -469,14 +513,24 @@
...
-                    $_POST['form_apptstatus'], $_POST['form_prefcat'] ?? null,
-                    $locationspec, $facility,
+                    $_POST['form_category'],
+                    $form_provider_ae,
+                    $pid,
+                    $_POST['form_title'],
+                    $_POST['form_comments'],
+                    $session->get('providerId'),
+                    $event_date,
+                    (($d = fixDate($_POST['form_enddate'] ?? '')) !== '0000-00-00' ? $d : null),
```
Firehed: ^

interface/main/calendar/add_edit_event.php
```diff
@@ -904,7 +904,7 @@ function setEventDate($start_date, $recurrence)
         $repeattype = $rspecs['event_repeat_on_num'] < 5 ? 5 : 6;
     }
 
-    $recurrence_end_date = ($row['pc_endDate'] && !Utilities::isDateEmpty($row['pc_endDate'])) ? $row['pc_endDate'] : null;
+    $recurrence_end_date = fixDate($row['pc_endDate'], null);
```
kojiromike: This call adds a new entry to `.phpstan/baseline/argument.type.php` because `$row['pc_endDate']` is `mixed` and `fixDate` expects `string|null`. The repo `CLAUDE.md` is firm that new baseline entries should be fixed at the source rather than suppressed.

Suggestion — narrow first, then call:

```suggestion
    $pcEndDate = $row['pc_endDate'] ?? null;
    $recurrence_end_date = fixDate(is_string($pcEndDate) ? $pcEndDate : null, null);
```

Applying the same shape in `portal/add_edit_event_user.php` (extracting `$_POST['form_enddate']` into a typed local once at the top of the block) should also address the "this looks weird to me" self-review comments there — the `mixed|null` input is what makes the inline expression awkward.

library/encounter_events.inc.php
```diff
@@ -371,7 +371,7 @@ function InsertEvent($args, $from = 'general')
             ") VALUES (?,?,?,?,?,?,NOW(),?,?,?,?,?,?,?,?,?,?,?,?,?,1,1,?,?,?)",
             [$args['form_category'],($args['new_multiple_value'] ?? ''),$args['form_provider'],$form_pid,$form_gid,
             $args['form_title'],$args['form_comments'],$session->get('authUserID'),$args['event_date'],
-            fixDate($args['form_enddate']),$args['duration'],$pc_recurrtype,serialize($args['recurrspec']),
+            fixDate($args['form_enddate'], null),$args['duration'],$pc_recurrtype,serialize($args['recurrspec']),
```
kojiromike: Same issue as `add_edit_event.php:907` — this call (plus the matching one further down at line 396) produces the bump from `count => 1` to `count => 2` in `.phpstan/baseline/argument.type.php`. Narrowing `$args['form_enddate']` from `mixed` to `string|null` before the call (e.g., a local typed via `is_string($args['form_enddate'] ?? null) ? ... : null`) removes the baseline entry instead of widening it.

sql/database.sql
```diff
@@ -8272,8 +8272,8 @@ CREATE TABLE `openemr_postcalendar_events` (
   `pc_counter` mediumint(8) unsigned default '0',
   `pc_topic` int(3) NOT NULL default '1',
   `pc_informant` varchar(20) default NULL,
-  `pc_eventDate` date NOT NULL default '0000-00-00',
-  `pc_endDate` date NOT NULL default '0000-00-00',
+  `pc_eventDate` date DEFAULT NULL,
```
kojiromike: Intentional to drop `NOT NULL` on `pc_eventDate` as well? The `pc_endDate` change clearly makes sense (recurring events can legitimately have no end), but an event with no start date is semantically meaningless. The strict-mode goal would also be satisfied by keeping `NOT NULL` and just removing the zero-date default, which would fail loud if anything tried to insert an event without a date rather than silently landing a NULL. The corresponding `UPDATE ... SET pc_eventDate = NULL WHERE pc_eventDate = '0000-00-00'` in the upgrade SQL would then need to become either a DELETE or a repair step depending on what should happen to legacy placeholder rows.

kojiromike (commented):
some comments inline. Claude wrote, I reviewed.

sql/database.sql
kojiromike: to wit: what does it mean if `pc_eventDate` is null?

kojiromike:
@aisle-analyzer

kojiromike (approved):
anyway lgtm

PR #11251, opened 2026-03-23
fix(schema): remove 0000-00-00 values from document_templates inserts

Firehed:
Towards #11179

Fixes one spot that prevents database.sql from importing into the db while in strict mode.

Note: nothing in the code (that I or Claude found) references _either_ of these columns.

Changes the zero date value in `document_templates` INSERTS to `now()` and `null` for `send_date` and `end_date` respectively.

Updates existing data to follow this format.

Claude made the change, I sanity-checked.

### DIFF
--- a/sql/8_0_0-to-8_0_1_upgrade.sql
+++ b/sql/8_0_0-to-8_0_1_upgrade.sql
@@ -247,3 +247,16 @@ ALTER TABLE `medex_recalls` MODIFY `r_created` timestamp NOT NULL DEFAULT CURREN
 #IfIndex onetime_auth pid
 ALTER TABLE `onetime_auth` DROP INDEX `pid`, ADD INDEX `pid` (`pid`, `onetime_token`(32));
 #EndIf
+
+--
+-- Fix document_templates zero dates for MySQL strict mode compatibility.
+-- See: https://github.com/openemr/openemr/issues/11179
+--
+
+#IfTable document_templates
+SET @currentSQLMode = (SELECT @@sql_mode);
+SET sql_mode = '';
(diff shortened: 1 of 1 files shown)


sql/8_0_0-to-8_0_1_upgrade.sql
```diff
@@ -247,3 +247,16 @@ ALTER TABLE `medex_recalls` MODIFY `r_created` timestamp NOT NULL DEFAULT CURREN
...
 #EndIf
+
+--
+-- Fix document_templates zero dates for MySQL strict mode compatibility.
+-- See: https://github.com/openemr/openemr/issues/11179
+--
+
+#IfTable document_templates
+SET @currentSQLMode = (SELECT @@sql_mode);
+SET sql_mode = '';
```
Firehed, line 258: This is weird, but the pattern is in use elsewhere.

Firehed:
Verified on a mock db that this + #11245 are enough that you can `mysql < sql/database.sql` in `SQL_MODE=TRADITIONAL` and get to a fairly workable place, now that the other cleanups landed.

`TRADITIONAL` is shorthand for `STRICT_TRANS_TABLES,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,TRADITIONAL,NO_ENGINE_SUBSTITUTION`

sql/8_0_0-to-8_0_1_upgrade.sql
bradymiller: I think this was done if a user is upgrading on a strict database connector (ie. in theory would be case in future when users are using a strict database connector and needing to upgrade it from stuff that would break it).

bradymiller (approved):

Issue #10858, opened 2026-02-22
chore(deps): align oe-module-claimrev-connect composer.json with root


kojiromike:
The `composer.json` in `interface/modules/custom_modules/oe-module-claimrev-connect/` has dependency versions that are significantly out of date compared to the root `composer.json`. Since these module-level files aren't used directly during install (the root handles everything), they should reflect what's actually installed.

**Misaligned dependencies:**

| Dependency | Module | Root |
|---|---|---|
| `php` | `>=7.1` | `>=8.2.0` |
| `symfony/event-dispatcher` | `^4.4.0` | `~6.4.25` |
| `nyholm/psr7` | `^1.4` | `^1.8.2` |

Update all three constraints to match root.

Related: #7796

Issue #10747, opened 2026-02-12
ClaimRev: Refactor ClaimRevApi to reduce code duplication


kojiromike:
## Summary

Every API method in `ClaimRevApi.php` repeats the same ~15-line pattern for curl setup, resulting in significant code duplication.

## Location

`interface/modules/custom_modules/oe-module-claimrev-connect/src/ClaimRevApi.php`

## Repeated Pattern

Each method contains:

```php
$bootstrap = new Bootstrap(OEGlobalsBag::getInstance()->getKernel()->getEventDispatcher());
$globalsConfig = $bootstrap->getGlobalConfig();
$api_server = $globalsConfig->getApiServer();

$content = 'content-type: application/json';
$bearer = 'authorization: Bearer ' . $token;
$headers = [$content, $bearer];

// ... endpoint-specific code ...

$ch = curl_init();
curl_setopt($ch, CURLOPT_URL, $url);
curl_setopt($ch, CURLOPT_HTTPHEADER, $headers);
curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
// ... POST options if needed ...

$result = curl_exec($ch);
$httpcode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
curl_close($ch);
```

This pattern appears in 8 methods: `uploadClaimFile`, `getReportFiles`, `getDefaultAccount`, `searchClaims`, `searchDownloadableFiles`, `getFileForDownload`, `getEligibilityResult`, `uploadEligibility`.

## Problems

1. DRY violation - same code repeated 8+ times
2. Bug fixes must be applied to every method
3. Inconsistent error handling across methods
4. No central place to add logging, retry logic, or timeout configuration

## Suggested Approach

Extract helper methods:

```php
private static function getGlobalConfig(): GlobalConfig
{
    // Cache and return config
}

private static function makeRequest(
    string $method,
    string $endpoint,
    string $token,
    ?array $payload = null
): array {
    // Centralized curl setup, execution, and error handling
}
```

Then each public method becomes simpler:

```php
public static function searchClaims($claimSearch, $token)
{
    $response = self::makeRequest('POST', '/api/ClaimView/v1/SearchClaims', $token, $claimSearch);
    return $response['success'] ? $response['data'] : false;
}
```

This also provides a natural place to add:
- Curl error logging
- Timeout configuration
- Retry logic for transient failures

Issue #10746, opened 2026-02-12
ClaimRev: Code cleanup - naming, namespaces, and dead code


kojiromike:
## Summary

Several code quality issues in the ClaimRev module that should be cleaned up.

## 1. Method Case Mismatch

The method `getAccessToken()` is defined with camelCase but called as `GetAccessToken()` with PascalCase in multiple files:

- `ClaimRevApi.php:24`
- `ClaimUpload.php:59`
- `EligibilityTransfer.php:39`
- `ReportDownload.php:26`

While PHP methods are case-insensitive, this violates PSR-1 and makes code harder to search/grep.

## 2. Missing Namespace Declarations

Two service files define global functions instead of namespaced classes:

- `src/Billing_Claimrev_Service.php` - defines `start_X12_Claimrev_send_files()` and `start_X12_Claimrev_get_reports()`
- `src/Eligibility_ClaimRev_Service.php` - defines `start_send_eligibility()`

These should be converted to static class methods within the `OpenEMR\Modules\ClaimRevConnector` namespace.

## 3. Unused Imports

Multiple files have unused `use` statements:

| File | Unused Import |
|------|---------------|
| ClaimSearch.php | `OpenEMR\Common\Crypto\CryptoGen` |
| EraSearch.php | `OpenEMR\Common\Crypto\CryptoGen` |
| Billing_Claimrev_Service.php | `OpenEMR\Common\Crypto\CryptoGen` |
| Eligibility_ClaimRev_Service.php | `OpenEMR\Common\Crypto\CryptoGen` |
| ClaimRevApi.php | `OpenEMR\Common\Http\HttpRestRequest` |

## 4. Empty/Dead Methods in Bootstrap.php

- `renderMainBodyScripts()` (lines 219-221) - empty method
- `subscribeToApiEvents()` (lines 311-313) - empty method
- `addMetadataConformance()` (line 347) - references non-existent `CustomSkeletonFHIRResourceService`
- `addApiScope()` (lines 330-343) - references non-existent `CustomSkeletonResource`

These appear to be skeleton module leftovers that should be removed.

Issue #10745, opened 2026-02-12
ClaimRev: Malformed email links missing mailto: protocol


kojiromike:
## Summary

The support and sales email links are missing the `mailto:` protocol, causing them to be treated as relative URLs.

## Location

`interface/modules/custom_modules/oe-module-claimrev-connect/public/index.php` lines 68-70

## Current Code

```php
<li> <?php echo xlt("Email Support"); ?>: <a href = "support@claimrev.com">support@claimrev.com</a> </li>
<li> <?php echo xlt("Email Sales"); ?>: <a href = "sales@claimrev.com">sales@claimrev.com</a> </li>
```

## Problem

Without `mailto:`, clicking the link tries to navigate to a relative URL like `/interface/modules/.../public/support@claimrev.com`, resulting in a 404 error.

## Fix

```php
<li><?php echo xlt("Email Support"); ?>: <a href="mailto:support@claimrev.com">support@claimrev.com</a></li>
<li><?php echo xlt("Email Sales"); ?>: <a href="mailto:sales@claimrev.com">sales@claimrev.com</a></li>
```

PR #11265, opened 2026-03-24
feat(claimrev): install ClaimRev Connect module as Composer dependency

claimrevolution:
## Summary

Replaces the bundled `interface/modules/custom_modules/oe-module-claimrev-connect/`
directory with a Composer dependency on `claimrevolution/oe-module-claimrev-connect`
(published on Packagist). The module is installed into the same on-disk path at
`composer install` time by the existing `openemr/oe-module-installer-plugin`,
so OpenEMR's module installer continues to discover it via the filesystem scan
with no runtime behavior change.

This is a structural pivot from the original v2.1.0 submission in this PR, per
the architectural feedback from @kojiromike, @adunsulag, and @stephenwaite
(https://github.com/openemr/openemr/pull/11265#issuecomment-4509190794).

## What changed

- **`composer.json`** — adds `"claimrevolution/oe-module-claimrev-connect": "^2.1"`
- **`composer.lock`** — regenerated for that require
- **`interface/modules/custom_modules/oe-module-claimrev-connect/`** — removed (69 files); these files now arrive via Composer

## Why this is better than the original submission

- **Single binary covers OpenEMR 7.x + 8.x.** The module's own `src/Compat/`
  layer uses `class_alias` to provide `OEGlobalsBag`, `CryptoInterface`, and
  `ServiceContainer` only when the host core lacks them, so the same release
  works on OE master, the 8.0.x patch line, and the 7.x line.
- **Decoupled release cadence.** Module updates ship to Packagist without
  riding the OpenEMR release train. Customers can `composer update
  claimrevolution/oe-module-claimrev-connect` to pull patches independently.
- **Decoupled CI surface.** PHPStan, code coverage, and security review
  for the module move to the standalone repo's own CI (which I control).
  This PR's reviewable scope reduces from 17,517 LOC to 71 files / +103 / −5,987.

## Verified installs

Validated end-to-end installs against three OpenEMR images using
`composer require claimrevolution/oe-module-claimrev-connect:^2.1`:

| Image | OE version | Result |
|---|---|---|
| `openemr/openemr:flex` | master | ✓ installed, module enabled, pages render |
| `openemr/openemr:8.0.0.3-2026-03-25` | 8.0.x patched | ✓ installed (shims activate for missing APIs) |
| `openemr/openemr:7.0.2` | 7.0.2 | ✓ installed (full shim layer activates) |

## Module repo + Packagist links

- Source: https://github.com/claimrevolution/oe-module-claimrev-connect
- Packagist: https://packagist.org/packages/claimrevolution/oe-module-claimrev-connect
- Tagged release used by this PR: `v2.1.3`


kojiromike:
Hi, I don't know when I'll be able to take a deep look, but does this take these changes and fixes into account? Many of them were relatively recently fixed in the module.

#10736
#10737
#10738
#10739
#10740
#10741
#10742
#10743
#10744
#10745
#10746
#10747
#10858

claimrevolution:
I synced from master before making changes, these are all included.



Brad Sharp

CTO

Cell: 918-943-0020

Office: 918-842-9564

Bixby, OK

www.claimrev.com<http://www.claimrev.com/>

Schedule a Meeting<https://apicrm.claimrev.com/widget/booking/VCFC8qPP5w2Qo7K6Ni6o>

________________________________

interface/modules/custom_modules/oe-module-claimrev-connect/tests/test_compat_nuclear.php
```diff
@@ -0,0 +1,150 @@
+<?php
```
kojiromike: **Remove this file from the module tree.**

This script is reachable via `https://host/interface/modules/custom_modules/oe-module-claimrev-connect/tests/test_compat_nuclear.php`. There's no `.htaccess`, no `if (PHP_SAPI !== 'cli')` guard, and the `OPENEMR_ROOT` default (`/var/www/localhost/htdocs/openemr`) matches the official OpenEMR Docker image, so the env-var check is satisfied by default. The script then `rename()`s `src/Core/OEGlobalsBag.php`, `src/BC/ServiceContainer.php`, and `src/Common/Crypto/CryptoInterface.php`.

The `register_shutdown_function` restoration only runs if the request completes. An attacker can drop the TCP connection mid-execution — or PHP-FPM can hit `max_execution_time` — leaving the rename in place. Subsequent requests then either fatal because core classes are missing, or silently use the module's shim as the canonical class for the rest of the install.

References: OWASP A01:2021 (Broken Access Control), CWE-285, CWE-552.

Move integration tests outside the web-served tree (e.g., a CI-only path) and convert to PHPUnit with `spl_autoload_register` overrides instead of `rename()`. The repo's `composer phpunit-isolated` framework is the model — see `tests/Tests/Isolated/Common/Utils/ValidationUtilsIsolatedTest.php`.

interface/modules/custom_modules/oe-module-claimrev-connect/tests/test_compat_8x.php
```diff
@@ -0,0 +1,52 @@
+<?php
```
kojiromike: **Remove this file from the module tree.**

Same web-exposure issue as `test_compat_nuclear.php`. This file sets `$ignoreAuth = 1` and `$_GET['site'] = 'default'` before including `interface/globals.php`, so a request to `https://host/interface/modules/custom_modules/oe-module-claimrev-connect/tests/test_compat_8x.php` runs the bootstrap path with auth disabled. It also leaks absolute filesystem paths via `ReflectionClass::getFileName()` in the response (CWE-200).

Integration scripts must not live in directories the web server reaches. Move under a CI-only path, or convert to PHPUnit isolated tests under the repo's `tests/Tests/Isolated/` tree.

interface/modules/custom_modules/oe-module-claimrev-connect/tests/test_compat_shims.php
```diff
@@ -0,0 +1,134 @@
+<?php
```
kojiromike: **Remove this file from the module tree.**

Same web-exposure pattern as the other two `test_compat_*.php` files. Tests don't belong alongside runtime code in a web-served directory. The repo's PHPUnit isolated test framework (`composer phpunit-isolated`, see `tests/Tests/Isolated/`) is the standard place for this kind of compatibility check.

interface/modules/custom_modules/oe-module-claimrev-connect/src/Compat/compat.php
```diff
@@ -0,0 +1,43 @@
+<?php
```
kojiromike: **Remove the OE 7.x compatibility layer.**

This PR targets `openemr/openemr` master, which is the 8.x line. The shims here use `class_alias()` to bind module-internal classes to upstream FQCNs (`OpenEMR\Core\OEGlobalsBag`, `OpenEMR\BC\ServiceContainer`, `OpenEMR\Common\Crypto\CryptoInterface`). Two concrete problems with shipping this on master:

1. **Static analysis goes blind.** PHPStan resolves `OpenEMR\Core\OEGlobalsBag` to the real class via the autoloader — it never sees the `class_alias`'d shim. The shim has a strict subset of the real API (no `getProjectDir()`, no `getWebRoot()`). Any future divergence between the shim and the real class is invisible to CI; the first signal is a `Call to undefined method` fatal in production on a 7.x install.

2. **Permanent dead code on master.** On 8.x the real classes already exist, so `compat.php` is by design a no-op on the only branch it actually runs on. Every line here is maintenance surface for the upstream codebase, paid forever, for users who don't get this code via this branch.

The standard pattern for cross-version module support is to keep the module in its own repository and use Composer version constraints (`"openemr/openemr": ">=7.0,<9.0"`) — that puts the compatibility responsibility in the consuming environment, not in upstream master.

This change implies removing `src/Compat/`, `public/compat_check.php`, and the three `tests/test_compat_*.php` files.

interface/modules/custom_modules/oe-module-claimrev-connect/src/Bootstrap.php
```diff
@@ -88,14 +92,35 @@ public function __construct(
...
+        $this->registerMenuItems();
 
         // we only add the rest of our event listeners and configuration if we have been fully setup and configured
         if ($this->globalsConfig->isConfigured()) {
-            $this->registerMenuItems();
+            // Ensure the core SFTP global is enabled so claims flow through
+            // the x12_remote_tracker table where our background service picks them up.
+            // Without this, users must manually check "Automatically SFTP Claims To X12 Partner"
+            // in Globals for ClaimRev to work.
+            if ($this->globalsConfig->getGlobalSetting(GlobalConfig::CONFIG_AUTO_SEND_CLAIM_FILES)) {
```
kojiromike: **Move this block out of `subscribeToEvents()`.**

This runs on every authenticated request when `CONFIG_AUTO_SEND_CLAIM_FILES` is set. Three concrete issues:

1. **Bootstrap is the wrong lifecycle hook for persistent state.** Module install/upgrade is what `ClaimRevModuleSetup` exists for; `ModuleManagerListener::enable()` is the natural place. Per-request writes to `globals` are write amplification with no audit trail — the admin Globals UI normally produces an `audit_master_log` entry, and bypassing it here breaks that contract.

2. **The check at L112 (`$glValue === null || (string) $glValue === '' || (string) $glValue === '0'`) overwrites an explicit `'0'`.** If a site administrator has *deliberately* disabled `auto_sftp_claims_to_x12_partner`, this code re-enables it on the next page load. There's no way for the admin to keep it off as long as the module is enabled and configured.

3. **Concurrency.** The SELECT/UPDATE pair isn't transactional. Under a thundering-herd login burst, all workers see the empty/`'0'` value and fire UPDATEs. On multi-site shared MySQL deployments this contributes row-lock pressure on `globals` for every other tenant.

For the broader pattern of "give my module typed config without forcing it through `$GLOBALS`," `opencoreemr/oce-lib-module-config` provides a layered accessor (YAML files → env vars → `OEGlobalsBag` fallback) that other OCE modules use. Worth a look as a constructive alternative — config sourcing without runtime mutation.

interface/modules/custom_modules/oe-module-claimrev-connect/src/Bootstrap.php
```diff
@@ -20,10 +20,13 @@
  * Note the below use statements are importing classes from the OpenEMR core codebase
  */
 use OpenEMR\BC\ServiceContainer;
+use OpenEMR\Common\Database\QueryUtils;
```
kojiromike: **Add `declare(strict_types=1);` at the top of every new PHP file.**

The repo CLAUDE.md mandates `declare(strict_types=1)` on every new file. Currently only `src/ModuleInput.php`, `src/TypeCoerce.php`, and `src/Dto/ClaimSearchResult.php` have it — 84 of the 87 new PHP files in this PR are missing it.

Without strict types, PHP silently coerces `"123abc"` to `123` when passed to an `int` parameter, hiding the boundary parsing bugs that the module's own `ModuleInput` is otherwise trying to prevent. Apply across all new files. The existing strict-typed files in this PR are the model.

Flagging on this file because it's the most-loaded entry point; the same comment applies to the other 83.

interface/modules/custom_modules/oe-module-claimrev-connect/public/eligibility_chat.php
```diff
@@ -0,0 +1,52 @@
...
+use OpenEMR\Modules\ClaimRevConnector\ClaimRevException;
+use OpenEMR\Modules\ClaimRevConnector\ModuleInput;
+
+header('Content-Type: application/json');
+
+if (!AclMain::aclCheckCore('acct', 'bill')) {
+    http_response_code(403);
+    echo json_encode(['success' => false, 'message' => 'Unauthorized']);
+    exit;
+}
```
kojiromike: **Add CSRF verification.**

ACL alone doesn't stop CSRF (CWE-352). An attacker page can host a `<form action="https://emr/.../eligibility_chat.php" method=POST>` with attacker-chosen `question` content and auto-submit from a logged-in billing user's browser — triggering a paid AI call billed to the tenant, with prompt content the attacker controls.

The pattern is already established in this module (e.g., `claim_status_api.php` and `payment_advice_post.php`). Add a `CsrfHelper::verifyCsrfToken(ModuleInput::postString('csrf_token'), 'eligibility')` check after the ACL check, and emit the token from the parent template that posts here.

Same gap exists in (please add CSRF to all):

- `eligibility_check_now.php`
- `eligibility_sync_native.php`
- `appointment_check_now.php`
- `claim_oe_status.php`
- `claim_sync_status.php`

interface/modules/custom_modules/oe-module-claimrev-connect/public/claim_export_csv.php
```diff
@@ -0,0 +1,38 @@
...
+use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Modules\ClaimRevConnector\ClaimRevException;
+use OpenEMR\Modules\ClaimRevConnector\ClaimsPage;
+
+if (!AclMain::aclCheckCore('acct', 'bill')) {
+    http_response_code(403);
+    header('Content-Type: application/json');
+    echo json_encode(['success' => false, 'message' => 'Unauthorized']);
+    exit;
+}
```
kojiromike: **Add CSRF verification — and don't pass raw `$_POST` to the service.**

Two issues on this endpoint:

1. **Missing CSRF (CWE-352).** This is a bulk PHI export endpoint — exactly the kind of target where CSRF protection matters most. An attacker page can silently exfiltrate a tenant's claims to their own server by submitting from a logged-in billing user's browser. Add a `CsrfHelper::verifyCsrfToken(...)` check after the ACL check, mirroring `claim_status_api.php`.

2. **L27: `ClaimsPage::exportCsv($_POST)` defeats the `ModuleInput` discipline.** The module's own `payment_advice_post.php` shows the right pattern: parse the request to typed values at the controller boundary, then pass typed values into the service. Passing raw `$_POST` means `ClaimsPage::exportCsv` receives `mixed` for every key and has to re-validate downstream.

Also confirm row-level scoping is enforced inside `ClaimsPage::exportCsv` itself — bulk export must respect facility/PHI scope, not just the `acct/bill` global ACL (45 CFR §164.502).

interface/modules/custom_modules/oe-module-claimrev-connect/public/claim_mark_worked.php
```diff
@@ -0,0 +1,45 @@
...
+use OpenEMR\Modules\ClaimRevConnector\ClaimRevException;
+use OpenEMR\Modules\ClaimRevConnector\ModuleInput;
+
+header('Content-Type: application/json');
+
+if (!AclMain::aclCheckCore('acct', 'bill')) {
+    http_response_code(403);
+    echo json_encode(['success' => false, 'message' => 'Unauthorized']);
+    exit;
+}
```
kojiromike: **Add CSRF verification.**

This is a state-changing write (toggles claim worked-status via the ClaimRev API). ACL alone doesn't stop CSRF — an attacker page can flip claim states from a billing user's browser. Add the same `CsrfHelper::verifyCsrfToken(...)` check pattern that `claim_status_api.php` uses, after the ACL check.

interface/modules/custom_modules/oe-module-claimrev-connect/src/PaymentAdvicePostingService.php
```diff
@@ -0,0 +1,720 @@
+<?php
```
kojiromike: **Required: minimum test coverage before merge.**

This service is 720 lines that move money — posting ERA/835 payment data to OpenEMR encounters. Codecov shows 0% patch coverage (430 missing lines). At minimum, three behaviors need pinned-down tests:

1. **Idempotency.** Posting the same `payment_advice_id` twice must produce one `ar_session` row and N `ar_activity` rows, not duplicates. This is the most common 835 failure mode (operator double-clicks Post, retry after timeout, re-import of a re-sent file). HIPAA X12 835 implementation guides require receivers to handle duplicate transmissions.

2. **Decimal-cents integrity.** Sum of line-item amounts equals the claim total when posted, with no float drift. Use a `@dataProvider` of edge cents and assert as string compare on a `DECIMAL` cast, not a float compare. The repo's `tests/Tests/Isolated/Common/Utils/ValidationUtilsIsolatedTest.php` is the template — and per repo CLAUDE.md, every `@dataProvider` method must carry `@codeCoverageIgnore Data providers run before coverage instrumentation starts.` to avoid dragging patch coverage.

3. **Partial-write atomicity.** If row N of N fails (FK violation, deadlock, disk-full), no orphaned `ar_session` row is left. Pick a contract — full rollback or session-marked-failed — and assert it. The `PaymentAdviceMockService` already provides the fixture inventory; just wire it to a `PHPUnit\Framework\TestCase`.

`composer phpunit-isolated` runs on the host without Docker — these tests are cheap to add.

interface/modules/custom_modules/oe-module-claimrev-connect/src/ReconciliationService.php
```diff
@@ -0,0 +1,342 @@
+<?php
```
kojiromike: **Required: minimum test coverage before merge.**

Discrepancy detection has no truth oracle currently — there's no test that says what counts as a real mismatch and what doesn't. Two scenarios that need pinned-down tests:

1. **False negative (real discrepancy missed).** OE encounter `billed=$100, paid=$80, adj=$20, balance=$0`; ClaimRev reports `paid=$80, adj=$10, w/o=$10, balance=$0`. Balances tie but adjustment composition differs — a real reconciliation tool must flag this. A naive `oe.balance === cr.balance` check will miss it. Pin the right answer.

2. **False positive (phantom discrepancy reported).** OE row updated at `23:59:59` local, ClaimRev snapshot at `00:00:01 UTC` the next day. A timezone-naive `updated_at > snapshot_at` filter flags every late-evening encounter as out-of-sync forever — operators learn to ignore the report. CWE-1339 applies to time arithmetic too.

Without these tests pinned down, the tool may silently teach operators to trust the wrong side when OE and ClaimRev disagree.

interface/modules/custom_modules/oe-module-claimrev-connect/src/ClaimRevModuleSetup.php
```diff
@@ -77,6 +121,90 @@ public static function getServiceRecord($name)
...
         return null;
     }
+    /**
+     * Reset any ClaimRev background services that are stuck in running state.
+     * If running = 1 and next_run is more than 10 minutes in the past,
+     * the service is stuck (PHP crash, OOM kill, etc.) and needs to be freed.
+     */
+    public static function resetStuckServices()
+    {
+        $sql = "UPDATE background_services SET running = 0 WHERE running = 1 AND next_run < (NOW() - INTERVAL 10 MINUTE) AND name LIKE '%ClaimRev%'";
```
kojiromike: **Heuristic stuck-detection will false-positive on healthy long batches.**

This query marks any `ClaimRev_*` service "stuck" if `running = 1 AND next_run < NOW() - INTERVAL 10 MINUTE`. But a perfectly healthy `ClaimRev_Send` job processing a large end-of-month batch (5,000 claims at ~200ms each = ~16 minutes) exceeds that threshold mid-run. The watchdog flips `running = 0`; the next polling tick re-enters the same job; **two PHP workers are now concurrently iterating `x12_remote_tracker` and posting claims to ClaimRev**. That's the exact double-submission case 837 deduplication is supposed to prevent at the *clearinghouse*, not at the submitter.

Two concrete fixes:

1. **Exclude self from the reset query** — the `name LIKE '%ClaimRev%'` matches `ClaimRev_Watchdog` itself. If the watchdog ever runs > 10 min, it can clear its own `running` flag mid-execution. Add `AND name != 'ClaimRev_Watchdog'`.

2. **Replace the heuristic with a heartbeat lease.** Add a `last_heartbeat DATETIME` column to `background_services`; runners update it every ~30s; stuck = `running = 1 AND last_heartbeat < NOW() - INTERVAL 10 MINUTE`. This distinguishes liveness from progress (the same distinction Kubernetes liveness probes vs. readiness probes draw). A long-but-healthy run is then distinguishable from a crashed one.

interface/modules/custom_modules/oe-module-claimrev-connect/src/Eligibility_Sweep_Service.php
```diff
@@ -0,0 +1,109 @@
...
+    if ($enabled === '') {
+        return;
+    }
+
+    // Check if today is a configured sweep day
+    $sweepDaysConfig = TypeCoerce::asString($globals->get(GlobalConfig::CONFIG_SWEEP_DAYS) ?? '1,4');
+    if ($sweepDaysConfig === '') {
+        $sweepDaysConfig = '1,4';
+    }
+    $sweepDays = array_map(intval(...), array_filter(explode(',', $sweepDaysConfig)));
```
kojiromike: **`array_filter` without a callback silently drops `'0'` (Sunday).**

From the PHP manual on `array_filter`: *"If no callback is supplied, all empty entries of array will be removed. See `empty()` for how PHP defines empty in this case."* `'0'` is empty under that definition.

A site administrator who configures sweep days as `'0,3'` (Sunday + Wednesday) will silently get only Wednesday — Sunday is dropped before `array_map(intval(...), ...)` ever sees it.

```suggestion
    $sweepDays = array_map(intval(...), array_filter(explode(',', $sweepDaysConfig), fn($s) => $s !== ''));
```

Better still: parse to a typed `SweepSchedule` value object up front so future readers don't have to re-derive the semantics from raw integer comparisons against `date('w')`.

kojiromike (commented):
Thanks for the substantial update. This is a large surface — 88 files, ~16k LOC, ~17 distinct features bundled together. Before going line-by-line, two structural points and one packaging point.

## Packaging — would you consider splitting?

The PR has at least four naturally independent clusters:

1. Payment Advice + posting service (~2,620 LOC)
2. Claims + Reconciliation + Status Sync (~2,800 LOC)
3. Eligibility Sweep + Calendar + Notifications + Watchdog (~1,900 LOC)
4. Reporting dashboards — AR Aging, Denial Analytics, Recoupment, Patient Balance (~2,580 LOC)

Each cluster is internally cohesive but the four don't depend on each other at runtime. A 16k-LOC review can't get the same scrutiny that four 2k–3k PRs can in sequence. I realize splitting is real work; I'm raising it because the review velocity loss compounds the longer the PR stays open.

## Two findings that need to be addressed before merge

**A. The OE 7.x compatibility layer doesn't belong on master.** `openemr/openemr` master targets the 8.x line. Backwards-compat shims for a version that isn't here add permanent maintenance surface, and — more importantly — they make static analysis impossible: PHPStan resolves `OpenEMR\Core\OEGlobalsBag` and friends to the real classes and never sees the `class_alias`'d shims, so any drift between the shim's API and the real class is invisible to CI. Cross-version module support typically lives in the module's own repository via Composer version constraints, not as runtime shims grafted onto upstream namespaces. Inline comment on `src/Compat/compat.php`.

**B. Tests live alongside runtime code, in a web-served directory.** `tests/test_compat_nuclear.php` `rename()`s core OpenEMR PHP files; `tests/test_compat_8x.php` and `test_compat_shims.php` set `$ignoreAuth = 1`. None of them have any web-access guard. Whatever the test value, this should not ship under `interface/modules/custom_modules/.../tests/`. Move under a CI-only path or to PHPUnit isolated tests under the repo's `tests/Tests/Isolated/` tree. Inline comments on each.

## Other notable items

- **Menu / page ACL mismatch.** `Bootstrap::addCustomModuleMenuItem` sets `$menuItem->acl_req = []`, so the menu item is visible to any authenticated user. Every page under `public/` enforces `acct/bill`. Set `acl_req = ["acct", "bill"]` so the menu doesn't lead users to a 403. (Existing code, in unchanged context — flagging here since it can't anchor inline.)

- **Standards regression: `declare(strict_types=1)` is missing on 84 of 87 new PHP files.** Repo CLAUDE.md mandates it for every new file. Without strict types, PHP coerces `"123abc"` to `123` silently, defeating the boundary parsing the module's own `ModuleInput` is doing correctly. Inline on `Bootstrap.php`.

- **0% patch coverage on 7,077 lines of financial code.** Codecov shows zero coverage on `PaymentAdvicePostingService` (430 missing lines, money-moving), `ReconciliationService` (303 missing lines, discrepancy detection), and others. The repo exposes `composer phpunit-isolated` which runs without Docker; `tests/Tests/Isolated/Common/Utils/ValidationUtilsIsolatedTest.php` is the template. At minimum, the highest-risk behaviors in those two files (idempotency, decimal cents, partial-write rollback for posting; truth-oracle for reconciliation) need pinned-down tests before merge. Inline on each.

- **AJAX endpoints missing CSRF.** Most state-changing endpoints have both ACL and CSRF (good — `claim_status_api.php` and `payment_advice_post.php` are the model). Eight are missing CSRF: `eligibility_chat.php`, `eligibility_check_now.php`, `eligibility_sync_native.php`, `appointment_check_now.php`, `claim_oe_status.php`, `claim_export_csv.php`, `claim_mark_worked.php`, `claim_sync_status.php`. ACL alone doesn't stop CSRF (CWE-352). Grouped inline on `eligibility_chat.php`; specific notes on the highest-risk ones.

- **Bootstrap performs SQL writes on every authenticated request.** `Bootstrap::subscribeToEvents()` runs `UPDATE globals` and `UPDATE background_services` whenever `CONFIG_AUTO_SEND_CLAIM_FILES` is set and the stored value isn't `'1'`. Bootstrap should register listeners; persistence belongs in install/upgrade. Inline.

- **Watchdog stuck-detection heuristic.** `resetStuckServices()` marks any `ClaimRev_*` service "stuck" if `next_run < NOW() - 10min`. A healthy long batch (e.g., end-of-month claim post run) will exceed that and get reset → next tick re-enters → two workers concurrently posting claims. Also reflexive (matches the watchdog itself). Inline.

I appreciate the engineering effort here — the typed `ModuleInput`, the CSRF helper, the parameterized queries in many places, and the recently-added PSR-3 logger usage in the Eligibility Sweep service all show care. The findings above are about reducing surface where it's not pulling its weight (compat shims, web-accessible tests) and pinning down the financial-correctness invariants before they ship.

claimrevolution:
I'm making more progress here, I pulled out the shims to a 7 branch and worked on what you have suggested so far, still working on all the PHP Stan stuff, hope to be done with it... eventually

claimrevolution:
what else can I do to get this merged in? I can't split it up in smaller PR's everything is tested together and I know it all works.

kojiromike:
@aisle-analyzer

kojiromike:
@claimrevolution I just ran `composer install && composer phpstan-reset-baseline` and it removed a bunch of stuff. Let me know if you get different results.

interface/modules/custom_modules/oe-module-claimrev-connect/public/payment_advice.php
```diff
@@ -0,0 +1,1199 @@
+<?php
```
bradymiller: ongoing review placeholder

kojiromike:
@claimrevolution I think we have a way to make this much easier for you. The gist is, make a separate repo with the claimrev module in it and make that depend on https://packagist.org/packages/openemr/oe-module-installer-plugin and publish that module to packagist. Then create a PR against this repo to remove the old claimrev module along with updating composer.lock and composer.json to add your new `oe-module-claimrev` (or whatever you decide to call it).

This approach will
1. Enable you to have a package that can be independently installed into any version of openemr, even old ones.
2. Allow you to set your own standards for testing and static analysis. The bar stops being 'make all the checks in the PR green' and becomes, 'make sure the module in composer.json works'. How you achieve that is your choice.
3. Isolate claimrev and openemr from each other's security profiles: A vulnerability in claimrev becomes a "dependency vulnerability" in openemr (so addressable far more simply than a current GHSA) and a vulnerability in openemr otherwise needn't delay a new release of the claimrev module.

I discussed this with @adunsulag and @stephenwaite and we agreed this seems like a good way forward. Please let me know your thoughts.

claimrevolution:
> 

How will this put me on the list of modules to install? without being in the code base claimrev wouldn't be listed.

kojiromike:
> How will this put me on the list of modules to install? without being in the code base claimrev wouldn't be listed.

It won't be in the repo, but it will be in the codebase for any runnable OpenEMR instance. That's what oe-module-installer-plugin does.

claimrevolution:
Per the suggestion from @kojiromike, @adunsulag, and @stephenwaite earlier in this thread (https://github.com/openemr/openemr/pull/11265#issuecomment-4509190794), I've force-pushed `claimrev-v2.1.0` with a complete rewrite of the PR scope.

The module is now published on Packagist as `claimrevolution/oe-module-claimrev-connect` (https://packagist.org/packages/claimrevolution/oe-module-claimrev-connect) and this PR collapses to a 3-file change against upstream:

- `composer.json` — single require line
- `composer.lock` — resolved entry
- `interface/modules/custom_modules/oe-module-claimrev-connect/` — removed (the older module ships from this tree; new releases come from the standalone repo)

The 17k-LOC module code, the OE 7.x compatibility shims, PHPStan / Codecov / security review — all of that now lives in the standalone repo's own CI: https://github.com/claimrevolution/oe-module-claimrev-connect

Title and description have been updated to reflect the new scope. Happy to take a closer look at the composer.lock churn or talk through the install-plugin flow if useful.

claimrevolution:
can we finish this up now?

kojiromike:
When CI runs `composer install` oe-module-installer-plugin does its job and copies oe-module-claimrev-connect back into custom_modules. Then the checks run, and fail. I'll add appropriate ignores to the path. Eventually, we need to get the autoloader so that modules can run out of `vendor`. But this is something of a front-controller problem.

kojiromike (approved):

Issue #11136, opened 2026-03-16
feat(crypto): Add support for libsodium-based internals


Firehed:
This offers a number of improvements over the current openssl-based encryption:

- It's bundled with all supported versions of PHP; no need to install or check for an extension
- The API is easier to get right
- Secure defaults
- Ignoring support for older data versions, the complete implementation is only a couple lines of code

Firehed:
I've done some additional research here under the context of how this may affect compliance and ONC certification. From what I was able to gather (and anyone that knows better, please offer a correction), the certification requirement is basically "follow [FIPS 140-2 Annex A](https://csrc.nist.gov/pubs/fips/140-2/upd2/final)". As of writing, that summarizes to "use AES".

There's only one AES tool within Sodium (`sodium_crypto_aead_aes256gcm_*`), and it's not always available; evidently it depends on the hardware on which it's run. It's a fairly high success rate, but not a guarantee. The "simpler" `sodium_crypto_secretbox*` and most of the `[X]ChaCha20-*` appear to be always available, but NOT currently in the NIST/FIPS approved cipher list.

I think it still makes sense to add support for one or two of the sodium family - probably aes256 and xchacha20-poly1305. If we add an ONC certification module (hi @kojiromike) it would have to check that either encryption is a) enabled with one of the AES paths or b) disabled but the underlying storage engines are appropriately encrypted at rest. The latter path is actually better for most deployments, since you can outsource all of the certification to the hosting vendor (easy) and there's less to go wrong within OpenEMR.

bradymiller:
Most simpleton installs would need ONC supported encryption out of box. Disabling encryption while having documents in web directory would be a bad combination.
Do we have an idea of how many environments would not support aes256?
What does going to libsodium really buy us in terms of better security?
And is this worth above issues?

Note this criteria may disappear if we wait a bit. Per discussion with @adunsulag , these criteria may be going away with ONC in future. Dropping of this criteria would also open the door to an easier route of supporting encryption streaming via libsodium (rather than needing to have entire file in memory).

Firehed:
That's great feedback and context, thanks @bradymiller  :)

I agree that the out-of-box experience should continue to have documents encrypted regardless of whether there's full-disk encryption. We should aim to migrate documents outside of the web directory (which would happen with the front controller) but that's a separate discussion, but they should absolutely remain encrypted until that happens.

> Do we have an idea of how many environments would not support aes256?

[...]

PR #11267, opened 2026-03-24
refactor(encryption): Updated key management

Firehed:
Fixes #

Not materially besides unit tests, though it's been a loooong series of conversations with AI tools.

To do:
- [x] Split out Message and MessageFormat; add any suitable tests and information
- [x] Split Keychain and its immediate dependencies to own PR, with additional tests
- [x] Split out the KeyStorage components to own PR. ADODB probably not unit-testable, but the rest should be (within reason). Note: they're not heavily used but ought to make a future-facing LazyKeychain with "normal" key ids possible
- [ ] Improve the "latest key" strategy (I've got a path here that almost works, but it's a bit goofy in the BC path)
- [x] Remove debugging changes
- [x] Further improve error handling
- [x] Split out the key generation?
- [ ] Fix N+1 key loading

Future scope:
- [ ] Message format v8 that specifies a key id (#11305)
- [ ] Sodium ciphers (#11136)
- [ ] Additional key management and rotation tooling (e.g. env var to control preferred key in DI setup)


ci/ciLibrary.source
```diff
@@ -334,7 +334,7 @@ build_test_e2e() {
     _exec php "${php_args[@]}" ./vendor/bin/phpunit \
         --log-junit junit-e2e.xml \
         --testsuite e2e \
-        --testdox \
```
Firehed: Debugging, will be reverted (but we probably do actually want this change)

sites/default/config.php
```diff
@@ -20,7 +20,7 @@
     //Enscript command used by Hylafax.
     $GLOBALS['more_secure']['hylafax_enscript'] = 'enscript -M Letter -B -e^ --margins=36:36:36:36';
     foreach ($GLOBALS['more_secure'] as $key => $value) {
-        $GLOBALS['more_secure'][$key] = $cryptoGen->encryptStandard(is_string($value) ? $value : null);
```
Firehed: Debugging, need encryption paths first

src/Encryption/Storage/PlaintextKeyInDbKeysTable.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace OpenEMR\Encryption\Keys\Storage;
+
+use Doctrine\DBAL\Connection;
+use OpenEMR\Encryption\Keys\KeyMaterial;
+
+readonly class PlaintextKeyInDbKeysTable implements KeyStorageInterface
+{
+    public function __construct(
+        private Connection $conn,
```
Firehed: This is dormant for now, but intended to support the autowiring. Needs more work on the connection management, since this needs the no-log/no-middleware connection to avoid recursion

src/Encryption/Keys/Storage/PlaintextKeyInDbKeysTableAdodb.php
```diff
@@ -0,0 +1,36 @@
...
+    public function getKey(string $identifier): KeyMaterial
+    {
+        $row = QueryUtils::querySingleRow(
+            'SELECT value FROM `keys` WHERE name = ?',
+            // 'value',
+            [$identifier],
+            log: false,
+        );
+        if ($row === false) {
+            throw new \Exception('No key found');
```
Firehed: All these exceptions need cleanup. Or maybe make the interface `tryGetKey(string $identifier): ?KeyMaterial`

src/Encryption/Keys/Keychain.php
```diff
@@ -0,0 +1,29 @@
...
+        CipherInterface $cipher,
+    ): void {
+        $this->mappings[$id] = $cipher;
+    }
+
+    // addLoader to defer key loading?
+
+    public function getCipher(string $keyId): CipherInterface
+    {
+        return $this->mappings[$keyId];
```
Firehed: Needs missing key handling (will just crash from no offset/wrong type)

src/Encryption/BCCrypto.php
```diff
@@ -0,0 +1,78 @@
...
+
+namespace OpenEMR\Encryption;
+
+use OpenEMR\Common\Crypto\{
+    CryptoInterface,
+    KeySource,
+};
+use Psr\Log\LoggerInterface;
+
+class BCCrypto implements CryptoInterface
```
Firehed: This probably should delegate to a "cleaner" class so that this just does the key id remapping. TBD. Easy enough to do that later.

High-level goal here is there's some sort of `CipherSuiteInterface` with a simple `encrypt(Plaintext): Message` and decrypt(Message): Plaintext` and code would iteratively move to that interface instead of CryptoInterface.

Firehed (commented):
Leaving notes for myself and anyone looking at this while it's in draft.

Also generally want a seprate key generation path and some sort of import type thing. Test bootstrapping and OpenEMR setup would use the former instead of the on-the-fly thing today (the BC layer probably needs it though)

src/Encryption/Keys/BCKeychain.php
```diff
@@ -0,0 +1,123 @@
...
+        self::tryLoadDbKey('six', $pkidb, $storageDir, $keychain);
+        self::tryLoadDbKey('seven', $pkidb, $storageDir, $keychain);
+
+        if ($createKeyIfNeeded !== null) {
+            if (!$keychain->hasKey($createKeyIfNeeded)) {
+                // Generate and store keys
+                // TODO: actually persist them!!
+                $key = KeyMaterial::generate(openssl_cipher_key_length('aes-256-cbc'));
+                $hmacKey = KeyMaterial::generate(32);
+                // FIXME: persist this data!
```
Firehed: Calling yet even more attention to this: it should be enough to smoke-test the application so a key _exists in memory_, but would be catastrophically bad in reality.

src/Encryption/MessageFormat.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenEMR\Encryption;
+
+enum MessageFormat: int
```
Firehed: I'm aware this is nearly identical to `KeyVersion` in `Crypto`. I intentionally did not re-use it in order to help separate data versions and key identities, which KeyVersion muddles.

PR #11292, opened 2026-03-25
chore(ci): Run PHPStan on 8.5

Firehed:
Fixes https://github.com/openemr/openemr-devops/issues/611

PHPStan should run on the highest supported released version of PHP so its aware of forward capabilities.

It will still analyze against 8.2: https://phpstan.org/config-reference#phpversion


kojiromike:
I'd just merge this once phpstan passes, without waiting for the other checks.

PR #11294, opened 2026-03-26
feat(encryption): Add encrypt methods to new ciphers

Firehed:
Extracted from #11267 

Adds `encrypt` methods to the new `CipherInterface` as part of a larger cryptography overhaul. Adds tests to ensure both backwards compatibility and overall correctness.

- Adds method to the interface
- Adds methods on the three implementing classes
  - Only the most modern cipher actually has a working implemented
  - Older ciphers throw

The failure on older cyphers more or less matches CryptoGen's behavior, where it always uses the latest key version/format.

Tests were from AI.

### DIFF
--- a/tests/Tests/Isolated/Encryption/Cipher/Aes256CbcHmacSha384Test.php
+++ b/tests/Tests/Isolated/Encryption/Cipher/Aes256CbcHmacSha384Test.php
@@ -204,12 +204,13 @@ final class Aes256CbcHmacSha384Test extends TestCase
         );
         $cipher2 = new Aes256CbcHmacSha384(
             key: new KeyMaterial('different_key___________________'), // 32 bytes
-            hmacKey: new KeyMaterial($this->fixtures->getTestKey('sevenb')),
+            hmacKey: new KeyMaterial('different_hmac__________________'), // 32 bytes
         );
 
         $ciphertext = $cipher1->encrypt(new Plaintext(CryptoFixtureManager::PLAINTEXT));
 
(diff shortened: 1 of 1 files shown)


Firehed:
Further note on last commit:

```
❯ ok give me a brief recap of what was happening there

⏺ The test encrypted with cipher1, then decrypted with cipher2 using a different encryption key but the same HMAC key.

  Since HMAC covers the ciphertext (not the plaintext), and the ciphertext wasn't modified, HMAC validation passed. The test then relied on openssl_decrypt returning false due to invalid PKCS7 padding when decrypting garbage with the wrong key.

  The flake: ~0.4% of the time, random garbage happens to end with valid-looking PKCS7 padding (e.g., last byte is 0x01). When that occurs, openssl_decrypt returns the garbage instead of false, no exception is thrown, and the test fails.

  The fix: Use different HMAC keys too. Now HMAC validation fails deterministically before we even get to the padding check.

❯ does this mean there's a risk in 1/256 keys?

⏺ No security risk. The ~1/256 case only happens when:

  1. HMAC key is correct (HMAC passes)
  2. Encryption key is wrong
  3. Garbage output happens to have valid padding

  In that case you get garbage back instead of an exception - but it's still garbage, not the real plaintext. An attacker can't read anything.

  In practice, if someone has the wrong encryption key, they almost certainly have the wrong HMAC key too. HMAC fails first, exception thrown, done.

  The flaky test was an artificial scenario (same HMAC key, different encryption key) that doesn't occur in real usage.
```

PR #11302, opened 2026-03-26
refactor(encryption): Message parser and format tooling

Firehed:
Extracted from #11267

Adds new infrastructure for the encryption overhaul that improves type safety for several string-like primitives. This paves the way to separate the message format from the key identifiers, which have historically been coupled. 

A future addition will define _format_ `8`, which will likely encode into something like `008:keyid-base64:ciphertext`. There may be a change to the parsing logic that will pass through unencrypted data, but both of these are out of scope for now.

- Key ID DTO class
- Message Format enum
- Message DTO class with parser

Tests got a light AI assist

### DIFF
--- a/src/Encryption/Keys/Id.php
+++ b/src/Encryption/Keys/Id.php
@@ -4,7 +4,7 @@
  * @package   OpenEMR
  * @link      https://www.open-emr.org
  * @author    Eric Stern <erics@opencoreemr.com>
- * @copyright Copyright (c) 2026 OpenCoreEMR
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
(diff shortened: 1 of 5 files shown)


src/Encryption/Keys/Id.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/**
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Eric Stern <erics@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR
```
kojiromike:
```suggestion
 * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
```

PR #11305, opened 2026-03-26
refactor(encryption): Simplify the modern key format handling, prepare for explicit key ids

Firehed:
Prepares the new encryption paths to support proper key rotation and decoupling the format from the key id.

- Removes MessageFormat enum cases v1-v7 in favor of `ImplicitKey`
- Adds a `UnusedCaseToSupportConditionals` so it's possible to set up `match()` constructs without PHPStan complaining
- Leaves a note about adding case `ExplicitKey` to `MessageFormat`, also maybe `Plaintext`
- Update parsing, detection, etc to match new logic

To simplify how data detection works, the `ExplicitKey` still will use the prefix-based approach, rather than something like JWE or another slightly-more-specified format. To detect data using a non-preferred key outside of the application runtime, you'd search for data that _doesn't_ begin with `008:yourcurrentkeyid:`.

Format considerations:
- Easily detectable
- Hard to badly screw up
- Not subject to algorithm swaps

I ruled out JWE because the official spec _requires_ `alg`; the only safe way to handle the JWT family of data formats including JWE is to ignore it in favor of the spec-_optional_ `kid`. Easy enough, but it's still otherwise a rather awkward fit

No, other than as a rubber-duck


Firehed:
Ignoring the gaps in test coverage; see #11321. The other areas are for placeholder code that's intended to be unreachable, but due to syntax isn't respecting the coverage ignore markers.

PR #11310, opened 2026-03-27
chore: prep for 8.1.0 and 8.1.1

bradymiller:
chore: change upgrade script to 8.1.0 (this will also go into rel-810)
chore: add upgrade script for 8.1.1


bradymiller:
@kojiromike , forgot about this when branch. need to add a new sql upgrade script for the new version. (also since we changed versions, needed to update the prior upgrade script, which will also bring to rel-810)

PR #11332, opened 2026-03-29
fix(users): allow clearing user profile fields

### LINKED ISSUE #11317: Unable to clear/delete user field
### Describe the bug

As an admin, I am unable to delete information in a user field, e.g. if I accidentally information to the Valedictory or Fed Tax ID field, I am only able to modify, but not completely clear the field.

### To Reproduce

https://github.com/user-attachments/assets/b37b279a-79b2-4dd1-8fcf-c303bf1b84ff

1. Admin > Users > select a user
2. Add a string into a user field, e.g. Valedictory or Federal Tax ID and save.
3. Open the user again and delete the string.
4. See how the string is not deleted. If I modify the field and as long as there is a value in the field, then the modification is saved.

### Expected behavior
1. I would expect that I would be able to delete values in editable fields for the user.

### Client configuration

- Browser:
- OpenEMR version: 8.0.0 (3)
- Operating system:

haoyu-haoyu:
Fixes #11317

This fixes the update path for user profile fields in `usergroup_admin.php`. Before this change, clearing fields like Valedictory or Federal Tax ID could silently keep the old value because the update logic treated empty strings as false and skipped the write.

The current version keeps the consolidated update flow, but only runs it when one of the string fields actually changed. Submitted empty strings are treated as intentional clears, while untouched fields keep their original database value exactly as stored, including `NULL`. That preserves the clearing behavior without rewriting unrelated nullable columns.

`facility_id` and `billing_facility_id` are still handled separately because they have additional facility lookup and sync logic.

### DIFF
--- a/interface/usergroup/usergroup_admin.php
+++ b/interface/usergroup/usergroup_admin.php
@@ -108,47 +108,47 @@ if (isset($_POST["privatemode"]) && $_POST["privatemode"] == "user_admin") {
             sqlStatement("update `groups` set user=? where user= ?", [trim((string) $_POST["username"]), $user_data["username"]]);
         }
 
-        if ($_POST["taxid"]) {
+        if (isset($_POST["taxid"])) {
             sqlStatement("update users set federaltaxid=? where id= ? ", [$_POST["taxid"], $_POST["id"]]);
         }
 
-        if ($_POST["state_license_number"]) {
+        if (isset($_POST["state_license_number"])) {
             sqlStatement("update users set state_license_number=? where id= ? ", [$_POST["state_license_number"], $_POST["id"]]);
         }
 
-        if ($_POST["drugid"]) {
+        if (isset($_POST["drugid"])) {
             sqlStatement("update users set federaldrugid=? where id= ? ", [$_POST["drugid"], $_POST["id"]]);
         }
 
-        if ($_POST["upin"]) {
+        if (isset($_POST["upin"])) {
             sqlStatement("update users set upin=? where id= ? ", [$_POST["upin"], $_POST["id"]]);
         }
 
-        if ($_POST["npi"]) {
+        if (isset($_POST["npi"])) {
             sqlStatement("update users set npi=? where id= ? ", [$_POST["npi"], $_POST["id"]]);
         }
 
-        if ($_POST["taxonomy"]) {
+        if (isset($_POST["taxonomy"])) {
             sqlStatement("update users set taxonomy = ? where id= ? ", [$_POST["taxonomy"], $_POST["id"]]);
         }
 
-        if ($_POST["lname"]) {
+        if (isset($_POST["lname"])) {
             sqlStatement("update users set lname=? where id= ? ", [$_POST["lname"], $_POST["id"]]);
         }
 
-        if ($_POST["suffix"]) {
+        if (isset($_POST["suffix"])) {
             sqlStatement("update users set suffix=? where id= ? ", [$_POST["suffix"], $_POST["id"]]);
         }
 
-        if ($_POST["valedictory"]) {
+        if (isset($_POST["valedictory"])) {
             sqlStatement("update users set valedictory=? where id= ? ", [$_POST["valedictory"], $_POST["id"]]);
         }
 
-        if ($_POST["job"]) {
+        if (isset($_POST["job"])) {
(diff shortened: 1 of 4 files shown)


interface/usergroup/usergroup_admin.php
```diff
@@ -108,47 +108,47 @@
             sqlStatement("update `groups` set user=? where user= ?", [trim((string) $_POST["username"]), $user_data["username"]]);
         }
 
-        if ($_POST["taxid"]) {
+        if (isset($_POST["taxid"])) {
             sqlStatement("update users set federaltaxid=? where id= ? ", [$_POST["taxid"], $_POST["id"]]);
         }
 
-        if ($_POST["state_license_number"]) {
+        if (isset($_POST["state_license_number"])) {
```
kojiromike: these if statements are very repetitive. This could be another thing to refactor easily, if not in this PR, then in another.

haoyu-haoyu:
@kojiromike Thank you for the feedback! The repetitive if-statements have been consolidated into a single fixed-shape UPDATE:

- A `$stringFields` map defines the POST key → DB column mapping
- Only changed values are detected (compared against `$user_data`)
- A single `UPDATE users SET ... WHERE id = ?` with all 12 fields is issued
- If nothing changed, no UPDATE is executed (no unnecessary audit-log noise)

**Re: Semgrep SQL injection finding** — This is a false positive. The SQL string is built from concatenated string *literals* (not variables):
```php
"UPDATE users SET federaltaxid = ?, state_license_number = ?, ..."
```
All column names are hardcoded; all values use `?` placeholders. No user input enters the SQL string.

Rebased onto latest master. Ready for re-review.

interface/usergroup/usergroup_admin.php
```diff
@@ -108,48 +108,74 @@
...
+                    $changedFields[$dbColumn] = $_POST[$postKey];
+                }
+            }
         }
 
-        if ($_POST["valedictory"]) {
-            sqlStatement("update users set valedictory=? where id= ? ", [$_POST["valedictory"], $_POST["id"]]);
-        }
+        if ($changedFields !== []) {
+            $preserveUserValue = static function (string $dbColumn) use ($changedFields, $user_data) {
```
kojiromike, line 146: The two-phase design (detect changes in a loop → reconstruct values via closure) adds indirection that can be eliminated. Since the UPDATE always writes all 12 columns regardless of which changed, you could build the params array in the same loop that detects changes:

```php
$params = [];
$changed = false;
foreach ($stringFields as $postKey => $dbColumn) {
    $current = is_array($user_data) && array_key_exists($dbColumn, $user_data)
        ? $user_data[$dbColumn]
        : null;
    if (isset($_POST[$postKey])) {
        $submitted = (string) $_POST[$postKey];
        // Treat NULL and '' as equivalent so NULLs aren't silently converted
        if (($current ?? '') !== $submitted) {
            $changed = true;
        }
        $params[$dbColumn] = $submitted;
    } else {
        $params[$dbColumn] = $current;
    }
}

if ($changed) {
    sqlStatement(
        "UPDATE users SET"
        . " federaltaxid = ?, state_license_number = ?,"
        . " federaldrugid = ?, upin = ?, npi = ?, taxonomy = ?,"
        . " lname = ?, fname = ?, suffix = ?, valedictory = ?,"
        . " specialty = ?, mname = ?"
        . " WHERE id = ?",
        [...array_values($params), $_POST['id']]
    );
}
```

This resolves the NULL handling and raw-`$_POST` issues in one pass, and removes the need for the `$changedFields` array and `$preserveUserValue` closure entirely.

kojiromike (commented):
The consolidation into a `$stringFields` map + single UPDATE is a good improvement over the per-field if/sqlStatement pattern. A few things need attention before merge — copilot raised some valid points that haven't been addressed yet:

1. **NULL→'' silent conversion** — When a column is NULL in the DB and the form submits `''` (which HTML text inputs always do for empty fields), the comparison `null !== ''` flags it as "changed" and overwrites NULL with `''`. This will silently convert every nullable column to `''` across the user table as admins make routine edits, contradicting the PR's stated goal of preserving NULLs.

2. **Raw `$_POST[$postKey]` in `$changedFields`** — The cast to `(string)` was done deliberately for `$submittedValue`, but `$changedFields` stores the raw mixed `$_POST` value instead. This also drives the baseline increase in `cast.string.php`.

I have one additional inline comment on simplifying the overall approach.

Issue #11330, opened 2026-03-28
feat(background-services): Add REST API endpoints for listing and triggering services


kojiromike:
**Parent:** #11325 (step 5 of 5)
**Depends on:** Step 2 (BackgroundServiceRunner), Step 4 (BackgroundServiceRegistry)

## Problem

There is no API to list or trigger background services. In containerized deployments, triggering a background job requires shell access to a running OpenEMR container. This makes it impossible to use lightweight Kubernetes CronJobs — you'd need a full OpenEMR container with PHP and the entire codebase just to invoke a CLI command.

A thin HTTP client (e.g., a k8s CronJob container with `curl`) using OAuth2 client credentials would be far simpler and more robust.

## Proposed Change

Add routes to `apis/routes/_rest_routes_standard.inc.php`, gated by `admin/super` ACL:

| Method | Endpoint | Description |
|--------|----------|-------------|
| `GET` | `/api/admin/background-services` | List all registered services |
| `GET` | `/api/admin/background-services/:name` | Get a single service's details |
| `POST` | `/api/admin/background-services/:name/run` | Trigger execution |

### `GET /api/admin/background-services`

Returns an array of service objects with: `name`, `title`, `active`, `execute_interval`, `running`, `last_run`, `next_run`.

Uses `BackgroundServiceRegistry::list()`.

### `GET /api/admin/background-services/:name`

Returns a single service object, or 404.

Uses `BackgroundServiceRegistry::get()`.

### `POST /api/admin/background-services/:name/run`

Request body (optional):
```json
{"force": true}
```

Response:
```json
{"status": "started", "service": "phimail"}
```

Possible `status` values: `started`, `already_running`, `not_active`, `not_found`, `not_due`.

Uses `BackgroundServiceRunner::run()`. The existing database-level locking (UPDATE + `affectedRows()`) prevents concurrent execution, so it's safe for multiple callers (API, cron, Ajax) to coexist.

### Authentication

Uses the existing OAuth2 flow. A service account with `admin/super` ACL and client credentials grant type can call these endpoints with a Bearer token.

**Example k8s CronJob usage:**
```bash
# Get token
TOKEN=$(curl -s -X POST https://openemr.example.com/oauth2/default/token \
  -d "grant_type=client_credentials&client_id=$ID&client_secret=$SECRET" \
  | jq -r .access_token)

# Trigger service
curl -X POST https://openemr.example.com/apis/default/api/admin/background-services/phimail/run \
[...]

Issue #11328, opened 2026-03-28
feat(background-services): Add CLI tooling for listing, running, and crontab generation


kojiromike:
**Parent:** #11325 (step 3 of 5)
**Depends on:** Step 2 (BackgroundServiceRunner extraction)

## Problem

A sysadmin who wants to run background services via cron must:
1. Query the `background_services` table directly to discover service names
2. Know the undocumented CLI calling convention: `php library/ajax/execute_background_services.php default phimail 0`
3. Manually translate `execute_interval` values into cron expressions

There is no self-documenting CLI interface.

## Proposed Change

Add a CLI entry point at `bin/background-services` with three subcommands:

### `list` — Show registered services

```bash
bin/background-services list [--site=default]

# NAME               TITLE                              ACTIVE  INTERVAL  LAST RUN
# phimail            phiMail Direct Messaging Service    yes     5 min     2026-03-28 10:15:00
# Email_Service      Email Service                       yes     2 min     2026-03-28 10:18:00
# UUID_Service       UUID Creation Service               yes     240 min   2026-03-28 06:00:00
# X12_SFTP           SFTP Claims to X12 Partner          no      1 min     never
# MedEx              MedEx Messaging Service             no      manual    never
```

### `run` — Execute a specific service

```bash
bin/background-services run <name> [--site=default] [--force]
```

Delegates to `BackgroundServiceRunner::run()`. Exit code reflects outcome (0 = ran, 1 = error, 2 = skipped/not due).

### `crontab` — Generate crontab entries

```bash
bin/background-services crontab [--site=default] [--php=/usr/bin/php]

# # OpenEMR Background Services (site: default)
# */5 * * * *   /usr/bin/php /path/to/bin/background-services run phimail --site=default
# */2 * * * *   /usr/bin/php /path/to/bin/background-services run Email_Service --site=default
# 0 */4 * * *   /usr/bin/php /path/to/bin/background-services run UUID_Service --site=default
```

Translates `execute_interval` (minutes) into reasonable cron expressions. Output is a starting point — sysadmins adjust as needed.

## Backward Compatibility

The old CLI interface (`php library/ajax/execute_background_services.php default phimail 0`) continues to work unchanged.

## Files

- `bin/background-services` — new CLI entry point
- Uses `BackgroundServiceRunner` from step 2
- Pattern reference: `library/allow_cronjobs.php` (existing CLI bootstrap helper)

PR #11334, opened 2026-03-29
refactor(background-services): extract orchestration into BackgroundServiceRunner

### LINKED ISSUE #11327: refactor(background-services): Extract orchestration logic into BackgroundServiceRunner
**Parent:** #11325 (step 2 of 5)

## Problem

All background service orchestration logic lives in two procedural functions inside `library/ajax/execute_background_services.php`:

- `execute_background_service_calls()` — reads the `background_services` table, acquires locks, includes files, calls functions, releases locks
- `background_shutdown()` — cleanup on abnormal exit

This script also handles CLI argument parsing, CSRF verification, and globals bootstrapping — mixing entry-point concerns with business logic. The orchestration cannot be reused from the planned CLI tool or REST API without duplicating code or shelling out.

## Proposed Change

Extract the orchestration into a service class:

```php
namespace OpenEMR\Services\Background;

class BackgroundServiceRunner
{
    /**
     * Run one or all background services.
     *
     * @param string|null $serviceName  Specific service name, or null for all
     * @param bool $force  Bypass interval check
     * @return array  Results keyed by service name (ran, skipped, not_found, already_running, etc.)
     */
    public function run(?string $serviceName = null, bool $force = false): array;
}
```

The existing `execute_background_services.php` becomes a thin wrapper that:
- Detects CLI vs Ajax
- Bootstraps globals / validates CSRF
- Delegates to `BackgroundServiceRunner::run()`

The locking mechanism (UPDATE + `affectedRows()`), interval checking, `require_once` + function dispatch, and shutdown cleanup all move into the service class.

## Why This Comes Before CLI and API

Both the CLI tool (step 3) and the REST API (step 5) need to call the same orchestration logic. Extracting it first avoids duplication and keeps the later PRs focused.

## Files

- `library/ajax/execute_background_services.php` — refactor to thin wrapper
- `src/Services/Background/BackgroundServiceRunner.php` — new class (extracted logic)

kojiromike:
## Summary

- Extracts service discovery, locking, execution, and shutdown logic from `library/ajax/execute_background_services.php` into `src/Services/Background/BackgroundServiceRunner.php`
- The existing script becomes a thin wrapper (CLI/Ajax detection, globals bootstrap, CSRF) using `filter_input()` for typed parameter access
- Uses `BackgroundServicesRow` type stub from `TableTypes` for fully typed DB rows at PHPStan level 10
- Uses `QueryUtils` methods instead of deprecated global SQL functions
- Replaces `global $service_name` shutdown pattern with instance property + closure
- Adds `background_services` to the CI `TableTypes.php` generation check
- Net reduction of PHPStan baseline entries
- Prerequisite for CLI tooling (#11328) and REST API (#11330)

## Test plan

- [x] `composer phpunit-isolated` — 7 new `BackgroundServiceRunnerTest` tests pass
- [x] `composer phpstan` — no errors
- [x] Pre-commit hooks pass (phpcs, phpstan, rector)
- [ ] Docker: login, verify background services still fire via Ajax polling

Closes #11327

### DIFF
--- a/library/ajax/execute_background_services.php
+++ b/library/ajax/execute_background_services.php
@@ -48,10 +48,10 @@
 
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Session\SessionWrapperFactory;
-use OpenEMR\Core\OEGlobalsBag;
+use OpenEMR\Services\Background\BackgroundServiceRunner;
 
 //ajax param should be set by calling ajax scripts
-$isAjaxCall = isset($_POST['ajax']);
+$isAjaxCall = filter_has_var(INPUT_POST, 'ajax');
 
 //if false ajax and this is a called from command line, this is a cron job and set up accordingly
 if (!$isAjaxCall && (php_sapi_name() === 'cli')) {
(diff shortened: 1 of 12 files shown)


library/ajax/execute_background_services.php
```diff
@@ -87,95 +87,9 @@
...
-        $res = sqlStatementNoLog($sql, [$service_name]);
-    }
-}
+$runner = new BackgroundServiceRunner();
+$runner->registerShutdownHandler();
 
-register_shutdown_function(background_shutdown(...));
-execute_background_service_calls();
-unset($service_name);
+$serviceName = $_REQUEST['background_service'] ?? null;
```
kojiromike: Can you get this from filter_input or another means to narrow and avoid the mixed type problem on 95?

library/ajax/execute_background_services.php
```diff
@@ -87,95 +87,9 @@
...
+$force = !empty($_REQUEST['background_force']);
```
kojiromike: don't use empty(). Do use filter_input

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -0,0 +1,148 @@
...
+
+        return QueryUtils::fetchRecordsNoLog($sql . ' ORDER BY sort_order', []);
+    }
+
+    /**
+     * @param array<string, mixed> $service
+     */
+    protected function acquireLock(array $service, bool $force): bool
+    {
+        $interval = intval($service['execute_interval'] ?? 0);
```
kojiromike: is phpstan only complaining about this being mixed because of the test, where the `getServices` mock doesn't have typing?

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -0,0 +1,148 @@
...
+    {
+        $results = [];
+        $services = $this->getServices($serviceName, $force);
+
+        if ($serviceName !== null && $services === []) {
+            return [['name' => $serviceName, 'status' => 'not_found']];
+        }
+
+        foreach ($services as $service) {
+            $name = (string) $service['name'];
```
kojiromike: Why is a cast necessary? the return type of getServices looks right...

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -0,0 +1,148 @@
...
+            $this->currentServiceName = $name;
+
+            if (empty($service['active']) || ($service['running'] ?? 0) == 1) {
```
kojiromike: don't use empty()

.phpstan/baseline/return.type.php
```diff
@@ -9081,6 +9081,11 @@
     'count' => 1,
     'path' => __DIR__ . '/../../src/Services/AppointmentService.php',
 ];
+$ignoreErrors[] = [
+    'message' => '#^Method OpenEMR\\\\Services\\\\Background\\\\BackgroundServiceRunner\\:\\:getServices\\(\\) should return list\\<array\\<string, mixed\\>\\> but returns list\\<array\\<mixed\\>\\>\\.$#',
```
kojiromike: This is leading to several additional phpstan errors downstream. Lock it down.

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -0,0 +1,181 @@
...
+            try {
+                $locked = $this->acquireLock($service, $force);
+            } catch (\Throwable) {
+                $results[] = ['name' => $name, 'status' => 'error'];
+                continue;
+            }
+
+            if (!$locked) {
+                $results[] = ['name' => $name, 'status' => 'locked'];
+                continue;
```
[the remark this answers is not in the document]
kojiromike: Out of scope for this extraction — the original code also conflated these cases. Can revisit if the CLI/API callers need finer granularity.

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * @param BackgroundServicesRow $service
+     */
+    protected function executeService(array $service): void
+    {
+        $requireOnce = $service['require_once'];
+        if ($requireOnce !== null && $requireOnce !== '') {
+            require_once(OEGlobalsBag::getInstance()->getString('fileroot') . $requireOnce);
```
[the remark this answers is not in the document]
kojiromike: Good suggestion, but out of scope for this extraction — the original code had the same unvalidated `require_once`. Path validation for background service includes would be a worthwhile follow-up.

Issue #11254, opened 2026-03-24
refactor: replace direct superglobal access with filter_input() across codebase


kojiromike:
The `ForbiddenRequestGlobalsRule` PHPStan rule flags direct access to `$_GET`, `$_POST`, `$_REQUEST`, `$_FILES`, `$_COOKIE`, and `$_SERVER`. The baseline currently has **8,068 errors** across **1,031 file entries**.

### Sub-issues

- #11255 — Rector rule to automate ~85% of conversions

### Baseline breakdown by superglobal

| Superglobal | Errors | Replacement |
|-------------|--------|-------------|
| `$_POST` | 5,006 | `filter_input(INPUT_POST, ...)` |
| `$_GET` | 1,565 | `filter_input(INPUT_GET, ...)` |
| `$_REQUEST` | 1,105 | Manual — no `INPUT_REQUEST` constant |
| `$_SERVER` | 259 | `filter_input(INPUT_SERVER, ...)` |
| `$_FILES` | 116 | Manual — no filter_input equivalent |
| `$_COOKIE` | 17 | `filter_input(INPUT_COOKIE, ...)` |

### Pattern analysis

~85% of errors are mechanically convertible via Rector rule (#11255):

| Pattern | Count | Rector? |
|---------|------:|---------|
| Simple key access `$_GET['key']` | 3,585 | Yes |
| Null coalesce `$_GET['key'] ?? x` | 1,288 | Yes |
| isset/!isset checks | 846 | Yes |
| empty/!empty checks | 1,142 | Yes (phase 2) |
| **Convertible subtotal** | **~6,860** | |
| Writes `$_POST['k'] = v` | 660 | No |
| `$_REQUEST` | 924 | No |
| `$_FILES` | 86 | No |
| Variable/nested keys | 101 | No |
| **Manual subtotal** | **~1,770** | |

### Breakdown by directory

| Directory | Errors | Notes |
|-----------|--------|-------|
| `interface/` | 6,134 | Legacy web UI — largest concentration |
| `library/` | 643 | Legacy utilities |
| `portal/` | 565 | Patient portal |
| `gacl/` | 240 | ACL admin (isolated) |
| `controllers/` | 157 | Legacy MVC controllers |
| `src/` | 137 | Modern PSR-4 code |
| Other | 192 | sphere, custom, contrib, etc. |

### Top 10 files

| Errors | File |
|--------|------|
| 396 | `interface/forms/eye_mag/save.php` |
[...]

PR #11348, opened 2026-03-30
refactor(php): route superglobal access through HttpRestRequest in src/

haoyu-haoyu:
## Summary

Stop accessing `$_GET`/`$_POST`/`$_SERVER` directly in **8 files** under `src/` (ClinicalDecisionRules + Patient Cards + OEEnvBag), routing reads through `HttpRestRequest::createFromGlobals()` / Symfony's `Request` instead.

### Why

Direct superglobal access is flagged by the project's `ForbiddenRequestGlobalsRule` PHPStan rule (#11254). Using the framework's typed accessors lets PHPStan see the right return types and removes the corresponding baseline entries.

### Scope

| Area | Files | Changes |
|------|-------|---------|
| ClinicalDecisionRules | 3 | `Common::get()/post()` go through a cached `Symfony\Request::createFromGlobals()` (with `Common::resetRequestCache()` for tests). `ControllerAjax` reads `plan_id` via `Common::get()`. `ControllerAlerts` validates that `id[]`/`access_control[]` align before delegating to `CdrAlertManager::update()`. |
| Patient Cards | 4 | `(string) $_POST['x']` → `$request->request->getString('x')`; integer ids → `getInt()`. The `team` array in `CareTeamViewCard` still goes through `->all()` because that field really is array-typed. |
| Core | 1 | `OEEnvBag::createInstance()` drops the redundant `is_array(getenv())` guard — `getenv()` with no arguments always returns an array. |

### Deliberately excluded

- **`$_SERVER['REDIRECT_URL']` in FHIR controllers** — `ForbiddenRequestGlobalsRule` does not currently flag `$_SERVER`, and the touched files are out of scope here.
- **`$_POST` writes** and **bulk `$_POST` pass** — out of scope; would need a different abstraction.
- **Files outside `src/`** — separate PRs per #11254.

### Notes

- `Common::get()/post()` cache the `Request` so the ~50 helper calls per CDR controller request stop rebuilding the parameter bags. Tests that mutate `$_GET`/`$_POST` between cases call `Common::resetRequestCache()` in `setUp()`.
- `CommonTest::testPost()` still asserts the array-return contract; the implementation goes through `->request->all()[$key]` which preserves array values that `InputBag::get()` would coerce/reject in Symfony 6.3+.

## Test plan

- [x] `CommonTest` (get/post/postString) passes via cached request + reset hook
- [x] `ControllerEditTest` passes
- [ ] Verify ClinicalDecisionRules plan CRUD (plan_id via `Common::get`)
- [ ] Verify Patient Cards save/update
- [ ] Verify Alert Manager submit with sparse checkboxes

Ref #11254


.phpstan/baseline/offsetAccess.notFound.php
Firehed, line 399: Make sure you've got the latest dependencies installed, I think your PHPStan may beout of date. I had this come up on a recent branch as well.

src/ClinicalDecisionRules/Interface/Common.php
```diff
@@ -51,8 +51,14 @@ public static function implode_funcs($glue, array $pieces, array $funcs): string
      */
     public static function get($var, $default = ''): string
     {
-        $val = $_GET[$var] ?? null;
-        return isset($val) && $val !== '' ? $val : $default;
+        // filter_input reads the original SAPI input, not the $_GET array.
+        // Fall back to $_GET for CLI/test environments where filter_input
+        // cannot see values set at runtime.
+        $val = filter_input(INPUT_GET, $var) ?? $_GET[$var] ?? null;
```
Firehed: This is a behavioral change which could have unintended consequences: `filter_input()` returns `false` if the variable isn't set, so the null-coalesce behavior is slightly different than before.

Same for `post`, and possibly other places.

Firehed (changes requested):
I like the direction this is headed! Leaving a couple preliminary notes.

haoyu-haoyu:
Thanks for the review @Firehed — both points addressed:

### 1. `filter_input()` false vs null handling

You're right — `filter_input()` returns `false` on failure (not just `null`), and `false ?? 'default'` evaluates to `false`, not `'default'`. Fixed across all files:

**`Common::get()` / `Common::post()`** — Replaced the `??` chain with an explicit `is_string()` guard:
```php
$val = filter_input(INPUT_GET, $var);
if (!is_string($val)) {
    $val = $_GET[$var] ?? null;  // fallback for CLI/test or filter failure
}
```
This correctly handles both `null` (not set) and `false` (failure), falling through to `$_GET`/`$_POST` in either case.

**ViewCard files** — Changed all `filter_input(...) ?? ''` to `?: ''` so that `false` also triggers the fallback to the default value (11 occurrences across 4 files).

### 2. PHPStan baseline

Regenerated via `composer phpstan-baseline` instead of hand-editing. The baseline diff is now smaller and correct (5 files, -110 lines vs the previous 10 files with incorrect entries).

haoyu-haoyu:
@Firehed Both issues have been addressed in the follow-up commit:

1. **`filter_input()` false/null semantics**: Replaced `??` with explicit `is_string()` type checks. `filter_input()` returns `null` (not set) or `false` (failure) — neither is a usable string. Falls back to `$_GET`/`$_POST` for CLI/test environments or when `filter_input` returns `false`.

2. **PHPStan baseline**: Regenerated with latest dependencies.

CI is green. Ready for re-review when you have a moment.

kojiromike:
Hi @haoyu-haoyu — this one is 15 days old, has a merge conflict, CI failures, and changes requested by @Firehed about the behavioral difference between `filter_input()` and direct superglobal access. There are also stale baseline entries and remaining `$_POST`/`$_GET` usage. Please rebase and address this feedback before opening new PRs. If you need help, feel free to ping me with questions.

haoyu-haoyu:
Thanks @Firehed and @kojiromike! Pushed an update addressing all the feedback:

**Rebase onto master** — force-push, 3 commits replayed onto current master; baseline conflicts resolved by taking upstream (CI will regenerate).

**Remove superglobal fallbacks:**
- `src/ClinicalDecisionRules/Interface/Common.php` — `Common::get()` and `Common::post()` now go through `HttpRestRequest::createFromGlobals()->query` / `->request` instead of falling back to `$_GET` / `$_POST`. Addresses both @Firehed's "behavioral change on filter_input" concern and Copilot's `ForbiddenRequestGlobalsRule` concern — the Symfony InputBag handles nested array values, missing keys, and CLI/test contexts cleanly.
- `src/Patient/Cards/CareTeamViewCard.php` — `handleFormSubmission()` reads every form field via the request's `request` InputBag, including the team array via `->all('team')`. `$_POST['save_care_team']` fallback gone.
- `src/Patient/Cards/InsuranceViewCard.php` — `getEligibilityOutput()` reads `status_update` from the request and no longer mutates `$_POST`.

**Dead config:**
- Deleted the top-level `phpstan-baseline.neon`. The project wires baselines through `.phpstan/baseline/loader.php`; the newly-added neon file wasn't referenced anywhere.

**Stale baseline entries** for `TreatmentPreferenceViewCard.php` / `CareExperiencePreferenceViewCard.php` — CI's baseline regeneration will drop these automatically now that the underlying code no longer triggers the errors.

Ready for another look.

src/Patient/Cards/CareExperiencePreferenceViewCard.php
```diff
@@ -188,18 +189,23 @@ public function getTextColorClass(): string
...
     private function handlePost(): void
     {
-        if (($_SERVER['REQUEST_METHOD'] ?? 'GET') !== 'POST') {
+        $request = HttpRestRequest::createFromGlobals();
+
+        if ($request->getMethod() !== 'POST') {
             return;
         }
-        if (($_POST['pref_type'] ?? '') !== 'care_experience') {
+        if ((string) $request->request->get('pref_type', '') !== 'care_experience') {
```
kojiromike:
```suggestion
        if ($request->request->getString('pref_type') !== 'care_experience') {
```

src/Patient/Cards/CareExperiencePreferenceViewCard.php
```diff
@@ -188,18 +189,23 @@ public function getTextColorClass(): string
...
             return;
         }
         CsrfUtils::checkCsrfInput(INPUT_POST, key: 'csrf_token', dieOnFail: true);
 
-        $action = $_POST['action'] ?? '';
+        $action = (string) $request->request->get('action', '');
```
kojiromike:
```suggestion
        $action = $request->request->getString('action');
```

The rule here is always, "narrow, don't cast". We get some value from $_REQUEST, or from an untyped database field. IRL it's probably usually a `string`, but phpstan can't know that, so it decides it's `mixed`. Then we try to use it in a typed context, and phpstan complains. The next thing we do is cast it, but you can't cast from mixed to another type anyway, so that just makes phpstan complain a different way.

But narrowing works. It lets phpstan know that we're properly guarding the type. If it can't be used as a string, it's probably broken anyway.

src/Patient/Cards/TreatmentPreferenceViewCard.php
```diff
@@ -188,18 +189,23 @@ public function getTextColorClass(): string
...
     private function handlePost(): void
     {
-        if (($_SERVER['REQUEST_METHOD'] ?? 'GET') !== 'POST') {
+        $request = HttpRestRequest::createFromGlobals();
+
+        if ($request->getMethod() !== 'POST') {
             return;
         }
-        if (($_POST['pref_type'] ?? '') !== 'treatment_intervention') {
+        if ((string) $request->request->get('pref_type', '') !== 'treatment_intervention') {
```
kojiromike:
```suggestion
        if ($request->request->getString('pref_type') !== 'treatment_intervention') {
```

src/Patient/Cards/TreatmentPreferenceViewCard.php
```diff
@@ -188,18 +189,23 @@ public function getTextColorClass(): string
...
             return;
         }
         CsrfUtils::checkCsrfInput(INPUT_POST, key: 'csrf_token', dieOnFail: true);
 
-        $action = $_POST['action'] ?? '';
+        $action = (string) $request->request->get('action', '');
```
kojiromike:
```suggestion
        $action = $request->request->getString('action');
```

src/Patient/Cards/TreatmentPreferenceViewCard.php
```diff
@@ -188,18 +189,23 @@ public function getTextColorClass(): string
...
         if ($action === 'save') {
```
kojiromike: this should be a switch/case or match expression

haoyu-haoyu:
Pushed 9c1bfc1ea addressing the latest round of feedback:

**@kojiromike — narrow, don't cast** ✅
- `CareExperiencePreferenceViewCard.php` and `TreatmentPreferenceViewCard.php` now use `$request->request->getString('pref_type')` and `->getString('action')` (and `getInt('id')` on the delete branch) instead of `(string)`/`filter_var(...) ?: 0` casts. The save branch keeps the `getString('id')` + `filter_var(..., FILTER_NULL_ON_FAILURE)` flow because `update()` distinguishes a real id from `null`.

**Copilot items** ✅
1. **`InsuranceViewCard` — only one key needed** — went straight to `->request->getString('status_update')`; no more `->all()`.
2. **`OEEnvBag` — `is_array(getenv())` always true** — dropped the redundant guard. `getenv()` with no arguments always returns `array<string, string>`; only the named-overload can return `false`. The corresponding baseline ignore is no longer needed (CI will regenerate).
3. **PR description / implementation mismatch** — both bullets covered by item 2 (no more is_array baseline) and by the description rewrite below.
4. **`Common::get()/post()` — cache the Request** — added a private `?Request` field with a `request()` accessor; the ~50 calls per CDR controller request now share one constructed Request. Also added `Common::resetRequestCache()` so `CommonTest::setUp()` can drop the snapshot between cases that mutate `$_GET`/`$_POST`.
5. **`ControllerAlerts` array length validation** — the active/passive/reminder checkbox arrays are sparse by design (only checked boxes get submitted), so they can't carry a length requirement. The fix `array_values()`-reindexes `id[]` and `access_control[]` (the only two dense per-row inputs) and aborts with HTTP 400 when their lengths don't match. `CdrAlertManager::update()` then iterates safely by zero-based offset.

CI regenerates the baselines, so the existing slices should drop the now-obsolete entries (`OEEnvBag` is_array, `Common` get/post return-type, etc.).

PR description updated to reflect the actual approach (HttpRestRequest + Symfony InputBag with narrow accessors, not filter_input fallback). Ready for another look.

haoyu-haoyu:
Pushed `ff37b3fe8f`:

**PHPStan CI failure** — `ControllerEditTest.php:33` anonymous class had untyped params. Dropped the `_get`/`_post` overrides (they were dead — `ControllerEdit`/`BaseController` don't declare those methods; the real code calls `Common::get`/`Common::post`). Typed the ctor with `RuleManager`/`CodeManager`.

**@kojiromike — switch/case** ✅ `TreatmentPreferenceViewCard` and `CareExperiencePreferenceViewCard` now use `switch ($request->request->getString('action'))` with `case 'save':`/`case 'delete':` instead of the if/elseif chain.

Ready for another look when you have a moment.

kojiromike:
@aisle-analyzer

kojiromike (approved):
Re-reviewed. The five points I raised on 2026-04-14 are all addressed:

- `getString('pref_type')` / `getString('action')` in both `CareExperiencePreferenceViewCard` and `TreatmentPreferenceViewCard` (narrow, don't cast)
- `switch` on `$action` in both files

Resolved my outdated threads. LGTM from my side once CI is green.

src/Patient/Cards/CareExperiencePreferenceViewCard.php
```diff
@@ -188,31 +189,38 @@ public function getTextColorClass(): string
...
+                } else {
+                    $this->service->insert($data);
+                    $this->flashMessage = xl('Preference saved');
+                }
+                break;
+            case 'delete':
+                $id = $request->request->getInt('id');
+                if ($id) {
+                    $this->service->delete($id);
+                    $this->flashMessage = xl('Preference deleted');
```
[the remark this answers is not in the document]
kojiromike: Tracked as #11853 (covers both preference cards). The IDOR predates this PR — the refactor here only changes how `id` is read, not the missing patient-ownership check. Out of scope for the superglobal cleanup; will be fixed in a follow-up.

src/Patient/Cards/TreatmentPreferenceViewCard.php
```diff
@@ -188,31 +189,38 @@ public function getTextColorClass(): string
...
+                } else {
+                    $this->service->insert($data);
+                    $this->flashMessage = xl('Preference saved');
+                }
+                break;
+            case 'delete':
+                $id = $request->request->getInt('id');
+                if ($id) {
+                    $this->service->delete($id);
+                    $this->flashMessage = xl('Preference deleted');
```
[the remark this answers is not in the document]
kojiromike: Tracked as #11853 (covers both preference cards). Pre-existing — out of scope for this PR.

src/Patient/Cards/InsuranceViewCard.php
```diff
@@ -161,8 +162,7 @@ private function getEligibilityOutput()
         $output = '';
         $pid = $this->pid;
         if (OEGlobalsBag::getInstance()->getBoolean("enable_eligibility_requests")) {
-            if (($_POST['status_update'] ?? '') === 'true') {
-                unset($_POST['status_update']);
+            if (HttpRestRequest::createFromGlobals()->request->getString('status_update') === 'true') {
                 $showEligibility = true;
                 $ok = EDI270::requestEligibleTransaction($pid);
                 if ($ok === true) {
```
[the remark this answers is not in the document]
kojiromike: Tracked as #11854. The missing CSRF token and ACL check predate this PR; out of scope for the superglobal refactor and will be addressed in a follow-up.

src/ClinicalDecisionRules/Interface/Common.php
```diff
@@ -64,8 +68,15 @@ public static function get($var, $default = ''): string
...
-        return isset($val) && $val !== '' ? $val : $default;
+        $val = Request::createFromGlobals()->request->all()[$var] ?? null;
+        if (is_array($val)) {
+            /** @var string[] $val */
+            return $val;
+        }
+        if (is_string($val) && $val !== '') {
+            return $val;
+        }
+        return $default;
```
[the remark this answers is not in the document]
kojiromike: Title updated to reflect the actual approach (`HttpRestRequest` rather than `filter_input()`). Description body already describes the Symfony Request path.

Issue #11312, opened 2026-03-27
Clean up CAMOS forms and C_Prescription after mitigateSqlTableUpperCase removal


kojiromike:
Follow-up to #11298 / #11299. The wrapper removal surfaced several cleanup opportunities in the CAMOS form files and `C_Prescription.class.php`. These were identified during PR review but are out of scope for the minimal wrapper-removal PR.

### CAMOS admin.php

- Remove redundant `?>` / `<?php` tag pairs
- Swap tmpfile condition to `if ($temp) {` (positive case first)
- Inline `$query*` variables directly into `sqlStatement()` calls
- Wrap export block in `try/finally` to ensure `fclose($temp)` runs (resource leak)
- Remove pointless if/else where both branches assign `$content = $value`
- Use single-quoted `Content-Disposition` header

### CAMOS new.php

- Use string interpolation instead of concatenation for simple variable strings
- Inline `$query` variable assignments directly into `sqlStatement()` calls
- Parameterize dynamically-built clone queries (currently uses `add_escape_custom()`)
- Replace `searchName()` (returns raw SQL fragments) with a function that returns PIDs for parameterized `IN (?)` clauses

### CAMOS notegen.php

- Format SQL with proper indentation and uppercase keywords

### CAMOS view.php

- Replace multi-part string concatenation with single interpolated query

### CAMOS rx_print.php

- Add space around `=` in SQL (`id =?` → `id = ?`)
- Use string interpolation instead of concatenation

### CAMOS save.php

- Use consistent SQL keyword capitalization

### controllers/C_Prescription.class.php

- Delete commented-out code (old `SELECT *` query)
- Fix nonsensical `== false` / `== true` conditions on `querySingleRow()` result (use `=== false` / `else`)
- Move HTML string building out of SQL `CONCAT()` — select individual fields and build strings in PHP

### Cross-cutting

- Migrate all `sqlStatement()` / `sqlFetchArray()` / `sqlQuery()` calls to `QueryUtils::` equivalents
- Handle `querySingleRow()` returning `false` (PHPStan offset-access errors)
- Fix `preg_replace()` in `C_Prescription` that incorrectly operates on the full result array instead of a single column
- Resolve as many PHPStan errors as possible in these files

PR #11349, opened 2026-03-30
refactor(prescription): parameterize SQL, fix loose comparisons, remove dead code

haoyu-haoyu:
## Summary

Clean up `controllers/C_Prescription.class.php` as follow-up to #11312. The CAMOS form files are handled separately in #11313; this PR addresses all C_Prescription items from the same issue.

## Changes

| Category | Before | After |
|----------|--------|-------|
| SQL injection | `add_escape_custom()` in string-concatenated SQL with `CONCAT()` building HTML in MySQL | Parameterized queries selecting individual columns, address strings built in PHP |
| Bug: `preg_replace` on array | `preg_replace($patterns, $replace, $res)` where `$res` is the full `sqlQuery()` result array | Build HTML directly from individual columns — no regex needed |
| Loose comparison | `$rxn == false` / `elseif ($rxn == true)` | `$rxn === false` / `else` |
| Dead code | Commented-out `SELECT *` query, `print_r()` debug calls | Deleted |
| XSS | Facility address output unescaped in CSS print view | Wrapped in `text()` |
| Superglobal | `$_GET['ids']` | `filter_input(INPUT_GET, 'ids')` |

### The `preg_replace` bug explained

The old code ran:
```php
$res = sqlQuery("SELECT CONCAT(...) AS addr FROM ...");
$res = preg_replace($patterns, $replace, $res);
echo $res['addr'];
```

`sqlQuery()` returns an associative array like `['addr' => '...']`. Passing this array to `preg_replace()` applies the patterns to every value in the array and returns a new array — but the keys and structure are silently mangled. The fix eliminates the regex entirely by building the HTML string directly from individual columns.

## Test plan

- [ ] Verify prescription PDF print (multiprint) shows correct facility and patient address
- [ ] Verify CSS prescription print shows correct addresses with proper HTML escaping
- [ ] Verify RxNorm drug interaction check works (especially the `=== false` path)
- [ ] Verify email prescription (getDefaultMailClientText) works
- [ ] Verify plain-text prescription output (multiprintplain) is unchanged

Ref #11312

### DIFF
--- a/.phpstan/baseline/binaryOp.invalid.php
+++ b/.phpstan/baseline/binaryOp.invalid.php
@@ -432,16 +432,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../controllers/C_Prescription.class.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Binary operation "\\." between \'SELECT  concat…\' and mixed results in an error\\.$#',
-    'count' => 2,
-    'path' => __DIR__ . '/../../controllers/C_Prescription.class.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Binary operation "\\." between \'SELECT concat\\(\\\\\'\\<b\\>\\\\\'…\' and mixed results in an error\\.$#',
-    'count' => 2,
-    'path' => __DIR__ . '/../../controllers/C_Prescription.class.php',
-];
-$ignoreErrors[] = [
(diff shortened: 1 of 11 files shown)


kojiromike (approved):
Solid cleanup — SQL parameterization, `preg_replace`-on-array bug fix, loose-to-strict comparisons, XSS escaping via `text()`, dead code removal, and `filter_input()` replacing raw `$_GET`. The new code is correct and well-scoped. The noisy baseline changes in unrelated files appear to be PHPStan cache artifacts and should resolve on merge.

PR #11362, opened 2026-03-30
fix(phpstan): emit ADOdb-accurate types in TableTypes generator

### LINKED ISSUE #11361: fix(phpstan): TableTypes generator should emit string types matching ADOdb runtime
The TableTypes generator (GeneratePhpstanTypesCommand) maps SQL column types to PHPStan types based on the schema — e.g. SQL `int` becomes PHPStan `int`. However, ADOdb's FetchRow() returns `string|null` for all column values at runtime, not native PHP types. A SQL `int` column comes back as `'5'`, not `5`.

This means the generated type stubs are inaccurate: PHPStan believes the values are `int` and flags `(int)` casts as unnecessary (Rector's RecastingRemovalRector removes them), but the casts are actually needed to avoid `TypeError` under `strict_types=1`.

The generator should emit types matching what ADOdb actually returns:
- SQL `int`, `tinyint`, `bigint`, etc. → PHPStan `numeric-string` (or just `string`)
- SQL `varchar`, `text`, etc. → PHPStan `string`
- SQL nullable columns → append `|null`
- SQL `NOT NULL` columns → no `|null`

This would make the type stubs truthful and let PHPStan catch real type mismatches (e.g. passing a string to an int parameter without casting).

Related: #11359 (nullability issue in the same generator)

kojiromike:
## Summary

ADOdb's FetchRow() returns all scalar values as `string|null` at runtime, regardless of the underlying SQL column type. The TableTypes generator previously mapped SQL integer/float types to PHPStan `int`/`float`, which was inaccurate — it caused Rector to remove necessary `(int)` casts and hid real type mismatches.

Now emits `numeric-string` for integer, float, decimal, and boolean SQL types, matching what code actually receives from ADOdb. This lets PHPStan catch real type mismatches (e.g. passing a numeric string to an `int` parameter without casting).

## Test plan

- [x] Pre-commit hooks pass (phpstan, rector, phpcs)
- [x] Regenerated TableTypes.php matches generator output
- [ ] Verify existing consumers still work (baseline updated for pre-existing type mismatches)

Closes #11361


src/Common/Command/GeneratePhpstanTypesCommand.php
```diff
@@ -130,25 +130,26 @@ private function preprocessSql(string $sql): string
     /**
      * Map SQL types to PHPStan types
      *
-     * Note: ADODB returns all values as strings by default, but we type them
-     * based on their semantic meaning for better static analysis.
+     * ADOdb's FetchRow() returns all scalar values as string|null, regardless
+     * of the underlying SQL column type. The types here reflect runtime
+     * reality, not the SQL schema's logical types.
```
kojiromike, line 135: @Firehed this was painful.

Firehed: I ran into something messy like that with some part of the rainforest id generation stuff. I'm sure it'll _never_ be a problem in reality 🙄

.phpstan/baseline/argument.type.php
```diff
@@ -74931,6 +74926,11 @@
     'count' => 1,
     'path' => __DIR__ . '/../../tests/Tests/Services/DocumentTest.php',
 ];
+$ignoreErrors[] = [
+    'message' => '#^Parameter \\#8 \\$owner of method Document\\:\\:createDocument\\(\\) expects int, string given\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../tests/Tests/Services/DocumentTest.php',
+];
```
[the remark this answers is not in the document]
kojiromike: Pre-existing baseline entry that shifted between baseline files due to the type change. The DocumentTest type mismatch is unrelated to this PR.

Issue #11259, opened 2026-03-24
fix: replace addslashes() with context-appropriate escaping


kojiromike:
`addslashes()` is used throughout the codebase in contexts where purpose-built escaping functions should be used instead. `addslashes()` only escapes quotes, backslashes, and NUL bytes — it is not sufficient for JavaScript, SQL, or CSV escaping.

### JavaScript context (~30 occurrences)

Should use `js_escape()` (or `js_escape_protected()` where newlines need preserving):

- `interface/patient_file/pos_checkout_ippf.php` — maskkeyup/maskblur handlers
- `interface/drugs/add_edit_drug.php` — alerts and maskkeyup handlers
- `interface/drugs/add_edit_lot.php`, `destroy_lot.php` — alerts
- `interface/billing/sl_eob_search.php` — alert
- `interface/main/calendar/add_edit_event.php` — alert and confirm
- `interface/forms/eye_mag/a_issue.php` — alerts
- `interface/forms/LBF/new.php` — JS string
- `interface/forms/CAMOS/new.php` — JS array building (notably already uses `js_escape()` for other values in the same line)
- `interface/fax/fax_dispatch.php` — alert
- `ccr/createCCR.php` — alert in script tag
- `custom/import_xml.php` — alert
- `library/custom_template/add_context.php`, `add_template.php`, `delete_category.php` — alerts and confirms

### `attr(addslashes(...))` in onclick handlers (~12 occurrences)

`addslashes()` is used for JS escaping inside `attr()` for HTML. Should use `attr(js_escape(...))` or preferably `data-` attributes:

- `interface/reports/destroyed_drugs_report.php`
- `interface/drugs/drug_inventory.php`
- `interface/fax/faxq.php` (5 occurrences)
- `interface/forms/eye_mag/SpectacleRx.php`
- `library/dated_reminder_functions.php` (3 occurrences)
- `interface/main/finder/multi_patients_finder.php`
- `interface/main/messages/messages.php` (2 occurrences)

### SQL/log string building (security smell)

`addslashes()` is not a safe SQL escaping mechanism. These should use parameterized queries or proper escaping:

- `gacl/admin/acl_admin.php` (3 occurrences)
- `gacl/admin/assign_group.php` (3 occurrences)
- `library/payment.inc.php` — log string building
- `interface/forms/eye_mag/save.php` — log string building

### CSV context

Should use `fputcsv()` or proper CSV quoting:

- `custom/BillingExport.csv.php` (4 occurrences)

[...]

Issue #10818, opened 2026-02-19
refactor(csv): Update various CSV generation paths to use built-in or library tooling


Firehed:
There are a number of places in the application that appear to be building CSVs by hand, one cell at a time. PHP has native support for this, in `fputcsv`. It does all of the escaping, encoding, and whatnot.

I see a dozen-odd places calling `csvEscape` which seems to be a strong indicator for this pattern.

There is an existing `SpreadSheetService` (which does this); I think it should be possible to adjust its internals and the migrate the other places to use it.

Firehed:
Related: #10415.

PR #11363, opened 2026-03-30
fix(csv): replace addslashes() with RFC 4180 CSV double-quote escaping

haoyu-haoyu:
## Summary

Replace 4 `addslashes()` calls in `custom/BillingExport.csv.php` with RFC 4180-compliant CSV escaping (`str_replace('"', '""', ...)`).

`addslashes()` escapes quotes with backslashes (`\"`), which is not valid CSV. RFC 4180 specifies that double quotes within a quoted field must be escaped by doubling them (`""`). This file manually builds CSV output using `fwrite()` with `,"value"` formatting, so the field values must use the CSV quoting convention.

### Changed functions

| Function | Before | After |
|----------|--------|-------|
| `fixString()` | `addslashes(trim($string))` | `str_replace('"', '""', trim($string))` |
| `fixMI()` | `addslashes(substr(...))` | `str_replace('"', '""', substr(...))` |
| `fixCPT()` | `addslashes($code)` | `str_replace('"', '""', $code)` |
| `fixJust()` | `addslashes(trim(...))` | `str_replace('"', '""', trim(...))` |

## Test plan

- [ ] Verify CSV export with patient data containing double quotes (e.g., name `O"Brien`)
- [ ] Verify CSV can be imported into Excel/LibreOffice without field misalignment

Ref #11259


Firehed (changes requested):
Before proceeding with this as-is, please also see #10818. While this seems to be an improvement, I think it's likely trading one issue with CSV exporting for another (or at least setting one up).

haoyu-haoyu:
Thanks for the review @Firehed — you're right, the original approach was still hand-rolling CSV escaping.

I've rewritten the commit to use PHP's built-in `fputcsv()` instead:

- All 4 `fwrite()` calls with manual string concatenation → `fputcsv($fh, [...])` 
- The `fixString`/`fixMI`/`fixCPT`/`fixJust` helpers now only do data formatting (trim, substr) — `fputcsv()` handles all RFC 4180 quoting/escaping
- No more `addslashes()` or `str_replace('"', '""', ...)` — that's `fputcsv`'s job

This aligns with the direction in #10818 for this file. The remaining ~15 report files that use `csvEscape()` with manual `echo` are a bigger refactor that could be tackled separately.

Firehed:
This is looking much better and safer, thank you! Can you kindly address the PHPStan warnings/errors? I think a couple `@param resource $fh` type annotations will handle a lot of it based on CI feedback. I'd prefer to avoid adding them into the baseline.

Feel free to tighten up anything else in the class if you're so inclined; it looks like a lot of the methods could be made private and made more type-safe, and I didn't see much pointing to it int he rest of the codebase.

Firehed (approved):
LGTM now, thanks!

(I suspect some of the PHPStan removals are a slightly outdated base, but no worries)

Firehed:
Note: I adjusted the commit message slightly to reflect the fputcsv changes. I hope you don't mind!

haoyu-haoyu:
@Firehed Thank you for the feedback! I've rebased onto the latest master. The code already has:

- `/** @var resource */ public $tmpfh` — typed property annotation
- `@param list<bool|float|int|string|null> $fields` — typed parameter for `writeCsvRow()`
- All helper methods (`fixString`, `fixMI`, `fixCPT`, `fixJust`, etc.) already marked `private` with `mixed` → `string` return types
- PHPStan baseline entries **reduced** (removed 6 entries from 3 baseline files)

CI is fully green after rebase. Ready for re-review when you have a moment.

Issue #10572, opened 2026-02-04
`tests/Tests/BC/DatabaseTest.php` (37 lines) is not listed in `phpunit.xml`, `phpunit-isolated.xml`, or `phpunit.integration.xml`. It will never be discovered by any test run.


kojiromike:

Firehed:
See also #10084

PR #11372, opened 2026-03-31
fix(tests): add integration tests to phpunit.xml and CI suite list

### LINKED ISSUE #11201: fix(tests): tests/Tests/Integration/ not listed in any phpunit.xml testsuite
## Summary

`tests/Tests/Integration/` is not included in any testsuite in `phpunit.xml`, so CI never discovers or runs these tests. Same pattern as #10572.

## Affected tests

- `tests/Tests/Integration/Api/ApiApplicationTest.php`
- `tests/Tests/Integration/RestControllers/Subscriber/AuthorizationListenerTest.php`
- `tests/Tests/Integration/RestControllers/Subscriber/SiteSetupListenerTest.php`

## Fix

Add the directory to an existing testsuite in `phpunit.xml`, or create a dedicated `integration` suite. Some of these tests currently have failures that would need to be fixed first.

haoyu-haoyu:
## Summary

- Add a dedicated `integration` testsuite to `phpunit.xml` pointing to `tests/Tests/Integration/`
- Add `"integration"` to the CI workflow's suite matrix in `.github/workflows/test.yml`

### Background

`tests/Tests/Integration/` contains `ApiApplicationTest` and `AuthorizationListenerTest` but was not listed in any testsuite in `phpunit.xml`, so CI never discovered or ran these tests. Same pattern as #10572.

Note: `phpunit.integration.xml` exists with a lightweight bootstrap but is never referenced by CI. This PR adds the directory to the main `phpunit.xml` so it runs in the Docker CI environment alongside other test suites.

Closes #11201

## Test plan

- [ ] CI discovers and attempts to run the integration test suite
- [ ] If any integration tests fail, they should be fixed in follow-up PRs

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -49,6 +49,9 @@ This file configures the kind that requires initialization.
     <testsuite name="common">
       <directory>tests/Tests/Common</directory>
     </testsuite>
+    <testsuite name="integration">
+      <directory>tests/Tests/Integration</directory>
+    </testsuite>
     <testsuite name="certification">
       <directory>tests/Tests/Certification/HIT1</directory>
     </testsuite>
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,9 +117,9 @@ jobs:
       id: compute-suites
       run: |
         if [[ '${{ steps.parse.outputs.e2e_enabled }}' = 'true' ]]; then
-          echo 'suites=["unit","api","fixtures","services","validators","controllers","common","email","e2e"]' >> "$GITHUB_OUTPUT"
+          echo 'suites=["unit","api","fixtures","services","validators","controllers","common","integration","email","e2e"]' >> "$GITHUB_OUTPUT"
         else
-          echo 'suites=["unit","api","fixtures","services","validators","controllers","common","email"]' >> "$GITHUB_OUTPUT"
+          echo 'suites=["unit","api","fixtures","services","validators","controllers","common","integration","email"]' >> "$GITHUB_OUTPUT"
         fi
 
     - name: Setup PHP
--- a/.phpstan/baseline/deadCode.unreachable.php
+++ b/.phpstan/baseline/deadCode.unreachable.php
@@ -294,6 +294,16 @@ $ignoreErrors[] = [
 $ignoreErrors[] = [
     'message' => '#^Unreachable statement \\- code above always terminates\\.$#',
     'count' => 1,
+    'path' => __DIR__ . '/../../tests/Tests/Integration/Api/ApiApplicationTest.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Unreachable statement \\- code above always terminates\\.$#',
+    'count' => 2,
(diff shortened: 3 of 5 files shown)


phpunit.xml
```diff
@@ -47,6 +47,9 @@ This file configures the kind that requires initialization.
     <testsuite name="common">
       <directory>tests/Tests/Common</directory>
     </testsuite>
+    <testsuite name="integration">
```
kojiromike, line 52: The integration tests have 3 pre-existing failures (19 tests, 77 assertions, 3 failures). Merging as-is would make the `integration` CI job fail on every PR.

Could you either fix the failing tests in this PR, or mark them as `$this->markTestIncomplete('Known failure — see #XXXX')` with tracking issues so CI stays green?

kojiromike (changes requested):
Good idea wiring up the orphaned integration tests into CI. However, the integration suite currently has 3 pre-existing test failures, which means merging this would introduce a persistently failing job on master. The tests need to either be fixed or marked incomplete before this can land.

haoyu-haoyu:
@kojiromike Thank you for the review! I've addressed the feedback:

The 3 pre-existing test failures are now marked as `$this->markTestIncomplete()` with tracking references to #11201:

1. **`ApiApplicationTest::testRunWithSkippedRoute`** — Returns 500 due to `get()` called on `false` during globals initialization
2. **`AuthorizationListenerTest::testOnRestApiSecurityCheckUpdatesRequestWithConstraints`** — Scope constraint array format mismatch after ScopeValidatorFactory changes
3. **`AuthorizationListenerTest::testOnRestApiSecurityCheckMergesConstraintsFromMultipleScopes`** — Same root cause as #2

The remaining **16 tests** (all using mocks) continue to pass without issues. CI should now stay green with the integration suite reporting 16 passes + 3 incomplete.

Also rebased onto latest master to pick up the recent upstream changes.

kojiromike:
Hi @haoyu-haoyu — this one is 14 days old with changes requested. As noted in my review, the integration tests have 3 pre-existing failures that would break CI on every PR if merged as-is. Please either fix those tests or mark them `markTestIncomplete` with tracking issues so CI stays green. Please address this before opening new PRs. If you need help, feel free to ping me with questions.

haoyu-haoyu:
Thanks @kojiromike — sorry for the delay. Update pushed:

**Rebase onto master** — clean single-commit replay; the intermediate merged PRs (#11349, #11373, #11423, #11424) are now reached via upstream rather than in my branch history.

**Tracking issue for the 3 incomplete tests** — opened #11609 to specifically track what each test needs before it can be restored to passing. Each `markTestIncomplete()` call now references #11609 (previously #11201, but that issue is itself resolved by this PR, so the self-reference would have gone stale on merge).

**Copilot's alternative suggestion** — use CI-level allowed-failure job rather than `markTestIncomplete` — is a valid path but a larger CI-workflow change. Your explicit guidance was "mark them as markTestIncomplete with a tracking issue so CI stays green", which is what this PR does. Happy to revisit if you'd prefer the allowed-failure route.

Ready for another look.

kojiromike (approved):

PR #11374, opened 2026-03-31
fix(oauth): guard against missing grant_type in token request

haoyu-haoyu:
## Summary

- Add null coalescing operator (`?? ''`) to `getParsedBody()['grant_type']` access in `AuthorizationController::oauthAuthorizeToken()`

### Background

The token endpoint accesses `$request->getParsedBody()['grant_type']` without null coalescing (line 1339), while the adjacent `code` key access (line 1337) correctly uses `?? null`. When `grant_type` is missing from the POST body, this causes:

1. PHP 8.x "Undefined array key" warning
2. A subsequent TypeError trying to assign `null` to the `string`-typed `$grantType` property

This is visible in the [demo server error log](https://demo.openemr.io/log/logPhp.txt).

With the `?? ''` default, the empty string safely fails all grant type comparisons (`=== 'authorization_code'`, `=== 'refresh_token'`, etc.), allowing the OAuth2 server to return the proper error response.

Ref #7598

## Test plan

- [ ] Send a POST to the token endpoint without `grant_type` — verify no PHP warning and proper OAuth2 error response
- [ ] Normal OAuth2 flows (authorization_code, refresh_token) continue to work

### DIFF
--- a/src/RestControllers/AuthorizationController.php
+++ b/src/RestControllers/AuthorizationController.php
@@ -1336,7 +1336,7 @@ class AuthorizationController
         // by the authorization grant flow.
         $code = $request->getParsedBody()['code'] ?? null;
         // grantType could be authorization_code, password or refresh_token.
-        $this->grantType = $request->getParsedBody()['grant_type'];
+        $this->grantType = $request->getParsedBody()['grant_type'] ?? '';
         $this->getSystemLogger()->debug("AuthorizationController->oauthAuthorizeToken() grant type received", ['grant_type' => $this->grantType]);
         if ($this->grantType === 'authorization_code') {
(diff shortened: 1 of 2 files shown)


Firehed (commented):
The change here looks ok, but can we update the tests for this class to prevent a regression?

haoyu-haoyu:
@Firehed Good call! I've added `testOauthAuthorizeTokenMissingGrantType` to `AuthorizationControllerTest.php` following the existing test pattern.

The test:
- Creates a POST request to `/oauth2/default/token` with `code` but **without `grant_type`**
- Asserts the response is HTTP 400 (Bad Request)
- Verifies the `?? ''` guard prevents the PHP undefined-array-key warning while letting the OAuth2 server return the proper error

PR #11401, opened 2026-04-02
fix(portal): exclude restricted-category documents from patient portal (#11284 #11279)

### LINKED ISSUE #11279: Patient Portal: Unable to download normal note if high sensitivity note included
### Describe the bug

In the patient portal, if a patient tries to download multiple notes associated with an encounter, and one of the notes is marked as High sensitivity, then none of the notes in subsequent DOS will show up

### To Reproduce

-  Create an clinical LBF form where the access control is set to High Sensitivity.

Image

- For a patient, create a SOAP note (or any note with normal sensitivity) for encounter 1 with DOS of yesterday. (this will be references as Note A). Add High sensitive note (Note B)
- Create encounter 2 with DOS of today with High sensitive note from step one (Note C) and a regular SOAP note (Note D).
- Log into Patient Portal > Medical Reports > Customized Medial History Report
- Generate the report with the following combinations:

| Scenario | Note A  | Note B (high sensitivy) | Note C (High) | Note D  | Outcome                |
|----------|---------|-------------------------|---------------|---------|------------------------|
| 1        | Include | Include                 | Include       | Include | Only Note A displayed  |
| 2        | Include |                         | Include       | Include | Only Note A displayed  |
| 3        |         | Include                 | Include       | Include | No notes displayed     |
| 4        | Include |                         |               | Include | Note A and D displayed |

- Scenario 2:
Image

Image

- Scenario 4

Image

Image

### Expected behavior
1. Display all notes in in the report that are selected that are of normal sensitivity. The order of note sensitivity should not impact the other notes from the report generation.

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

### LINKED ISSUE #11284: Patients able to access document marked as high sensitivity
### Describe the bug

Patients are able to access documents that put in a High sensitive document category. Patients are also able to download the document if the encounter is marked as High sensitivity.

### To Reproduce

1. Create a new document category and set Access to "High sensitivity"
2. Add a document to the patient to the High sensitivity category.
3. Log into patient portal and see that you can download document.
    1. User in EMR with ACL to High sensitivity is not able to access document
4.  Associate the high sensitive document to an encounter with normal sensitivity
    1. User in EMR with ACL to High sensitivity is not able to access document
    2. Patient able to access document via Portal.
5. Mark the encounter as High Sensitivity
    1. User in EMR with ACL to High sensitivity is not able to access document
    2. Patient able to access document via Portal. Patient unable to access other contents of the note that is marked as high sensitivity 

- **Accessing in EMR with user that does not have High sensitivity ACL**
Image

Image

Image

- **Able to access and download as a patient**
Image

### Expected behavior

High sensitivity for documents should be respected on the patient portal. There is a use case where I would want to have documents associated to the patient but do not want the patient to access that specific document(s) that are considered High sensitivity.

### Client configuration

- Browser:
- OpenEMR version:
- Operating system:

craigrallen:
## Summary

Fixes two related patient portal security/UX bugs:

### Fix #11284 — Patients able to access high-sensitivity documents

**Root cause:** `get_patient_documents.php` and `document_downloads_action.php` fetched all documents for the patient without checking the document category's `aco_spec` field. Documents in categories with restricted ACO specs (e.g. `patients|high`, `admin|super`) were visible and downloadable by patients.

**Fix:** Both files now exclude documents where ANY assigned category has an `aco_spec` other than `patients|docs` (the standard patient-accessible value). A `NOT EXISTS` subquery is used so that a document is hidden if it belongs to even one restricted category.

### Fix #11279 — High-sensitivity note in portal report breaks subsequent notes

**Root cause:** `lbf_report()` called `AccessDeniedHelper::deny()` — a `never`-returning function — when the portal requested a restricted LBF form. This terminated the entire HTTP response, preventing all subsequent (normal-sensitivity) notes from appearing in the same Customized Medical History Report.

**Fix:** When `OEGlobalsBag::patient_portal_onsite_two` is set (portal context), the function returns early instead of calling `deny()`. The report continues rendering the remaining items. An EMR user without the required ACL still hits the deny path as before.

## Changed files

- `portal/get_patient_documents.php` — filter restricted docs with `NOT EXISTS`
- `portal/report/document_downloads_action.php` — same filter on download; also fixes bug where `$file` could be empty after sqlQuery but was not checked before use
- `interface/forms/LBF/report.php` — return early in portal context instead of calling `deny()`

## Tests added

- `tests/Tests/Isolated/Portal/PortalDocumentSensitivityTest.php` — 6 tests covering:
  - SQL `NOT EXISTS` filter present in listing file
  - SQL `NOT EXISTS` filter present in download action
  - aco_spec accessibility data provider (6 spec values)
  - lbf_report returns early for portal context

Fixes #11284
Fixes #11279


portal/report/document_downloads_action.php
```diff
@@ -35,12 +35,26 @@
...
+              AND NOT EXISTS (
+                  SELECT 1
+                  FROM `categories_to_documents` AS ctd
+                  INNER JOIN `categories` AS c ON c.id = ctd.category_id
+                  WHERE ctd.document_id = d.id
+                    AND c.`aco_spec` != 'patients|docs'
+              )";
     $file = sqlQuery($sql, [$documentId]);
-    if ($file['foreign_id'] != $pid && $file['foreign_id'] != $pid) {
+    if (empty($file) || $file['foreign_id'] != $pid) {
```
kojiromike: New code should not use `empty()` — it's forbidden by the PHPStan ruleset (`empty.notAllowed`). Since `sqlQuery()` returns `false` on no results:

```suggestion
    if ($file === false || $file['foreign_id'] != $pid) {
```

portal/report/document_downloads_action.php
```diff
@@ -35,12 +35,26 @@
 
 // Process each selected document
 foreach ($documentIds as $documentId) {
-    $sql = "SELECT url, id, mimetype, `name`, `foreign_id` FROM `documents` WHERE `id` = ? AND `deleted` = 0";
+    // Verify the document belongs to this patient, is not deleted, and does not
+    // belong to any restricted category.  Using NOT EXISTS with aco_spec check
+    // ensures the portal cannot serve high-sensitivity or admin-only content
+    // even when document IDs are submitted directly (bypass of the listing UI).
+    $sql = "SELECT d.url, d.id, d.mimetype, d.`name`, d.`foreign_id`
```
kojiromike: This SQL string is a constant assigned inside a `foreach` loop — it gets reassigned on every iteration but never changes. Move it before the loop to make the intent clear and avoid the redundant assignment.

portal/get_patient_documents.php
```diff
@@ -39,16 +39,30 @@
...
 /**
  * @Global $pid Patient id setup during verify_session.php
  */
 $fres = sqlStatement($sql, [$pid]);
 
 $documents = [];
 while ($file = sqlFetchArray($fres)) {
-    // Find the document category
+    // Find the document category (already confirmed to be patients|docs above)
     $sql = "SELECT name, lft, rght FROM `categories`, `categories_to_documents`
```
kojiromike: Same here — this SQL string is a constant reassigned on every iteration of the `while` loop. Move it before the loop.

kojiromike (changes requested):
The security fix is well-reasoned — the `NOT EXISTS` subquery correctly hides documents in restricted categories from the portal, and the download action independently enforces the filter (defense in depth). A few items to address before merge.

The PHPStan baseline includes unrelated count bumps for `handle_note.php` (`empty.notAllowed` 15->18, `deprecatedSqlFunction` 1->3, `forbiddenRequestGlobals` 27->28) that aren't related to this PR's changes — suggests a stale rebase. Please rebase onto current master and regenerate with `composer phpstan-baseline`.

kojiromike:

Reviewers have requested changes. Wait for revision before backporting.

craigrallen:
@kojiromike I rebased this onto current master and addressed the requested loop/query and false-handling changes. I also migrated the changed queries to QueryUtils and added strict types to the new test. Targeted isolated tests, PHPStan, and PHPCS pass. Fresh CI is running; re-review when convenient.

interface/forms/LBF/report.php
```diff
@@ -36,6 +36,16 @@ function lbf_report($pid, $encounter, $cols, $id, $formname, $no_wrap = false):
     }
     if (!AclMain::aclCheckCore('admin', 'super') && !empty($LBF_ACO)) {
         if (!AclMain::aclCheckCore($LBF_ACO[0], $LBF_ACO[1])) {
+            // When rendering from the patient portal the ACL check cannot call
+            // AccessDeniedHelper::deny() because that terminates the entire
+            // request, which would prevent subsequent (non-sensitive) notes from
+            // appearing in the same Customized Medical History Report.
+            // Instead, silently skip this restricted form and let the report
+            // continue rendering the remaining items.
+            if (OEGlobalsBag::getInstance()->get('patient_portal_onsite_two') !== null) {
```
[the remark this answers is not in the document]
craigrallen: Fixed in 8a1d7e07f: the LBF report now obtains the active Symfony session through SessionWrapperFactory and delegates the portal discriminator to LbfReportAccessGuard. The prior source-text assertion was replaced with executable portal/core session regression tests (10 isolated tests, 16 assertions pass).

PR #11402, opened 2026-04-02
fix(portal): store patient real name as message recipient, not portal_username (#11202)

### LINKED ISSUE #11202: Chart notes display portal_username instead of patient name for recipient

craigrallen:
## Summary

Fixes #11202 — chart notes in the secure messaging interface displayed `portal_username` (e.g. `zhportal`) in the recipient field instead of the patient's real name (e.g. `First Last`).

## Root cause

`handle_note.php` stored `$_POST['recipient_name']` directly into the `onsite_mail` table. The Angular front-end set this value from `compose.recipient_name`, which was populated by a jQuery selector reading the selected option text at submit time. Under some conditions the value was unset or fell through to `compose.selrecip` (the `portal_username`), resulting in the internal username being stored as the display name.

## Fix

Server-side name resolution in `handle_note.php`: when `recipient_id` is provided, the real name is looked up from:

1. `patient_data JOIN patient_access_onsite` (portal patient — `recipient_id` is `portal_username`)
2. `users` table (staff fallback for staff-to-staff messages)

The client-submitted `recipient_name` is only used as a last resort, removing the dependency on client-side rendering state and closing a potential display-name spoofing vector.

## Changed files

- `portal/messaging/handle_note.php` — server-side recipient name resolution

## Tests added

- `tests/Tests/Isolated/Portal/PortalMessageRecipientNameTest.php` — 3 tests:
  - Verifies `patient_access_onsite` JOIN is present
  - Verifies resolution occurs before `sendMail()`
  - Verifies `users` table fallback for non-portal recipients

Fixes #11202


portal/messaging/handle_note.php
```diff
@@ -105,7 +105,32 @@
 $sid = $_POST['sender_id'] ?? null;
 $sn = $_POST['sender_name'] ?? null;
 $rid = $_POST['recipient_id'] ?? null;
-$rn = $_POST['recipient_name'] ?? null;
+// Resolve recipient_name server-side so the stored display name is always the
+// patient's real name (fname + lname) rather than whatever the client submitted.
+// This prevents portal_username from appearing in the message list when a
+// clinician sends a chart note to a patient (issue #11202).
+if (!empty($rid)) {
```
kojiromike: New code should use explicit checks instead of `empty()` — it's forbidden by the PHPStan ruleset (`empty.notAllowed`) and the baseline was bumped to accommodate this rather than writing clean code.

```suggestion
if ($rid !== null && $rid !== '') {
```

portal/messaging/handle_note.php
```diff
@@ -105,7 +105,32 @@
...
+    // First try to look up as a portal patient (rid = portal_username).
+    $recipPatient = sqlQuery(
```
kojiromike: `sqlQuery()` is deprecated. New code should use `QueryUtils::querySingleRow()` instead:

```suggestion
    $recipPatient = QueryUtils::querySingleRow(
```

portal/messaging/handle_note.php
```diff
@@ -105,7 +105,32 @@
...
+        "SELECT CONCAT(pd.fname, ' ', pd.lname) AS full_name
```
kojiromike: `CONCAT()` returns NULL if either argument is NULL. Consider `CONCAT_WS` which skips NULLs:

```suggestion
        "SELECT CONCAT_WS(' ', pd.fname, pd.lname) AS full_name
```

kojiromike (changes requested):
The fix direction is right — resolving `recipient_name` server-side instead of trusting `$_POST` is the correct approach. A few items need attention before this is mergeable.

The baseline includes unrelated count bumps for files not in this PR's diff (e.g., `interface/forms/LBF/report.php`, `portal/report/document_downloads_action.php`), suggesting a stale rebase. Please rebase onto current master and regenerate the baseline with `composer phpstan-baseline` so only counts affected by this PR change.

The tests are file-content assertions (checking that certain strings exist in source) rather than behavioral tests — they don't verify the logic actually works. Consider replacing them with integration tests that exercise the lookup paths with test data, or remove them if the effort isn't justified.

kojiromike:

Reviewers have requested changes. Wait for revision before backporting.

portal/messaging/handle_note.php
```diff
@@ -119,7 +119,34 @@
...
+    if (is_string($patientName) && trim($patientName) !== '') {
+        $rn = $patientName;
+    } else {
+        // Fall back to EMR user lookup (rid = username) for staff-to-staff messages.
+        $recipUser = QueryUtils::querySingleRow(
+            "SELECT CONCAT_WS(' ', fname, lname) AS full_name FROM users WHERE username = ?",
+            [$rid]
+        );
+        $userName = $recipUser === false ? null : ($recipUser['full_name'] ?? null);
+        $rn = is_string($userName) && trim($userName) !== '' ? $userName : $rid;
```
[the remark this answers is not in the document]
craigrallen: I am keeping the server-resolved fallback intentionally. The submitted recipient_name is client-controlled and may be stale or tampered with; using it after both authoritative lookups fail would persist unverified display data. Falling back to the stable recipient identifier preserves traceability without trusting the request body.

craigrallen:
@kojiromike I rebased this onto current master and addressed the requested recipient-name changes: explicit ID handling, QueryUtils::querySingleRow(), CONCAT_WS(), and server-side patient/staff name resolution. The rejected source-content test was removed and the baseline synchronized. Fresh CI is running; re-review when convenient.

PR #11404, opened 2026-04-02
refactor(config): replace encrypted more_secure globals with constants

### LINKED ISSUE #11403: chore(config): Remove `more_secure` encryption setup from sites/x/config.php
Encrypting a string that's available in source code of an open source project only to decrypt it later is... an interesting way to guard against an extremely specific risk vector. We should just move the values to a constant for now which bypasses the risk the encryption aims to guard against (someone can modify globals but somehow not get access to the encryption tooling).

This makes no claim as to whether it should still be there at all. It seems _extremely_ niche (you'd use the print button in the browser...), and this should invite discussion about its continued existence. But we can at least bypass some complexity for almost free here.

Firehed:

Fixes #11403

Note: this is technically a very small BC break for the rare case of anyone both having customized `sites/{}/config.php` file in their installation AND using one of these paths, but it's far more more likely that other recent changes would have already caused snags. This will actually reduce the chance of future disruption.

The `more_secure` mechanism in `config.php` encrypted shell commands (`print_command` for `lpr`, `hylafax_enscript` for fax) in memory to protect against runtime `$GLOBALS` manipulation.

This complexity is no longer necessary:
- The commands are hardcoded in `config.php` and not user-editable
- Simple constants provide the same protection with less overhead

This PR replaces the encrypted globals with simple constants:
- `OPENEMR_PRINT_COMMAND` - used by statement printing
- `OPENEMR_HYLAFAX_ENSCRIPT` - used by fax dispatch

Net result: -63 lines, no runtime encryption overhead, clearer code.

### DIFF
--- a/interface/fax/fax_dispatch.php
+++ b/interface/fax/fax_dispatch.php
@@ -19,7 +19,6 @@ require_once("$srcdir/forms.inc.php");
 require_once("$srcdir/options.inc.php");
 require_once("$srcdir/gprelations.inc.php");
 
-use OpenEMR\BC\ServiceContainer;
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Session\SessionWrapperFactory;
@@ -313,7 +312,7 @@ if ($_POST['form_save']) {
         $cpstring = str_replace('{MESSAGE}', $form_message, $cpstring);
         fwrite($tmph, $cpstring);
         fclose($tmph);
-        $tmp0 = exec("cd " . escapeshellarg($webserver_root . '/custom') . "; " . escapeshellcmd((ServiceContainer::getCrypto())->decryptStandard(OEGlobalsBag::getInstance()->get('more_secure')['hylafax_enscript'])) .
+        $tmp0 = exec("cd " . escapeshellarg($webserver_root . '/custom') . "; " . escapeshellcmd(OPENEMR_HYLAFAX_ENSCRIPT) .
         " -o " . escapeshellarg($tmpfn2) . " " . escapeshellarg($tmpfn1), $tmp1, $tmp2);
         if ($tmp2) {
               $info_msg .= "enscript returned $tmp2: $tmp0 ";
--- a/sites/default/config.php
+++ b/sites/default/config.php
@@ -1,28 +1,16 @@
 <?php
 
-use OpenEMR\BC\ServiceContainer;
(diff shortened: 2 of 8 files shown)


interface/fax/fax_dispatch.php
```diff
@@ -313,7 +312,7 @@ function mergeTiffs()
         $cpstring = str_replace('{MESSAGE}', $form_message, $cpstring);
         fwrite($tmph, $cpstring);
         fclose($tmph);
-        $tmp0 = exec("cd " . escapeshellarg($webserver_root . '/custom') . "; " . escapeshellcmd((ServiceContainer::getCrypto())->decryptStandard(OEGlobalsBag::getInstance()->get('more_secure')['hylafax_enscript'])) .
+        $tmp0 = exec("cd " . escapeshellarg($webserver_root . '/custom') . "; " . escapeshellcmd(OPENEMR_HYLAFAX_ENSCRIPT) .
```
kojiromike, line 315: symfony/process? 🙏

sites/default/config.php
```diff
@@ -1,28 +1,16 @@
...
+// This is the command to be used for printing (without the filename).
+// The word following "-P" should be the name of your printer. This
+// example is designed for 8.5x11-inch paper with 1-inch margins,
+// 10 CPI, 6 LPI, 65 columns, 54 lines per page.
+// If lpr services are installed on Windows this setting will be similar.
+// Otherwise configure it as needed (print /d:PRN) might be an option for Windows parallel printers.
+const OPENEMR_PRINT_COMMAND = 'lpr -P HPLaserjet6P -o cpi=10 -o lpi=6 -o page-left=72 -o page-top=72';
+
+// Enscript command used by HylaFAX (fax_dispatch.php).
+const OPENEMR_HYLAFAX_ENSCRIPT = 'enscript -M Letter -B -e^ --margins=36:36:36:36';
```
kojiromike, line 13: @bradymiller do we still need these custom commands?

bradymiller: pretty sure still need the custom commands. But the encryption can definitely go (I must of had way to much coffee when i decided to do that encryption :) ).

bradymiller: historically, as i recall this used to be a global config settings and was reported as a vulnerability and the fix was to move the setting to hard-coded files.

interface/fax/fax_dispatch.php
Firehed: Follow-up in #11405, trying to unblock other stuff asap/

PR #11407, opened 2026-04-02
fix(calendar): route freq_type 5 to REPEAT_ON branch to prevent infin…

stephenwaite:
…ite loop

Fixes infinite loop when saving recurring events with freq_type 5 (nth weekday of month). The REPEAT_ON threshold check in add_edit_event.php was changed from > 4 to > 6 to accommodate new repeat options, but this inadvertently excluded freq_type 5 from the REPEAT_ON branch, causing it to be stored as a REPEAT event with an unhandled freq_type, resulting in an infinite loop in the __increment() fast-forward loop.

Changed threshold condition from $my_repeat_type > 6 to $my_repeat_type > 6 || $my_repeat_type == 5 in add_edit_event.php to correctly route freq_type 5 events into the REPEAT_ON branch.

### DIFF
--- a/interface/main/calendar/add_edit_event.php
+++ b/interface/main/calendar/add_edit_event.php
@@ -381,7 +381,7 @@ if (!empty($_POST['form_action']) && ($_POST['form_action'] == "duplicate" || $_
         $event_date = setEventDate($_POST['form_date'], $my_repeat_freq);
     } elseif (!empty($_POST['form_repeat'])) {
         $my_recurrtype = 1;
-        if ($my_repeat_type > 6 || $my_repeat_type == 5) { // Changed from 4 to 6 to accommodate new options.
+        if ($my_repeat_type > 6 || $my_repeat_type == 5) { // 5 is nth weekday (REPEAT_ON); >6 handles 7th, 8th, 9th occurrences
             $my_recurrtype = 2;
             $time = strtotime((string) $event_date);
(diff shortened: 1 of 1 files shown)


interface/main/calendar/add_edit_event.php
```diff
@@ -381,7 +381,7 @@ function setEventDate($start_date, $recurrence)
         $event_date = setEventDate($_POST['form_date'], $my_repeat_freq);
     } elseif (!empty($_POST['form_repeat'])) {
         $my_recurrtype = 1;
-        if ($my_repeat_type > 6) { // Changed from 4 to 6 to accommodate new options.
```
kojiromike, line 384: The condition `$my_repeat_type > 6 || in_array($my_repeat_type, [5, 6])` is logically equivalent to `$my_repeat_type >= 5` (or the original `> 4`). Since `in_array($my_repeat_type, [5, 6])` covers exactly the gap between `> 4` and `> 6`, the compound expression could be simplified:

```suggestion
        if ($my_repeat_type > 4) { // Types 5+ use REPEAT_ON: 5=nth weekday, 6=last weekday, 7-9=nth occurrences
```

This is what the condition was before the regression, just with a more descriptive comment. Alternatively `>= 5` reads the same way.

kojiromike (commented):
The fix is correct — `freq_type` 5 (and 6) must enter the `REPEAT_ON` branch. One inline comment about simplifying the condition.

PR #11409, opened 2026-04-02
fix(finder): cast data_type to int in getSearchClass() for match comp…

stephenwaite:
…atibility

Fixes [community forum issue](https://community.open-emr.org/t/new-search-issues-on-version-8/26674).

PHP's match uses strict comparison (===), so string values returned
by sqlFetchArray() would never match integer arms, causing all fields
to be excluded from the search URL and New/Search to return all patients
regardless of criteria entered.

cast data_type to int in getSearchClass() for match compatibility

### DIFF
--- a/interface/new/new_comprehensive.php
+++ b/interface/new/new_comprehensive.php
@@ -52,9 +52,9 @@ $session = SessionWrapperFactory::getInstance()->getActiveSession();
 // 2 = select list
 // 0 = not searchable
 //
-function getSearchClass($data_type)
+function getSearchClass(int $data_type)
 {
-    return match ((int)$data_type) {
+    return match ($data_type) {
         // facilities
         1, 10, 11, 12, 13, 14, 26, 35 => 2,
         // date
@@ -333,7 +333,7 @@ while ($lrow = sqlFetchArray($lres)) {
         continue;
     }
 
-    $data_type = $lrow['data_type'];
+    $data_type = (int)$lrow['data_type'];
     $fldname = "form_$field_id";
     switch (getSearchClass($data_type)) {
         case 1:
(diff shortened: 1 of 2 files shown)


interface/new/new_comprehensive.php
```diff
@@ -54,7 +54,7 @@
 //
 function getSearchClass($data_type)
 {
-    return match ($data_type) {
+    return match ((int)$data_type) {
```
kojiromike: it'd be better to use filter_input on the request that's getting passed to this with the FILTER_VALIDATE_INT flag. That would move the validation to the edge and fix the issue where it occurs. Then this function can be properly typed.

stephenwaite: it comes from the sql fetch array call unfortunately

stephenwaite:
easydev not running same version of php and :elephant: as gh, https://github.com/openemr/openemr-devops/pull/628

interface/new/new_comprehensive.php
```diff
@@ -333,7 +333,7 @@ function searchme() {
         continue;
     }
 
-    $data_type = $lrow['data_type'];
+    $data_type = (int)$lrow['data_type'];
```
kojiromike, line 336:
```suggestion
    $data_type = is_numeric($lrow['data_type']) ? (int) $lrow['data_type'] : 0;
```

interface/new/new_comprehensive.php
```diff
@@ -907,7 +907,7 @@ function comprehensive_end_group(): void
         continue;
     }
 
-    switch (getSearchClass($lrow['data_type'])) {
+    switch (getSearchClass((int)$lrow['data_type'])) {
```
kojiromike, line 910:
```suggestion
    $data_type = is_numeric($lrow['data_type']) ? (int) $lrow['data_type'] : 0;
    switch (getSearchClass($data_type) {
```

kojiromike:
Alternatively: https://github.com/openemr/openemr/pull/11557

stephenwaite:
closing in favor of #11557 but may later create a backport for rel-800, thanks @kojiromike !

Issue #9701, opened 2025-12-09
Improve HTTP status code specificity in Portal API error responses


kojiromike:
## Overview

The Patient Portal API currently returns `401 Unauthorized` for all exceptions via the `RenderErrorJSON` method in `AppBasePortalController.php:114`. This could be enhanced to return more specific HTTP status codes based on the error type.

## Benefits

Returning more specific status codes would:
- Improve REST API semantics
- Make it easier for API consumers to programmatically handle different error conditions
- Provide clearer distinction between authentication, authorization, validation, and server errors

## Current Behavior

In `PatientController.php`, various error types all currently return 401:
- Invalid JSON in request body
- Authorization/permission failures  
- Validation errors
- Generic errors
- Invalid IDs

## Suggested Enhancements

These errors could be differentiated with more appropriate status codes:
- Invalid JSON or IDs → **400 Bad Request**
- Permission failures (authenticated but lacking permission) → **403 Forbidden**
- Validation errors → **400 Bad Request** or **422 Unprocessable Entity**
- Generic server errors → **500 Internal Server Error**

## Potential Implementation Approaches

1. **Custom exception classes** that carry their own status codes (UnauthorizedException, ForbiddenException, BadRequestException, etc.) and enhance RenderExceptionJSON to check exception type
2. **Status code parameter** added to RenderErrorJSON to allow callers to specify the appropriate code
3. **http_response_code()** calls before throwing exceptions where more specific status codes are needed

## Affected Files

- `portal/patient/libs/Controller/AppBasePortalController.php`
- Multiple controller files that use RenderExceptionJSON/RenderErrorJSON

PR #11411, opened 2026-04-02
chore(logging): Add and initialize basic fallback exception handler

Firehed:
Ref #8756, #9701, #8964, probably others.

When OpenEMR throws an exception and it's not caught, we tend to get the (in)famous "white screen of death". Depending on `php.ini` settings you may get slightly different output, but the end result is the same - absolutely no useful information at all.

This makes an error handler class, and initializes it in `interface/globals.php` which serves as a bootstrap for many paths. With this, in the event of an uncaught exception, it'll be logged and _something_ will be displayed on the page. If you have `OPENEMR__ENVIRONMENT` set to `dev` (used in version and kernel), it'll render the stack trace in addition to logging it.

_This is a very basic start._ Future enhancements (vaguely noted in the code) include things like:
- Better awareness of log levels and runtime environment
- Customizing the failure information based on exception type (e.g. a `RuntimeException` may send a 4xx class; a request indicating it wants `json` back might format the output as such)
- Pluggable and chainable error handlers (remote logging, telemetry, etc)

There's a lot more we can do to enhance this, but we've gotta start somewhere, and this should be a non-disruptive enhancement that more or less guarantees uncaught exceptions never go to the void.

Practical runtime changes:
- What was a "white screen of death" will now serve a 500 response type, where previously it may not have
- It will now send a nondescript message in `Content-type: text/plain`
- If an envvar indicates dev mode, it'll render the stack trace. This is not a default anywhere.

- Adds `ErrorHandler` class
- Adds basic smoke tests
- Changes the loading order slightly within `interface/globals.php` to make the handler install as early as possible
- Configures the exception handling path (only) in `interface/globals.php`
- Sets up strict mode in `bootstrap.php`, which is only used by some next-generation prototypes for now.

Tests were AI generated.

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -13,16 +13,6 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
-// Set up autoloader as early as possible
-require_once dirname(__DIR__) . '/vendor/autoload.php';
-
-// Checks if the server's PHP version is compatible with OpenEMR:
-$response = OpenEMR\Common\Compatibility\Checker::checkPhpVersion();
-if ($response !== true) {
-    http_response_code(500);
-    die(htmlspecialchars($response));
-}
-
 use Dotenv\Dotenv;
 use OpenEMR\BC\ServiceContainer;
 use OpenEMR\Common\Database\QueryUtils;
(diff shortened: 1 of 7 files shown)


interface/globals.php
```diff
@@ -34,8 +24,56 @@
...
+}
+
+// The webserver_root and web_root are now automatically collected.
+// If not working, can set manually below.
+// Auto collect the full absolute directory path for openemr.
+$webserver_root = dirname(__FILE__, 2);
+if (IS_WINDOWS) {
+ //convert windows path separators
+    $webserver_root = str_replace("\\", "/", $webserver_root);
+}
```
kojiromike:
```suggestion
if (DIRECTORY_SEPARATOR !== '/') {
    $webserver_root = str_replace(DIRECTORY_SEPARATOR, "/", $webserver_root);
}
```

Or even omit the `if` entirely?

src/Core/ErrorHandler.php
```diff
@@ -0,0 +1,104 @@
...
+        if ($errno === E_USER_DEPRECATED || $errno === E_DEPRECATED) {
+            if (defined('PHPUNIT_COMPOSER_INSTALL')) {
+                throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
+            } else {
+                $this->logger->warning('Deprecated: {message} ({file}:{line})', [
+                    'message' => $errstr,
+                    'file' => $errfile,
+                    'line' => $errline,
+                ]);
+            }
```
kojiromike: no else needed here.

interface/globals.php
Firehed: Probably safe. I was aiming for minimal change, but IIRC pretty much everything in PHP will resolve slashes of either type on windows so it's mostly ignorable.

tests/Tests/Isolated/Core/ErrorHandlerTest.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+/**
+ * @package   OpenEMR
+ *
+ * @link      https://www.open-emr.org
+ *
+ * @author    Eric Stern <eric@lwt.io>
```
msummers42: These copyright lines look suspect. Should represent OpenCoreEMR.

Firehed: lol good find, Claude likes to do that sometimes. Will fix!

interface/globals.php
```diff
@@ -34,8 +24,56 @@
...
+
 $logger = ServiceContainer::getLogger();
 
+// Set up exception handling: ensure that any uncaught exceptions have some
+// guaranteed way of reaching the logs, regardless of other settings.
+$handler = new \OpenEMR\Core\ErrorHandler(
+    logger: $logger,
+    shouldDisplayErrors: ($_ENV['OPENEMR__ENVIRONMENT'] ?? null) === 'dev',
+);
+$handler->installExceptionHandler();
```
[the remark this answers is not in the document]
Firehed: Unavoidable and not especially important. Worst (and fairly unlikely case) is existing behavior. It's necessary to load dotenv first so the logger can be configured.

src/Core/ErrorHandler.php
```diff
@@ -0,0 +1,104 @@
...
+
+        // For now, just crash out with minimal detail as before.
+        // In the future, this can be smarter about request context (accept
+        // headers), switching on exception type to yield the correct http
+        // code, etc.
+        http_response_code(500);
+        header('Content-type: text/plain');
+        echo 'An error has occurred.';
+
+        if ($this->shouldDisplayErrors) {
```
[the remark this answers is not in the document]
Firehed: I've added a comment explaining that this is intentional.

Issue #11405, opened 2026-04-02
chore(security): Migrate all exec calls to symfony/process


Firehed:
This improves security (more consistent escaping, etc) and provides an easier-to-use API.

- [ ] interface/billing/sl_eob_search.php
- [ ] interface/fax/fax_dispatch.php

PR #11414, opened 2026-04-03
refactor(billing): use symfony/process for print command

Firehed:

Part of #11405 - migrating shell exec calls to use symfony/process.

This is also partially to get symfony/process as an explicit dependency ahead of some other work that needs it.

- Replace `exec()` with Symfony Process using array arguments for the print command in `sl_eob_search.php`
- This is safer (no shell injection possible) and more maintainable
- Uses array arguments instead of shell string, so no manual escaping needed

---

Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -120,7 +120,7 @@
         "symfony/http-client": "~6.4.26",
         "symfony/http-foundation": "~6.4.26",
         "symfony/http-kernel": "~6.4.27",
-        "symfony/process": "^7.4",
+        "symfony/process": "^6.4",
         "symfony/psr-http-message-bridge": "~6.4.24",
         "symfony/yaml": "~6.4.26",
         "twig/twig": "^3.22.2",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -120,6 +120,7 @@
         "symfony/http-client": "~6.4.26",
         "symfony/http-foundation": "~6.4.26",
         "symfony/http-kernel": "~6.4.27",
+        "symfony/process": "^7.4",
```
kojiromike: hmm, for consistency should this be ~6.4...? Especially because with the dependabot group, this will confusingly bump symfony/* to 6.4.x, but symfony/process to 7.4.x...

Firehed: 7.4 is already installed, this just formalizes it as an explicit dependency. I doubt it would matter as far as the API goes but doing so is a downgrade. 

I'd say if anything we should try to get the others up to 7.x.

kojiromike: I agree about getting the others up to 7, and that it probably won't matter, but I still think it's a confusing inconsistency to have one symfony component explicitly on 7 while most are on 6.

kojiromike (approved):

PR #11, opened 2013-04-12
Update database.sql

garcianc:
Added two indexes to the 'form_encounter' table on 'encounter' and 'date'. This is to speed up generation of the encounters report, which was hanging or taking an extremely long time to run in our systems.

PR #11416, opened 2026-04-03
feat(oidc): OIDC Implementation

zmilan:
# Enterprise SSO via OIDC / GCIP

#### Deployment status: 
This PR is self-contained and deployable today. User↔IdP linking currently requires manual pre-provisioning (SQL seed of the oidc_external_identity table or direct DB entry), which is sufficient for initial rollout and pilots.                                                                     

Automated pre-provisioning, bulk sync from the IdP, and IdP-driven deactivation/revocation all flow through the Admin API (phase 1 (#11222) and phase 2 (#11446)). Once Admin API Phase 2 lands, a follow-up PR will add external_identity field handling to the user create/update endpoints, plus the session-revocation endpoint, closing the loop for orchestration-layer-driven user management.

Adds **enterprise single sign-on** for staff/clinician users via OIDC (OpenID Connect), with a concrete integration for **Google Cloud Identity Platform (GCIP) / Firebase Auth** as the first provider. The architecture is two-layered: a provider-agnostic core OIDC layer in `src/Common/Auth/Oidc/` and a GCIP-specific custom module in `interface/modules/custom_modules/oe-module-gcip-auth/`. Patient Portal is out of scope.

**Core OIDC Layer (`src/Common/Auth/Oidc/`)**
- Add `OidcDiscoveryClient` — fetches and caches `.well-known/openid-configuration` metadata from external OIDC providers (we serve this endpoint but never consumed one before)
- Add `OidcTokenValidator` — validates external ID tokens (JWT signature via JWKS, iss/aud/exp/iat checks, claim mapping to `NormalizedIdentity`)
- Add `OidcSessionRefreshHandler` — testable handler for the silent token-refresh flow, extracted from the procedural AJAX endpoint
- Add `OidcSessionHelper` — session-metadata read/write for OIDC token expiry, issuer/subject pinning, per-session refresh cooldown
- Add OIDC event system (`OidcLoginRequestEvent`, `OidcAuthenticationEvent`, `OidcTokenReceivedEvent`, `OidcLogoutEvent`) using Symfony EventDispatcher — zero-impact when no module listens
- Add `ExternalIdentityRepository` — maps (issuer, subject) → local `users.id` via `oidc_external_identity` table; atomic upsert via `INSERT ... ON DUPLICATE KEY UPDATE`
- Add `NormalizedIdentity` / `ExternalIdentityMapping` — typed value objects for external identity data
- Add `ClaimMapperInterface` with `StandardClaimMapper` (full OIDC claims) and `MinimalClaimMapper` (sub+iss only, for refresh path)
- Add `LocalUserDirectoryInterface` + `DatabaseLocalUserDirectory` — injectable contract for local user/ACL lookups, replacing inline `QueryUtils` calls
- Add `OidcLoginAuditLoggerInterface` + `OidcRefreshAuditLoggerInterface` with database-backed implementations — domain-specific audit logging replacing `EventAuditLogger` singleton coupling
- Add PSR-16 cache backends (`FilesystemCache`, `DatabaseCache`, `RedisCache`) for JWKS and discovery caching
- Add `TokenRevocationService` for token lifecycle management
- Add JTI replay protection in `OidcTokenValidator` using `JWTRepository` — synthetic key fallback `sha256(iss|sub|iat)` for providers that omit `jti` (Firebase/GCIP)
- Add `OidcLoginCsrf` — CSRF generation/validation for the pre-auth OIDC token POST
- Extract `IpLoginRateLimiter` from `AuthUtils` with `IpBlockStatus` value object — wire IP-based rate limiting into the OIDC login path in `auth.inc.php`
- Add PSR-16 cache support to existing `JsonWebKeySet` (closes longstanding TODO), with `getSigningKeyAsPem($kid, $alg)` and key-rotation `refresh()`
- Consolidate parallel JWKS stack — delete `Oidc/Token/JwksClient`, `JsonWebKey`, `JwksException` (~330 lines); `OidcTokenValidator` now uses the unified `JsonWebKeySet`
- Harden OAuth2/JWT type surface — narrow `string` → `non-empty-string` across `JWTClientAuthenticationService`, `JsonWebKeyParser`, `AccessTokenEntity`, `OEIdTokenResponse` (renamed from `IdTokenSMARTResponse`), `ClientCredentialsAssertionGenerator`, and 3 controller callers; replace deprecated `LocalFileReference::file()` with `InMemory::plainText()`; replace `Configuration::forUnsecuredSigner()` with direct `Parser`+`Validator` usage; remove ~30 stale PHPStan baseline entries

**OIDC Dependency Cleanup & JWT 5.x Migration**
- **Dropped `steverhoades/oauth2-openid-connect-server ^3.0.1`** — package is abandoned (v3.0.1 is the latest tag; master still references `Lcobucci\JWT\Signer\Key\LocalFileReference` which was removed in lcobucci/jwt 5.x, so its composer constraint `^5.0` is not correct). Ported all 9 package types to `OpenEMR\Common\Auth\OpenIDConnect\*` (`ClaimExtractor`, `ClaimSetEntity`, `ClaimSetInterface`, `ClaimSetEntityInterface`, `ClaimSetRepositoryInterface`, `IdentityProviderInterface`; collapsed the redundant `ScopeInterface` into `ClaimSetEntityInterface`) with modern PHP 8.2+ style (strict types, readonly value objects, native type declarations). Full MIT permission notice preserved at `src/Common/Auth/OpenIDConnect/LICENSE.steverhoades-oauth2-openid-connect-server`.
- **Bumped `lcobucci/jwt` `^4.3.0` → `^5.5`** (tightened `lcobucci/clock` to `^3.3.1` for PSR-20 support). This bump was previously blocked by the `steverhoades` / `LocalFileReference` conflict; removing the package unblocks it. JWT 5.x migration touchpoints: `Lcobucci\Clock\Clock` → `Psr\Clock\ClockInterface` in `OidcTokenValidator` + `FakeClock`; `new Builder(...)` → `Builder::new(...)` in `OEIdTokenResponse` (constructor deprecated in 5.5); narrowed claim name to non-empty-string before `withClaim()`; removed tautological asserts now covered by 5.x return types
- **Renamed `IdTokenSMARTResponse` → `OEIdTokenResponse`** and re-parented it onto League's `BearerTokenResponse` directly, inlining the id_token issuance logic (~40 LoC from steverhoades's `IdTokenResponse`). Replaced deprecated `SystemLoggerAwareTrait` with a constructor-injected PSR `LoggerInterface` (default `NullLogger`) implementing `LoggerAwareInterface`.
- **Deleted `IdentityRepository`** — was `@deprecated` with zero in-repo usages (verified via cross-repo grep). Anyone still using it should switch to `UserRepository`.
- **Modernized `UserEntity`** — deleted dead `$userRole` property (never read anywhere) and the unused `$claimsType` / `getClaimsType` / `setClaimsType` cluster with its commented-out client_credentials logic; typed `$identifier` as `?string`; added types to `getIdentifier()` / `setIdentifier()`. Small polish on `ClaimRepository` (`const` → `public const`), `UserRepository` (typed `$fhirBaseUrl`), and `ScopeRepository` (`declare(strict_types=1)`)
- **Breaking changes for integrators:** downstream modules that import `OpenIDConnectServer\*` directly, extend `IdTokenSMARTResponse`, or use the deleted `IdentityRepository` must update — swap imports to the `OpenEMR\Common\Auth\OpenIDConnect\*` namespace, rename `IdTokenSMARTResponse` references to `OEIdTokenResponse`, and switch off `IdentityRepository` to `UserRepository`. No API signature changes beyond the namespace/class rename. Callers of `$responseType->setSystemLogger(...)` should switch to `setLogger(...)` (PSR-3)
- **SMART on FHIR / ONC certification:** id_token wire format is unchanged — the JWT builder has always been ours (`getBuilder()` override), so the port changes code paths, not payload. Existing `OEIdTokenResponseTest` + `SMARTSessionTokenContextIntegrationTest` suites pass

**GCIP Custom Module (`oe-module-gcip-auth/`)**
- Add `GcipAuthHandler` — event listener for `OidcLoginRequestEvent`; validates Firebase ID token, resolves local user via external identity + ACL, sets session, dispatches post-auth event
- Add Firebase JS SDK v9 login form (`gcip-login.html.twig`) with SRI integrity hashes
- Add silent session refresh via `gcip-session-refresh.js` — proactively refreshes Firebase tokens before 1-hour expiry
- Add `GcipConfigService` — module settings CRUD (project ID, API key, auth domain, issuer, client ID, tenant IDs)
- Add `AdminController` — AJAX config management endpoint with CSRF protection
- Add `GcipClaimMapper` — handles Firebase-specific `firebase.tenant` claim structure
- Add `Bootstrap` — wires module into OpenEMR's event system; conditional: zero impact when module is disabled

**Database Schema**
- Add `oidc_external_identity` table (unique keys on `user_id` and `issuer+external_id`)
- Add `oidc_cache` table for `DatabaseCache` PSR-16 backend
- Add `oidc_token_revocation` table for token lifecycle
- Add upgrade SQL (`patch.sql`) for existing installations

**Auth Flow Integration**
- Hook into `auth.inc.php` via `OidcLoginRequestEvent` dispatch — if an OIDC module handles the event, standard password auth is bypassed; otherwise unchanged
- OIDC session refresh endpoint at `library/ajax/oidc_session_refresh.php` (thin HTTP shim over `OidcSessionRefreshHandler`)
- Session metadata stored for issuer/subject pinning on refresh (prevents token substitution)

**Testing (139 files changed, 12087 insertions)**
- `OidcTokenValidatorTest` — 26 isolated tests (signature, claims, algorithm restriction, replay, clock skew)
- `GcipAuthHandlerTest` — 12 isolated tests covering all 9 login branches
- `OidcSessionRefreshHandlerTest` — 6 isolated tests (discovery, validation, issuer/subject pinning)
- `JsonWebKeySetCacheTest` — 4 isolated tests (cache hit/miss, HTTP fallback, read-failure recovery)
- `JsonWebKeySetSigningKeyTest` — 9 isolated tests (kid match, alg mismatch, rotation refetch, non-RSA rejection)
- `GcipConfigServiceTest` — 18 integration tests
- `IpLoginRateLimiterTest` — 14 integration tests
- `AuthUtilsIpDelegationTest` — 4 integration tests
- `DatabaseCacheTest` — integration test for PSR-16 backend
- `IpBlockStatusTest`, `CacheEntryTest`, `ValidatedTokenTest`, `OidcValidationParametersTest`, `NormalizedIdentityTest`, `ExternalIdentityMappingTest`, `MinimalClaimMapperTest`, `StandardClaimMapperTest`, `GcipClaimMapperTest`, `OidcProviderMetadataTest`, `OidcDiscoveryClientTest`, `JsonWebKeyTest`, `JwksClientTest`, `OidcSessionHelperTest`, `OidcLoginCsrfTest`, `OidcLogoutEventTest`, `OidcAuthenticationEventTest`, `OidcLoginRequestEventTest`, `OidcTokenReceivedEventTest`, `FilesystemCacheTest`, `RedisCacheTest`, `ClientCredentialsAssertionGeneratorTest`, `CustomClientCredentialsGrantTest` — isolated value-object, mapper, and event tests

**Security**
- Algorithm restriction: RS256/384/512 only; explicit `none` and symmetric rejection (tested)
- CSRF on both login (pre-auth) and refresh (post-auth) with separate tokens
- Issuer + subject pinning on session refresh (prevents token substitution)
- Per-session refresh rate limiting (30s cooldown, HTTP 429)
- JTI replay protection on all ID-token validation paths
- IP-based login rate limiting with audit logging
- No token content in error responses
- SRI integrity hashes on all Firebase SDK scripts
- Graceful degradation: module disable restores standard login


kojiromike:
@aisle-analyzer do your thing

src/Common/Auth/Oidc/Discovery/OidcDiscoveryClient.php
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Fetch provider metadata, using the cache when available.
+     *
+     * @param string $issuerUrl The provider's issuer URL (e.g. "https://accounts.google.com").
+     */
+    public function getMetadata(string $issuerUrl): OidcProviderMetadata
+    {
+        $cacheKey = self::CACHE_KEY_PREFIX . hash('sha256', $issuerUrl);
+
```
kojiromike, line 46: The cache key is derived from `$issuerUrl` alone, with no component identifying which OpenEMR site the lookup is for. In the multi-site deployment model where several sites share a single MySQL database (see `sites/` and the documented multi-site architecture), this means site A and site B that both authenticate against the same issuer URL will read and write the same `oidc_discovery_<hash>` row in `oidc_cache`. A compromised or misconfigured site can poison the discovery document (and indirectly the cached JWKS) for every other site that shares the DB.

The same concern applies to `RedisCache` and `FilesystemCache` when the Redis instance or filesystem path is shared across sites.

Would it make sense to namespace the cache key by `$_SESSION['site_id']` (or an explicit site identifier passed into the client), so cached discovery documents can't cross sites?

```suggestion
        $cacheKey = self::CACHE_KEY_PREFIX . ($_SESSION['site_id'] ?? 'default') . '_' . hash('sha256', $issuerUrl);
```

(The same prefixing would need to land in `refresh()` and anywhere else a cache key is constructed from tenant-scoped input.)

sql/database.sql
```diff
@@ -14145,6 +14145,38 @@ KEY `accounts_id` (`user_id`),
...
+    `expires_at` DATETIME DEFAULT NULL COMMENT 'NULL = no expiration',
+    PRIMARY KEY (`cache_key`),
+    KEY `idx_expires_at` (`expires_at`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+DROP TABLE IF EXISTS `oidc_token_revocation`;
+CREATE TABLE `oidc_token_revocation` (
+    `jti` VARCHAR(512) NOT NULL COMMENT 'JWT ID claim',
+    `revoked_at` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    `token_expiry` DATETIME NOT NULL COMMENT 'When token would naturally expire',
```
kojiromike, line 14176: `PRIMARY KEY (jti(255))` on a `VARCHAR(512)` column under `utf8mb4` silently truncates the indexed prefix: any two JTIs that agree in their first 255 characters collide at the primary-key level, and `INSERT` on the second one fails with a duplicate-key error. The reverse is worse: because revocation lookups go through this PK, a longer JTI whose first 255 chars match a revoked JTI will appear revoked, and a lookup of a revoked-but-long JTI can miss the row it meant to match.

Most IdPs produce short JTIs, but OIDC does not bound the length (`jti` is just a case-sensitive string unique within the issuer), and Firebase/GCIP have historically emitted longer opaque IDs.

Two options that avoid the silent-truncation trap:

1. Shrink the column and validate at write time: `\`jti\` VARCHAR(255) NOT NULL` with an explicit length check in `TokenRevocationService::revoke()` that rejects oversize JTIs.
2. Store the hash: keep `jti` as-is for audit, add a `jti_hash CHAR(64) NOT NULL` column populated from `hash('sha256', $jti)` and make that the PK.

Option 2 also gives you fixed-width comparisons and a natural fit for the `VARCHAR(512)` tolerance the schema comment implies.

(Same concern applies to `oidc_external_identity.uq_issuer_external_id` on lines 14158 — `(\`issuer\`(255), \`external_id\`(255))` can collide for providers that emit long `sub` values.)

sql/database.sql
```diff
@@ -14145,6 +14145,38 @@ KEY `accounts_id` (`user_id`),
 KEY `clients_id` (`client_id`)
 ) ENGINE=InnoDB;
 
+DROP TABLE IF EXISTS `oidc_external_identity`;
+CREATE TABLE `oidc_external_identity` (
+    `id` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
```
kojiromike, line 14150: The column comment says `'FK to users.id'` but there's no `FOREIGN KEY` constraint, and deliberately so — OpenEMR's schema doesn't use DDL FKs anywhere (a grep of the full `database.sql` returns zero matches). That convention is fine, but it means the *application* has to own the cleanup on user deletion.

I don't see anywhere in the PR that deletes `oidc_external_identity` rows when a user is deleted (or reassigned / merged). In a long-lived installation this creates two problems:

1. **Orphan rows.** `users.id` values can be reused (the table is `AUTO_INCREMENT BIGINT` but rows can be purged and IDs recycled in some flows). A stale `oidc_external_identity` row for an old user would bind an IdP identity to whatever new local user inherits that `user_id` — i.e., unauthenticated access to the wrong account.
2. **Audit drift.** The `UNIQUE KEY uq_user_id` means the orphan also blocks a new legitimate mapping for that user.

Could you add a cleanup step to the user-deletion path (or whichever admin code path deactivates/removes users) that also deletes from `oidc_external_identity`? A unit or integration test pinning "delete user → mapping row is gone" would document the invariant.

library/ajax/oidc_session_refresh.php
```diff
@@ -0,0 +1,129 @@
...
+// 3. Per-session rate limiting
+if (OidcSessionHelper::isRefreshOnCooldown(time())) {
+    http_response_code(429);
+    echo json_encode(['error' => 'too_many_requests']);
+    exit;
+}
+
+// 4. CSRF validation
+if (!CsrfUtils::verifyCsrfToken(filter_input(INPUT_POST, 'csrf_token_form') ?? '', $session)) {
+    $username = is_string($session->get('authUser')) ? $session->get('authUser') : '';
```
kojiromike, line 61: `filter_input()` returns `string|false|null` — see [the PHP manual](https://www.php.net/manual/en/function.filter-input.php). `??` only catches `null`, so when the filter fails (e.g., the client sends `csrf_token_form` as an array instead of a scalar, which is a known way to probe filter-based validation), the expression yields `false`, which is then passed to `CsrfUtils::verifyCsrfToken()` as the *submitted* token. Depending on how `verifyCsrfToken()` normalizes its input, this can either short-circuit the check or compare `false` against the stored token via loose-ish semantics.

Use `?:` to collapse `false`, `null`, and `''` to the same empty-string fallback:

```suggestion
if (!CsrfUtils::verifyCsrfToken(filter_input(INPUT_POST, 'csrf_token_form') ?: '', $session)) {
```

The same pattern shows up in four other places in this PR that I noticed — all worth the same fix:

- `library/auth.inc.php` — `$oidcIdToken = filter_input(INPUT_POST, 'oidc_id_token') ?? '';`
- `library/auth.inc.php` — `$oidcCsrfToken = filter_input(INPUT_POST, 'oidc_csrf_token') ?? '';`
- `interface/modules/custom_modules/oe-module-gcip-auth/src/Controller/AdminController.php` — two instances (`csrf_token`, `action`)

The CSRF-token ones are the ones that matter most from a security standpoint; the others are defense-in-depth.

src/Common/Auth/Oidc/Cache/DatabaseCache.php
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        return $success;
+    }
+
+    /**
+     * Remove all expired entries. Intended to be called periodically (cron or on-demand).
+     */
+    public function purgeExpired(): void
+    {
```
kojiromike, line 162: `DatabaseCache::purgeExpired()` (here), `RedisCache::purgeExpired()`, and `TokenRevocationService::purgeExpired()` are all defined and tested, but I can't find anywhere in the PR that actually invokes them outside of tests — no cron entry, no background-service registration in `background_services`, no hook into any admin-initiated cleanup task.

The practical consequence for a long-running install:

- `oidc_cache` grows with every discovery refresh and every JWKS rotation. Expired rows are read-lazy-purged in `DatabaseCache::get()` (good), but rows that are never looked up again (e.g., keys from rotated-out issuers) stay forever.
- `oidc_token_revocation` grows with every revocation. Those entries only need to live until `token_expiry`, but nothing removes them after that. In an environment with short-lived tokens and active revocation, this table will dominate the OIDC footprint within weeks.

Could these be wired into `background_services` (or whatever scheduling mechanism the rest of the module uses), with a default interval of a few hours? At minimum, a note in the module README that operators must call these manually would be better than nothing, but I'd prefer a default-on cleanup job.

kojiromike (commented):
Thanks for the sustained work on this, @zmilan. The OIDC pieces are well-structured and the test coverage on the isolated units is solid.

The AISLE-bot pass (#1–#11 above) and Copilot's `ExternalIdentityRepository::save()` note already surface most of the security-critical items I'd have flagged, so those aren't repeated here. The inline comments below are concerns I don't think have been raised yet — mostly schema / multi-site / lifecycle issues that aren't visible from a straight token-validation review.

A few things that fit better as body notes than inline comments:

**Scope.** This PR bundles a handful of independently-shippable changes:

1. OIDC core (discovery, validator, session refresh handler, event dispatcher)
2. PSR-16 cache abstraction with three backends
3. GCIP custom module under `interface/modules/custom_modules/oe-module-gcip-auth/`
4. `library/ajax/login_counter_ip_tracker.php` hardening (the `(int) $_POST[...]` rewrites around L3781–3817)

Item 4 is unrelated to OIDC and could land as a separate PR immediately — it would also remove ~40 lines from this diff and make the scope more obviously "OIDC". Items 1–3 are more tightly coupled but would still be easier to review and revert independently. The cache abstraction in particular has three backends, each with their own semantics, and could stand on its own.

**"Graceful degradation" wording.** The PR description and module docs describe the module as degrading gracefully when disabled. That's not quite what happens: users who were provisioned via OIDC and do not have a usable local password are locked out when the module is off. That is the correct fail-closed posture — but please call it that in the docs rather than "graceful", since an operator reading "graceful degradation" may assume OIDC users can still log in by some other means.

**Dedup note.** Several AISLE findings overlap with things I'd also have raised (missing `exp` enforcement, revocation list not wired into the validator, server-side tenant allowlist, SSRF on `jwks_uri` / issuer discovery, etc.). I've deferred to those comments rather than restate them — but they all still stand as blocking from my perspective.

zmilan:
@kojiromike about this part: `"Graceful degradation" wording. The PR description and module docs describe the module as degrading gracefully when disabled. That's not quite what happens: users who were provisioned via OIDC and do not have a usable local password are locked out when the module is off. That is the correct fail-closed posture — but please call it that in the docs rather than "graceful", since an operator reading "graceful degradation" may assume OIDC users can still log in by some other means."`  - we are not provisioning users via OIDC, all users needs to be set in OE and just connected to OIDC with a record in `oidc_external_identity` database table... so, `"users who were provisioned via OIDC and do not have a usable local password are locked out when the module is off"` is not something that could happen.

zmilan:
@kojiromike, just to clarify, should I split this PR into smaller ones?

kojiromike:
> @kojiromike, just to clarify, should I split this PR into smaller ones?

Yes, if it's reasonable to do so.

> we are not provisioning users via OIDC, all users needs to be set in OE and just connected to OIDC with a record in oidc_external_identity database table...

Does this mean that a user can always log in with their username and password, even if OIDC is set up?

zmilan:
> > @kojiromike, 
> 
> Does this mean that a user can always log in with their username and password, even if OIDC is set up?

No, once when OIDC/GCIP is activated, it takes over login/authentication (using Symphony events). So, basically, only one of them can be active and handling authentication at the time. But it is intentionally designed that way that OIDC can be switched on/off per request (GCIP is down, or something else, only what needs to be done is turn it off, and the OE authentication mechanism is back).
But, also, we intentionally decided not to provision new users using OIDC/GCIP, at least not for now. But instead, just connect them to already provisioned users in OE.

kojiromike:
> basically, only one of them can be active and handling authentication at the time. …also, we intentionally decided not to provision new users using OIDC/GCIP, at least not for now.

That is fine for this scope. I just wanted to think through the overall operational gaps to roll out OIDC effectively, especially at scale.

kojiromike:
Correction on my review above re: item 4 in the scope-split list (`login_counter_ip_tracker.php`):

I described it as "hardening" around some `(int) $_POST[...]` casts, but re-reading the diff against master, those casts already exist on both the `empty()` guards and the call sites. This PR's only change in that file is swapping `AuthUtils::*` static calls to `IpLoginRateLimiter::*` on the five admin actions.

So the accurate framing for a split PR is **service swap / refactor**, not hardening — no additional changes needed beyond what's already in the diff. The line numbers I cited were positions in the aggregated PR diff rather than file lines, which contributed to the confusion. Sorry for the noise.

Issue #10670, opened 2026-02-09
Meta: PHPStan Rules & Baseline Update


Firehed:
As of #10444, OpenEMR now runs PHPStan at **level 10** (the strictest level). We went from ~20k to ~160k errors being tracked in the baseline[^bl] files. This doesn't mean there are _new_ errors, but that more existing errors (and warnings) are now _known_.

This issue serves as a reference for contributors and for PRs that reduce the baseline.

## What This Means for Contributors

- **New code** must pass level 10 analysis
- **Existing code** is unaffected (for now)
- Refactors that "trade" errors are generally welcome; preferably those that have a net reduction of errors

In the next several days, expect some merge conflicts to arise on the baseline files. We expect this to settle down quickly as more PRs have incorporated the rule change. To quickly resolve them, run:

```bash
git checkout master .phpstan/baseline && composer phpstan-baseline
```

This won't fix the errors, but should unstick conflicts.

### Open PRs

If you have a PR that was opened prior to this change landing, please merge `master` and run PHPStan to check for new errors. We don't want to disrupt existing work as a result and will be pretty forgiving about error detection. Try to fix anything obviously bad, but some of the pickier errors (especially changes relying on the old globals paths) will be forgiven :)

### Writing New Code

New or modified code must be fully type-safe. This includes:

- **Type hints on parameters and return types** — PHPStan will reject `function foo($bar)` in new code
- **Consistent types** — If a function returns `?string`, don't return `false`
- **Null safety** — Check for null before calling methods on nullable types (however, avoid nullable types when possible)
- **Array shapes** — Use `@param array{key: type}` for complex arrays when needed. Intermediate objects ("DTO"s) may also be used.

### Running PHPStan Locally

Before pushing, run:

```bash
composer phpstan
```

This uses the same configuration as CI.

This may take a few minutes the first time you run it. Subsequent runs should be faster since PHPStan has an internal cache.

## Fixing PHPStan Errors

If you see a new error in code you're modifying:

1. Fix the type issue — Add proper type hints, handle null cases, etc.
2. If legacy code prevents a clean fix, consider whether fixing the surrounding code is in scope

Do not manually edit baseline files. When fixing issues, regenerate the baseline with:

```
composer phpstan-baseline
```

# We need your help!

PRs that reduce the baseline by reducing errors are enthusiastically welcomed! Feel free to reference this issue when opening PRs if there's not something more specific. 

The baseline is organized into files by error type in .phpstan/baseline/. 

[...]

PR #11421, opened 2026-04-04
chore(billing): remove unused BillingExport external billing hook

### LINKED ISSUE #11367: chore: Clean up BillingExport
While reviewing #11363, I realized just how unconventional this is:

- There's a comment in the file header that basically says "to use this, rename it"
- It appears to require further customization to work, even after being renamed
- The post-renaming _presence of the file_ is used to conditionally render HTML fragments and change javascript variables
- The class that uses it (`GeneratorExternal`) has a high likelihood of crashing since it tries to use the class regardless of whether it's included

I don't have enough context on the file to know what makes sense here. It appears to be >20 years old based on the header. If nobody knows of it being actively used, it seems better to remove it outright.  I'd like some input here!

mw2000:

Removes the unused `BillingExport` external billing integration. Closes #11367.

`custom/BillingExport.csv.php` was an opt-in integration point that allowed practices with external billing systems to receive claim exports in a custom CSV format. Activating it required manually renaming the file on disk. As discussed in #11367, X12 837P generation now serves this need for modern users, and the feature appears to have no active users in the community.

- Delete `custom/BillingExport.csv.php` — the template file itself
- Delete `src/Billing/BillingProcessor/Tasks/GeneratorExternal.php` — the task that depended on it
- Remove `bn_external` dispatch branch from `BillingProcessor.php`
- Remove `$EXPORT_INC` variable, all `file_exists($EXPORT_INC)` conditionals, the Export Billing button, and the `SubmitTheScreenExportOFX()` JS function from `interface/billing/billing_report.php`
- Remove `file_exists` conditional around the Export OFX link in `interface/reports/criteria.tab.php`; the View Log link is now unconditionally visible
- Clean up 13 PHPStan baseline files — removed all stale entries pointing to the deleted files


kojiromike:
PHPStan is still failing — the baseline files need to be regenerated rather than hand-edited. See my comment on #11422 for the fix, and #10670 for context on how the baseline works.

```bash
composer install          # ensure dependencies are fresh
composer phpstan-baseline # regenerate all baseline files
```

mw2000:
@kojiromike ty for the review on all 3 PRs 😄 

regenerating with `composer phpstan` fixed the CI failures on the PRs

src/RestControllers/AuthorizationController.php
```diff
@@ -327,22 +327,31 @@ public function clientRegistration(HttpRestRequest $request): ResponseInterface
...
+                        $params[$key] = implode('|', $stringValues);
                     } else if (in_array($key, ['dsi_source_attributes'])) {
                         $params[$key] = $data->all($key);
                     } elseif ($key === 'jwks') {
                         $jwks = $data->all('jwks');
-                        if (is_string($jwks)) {
-                            $jwks = json_decode($jwks, true, 512, JSON_THROW_ON_ERROR);
-                        }
-                        $params[$key] = json_encode($jwks);
+                        $params[$key] = json_encode($jwks, JSON_THROW_ON_ERROR);
```
kojiromike, line 343: This is a behavior change worth calling out explicitly, separate from the scope concern that the OAuth diff doesn't belong in this PR at all.

The previous code handled `$jwks` arriving as a JSON string:

```php
if (is_string($jwks)) {
    $jwks = json_decode($jwks, true, 512, JSON_THROW_ON_ERROR);
}
$params[$key] = json_encode($jwks);
```

The new code passes whatever `$data->all('jwks')` returns straight to `json_encode`. If a client submits `jwks` as a JSON-encoded string (e.g. `"jwks": "{\"keys\":[...]}"`), the round trip now produces a JSON-encoded quoted string rather than a structured object — silently storing a corrupt value. The new `testClientRegistrationEncodesJwksArrayBeforeRedirectValidation` test only exercises the array case.

Is the intent to no longer accept string-form `jwks`? If so, that's a reasonable narrowing but it should be a deliberate, documented rejection — `throw new OAuthServerException("jwks must be an object", 0, 'invalid_client_metadata')` when `is_string($jwks)` — rather than a silent acceptance that stores garbage. If not, the `is_string` decode branch needs to stay.

src/RestControllers/AuthorizationController.php
```diff
@@ -327,22 +327,31 @@ public function clientRegistration(HttpRestRequest $request): ResponseInterface
 
             foreach ($keys as $key => $supported_values) {
                 if ($data->has($key)) {
-                    if (in_array($key, ['contacts', 'redirect_uris', 'request_uris', 'post_logout_redirect_uris', 'grant_types', 'response_types', 'default_acr_values'])) {
-                        $params[$key] = implode('|', $data->all($key));
+                    if (in_array($key, ['contacts', 'redirect_uris', 'request_uris', 'post_logout_redirect_uris', 'grant_types', 'response_types', 'default_acr_values'], true)) {
```
kojiromike, line 330: Strict `in_array` here and at line 352 is a real improvement and the right call. Worth landing — just not in a PR titled "remove unused BillingExport hook." Splitting the OAuth changes into their own PR would also let them get the OAuth-specific review attention they deserve (and let this cleanup land independently).

kojiromike (changes requested):
The BillingExport deletion itself looks clean — `custom/BillingExport.csv.php` and `GeneratorExternal.php` are gone, the `bn_external` dispatch branch and the `$EXPORT_INC` / `file_exists` toggles in `billing_report.php` and `criteria.tab.php` are removed consistently, the dead `SubmitTheScreenExportOFX()` JS is gone, and the baseline files were regenerated rather than hand-edited. Negative coverage (no remaining call sites) is the appropriate "test" here.

Two concerns:

1. **The PR also contains unrelated OAuth changes.** `src/RestControllers/AuthorizationController.php` and `tests/Tests/RestControllers/Authorization/AuthorizationControllerTest.php` are substantive edits to SMART client-registration metadata validation (strict `in_array`, type narrowing instead of implicit casts, dropping the `is_string($jwks)` decode branch, body rewinds, three new tests). These are defensible improvements but they sit on an OAuth trust boundary and have nothing to do with removing BillingExport. They should be split into their own PR — the issue link (#11367) doesn't cover them, and a reviewer looking at this PR by its title will not give the OAuth diff the attention it needs. Inline comment on the most consequential line below.

2. **`.composer-require-checker.json` orphan.** The symbol whitelist still has a `"BillingExport"` entry pointing at the now-deleted file. Worth removing in the same cleanup so the whitelist stays accurate.

Test coverage: the BillingExport removal is appropriately tested by negative search (no call sites remain — verified). The OAuth changes have three new tests, but they cover only the new strict-validation paths, not the behavioral change to `jwks` handling (see inline comment).

PR #11422, opened 2026-04-04
refactor(billing): replace exec(unzip) with ZipArchive for ERA file uploads

### LINKED ISSUE #11413: refactor: Replace exec('unzip') with ZipArchive
## Summary

The codebase uses exec("unzip -p ...") to extract zip files in a couple of places:

- interface/billing/era_payments.php:126
- interface/billing/sl_eob_search.php:968

Both have the comment "Probably won't work on Windows" which confirms the portability issue.

## Proposal

Replace these with PHP's native ZipArchive class, which would:
- Work cross-platform (including Windows)
- Remove the external unzip dependency
- Avoid shell execution security considerations

## Considerations

ZipArchive requires the php-zip extension, which may not be installed by default on all systems. This should be verified as a required extension before making the change.

---

_This issue was written with AI assistance and reviewed by a human._

mw2000:

Closes #11413.

Replaces two `exec("unzip -p ...")` shell calls used to extract ZIP-compressed ERA file uploads with PHP's native `ZipArchive::getFromIndex(0)`. Both call sites had the comment _"Probably won't work on Windows"_ — this resolves that portability gap without any change in behavior on Linux/macOS.

- `interface/billing/era_payments.php` — replace `exec("unzip -p ...")` + `rename`/`unlink` dance with `ZipArchive::open()` + `getFromIndex(0)` + `file_put_contents()`; add error messages to `$alertmsg` on open/read failure
- `interface/billing/sl_eob_search.php` — same replacement at the equivalent upload handler

`ext-zip` is already a declared required dependency in `composer.json` and `ZipArchive` is used widely throughout the codebase, so no new requirements are introduced.

### DIFF
--- a/interface/billing/era_payments.php
+++ b/interface/billing/era_payments.php
@@ -120,6 +120,7 @@ elseif (!empty($_FILES['form_erafile']['size'])) {
     if (!is_string($tmp_name)) {
         $alertmsg .= xl("Invalid file upload") . " ";
     } else {
+        $shouldParseEra = true;
         // Handle .zip extension if present.
         if (strtolower(substr((string) $_FILES['form_erafile']['name'], -4)) == '.zip') {
             $zip = new ZipArchive();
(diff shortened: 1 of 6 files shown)


interface/billing/era_payments.php
```diff
@@ -120,13 +120,26 @@ function era_payments_callback(array &$out): void
...
+                    $alertmsg .= xl("Unable to read file from ZIP archive") . " ";
+                    $contents = '';
+                }
+
+                file_put_contents($tmp_name, $contents);
+            } else {
+                $alertmsg .= xl("Unable to open ZIP archive") . " ";
+            }
+        }
+        $parseResult = ParseERA::parseERA($tmp_name, 'era_payments_callback');
```
kojiromike: When ZIP open or read fails above, `$alertmsg` is set but execution falls through here — `parseERA()` will try to parse the original ZIP binary (or empty string) as ERA content. Would it make sense to guard this with something like `if (!$alertmsg)` or skip the parse on failure? The same pattern exists in `sl_eob_search.php`.

(This is pre-existing behavior — the old `exec` path had the same issue — but since you're already improving error handling here, it would be nice to close it.)

kojiromike (commented):
The `exec(unzip)` to `ZipArchive` swap is clean and `ext-zip` is already a declared dependency. One concern about error handling — see inline.

kojiromike:
PHPStan is failing because the baseline files are out of sync with the code changes. When you add, remove, or modify code that has PHPStan baseline entries, the baseline needs to be regenerated — don't edit the baseline files by hand.

To fix:

```bash
composer install          # ensure dependencies are fresh
composer phpstan-baseline # regenerate all baseline files
```

Then commit the updated baseline files. This will remove stale entries for deleted code and adjust counts for modified code automatically.

The same applies to #11421 — both PRs need a `composer phpstan-baseline` run after rebasing onto current master. See #10670 for more context on how the baseline works.

kojiromike:
These copilot findings aren't blockers. I created followup issues for the important ones. Merging.

PR #11423, opened 2026-04-04
refactor(background-services): distinguish lock failure reasons in BackgroundServiceRunner

### LINKED ISSUE #11350: refactor(background-services): distinguish lock failure reasons in BackgroundServiceRunner
Currently `BackgroundServiceRunner::run()` returns a single `locked` status when `acquireLock()` returns false, but the underlying SQL can fail for multiple reasons: the service is already running (another process holds the lock) vs the service is not yet due (interval hasn't elapsed). Distinguishing these (e.g. `already_running` vs `not_due`) would give CLI and API callers better diagnostic output.

Originated from Copilot review on #11334.

mw2000:

Closes #11350.

`BackgroundServiceRunner::run()` previously emitted a single `'locked'` status when `acquireLock()` returned false, regardless of whether the cause was another process holding the DB lock or the service interval not yet having elapsed. This change distinguishes the two cases so CLI and API callers get actionable diagnostic output.

**`BackgroundServiceRunner`**
- Change `acquireLock()` signature from `bool` to `?string` — returns `null` on success (lock acquired), or a reason string on failure
- When the `UPDATE` affects 0 rows, a follow-up `SELECT` on `running` determines which condition blocked the lock:
  - `'already_running'` — `running = 1`, another process holds the lock
  - `'not_due'` — interval has not elapsed (`NOW() <= next_run`)
- Update `run()` docblock to document all possible status values

**`BackgroundServicesCommand`**
- Replace the single `'locked'` match arm with distinct messages for `'already_running'` and `'not_due'`
- Both map to exit code `2` (unchanged behaviour)

**`BackgroundServiceRestController`**
- Replace `'locked'` in the HTTP status match with `'already_running'` and `'not_due'`
- Both continue to return `HTTP 409 Conflict` (unchanged behaviour)

**`BackgroundServiceRunnerTest`**
- Update the stub's `acquireLock()` from `bool` to `?string`
- Replace the single `testRunReturnsLockedWhenLockFails` test with two separate tests: `testRunReturnsAlreadyRunningWhenLockFailsDueToRunningProcess` and `testRunReturnsNotDueWhenLockFailsDueToInterval`

### DIFF
--- a/src/Services/Background/BackgroundServiceRunner.php
+++ b/src/Services/Background/BackgroundServiceRunner.php
@@ -37,6 +37,13 @@ class BackgroundServiceRunner
      * @param string|null $serviceName Specific service name, or null for all
      * @param bool $force Bypass interval check
      * @return list<array{name: string, status: string}> Results per service
+     *   Possible status values:
+     *   - 'executed'        — service ran successfully
+     *   - 'skipped'         — inactive, already running (in-memory check), or manual-mode without --force
+     *   - 'already_running' — another process holds the DB lock (running = 1)
+     *   - 'not_due'         — interval has not elapsed yet (NOW() <= next_run)
+     *   - 'error'           — exception during lock acquisition or execution
+     *   - 'not_found'       — requested service name does not exist
      */
     public function run(?string $serviceName = null, bool $force = false): array
     {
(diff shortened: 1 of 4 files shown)


src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -137,16 +144,44 @@ protected function getServices(?string $serviceName, bool $force): array
...
+            'SELECT running FROM background_services WHERE name = ?',
+            [$service['name']],
+            false,
+        );
+
+        if ($row === false) {
+            return 'error';
+        }
+
+        if (($row['running'] ?? '0') == 1) {
```
kojiromike: Minor: the `running` column comes back from MySQL as a string, so this loose comparison (`'0' == 1` is false, `'1' == 1` is true) happens to work — but would a strict check be safer here?

```suggestion
        if (($row['running'] ?? '0') === '1') {
```

kojiromike (approved):
Clean split of the ambiguous `'locked'` status into actionable `'already_running'` and `'not_due'` — the docblock, command output, and REST controller all handle both new values consistently. One minor nit inline.

PR #11427, opened 2026-04-05
refactor(calendar): replace hardcoded day/month arrays with int-backed enums

### LINKED ISSUE #10074: Implement `literal-string` type for xl and other translation functions
phpstan supports a [literal-string](https://phpstan.org/writing-php-code/phpdoc-types#other-advanced-string-types) type that we can use to find where functions like xl are called with variables. However, it won't work without special effort, or until we get phpstan to level 8, so I'm just opening this issue as a reminder. We can always add the type hint in the comment, and it'll show up if/when we get phpstan to the level.

kojiromike:
## Summary

- Introduce `DayOfWeek` and `Month` int-backed enums in `src/Common/Calendar/` with `label()` methods using exhaustive `match` expressions calling `xl()` on literal strings
- Replace all hardcoded day-of-week and month name arrays across `date_functions.php`, `appointments.inc.php`, `globals.inc.php`, `add_edit_event.php`, and the PostCalendar language file with enum usage
- PHPStan sees each `match` arm as `literal-string`, eliminating the `mixed`-type violations from array-indexed `xl()` calls
- Removes 155 stale `constant.notFound` baseline entries for the old `_CAL*` defines

Closes #10074 (partially — adds literal-string enforcement for day/month translations)

## Test plan

- [ ] Verify date formatting displays correctly in English and at least one non-English language
- [ ] Check calendar event creation/editing shows translated day names
- [ ] Verify appointment repeat-on-day selector shows translated labels
- [ ] Run `composer phpstan` — should pass clean

### DIFF
--- a/src/Common/Calendar/DayOfWeek.php
+++ b/src/Common/Calendar/DayOfWeek.php
@@ -1,0 +1,45 @@
+<?php
+
+/**
+ * Day-of-week enum with translatable labels.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Michael A. Smith <michael@opencoreemr.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Common\Calendar;
+
+enum DayOfWeek: int
+{
+    case Sunday = 0;
+    case Monday = 1;
+    case Tuesday = 2;
+    case Wednesday = 3;
+    case Thursday = 4;
+    case Friday = 5;
+    case Saturday = 6;
+
+    /**
+     * Return the translated day name.
+     *
+     * Call xl() on each literal arm so translation-key extraction tools
+     * (e.g. xgettext, grep) can find every key statically.
+     */
+    public function label(): string
+    {
+        return match ($this) {
+            self::Sunday => xl('Sunday'),
+            self::Monday => xl('Monday'),
+            self::Tuesday => xl('Tuesday'),
+            self::Wednesday => xl('Wednesday'),
+            self::Thursday => xl('Thursday'),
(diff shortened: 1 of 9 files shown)


src/Common/Calendar/DayOfWeek.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        return xl(match ($this) {
+            self::Sunday => 'Sunday',
+            self::Monday => 'Monday',
+            self::Tuesday => 'Tuesday',
+            self::Wednesday => 'Wednesday',
+            self::Thursday => 'Thursday',
+            self::Friday => 'Friday',
+            self::Saturday => 'Saturday',
+        });
```
kojiromike: call `xl` on literal strings, and add a comment to the code explaining why.

```suggestion
        return match ($this) {
            self::Sunday => xl('Sunday'),
            self::Monday => xl('Monday'),
            self::Tuesday => xl('Tuesday'),
            self::Wednesday => xl('Wednesday'),
            self::Thursday => xl('Thursday'),
            self::Friday => xl('Friday'),
            self::Saturday => xl('Saturday'),
        };
```

src/Common/Calendar/Month.php
```diff
@@ -0,0 +1,49 @@
...
+            self::April => 'April',
+            self::May => 'May',
+            self::June => 'June',
+            self::July => 'July',
+            self::August => 'August',
+            self::September => 'September',
+            self::October => 'October',
+            self::November => 'November',
+            self::December => 'December',
+        });
```
kojiromike: call xl on each line on the literal strings, and add a code comment explaining why.

src/Common/Calendar/DayOfWeek.php
kojiromike: Fixed — moved `xl()` onto each literal arm and added a docblock explaining why (translation-key extraction tools need to see each key statically).

src/Common/Calendar/Month.php
kojiromike: Fixed — same change as DayOfWeek, with the same docblock explaining the rationale.

PR #11430, opened 2026-04-05
fix(config): stop suppressing E_USER_WARNING and E_USER_DEPRECATED

### LINKED ISSUE #11408: globals.php unconditionally suppresses E_USER_WARNING and E_USER_DEPRECATED
`interface/globals.php` line 826:

```php
```

This runs on every request and strips `E_USER_WARNING` and `E_USER_DEPRECATED` from the error reporting level unconditionally. This silences warnings that OpenEMR's own code intentionally triggers — Smarty plugin parameter validation (`{xlj}`, `{xlt}`, `{xla}` missing `t` param), clickmap API violations, FHIR parser issues, and module deprecation notices.

These are the application telling operators something is wrong, and they're being suppressed before they can reach the error log. In deployments where `display_errors = Off` (i.e., any production environment), these errors simply vanish.

The `user_debug > 1` block below it makes this worse — it additionally suppresses `E_WARNING` and `E_NOTICE` while enabling `display_errors`, which seems inverted from the intent (show more errors to the developer, but actually report fewer).

Suggested fix: remove line 826 entirely, or at minimum gate it behind a configuration flag so operators can opt in to full error reporting.

haoyu-haoyu:
## Summary

Remove the unconditional `error_reporting()` mask in `interface/globals.php` that silences `E_USER_WARNING` and `E_USER_DEPRECATED` on every HTTP request, and replace it with an explicit `error_reporting(E_ALL)`.

## Background

Line 792 was introduced in **PR #4459 (June 2021)** with the commit message *"Temporary fix for deprecated warnings turning into exceptions"* as a workaround for PHP 8.0. Five years later, it still silences all application-level warnings.

## What gets silenced today

| Component | Warning | Impact |
|-----------|---------|--------|
| Smarty `{xlt}`, `{xla}`, `{xlj}` | Missing `t` parameter | Silent rendering failures |
| FHIR Response Parser | Unmapped property warnings | Silent data loss |
| LBF Validation | Invalid JSON config | Silent form validation bypass |
| `xl()` translation | Function not defined | Silent missing translations |

In production (`display_errors = Off`), none of these reach the error log.

## Changes

```diff
-error_reporting(error_reporting() & ~E_USER_DEPRECATED & ~E_USER_WARNING);
-// user debug mode
+// Report all errors so nothing is silently suppressed.
+error_reporting(E_ALL);
+// user debug mode — controls display_errors
 if ($globalsBag->getInt('user_debug', 0) > 1) {
-    error_reporting(error_reporting() & ~E_WARNING & ~E_NOTICE & ~E_USER_WARNING & ~E_USER_DEPRECATED);
     ini_set('display_errors', 1);
 }
```

The error reporting level is now explicitly `E_ALL`, so all warnings (including `E_USER_WARNING` and `E_USER_DEPRECATED`) are logged. The `user_debug` block is simplified to only control `display_errors`.

Fixes #11408

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -789,8 +789,9 @@ $globalsBag->set('groupname', $groupname);
 // Override temporary_files_dir
 $globalsBag->set('temporary_files_dir', rtrim(sys_get_temp_dir(), '/'));
 
-// user debug mode — only controls display_errors; the error reporting
-// level is left to php.ini so administrators can decide what to log.
+// Report all errors so nothing is silently suppressed.
+error_reporting(E_ALL);
+// user debug mode — controls display_errors
 if ($globalsBag->getInt('user_debug', 0) > 1) {
     ini_set('display_errors', 1);
 }
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -789,10 +789,9 @@ $globalsBag->set('groupname', $groupname);
 // Override temporary_files_dir
 $globalsBag->set('temporary_files_dir', rtrim(sys_get_temp_dir(), '/'));
 
-error_reporting(error_reporting() & ~E_USER_DEPRECATED & ~E_USER_WARNING);
-// user debug mode
+// user debug mode — only controls display_errors; the error reporting
+// level is left to php.ini so administrators can decide what to log.
 if ($globalsBag->getInt('user_debug', 0) > 1) {
-    error_reporting(error_reporting() & ~E_WARNING & ~E_NOTICE & ~E_USER_WARNING & ~E_USER_DEPRECATED);
     ini_set('display_errors', 1);
 }
 


Firehed:
I think we ought to explicitly set `E_ALL` rather than defer to `php.ini`... but I could be persuaded otherwise. I want _at minimum_ to ensure that CI runs at E_ALL (the new front-controller and bootstrap initializes this way, but it's not active/integrated yet) so there's a way to forcibly surface the issues currently being suppressed.

haoyu-haoyu:
Makes sense, explicitly setting E_ALL is better than hoping php.ini has the right value. I'll update the PR to use `error_reporting(E_ALL)` instead of removing the call entirely.

kojiromike (commented):
Code change looks correct. Forcing `E_ALL` is the right call per Firehed's feedback -- it guarantees nothing is silently swallowed regardless of the server's php.ini.

One housekeeping item: the PR description is now stale. It still says "The error reporting level is now fully controlled by `php.ini`, where administrators can configure it per environment" and shows a diff snippet that omits the `error_reporting(E_ALL)` line. Since the code now explicitly sets `E_ALL` (the opposite of deferring to php.ini), the description should be updated to match.

Issue #11391, opened 2026-04-02
Remove dead $payment_key code in portal_payment.php


kojiromike:
`$payment_key` is computed in `portal/portal_payment.php` (around the receipt-rendering block) but never used anywhere in the file. It should be removed to reduce confusion.

PR #11437, opened 2026-04-06
fix(portal): migrate $_POST/$_REQUEST to filter_input() in portal_payment.php

### LINKED ISSUE #11392: Migrate remaining $_POST reads in portal_payment.php to filter_input()
`portal/portal_payment.php` still reads several form fields directly from `$_POST` (e.g., `form_save`, `form_method`, `form_source`, `form_upay`, `form_prepayment`). These should be converted to `filter_input(INPUT_POST, ...)` with appropriate validation filters to comply with the `ForbiddenRequestGlobals` PHPStan rule and reduce the baseline.

### LINKED ISSUE #11393: Validate form_prepayment as float in portal_payment.php
`$_POST['form_prepayment']` in `portal/portal_payment.php` is used in SQL inserts and `frontPayment()` without numeric validation. It should be read via `filter_input(INPUT_POST, 'form_prepayment', FILTER_VALIDATE_FLOAT)` with a fallback, and invalid values should be rejected before reaching the database.

craigrallen:
## Summary

Addresses **#11392** and **#11393**. (#11391 was already resolved by #11396.)

### Changes

- All `$_POST` and `$_REQUEST` accesses in `portal/portal_payment.php` replaced with `filter_input()` calls
- `form_prepayment` validated as `FILTER_VALIDATE_FLOAT` with a non-negative guard — invalid or negative values fall back to `0.0` instead of reaching SQL or `frontPayment()`
- String fields (`form_method`, `form_source`, `radio_type_of_payment`) sanitized with `FILTER_SANITIZE_SPECIAL_CHARS`
- `form_upay` read with `FILTER_REQUIRE_ARRAY`
- GET params (`pid`, `hidden_patient_code`, `recid`, `time`, `patient`) also migrated
- 3 entries removed from `.phpstan/baseline/openemr.forbiddenRequestGlobals.php`

### Tests

`tests/Tests/Unit/Portal/PortalPaymentInputTest.php` — 14 unit tests covering:
- Prepayment validation (positive, zero, negative, non-numeric, null, false, data-provider cases)
- `filter_input()` behavior contracts (XSS sanitization, int validation)

Closes #11392
Closes #11393


kojiromike:

One CI failure on this PR. Portal hardening, not urgent — verify CI before backporting.

portal/portal_payment.php
```diff
@@ -125,14 +134,15 @@
...
+            [0, $form_pid, $session->get('authUserID'), 0, $form_source, $form_prepayment, $NameNew, $form_method]
         );
 
-        frontPayment($form_pid, 0, $form_method, $form_source, $_REQUEST['form_prepayment'], 0, $timestamp);//insertion to 'payments' table.
+        frontPayment($form_pid, 0, $form_method, $form_source, $form_prepayment, 0, $timestamp);//insertion to 'payments' table.
     }
 
-    if ($_POST['form_upay'] && $_REQUEST['radio_type_of_payment'] != 'pre_payment') {
-        foreach ($_POST['form_upay'] as $enc => $payment) {
+    $form_upay = filter_input(INPUT_POST, 'form_upay', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY);
```
kojiromike: `FILTER_DEFAULT` leaves each `form_upay` entry as an unfiltered string, and they're cast to `float` on the next iteration. Validating each entry as a float at the boundary would catch malformed values without relying on `(float)$payment` silently coercing garbage to 0.0:

```suggestion
    $form_upay = filter_input(INPUT_POST, 'form_upay', FILTER_VALIDATE_FLOAT, ['flags' => FILTER_REQUIRE_ARRAY]);
```

(The truthy guard on the next line continues to work — invalid entries become `false`, and `if ($amount = (float)$payment)` filters them out.)

portal/portal_payment.php
```diff
@@ -103,14 +104,22 @@
 $alertmsg = ''; // anything here pops up in an alert box
 
 // If the Save button was clicked...
-if ($_POST['form_save'] ?? '') {
-    $form_pid = $isPortal ? $pid : $_POST['form_pid'];
-    $form_method = trim((string) $_POST['form_method']);
-    $form_source = trim((string) $_POST['form_source']);
+if (filter_input(INPUT_POST, 'form_save')) {
+    $form_save_pid = filter_input(INPUT_POST, 'form_pid', FILTER_SANITIZE_SPECIAL_CHARS);
+    $form_pid = $isPortal ? $pid : (int) ($form_save_pid ?? 0);
```
kojiromike, line 109: `form_pid` flows directly into `(int)` two lines later, so `FILTER_SANITIZE_SPECIAL_CHARS` is doing no work here — `FILTER_VALIDATE_INT` matches the actual use:

```suggestion
    $form_pid = $isPortal ? $pid : (int) (filter_input(INPUT_POST, 'form_pid', FILTER_VALIDATE_INT) ?? 0);
```

Then drop the `$form_save_pid` line above.

portal/portal_payment.php
```diff
@@ -103,14 +104,22 @@
...
+    $form_method = trim((string) filter_input(INPUT_POST, 'form_method', FILTER_SANITIZE_SPECIAL_CHARS));
+    $form_source = trim((string) filter_input(INPUT_POST, 'form_source', FILTER_SANITIZE_SPECIAL_CHARS));
+    // Validate form_prepayment as a non-negative float; reject invalid values.
+    $form_prepayment_raw = filter_input(INPUT_POST, 'form_prepayment', FILTER_VALIDATE_FLOAT);
+    if ($form_prepayment_raw === false || $form_prepayment_raw === null || $form_prepayment_raw < 0.0) {
+        $form_prepayment_raw = 0.0;
+    }
+    $form_prepayment = $form_prepayment_raw;
+    $radio_type_of_payment = filter_input(INPUT_POST, 'radio_type_of_payment', FILTER_SANITIZE_SPECIAL_CHARS) ?? '';
```
kojiromike, line 118: `$radio_type_of_payment` is then compared with `===` against a closed set (`'pre_payment'`, `'copay'`, `'invoice_balance'`, `'cash'`). Sanitizing HTML entities on a value that's only used for equality against known strings is the wrong filter — any character mangled by `FILTER_SANITIZE_SPECIAL_CHARS` would just cause the `===` to silently fall through every branch. Either use `FILTER_DEFAULT` and check membership with `in_array($radio_type_of_payment, ['pre_payment', 'copay', 'invoice_balance', 'cash'], true)`, or `FILTER_VALIDATE_REGEXP` against the same set, falling back to `''` so unknown inputs are explicitly rejected.

portal/portal_payment.php
kojiromike, line 143: `FILTER_DEFAULT` leaves each `form_upay` entry as an unfiltered string, and they're cast to `float` on the next iteration. Validating each entry as a float at the boundary would catch malformed values without relying on `(float)$payment` silently coercing garbage to 0.0:

```suggestion
    $form_upay = filter_input(INPUT_POST, 'form_upay', FILTER_VALIDATE_FLOAT, ['flags' => FILTER_REQUIRE_ARRAY]);
```

(The truthy guard on the next line continues to work — invalid entries become `false`, and `if ($amount = (float)$payment)` filters them out.)

portal/portal_payment.php
```diff
@@ -303,13 +313,13 @@
...
+}//if (filter_input(INPUT_POST, 'form_save'))
 
-if (($_POST['form_save'] ?? null) || ($_REQUEST['receipt'] ?? null)) {
-    if (($_REQUEST['receipt'] ?? null)) {
-        $form_pid = $isPortal ? $pid : $_GET['patient'];
-        $timestamp = decorateString('....-..-.. ..:..:..', $_GET['time']);
+if (filter_input(INPUT_POST, 'form_save') || filter_input(INPUT_GET, 'receipt')) {
+    if (filter_input(INPUT_GET, 'receipt')) {
+        $form_pid = $isPortal ? $pid : (int) filter_input(INPUT_GET, 'patient', FILTER_VALIDATE_INT);
+        $timestamp = decorateString('....-..-.. ..:..:..', (string) filter_input(INPUT_GET, 'time', FILTER_SANITIZE_SPECIAL_CHARS));
```
kojiromike, line 322: `$_GET['time']` is a timestamp template literal matching `....-..-.. ..:..:..`. `FILTER_SANITIZE_SPECIAL_CHARS` happens to be harmless on the digits/dashes/colons/space it expects, but the *intent* of the field is a regex-constrained shape, not HTML escaping. `FILTER_VALIDATE_REGEXP` against `^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$` would actually reject malformed input rather than pass it through to `decorateString()`.

portal/portal_payment.php
```diff
@@ -154,7 +164,7 @@
                 }
 
                 //----------------------------------------------------------------------------------------------------
-                if ($_REQUEST['radio_type_of_payment'] == 'copay') {//copay saving to ar_session and ar_activity tables
+                if ($radio_type_of_payment === 'copay') {//copay saving to ar_session and ar_activity tables
```
kojiromike, line 167: `Recorder::recordActivity()` (around `src/PaymentProcessing/Recorder.php:143`) declares an `array{..., codeType: string, code: string, modifier: string, postUser: string, sessionId: string, ...}` shape, but the call site below passes `$Codetype`/`$Code`/`$Modifier` (which are `mixed` from `sqlFetchArray()`), `$session->get('authUserID')` (mixed), and `$session_id` (`int|string` from `sqlInsert`). This is what PHPStan is flagging on the failing 8.5 job — three `recordActivity()` parameter-type errors. The fix belongs at this call site (narrow with `is_string()`/`(string)` per value, or build a typed DTO), not in Recorder.

tests/Tests/Unit/Portal/PortalPaymentInputTest.php
```diff
@@ -0,0 +1,147 @@
...
+     * @return float Sanitized, non-negative float
+     */
+    private function validatePrepayment(mixed $raw): float
+    {
+        $filtered = filter_var($raw, FILTER_VALIDATE_FLOAT);
+        if ($filtered === false || $filtered === null || $filtered < 0.0) {
+            return 0.0;
+        }
+        return $filtered;
+    }
```
kojiromike, line 45: This method is a hand-copy of the validation block at `portal/portal_payment.php:113-117`. A test against a re-implementation of production code never fails on changes to the production code — if the script's guard were reverted today, every assertion below would still pass green. Pull the validator into a tiny class in `src/Portal/` (e.g. `PrepaymentValidator::normalize(mixed $raw): float`) and have *both* `portal_payment.php` and this test call it. That way the test actually covers the migration.

tests/Tests/Unit/Portal/PortalPaymentInputTest.php
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Mirrors the form_prepayment validation logic from portal_payment.php.
+     *
+     * @param mixed $raw Raw value as received from filter_input()
+     * @return float Sanitized, non-negative float
+     */
+    private function validatePrepayment(mixed $raw): float
+    {
+        $filtered = filter_var($raw, FILTER_VALIDATE_FLOAT);
+        if ($filtered === false || $filtered === null || $filtered < 0.0) {
```
kojiromike, line 41: `filter_var($raw, FILTER_VALIDATE_FLOAT)` returns `float|false`, never `null`, so the `=== null` check is dead — PHPStan flags this on the failing 8.5 job as "Strict comparison using === between float and null will always evaluate to false." Drop the `=== null` arm.

tests/Tests/Unit/Portal/PortalPaymentInputTest.php
```diff
@@ -0,0 +1,147 @@
...
+    #[Test]
+    public function testFalsePrepaymentFallsBackToZero(): void
+    {
+        $this->assertSame(0.0, $this->validatePrepayment(false));
+    }
+
+    /**
+     * @return array<string, array{string, float}>
+     */
+    public static function validPrepaymentProvider(): array
```
kojiromike, line 96: Data providers run before coverage instrumentation starts, so without an explicit ignore this method registers as uncovered. Per the in-repo CLAUDE.md (Testing → "Data providers: mark as `@codeCoverageIgnore`"):

```suggestion
    /**
     * @return array<string, array{string, float}>
     *
     * @codeCoverageIgnore Data providers run before coverage instrumentation starts.
     */
    public static function validPrepaymentProvider(): array
```

tests/Tests/Unit/Portal/PortalPaymentInputTest.php
```diff
@@ -0,0 +1,147 @@
...
+        $unsafe = '<script>alert(1)</script>';
+        $sanitized = filter_var($unsafe, FILTER_SANITIZE_SPECIAL_CHARS);
+        $this->assertStringNotContainsString('<script>', (string) $sanitized);
+    }
+
+    #[Test]
+    public function testFilterInputValidateIntRejectsFloat(): void
+    {
+        $result = filter_var('3.14', FILTER_VALIDATE_INT);
+        $this->assertFalse($result, 'FILTER_VALIDATE_INT must reject float strings');
```
kojiromike, line 131: This asserts PHP's own `filter_var()` behaviour, not anything in `portal_payment.php` — and PHPStan already flags it as always-true (`assertFalse(false)`). We don't add OpenEMR tests to re-verify what PHPStan guarantees. Drop this test and its three `filter_input() patterns` siblings; none of them exercise portal payment code.

tests/Tests/Unit/Portal/PortalPaymentInputTest.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    #[Test]
+    public function testFilterInputValidateIntAcceptsIntegerString(): void
+    {
+        $result = filter_var('42', FILTER_VALIDATE_INT);
+        $this->assertSame(42, $result);
```
kojiromike, line 138: Same point: `assertSame(42, $result)` where `$result` is a statically-known literal asserts PHP's `FILTER_VALIDATE_INT` contract, which PHPStan already enforces. Remove — see the note above on the whole `filter_input() patterns` block.

kojiromike (changes requested):
PHPStan CI is failing on this branch with errors introduced by the PR itself — those need to land before this can merge.

**1. Most of the new test file asserts PHP/PHPStan guarantees, not OpenEMR behaviour.** The `filter_input() patterns` block (`testFilterInputValidateIntRejectsFloat`, `testFilterInputValidateIntAcceptsIntegerString`, `testFilterInputValidateIntRejectsNonNumeric`, `testFilterInputSanitizeSpecialCharsStripsHtml`) asserts PHP's own `filter_var()` contract. That's the PHP project's test suite's job — and the static type facts those tests pin down are exactly what PHPStan already enforces on this repo. We don't write tests in OpenEMR to re-verify what PHPStan guarantees; PHPStan even flags two of them as always-true (`assertFalse($result)` / `assertSame(42, $result)` on statically-known literals). Drop this whole block.

**2. The remaining tests don't exercise the production code either.** `PortalPaymentInputTest::validatePrepayment()` is a hand-copy of the validation block from `portal/portal_payment.php`. If the production guard were reverted, every assertion would still pass — so the test cannot fail on the change it claims to cover, and Codecov shows 0% patch coverage on `portal/portal_payment.php` (27 lines missing). Either extract the validator into a small testable class in `src/` that `portal_payment.php` calls (and test *that*), or drop the test rather than imply coverage that doesn't exist. Net: as written, the test file should either be removed or rebuilt around an extracted validator — none of it currently tests this PR.

**3. The PHPStan failures on this PR are real.** `Recorder::recordActivity()` declares an `array{patientId: string, ..., sessionId: string, ...}` shape; the call site now passes `mixed`/`int` values (post-`filter_input` typing, plus `sessionId` is `string|int` from `sqlInsert`). Narrow each value to `string` at the call site — guards, not `(string)` casts ("narrow, don't cast"). See inline.

**4. Filter choices on closed-set / numeric inputs.** `radio_type_of_payment` is compared with `===` against a known enum — sanitizing HTML special chars is the wrong tool; validate against the allowed list. `form_pid` is cast to `int`, so `FILTER_VALIDATE_INT` matches the use better than `FILTER_SANITIZE_SPECIAL_CHARS`. See inline.

Issue coverage: the prepayment guard does close #11393's intent (negative/non-numeric values can no longer reach `frontPayment()`); #11392's "no direct superglobal access" is addressed for this file once PHPStan passes.

kojiromike (changes requested):
Hi, thanks for trying to keep this PR up-to-date. `filter_input` has been superseded in our standards by the Symfony request object.

craigrallen:
@kojiromike I rebased this onto current master and addressed both review rounds: migrated the page to CurrentRequest/Symfony request bags, validated closed-set and numeric inputs, narrowed Recorder values, removed the pseudo-tests, and synchronized the relevant PHPStan baseline. Targeted PHPStan and PHPCS pass. A fresh CI run is in progress; re-review when convenient.

portal/portal_payment.php
```diff
@@ -332,14 +339,19 @@
...
+$receiptRequested = $request->query->getBoolean('receipt');
+if ($alertmsg === '' && ($formSave !== '' || $receiptRequested)) {
+    if ($receiptRequested) {
         $form_pid = $pid;
-        $timestamp = decorateString('....-..-.. ..:..:..', filter_input(INPUT_GET, 'time') ?: '');
+        $receiptTime = $request->query->getString('time');
+        $receiptTime = preg_match('/^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}$/D', $receiptTime) === 1
+            ? $receiptTime
+            : '';
+        $timestamp = decorateString('....-..-.. ..:..:..', $receiptTime);
```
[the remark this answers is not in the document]
craigrallen: Fixed in a24ab17cf: receipt timestamps now require both the expected shape and a round-trippable DateTimeImmutable value; invalid dates/times disable the receipt path before any payment lookup.

PR #11446, opened 2026-04-06
feat(api): add admin user management endpoints (PUT/DELETE /api/admin/users)

zmilan:
Implements phase 2 for #8965                                                         
                                         
It is based on #11222                                                                            
Adds **Phase 2** of the Configuration Management API (issue #8965): 
PUT and DELETE endpoints completing the user CRUD story under /api/admin/users. 
Users can now be updated and deactivated (soft delete) programmatically.                
                                                            
- Add PUT /api/admin/users/{uuid} — update user fields (profile, facilities, ACL groups)
- Add DELETE /api/admin/users/{uuid} — deactivate user (soft delete, sets active=0)                                                                     
- Add DATABASE_UPDATE_CONTEXT to UserValidator (all fields optional, at least one required, rejects username/password)                                      
- Override validate() in UserValidator for pre-validation checks (rejected fields, empty body)                                                           
- Add updateUser() to UserManagementService — validates, resolves UUID, checks facilities, updates columns + ACL groups in transaction, audit logs          
- Add deactivateUser() to UserManagementService — resolves UUID, sets active=0, audit logs                                                          
- Add private buildSetClause() helper — explicit field allowlist, no schema introspection                                                                 
- Add put() and delete() to UserManagementRestController with OpenAPI attributes                                                                   
- Register PUT and DELETE routes with admin/super ACL
- Add 15 PHPUnit API tests (PUT happy/sad path, DELETE happy/sad path, 403 auth for both)
                                                                                
#### Assisted-by: Claude Code


src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,537 @@
...
+    }
+
+    /**
+     * Deactivate a user (soft delete: set active=0).
+     *
+     * @param string $uuid UUID of the user to deactivate
+     * @return ProcessingResult
+     */
+    public function deactivateUser(string $uuid): ProcessingResult
+    {
```
kojiromike, line 452: This should check whether the target user is the currently authenticated admin. In a single-admin deployment, self-deactivation locks everyone out.

```suggestion
    public function deactivateUser(string $uuid): ProcessingResult
    {
        $processingResult = new ProcessingResult();

        if (!UuidRegistry::isValidStringUUID($uuid)) {
            $processingResult->setValidationMessages(['uuid' => ['Invalid UUID format']]);
            return $processingResult;
        }

        // Prevent self-deactivation
        $session = SessionWrapperFactory::getInstance()->getActiveSession();
        $currentUser = $session->get('authUserID');
```
Then after resolving the UUID to a user record, compare `$userId` against `$currentUser` and reject if they match.

src/Validators/Admin/UserValidator.php
```diff
@@ -0,0 +1,145 @@
...
+                $context->optional('taxonomy')->lengthBetween(0, 255);
+                $context->optional('specialty')->lengthBetween(0, 255);
+                $context->optional('calendar')->inArray([0, 1, '0', '1']);
+                $context->optional('portal_user')->inArray([0, 1, '0', '1']);
+                $context->optional('federaltaxid')->lengthBetween(0, 255);
+                $context->optional('state_license_number')->lengthBetween(0, 255);
+                $context->optional('federaldrugid')->lengthBetween(0, 255);
+                $context->optional('upin')->lengthBetween(0, 255);
+                $context->required('access_group')->isArray()->callback(function ($value): bool {
+                    if (!is_array($value) || $value === []) {
```
kojiromike, line 53: The callback validates that values are non-empty strings, but doesn't check that the group names correspond to real ACL groups. `AclExtended::setUserAro()` silently ignores non-matching titles, so passing `["FakeGroup"]` creates a user with zero ACL memberships while the API reports success.

Consider querying `AclExtended::aclGetGroupTitleList()` (or equivalent) and validating each value exists.

kojiromike (commented):
Good continuation of #11222. Auth checks are correct, SQL is parameterized, tests cover happy/sad paths + authz.

Issues to address:

1. No self-deactivation guard. `deactivateUser()` doesn't check whether the target UUID belongs to the currently authenticated admin. In a single-admin deployment, an admin can lock everyone out by deactivating their own account.
2. No validation of `access_group` values against real ACL groups. I mentioned this in #11222. `AclExtended::setUserAro()` silently ignores non-matching group titles, so `{"access_group": ["NonExistentGroup"]}` on PUT can strip a user's ACL memberships while the API reports success. The validator should verify each group title exists before proceeding.
3. Three separate field allowlists (controller `WHITELISTED_FIELDS`, service `buildSetClause` `$allowedFields`, validator `$updatableFields`) must stay in sync manually. Consider a way to unify them.

src/Services/Admin/UserManagementService.php
zmilan: Good point, thanks. It is fixed now. I approach it differently, by passing the authenticated user UUID, instead of using the session inside.

src/Validators/Admin/UserValidator.php
zmilan: Great catch, thanks. It is fixed now.

zmilan:
@kojiromike, your point "3. Three separate field allowlists (controller WHITELISTED_FIELDS, service buildSetClause $allowedFields, validator $updatableFields) must stay in sync manually. Consider a way to unify them." is also addressed.

src/Services/Admin/UserManagementService.php
```diff
@@ -0,0 +1,592 @@
...
+                        'access_group' => ['Invalid access group(s): ' . implode(', ', $invalidGroups)],
+                    ]);
+                    return $processingResult;
+                }
+
+                $fname = isset($data['fname']) ? trim(self::strVal($data['fname'])) : null;
+                $mname = isset($data['mname']) ? trim(self::strVal($data['mname'])) : null;
+                $lname = isset($data['lname']) ? trim(self::strVal($data['lname'])) : null;
+                if ($fname === null || $mname === null || $lname === null) {
+                    $currentUser = QueryUtils::querySingleRow("SELECT fname, mname, lname FROM users WHERE id = ?", [$userId]);
```
kojiromike, line 411: Minor: new SQL string literals in this file use double quotes (here and in the `querySingleRow`/`sqlStatementThrowException` calls throughout `updateUser`/`deactivateUser`/`createUser`). Since none of them interpolate, single quotes are the project default for non-interpolated strings (per the repo's PHP coding standards) — they make accidental `$var` injection into a query physically impossible and are a cleaner review signal. Not blocking, but worth a sweep on the new lines.

kojiromike (commented):
Re-reviewed after the latest commit (e70bc9e). The prior round's findings are addressed well:

- Self-deactivation guard now present on both PUT (`active=0`) and DELETE, via an injected authenticated-user UUID rather than reading the session inside the service — clean approach, and both paths are covered by tests (`testPutSelfDeactivateReturns400`, `testDeleteSelfDeactivationReturns400`).
- `access_group` values are now validated against real ACL groups (`validateAccessGroups()` / `aclGetGroupTitleList()`), with create + update + mixed-valid/invalid tests.
- The update field allowlist is unified behind `UserManagementService::UPDATABLE_FIELDS`, shared by the validator and `buildSetClause()`. (The controller's `WHITELISTED_FIELDS` is a separate GET-search allowlist with a different purpose, so keeping it distinct is fine.)
- JSON bodies are now decoded with `json_decode($body, true)` + `json_last_error()` returning 400 on malformed input; the malformed-UUID 500 risk is closed by the `isValidStringUUID()` guard ahead of `uuidToBytes()`; and the OpenAPI/docblock gaps Copilot flagged are filled.

The `UserService::getSelectColumns()` extraction is a faithful refactor — `getAll()` is byte-for-byte equivalent and `search()` only reorders `username`/`last_updated`, which is immaterial since rows are keyed by column name.

One convention note inline. Test coverage for the new endpoints and guards is thorough.

One thing to surface: the self-deactivation guard prevents an admin from locking *themselves* out, but nothing prevents the last remaining active admin from being deactivated by a *different* admin. Worth a follow-up if "last admin" protection is a goal.

PR #11448, opened 2026-04-06
feat(encryption): Key material storage APIs

Firehed:
Extracts and enhances the key material storage components from #11267. These are needed so that keychains (etc) will be able to load key material from a variety of sources (for now, just disk and the `keys` table; eventually, this might support vaults, HSMs, etc).

Dormant for now, integration and related wiring coming shortly.

- Adds primitive-wrapping `KeyMaterialId` class
- Adds `KeyStorage` interface
- Adds two implementations that are needed for interacting with existing `CryptoGen`-originated keys

Tests were added by AI


Firehed:
CI failing due to a test flake, landing anyway to avoid waiting an extra 10min.

PR #11456, opened 2026-04-07
refactor(menu): fix literal-string violations in xl/xlt calls

kojiromike:
Part 5 of a series fixing ~225 PHPStan literal-string violations in xl()/xlt()/xla() calls.

## Summary

Narrows mixed values from JSON-decoded menu entries and DB rows to `string` before passing them to translation helpers in `src/Menu/`. Switches unconditional `xlt()` calls on dynamic labels to the appropriate wrappers (`xlt()` with a scoped `@phpstan-ignore` for menu labels, `xl_form_title()` for form/category labels), which accept runtime strings without literal-string violations.

## Escaping policy

Menu labels **remain HTML-escaped at rest** in `MenuRole::menuUpdateEntries()`, via `xlt()` applied to the narrowed label. This is because `twAddTab()` in `src/Tabs/TabsWrapper.php` concatenates the label directly into the tab anchor and a close-button `id` attribute without escaping. The long-term fix (escape at the render site) is tracked in #11478; until then, this PR preserves the pre-PR "labels are HTML-safe at rest" contract.

Custom menu filenames in `MainMenuRole`/`PatientMenuRole` are translated and escaped at output via `xlt(substr(...))` with a scoped `@phpstan-ignore argument.type` (since the substring is dynamic). `xlt()` is the same as `text(xl(...))` by definition.

## Files

- `src/Menu/MenuRole.php` — narrow `$entry->label` before translating in `menuUpdateEntries`; uses `xlt()` (escaped at rest) due to the `twAddTab()` concern above
- `src/Menu/MainMenuRole.php` — narrow `$reglastcat`/`$title`/`$category` in `updateVisitForms` and `updateBlankForms`; custom menu filenames use `xlt(substr(...))` so they're always translated regardless of `translate_lists`. Switched `preg_match("/\.json$/", ...)` to `str_ends_with()` per review.
- `src/Menu/PatientMenuRole.php` — narrow `$mod_nick_name` and assign raw to `$subEntry->label` (`MenuRole::menuUpdateEntries()` will translate + escape during the recursive pass); custom menu filenames use `xlt(substr(...))` same as `MainMenuRole`. Same `str_ends_with()` cleanup.

## Follow-up issues filed during review

- openemr/openemr#11478 — escape at the render site in `TabsWrapper::twAddTab()` so menu labels can be stored unescaped.
- openemr/openemr#11487 — extract a shared helper from `updateVisitForms` and `updateBlankForms`.

## Baseline impact

PHPStan baselines only go down:

- `argument.type.php`: 30 entries removed (covers all literal-string + several `mixed`-passed-to-string violations on these three files)
- `empty.notAllowed.php`: MainMenuRole 6 → 4
- `offsetAccess.nonOffsetAccessible.php`: MainMenuRole 4 → 2

No new baseline entries were introduced.

## Test plan

- [x] `composer phpstan` — clean, no errors
- [x] `composer phpstan-baseline` — only deletions/decreases for menu files
- [x] Pre-commit hooks (phpcs, phpcbf, rector, codespell, require-checker) — all passing

### DIFF
--- a/src/Menu/MainMenuRole.php
+++ b/src/Menu/MainMenuRole.php
@@ -49,7 +49,7 @@ class MainMenuRole extends MenuRole
         $mainMenuRole = $this->getMenuRole();
 
         // Load the selected menu
-        if (preg_match("/.json$/", $mainMenuRole)) {
+        if (str_ends_with($mainMenuRole, '.json')) {
             // load custom menu (includes .json in id)
             $menu_parsed = json_decode(file_get_contents(OEGlobalsBag::getInstance()->get('OE_SITE_DIR') . "/documents/custom_menus/" . $mainMenuRole));
         } else {
@@ -90,10 +90,12 @@ class MainMenuRole extends MenuRole
             $dHandle = opendir($customMenuDir);
             while (false !== ($menuCustom = readdir($dHandle))) {
                 // Only process files that contain *.json
-                if (preg_match("/.json$/", $menuCustom)) {
+                if (str_ends_with($menuCustom, '.json')) {
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
+                    // Drop the .json suffix and translate the name. Custom
+                    // menu filenames are dynamic, hence the @phpstan-ignore.
+                    // @phpstan-ignore argument.type (custom menu filenames are dynamic)
                     $output .= xlt(substr($menuCustom, 0, -5));
                     $output .= "</option>";
                 }
@@ -132,17 +134,18 @@ class MainMenuRole extends MenuRole
         $regrows = getFormsByCategory('1', false);
         foreach ($regrows as $entry) {
             $option_id = $entry['directory'];
-            $title = trim($entry['nickname'] ?? '');
-            if (empty($title)) {
-                $title = $entry['name'];
-            }
-            if ($entry['category'] != $reglastcat) {
+            $nickname = is_string($entry['nickname'] ?? null) ? trim($entry['nickname']) : '';
+            // Preserve the old empty() semantics: '' AND '0' both fall through
+            // to the form's name. (empty('0') returns true.)
+            $title = ($nickname !== '' && $nickname !== '0') ? $nickname : (is_string($entry['name'] ?? null) ? $entry['name'] : '');
+            $category = is_string($entry['category'] ?? null) ? $entry['category'] : '';
+            if ($category != $reglastcat) {
                 // New category. Close out the previous one if it exists.
                 if ($reglastcat) {
                     array_push($menu_list->children, $catEntry);
                 }
                 // Create the new category's object.
-                $reglastcat = $entry['category'];
+                $reglastcat = $category;
                 $catEntry = new \stdClass();
                 $catEntry->label = xl_form_title($reglastcat);
                 $catEntry->icon = 'fa-caret-right';
@@ -183,17 +186,18 @@ class MainMenuRole extends MenuRole
         $regrows = getFormsByCategory('1', true);
         foreach ($regrows as $entry) {
             $option_id = $entry['directory'];
-            $title = trim($entry['nickname'] ?? '');
-            if (empty($title)) {
-                $title = $entry['name'];
-            }
-            if ($entry['category'] != $reglastcat) {
+            $nickname = is_string($entry['nickname'] ?? null) ? trim($entry['nickname']) : '';
+            // Preserve the old empty() semantics: '' AND '0' both fall through
+            // to the form's name. (empty('0') returns true.)
+            $title = ($nickname !== '' && $nickname !== '0') ? $nickname : (is_string($entry['name'] ?? null) ? $entry['name'] : '');
+            $category = is_string($entry['category'] ?? null) ? $entry['category'] : '';
+            if ($category != $reglastcat) {
                 // New category. Close out the previous one if it exists.
                 if ($reglastcat) {
                     array_push($menu_list->children, $catEntry);
                 }
                 // Create the new category's object.
-                $reglastcat = $entry['category'];
+                $reglastcat = $category;
                 $catEntry = new \stdClass();
                 $catEntry->label = xl_form_title($reglastcat);
                 $catEntry->icon = 'fa-caret-right';
--- a/src/Menu/MenuRole.php
+++ b/src/Menu/MenuRole.php
@@ -52,8 +52,13 @@ abstract class MenuRole implements MenuRoleInterface
                 }
             }
 
-            // Translate the labels
-            $entry->label = xlt($entry->label);
+            // Translate the labels with HTML escaping. At least one render
+            // site (twAddTab in src/Tabs/TabsWrapper.php) concatenates the
+            // label directly into an anchor and an id attribute without
+            // escaping, so labels must be HTML-safe at rest.
+            $rawLabel = $entry->label ?? null;
+            // @phpstan-ignore argument.type (menu labels are dynamic content)
+            $entry->label = xlt(is_string($rawLabel) ? $rawLabel : '');
             // Recursive update of children
             if (isset($entry->children)) {
                 $this->menuUpdateEntries($entry->children);
--- a/src/Menu/PatientMenuRole.php
+++ b/src/Menu/PatientMenuRole.php
@@ -53,7 +53,7 @@ class PatientMenuRole extends MenuRole
         $patientMenuRole = $this->getMenuRole();
 
         // Load the selected menu
-        if (preg_match("/.json$/", $patientMenuRole)) {
+        if (str_ends_with($patientMenuRole, '.json')) {
             // load custom menu (includes .json in id)
             $menu_parsed = json_decode(file_get_contents(OEGlobalsBag::getInstance()->get('OE_SITE_DIR') . "/documents/custom_menus/patient_menus/" . $patientMenuRole));
         } else {
@@ -97,10 +97,12 @@ class PatientMenuRole extends MenuRole
             $dHandle = opendir($customMenuDir);
             while (false !== ($menuCustom = readdir($dHandle))) {
                 // Only process files that contain *.json
-                if (preg_match("/.json$/", $menuCustom)) {
+                if (str_ends_with($menuCustom, '.json')) {
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
+                    // Drop the .json suffix and translate the name. Custom
+                    // menu filenames are dynamic, hence the @phpstan-ignore.
+                    // @phpstan-ignore argument.type (custom menu filenames are dynamic)
                     $output .= xlt(substr($menuCustom, 0, -5));
                     $output .= "</option>";
                 }
@@ -159,13 +161,22 @@ class PatientMenuRole extends MenuRole
                 }
 
                 $relative_link = "../../modules/" . $modulePath . "/public/" . $hookrow['path'];
-                $mod_nick_name = $hookrow['menu_name'] ?: 'NoName';
+                // Preserve the old `?:` semantics: '' AND '0' both fall through
+                // to 'NoName'. (The previous code used `$hookrow['menu_name'] ?: 'NoName'`,
+                // and `'0' ?: 'NoName'` returns 'NoName'.)
+                $menuName = $hookrow['menu_name'] ?? null;
+                $mod_nick_name = is_string($menuName) && $menuName !== '' && $menuName !== '0'
+                    ? $menuName
+                    : 'NoName';
 
                 $subEntry = new \stdClass();
                 $subEntry->requirement = 0;
                 $subEntry->target = 'main';
                 $subEntry->menu_id = $hookrow['mod_id'];
-                $subEntry->label = xlt($mod_nick_name);
+                // Assign the raw module menu name; MenuRole::menuUpdateEntries()
+                // will translate and HTML-escape it via xlt() during the
+                // recursive pass over children.
+                $subEntry->label = $mod_nick_name;
                 $subEntry->url = $relative_link;
                 $subEntry->on_click = 'top.restoreSession()';
                 $subEntry->pid = 'false';
--- a/.phpstan/baseline/empty.notAllowed.php
+++ b/.phpstan/baseline/empty.notAllowed.php
@@ -4188,7 +4188,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#',
-    'count' => 6,
+    'count' => 4,
     'path' => __DIR__ . '/../../src/Menu/MainMenuRole.php',
 ];
 $ignoreErrors[] = [
--- a/.phpstan/baseline/offsetAccess.nonOffsetAccessible.php
+++ b/.phpstan/baseline/offsetAccess.nonOffsetAccessible.php
@@ -44353,7 +44353,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Cannot access offset \'category\' on mixed\\.$#',
-    'count' => 4,
+    'count' => 2,
     'path' => __DIR__ . '/../../src/Menu/MainMenuRole.php',
 ];
 $ignoreErrors[] = [


src/Menu/MenuRole.php
```diff
@@ -52,8 +52,9 @@ protected function menuUpdateEntries(&$menu_list)
                 }
             }
 
-            // Translate the labels
-            $entry->label = xlt($entry->label);
+            // Translate the labels (stored as plain strings; render sites escape).
+            $rawLabel = $entry->label ?? null;
+            $entry->label = xl_form_title(is_string($rawLabel) ? $rawLabel : '');
```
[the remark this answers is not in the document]
kojiromike: Good catch — you're right that `xl_form_title()` is gated by `translate_form_titles`, which is meant for patient-note form titles. Menu labels should translate unconditionally. Fixed in 9104220 by switching to `xl()` with a local `@phpstan-ignore argument.type` (matching the pattern in `library/translation.inc.php`).

src/Menu/PatientMenuRole.php
```diff
@@ -159,13 +159,15 @@ protected function updateModulesDemographicsMenu(&$menu_list)
...
+                $mod_nick_name = is_string($hookrow['menu_name'] ?? null) && $hookrow['menu_name'] !== ''
+                    ? $hookrow['menu_name']
+                    : 'NoName';
 
                 $subEntry = new \stdClass();
                 $subEntry->requirement = 0;
                 $subEntry->target = 'main';
                 $subEntry->menu_id = $hookrow['mod_id'];
-                $subEntry->label = xlt($mod_nick_name);
+                $subEntry->label = xl_form_title($mod_nick_name);
```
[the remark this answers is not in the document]
kojiromike: Same fix as the `MenuRole` thread — module menu names aren't form titles, so they shouldn't be gated by `translate_form_titles`. Switched to `xl()` with a local PHPStan ignore in 9104220.

src/Menu/MenuRole.php
```diff
@@ -52,8 +52,11 @@ protected function menuUpdateEntries(&$menu_list)
                 }
             }
 
-            // Translate the labels
-            $entry->label = xlt($entry->label);
+            // Translate the labels (stored as plain strings; render sites escape).
+            // Menu labels are not form titles, so translate unconditionally via xl().
+            $rawLabel = $entry->label ?? null;
+            // @phpstan-ignore argument.type (menu labels are dynamic content)
+            $entry->label = xl(is_string($rawLabel) ? $rawLabel : '');
```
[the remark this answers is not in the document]
kojiromike: You're right — I verified that `twAddTab` in `src/Tabs/TabsWrapper.php` concatenates `label` directly into the anchor text AND into an `id` attribute, with no escaping. My refactor broke the implicit "labels are HTML-safe at rest" contract.

Reverted to `xlt()` in `970bd45d` so the pre-PR escaping behavior is restored. Kept the `is_string()` narrowing and the `@phpstan-ignore` so the literal-string check still passes.

Filed openemr/openemr#11474 to track the long-term fix: escape at the render site (via jQuery `.text()` and a panel-id-based close-button id) instead of relying on upstream escaping. That's out of scope for this PR.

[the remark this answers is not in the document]
kojiromike: (Correction: the follow-up issue is openemr/openemr#11478, not 11474 — fat-fingered the link in the previous reply.)

src/Menu/MainMenuRole.php
```diff
@@ -93,8 +93,8 @@ public function displayMenuRoleSelector($selected = "")
                 if (preg_match("/.json$/", $menuCustom)) {
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
-                    $output .= xlt(substr($menuCustom, 0, -5));
+                    // Drop the .json and translate the name (custom menu names are dynamic).
+                    $output .= text(xl_list_label(substr($menuCustom, 0, -5)));
```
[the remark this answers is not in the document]
kojiromike: Fixed in `281a23f2` per your suggestion — switched to `xl()` with a scoped `@phpstan-ignore argument.type` so custom menu filenames are translated unconditionally, matching the pre-PR `xlt()` behavior.

src/Menu/PatientMenuRole.php
```diff
@@ -100,8 +100,8 @@ public function displayMenuRoleSelector($selected = "")
                 if (preg_match("/.json$/", $menuCustom)) {
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
-                    $output .= xlt(substr($menuCustom, 0, -5));
+                    // Drop the .json and translate the name (custom menu names are dynamic).
+                    $output .= text(xl_list_label(substr($menuCustom, 0, -5)));
```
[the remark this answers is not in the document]
kojiromike: Same fix as the MainMenuRole thread — also applied to `PatientMenuRole.php` in `281a23f2`.

src/Menu/MenuRole.php
```diff
@@ -52,8 +52,13 @@ protected function menuUpdateEntries(&$menu_list)
...
 
-            // Translate the labels
-            $entry->label = xlt($entry->label);
+            // Translate the labels with HTML escaping. At least one render
+            // site (twAddTab in src/Tabs/TabsWrapper.php) concatenates the
+            // label directly into an anchor and an id attribute without
+            // escaping, so labels must be HTML-safe at rest.
+            $rawLabel = $entry->label ?? null;
+            // @phpstan-ignore argument.type (menu labels are dynamic content)
+            $entry->label = xlt(is_string($rawLabel) ? $rawLabel : '');
```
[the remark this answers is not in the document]
kojiromike: Good catch. Updated the PR description to add an "Escaping policy" section that documents the exception: `MenuRole::menuUpdateEntries` still stores labels escaped-at-rest via `xlt()` because `twAddTab()` in `TabsWrapper.php` concatenates the label into unescaped HTML. Linked the long-term follow-up (openemr/openemr#11478) in the description. The per-file notes now call out which structures are escaped-at-rest vs render-time.

src/Menu/MainMenuRole.php
```diff
@@ -93,8 +93,11 @@ public function displayMenuRoleSelector($selected = "")
                 if (preg_match("/.json$/", $menuCustom)) {
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
```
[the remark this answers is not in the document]
kojiromike: Fixed in `18310e84` — escaped the dot in both `MainMenuRole` regex sites (the menu file selector at line 52 and the custom menu dropdown at line 93). Same fix applied to the corresponding sites in `PatientMenuRole`.

src/Menu/PatientMenuRole.php
```diff
@@ -100,8 +100,11 @@ public function displayMenuRoleSelector($selected = "")
                 if (preg_match("/.json$/", $menuCustom)) {
```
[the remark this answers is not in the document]
kojiromike: Fixed in `18310e84` along with the matching `MainMenuRole` regex sites — escaped the dot in both `preg_match` calls in `PatientMenuRole`.

src/Menu/PatientMenuRole.php
```diff
@@ -159,13 +162,17 @@ protected function updateModulesDemographicsMenu(&$menu_list)
...
+                    : 'NoName';
 
                 $subEntry = new \stdClass();
                 $subEntry->requirement = 0;
                 $subEntry->target = 'main';
                 $subEntry->menu_id = $hookrow['mod_id'];
-                $subEntry->label = xlt($mod_nick_name);
+                // Module menu names are dynamic, not form titles — translate unconditionally via xl().
+                // @phpstan-ignore argument.type (module menu names are dynamic content)
+                $subEntry->label = xl($mod_nick_name);
```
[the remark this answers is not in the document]
kojiromike: Good catch — that was a double-translation. Fixed in `18310e84`: assigned the raw `$mod_nick_name` to `$subEntry->label` and let `MenuRole::menuUpdateEntries()` apply `xlt()` (translation + escaping) in one place during the recursive pass over children. Removed the redundant `xl()` call and its `@phpstan-ignore`.

src/Menu/MainMenuRole.php
```diff
@@ -49,7 +49,7 @@ public function getMenu()
         $mainMenuRole = $this->getMenuRole();
 
         // Load the selected menu
-        if (preg_match("/.json$/", $mainMenuRole)) {
+        if (preg_match("/\.json$/", $mainMenuRole)) {
```
kojiromike: Should this be `str_ends_with`?

src/Menu/MainMenuRole.php
```diff
@@ -90,11 +90,14 @@ public function displayMenuRoleSelector($selected = "")
             $dHandle = opendir($customMenuDir);
             while (false !== ($menuCustom = readdir($dHandle))) {
                 // Only process files that contain *.json
-                if (preg_match("/.json$/", $menuCustom)) {
+                if (preg_match("/\.json$/", $menuCustom)) {
```
kojiromike: `str_ends_with`?

src/Menu/MainMenuRole.php
```diff
@@ -90,11 +90,14 @@ public function displayMenuRoleSelector($selected = "")
...
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
-                    $output .= xlt(substr($menuCustom, 0, -5));
+                    // Drop the .json and translate the name. Use xl() rather
+                    // than xl_list_label() so custom menu filenames are
+                    // translated unconditionally, matching the pre-PR xlt().
+                    // @phpstan-ignore argument.type (custom menu filenames are dynamic)
+                    $output .= text(xl(substr($menuCustom, 0, -5)));
```
kojiromike: is `text(xl())` the same as `xlt()`?

src/Menu/MainMenuRole.php
```diff
@@ -183,17 +185,16 @@ protected function updateBlankForms(&$menu_list)
         $regrows = getFormsByCategory('1', true);
         foreach ($regrows as $entry) {
             $option_id = $entry['directory'];
-            $title = trim($entry['nickname'] ?? '');
-            if (empty($title)) {
-                $title = $entry['name'];
-            }
-            if ($entry['category'] != $reglastcat) {
+            $nickname = is_string($entry['nickname'] ?? null) ? trim($entry['nickname']) : '';
```
kojiromike, line 189: Is this loop almost identical to the loop on line 136? is it worth a followup ticket to abstract and DRY?

src/Menu/MenuRole.php
kojiromike, line 61: is there no special function for translating labels?

src/Menu/PatientMenuRole.php
```diff
@@ -53,7 +53,7 @@ public function getMenu()
         $patientMenuRole = $this->getMenuRole();
 
         // Load the selected menu
-        if (preg_match("/.json$/", $patientMenuRole)) {
+        if (preg_match("/\.json$/", $patientMenuRole)) {
```
kojiromike: `str_ends_with`?

src/Menu/PatientMenuRole.php
```diff
@@ -97,11 +97,14 @@ public function displayMenuRoleSelector($selected = "")
             $dHandle = opendir($customMenuDir);
             while (false !== ($menuCustom = readdir($dHandle))) {
                 // Only process files that contain *.json
-                if (preg_match("/.json$/", $menuCustom)) {
+                if (preg_match("/\.json$/", $menuCustom)) {
```
kojiromike: `str_ends_with`?

src/Menu/PatientMenuRole.php
```diff
@@ -97,11 +97,14 @@ public function displayMenuRoleSelector($selected = "")
...
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
-                    $output .= xlt(substr($menuCustom, 0, -5));
+                    // Drop the .json and translate the name. Use xl() rather
+                    // than xl_list_label() so custom menu filenames are
+                    // translated unconditionally, matching the pre-PR xlt().
+                    // @phpstan-ignore argument.type (custom menu filenames are dynamic)
+                    $output .= text(xl(substr($menuCustom, 0, -5)));
```
kojiromike: `xlt`?

src/Menu/PatientMenuRole.php
```diff
@@ -97,11 +97,14 @@ public function displayMenuRoleSelector($selected = "")
...
                 // Only process files that contain *.json
-                if (preg_match("/.json$/", $menuCustom)) {
+                if (preg_match("/\.json$/", $menuCustom)) {
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
-                    $output .= xlt(substr($menuCustom, 0, -5));
+                    // Drop the .json and translate the name. Use xl() rather
+                    // than xl_list_label() so custom menu filenames are
+                    // translated unconditionally, matching the pre-PR xlt().
```
kojiromike: What does "translated unconditionally" mean? All translation is conditional upon there being a translation in the system.

src/Menu/MainMenuRole.php
kojiromike: Switched to `str_ends_with($mainMenuRole, '.json')` in `c417cd3a` — much cleaner than the regex.

src/Menu/MainMenuRole.php
kojiromike: Same fix applied here in `c417cd3a` — `str_ends_with((string) $menuCustom, '.json')`. Rector then trimmed the redundant `(string)` cast since `readdir()` returns `string|false` and the `false !==` guard above narrows it.

src/Menu/MainMenuRole.php
kojiromike: Yes — `xlt()` is literally `function xlt($key) { return text(hsc_private_xl_or_warn($key)); }` (which calls `xl()` internally). Switched to `xlt(...)` in `c417cd3a`. The `@phpstan-ignore` is still needed because the substring is dynamic content, but it's now the same shape as the rest of the file.

src/Menu/MainMenuRole.php
kojiromike: Yes, they're duplicates differing only in the form source, the target/URL, and the ACO handling. Filed openemr/openemr#11487 to track extracting a shared helper. Out of scope for this PR.

src/Menu/MenuRole.php
kojiromike: `xlt()` (which is what's called here after the `is_string` narrowing) is the function — it does both translation and HTML escaping in one shot. There are also `xl_form_title()` and `xl_list_label()` but those are gated on the `translate_form_titles` / `translate_lists` globals respectively, which is the wrong contract for menu labels (we want them translated regardless of those settings).

src/Menu/PatientMenuRole.php
kojiromike: Switched to `str_ends_with($patientMenuRole, '.json')` in `c417cd3a`.

src/Menu/PatientMenuRole.php
kojiromike: Same fix — `str_ends_with($menuCustom, '.json')` in `c417cd3a` (Rector dropped the `(string)` cast as redundant).

src/Menu/PatientMenuRole.php
kojiromike: Yes — switched `text(xl(...))` to `xlt(...)` in `c417cd3a`. They're identical (xlt() = text(xl()) by definition), but the shorthand reads cleaner.

src/Menu/PatientMenuRole.php
kojiromike: You're right, that wording was muddled. I meant "the lookup runs regardless of whether `translate_lists` is enabled" — not "translation is guaranteed." Dropped the misleading comment in `c417cd3a`. Translation always depends on a row existing in the database; the only thing the choice of helper controls is whether the lookup happens at all (`xl()` always looks up; `xl_list_label()` short-circuits when `translate_lists` is off).

src/Menu/MainMenuRole.php
```diff
@@ -90,10 +90,12 @@ public function displayMenuRoleSelector($selected = "")
...
-                if (preg_match("/.json$/", $menuCustom)) {
+                if (str_ends_with($menuCustom, '.json')) {
                     $selectedTag = ($selected == $menuCustom) ? "selected" : "";
                     $output .= "<option value='" . attr($menuCustom) . "' " . $selectedTag . ">";
-                    // Drop the .json and translate the name
+                    // Drop the .json suffix and translate the name. Custom
+                    // menu filenames are dynamic, hence the @phpstan-ignore.
+                    // @phpstan-ignore argument.type (custom menu filenames are dynamic)
                     $output .= xlt(substr($menuCustom, 0, -5));
                     $output .= "</option>";
```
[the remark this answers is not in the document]
kojiromike: Updated the PR description to match the code: it now says custom menu filenames are translated via `xlt(substr(...))` with a scoped `@phpstan-ignore`, and notes that `xlt()` is the same as `text(xl(...))` by definition. Also added a section listing the follow-up issues filed during review (#11478 and #11487).

src/Menu/MainMenuRole.php
```diff
@@ -132,17 +134,16 @@ protected function updateVisitForms(&$menu_list)
         $regrows = getFormsByCategory('1', false);
         foreach ($regrows as $entry) {
             $option_id = $entry['directory'];
-            $title = trim($entry['nickname'] ?? '');
-            if (empty($title)) {
-                $title = $entry['name'];
-            }
-            if ($entry['category'] != $reglastcat) {
+            $nickname = is_string($entry['nickname'] ?? null) ? trim($entry['nickname']) : '';
+            $title = $nickname !== '' ? $nickname : (is_string($entry['name'] ?? null) ? $entry['name'] : '');
```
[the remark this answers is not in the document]
kojiromike: Good catch — my replacement of `empty()` with `!== ''` did silently change behavior for `nickname = '0'`. Fixed in `3c51349d` per your suggestion: `($nickname !== '' && $nickname !== '0')` preserves the old `empty('0') === true` semantics. Applied to both `updateVisitForms` and `updateBlankForms` since they share the pattern.

src/Menu/PatientMenuRole.php
```diff
@@ -159,13 +161,18 @@ protected function updateModulesDemographicsMenu(&$menu_list)
                 }
 
                 $relative_link = "../../modules/" . $modulePath . "/public/" . $hookrow['path'];
-                $mod_nick_name = $hookrow['menu_name'] ?: 'NoName';
+                $mod_nick_name = is_string($hookrow['menu_name'] ?? null) && $hookrow['menu_name'] !== ''
+                    ? $hookrow['menu_name']
```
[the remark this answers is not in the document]
kojiromike: Same edge case as the previous nickname fix — `$menu_name ?: 'NoName'` falls back for both `''` and `'0'`. Fixed in `195b22a3` with the explicit `&& $menuName !== '0'` check, plus the `$menuName` extraction from your suggestion to make the chain readable.

src/Menu/MainMenuRole.php
```diff
@@ -49,7 +49,7 @@ public function getMenu()
         $mainMenuRole = $this->getMenuRole();
 
         // Load the selected menu
-        if (preg_match("/.json$/", $mainMenuRole)) {
+        if (str_ends_with($mainMenuRole, '.json')) {
             // load custom menu (includes .json in id)
             $menu_parsed = json_decode(file_get_contents(OEGlobalsBag::getInstance()->get('OE_SITE_DIR') . "/documents/custom_menus/" . $mainMenuRole));
         } else {
```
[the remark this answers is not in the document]
kojiromike: Valid concern, but pre-existing. This PR only changed line 52 from `preg_match("/\.json$/", ...)` to `str_ends_with(...)` — the `file_get_contents()` path concatenation at line 54 is unchanged. Also note that `main_menu_role` is admin-controlled (via `interface/usergroup/`), so it's hardening rather than a remotely exploitable issue.

Filed openemr/openemr#11489 to track adding basename-only validation (or whitelisting against the actual files discovered in `displayMenuRoleSelector()`). Out of scope for this PR.

src/Menu/PatientMenuRole.php
```diff
@@ -53,7 +53,7 @@ public function getMenu()
         $patientMenuRole = $this->getMenuRole();
 
         // Load the selected menu
-        if (preg_match("/.json$/", $patientMenuRole)) {
+        if (str_ends_with($patientMenuRole, '.json')) {
             // load custom menu (includes .json in id)
             $menu_parsed = json_decode(file_get_contents(OEGlobalsBag::getInstance()->get('OE_SITE_DIR') . "/documents/custom_menus/patient_menus/" . $patientMenuRole));
         } else {
```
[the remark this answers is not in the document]
kojiromike: Same concern as the `MainMenuRole` thread, pre-existing in `PatientMenuRole::getMenu()`. Tracked in openemr/openemr#11489 — the suggested fix covers both files.

Issue #11458, opened 2026-04-07
refactor(install): replace Installer::load_file() line-parser with mysqli_multi_query()


kojiromike:
## Background

`Installer::load_file()` in `library/classes/Installer.class.php` hand-rolls a SQL dump parser: read line-by-line via `fgets($fd, 1024)`, `rtrim()` each result, skip `--` and `#` comments, concatenate into a buffer, split statements on trailing `;`, execute each one.

This state machine has produced at least two bugs so far:

1. **#10935 / #11003** — `rtrim()` strips the newline between short lines, and direct concatenation fuses tokens across line boundaries (e.g. `= 0\nWHERE` → `= 0WHERE`).
2. **Latent, unreported** — a `--` or `#` comment longer than 1023 bytes gets chunked across `fgets()` calls. Only the first chunk starts with `--`/`#` and gets skipped; every subsequent chunk is treated as SQL and injected into the current statement buffer. A stray `;` in the comment prose triggers the installer to execute a comment fragment as SQL.

Both bugs come from the same root cause: we are manually doing a job MySQL's own parser already does.

## Proposal

Replace the line-by-line loop with `mysqli_multi_query()`:

```php
$sql = file_get_contents($filename);
if ($sql === false) { /* error */ }

$batch = "SET autocommit=0;\nSTART TRANSACTION;\n" . $sql . "\nCOMMIT;\nSET autocommit=1;\n";

if (!mysqli_multi_query($this->dbh, $batch)) {
    // handle error
}
do {
    $result = mysqli_store_result($this->dbh);
    if ($result) { mysqli_free_result($result); }
} while (mysqli_more_results($this->dbh) && mysqli_next_result($this->dbh));

if (mysqli_errno($this->dbh)) {
    // report error
}
```

This deletes the whole parser, `getLine()` wrapper, and the cargo-culted 1024-byte buffer. MySQL natively handles `--` and `#` comments, `/* */` block comments, `;` inside string literals, hex literals, escaped quotes, and multi-byte characters.

### Things to verify / decide

1. **Transaction wrapping.** Current code wraps each file in `SET autocommit=0; START TRANSACTION; ... COMMIT; SET autocommit=1;` for InnoDB speedup. `multi_query` passes these through fine — prepend/append to the batch as shown above.

[...]

Issue #10935, opened 2026-02-27
fix(install): Installer::load_file() drops whitespace between SQL lines


kojiromike:
## Bug

`Installer::load_file()` in `library/classes/Installer.class.php` reads a SQL dump file line by line, strips trailing whitespace (including newlines) from each line with `rtrim()`, then concatenates the lines directly with no separator:

```php
$line = $this->getLine($fd, 1024);
$line = rtrim($line);
// ...
$query .= $line;
```

This silently joins multi-line SQL statements without a space. For example:

```sql
UPDATE layout_options SET uor = 0
WHERE form_id = 'DEM'
```

becomes:

```
UPDATE layout_options SET uor = 0WHERE form_id = 'DEM'
```

…which is a syntax error.

This hasn't broken installs in practice because the SQL dumps shipped with OpenEMR use indented continuation lines — the leading whitespace on the next line provides the separator. Any multi-line statement where a continuation line starts at column 0 (no leading whitespace) will break.

## Why a one-character fix isn't quite enough

The obvious fix — `$query .= " " . $line;` — looks right but introduces a regression, because of how `getLine()` reads the file:

```php
protected function getLine($stream, int $length): string|false
{
    return fgets($stream, $length);  // length = 1024
}
```

`fgets($fd, 1024)` reads **up to 1023 bytes**, stopping early on a newline or EOF. For a short line (< 1023 bytes), it returns the whole line including the trailing `\n`. For a long line (≥ 1023 bytes), the first call returns 1023 bytes with no newline and the next call continues reading from where it left off.

OpenEMR's bundled SQL dumps contain hex literals several KB long — notably the `document_templates` fixture, which inserts HTML document content as `0x3c68746d6c3e...` literals that far exceed 1 KB. These lines get split into multiple `fgets()` chunks.

The old buggy `$query .= $line;` happened to reassemble those chunks losslessly, because no separator means byte-for-byte concatenation. Inserting an unconditional space between every `$query .=` corrupts the hex literal mid-token:

```
0x3c68746d6c... 520627574746f6e...
              ^ inserted space breaks the literal
```

…producing `You have an error in your SQL syntax ... near '520627574746f6e...'` during install. This was the cascade failure across every "Test All Configurations" job on the first attempt at #11003.

## Fix options

**Option A (minimal, preferred):** Remove the 1024-byte limit so `fgets()` always returns a complete line. Every call then yields exactly one logical line, and `$query .= " " . $line;` is safe:

```php
[...]

PR #9275, opened 2025-10-28
Observation Treatment Intervention Preference - Observation Care Experience Preference Cards

sjpadgett:
modify UtilsService createCoding() to not set elements that have empty strings.

New Dashboard cards Observation user input.

Fixes #9295
New Dashboard cards Observation user input.

PR #11461, opened 2026-04-07
refactor(install): drop 1024-byte fgets() buffer in Installer::getLine()

### LINKED ISSUE #11459: refactor(install): drop 1024-byte fgets() buffer limit in Installer::getLine()
## Background

`Installer::getLine()` in `library/classes/Installer.class.php` wraps `fgets()` with a 1024-byte length limit:

```php
protected function getLine($stream, int $length): string|false
{
    return fgets($stream, $length);  // length = 1024 at every call site
}
```

The limit was introduced in August 2010 (commit 7da2b7a44d). It is cargo-culted from pre-PHP-4.3 code, when the `length` argument to `fgets()` was mandatory and `1024` was the textbook example buffer size. PHP 4.3 made the argument optional in December 2002 and PHP 4 went out of security support in August 2008 — two years before this line was committed. The limit has been untouched for fifteen years and serves no purpose: SQL dumps are a few MB at most, OpenEMR's PHP memory limit is orders of magnitude higher, and the file contents are bundled (not untrusted input).

The 1024 is not just dead weight — it is the direct cause of two bugs:

1. Lines longer than 1023 bytes get split across multiple `fgets()` calls, and `load_file()` treats each chunk as a complete line. Hex literals in the `document_templates` fixture span several KB and are always chunked. This made the minimal fix for #10935 corrupt the literals (see #11003 for the full story).
2. A `--` or `#` comment longer than 1023 bytes is chunked the same way. Only the first chunk is recognized as a comment and skipped; subsequent chunks get injected into whatever statement is accumulating. Latent, never reported, but loaded.

#11003 works around both problems by tracking whether each `fgets()` chunk ended in a newline and only inserting the statement separator at real line boundaries. That works, but it is extra state machine piled on top of a state machine that shouldn't need it.

## Proposal

Drop the length argument entirely:

```php
protected function getLine($stream): string|false
{
    return fgets($stream);
}
```

`fgets()` without a length reads until newline or EOF, so every call returns a complete logical line. Then:

- The EOL-tracking logic added in #11003 (`$prevReachedEol`) becomes dead code and can be deleted.
- The companion regression test for chunked long lines (`testLoadFileDoesNotInsertSeparatorBetweenChunksOfSameLine`) can be removed — the scenario it covers is no longer possible.
- The mid-comment-chunk bug (bug 2 above) is fixed as a side effect, because `--`/`#` comments always arrive as complete lines.
- All existing call sites of `getLine()` need their `1024` argument removed.

## Out of scope

- Replacing the whole line-by-line parser with `mysqli_multi_query()`. That is the larger cleanup tracked in #11458. This issue is the narrower, lower-risk precursor: delete the buffer limit and the scaffolding it required, without touching the parser shape.

## Relation to #11003 / #11458

- #11003 (merged) lands the minimal fix for #10935 with EOL tracking.
- **This issue** deletes the buffer limit and the EOL scaffolding.
- #11458 replaces the whole parser with `mysqli_multi_query()`, subsuming this issue if it lands first.

kojiromike:
Closes #11459.

The `1024` length argument on `fgets()` in `Installer::getLine()` was introduced in August 2010 and cargo-culted from pre-PHP-4.3 code, when the argument was mandatory and `1024` was the textbook example. PHP 4.3 made it optional in December 2002 and PHP 4 went out of security support in August 2008 — two years before this line was committed. It has served no purpose for fifteen years, and it was the root cause of two bugs:

1. **#10935 / #11003.** Lines longer than 1023 bytes (notably the multi-KB hex literals in the `document_templates` fixture) were split across multiple `fgets()` calls. The minimal fix for #10935 inserted a space between concatenated lines and ended up inserting one into the middle of a hex literal, breaking install in CI. #11003 worked around this by tracking whether each chunk ended in a newline and only inserting the separator at real line boundaries.
2. **Latent.** A `--` or `#` comment longer than 1023 bytes got chunked the same way. Only the first chunk was recognized as a comment and skipped; subsequent chunks got injected into whatever statement was accumulating. Never reported in the wild because nobody wrote a long enough comment, but the footgun was loaded.

## Change

Drop the length argument. `fgets($fd)` reads until newline or EOF, so every call returns a complete logical line. This:

- deletes the `\$prevReachedEol` tracking that #11003 added,
- lets `load_file()` unconditionally insert the space separator,
- removes the companion regression test for the chunked-line case (the scenario is no longer possible),
- and fixes the latent long-comment bug as a side effect.

Net: **-74 lines**, and the bug class is gone at the root.

## Not in scope

Replacing the whole line-by-line parser with `mysqli_multi_query()`. That is the larger cleanup tracked in #11458 and subsumes this change if it lands first.

## Test plan

- [x] `composer phpunit-isolated` — full isolated suite passes (113 → 112 tests; the chunked-line regression test was removed as obsolete)
- [x] `composer phpstan` — clean (one now-obsolete baseline entry removed)
- [ ] CI install + "Test All Configurations" matrix — confirms the `document_templates` hex literal still loads correctly without the EOL scaffolding

### DIFF
--- a/library/classes/Installer.class.php
+++ b/library/classes/Installer.class.php
@@ -471,38 +471,26 @@ class Installer
             return false;
         }
 
-        // Track whether the previous fgets() chunk reached an actual line
-        // boundary (ended in "\n") or was a partial chunk of a long line that
-        // exceeded the 1024-byte buffer. We can only insert a separator at
-        // real line boundaries — inserting one mid-line would break tokens
-        // like hex literals (0x...) that are split across chunks.
-        $prevReachedEol = true;
         while (!$this->atEndOfFile($fd)) {
-            $rawLine = $this->getLine($fd, 1024);
+            $rawLine = $this->getLine($fd);
             if ($rawLine === false) {
                 continue;
             }
-            $reachedEol = str_ends_with($rawLine, "\n");
             $line = rtrim($rawLine);
             if ($line === "" || str_starts_with($line, "--") || str_starts_with($line, "#")) {
-                // A skipped line still counts as a line boundary if it
-                // actually ended in a newline.
-                if ($reachedEol) {
-                    $prevReachedEol = true;
-                }
                 continue;
             }
 
-            // Insert a single space separator only when we are starting a
-            // new logical line (previous chunk ended in newline) and there
-            // is already buffered query text. This fixes #10935 — without
-            // a separator, a continuation line at column 0 would fuse with
-            // the previous token (e.g. "= 0" + "WHERE" → "0WHERE").
-            if ($prevReachedEol && $query !== "") {
+            // Insert a single space separator between concatenated lines.
+            // Without it, a continuation line starting at column 0 would
+            // fuse with the previous token (e.g. "= 0" + "WHERE" → "0WHERE",
+            // see #10935). getLine() is unbounded, so every call returns a
+            // complete logical line and the separator is always safe to
+            // insert.
+            if ($query !== "") {
                 $query .= " ";
             }
             $query .= $line;
-            $prevReachedEol = $reachedEol;
             $chr = substr($query, strlen($query) - 1, 1);
             if ($chr == ";") { // valid query, execute
                 $query = rtrim($query, ";");
@@ -2140,12 +2128,11 @@ SETHLP;
      * @codeCoverageIgnore
      *
      * @param resource $stream
-     * @param int $length
      * @return string|false
      */
-    protected function getLine($stream, int $length): string|false
+    protected function getLine($stream): string|false
     {
-        return fgets($stream, $length);
+        return fgets($stream);
     }
 
     /**
--- a/tests/Tests/Isolated/library/classes/InstallerTest.php
+++ b/tests/Tests/Isolated/library/classes/InstallerTest.php
@@ -807,7 +807,7 @@ class InstallerTest extends TestCase
 
         $mockInstaller->expects($this->exactly(2))
             ->method('getLine')
-            ->with($mockResource, 1024)
+            ->with($mockResource)
             ->willReturnOnConsecutiveCalls(
                 "CREATE TABLE users;",
                 "INSERT INTO users VALUES (1, 'admin');"
@@ -882,7 +882,7 @@ class InstallerTest extends TestCase
 
         $mockInstaller->expects($this->once())
             ->method('getLine')
-            ->with($mockResource, 1024)
+            ->with($mockResource)
             ->willReturn("CREATE TABLE users;");
 
         $mockInstaller->expects($this->exactly(3))
@@ -3409,8 +3409,8 @@ class InstallerTest extends TestCase
      *
      *   UPDATE layout_options SET uor = 0WHERE form_id = 'DEM'
      *
-     * The fix inserts a single space between concatenated lines at real
-     * line boundaries, restoring valid token separation.
+     * The fix inserts a single space between concatenated lines, restoring
+     * valid token separation.
      */
     public function testLoadFileInsertsSpaceSeparatorBetweenSqlLines(): void
     {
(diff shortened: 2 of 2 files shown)


library/classes/Installer.class.php
```diff
@@ -471,38 +471,26 @@ public function load_file(string $filename, string $title): string|false
...
-        $prevReachedEol = true;
         while (!$this->atEndOfFile($fd)) {
-            $rawLine = $this->getLine($fd, 1024);
+            $rawLine = $this->getLine($fd);
             if ($rawLine === false) {
                 continue;
             }
-            $reachedEol = str_ends_with($rawLine, "\n");
             $line = rtrim($rawLine);
             if ($line === "" || str_starts_with($line, "--") || str_starts_with($line, "#")) {
```
kojiromike, line 480: is this comment detection worth it? What if a line starts with space, and then has a comment character? OTOH does it do any harm to give mysql a comment here?

library/classes/Installer.class.php
```diff
@@ -2140,12 +2128,11 @@ protected function fileExists(string $fileName): bool
...
      *
      * @param resource $stream
-     * @param int $length
      * @return string|false
      */
-    protected function getLine($stream, int $length): string|false
+    protected function getLine($stream): string|false
     {
-        return fgets($stream, $length);
+        return fgets($stream);
```
kojiromike, line 2135: is fgets the right, modern way to read a file/line?

library/classes/Installer.class.php
kojiromike: Great question, and the answer turned up a latent bug. Filed as #11465.

**Short version:** the comment check is worth it for column-zero `--`/`#` comments (which is what the bundled dumps mostly have), but it silently breaks on indented ones — exactly the case you flagged.

A line like `    -- comment` does not match `str_starts_with($line, "--")` because of the leading whitespace, so it is not skipped. It gets appended to the current statement buffer. Because `load_file()` strips the newline and concatenates with a space, the result sent to MySQL is `    -- comment    NEXT STATEMENT;`. MySQL's `--` comment runs to end-of-line, and the concatenated buffer has no end-of-line, so the whole thing is parsed as a comment and the statement is silently dropped.

`sql/database.sql` already has this pattern. Lines 15190 and 15214 (introduced in #9275 in November 2025) both have indented `--` comments immediately preceding statements. The 15214 one precedes an `INSERT INTO list_options` for `treatment_intervention_preferences` that, as far as I can tell, has never actually been inserted on any fresh install via this code path.

So: "does it do any harm to give mysql a comment here?" — unfortunately, yes, because mysql cannot tell where the comment ends once we strip the newlines. The real fix (in #11465) is either to concatenate with `\n` instead of a space (so mysql can terminate `--` naturally, and we can delete the comment check) or to skip the whole hand-rolled parser via #11458.

Keeping this PR scoped to "drop the 1024 buffer" — the comment-handling fix is a separate, more invasive change.

library/classes/Installer.class.php
kojiromike: Fair question. Short answer: `fgets()` without a length limit is still the right, standard, line-oriented read primitive in modern PHP. The options are:

- **`fgets($fd)`** — what this PR leaves us with. Reads one line per call, streams the file, zero dependencies, every PHP version supports it.
- **`SplFileObject`** — OO wrapper, iterable (`foreach ($file as $line)`), slightly nicer API. Not strictly better here; the class has some sharp edges around how it handles the trailing newline and EOF, and it adds no real value over `fgets` for this call site.
- **`file($filename)`** — returns all lines as an array. Eager, loads the entire file into memory up front. Fine for small files but we already took the "it's fine to read the whole thing into memory" step — at which point `file_get_contents()` is a better fit.
- **`file_get_contents()` + parsing** — this is where #11458 lands. `mysqli_multi_query()` wants one big string, not a line iterator, so the whole `getLine()` wrapper goes away.

So I'd leave `fgets()` here. The broader "modernize the I/O and delete this hand-rolled parser" work is exactly what #11458 tracks, and it replaces the loop wholesale rather than swapping one line-reader for another. This PR is scoped to "drop the 1024 buffer and the scaffolding it required" — changing the I/O primitive would be scope creep for no real benefit.

Issue #11474, opened 2026-04-07
refactor(direct-messaging): handle fgets() false return on every phiMail protocol read


kojiromike:
## Background

The phiMail protocol parsers in `library/direct_message_check.inc.php` and `ccr/transmitCCD.php` call `fgets($fp)` at many points and immediately use the result as a string — passing it to `str_starts_with()`, `explode()`, loose comparison with `==`, or `phimail_read_blob()` as a byte count — without checking whether the read actually succeeded.

`fgets()` returns `false` on EOF, error, or connection drop. In PHP 8+ this means:

- `str_starts_with(false, ...)` → `TypeError`
- `explode(';', false)` → `TypeError`
- `false == "NONE\n"` is `false` so the path "works", but `false == 0` is `true`, which can mask attachment-count desyncs
- `phimail_read_blob($fp, false)` likely yields a 0-byte temp file rather than a recognized read failure

This is **pre-existing** — it has nothing to do with the recent `fgets()` length-limit removal in #11470. The same `false` return was possible with the original 256/512/1024 length arguments. Removing the length argument did not introduce the issue, but does make it more visible because the surrounding code is now under closer scrutiny.

## Confirmed sites in `library/direct_message_check.inc.php`

Flagged by copilot's review on #11470:

- **Line 180** — `$ret = fgets($fp);` followed by `str_starts_with($ret, "STATUS")` and `explode(" ", trim($ret), 4)` in the CHECK loop. False would TypeError.
- **Line 287** — `$mime_type = fgets($fp);` followed by `explode(";", $mime_type)`. False would TypeError.
- **Line 292** — `$body_len = fgets($fp);` passed straight into `phimail_read_blob($fp, $body_len)` as the byte count. Should validate it's a non-empty digit string and coerce to int.
- **Line 300** — `$att2 = fgets($fp);` compared with `$att` via loose `==`. If the read fails and `$att === 0`, the comparison succeeds and a desync is masked.
- **Line 366** — `$att_len = fgets($fp);` passed into `phimail_read_blob()` as a length without validation, same as line 292.

The `attinfo` reads at lines 311–313 already use `=== false` checks, and the body/attachment blob reads at lines 360 already check `=== false`. The pattern is inconsistent across the file — some reads are guarded, most are not.

## Likely sites in `ccr/transmitCCD.php`

[...]

Issue #11463, opened 2026-04-07
fix(direct-messaging): drop fgets() length limits in phiMail protocol parsers


kojiromike:
## Background

Two files read phiMail Direct Messaging protocol responses with fixed-length `fgets()`:

- `library/direct_message_check.inc.php` — nine call sites (lines 180, 283, 287, 292, 300, 311–313, 360, 366, 494) using lengths of 512 and 1024.
- `ccr/transmitCCD.php` — six call sites (lines 75, 81, 207, 226, 252, 276) using lengths of 256, 512, and 1024.

These are essentially the same protocol parser copy-pasted into two files. Both use exact-string equality to test responses, e.g.:

```php
$ret = fgets($fp, 512);
if ($ret == "NONE\n") { ... }
```

and

```php
if ($ret != "BEGIN\n") { ... }
```

If the phiMail server ever sends a response line longer than the buffer, `fgets()` returns a mid-line chunk, the equality check fails, and the protocol state machine goes off the rails. Email headers (`Received:`, `DKIM-Signature:`, long `Message-ID`) routinely exceed 1 KB in the wild, and the header-reading loop in `direct_message_check.inc.php:283` uses:

```php
while (($next_hdr = fgets($fp, 1024)) != "\n")
```

A long header causes the loop to treat a mid-header chunk as a new header, corrupting MIME parsing.

This has likely been working by accident for years — whoever runs Direct Messaging in production has not hit a server sending long enough lines to expose it. It is a latent protocol bug, not a hypothetical one.

## Fix

Drop every `fgets()` length argument in both files. `fgets()` without a length reads until newline or EOF, so every call returns a complete logical line. All the exact-string comparisons then work as written.

## Scope

Mechanical: ~15 one-line edits across two files. No test coverage exists for these protocol paths as far as I can tell — verify during the PR. If there are no tests, do not block on adding them; this is a bug fix for latent behavior.

## Risk

Low. `fgets()` without a length is strictly more correct for line-oriented protocol reads. The only way this change could regress behavior is if the current code was relying on `fgets()` to bail out early on oversized responses, and nothing in the surrounding logic suggests that.

## Related

- Found by the bycatch audit during #11461 (`Installer::getLine()` fgets buffer removal).
- Same root cause as #10935 / #11003 / #11459, but in a completely different subsystem.

PR #11476, opened 2026-04-07
refactor(direct-messaging): wrap phiMail socket lifecycle in try/finally

### LINKED ISSUE #11471: refactor(direct-messaging): wrap phiMail socket lifecycle in try/finally to prevent handle leaks
## Background

The phiMail Direct Messaging protocol clients in `ccr/transmitCCD.php` and `library/direct_message_check.inc.php` open a socket via `phimail_connect()` and rely on scattered `phimail_close($fp)` calls along every return path to release it. Several early-return paths skip the close entirely, so any protocol-level error leaks the socket handle until the request ends.

This was noticed as bycatch while reviewing #11463 / #11470 (the `fgets()` length-limit fix). The same two files are involved.

### `ccr/transmitCCD.php::transmitMessage()`

`phimail_connect()` at line 57, `phimail_close()` at line 107. Between them, these early returns skip the close:

- line 67 — AUTH failed
- line 72 — TO failed (`RECIPIENT_NOT_ALLOWED`)
- line 89 — TEXT OK check failed
- line 95 — SET FINAL 1 failed
- line 100 — SET FINAL 0 failed

Five real leak paths.

### `ccr/transmitCCD.php::transmitCCD()`

Same shape. `phimail_connect()` at line 189, `phimail_close()` at line 278. Early returns at 199/204/234/246/260/271 skip the close.

### `library/direct_message_check.inc.php::phimail_check()`

`phimail_connect()` at the top of the function followed by a `while (1)` loop with many conditional branches. `phimail_close()` appears in several places inside the loop, but the control flow is complex enough that it is hard to argue any specific path is safe — exactly the situation try/finally exists for.

## Fix

Wrap the section from `phimail_connect()` through the last use of `$fp` in a try/finally, and move `phimail_close($fp)` into the `finally` block. Delete the scattered in-body `phimail_close()` calls. The intent and control flow become obvious:

```php
$fp = phimail_connect($err);
if ($fp === false) {
    return "$config_err $err";
}
try {
    // all the existing logic, minus phimail_close() calls
} finally {
    phimail_close($fp);
}
```

## Scope

Mechanical refactor of three functions across two files. No behavior change on the happy path. On error paths, the socket is now deterministically released.

## Related

- #11463 / #11470 — `fgets()` length-limit fix in the same files; the audit that surfaced this.
- Larger resource-cleanup audit pending — this issue covers phiMail specifically because the two files are tightly coupled and share a copy-pasted parser.

kojiromike:
## Summary

Fixes #11471.

The phiMail Direct Messaging protocol clients in `ccr/transmitCCD.php` (`transmitMessage`, `transmitCCD`) and `library/direct_message_check.inc.php` (`phimail_check`) opened a socket via `phimail_connect()` and relied on scattered `phimail_close($fp)` calls along every return path to release it. Many early-return paths skipped the close entirely, so any protocol-level error — AUTH failed, TO failed, SET FINAL failed, SHOW failed, mid-loop SQL failure, header read failure, attachment read/store failure, and more — leaked the socket handle until the request ended.

This PR wraps the body of each function from `phimail_connect()` through the last use of `$fp` in a `try`/`finally`, and moves `phimail_close($fp)` into the `finally` block. Every in-body `phimail_close()` call is deleted. The control flow becomes obvious and the socket is deterministically released on every exit path, including unexpected exceptions from any of the helpers (`phimail_read_blob`, `phimail_store`, SQL helpers, `addPnote`, `setGpRelation`, etc.).

## Narrow try/finally scope

The crypto setup (`$phimail_username` / `$cryptoGen` / `$phimail_password`) and the `$notifyUsername` fallback in `phimail_check()` are hoisted above `phimail_connect()` so the `try` block contains only socket-using code. The principle — anything that can be computed before the resource is acquired does not belong inside the `try` — is now documented as a coding convention in ocemr-docs.

## Related cleanups

While reindenting the affected blocks, this PR also addresses review feedback on the surrounding code quality:

- **SET FINAL duplication.** Collapse the `SET FINAL 0/1` if/else duplication in both `transmitMessage()` and `transmitCCD()` into a single ternary on the flag value — the two branches differed only in `'0'` vs `'1'`.
- **Format dispatch.** Convert `transmitCCD()`'s `$format_type` if/elseif/elseif/else chain to a `match` expression. Full enum conversion for `$format_type` and `$xml_type` is tracked separately in #11483 because it touches every caller.
- **Narrow, don't cast.** Replace `(string) $url` / `strrpos((string) $url, ...)` in `phimail_check()` with an explicit `is_string($url)` narrow plus log + return on the false case. This incidentally drops two pre-existing entries each from the `binaryOp.invalid.php` and `cast.string.php` PHPStan baselines.

## `phimail_write_expect_OK` no longer closes

`phimail_write_expect_OK()` previously called `phimail_close($fp)` internally on the non-OK path. With the lifecycle now managed by `try`/`finally` in the caller, having the helper close as a side effect was both misleading and caused an actual pre-existing double-close bug in several paths (where the caller then explicitly called `phimail_close()` again on top of the helper's internal close). The helper is now pure "write + check OK" — callers own the socket lifecycle.

Correspondingly, `phimail_connect()`'s own INFO VER failure path (the `C5` error branch) now explicitly closes before setting `$fp = false`, since that site used to rely on `phimail_write_expect_OK`'s internal close.

## Scope

Mechanical. No behavior change on the happy path. On every error path the socket is closed exactly once, where previously it was either leaked or double-closed. The diff is dominated by reindentation inside the `try` blocks; the logic itself is unchanged.

## Test plan

- [x] `php -l` clean on both files
- [x] PHPStan level 10, phpcs, rector, composer-require-checker — all pre-commit hooks pass
- [ ] Manual exercise of Direct Messaging requires a phiMail server; relying on review for the protocol logic and CI for the static checks

## Related

- #11463 / #11470 — the `fgets()` length-limit fix in the same files; this issue was bycatch from that audit.
- #11474 — unchecked `fgets()` `false` returns in the same parsers (still open; independent of this PR).
- #11483 — follow-up: convert `$format_type` and `$xml_type` to backed enums (surfaced during review of this PR).

### DIFF
--- a/ccr/transmitCCD.php
+++ b/ccr/transmitCCD.php
@@ -54,57 +54,54 @@ function transmitMessage($message, $recipient, $verifyFinalDelivery = false)
         return("$config_err " . ErrorConstants::ERROR_CODE_MESSAGING_DISABLED);
     }
 
+    $phimail_username = OEGlobalsBag::getInstance()->getString('phimail_username');
+    $cryptoGen = ServiceContainer::getCrypto();
+    $phimail_password = $cryptoGen->decryptStandard(OEGlobalsBag::getInstance()->getString('phimail_password'));
+
     $fp = phimail_connect($err);
     if ($fp === false) {
         return("$config_err $err");
     }
 
-    $phimail_username = OEGlobalsBag::getInstance()->getString('phimail_username');
-    $cryptoGen = ServiceContainer::getCrypto();
-    $phimail_password = $cryptoGen->decryptStandard(OEGlobalsBag::getInstance()->getString('phimail_password'));
-    $ret = phimail_write_expect_OK($fp, "AUTH $phimail_username $phimail_password\n");
-    if ($ret !== true) {
-        return("$config_err " . ErrorConstants::ERROR_CODE_AUTH_FAILED);
-    }
+    try {
+        $ret = phimail_write_expect_OK($fp, "AUTH $phimail_username $phimail_password\n");
+        if ($ret !== true) {
+            return("$config_err " . ErrorConstants::ERROR_CODE_AUTH_FAILED);
+        }
 
-    $ret = phimail_write_expect_OK($fp, "TO $recipient\n");
-    if ($ret !== true) {
-        return( xl(ErrorConstants::RECIPIENT_NOT_ALLOWED) . " " . $ret );
-    }
+        $ret = phimail_write_expect_OK($fp, "TO $recipient\n");
+        if ($ret !== true) {
+            return( xl(ErrorConstants::RECIPIENT_NOT_ALLOWED) . " " . $ret );
+        }
 
-    $ret = fgets($fp); //ignore extra server data
+        $ret = fgets($fp); //ignore extra server data
 
-    $text_out = $message;
+        $text_out = $message;
 
-    $text_len = strlen((string) $text_out);
-    phimail_write($fp, "TEXT $text_len\n");
-    $ret = @fgets($fp);
-    if ($ret != "BEGIN\n") {
+        $text_len = strlen((string) $text_out);
+        phimail_write($fp, "TEXT $text_len\n");
+        $ret = @fgets($fp);
+        if ($ret != "BEGIN\n") {
+            return("$config_err " . ErrorConstants::ERROR_CODE_MESSAGE_BEGIN_FAILED);
+        }
+
+        $ret = phimail_write_expect_OK($fp, $text_out);
+        if ($ret !== true) {
+            return("$config_err " . ErrorConstants::ERROR_CODE_MESSAGE_BEGIN_OK_FAILED);
+        }
+
+        $finalFlag = $verifyFinalDelivery ? '1' : '0';
+        $ret = phimail_write_expect_OK($fp, "SET FINAL $finalFlag\n");
+        if ($ret !== true) {
+            return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
+        }
+
+        phimail_write($fp, "SEND\n");
+        $ret = fgets($fp);
+        phimail_write($fp, "OK\n");
+    } finally {
         phimail_close($fp);
-        return("$config_err " . ErrorConstants::ERROR_CODE_MESSAGE_BEGIN_FAILED);
     }
-
-    $ret = phimail_write_expect_OK($fp, $text_out);
-    if ($ret !== true) {
-        return("$config_err " . ErrorConstants::ERROR_CODE_MESSAGE_BEGIN_OK_FAILED);
-    }
-
-    if ($verifyFinalDelivery) {
-        $ret = phimail_write_expect_OK($fp, "SET FINAL 1\n");
-        if ($ret !== true) {
-            return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
-        }
-    } else {
-        $ret = phimail_write_expect_OK($fp, "SET FINAL 0\n");
-        if ($ret !== true) {
-            return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
-        }
-    }
-
-    phimail_write($fp, "SEND\n");
-    $ret = fgets($fp);
-    phimail_write($fp, "OK\n");
-    phimail_close($fp);
 
 
     if (substr($ret, 5) == "ERROR") {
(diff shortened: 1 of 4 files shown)


ccr/transmitCCD.php
```diff
@@ -59,52 +59,54 @@ function transmitMessage($message, $recipient, $verifyFinalDelivery = false)
...
-    if ($ret !== true) {
-        return("$config_err " . ErrorConstants::ERROR_CODE_AUTH_FAILED);
-    }
-
-    $ret = phimail_write_expect_OK($fp, "TO $recipient\n");
-    if ($ret !== true) {
-        return( xl(ErrorConstants::RECIPIENT_NOT_ALLOWED) . " " . $ret );
-    }
+    try {
+        $phimail_username = OEGlobalsBag::getInstance()->getString('phimail_username');
```
kojiromike: yikes, does _all this_ have to go in try? We should have a principle of only putting into try what we actually need. For example, if we are trying to avoid forgetting to close a resource, then we shouldn't have the try statement include stuff that happens before we open that resource.

kojiromike: Agreed, the principle is right. Pushed `35e5b13`: the `$phimail_username` / `$cryptoGen` / `$phimail_password` setup doesn't touch `$fp`, so I hoisted it above `phimail_connect()` in all three functions. Same for `$notifyUsername` in `phimail_check()`. After this change, each `try` block contains only socket-using code, and the visual scope of the resource lifetime matches the code's intent. The text-assembly in `transmitCCD` stays inside the try because it's interleaved with `phimail_write()` calls — moving it out would fragment the protocol sequence.

ccr/transmitCCD.php
```diff
@@ -54,57 +54,60 @@ function transmitMessage($message, $recipient, $verifyFinalDelivery = false)
...
+            $ret = phimail_write_expect_OK($fp, "SET FINAL 1\n");
+            if ($ret !== true) {
+                return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
+            }
+        } else {
+            $ret = phimail_write_expect_OK($fp, "SET FINAL 0\n");
+            if ($ret !== true) {
+                return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
+            }
+        }
```
kojiromike: This block is very repetitive. Isn't the `SET FINAL 0/1` the only difference? Write this better.

ccr/transmitCCD.php
```diff
@@ -186,96 +189,98 @@ function transmitCCD($pid, $ccd_out, $recipient, $requested_by, $xml_type = "CCD
...
+        if ($format_type == 'html') {
+            $add_type = "TEXT";
+        } elseif ($format_type == 'pdf') {
+            $add_type = "RAW";
+        } elseif ($format_type == 'xml') {
+            $add_type = $xml_type == "CCR" ? "CCR" : "CDA";
+        } else {
+            // unsupported format
+            return ("$config_err " . ErrorConstants::ERROR_CODE_INVALID_FORMAT_TYPE);
+        }
```
kojiromike: $format_type should be an enum and this should use a match expression.

ccr/transmitCCD.php
```diff
@@ -186,96 +189,98 @@ function transmitCCD($pid, $ccd_out, $recipient, $requested_by, $xml_type = "CCD
...
+            $ret = phimail_write_expect_OK($fp, "SET FINAL 1\n");
+            if ($ret !== true) {
+                return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
+            }
+        } else {
+            $ret = phimail_write_expect_OK($fp, "SET FINAL 0\n");
+            if ($ret !== true) {
+                return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
+            }
         }
```
kojiromike: we saw this code before and I commented on it above.

library/direct_message_check.inc.php
```diff
@@ -155,336 +156,324 @@ function phimail_connect(&$phimail_error)
...
+                        $doc_id = phimail_store($name, $mime_type_main, $body);
+                        if (!$doc_id) {
+                            phimail_logit(0, "M7 store non-text body failed");
+                            return;
+                        }
+
+                        $idnum = $doc_id['doc_id'];
+                        $all_doc_ids[] = $idnum;
+                        $url = $doc_id['url'];
+                        $url = substr((string) $url, strrpos((string) $url, "/") + 1);
```
kojiromike: narrow, don't cast

ccr/transmitCCD.php
kojiromike: Fixed in `4decb42`. The two branches differed only in `'1'` vs `'0'`, so collapsed to:

```php
$finalFlag = $verifyFinalDelivery ? '1' : '0';
$ret = phimail_write_expect_OK($fp, "SET FINAL $finalFlag\n");
if ($ret !== true) {
    return( xl(ErrorConstants::ERROR_MESSAGE_SET_DISPOSITION_NOTIFICATION_FAILED) . " " . $ret );
}
```

Applied to both `transmitMessage()` and `transmitCCD()`.

ccr/transmitCCD.php
kojiromike: Same fix applied here (`4decb42`).

ccr/transmitCCD.php
kojiromike: Converted the if/elseif chain to a `match` expression in `4decb42`:

```php
$add_type = match ($format_type) {
    'html' => "TEXT",
    'pdf' => "RAW",
    'xml' => $xml_type == "CCR" ? "CCR" : "CDA",
    default => null,
};
if ($add_type === null) {
    return ("$config_err " . ErrorConstants::ERROR_CODE_INVALID_FORMAT_TYPE);
}
```

The full enum conversion (both `$format_type` and `$xml_type`) is tracked separately in #11483 because it touches every caller of `transmitCCD()` and adds two enum class files — too much scope for this PR. With proper enums the `default => null` branch and the `ERROR_CODE_INVALID_FORMAT_TYPE` path go away entirely and the `match` becomes exhaustive.

library/direct_message_check.inc.php
kojiromike: Fixed in `4decb42`. Replaced `(string) $url` and `strrpos((string) $url, "/")` with an explicit narrow:

```php
$url = $doc_id['url'];
if (!is_string($url)) {
    phimail_logit(0, "M7 store non-text body returned non-string url");
    return;
}
$url = substr($url, strrpos($url, "/") + 1);
```

This drops two pre-existing entries each from `binaryOp.invalid.php` and `cast.string.php` baselines (regenerated in the same commit). The deeper fix is to type `phimail_store()`'s return properly so `$doc_id['url']` is `string` at the source — that would be a much bigger touch and is left for a follow-up.

PR #11488, opened 2026-04-08
refactor(database): migrate sqlBeginTrans call sites to QueryUtils::inTransaction

### LINKED ISSUE #11472: refactor(db): guard sqlBeginTrans with try/finally so exceptions don't abandon open transactions
## Background

Several call sites open a database transaction with `sqlBeginTrans()` and commit with `sqlCommitTrans()` without any `try`/`finally` or `catch` protecting the range in between. If any statement between the two throws, the transaction is never committed or rolled back — it sits open on the connection until the request ends.

Depending on the MySQL isolation level and connection pooling, this can hold row locks, block other writers, or leave partial work visible to subsequent queries in the same request.

Noticed during a resource-cleanup audit that started from #11463 / #11470 / #11471.

## Confirmed example

`interface/orders/receive_hl7_results.inc.php` around lines 1205–1230:

```php
sqlBeginTrans();
$procedure_order_seq = sqlQuery(
    "SELECT IFNULL(MAX(procedure_order_seq),0) + 1 AS increment FROM procedure_order_code "
    . "WHERE procedure_order_id = ? ",
    [$in_orderid]
);
sqlInsert(
    "INSERT INTO procedure_order_code SET "
    . "procedure_order_id = ?, procedure_order_seq = ?, procedure_code = ?, "
    . "procedure_name = ?, procedure_type = ?, transport = ?, procedure_source = '2'",
    [$in_orderid, $procedure_order_seq['increment'], $in_procedure_code, $in_procedure_name, $code_type, $code_transport]
);
$pcrow = sqlQuery($pcquery, $pcqueryargs);
sqlCommitTrans();
```

If any of those three SQL calls throws, `sqlCommitTrans()` is never reached and no `sqlRollbackTrans()` runs.

## Other likely sites

A grep audit surfaced several other `sqlBeginTrans`/`sqlCommitTrans` pairs that appear to have the same shape. These should be spot-checked as part of this ticket rather than taken on faith:

- `interface/forms/procedure_order/procedure_order_save_functions.php`
- `src/Services/PatientTransactionService.php`
- `src/Billing/BillingUtilities.php`
- `interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/CcdaServiceDocumentRequestor.php`

## Fix shape

The minimal fix is the standard try/catch/rethrow:

```php
sqlBeginTrans();
try {
    // existing body
    sqlCommitTrans();
} catch (\Throwable $e) {
    sqlRollbackTrans();
    throw $e;
}
```

A nicer long-term fix is a `withTransaction(callable $body)` helper in the database layer so call sites never have to remember the pattern. Worth discussing as part of the ticket, not a prerequisite.

## Scope

Mechanical, file-by-file. No behavior change on the happy path. On exception paths, transactions are deterministically rolled back instead of abandoned.

## Related

- #11471 — phiMail socket cleanup (same resource-cleanup audit)

kojiromike:
## Summary

Fixes #11472.

Several call sites opened a database transaction with `sqlBeginTrans()` and committed with `sqlCommitTrans()` without `try`/`catch` protecting the range in between. If any statement between the two threw, the transaction was never committed or rolled back — it sat open on the connection until the request ended. Other sites had a manual `try`/`catch` around the same pattern, correct but verbose and inconsistent with the project's official `QueryUtils::inTransaction()` helper. The existing `openemr.deprecatedSqlFunction` PHPStan rule has been flagging every bare call via the baseline.

This PR migrates **every** bare `sqlBeginTrans` / `sqlCommitTrans` / `sqlRollbackTrans` call site to `QueryUtils::inTransaction(callable)`. The helper already does the right thing: start, run the closure, commit on success, rollback and rethrow on `\Throwable`.

## Sites migrated

| File | Was | Becomes |
|---|---|---|
| `src/Billing/BillingUtilities.php` | unprotected | wrapped |
| `src/Common/Uuid/UuidMapping.php` | manual try/catch/rethrow | collapsed |
| `src/Services/PatientTransactionService.php::insert` | unprotected + 2× `return false` leak paths | wrapped, false contract preserved via private sentinel exception (see below) |
| `src/Services/PatientTransactionService.php::update` | unprotected | wrapped |
| `interface/orders/receive_hl7_results.inc.php` | unprotected | wrapped |
| `interface/forms/procedure_order/procedure_order_save_functions.php` | unprotected | wrapped |
| `interface/forms/procedure_order/handle_deletions.php` | manual try/catch/rethrow | collapsed |
| `interface/modules/zend_modules/.../CodeTypeEventsSubscriber.php` (×2) | manual try/catch with catch-log-continue | inTransaction inside the existing catch-log-continue |
| `interface/modules/zend_modules/.../EncounterccdadispatchTable.php` | manual try/catch/rethrow | collapsed |
| `interface/modules/custom_modules/oe-module-dorn/src/ReceiveHl7Results.php` | unprotected | wrapped |

## `PatientTransactionService::insert` false-return contract

The current method has two `return false` early-exit paths *inside* the transaction body, which leak the transaction. Migrating to `inTransaction()` requires a decision on those paths because returning normally from the closure causes a commit.

Chosen behavior: throw a private sentinel exception (`PatientTransactionInsertFailedException`, `@internal` in `src/Services/`) from inside the closure to trigger rollback, then catch it immediately in `insert()` and return `false` — preserving the existing contract. Cleaner than sentinel magic codes on a generic `RuntimeException` because the class name documents intent and the catch can't accidentally swallow an unrelated exception.

## String-literal cleanup in touched code

Per review feedback, while the diff is already reindenting the SQL inside each closure, also bring the touched strings up to the project's string-literal conventions:

- **Multi-line concatenated SQL → nowdocs.** Every `"SELECT ..." . "FROM ..."`-style concatenation in the diff becomes a `<<<'SQL'` nowdoc (or a `<<<SQL` heredoc in the one place where a dynamic `$claimset` SQL fragment is interpolated — `BillingUtilities` claims INSERT).
- **Single-line double-quoted SQL → single-quoted.** `BillingUtilities`' version-lookup SELECT was double-quoted for no reason; now single-quoted.
- **Bare `sqlQuery`/`sqlStatement` → `QueryUtils::*`.** In `procedure_order_save_functions.php` and `CodeTypeEventsSubscriber.php`, replace with `QueryUtils::fetchSingleValue` and `QueryUtils::sqlStatementThrowException`.
- **`empty($code_id)` → `$code_id === null || $code_id === ''`.** Per the project's avoid-empty() rule.
- **`!empty($logger) && is_callable($logger)` → `is_callable($logger)`.** An empty value can never be callable, so the first half is redundant. All three occurrences in `CodeTypeEventsSubscriber` simplified.

## Baseline impact

- **21 entries dropped** from `.phpstan/baseline/openemr.deprecatedSqlFunction.php` (110 lines) — the rule was already telling us to migrate.
- **4 entries dropped** from `.phpstan/baseline/openemr.emptyIsNotAllowed.php` — from the explicit null/empty checks.
- **Additional drops** from `offsetAccess.nonOffsetAccessible.php` and related baselines as incidental side effects of the narrower types and the nowdoc conversions.

No new baseline entries. Tracked PHPStan rules are strictly reduced.

## Scope

Mechanical refactor with no behavior change on the happy path. On every error path the transaction is now deterministically rolled back instead of leaked. Uses the existing `QueryUtils::inTransaction()` helper rather than introducing a new abstraction.

## Test plan

- [x] `php -l` clean on all 12 changed files
- [x] PHPStan level 10, phpcs, rector, composer-require-checker — all pre-commit hooks pass
- [x] `composer phpstan-baseline` regenerated with only expected drops

## Related

- #11471 / #11476 — phiMail socket cleanup (parallel resource-cleanup work in the same audit)
- #11490 — CPT4 mapping "Skipping option_id" log + missing `continue` (pre-existing bug copilot surfaced during review of this PR; intentionally out of scope)
- opencoreemr/ocemr-docs#67 — "Narrow try/finally Scope" convention, motivated by this audit
- opencoreemr/ocemr-docs#69 — heredoc/nowdoc for embedded languages, motivated by this audit
- opencoreemr/ocemr-docs#70 — string quoting discipline, motivated by review of this PR

### DIFF
--- a/interface/forms/procedure_order/procedure_order_save_functions.php
+++ b/interface/forms/procedure_order/procedure_order_save_functions.php
@@ -10,6 +10,7 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Common\Uuid\UuidRegistry;
 
@@ -634,27 +635,32 @@ function saveProcedureAnswers($formid, $poseq, $ptid, $postData, $index): void
         }
 
         foreach ($data as $datum) {
-            sqlBeginTrans();
-            $answer_seq = sqlQuery(
-                "SELECT IFNULL(MAX(answer_seq), 0) + 1 AS increment
-                 FROM procedure_answers
-                 WHERE procedure_order_id = ?
-                   AND procedure_order_seq = ?
-                   AND question_code = ?",
-                [$formid, $poseq, $qcode]
-            );
+            QueryUtils::inTransaction(function () use ($formid, $poseq, $qcode, $datum, $pcode): void {
+                $answerSeq = QueryUtils::fetchSingleValue(
+                    <<<'SQL'
+                    SELECT IFNULL(MAX(answer_seq), 0) + 1 AS increment
+                    FROM procedure_answers
+                    WHERE procedure_order_id = ?
+                      AND procedure_order_seq = ?
+                      AND question_code = ?
+                    SQL,
+                    'increment',
+                    [$formid, $poseq, $qcode]
+                );
 
-            sqlStatement(
-                "INSERT INTO procedure_answers SET
-                 procedure_order_id = ?,
-                 procedure_order_seq = ?,
-                 question_code = ?,
-                 answer_seq = ?,
-                 answer = ?,
-                 procedure_code = ?",
-                [$formid, $poseq, $qcode, $answer_seq['increment'], trim((string) $datum), $pcode]
-            );
-            sqlCommitTrans();
+                QueryUtils::sqlStatementThrowException(
+                    <<<'SQL'
+                    INSERT INTO procedure_answers SET
+                        procedure_order_id = ?,
+                        procedure_order_seq = ?,
+                        question_code = ?,
+                        answer_seq = ?,
+                        answer = ?,
+                        procedure_code = ?
+                    SQL,
+                    [$formid, $poseq, $qcode, $answerSeq, trim((string) $datum), $pcode]
+                );
+            });
         }
     }
 }
--- a/interface/modules/custom_modules/oe-module-dorn/src/ReceiveHl7Results.php
+++ b/interface/modules/custom_modules/oe-module-dorn/src/ReceiveHl7Results.php
@@ -687,32 +687,37 @@ class ReceiveHl7Results
                         $lkup = $this->lookupTestCode($lab_id, $in_procedure_code);
                         $code_type = ($lkup['procedure_type'] ?? '') ? trim($lkup['procedure_type']) : '';
                         $code_transport = ($lkup['transport'] ?? '') ? trim($lkup['transport']) : '';
-                        sqlBeginTrans();
-                        $procedure_order_seq = sqlQuery(
-                            "SELECT IFNULL(MAX(procedure_order_seq),0) + 1 AS increment FROM procedure_order_code " .
-                            "WHERE procedure_order_id = ? ",
-                            [$in_orderid]
-                        );
-                        sqlInsert(
-                            "INSERT INTO procedure_order_code SET " .
-                            "procedure_order_id = ?, " .
-                            "procedure_order_seq = ?, " .
-                            "procedure_code = ?, " .
-                            "procedure_name = ?, " .
-                            "procedure_type = ?, " .
-                            "transport = ?, " .
-                            "procedure_source = '2'",
-                            [
-                                $in_orderid,
-                                $procedure_order_seq['increment'],
-                                $in_procedure_code,
-                                $in_procedure_name,
-                                $code_type,
-                                $code_transport
-                            ]
-                        );
-                        $pcrow = sqlQuery($pcquery, $pcqueryargs);
-                        sqlCommitTrans();
+                        $pcrow = QueryUtils::inTransaction(function () use ($in_orderid, $in_procedure_code, $in_procedure_name, $code_type, $code_transport, $pcquery, $pcqueryargs) {
+                            $procedure_order_seq = sqlQuery(
+                                <<<'SQL'
+                                SELECT IFNULL(MAX(procedure_order_seq), 0) + 1 AS increment
+                                FROM procedure_order_code
+                                WHERE procedure_order_id = ?
+                                SQL,
+                                [$in_orderid]
+                            );
+                            sqlInsert(
+                                <<<'SQL'
+                                INSERT INTO procedure_order_code SET
+                                    procedure_order_id = ?,
+                                    procedure_order_seq = ?,
+                                    procedure_code = ?,
+                                    procedure_name = ?,
+                                    procedure_type = ?,
+                                    transport = ?,
+                                    procedure_source = '2'
+                                SQL,
+                                [
+                                    $in_orderid,
+                                    $procedure_order_seq['increment'],
+                                    $in_procedure_code,
+                                    $in_procedure_name,
+                                    $code_type,
+                                    $code_transport,
+                                ]
+                            );
+                            return sqlQuery($pcquery, $pcqueryargs);
+                        });
                     } else {
                         // Dry run, make a dummy procedure_order_code row.
                         $pcrow = [
--- a/interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
+++ b/interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
@@ -227,23 +227,26 @@ class CodeTypeEventsSubscriber implements EventSubscriberInterface
     private function updateSNOMEDCTMappingsForList($mappings, $list_id, $logger = null)
     {
         // update our list options
+        $sql = null;
+        $values = null;
         try {
-            \sqlBeginTrans();
-            foreach ($mappings as $option_id => $code_id) {
-                $sql = "UPDATE list_options SET codes=CONCAT('SNOMED-CT:', ?) WHERE list_id=? AND option_id=?";
-                $values = [$code_id, $list_id, $option_id];
-                QueryUtils::sqlStatementThrowException($sql, $values);
-                if (!empty($logger) && is_callable($logger)) {
-                    $logger(xl('Success') . ' - (sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
+            QueryUtils::inTransaction(function () use ($mappings, $list_id, $logger, &$sql, &$values): void {
+                foreach ($mappings as $option_id => $code_id) {
+                    $sql = <<<'SQL'
+                    UPDATE list_options SET codes = CONCAT('SNOMED-CT:', ?) WHERE list_id = ? AND option_id = ?
+                    SQL;
+                    $values = [$code_id, $list_id, $option_id];
+                    QueryUtils::sqlStatementThrowException($sql, $values);
+                    if (is_callable($logger)) {
+                        $logger(xl('Success') . ' - (sql=`' . $sql . '`, values=`' . var_export($values, true) . '`)');
+                    }
                 }
-            }
-            \sqlCommitTrans();
+            });
         } catch (\Throwable $exception) {
             ServiceContainer::getLogger()->error($exception->getMessage(), ['exception' => $exception]);
-            if (!empty($logger) && is_callable($logger)) {
-                $logger(xl('Failed') . ' - (sql=`"' . ($sql ?? 'N/A') . '`, values=`' . var_export($values ?? [], true) . "`)");
+            if (is_callable($logger)) {
+                $logger(xl('Failed') . ' - (sql=`' . ($sql ?? 'N/A') . '`, values=`' . var_export($values ?? [], true) . '`)');
             }
-            \sqlRollbackTrans();
         }
     }
 
@@ -251,27 +254,35 @@ class CodeTypeEventsSubscriber implements EventSubscriberInterface
     {
         // update our list options
         try {
-            \sqlBeginTrans();
-            foreach (self::CPT4_ENCOUNTER_TYPE_MAPPINGS as $option_id => $code_text) {
-                $code_id = QueryUtils::fetchSingleValue("SELECT `code` FROM codes WHERE code_text =? "
-                    . " AND code_type IN (SELECT ct_id FROM code_types WHERE ct_key = 'CPT4')", 'code', [$code_text]);
-                if (empty($code_id)) {
-                    ServiceContainer::getLogger()->debug(
-                        "Failed to find cpt4 code in codes with code_text. Skipping option_id",
-                        ['code_text' => $code_text, 'option_id' => $option_id]
+            QueryUtils::inTransaction(function () use ($logger): void {
+                foreach (self::CPT4_ENCOUNTER_TYPE_MAPPINGS as $option_id => $code_text) {
+                    $code_id = QueryUtils::fetchSingleValue(
+                        <<<'SQL'
+                        SELECT `code` FROM codes
+                        WHERE code_text = ?
+                          AND code_type IN (SELECT ct_id FROM code_types WHERE ct_key = 'CPT4')
+                        SQL,
+                        'code',
+                        [$code_text]
                     );
+                    if ($code_id === null || $code_id === '') {
+                        ServiceContainer::getLogger()->debug(
+                            'Failed to find cpt4 code in codes with code_text. Skipping option_id',
+                            ['code_text' => $code_text, 'option_id' => $option_id]
+                        );
+                    }
+                    $sql = <<<'SQL'
+                    UPDATE list_options SET codes = CONCAT('CPT4:', ?) WHERE list_id = ? AND option_id = ?
+                    SQL;
+                    $values = [$code_id, self::LIST_ID_ENCOUNTER_TYPES, $option_id];
+                    if (is_callable($logger)) {
+                        $logger('(sql=`' . $sql . '`, values=`' . var_export($values, true) . '`)');
+                    }
+                    QueryUtils::sqlStatementThrowException($sql, $values);
                 }
-                $sql = "UPDATE list_options SET codes=CONCAT('CPT4:', ?) WHERE list_id=? AND option_id=?";
-                $values = [$code_id, self::LIST_ID_ENCOUNTER_TYPES, $option_id];
-                if (!empty($logger) && is_callable($logger)) {
-                    $logger('(sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
-                }
-                QueryUtils::sqlStatementThrowException($sql, $values);
-            }
-            \sqlCommitTrans();
+            });
         } catch (\Throwable $exception) {
             ServiceContainer::getLogger()->error($exception->getMessage(), ['exception' => $exception]);
-            \sqlRollbackTrans();
         }
     }
 }
--- a/interface/orders/receive_hl7_results.inc.php
+++ b/interface/orders/receive_hl7_results.inc.php
@@ -1202,32 +1202,37 @@ function receive_hl7_results(&$hl7, &$matchreq, $lab_id = 0, $direction = 'B', $
                     $lkup = lookupTestCode($lab_id, $in_procedure_code);
                     $code_type = ($lkup['procedure_type'] ?? '') ? trim($lkup['procedure_type']) : '';
                     $code_transport = ($lkup['transport'] ?? '') ? trim($lkup['transport']) : '';
-                    sqlBeginTrans();
-                    $procedure_order_seq = sqlQuery(
-                        "SELECT IFNULL(MAX(procedure_order_seq),0) + 1 AS increment FROM procedure_order_code " .
-                        "WHERE procedure_order_id = ? ",
-                        [$in_orderid]
-                    );
-                    sqlInsert(
-                        "INSERT INTO procedure_order_code SET " .
-                        "procedure_order_id = ?, " .
-                        "procedure_order_seq = ?, " .
-                        "procedure_code = ?, " .
-                        "procedure_name = ?, " .
-                        "procedure_type = ?, " .
-                        "transport = ?, " .
-                        "procedure_source = '2'",
-                        [
-                            $in_orderid,
-                            $procedure_order_seq['increment'],
-                            $in_procedure_code,
-                            $in_procedure_name,
-                            $code_type,
-                            $code_transport
-                        ]
-                    );
-                    $pcrow = sqlQuery($pcquery, $pcqueryargs);
-                    sqlCommitTrans();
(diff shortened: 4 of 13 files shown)


interface/forms/procedure_order/procedure_order_save_functions.php
```diff
@@ -634,27 +635,27 @@ function saveProcedureAnswers($formid, $poseq, $ptid, $postData, $index): void
...
-                 procedure_order_seq = ?,
-                 question_code = ?,
-                 answer_seq = ?,
-                 answer = ?,
-                 procedure_code = ?",
-                [$formid, $poseq, $qcode, $answer_seq['increment'], trim((string) $datum), $pcode]
-            );
-            sqlCommitTrans();
+            QueryUtils::inTransaction(function () use ($formid, $poseq, $qcode, $datum, $pcode): void {
+                $answer_seq = sqlQuery(
```
kojiromike: if we're changing this anyway, why not change to QueryUtils everywhere? Also, really, prefer a nowdoc or heredoc to these multiline string forms.

src/Services/PatientTransactionService.php
```diff
@@ -133,19 +133,19 @@ public function getAll($pid)
...
+                throw new \RuntimeException('Failed to insert patient transaction');
+            }
+
+            $lbtDataId = $this->insertTransactionForm($transactionId, $data);
+            if ($lbtDataId == false) {
+                throw new \RuntimeException('Failed to insert patient transaction form');
+            }
+
+            return ["id" => $transactionId, "form_id" => $lbtDataId];
+        });
```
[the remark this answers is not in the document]
kojiromike: You're right that the unilateral contract change was the wrong call. Fixed in `8df2191`.

Same general shape as your suggestion (throw inside the closure, catch around `inTransaction`, return false), but used a private sentinel exception class (`PatientTransactionInsertFailedException`, marked `@internal`) rather than magic exception codes — clearer intent and no risk of catching an unrelated `RuntimeException` from somewhere deeper in the stack:

```php
public function insert($pid, $data)
{
    try {
        return QueryUtils::inTransaction(function () use ($pid, $data) {
            $transactionId = $this->insertTransaction($pid, $data);
            if ($transactionId == false) {
                throw new PatientTransactionInsertFailedException('Failed to insert patient transaction');
            }
            $lbtDataId = $this->insertTransactionForm($transactionId, $data);
            if ($lbtDataId == false) {
                throw new PatientTransactionInsertFailedException('Failed to insert patient transaction form');
            }
            return ["id" => $transactionId, "form_id" => $lbtDataId];
        });
    } catch (PatientTransactionInsertFailedException) {
        return false;
    }
}
```

The fact that `TransactionRestController::CreateTransaction` doesn't actually handle the false case (it passes the result straight to `ProcessingResult::addData`, which then produces a 201 response with garbage data) is a separate concern. I'll file a follow-up issue for the controller's error handling rather than bundling it with the resource-cleanup migration.

interface/forms/procedure_order/procedure_order_save_functions.php
kojiromike: Done in `8df2191`. Replaced the bare `sqlQuery` + array dereference with `QueryUtils::fetchSingleValue`, the bare `sqlStatement` with `QueryUtils::sqlStatementThrowException`, and reformatted both SQL strings as nowdocs (no interpolation, so `<<<'SQL'` is the right form). Drops two more pre-existing entries from `openemr.deprecatedSqlFunction.php` and one from `offsetAccess.nonOffsetAccessible.php` as an incidental cleanup.

Adding the nowdoc-for-long-embedded-strings convention to ocemr-docs in a separate PR right after this.

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
```diff
@@ -227,51 +227,51 @@ private function updateSNOMEDCTMappings($logger = null)
...
-                if (!empty($logger) && is_callable($logger)) {
-                    $logger(xl('Success') . ' - (sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
+            QueryUtils::inTransaction(function () use ($mappings, $list_id, $logger, &$sql, &$values): void {
+                foreach ($mappings as $option_id => $code_id) {
+                    $sql = "UPDATE list_options SET codes=CONCAT('SNOMED-CT:', ?) WHERE list_id=? AND option_id=?";
+                    $values = [$code_id, $list_id, $option_id];
+                    QueryUtils::sqlStatementThrowException($sql, $values);
+                    if (!empty($logger) && is_callable($logger)) {
+                        $logger(xl('Success') . ' - (sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
+                    }
```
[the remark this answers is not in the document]
kojiromike: Not a bug, and not touched by this PR.

The source is a single-quoted PHP string: `' - (sql=` `"'`. PHP single-quoted strings only interpret two escape sequences: `\\` and `\'`. Every other character — including `"` — is a literal. So `' - (sql=` `"'` contains the characters ` - (sql=` `"` with no backslash. The log output is `Success - (sql=` `"<SQL>` `, values=` `<array>` `)` with literal double-quotes, not `\"`.

Also, this specific log line is unchanged content from master — the diff only reindents it one more level because of the closure body, but the string literal itself is identical. If there were a real quoting bug here it would be a pre-existing issue outside the scope of this resource-cleanup migration.

(The `"` inside the first backtick pair without a matching `"` before the second backtick pair *is* asymmetric, but that's a pre-existing stylistic quirk in the log format string, not a PHP-string-escaping bug, and also not part of this PR.)

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
```diff
@@ -227,51 +227,51 @@ private function updateSNOMEDCTMappings($logger = null)
...
                 }
-            }
-            \sqlCommitTrans();
+            });
         } catch (\Throwable $exception) {
             ServiceContainer::getLogger()->error($exception->getMessage(), ['exception' => $exception]);
             if (!empty($logger) && is_callable($logger)) {
                 $logger(xl('Failed') . ' - (sql=`"' . ($sql ?? 'N/A') . '`, values=`' . var_export($values ?? [], true) . "`)");
             }
```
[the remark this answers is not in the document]
kojiromike: Same answer as the parallel thread above: `"` inside a PHP single-quoted string is a literal double-quote, not `\"`. PHP single-quoted strings only process `\\` and `\'` as escapes; every other character is a literal. No backslash is emitted in the log output. And this line is unchanged content from master — the reindent shifted the line number but the string literal is byte-identical.

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
```diff
@@ -227,51 +227,51 @@ private function updateSNOMEDCTMappings($logger = null)
...
+                            "Failed to find cpt4 code in codes with code_text. Skipping option_id",
+                            ['code_text' => $code_text, 'option_id' => $option_id]
+                        );
+                    }
+                    $sql = "UPDATE list_options SET codes=CONCAT('CPT4:', ?) WHERE list_id=? AND option_id=?";
+                    $values = [$code_id, self::LIST_ID_ENCOUNTER_TYPES, $option_id];
+                    if (!empty($logger) && is_callable($logger)) {
+                        $logger('(sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
+                    }
+                    QueryUtils::sqlStatementThrowException($sql, $values);
```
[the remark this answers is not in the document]
kojiromike: Same answer as the two parallel threads: `"` inside a PHP single-quoted string is a literal double-quote. `'(sql=` `"'` contains the characters `(sql=` `"` with no backslash — PHP only escapes `\\` and `\'` inside single quotes. The log output is `(sql=` `"<SQL>` `, values=` `<array>` `)` with literal double-quotes. And this line is unchanged from master; only the indentation shifted.

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
```diff
@@ -227,51 +227,51 @@ private function updateSNOMEDCTMappings($logger = null)
...
-                    $logger('(sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
+            QueryUtils::inTransaction(function () use ($logger): void {
+                foreach (self::CPT4_ENCOUNTER_TYPE_MAPPINGS as $option_id => $code_text) {
+                    $code_id = QueryUtils::fetchSingleValue("SELECT `code` FROM codes WHERE code_text =? "
+                        . " AND code_type IN (SELECT ct_id FROM code_types WHERE ct_key = 'CPT4')", 'code', [$code_text]);
+                    if (empty($code_id)) {
+                        ServiceContainer::getLogger()->debug(
+                            "Failed to find cpt4 code in codes with code_text. Skipping option_id",
+                            ['code_text' => $code_text, 'option_id' => $option_id]
+                        );
```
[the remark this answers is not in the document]
kojiromike: Valid — the log message says "Skipping" but the loop actually falls through and runs the UPDATE with an empty `$code_id`, clobbering the target `list_options.codes` with a sentinel `CPT4:` string. Real latent bug.

Pre-existing though: my diff only reindents this block inside the `inTransaction` closure; the missing `continue` is unchanged from master. Tracking the one-line fix in #11490 rather than expanding the scope of this resource-cleanup PR.

interface/modules/custom_modules/oe-module-dorn/src/ReceiveHl7Results.php
```diff
@@ -687,32 +687,32 @@ private function receiveHl7Results(&$hl7, &$matchreq, $lab_id = 0, $direction =
...
-                                $code_type,
-                                $code_transport
-                            ]
-                        );
-                        $pcrow = sqlQuery($pcquery, $pcqueryargs);
-                        sqlCommitTrans();
+                        $pcrow = QueryUtils::inTransaction(function () use ($in_orderid, $in_procedure_code, $in_procedure_name, $code_type, $code_transport, $pcquery, $pcqueryargs) {
+                            $procedure_order_seq = sqlQuery(
+                                "SELECT IFNULL(MAX(procedure_order_seq),0) + 1 AS increment FROM procedure_order_code " .
+                                "WHERE procedure_order_id = ? ",
```
kojiromike: It's silly to use concatenation to join literal strings over a code newline. Don't do that. I don't care if it was written this way before. Reformat this properly using a nowdoc. Do that for other stuff that was already changed in this pr.

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
```diff
@@ -227,51 +227,51 @@ private function updateSNOMEDCTMappings($logger = null)
...
-            \sqlBeginTrans();
-            foreach ($mappings as $option_id => $code_id) {
-                $sql = "UPDATE list_options SET codes=CONCAT('SNOMED-CT:', ?) WHERE list_id=? AND option_id=?";
-                $values = [$code_id, $list_id, $option_id];
-                QueryUtils::sqlStatementThrowException($sql, $values);
-                if (!empty($logger) && is_callable($logger)) {
-                    $logger(xl('Success') . ' - (sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
+            QueryUtils::inTransaction(function () use ($mappings, $list_id, $logger, &$sql, &$values): void {
+                foreach ($mappings as $option_id => $code_id) {
+                    $sql = "UPDATE list_options SET codes=CONCAT('SNOMED-CT:', ?) WHERE list_id=? AND option_id=?";
```
kojiromike: Respect the difference between single and double quotes, especially for injection-sensitive contexts like sql. That is, if it doesn't need interpolation, make interpolation impossible by using single quotes.

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
```diff
@@ -227,51 +227,51 @@ private function updateSNOMEDCTMappings($logger = null)
...
-                }
-                $sql = "UPDATE list_options SET codes=CONCAT('CPT4:', ?) WHERE list_id=? AND option_id=?";
-                $values = [$code_id, self::LIST_ID_ENCOUNTER_TYPES, $option_id];
-                if (!empty($logger) && is_callable($logger)) {
-                    $logger('(sql=`"' . $sql . '`, values=`' . var_export($values, true) . "`)");
+            QueryUtils::inTransaction(function () use ($logger): void {
+                foreach (self::CPT4_ENCOUNTER_TYPE_MAPPINGS as $option_id => $code_text) {
+                    $code_id = QueryUtils::fetchSingleValue("SELECT `code` FROM codes WHERE code_text =? "
+                        . " AND code_type IN (SELECT ct_id FROM code_types WHERE ct_key = 'CPT4')", 'code', [$code_text]);
+                    if (empty($code_id)) {
```
kojiromike: empty is not allowed

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
```diff
@@ -227,51 +227,51 @@ private function updateSNOMEDCTMappings($logger = null)
...
+                        ServiceContainer::getLogger()->debug(
+                            "Failed to find cpt4 code in codes with code_text. Skipping option_id",
+                            ['code_text' => $code_text, 'option_id' => $option_id]
+                        );
+                    }
+                    $sql = "UPDATE list_options SET codes=CONCAT('CPT4:', ?) WHERE list_id=? AND option_id=?";
+                    $values = [$code_id, self::LIST_ID_ENCOUNTER_TYPES, $option_id];
+                    if (!empty($logger) && is_callable($logger)) {
```
kojiromike: if logger were empty, could it be callable?

interface/orders/receive_hl7_results.inc.php
```diff
@@ -1202,32 +1202,32 @@ function receive_hl7_results(&$hl7, &$matchreq, $lab_id = 0, $direction = 'B', $
...
-                            $in_procedure_name,
-                            $code_type,
-                            $code_transport
-                        ]
-                    );
-                    $pcrow = sqlQuery($pcquery, $pcqueryargs);
-                    sqlCommitTrans();
+                    $pcrow = QueryUtils::inTransaction(function () use ($in_orderid, $in_procedure_code, $in_procedure_name, $code_type, $code_transport, $pcquery, $pcqueryargs) {
+                        $procedure_order_seq = sqlQuery(
+                            "SELECT IFNULL(MAX(procedure_order_seq),0) + 1 AS increment FROM procedure_order_code " .
```
kojiromike: definitely rewrite this using nowdocs

src/Billing/BillingUtilities.php
```diff
@@ -1673,30 +1674,30 @@ public static function updateClaim(
...
-                    "patient_id = ?, " .
-                    "encounter_id = ?, " .
-                    "bill_time = NOW(), status=? ," .
-                    "version = ?";
-                array_push($sqlBindArray, $status, $version['increment']);
-            }
+                $sqlBindArray = [];
+                array_push($sqlBindArray, $patient_id, $encounter_id);
+                if ($crossover <> 1) {
+                    $sql = "INSERT INTO claims SET " .
```
kojiromike: nowdocs!

interface/modules/custom_modules/oe-module-dorn/src/ReceiveHl7Results.php
kojiromike: Fair — pushed `05553c2` converting every multi-line concatenated SQL string in the PR's diff to a nowdoc. Covered: the SELECT/INSERT pair here in `oe-module-dorn/ReceiveHl7Results.php`, the parallel SELECT/INSERT pair in `interface/orders/receive_hl7_results.inc.php`, the two INSERTs in `BillingUtilities` (one stays a heredoc — not nowdoc — because it interpolates a dynamic `$claimset` SQL fragment), and the SNOMED-CT / CPT4 `UPDATE list_options` strings plus the CPT4 lookup SELECT in `CodeTypeEventsSubscriber`. Same commit also fixes the other comments on this file/PR.

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
kojiromike: Fixed in `05553c2`. Both `UPDATE list_options` strings (the `SNOMED-CT:` one here and the `CPT4:` one below) are now nowdocs — the SQL contains literal single quotes around the prefix, so either a single-quoted PHP string with escaped inner quotes or a nowdoc would work; the nowdoc avoids the backslashes entirely and makes interpolation impossible, which matches the convention.

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
kojiromike: Fixed in `05553c2`. Replaced `if (empty($code_id))` with the explicit `if ($code_id === null || $code_id === '')` check per the project's avoid-empty() rule. Drops four pre-existing entries from `.phpstan/baseline/openemr.emptyIsNotAllowed.php` as a side effect.

(The "Skipping option_id" log + missing `continue` is a separate pre-existing bug I filed as #11490 during the earlier review round — leaving that out of this commit to keep the scope of this PR on resource cleanup.)

interface/modules/zend_modules/module/CodeTypes/src/CodeTypes/Listener/CodeTypeEventsSubscriber.php
kojiromike: No. `is_callable(null)`, `is_callable('')`, `is_callable(0)`, `is_callable(false)` all return false, so `!empty($logger) && is_callable($logger)` is equivalent to just `is_callable($logger)`. Dropped the redundant `!empty()` check from all three occurrences in this file in `05553c2`.

interface/orders/receive_hl7_results.inc.php
kojiromike: Done in `05553c2`. Both the SELECT (increment lookup) and the INSERT INTO procedure_order_code are now nowdocs.

src/Billing/BillingUtilities.php
kojiromike: Done in `05553c2`. Both `INSERT INTO claims` strings are now heredoc/nowdoc form — the first stays a heredoc (not nowdoc) because it interpolates the dynamic `$claimset` SQL fragment into the `bill_time = NOW() $claimset` clause, which is a pre-existing dynamic-SQL-fragment pattern independent of this PR. The parallel INSERT without interpolation is a nowdoc. Also converted the single-line SELECT above them from double-quoted to single-quoted (no interpolation, no reason to leave it double-quoted).

Issue #9078, opened 2025-10-07
FHIR Write Operations: Encounter


kojiromike:
Enable write operations (POST/PUT) for FHIR Encounter resource. Encounter is the core clinical workflow resource and high priority for interoperability.

## Current State
- ✅ FhirEncounterService exists with read operations (`parseOpenEMRRecord`)
- ✅ EncounterValidator exists
- ❌ Uses `FhirServiceBaseEmptyTrait` - stub implementations for write operations
- ❌ EncounterService (776 lines) has NO insert/update methods
- ❌ Missing `parseFhirResource()` to convert FHIR → OpenEMR format
- ❌ Comment in controller: `// implement put post in future` (FhirEncounterRestController.php:31)

## Complexity
**High** - Complex data model with participants, locations, periods, hospitalization

## Estimated Time
**8-12 hours**

## Dependencies
- Patient ✅ (already writable)
- Practitioner ✅ (already writable)
- Location (see issue #103)

## Blocks
- Many clinical resources reference encounters (Observation, Condition, Procedure, etc.)

### 1. Service Layer - EncounterService
- Add `insert()` method to EncounterService
- Add `update()` method to EncounterService
- Handle complex encounter data model (participants, locations, periods)
- File: `src/Services/EncounterService.php`

### 2. Service Layer - FhirEncounterService
- Implement `parseFhirResource()` to convert FHIR → OpenEMR format
  - Parse encounter status, class, type
  - Parse period (start/end dates)
  - Parse participants (practitioners, their roles)
  - Parse locations
  - Parse hospitalization data
- Implement `insertOpenEMRRecord()` in FhirEncounterService
- Implement `updateOpenEMRRecord()` in FhirEncounterService
- Remove `FhirServiceBaseEmptyTrait` usage
- File: `src/Services/FHIR/FhirEncounterService.php`

### 3. REST Controller
- Add `post()` method to FhirEncounterRestController
- Add `put()` or `patch()` method for updates
- Include FhirValidationService validation
- File: `src/RestControllers/FHIR/FhirEncounterRestController.php`
- Remove `// implement put post in future` comment

### 4. Route Registration
- Add POST /fhir/Encounter route with ACL check
- Add PUT /fhir/Encounter/:uuid route with ACL check
- File: `apis/routes/_rest_routes_fhir_r4_us_core_3_1_0.inc.php`

### 5. Testing
- Create/update integration tests
- Test ACL enforcement (own encounters vs all encounters)
- Test data validation
[...]

Issue #9076, opened 2025-10-07
Enable Write Operations for FHIR Resources


kojiromike:
Enable write operations (POST/PUT) for FHIR resources in OpenEMR. Currently only Patient, Practitioner, and Organization support write operations. This epic tracks the implementation of write support for all remaining clinical FHIR resources.

## Current State
- **Write-enabled resources** (3): Patient, Practitioner, Organization
- **Read-only resources** (20+): All other FHIR resources
- **Infrastructure**: FhirServiceBase provides insert/update abstractions, FhirValidationService handles FHIR validation, RestConfig provides ACL integration via GACL

## Success Criteria
- All prioritized FHIR resources support POST (create) and PUT (update) operations
- All write operations enforce appropriate ACL permissions via GACL
- All write operations validate input data and prevent data loss
- Integration tests verify functionality and security
- Code follows existing OpenEMR/FHIR patterns

## Dependencies & Prerequisites
The following resources must be writable before others can reference them:
- **Patient** ✅ (already implemented) - Required by almost all clinical resources
- **Practitioner** ✅ (already implemented) - Required by Encounter, Appointment, etc.
- **Organization** ✅ (already implemented) - Required by many resources
- **Location** - Required by Encounter, Appointment

#### 1. Location
- **Why**: Required by Encounter and Appointment
- **Complexity**: Low-Medium (simple resource)
- **Estimate**: 6-8 hours
- **Current State**:
  - ✅ FhirLocationService exists with read operations
  - ❌ Uses `FhirServiceBaseEmptyTrait` - no write operations
  - ❌ No LocationService with insert/update methods
- **ACL**: "admin"/"super" (administrative resource)
- **Blocks**: Encounter, Appointment

#### 2. Encounter
- **Why**: Core clinical workflow, high interoperability priority
- **Complexity**: High (complex data model with participants, locations, periods)
- **Estimate**: 8-12 hours
- **Current State**:
  - ✅ FhirEncounterService exists with read operations
  - ✅ EncounterValidator exists
  - ❌ EncounterService (776 lines) has NO insert/update methods
  - ❌ Missing `parseFhirResource()` implementation
  - ❌ Comment in controller: `// implement put post in future`
- **ACL**: "encounters"/"auth" (own encounters) or "encounters"/"auth_a" (all encounters)
- **Dependencies**: Patient, Practitioner, Location
- **Enables**: Many clinical resources reference encounters

#### 3. AllergyIntolerance
[...]

Issue #8138, opened 2025-03-15
feat: Add onset date and recordedDate to FHIR AllergyIntolerance resource


adunsulag:
### Describe the problem

As an API consumer I would like to be able to see the onset date and recordedDate when I retrieve a list of AllergyIntolerance resources.  We are already tracking that information in the EMR so we should go ahead and populate those values.

You can see the datatypes in the Snapshot Table here: https://hl7.org/fhir/us/core/STU6.1/StructureDefinition-us-core-allergyintolerance.html

PR #11507, opened 2026-04-09
feat(fhir): add R4 write support for 16 resources plus 7 documented deferrals

rollingventures:
## Summary

- Adds FHIR R4 **create (POST)** and **update (PUT)** support for **5 clinical resources** that previously only supported read operations:
  - **AllergyIntolerance** — POST + PUT
  - **Condition** — POST + PUT
  - **Encounter** — POST + PUT
  - **Appointment** — POST only (underlying service lacks full update method)
  - **Immunization** — POST + PUT (service layer built from stubs)
- Adds generic `post()`/`put()` methods to `FhirGenericRestController` so future resources can reuse the same write path
- Includes CRUD tests and FHIR R4 JSON test fixtures for each resource
- Builds out `ImmunizationService::insert()`/`update()` and `ImmunizationValidator` rules (previously stubbed as "Method not implemented yet")

## Context

OpenEMR's FHIR API supports reads for 35 resources but writes for only 3 (Patient, Organization, Practitioner). This PR begins closing that gap by enabling writes for clinical resources whose underlying OpenEMR services support (or now support) write operations.

Coverage write support was evaluated but deferred — the `CoverageValidator` requires 15+ subscriber demographic fields (name, DOB, address, etc.) that FHIR Coverage doesn't carry. Those belong in RelatedPerson/Patient resources and need a design decision from maintainers.

Refs #9076 (FHIR write operations epic), #9078 (Encounter writes), #8138 (AllergyIntolerance dates)

### Infrastructure
- **`FhirGenericRestController`** — new `post()`, `put()`, `deserializeFhirResource()` methods with FHIR validation, resourceType/class_exists guards, ACL checks, patient-portal blocking, and session passthrough to services

### AllergyIntolerance (POST/PUT)
- `parseFhirResource()` via `jsonSerialize()` — maps patient, code, clinicalStatus, criticality, verificationStatus, reaction, onset, notes
- Delegates to `AllergyIntoleranceService::insert()`/`update()`

### Condition (POST/PUT)
- `parseFhirResource()` — maps category, code (SNOMED-CT/ICD-10/ICD-9), subject, clinical/verification status, onset/abatement dates
- Delegates to `ConditionService::insert()`/`update()`

### Encounter (POST/PUT)
- `parseFhirResource()` — maps multi-participant handling (primary performer vs referrer), class, period, reasonCode, serviceProvider→facility_id, hospitalization discharge disposition
- Resolves `provider_uuid`→`provider_id`, `referrer_uuid`→`referring_provider_id`, injects `user`/`group` from session
- Dynamic `pc_catid` lookup via `openemr_postcalendar_categories` (no hardcoded IDs)
- Delegates to `EncounterService::insertEncounter()`/`updateEncounter()`

### Appointment (POST only)
- `parseFhirResource()` — reverse status mapping (13 FHIR→OpenEMR statuses), appointmentType→pc_catid lookup, participant reference resolution (Patient→pid, Practitioner→pc_aid, Location→pc_facility), start/end→duration calculation
- POST only — `AppointmentService` has `insert()` but no full `update()` method
- Delegates to `AppointmentService::insert()`

### Immunization (POST/PUT) — service layer built from scratch
- **`ImmunizationService`** — implemented `insert()` and `update()` (previously stubbed as "Method not implemented yet")
- **`ImmunizationValidator`** — added validation rules: `patient_id`, `cvx_code`, `administered_date` required for insert; `uuid` required for update
- `parseFhirResource()` — maps vaccineCode, status, lotNumber, expirationDate, site, doseQuantity, performer, encounter
- Delegates to `ImmunizationService::insert()`/`update()`

### Routes
- 11 new routes in `_rest_routes_fhir_r4_us_core_3_1_0.inc.php` with OpenAPI annotations
- Capability statement auto-updates from route definitions

### Key design decisions
- All `parseFhirResource()` methods use `$fhirResource->jsonSerialize()` to work with normalized arrays — the FHIR R4 PHP library (dcarbone/php-fhir) does not deeply hydrate nested objects, so calling type-specific getters on sub-elements causes "method on array" errors
- `instanceof` guards on all `parseFhirResource()` methods reject mismatched resource types
- `deserializeFhirResource()` validates `resourceType` is non-empty and resolves to a known class

### Not included (deferred)
- **Coverage** — `CoverageValidator` requires subscriber demographics not in FHIR Coverage
- **~15 other resources** (CarePlan, CareTeam, Device, Goal, Location, Medication, etc.) — underlying OpenEMR services lack write methods entirely

## Test plan

- [x] phpcs passes (all files)
- [x] phpstan level 10 passes (0 errors)
- [x] rector passes (0 changes needed)
- [x] semgrep passes (0 findings)
- [ ] `FhirAllergyIntoleranceServiceCrudTest` — insert, insertWithErrors, update, updateWithErrors
- [ ] `FhirConditionServiceCrudTest` — insert, insertWithErrors, update, updateWithErrors
- [ ] `FhirEncounterServiceCrudTest` — insert, insertWithErrors, updateWithErrors
- [ ] `FhirAppointmentServiceCrudTest` — insert, insertWithErrors
- [ ] `FhirImmunizationServiceCrudTest` — insert, insertWithErrors, update, updateWithErrors
- [ ] Verify `GET /fhir/metadata` shows new "create"/"update" interactions
- [ ] Existing FHIR read tests still pass (no regressions)


src/RestControllers/FHIR/FhirGenericRestController.php
```diff
@@ -147,4 +152,106 @@ public function getAll(?array $searchParams = null): JsonResponse|Response
...
+        if (!class_exists($className)) {
+            return RestControllerHelper::responseHandler(
+                UtilsService::createOperationOutcomeResource('error', 'invalid', 'Unknown resourceType: ' . $resourceType),
+                null,
+                400
+            );
+        }
+
+        unset($fhirJson['resourceType']);
+        return new $className($fhirJson);
```
adunsulag, line 460: The challenge with the data population here in the constructor is that unfortunately the existing FHIR resources (which were generated from a specific version of the fhir carbon tool) is that nested child resources are often not hydrated as PHP objects and instead remain their array equivalents which leads to problematic behaviors. I believe a later version of R4 resources generated via php carbon rectifies this problem but it'd be a bit of undertaking to update the resources to do that (https://github.com/dcarbone/php-fhir).   

You may get away with this for primitive values, but anything nested will require some additional deserialization logic or updating those resources to the newer versions of dcarbon.

rollingventures: Good catch — we hit exactly this issue in CI (`Call to a member function getReference() on array`). 

The fix (in commit b8eade8) switches all three `parseFhirResource()` implementations to use `$fhirResource->jsonSerialize()` to get a normalized array, then works with that directly instead of calling type-specific getters on the partially-hydrated objects. This avoids the hydration problem entirely while keeping the `instanceof` guard for type safety at the top of each method.

The `deserializeFhirResource()` in the controller still uses the constructor for validation purposes (it catches `InvalidArgumentException` and `Error` from unknown fields), but the actual field extraction in the service layer never relies on nested objects being hydrated.

src/RestControllers/FHIR/FhirGenericRestController.php
```diff
@@ -146,4 +150,75 @@ public function getAll(?array $searchParams = null): JsonResponse|Response
...
+            return RestControllerHelper::responseHandler($validationResult, null, 400);
+        }
+
+        $resourceType = $fhirJson['resourceType'] ?? '';
+        $className = 'OpenEMR\FHIR\R4\FHIRDomainResource\FHIR' . $resourceType;
+        unset($fhirJson['resourceType']);
+        $fhirResource = new $className($fhirJson);
+
+        $processingResult = $this->getFhirService()->insert($fhirResource);
+        return RestControllerHelper::handleFhirProcessingResult($processingResult, 201);
```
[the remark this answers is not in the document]
kojiromike: The `instanceof` guards in each service's `parseFhirResource()` now catch mismatches and throw `InvalidArgumentException`. However, this propagates as an unhandled 500 rather than a clean 400 OperationOutcome. Consider catching `InvalidArgumentException` in the controller's `post()`/`put()` methods and returning a proper 400 response.

src/Services/FHIR/FhirAllergyIntoleranceService.php
```diff
@@ -259,6 +260,184 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+
+        // Category is always 'medication' in OpenEMR (US Core requires it)
+        // We accept it from the FHIR resource but don't need to store it since
+        // OpenEMR allergy type is always 'allergy' in the lists table
+
+        // Patient reference -> puuid (required in US Core)
+        $patient = $fhirResource->getPatient();
+        if ($patient && $patient->getReference()) {
+            $reference = (string) $patient->getReference();
+            $data['puuid'] = str_replace('Patient/', '', $reference);
```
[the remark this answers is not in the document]
kojiromike: Still using `str_replace('Patient/', '', ...)` in the current code. This also affects `FhirConditionService` and `FhirEncounterService` (for Patient, Practitioner, and Organization references). FHIR references can be absolute URLs, so `str_replace` will produce garbage for those. Consider using `UtilsService::getUuidFromReference()` or at minimum extracting the last path segment.

src/Services/FHIR/FhirConditionService.php
```diff
@@ -129,6 +130,153 @@ public function getAll($fhirSearchParameters, $puuidBind = null): ProcessingResu
...
+                }
+            }
+        }
+
+        // Subject -> puuid
+        $subject = $fhirResource->getSubject();
+        if ($subject && $subject->getReference()) {
+            $reference = (string) $subject->getReference();
+            $data['puuid'] = str_replace('Patient/', '', $reference);
+        }
```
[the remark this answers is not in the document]
kojiromike: Same as the AllergyIntolerance thread — `str_replace('Patient/', '', ...)` is still used here. Applies to subject reference extraction.

kojiromike (commented):
Nice progress on the FHIR write operations. Many of the earlier Copilot findings have been addressed — resolved those threads. The `jsonSerialize()` approach for handling partially-hydrated resources is a clean solution.

A few new items beyond the still-open `str_replace` and route-mismatch threads:

### Sanitize `resourceType` before class instantiation (`FhirGenericRestController`)

`$resourceType` comes from user-supplied JSON and is used to construct a class name:
```php
$className = 'OpenEMR\FHIR\R4\FHIRDomainResource\FHIR' . $resourceType;
```
The `class_exists` check limits the blast radius, but backslashes in `$resourceType` could navigate the namespace tree. A simple guard like `/^[A-Za-z]+$/` on `$resourceType` before constructing the class name would close this off.

### PUT should validate body `id` matches URL `$fhirId`

The `put()` method passes `$fhirId` from the URL to `update()`, but never checks that the resource body's `id` (if present) matches. FHIR spec requires servers to reject updates where the body id differs from the URL id. A mismatch would silently use the URL's id, which could confuse clients.

### `UuidRegistry::uuidToBytes()` on unsanitized reference values

In `resolveProviderUuids()` and similar, UUIDs extracted via `str_replace` from FHIR references are passed directly to `UuidRegistry::uuidToBytes()`. A malformed reference like `Practitioner/not-a-uuid` will throw an uncaught exception. Wrap in try/catch or validate UUID format first.

### `empty()` usage in new code

Several places use `empty()` (e.g., `empty($resourceType)`, `!empty($json['id'])`). At PHPStan level 10 with strict rules, `empty()` is flagged because `empty("0")` returns `true`. Prefer explicit checks like `$resourceType === ''` or `$resourceType === null`.

src/RestControllers/FHIR/FhirGenericRestController.php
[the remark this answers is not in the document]
kojiromike: Still open — the `instanceof` guards catch mismatches but the resulting `InvalidArgumentException` propagates as a 500 instead of a clean 400 OperationOutcome. Please either add a catch in the controller or explain why a 500 is acceptable for client-supplied type mismatches.

src/Services/FHIR/FhirAllergyIntoleranceService.php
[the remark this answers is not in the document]
kojiromike: Still open — `str_replace('Patient/', '', ...)` doesn't handle absolute FHIR references. Please fix with a code change (e.g., `UtilsService::getUuidFromReference()` or extracting the last path segment) or explain why only relative references need to be supported.

src/Services/FHIR/FhirConditionService.php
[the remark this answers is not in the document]
kojiromike: Still open — same `str_replace` issue as the AllergyIntolerance thread. Please address with a code change or explain why this is acceptable.

src/RestControllers/FHIR/FhirGenericRestController.php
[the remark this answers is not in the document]
kojiromike: cc @Firehed do we have a good standard way to make exceptions result in the right http status code here?

src/Services/FHIR/FhirAllergyIntoleranceService.php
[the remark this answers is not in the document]
kojiromike: There's already a standard helper for this: `UtilsService::parseReference()` (`src/Services/FHIR/UtilsService.php:447-470`). It uses `parse_url()` to handle both relative (`Patient/uuid`) and absolute (`https://example.org/fhir/Patient/uuid`) references, and returns `['localResource' => bool, 'uuid' => ?string, 'type' => ?string]`.

It's already used in `FhirPatientService` and `FhirQuestionnaireResponseFormService`. The caveat is it expects a `FHIRReference` object, so if you're working from the jsonSerialized array you'd need to construct one or extract the reference string and use a similar `parse_url` + `explode` approach directly.

src/RestControllers/FHIR/FhirGenericRestController.php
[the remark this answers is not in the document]
kojiromike: For a concrete fix path: other FHIR controllers (e.g., `FhirPatientRestController::post()`) handle validation failures by returning `RestControllerHelper::responseHandler($errors, null, 400)`. The `InvalidArgumentException` from `parseFhirResource()` currently bubbles all the way up to `HttpRestRouteHandler`, which wraps any uncaught `\Throwable` as a 500.

The simplest fix would be to catch `InvalidArgumentException` in `FhirGenericRestController::post()` and `put()` and return a 400:

```php
try {
    $processingResult = $this->getFhirService()->insert($fhirResource);
} catch (\InvalidArgumentException $e) {
    return RestControllerHelper::responseHandler(
        ['validationErrors' => [$e->getMessage()]],
        null,
        400
    );
}
```

This is consistent with the existing validation error flow and keeps the `instanceof` guards in the service layer where they belong.

apis/routes/_rest_routes_fhir_r4_us_core_3_1_0.inc.php
```diff
@@ -394,7 +404,10 @@
      */
     "POST /fhir/Condition" => function (HttpRestRequest $request, OEGlobalsBag $globalsBag) {
         RestConfig::request_authorization_check($request, "patients", "med");
-        $data = (array) (json_decode(file_get_contents("php://input"), true));
+        $data = RestControllerHelper::parseJsonRequestBody(true);
+        if ($data instanceof \Symfony\Component\HttpFoundation\Response) {
```
adunsulag: We have a policy that class definitions go in the use statement instead of inline in the code.

adunsulag (changes requested):
Only a minor change on the namespace usage.  Other than that it looks fine to me.

kojiromike:
@aisle-analyzer

rollingventures:
## FHIR R4 specification cross-check + roadmap

To validate the write paths added in this PR, I ran a manual cross-check against the FHIR R4 resource specifications at [hl7.org/fhir/R4](https://hl7.org/fhir/R4/resourcelist.html). This comment documents what was checked, what was found, and what remains.

### Verification approach

1. **Layer 1 — resource list:** Confirmed every resource name we touched exists in R4 with correct spelling and maturity level (cross-referenced against [hl7.org/fhir/R4/resourcelist.html](https://hl7.org/fhir/R4/resourcelist.html)). No typos, no mismatches.
2. **Layer 2 — per-resource structural conformance:** Fetched the R4 spec page for each of the 11 resources to which writes were added, and compared required-field cardinality (1..1 / 1..*), element types, Reference targets, and modifier-element status against our `parseFhirResource` extractions.

### Resources shipped with full write support (POST/PUT)

| Resource | R4 required fields | Our impl handles | Notes |
|---|---|---|---|
| AllergyIntolerance | (existing) | ✅ | Pre-existing |
| Appointment | (existing) | ✅ | Pre-existing |
| Condition | (existing) | ✅ | Pre-existing |
| Encounter | (existing) | ✅ | Pre-existing |
| Immunization | (existing) | ✅ | Pre-existing |
| Organization | (existing) | ✅ | Pre-existing |
| Patient | (existing) | ✅ | Pre-existing |
| Practitioner | (existing) | ✅ | Pre-existing |
| Coverage | status, beneficiary, payor | ✅ all three | See *Documented spec narrowings* below |
| MedicationRequest | status, intent, medication[x], subject | ✅ all four | See *Documented spec narrowings* |
| CarePlan | status, intent, subject | ✅ status + subject; intent has storage gap | See *Documented spec narrowings* |
| Person | (none required at top level) | n/a — we add NPI requirement | OpenEMR practical constraint, documented in route |
| Medication | (none required at top level) | n/a — we add `name` requirement | OpenEMR `drugs.name` is NOT NULL |
| Device | (none required at top level) | n/a — we add `patient` requirement | OpenEMR `lists.pid` is NOT NULL |
| RelatedPerson | patient | ✅ patient required; we also require relationship | Practical constraint for read-side JOIN |
| PractitionerRole | (none required at top level) | n/a — we add practitioner+organization requirement | OpenEMR `facility_user_ids` keys on (uid, facility_id) |
| ServiceRequest | status, intent, subject | ✅ status + subject; intent has storage gap | See *Documented spec narrowings* |
| Goal | lifecycleStatus, description, subject | ✅ all three | Subject restricted to Patient (R4 allows Group/Organization) |
| CareTeam | (none required at top level) | n/a — we add subject requirement | OpenEMR `care_teams.pid` is the partitioning column |

### Documented spec narrowings

Four real deviations from R4 where the OpenEMR data model is narrower than the spec allows. None are bugs in the write code — they're surfaced honestly in the OpenAPI route descriptions so clients don't get round-trip surprises:

- **Coverage.payor:** R4 allows `Reference(Organization | Patient | RelatedPerson)`. OpenEMR's `insurance_companies.id` only models Organization payors. Non-Organization references will 422.
- **MedicationRequest.medication[x]:** R4 allows `medicationCodeableConcept` OR `medicationReference`. OpenEMR only persists the inline coding; a `medicationReference` is dropped on write. Clients should use `medicationCodeableConcept` with RxNorm for round-trip compatibility.
- **CarePlan.intent (1..1 required):** OpenEMR's `form_care_plan` has no column for intent. The read side hardcodes `intent='plan'`; the write silently treats other values as `'plan'` on round-trip. Information loss but not a parse failure.
- **ServiceRequest.intent (1..1 required):** OpenEMR's `procedure_order` has no column for intent. Same round-trip behavior as CarePlan.

### Resources deferred with `405 Method Not Allowed` + FHIR OperationOutcome

Seven resources have GET support in OpenEMR but writes were deferred. Each POST/PUT returns a 405 with an OperationOutcome explaining the constraint. Honest reasons:

- **Location** — virtual projection over `patient_data`/`users`/`facility`; no single entity to write to
- **Provenance** — OpenEMR generates Provenance synthetically from other domain resources at read time and has no provenance table to write to. The FHIR spec doesn't forbid client writes, but our implementation has no destination for them.
- **Group** — computed aggregation (patients-by-provider); no persistent storage
- **DiagnosticReport** — federated multi-service (Laboratory + ClinicalNotes); needs per-sub-service routing strategy
- **DocumentReference** — federated three-service; the `$docref` operation remains the primary write path
- **MedicationDispense** — federated; pharmacy persistence varies by deployment
- **Procedure** — federated (clinical + surgery); overlaps with ServiceRequest writes

Two resources are out of scope and not addressed:

- **Observation** — 10-way federation (Vitals, Laboratory, SocialHistory, SDOH, etc.); each sub-service has its own persistence model. Each is roughly a Coverage-sized implementation on its own.
- **Specimen**, **Media** — no underlying OpenEMR domain service exists

Two are spec-read-only by design and intentionally not exposed for writes:

- **OperationDefinition**, **ValueSet** — FHIR foundation/terminology resources

### Verification not yet performed in this PR

Three remaining layers are deliberately out of scope here; flagging them so reviewers know what *isn't* being claimed:

1. **US Core profile conformance** — most resources we touched declare `USCGI_PROFILE_URI` (3.1.0 / 6.1.0 / 7.0.0 / 8.0.0). The base R4 cross-check above does **not** verify Must Support elements added by US Core profiles. For ONC certification, the relevant US Core StructureDefinition for each resource (e.g. [us-core-coverage](http://hl7.org/fhir/us/core/StructureDefinition/us-core-coverage)) should be independently validated.
2. **Inferno test kit validation** — the definitive ONC-aligned conformance check. Should be run against a deployed instance with `feature/fhir-writes` applied; this PR does not include Inferno CI integration.
3. **DB-backed CRUD tests** — every resource ships with a `Fhir{X}ServiceCrudTest` covering insert/update/error paths, but these are integration tests requiring a live DB. They run in the OpenEMR test container; they haven't been executed in this PR's CI yet (we run them manually via `openemr-cmd at` / `composer phpunit-isolated` doesn't cover them since they need MySQL).

### Roadmap to "writes complete"

#### Immediate (this PR)
1. Address Copilot's automated review findings.
2. Address human-reviewer feedback (style, naming, etc.).
3. Run the full test suite in the OpenEMR docker container: `openemr-cmd unit-test`, `api-test`, `services-test`. Fix any regressions in the new CRUD tests.
4. **Update PR title** — currently reads "AllergyIntolerance, Condition, Encounter" but actually covers 11 new resources plus 7 documented 405 endpoints.

#### Near-term (next PR or two)
1. **US Core profile conformance audit per resource.** For each `USCGI_PROFILE_URI` declared, walk the Must Support elements in the StructureDefinition and confirm the write path accepts them (even if silently dropping).
2. **Inferno test kit run** against a staging environment. File issues for each failed test against this branch's behavior. Triage into spec bugs vs. documented narrowings.
3. **Federated resource writes (separate PRs).** Each of the four 405'd federated resources gets its own PR with a per-sub-service routing strategy. Recommended order: Procedure (closest to ServiceRequest, already-built infrastructure) → DiagnosticReport (Laboratory only first, then ClinicalNotes) → DocumentReference (ClinicalNotes only first) → MedicationDispense (LocalDispensary only, since that's the only sub-service implemented).

#### Mid-term (separate workstream)
1. **Observation writes** — 10 sub-services × ~6 hours each = a multi-week effort. Best done one category at a time (Vitals → SocialHistory → Lab → etc.), each as its own PR with US Core profile conformance.
2. **Specimen and Media** — require new OpenEMR domain services to be built first; pharmacy / file-upload concerns interact with this.
3. **Provenance write support (optional)** — if OpenEMR ever adds an audit table that maps to FHIR Provenance, this becomes implementable.

#### Long-term (post-write coverage)
1. **FHIR `Bundle` POST (transaction)** — not implemented in this PR. Clients posting bundles will fail. Bundle support is its own design exercise (transactional semantics across multiple resources, conditional create/update via `If-None-Exist`, etc.).
2. **PATCH operations** — only PUT/POST are implemented. FHIR also defines PATCH (JSON Patch / FHIRPath Patch). Not required for most clients but on the spec.
3. **Conditional updates** (PUT with `If-Match`, POST with `If-None-Exist`) — concurrency / idempotency support for distributed clients.
4. **`_history` / version retrieval** — read-side capability for prior versions of a resource. Pairs with versioned writes (`Meta.versionId` increments).

### What this PR is and isn't

**Is:** a substantial, reviewable foundation for FHIR R4 write support across 11 new resources, with documented constraints and 405 endpoints for resources where writes need separate design work.

**Isn't:** a complete US Core 8.0 certification-ready implementation, an Inferno-passing endpoint, or a Bundle/PATCH-capable server. Those are explicitly downstream.

---

Posting this so the review process has the full picture. Happy to expand any section in further detail if reviewers want — particularly the US Core profile conformance plan or the Inferno integration approach.

kojiromike:
@aisle-analyzer

src/RestControllers/FHIR/FhirGenericRestController.php
```diff
@@ -147,4 +171,230 @@ public function getAll(?array $searchParams = null): JsonResponse|Response
...
+        $fhirResource = $this->deserializeFhirResource($fhirJson);
+        if ($fhirResource instanceof Response) {
+            return $fhirResource;
+        }
+
+        try {
+            $processingResult = $this->getFhirService()->insert($fhirResource);
+        } catch (\InvalidArgumentException $e) {
+            return RestControllerHelper::responseHandler(
+                UtilsService::createOperationOutcomeResource('error', 'invalid', $e->getMessage()),
```
kojiromike: Two concerns on this catch:

1. `$e->getMessage()` lands in `OperationOutcome.diagnostics` (user-facing). Service-layer exceptions can carry SQL fragments or paths — log with an id and return a generic diagnostic.
2. The whole write path lacks audit emission. No `EventAuditLogger::newEvent` here or on the success path in the wrapped services. Chart-side writes use it; FHIR writes mutating PHI should too (HIPAA §164.312(b)).

src/RestControllers/FHIR/FhirGenericRestController.php
```diff
@@ -147,4 +171,230 @@ public function getAll(?array $searchParams = null): JsonResponse|Response
...
+        if ($compartmentCheck instanceof Response) {
+            return $compartmentCheck;
+        }
+
+        $fhirResource = $this->deserializeFhirResource($fhirJson);
+        if ($fhirResource instanceof Response) {
+            return $fhirResource;
+        }
+
+        try {
```
kojiromike, line 280: PUT should reject when `$fhirJson['id']` is present and `!== $fhirId` — FHIR R4 §3.1.0.5.1 requires URL id and body id to match. Otherwise a caller authorized for resource A can `PUT /fhir/X/A` with body `{"id":"B"}` and (depending on which key the service treats as operative) mutate B.

```suggestion
        if (isset($fhirJson['id']) && (string)$fhirJson['id'] !== $fhirId) {
            return RestControllerHelper::responseHandler(
                UtilsService::createOperationOutcomeResource('error', 'invalid', 'id in URL must match id in resource body'),
                null,
                400
            );
        }
        try {
```

src/RestControllers/FHIR/FhirGenericRestController.php
```diff
@@ -147,4 +171,230 @@ public function getAll(?array $searchParams = null): JsonResponse|Response
...
+        return RestControllerHelper::handleFhirProcessingResult($processingResult, 201);
+    }
+
+    /**
+     * Updates an existing FHIR resource
+     * @param string $fhirId The FHIR resource id (uuid)
+     * @param array $fhirJson The updated FHIR resource (complete resource)
+     * @return Response 200 if the resource is updated, 400 if invalid
+     */
+    public function put(string $fhirId, array $fhirJson): Response
```
kojiromike, line 232: No `If-Match` / `ETag` precondition on PUT — concurrent writers silently lose data, and a client retry after a network blip creates a duplicate write. FHIR R4 §3.1.0.5.1 defines `If-Match: W/"<versionId>"`. Reasonable to defer to a follow-up if not in scope, but worth tracking.

src/RestControllers/FHIR/FhirGenericRestController.php
```diff
@@ -147,4 +171,230 @@ public function getAll(?array $searchParams = null): JsonResponse|Response
...
+        $fhirResource = $this->deserializeFhirResource($fhirJson);
+        if ($fhirResource instanceof Response) {
+            return $fhirResource;
+        }
+
+        try {
+            $processingResult = $this->getFhirService()->update($fhirId, $fhirResource);
+        } catch (\InvalidArgumentException $e) {
+            return RestControllerHelper::responseHandler(
+                UtilsService::createOperationOutcomeResource('error', 'invalid', $e->getMessage()),
```
kojiromike: Same two concerns as the `post()` catch: `$e->getMessage()` leaks internal text into `OperationOutcome.diagnostics`, and no audit event is emitted for successful or failed FHIR writes.

src/RestControllers/FHIR/FhirGenericRestController.php
```diff
@@ -147,4 +171,230 @@ public function getAll(?array $searchParams = null): JsonResponse|Response
...
+                $this->getHttpRestRequest(),
+                $aclCheck['section'],
+                $aclCheck['subSection'],
+                $aclCheck['aclPermission']
+            );
+        }
+
+        $fhirValidationService = new FhirValidationService();
+        $validationResult = $fhirValidationService->validate($fhirJson);
+        if (!empty($validationResult)) {
```
kojiromike: Repo `CLAUDE.md` forbids `empty()` in new code (`empty('0') === true`). Use an explicit check — e.g. `if ($validationResult !== [] && $validationResult !== null)`, or tighten `validate()`'s return contract so the boundary is unambiguous. Flagging once as a representative example — same pattern repeats across the new `parseFhirResource` methods.

src/Services/FHIR/FhirAppointmentService.php
```diff
@@ -261,6 +266,275 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+     * Returns the first active facility id, or null if none exist.
+     * Used as a fallback when a FHIR Appointment doesn't include a Location
+     * participant but the underlying service requires pc_facility.
+     *
+     * @return int|null
+     */
+    private function getDefaultFacilityId(): ?int
+    {
+        $facility = QueryUtils::querySingleRow(
+            "SELECT id FROM facility WHERE inactive = 0 ORDER BY id ASC LIMIT 1",
```
kojiromike: `LIMIT 1` with no site filter — in multi-site OpenEMR this can return another tenant's facility as the default. We're not actively developing multi-site, but new code shouldn't introduce a fresh cross-tenant write path. Either scope the query to the current site or refuse to default and require the FHIR caller to supply `serviceProvider`. Same shape in `FhirEncounterService::getDefaultEncounterCategoryId()`.

src/Services/FHIR/FhirEncounterService.php
```diff
@@ -289,6 +291,256 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+            // Only honour the assignment if the caller is admin/users, or the
+            // requested provider IS the caller. Otherwise drop it.
+            if (
+                $providerId !== false
+                && ($canAssignAnyProvider || ($authUserId !== '' && (string) $providerId === $authUserId))
+            ) {
+                $record['provider_id'] = $providerId;
+            }
+        }
+        unset($record['provider_uuid']);
```
kojiromike: When a non-admin caller POSTs an Encounter naming another practitioner as performer, this silently drops `provider_uuid` and returns 201 — the client believes the attribution stuck; downstream billing/audit attribute the visit to whoever `insertEncounter` defaults to. FHIR R4 §2.40 expects either a 4xx with `OperationOutcome`, or a 201 with a `Warning`-severity `OperationOutcome.issue` documenting the override. Silent demotion is worse than failing.

src/Services/FHIR/FhirEncounterService.php
```diff
@@ -289,6 +291,256 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+        // Participant -> provider_uuid and referrer_uuid
+        if (!empty($json['participant'])) {
+            foreach ($json['participant'] as $participant) {
+                $reference = $participant['individual']['reference'] ?? null;
+                if (!is_string($reference) || $reference === '') {
+                    continue;
+                }
+                $parsed = UtilsService::parseReferenceString($reference, 'Practitioner');
+                if (empty($parsed['uuid']) || !\OpenEMR\Common\Uuid\UuidRegistry::isValidStringUUID($parsed['uuid'])) {
+                    continue;
```
kojiromike: `continue` past an unresolvable practitioner reference silently writes a partial Encounter. A POST with `participant: [{<real-doctor>}, {<bogus-uuid>}]` stores the encounter with only the first participant, no warning to the caller. FHIR R4 requires unresolvable references be rejected — throw `\InvalidArgumentException` here and the controller's existing catch turns it into a clean 400.

src/Validators/ImmunizationValidator.php
```diff
@@ -14,12 +16,55 @@
...
+        // insert validations
+        $this->validator->context(
+            self::DATABASE_INSERT_CONTEXT,
+            function (Validator $context): void {
+                $context->required('patient_id')->numeric();
+                $context->required('cvx_code')->lengthBetween(1, 255);
+                $context->required('administered_date')->datetime('Y-m-d');
+                $context->optional('completion_status')->lengthBetween(1, 255);
+                $context->optional('manufacturer')->lengthBetween(1, 255);
+                $context->optional('lot_number')->lengthBetween(1, 255);
```
kojiromike: Safety-critical gaps in this validator:
- `lot_number` is `lengthBetween(1, 255)` with no character class — accepts newlines/control chars/HTML (stored-XSS path if any view renders the column unescaped).
- `cvx_code` accepts any 1–255 char string; should enumerate against the CVX code system.
- `amount_administered` has no validator at all (need a numeric range).
- `expiration_date` is allowed in the past with no warning.

tests/Tests/Services/FHIR/FhirEncounterServiceCrudTest.php
```diff
@@ -0,0 +1,98 @@
...
+
+        $this->fhirEncounterService = new FhirEncounterService();
+        $this->fhirEncounterService->setSystemLogger(new SystemLogger(Level::Critical));
+    }
+
+    protected function tearDown(): void
+    {
+        $this->fixtureManager->removePatientFixtures();
+        // Clean up test encounters
+        QueryUtils::sqlStatementThrowException("DELETE FROM form_encounter WHERE reason LIKE 'test-fixture%'");
```
kojiromike: `LIKE 'test-fixture%'` silently no-ops if the fixture text drifts — orphans accumulate across runs. Also: no `try`/`finally`, so a fatal in `setUp` skips `tearDown` entirely. Capture the inserted uuid in `setUp` and delete by exact uuid in a `finally`. Same pattern across the other CRUD tests.

src/RestControllers/FHIR/FhirImmunizationRestController.php
```diff
@@ -197,4 +197,73 @@ public function getAll($searchParams, $puuidBind = null)
...
+            )
+        ),
+        responses: [
+            new OA\Response(response: '201', description: 'Immunization resource created'),
+            new OA\Response(response: '400', ref: '#/components/responses/badrequest'),
+            new OA\Response(response: '401', ref: '#/components/responses/unauthorized'),
+        ],
+        security: [['openemr_auth' => []]]
+    )]
+    public function post(array $fhirJson): void
```
kojiromike, line 227: Stub `post()` hosting OpenAPI attributes only — the real handler is `FhirGenericRestController::post()`. The `responses` here don't include 403/413/422 the generic controller can actually emit, and there's no CI gate detecting drift. Either generate OpenAPI from the route table, or add a test asserting the per-resource stub's documented `responses` match what the generic controller can return.

tests/Tests/Services/FHIR/FhirEncounterServiceCrudTest.php
```diff
@@ -0,0 +1,98 @@
...
+use OpenEMR\Services\FHIR\FhirEncounterService;
+use OpenEMR\Tests\Fixtures\FixtureManager;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * FHIR Encounter Service CRUD Tests
+ *
+ * @package   OpenEMR
+ * @link      http://www.open-emr.org
```
kojiromike, line 21: Two header issues across all 16 new test files:
- Missing `declare(strict_types=1);` after the opening `<?php` (project convention).
- `@author OpenEMR Contributors` / `@copyright … OpenEMR Contributors` — should be a named author and the contributing entity's copyright. "OpenEMR Contributors" isn't real attribution.

kojiromike (changes requested):
Twelve inline comments. Highest priority below.

**Security / correctness**
- PUT id-mismatch (URL vs body) not enforced — FHIR-spec IDOR vector
- `resolveProviderUuids` silently drops client-supplied attribution; participant loop silently `continue`s past unresolvable references — both write partial records with no warning
- `$e->getMessage()` leaks into `OperationOutcome.diagnostics`
- No audit emission anywhere on the write path (HIPAA §164.312(b)) — chart-side writes use `EventAuditLogger`; FHIR writes should too

**Operational**
- No `If-Match` / `ETag` on PUT — duplicate writes on retry; lost-update on concurrency
- `EncounterService::insertEncounter` (not in this diff) is not transactional; `sqlInsert` + `addForm` partial-fail leaves orphan `form_encounter` rows. Now newly reachable from the public network. Worth handling separately — probably a follow-up
- `LIMIT 1` defaults in `getDefaultFacilityId` / `getDefaultEncounterCategoryId` ignore multi-site scope. Multi-site isn't actively developed but new code shouldn't introduce a fresh cross-tenant write path
- Test cleanup uses `LIKE 'test-fixture%'` with no `try`/`finally` — orphan accumulation if fixture text drifts

**Validator gaps** — `ImmunizationValidator` is missing range/format checks on `lot_number`, `cvx_code`, `amount_administered`, `expiration_date`.

**Architecture** — the per-resource `RestController::post()/put()` stubs exist only to host OpenAPI attributes; they will drift from the generic controller's actual behavior. Worth either deriving OpenAPI from the route table or adding a CI gate.

**PR body discrepancy** — the body says "5 resources" but the diff wires 16 + 7 documented deferral stubs. Please reconcile before merge.

Not blocking but flagged once: `empty()` use in new code (one representative inline; same pattern across new `parseFhirResource` methods) and test-file headers (`declare(strict_types=1)` + named-author attribution).

rollingventures:
The 3 PR-level Aisle findings posted on 2026-05-28 are addressed in commit `ce9cb890a`:

- **🟡 Internal exception messages leaked into FHIR error responses** (CWE-209) — `RestControllerHelper::handleFhirProcessingResult()` now logs the raw `internalErrors` server-side with a correlation id and returns a generic message (`"An internal error occurred (incident <id>)"`) to the client.
- **🟠 User-controlled class instantiation in `FhirValidationService::validate()`** (CWE-470) — second copy of the same gadget I'd fixed in `FhirGenericRestController`. `resourceType` now must match `^[A-Z][A-Za-z0-9]*$` before any class resolution, and the resolved class must be a `FHIRDomainResource` subclass before instantiation.
- **🟠 Stored XSS via `Appointment.comment` → `pc_hometext`** (CWE-79) — two-prong fix: `printed_fee_sheet.php` now wraps the value in `text()`; `FhirAppointmentService::parseFhirResource()` calls `strip_tags()` on `comment` at the write boundary as defense in depth.

CI green on tip `6d0dc21bb`.

src/RestControllers/FHIR/FhirGenericRestController.php
kojiromike: Deferred to #12378 (If-Match/ETag preconditions on FHIR PUT). Out of scope for this PR; tracking separately.

src/RestControllers/FHIR/FhirImmunizationRestController.php
kojiromike: Deferred to #12379 (OpenAPI response-drift gate). Out of scope for this PR; tracking separately.

kojiromike (approved):
Namespace nit (inline `\Symfony\...\Response` → `use` import) is addressed in the current code. The two items I'd flagged are deferred and tracked: If-Match/ETag preconditions in #12378, OpenAPI response-drift gate in #12379 — both out of scope here. CI is green. LGTM.

src/Services/FHIR/FhirAllergyIntoleranceService.php
```diff
@@ -243,6 +243,159 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+        // Code -> title and diagnosis
+        if (!empty($json['code']['coding'])) {
+            $diagnosisParts = [];
+            foreach ($json['code']['coding'] as $coding) {
+                $system = $coding['system'] ?? '';
+                $codeValue = $coding['code'] ?? '';
+                $display = $coding['display'] ?? '';
+                if (!empty($codeValue)) {
+                    $prefix = match ($system) {
+                        'http://snomed.info/sct' => 'SNOMED-CT',
```
adunsulag: CodeTypesService::getCodeTypeForSystemUrl()  should be used here for the matching of the system url to OpenEMR's codetypes.  If that method for some reason isn't fit for purpose you should leverage the constants in FhirCodeSystemConstants.

src/Services/FHIR/FhirAllergyIntoleranceService.php
```diff
@@ -243,6 +243,159 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+        }
+
+        // Reaction -> reaction code
+        if (!empty($json['reaction'][0]['manifestation'][0]['coding'])) {
+            $reactionParts = [];
+            foreach ($json['reaction'][0]['manifestation'][0]['coding'] as $coding) {
+                $codeValue = $coding['code'] ?? '';
+                if (!empty($codeValue)) {
+                    $system = $coding['system'] ?? '';
+                    $prefix = match ($system) {
```
adunsulag: See comment on line 256 for same problem here.

src/Services/FHIR/FhirAppointmentService.php
```diff
@@ -261,6 +266,256 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+            if ($catId !== false) {
+                $data['pc_catid'] = $catId;
+            }
+        }
+
+        // Default pc_title from appointmentType display or "Office Visit"
+        if (!empty($json['appointmentType']['coding'][0]['display'])) {
+            $data['pc_title'] = $json['appointmentType']['coding'][0]['display'];
+        } else {
+            $data['pc_title'] = 'Office Visit';
```
adunsulag: You'll want to use xl_appt_category for translation here if you're goig to default the category title.

src/Services/FHIR/FhirAppointmentService.php
```diff
@@ -261,6 +266,256 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+                    // OR is assigning the appointment to themselves.
+                    if (
+                        $providerId !== false
+                        && ($canAssignAnyProvider
+                            || ($authUserId !== '' && (string) $providerId === $authUserId))
+                    ) {
+                        $data['pc_aid'] = $providerId;
+                    }
+                } elseif ($parsed['type'] === 'Location') {
+                    $facilityUuidBytes = UuidRegistry::uuidToBytes($parsed['uuid']);
```
adunsulag, line 386: Probably need to validate this facility location against facility service locations, otherwise people could schedule appointments against billing locations or even internal facilities that an organization has made that they don't want to have scheduled against.

src/Services/FHIR/FhirAppointmentService.php
```diff
@@ -261,6 +266,256 @@ public function parseOpenEMRRecord($dataRecord = [], $encode = false)
...
+            ]);
+            return $processingResult;
+        }
+        if (empty($openEmrRecord['pc_billing_location']) && !empty($openEmrRecord['pc_facility'])) {
+            $openEmrRecord['pc_billing_location'] = $openEmrRecord['pc_facility'];
+        }
+
+        // Default pc_catid if not provided (required by validator)
+        if (empty($openEmrRecord['pc_catid'])) {
+            $openEmrRecord['pc_catid'] = 9; // Office Visit is typically 9
```
adunsulag: I wonder if this should be extracted into a global setting, or extract the magic number into a constant.

src/Services/FHIR/FhirCarePlanService.php
```diff
@@ -701,6 +704,269 @@ protected function searchForOpenEMRRecords($openEMRSearchParameters): Processing
...
+        return $dt === false ? null : $dt->format('Y-m-d');
+    }
+
+    /**
+     * OpenEMR's form_care_plan.code column stores codes prefixed by code-type
+     * (e.g. "SNOMED-CT:182840001"). The read side splits this via CodeTypesService.
+     */
+    private function prefixCodeForStorage(string $system, string $code): string
+    {
+        $prefix = match ($system) {
```
adunsulag: Another area where the CodeTypesService should be used.

src/Services/FHIR/FhirConditionService.php
```diff
@@ -119,6 +121,138 @@ public function getAll($fhirSearchParameters, $puuidBind = null): ProcessingResu
...
+        // Code -> title and diagnosis
+        if (!empty($json['code']['coding'])) {
+            $diagnosisParts = [];
+            foreach ($json['code']['coding'] as $coding) {
+                $system = $coding['system'] ?? '';
+                $codeValue = $coding['code'] ?? '';
+                $display = $coding['display'] ?? '';
+                if (!empty($codeValue)) {
+                    $prefix = match ($system) {
+                        'http://snomed.info/sct' => 'SNOMED-CT',
```
adunsulag: I suggest searching everywhere where SNOMED-CT (or have claude do it) and have it utilize the CodeTypesService class.

src/Services/FHIR/FhirConditionService.php
```diff
@@ -119,6 +121,138 @@ public function getAll($fhirSearchParameters, $puuidBind = null): ProcessingResu
...
+            }
+        }
+
+        // VerificationStatus -> verification
+        if (!empty($json['verificationStatus']['coding'][0]['code'])) {
+            $data['verification'] = $json['verificationStatus']['coding'][0]['code'];
+        }
+
+        // OnsetDateTime -> begdate (ConditionValidator expects Y-m-d)
+        if (!empty($json['onsetDateTime'])) {
```
adunsulag, line 204: Note I think you're going to run into problems anywhere you are accepting datetime values.  The specification for the DateTime value allows a pretty wide range of values https://build.fhir.org/datatypes.html#dateTime.

For example YYYY, YYYY-MM, YYYY-MM-DD or YYYY-MM-DDThh:mm:ss+zz:zz are all valid allowed values. 

You may need to pull the code for the DateTime parser in DateSearchField::createDateComparableValue.

Another option would be to just reject the value or throw a not supported operation if the DateTime is not a full ISO8601 value.  

On read requests we didn't have to worry about it as we controlled the format, but on input we have a wider range of inputs for datetime.

src/Services/FHIR/FhirCoverageService.php
```diff
@@ -2,7 +2,9 @@
 
 namespace OpenEMR\Services\FHIR;
 
+use OpenEMR\Common\Database\QueryUtils;
```
adunsulag, line 5: Putting a marker here for me to continue my review later.

adunsulag (changes requested):
Hey @rollingventures I've been going through the PR, have to stop for today but will continue more tomorrow. Note a few issues on DateTime parsing as well as on code system mapping.  There's one minor information disclosure issue on the appointment service that will need to be addressed as well.

src/Services/FHIR/FhirAllergyIntoleranceService.php
kojiromike: Done — replaced the inline `match($system)` with `CodeTypesService::getOpenEMRCodeForSystemAndCode()`, which is the inverse of the read-side `BaseService::addCoding()` → `getSystemForCodeType()` path, so read and write share one source of truth for supported systems. This also fixed a latent bug here: the old map emitted an `RXNORM:` prefix, but the seeded `code_types.ct_key` is `RXCUI`, so RxNorm-coded allergies were being stored with an unknown code type.

src/Services/FHIR/FhirAllergyIntoleranceService.php
kojiromike: Done — the reaction-manifestation coding now goes through `CodeTypesService::getOpenEMRCodeForSystemAndCode()` as well.

src/Services/FHIR/FhirAppointmentService.php
kojiromike: Done — the fallback title now goes through `xl_appt_category()` (gated on `translate_appt_categories`), and the literal moved to a `DEFAULT_CATEGORY_TITLE` constant.

src/Services/FHIR/FhirAppointmentService.php
kojiromike: Extracted both magic numbers into named constants — `DEFAULT_CATEGORY_ID = 9` and `DEFAULT_DURATION_SECONDS = 900`, each with a docblock explaining the value. I kept them as constants rather than global settings for now; if you'd prefer these be admin-configurable globals, that's a reasonable follow-up — happy to file it.

src/Services/FHIR/FhirCarePlanService.php
kojiromike: Done — `prefixCodeForStorage()` here now delegates to `CodeTypesService::getOpenEMRCodeForSystemAndCode()`.

src/Services/FHIR/FhirConditionService.php
kojiromike: Searched the write side for the inline system→prefix maps and converted all of them to `CodeTypesService::getOpenEMRCodeForSystemAndCode()`: Condition (here), AllergyIntolerance (code + reaction), CarePlan, Goal, and ServiceRequest. To keep that swap from dropping ICD codes, I taught `CodeTypesService` the US Core `icd-10-cm` / `icd-9-cm` system URLs (the existing map only knew the base `icd-10`), backed by new `FhirCodeSystemConstants::HL7_ICD10_CM` / `HL7_ICD9_CM` constants. Added Condition CRUD coverage asserting `icd-10-cm`/`icd-9-cm` round-trip to the `ICD10:`/`ICD9:` prefixes.

src/Services/FHIR/FhirAppointmentService.php
kojiromike: Agreed — this needs validating the Location against the schedulable service-location set, not just any resolvable `facility.id`. It's enough of a design decision (which facility flags define "schedulable", and the matching error response) that I've filed it as a follow-up rather than expanding this PR: #12430.

src/Services/FHIR/FhirConditionService.php
kojiromike: Good catch on the wider input surface. For now I've made the Condition write path reject partial-precision values rather than silently truncating them: `parseFullDateTimeToDate()` requires at least a full `YYYY-MM-DD` (full date-times are accepted too) and drops the field for `YYYY` / `YYYY-MM`, so a year-only onset no longer lands as a wrong calendar day. Full partial-precision handling (likely reusing `DateSearchField::createDateComparableValue`, and deciding how each precision maps to OpenEMR's date columns) is tracked in #12429.

sjpadgett:
@rollingventures 
Hi Chris, this PR solves a important piece of my current Questionnaire project here issue #13266
Completion here would save me considerable work speeding up my Questionnaire workflow project.

If you've become too busy to finish would you be opposed if I was to jump in and help finish the review and any items needed to complete the project resulting in a merge?

You can give me a call or PM me if you'd like to discuss at length.

rollingventures:
please take any working PR's and do whatever you would like. i have so much work that it looks like i am coming out of retirement.

have at it!
cdx

sjpadgett:
> i have so much work that it looks like i am coming out of retirement.

lol Chris! Yep I get the coming out of retirement part! Still, it keeps our brain churning.

Thanks for your hard work on this and I hope I can bring it home while ensuring you receive the credit you deserve.
Be safe and good luck on whatever you're up to.
-- J

sjpadgett:
Continuing this work in a new PR

Thanks @rollingventures — picking this up per your note above. Maintainers have asked that it continue as a separate PR rather than as pushes into this branch, so the work moves to #13281 ("continues #11507").

To be explicit about credit: the new branch is cut from 0043797 with your commits carried over intact, no squash, so your authorship stands on every line you wrote. Anything that has to be squashed later carries Co-authored-by: rollingventures. This PR is where the design conversation happened, and I'll keep linking back to it rather than restating your reasoning as my own.

@bradymiller / @adunsulag — once the successor PR is up, could one of you close this one with a pointer to it? I'd rather a maintainer do that than close someone else's PR myself.

What carries over

Everything. The successor branch is this branch plus a master merge and the fixes noted below — no resource support is dropped. The review record here stays the reference for:

the FHIR R4 spec cross-check and the four documented R4 narrowings (Coverage.payor, MedicationRequest.medication[x], CarePlan.intent, ServiceRequest.intent)
the rationale for the seven 405-deferred resources and the two out-of-scope ones
the ~30 threads whose fixes are already in 0043797 — I've carried those over as resolved in the successor PR's checklist so nobody re-reviews April

Two items were deferred here and stay deferred, tracked separately: If-Match/ETag preconditions → #12378, OpenAPI response-drift gate → #12379.

Two fixes going in on top of your work

Flagging these here since they touch decisions made in this PR:

The seven 405-deferred resources register real POST/PUT routes, and RestControllerHelper::addRequestMethods() derives the CapabilityStatement from the route map — so /fhir/metadata would advertise create/update for all seven. Fixed with a factory that registers the 405 handler and excludes it from the capability statement, keeping the route file as the single source of truth.
@adunsulag's 6/08 point about the FHIR dateTime surface applies to all twelve write services, not just the Condition path that got patched. date_create_immutable('2024') — legal FHIR year precision — parses as a clock time and stores today; '2024-02-30' rolls forward instead of erroring; offset-bearing values are stored in the sender's wall clock. Replaced with one shared parser that throws into the existing 400 path.
Why it isn't landing as one PR

No criticism of the work — 124 files simply hasn't gotten a completed review in two attempts, and review here stopped at FhirCoverageService.php:5 on 6/08. The value in the branch is the write plumbing, which is reviewable on its own, so the successor PR is scoped to that and the rest follows in three stacked PRs. Sequence and rationale are in the successor PR description.

For context on why I'm pushing: #13266 (questionnaire-driven clinical workflows) can't start until the generic FHIR write path exists, and I'd rather finish this work than build a parallel one.

@adunsulag — the remaining review would pick up on the successor PR rather than here. The scoped first PR is ~15 files, which should be a much smaller ask than resuming from the Coverage marker in a 124-file diff.

bradymiller:
Closing this. Continued at https://github.com/openemr/openemr/pull/13281

PR #11511, opened 2026-04-09
fix(install): use newline separator in SQL statement concatenation

### LINKED ISSUE #11465: fix(install): indented SQL comments silently swallow following statements in Installer::load_file()
## Bug

`Installer::load_file()` skips comment lines with:

```php
$line = rtrim($rawLine);
if ($line === "" || str_starts_with($line, "--") || str_starts_with($line, "#")) {
    continue;
}
```

The `str_starts_with` checks run against the `rtrim()`'d line, so they only catch comments that begin in column zero. A line like `    -- Table for storing ...` has leading whitespace and is **not** skipped — it gets appended to the current statement buffer.

Combined with the fact that `load_file()` strips the trailing newline and concatenates lines with a space separator, the comment then poisons the statement sent to MySQL. MySQL's `--` comment runs to end-of-line, and in the concatenated buffer there is no end-of-line, so `-- comment STATEMENT;` becomes one giant comment and the statement is silently dropped.

## Example

`sql/database.sql:15214–15216` (introduced in #9275, November 2025):

```sql
    -- ------------------------------------- Parent lists under `lists`--------------------------------------------------------------------
    INSERT INTO `list_options` (`list_id`,`option_id`,`title`,`seq`)
    VALUES  ('lists','treatment_intervention_preferences','Treatment Intervention Preferences',1);
```

After `load_file()` processes those three lines, the buffer sent to MySQL is:

```
    -- ------- Parent lists under `lists`------    INSERT INTO `list_options` (`list_id`,`option_id`,`title`,`seq`)    VALUES  ('lists','treatment_intervention_preferences','Treatment Intervention Preferences',1);
```

MySQL sees a leading `--`, treats the whole buffer as a comment, and the `INSERT` is never executed. The `treatment_intervention_preferences` row under `lists` is silently missing from a fresh install via this code path.

`sql/database.sql:15190` has the same pattern, though its target (`DROP TABLE IF EXISTS patient_care_experience_preferences`) is idempotent on a fresh install so the practical impact is smaller.

## Scope of impact

- Any multi-line SQL dump loaded via `Installer::load_file()` where an indented `--` or `#` comment immediately precedes a statement whose first character is also indented. On fresh-install path, any `INSERT` that lands after such a comment is silently skipped.
- The bug only surfaces for fresh installs that run the installer against `sql/database.sql`. Docker images, release tarballs, and environments that seed the database via migrations or a pre-built snapshot may be unaffected.
- #11003 (the prior fix for #10935) did not introduce this — the bug predates it. The comment check was written for column-zero comments and has silently lost data for any indented comment ever since.

## Fix options

1. **Use `\n` as the line separator instead of a space.** MySQL then terminates `--` comments naturally, and the comment-skipping loop becomes optional (MySQL parses comments on its own). Smallest diff, and it means the hand-rolled parser no longer has to understand SQL comment syntax.
2. **`trim()` before the comment check**, so indented comments are also skipped. Simple, but preserves the hand-rolled comment-parser approach and doesn't address the underlying "hand-rolled parser re-implementing SQL grammar" problem.
3. **Delete the hand-rolled parser entirely** and use `mysqli_multi_query()`. Tracked in #11458. Subsumes this issue.

Recommendation: fix via option 1 as a narrow follow-up to #11461, unless #11458 is about to land, in which case punt.

## How this was found

Raised during self-review of #11461 ("is this comment detection worth it? What if a line starts with space, and then has a comment character?"). The comment detection as written was worth it for column-zero comments and *actively harmful* for indented ones — exactly the scenario called out.

## Related

- #10935 / #11003 — concatenation-without-separator bug in the same function.
- #11459 / #11461 — drops the 1024-byte fgets() buffer in the same function.
- #11458 — replace `load_file()` with `mysqli_multi_query()` entirely.

haoyu-haoyu:
Change `$query .= " "` to `$query .= "\n"` in Installer.class.php so that `--` SQL comments don't swallow the following statement. With space separators, MySQL's end-of-line comment syntax runs to the end of the concatenated string.

Fixes #11465

### DIFF
--- a/library/classes/Installer.class.php
+++ b/library/classes/Installer.class.php
@@ -481,14 +481,12 @@ class Installer
                 continue;
             }
 
-            // Insert a single space separator between concatenated lines.
-            // Without it, a continuation line starting at column 0 would
-            // fuse with the previous token (e.g. "= 0" + "WHERE" → "0WHERE",
-            // see #10935). getLine() is unbounded, so every call returns a
-            // complete logical line and the separator is always safe to
-            // insert.
+            // Insert a newline separator between concatenated lines.
+            // A space would fuse a SQL "--" comment with the following line,
+            // swallowing it (see #11465). A newline terminates the comment
+            // at the line boundary as SQL requires.
             if ($query !== "") {
-                $query .= " ";
+                $query .= "\n";
             }
             $query .= $line;
             $chr = substr($query, strlen($query) - 1, 1);
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -12691,7 +12691,8 @@ INSERT INTO ccda_sections (ccda_sections_id, ccda_components_id, ccda_sections_f
 INSERT INTO ccda_sections (ccda_sections_id, ccda_components_id, ccda_sections_field, ccda_sections_name, ccda_sections_req_mapping) values('45','9','unstructured_doc','Document','0');
 INSERT INTO `ccda_sections` (`ccda_sections_id`, `ccda_components_id`, `ccda_sections_field`, `ccda_sections_name`, `ccda_sections_req_mapping`) VALUES
 ('46', '3', 'medical_devices', 'Medical Devices', '0'),
-('47', '3', 'goals', 'Goals', '0');-- --------------------------------------------------------
+('47', '3', 'goals', 'Goals', '0');
+-- --------------------------------------------------------
 
 --
 -- Table structure for table `ccda_field_mapping`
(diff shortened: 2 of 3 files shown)


kojiromike:
Copilot is right – the bug case should be tested directly, but also the tests are failing because they assume the concatenation string is a " ".

haoyu-haoyu:
@kojiromike Test updated: renamed to `testLoadFileInsertsNewlineSeparatorBetweenSqlLines` and assertions changed from `'0 WHERE'` to `"0\nWHERE"` to match the new newline separator.

kojiromike:
Please address the failing tests.

PR #11513, opened 2026-04-09
fix(security): validate menu role filenames to prevent path traversal

### LINKED ISSUE #11489: MenuRole: validate user-supplied menu role filenames before file_get_contents
`MainMenuRole::getMenu()` and `PatientMenuRole::getMenu()` read `$user['main_menu_role']` / `$user['patient_menu_role']` from the users table, and when the value ends with `.json`, concatenate it directly into a filesystem path passed to `file_get_contents()`:

```php
if (str_ends_with($mainMenuRole, '.json')) {
    $menu_parsed = json_decode(file_get_contents($OE_SITE_DIR . "/documents/custom_menus/" . $mainMenuRole));
}
```

There is no validation/normalization of the value before it reaches the filesystem. The field is admin-controlled (set via `interface/usergroup/`), so this is hardening rather than a remotely exploitable issue, but the pre-existing code accepts values like `../somewhere/other.json` and would attempt to read arbitrary `.json` files relative to the custom menus directory.

Suggested fix:

- Reject anything containing `..`, `/`, or `\` (i.e., enforce basename-only).
- Or whitelist against the actual files discovered by `displayMenuRoleSelector()` (`scandir()` of the custom menus directory, filter to `.json`, and require an exact match).

Same fix needed in both `MainMenuRole::getMenu()` and `PatientMenuRole::getMenu()`.

Noticed by copilot review of #11456. Pre-existing code, not introduced by that PR.

haoyu-haoyu:
Reject menu role filenames containing `/`, `\`, or `..` before passing to `file_get_contents()` in MainMenuRole and PatientMenuRole. Prevents admin-supplied filenames from reading arbitrary JSON files outside the custom_menus directory.

Fixes #11489

### DIFF
--- a/src/Menu/MainMenuRole.php
+++ b/src/Menu/MainMenuRole.php
@@ -12,7 +12,7 @@
 
 namespace OpenEMR\Menu;
 
-use OpenEMR\Common\Logging\SystemLogger;
+use OpenEMR\BC\ServiceContainer;
 use OpenEMR\Core\OEGlobalsBag;
 use OpenEMR\Services\UserService;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
@@ -51,7 +51,7 @@ class MainMenuRole extends MenuRole
 
         // Validate that the menu role filename is a basename only (no path traversal)
         if ($mainMenuRole !== basename($mainMenuRole) || str_contains($mainMenuRole, '..')) {
-            (new SystemLogger())->error("Invalid menu role filename rejected", ['filename' => $mainMenuRole]);
+            ServiceContainer::getLogger()->error("Invalid menu role filename rejected", ['filename' => $mainMenuRole]);
             die("\nInvalid menu role filename.");
         }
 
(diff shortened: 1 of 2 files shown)


kojiromike:
please address phpstan findings.

haoyu-haoyu:
@kojiromike Replaced `error_log()` with `(new SystemLogger())->error()` in both MainMenuRole and PatientMenuRole. PHPStan custom rule `openemr.forbiddenErrorLog` should be satisfied now.

kojiromike:
@haoyu-haoyu phpstan is still failing, because we use a service locator, not `new`. Please see the error message.

kojiromike (approved):
LGTM as a stopgap — this is strictly better than the current state on `main` and blocks the obvious traversal vectors.

Dismissing the Copilot suggestions:
- **Fall back to default instead of `die()`:** Disagree. This is security validation — silently falling back to `"standard"` would mask a potential compromise. Failing loudly is the right call here.
- **Explicit backslash check:** Technically valid but academic in practice — on Linux `\` isn't a path separator, so `file_get_contents()` would just fail to find the file. The `basename()` check is sufficient for the actual threat.

Filed a follow-up issue for the deeper architectural fix (allowlist at write boundary, `realpath()` at read boundary, consolidate duplicated logic).

PR #11514, opened 2026-04-09
chore: drop unnecessary fgets() length limits

### LINKED ISSUE #11464: chore: drop cargo-culted fgets() length limits in miscellaneous call sites
## Background

Janitorial follow-up to the `fgets()` audit performed during #11461. Three remaining call sites pass a fixed length argument where none is needed. None of them appears to actively corrupt data today, but all three are fragile, unnecessary, and hold the same shape as the cargo-culted pattern fixed elsewhere.

The `length` argument to `fgets()` has been optional since PHP 4.3 (December 2002). When omitted, `fgets()` reads until newline or EOF, returning a complete logical line every call. That is what all three call sites want.

### 1. `library/smarty_legacy/smarty/plugins/function.fetch.php`

Lines 43, 178, 197 — `fgets($fp, 4096)` in three places.

- Line 43 reads from a local file opened for display; length truncation would silently corrupt the output on long lines.
- Lines 178 and 197 read HTTP headers and body from a socket. A `Set-Cookie` or `Received:` header longer than 4095 bytes would be split mid-header and concatenated; the downstream `preg_split("!\r\n\r\n!", $content, 2)` would then find the wrong header/body boundary.

This is a legacy Smarty plugin. Fix is three one-line edits.

### 2. `modules/sms_email_reminder/sms_clickatell.php:354`

```php
while ($line = @fgets($handler, 1024))
```

Reads HTTP responses from the Clickatell SMS gateway. Concatenated into `$result`, then downstream `substr($result, 4)` / `substr($result, 0, 2)` parse positions. Works by accident for short responses, fragile for error payloads.

### 3. `controllers/C_Prescription.class.php:749`

```php
while ($line = fgets($f, 1000))
```

Reads addendum text for PDF rendering. Worst case on a long line is a mid-word break in the rendered PDF, not data corruption. Still unnecessary.

## Fix

Drop the length argument from each call site. Total change is ~5 one-line edits across three files. No behavior change for any input shorter than the original limits.

## Out of scope

- The phiMail protocol parsers (`library/direct_message_check.inc.php`, `ccr/transmitCCD.php`) — tracked separately in #11463 because they are a distinct subsystem and have a higher likelihood of real-world impact.
- `src/Services/Utils/SQLUpgradeService.php` — tracked separately in #11462 because it is the same bug class as #10935 and deserves its own PR.
- `library/classes/smtp/smtp.php:107` — uses `fgets(conn, 100)` inside a chunk-reassembly loop that correctly handles splits. Not a bug. Worth cleaning up eventually but not in this pass.

## Related

- Found by the bycatch audit during #11461.

haoyu-haoyu:
Remove the second argument from 5 `fgets()` calls across 3 files. The length parameter has been optional since PHP 4.3 and these fixed limits (1000-4096 bytes) silently truncate long lines.

Fixes #11464

### DIFF
--- a/library/smarty_legacy/smarty/plugins/function.fetch.php
+++ b/library/smarty_legacy/smarty/plugins/function.fetch.php
@@ -39,8 +39,8 @@ function smarty_function_fetch($params, &$smarty)
 
         // fetch the file
         if($fp = @fopen($params['file'],'r')) {
-            while(!feof($fp)) {
-                $content .= fgets($fp);
+            while(($line = fgets($fp)) !== false) {
+                $content .= $line;
             }
             fclose($fp);
         } else {
@@ -174,8 +174,8 @@ function smarty_function_fetch($params, &$smarty)
                     }
 
                     fwrite($fp, "\r\n");
-                    while(!feof($fp)) {
-                        $content .= fgets($fp);
+                    while(($line = fgets($fp)) !== false) {
+                        $content .= $line;
                     }
                     fclose($fp);
                     $csplit = preg_split("!\r\n\r\n!",$content,2);
@@ -193,8 +193,8 @@ function smarty_function_fetch($params, &$smarty)
         } else {
             // ftp fetch
             if($fp = @fopen($params['file'],'r')) {
-                while(!feof($fp)) {
-                    $content .= fgets($fp);
+                while(($line = fgets($fp)) !== false) {
+                    $content .= $line;
                 }
                 fclose($fp);
             } else {
--- a/library/smarty_legacy/smarty/plugins/function.fetch.php
+++ b/library/smarty_legacy/smarty/plugins/function.fetch.php
@@ -40,7 +40,7 @@ function smarty_function_fetch($params, &$smarty)
         // fetch the file
         if($fp = @fopen($params['file'],'r')) {
             while(!feof($fp)) {
-                $content .= fgets ($fp,4096);
+                $content .= fgets($fp);
             }
             fclose($fp);
         } else {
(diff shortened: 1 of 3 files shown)


haoyu-haoyu:
Thanks for the thorough review, @Copilot! All 4 suggestions addressed:

1-3. **`function.fetch.php`**: Replaced all 3 `while (!feof())` loops with `while (($line = fgets($fp)) !== false)` to prevent infinite spin on read errors.

4. **`C_Prescription.class.php`**: Added `fopen()` return check + `try/finally/fclose()` for proper resource cleanup.

Also fixed `sms_clickatell.php` to use strict `!== false` check instead of loose truthiness (avoids silently dropping a line equal to `"0"`).

kojiromike:
Please address Phpstan failure

haoyu-haoyu:
Fixed in 0d32398 — removed the stale `fgets expects resource, resource|false` baseline entry for `C_Prescription.class.php`. The `fopen()` return check from the previous commit narrows the type to `resource` before `fgets` is called, so PHPStan no longer reports that error.

library/smarty_legacy/smarty/plugins/function.fetch.php
```diff
@@ -39,8 +39,8 @@ function smarty_function_fetch($params, &$smarty)
 
         // fetch the file
         if($fp = @fopen($params['file'],'r')) {
-            while(!feof($fp)) {
-                $content .= fgets ($fp,4096);
+            while(($line = fgets($fp)) !== false) {
+                $content .= $line;
             }
             fclose($fp);
```
kojiromike, line 45: You addressed the try/finally issue in one place, but it occurs in several other places in this PR, too. I think it should be addressed, but it isn't the same as the fgets issue this PR purports to solve. So maybe open a second PR for that, please? Or at least create an issue for follow-up.

haoyu-haoyu:
Good call — reverted the try/finally change in 0a28cb0 to keep this PR focused on fgets length limits only. Opened #11542 to track the resource cleanup separately.

kojiromike (approved):

PR #11515, opened 2026-04-09
fix(cdr): stop translating DB identifiers in render_select

### LINKED ISSUE #11469: cdr: render_select translates option labels — DB identifier lists shouldn't be translated
`RuleTemplateExtension::render_select()` calls `xlt($option['label'])` on every option label, regardless of what the option list represents. This is wrong for `RuleCriteriaDatabaseCustom::getTableNameOptions()`, which returns raw database table names (e.g. `patient_data`, `form_encounter`) — those identifiers should be displayed verbatim, not pushed through the translation layer.

It's also a `literal-string` type violation: `xlt()` expects `literal-string` but `$option['label']` is a runtime value, so the call sits in the PHPStan baseline.

The cleanest fix is to:
1. Stop translating in `render_select` — change `xlt($option['label'])` to `text($option['label'])`.
2. Have option providers that DO need translation pre-translate at the source. The two callers that need this today are `RuleCriteriaSex::getOptions()` and `RuleCriteriaLifestyle::getOptions()`.

This is a follow-up to PR #11444 which removed the (separate) `xl()` wrapping in `getTableNameOptions()`. The render-time translation in `render_select` was flagged in review but is out of scope for the literal-string cleanup PR.

haoyu-haoyu:
Change `xlt($option['label'])` to `text($option['label'])` in RuleTemplateExtension.php. Database table names passed by `getTableNameOptions()` should be displayed verbatim, not translated. Callers that need translation already pre-translate their labels.

Fixes #11469

### DIFF
--- a/src/ClinicalDecisionRules/Interface/RuleTemplateExtension.php
+++ b/src/ClinicalDecisionRules/Interface/RuleTemplateExtension.php
@@ -21,7 +21,7 @@ class RuleTemplateExtension
                 <option id="<?php echo attr($option['id']); ?>"
                         value="<?php echo attr($option['id']); ?>"
                     <?php echo $args['value'] == $option['id'] ? "SELECTED" : "" ?>>
-                    <?php echo text($option['label']); ?>
+                    <?php echo text((string)($option['label'] ?? '')); ?>
                 </option>
             <?php } ?>
 
--- a/src/ClinicalDecisionRules/Interface/RuleTemplateExtension.php
+++ b/src/ClinicalDecisionRules/Interface/RuleTemplateExtension.php
@@ -21,7 +21,7 @@ class RuleTemplateExtension
                 <option id="<?php echo attr($option['id']); ?>"
                         value="<?php echo attr($option['id']); ?>"
                     <?php echo $args['value'] == $option['id'] ? "SELECTED" : "" ?>>
-                    <?php echo xlt($option['label']); ?>
+                    <?php echo text($option['label']); ?>
                 </option>
             <?php } ?>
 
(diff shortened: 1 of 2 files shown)


kojiromike:
Please fix the phpstan baseline. Better still, fix the phpstan issue.

haoyu-haoyu:
@kojiromike Fixed the PHPStan issue at the source: cast to string with `text((string)($option['label'] ?? ''))`. Updated baseline counts: xlt count 2→1 (removed call), cast.string count 1→2 (added cast).

src/ClinicalDecisionRules/Interface/RuleTemplateExtension.php
```diff
@@ -21,7 +21,7 @@ public static function render_select($args)
                 <option id="<?php echo attr($option['id']); ?>"
                         value="<?php echo attr($option['id']); ?>"
                     <?php echo $args['value'] == $option['id'] ? "SELECTED" : "" ?>>
-                    <?php echo text($option['label']); ?>
+                    <?php echo text((string)($option['label'] ?? '')); ?>
```
kojiromike: If you can't fix the typing at the source (and that's admittedly quite hard in openemr right now), the next best thing is the rule "narrow, don't cast", meaning, we actually already know that `$option['label']` is probably not truly `mixed`, but is probably either `string` or `?string`, so something like…

```
<?php echo text(is_string($option['label'] ?? null) ? $option['label'] : '') ?>
```

haoyu-haoyu:
Thanks @kojiromike — good point on "narrow, don't cast". Updated in 6aab579:

```php
<?php echo text(is_string($option['label'] ?? null) ? $option['label'] : ''); ?>
```

Also reverted the `cast.string` baseline count back to 1 since no `(string)` cast remains.

PR #11519, opened 2026-04-09
feat(storage): Flysystem setup; first example integration

### LINKED ISSUE #11505: chore(storage): Add support for `flysystem` and encourage its use
[`flysystem`](https://flysystem.thephpleague.com/docs/) is a filesystem abstraction layer for PHP. I believe that adopting it will solve a number of pervasive problems and challenges in OpenEMR:

- It has a "fail loudly" design. The core APIs either succeed or throw an exception. In contrast, many of the low-level file utilities in PHP require checking for sentinel types in return values, several of which invite bugs (e.g. some things return `int|false`, and returning `0` isn't necessarily an error; `==` comparisons are broken in these cases)
- It supports a large number of different storage systems, both local and remote. Especially for tasks like document storage, this opens the possibility of storing data in S3 (and its clones) with only a couple lines of config change. This can _vastly_ simplify operating OpenEMR at scale.
- The storage engines are extensible; if someone running OpenEMR needs something not supported out of the box (fairly unlikely given what's supported), they'd just need to implement a couple interfaces and change config, but not hundreds of touch points
- Central config a) eliminates probably hundreds of string concatenation calls around `OE_SITE_DIR` (etc), and b) paves the road to getting data outside of the web root.
- Code relying on an abstraction rather than direct FS calls becomes much more testable

Aside from a cleaner and more extensible API, we'll probably eliminate a whole bunch of security risks.

At a brief glance, there's ~500 total `fopen`/`file_get_contents`/`file_put_contents` calls, so this is no small task and we SHOULD NOT aim to do it all at once. There will be some gaps and some false-positives in that count, but it's a general sense of the scope.

Firehed:
## Summary

Introduces Flysystem as a filesystem abstraction layer for OpenEMR, enabling:
- Testable file I/O via in-memory adapters or mocks
- Future remote storage support (S3, etc.)
- Centralized storage configuration

Fixes #11505

## Changes

- **Storage abstraction**: `Location` enum + `Manager`/`ManagerInterface` for accessing filesystems by semantic location
- **ServiceContainer integration**: `getStorageManager()` wires up default local filesystems based on `OE_SITE_DIR`
- **First migration**: `BillingLogger` converted from `file_get_contents`/`file_put_contents` to `FilesystemOperator`
- **Test infrastructure**: `league/flysystem-memory` added for isolated testing (mocks can be used situationally too)

## Usage

```php
// DI, preferred:
class Foo
{
  private readonly FilesystemOperator $storage;
  public function __construct(Storage\ManagerInterface $manager)
  {
    $this->storage = $manager->getStorage(Location::Foo);
  }

  public function doThing()
  {
    $this->storage->write('foo.txt', 'some story');
  }
}
```

```php
// Legacy, where DI isn't fully available:
$fs = ServiceContainer::getStorageManager()->getStorage(Location::Documents);
$content = $fs->read('edi/process_bills.log');
$fs->write('edi/process_bills.log', $content);

// In tests, inject mocked ManagerInterface with InMemoryFilesystemAdapter or FilesystemOperator mock
```


src/Billing/BillingProcessor/BillingLogger.php
```diff
@@ -98,7 +111,7 @@ public function onLogComplete()
             if (OEGlobalsBag::getInstance()->getBoolean('drive_encryption')) {
                 $this->hlog = $this->cryptoGen->encryptStandard($this->hlog, keySource: KeySource::Database);
             }
-            file_put_contents(OEGlobalsBag::getInstance()->get('OE_SITE_DIR') . "/documents/edi/process_bills.log", $this->hlog);
+            $this->filesystem->write(self::LOG_PATH, $this->hlog);
```
bradymiller, line 114: Image

Issue #8177, opened 2025-03-26
Inventory of dependency issues


bradymiller:
Below is an inventory of dependency issues that need addressing.
 
* Composer.json (https://packagist.org/)
  * OpenEMR breaks with following composer updates:
    * laminas-servicemanager 4.4.0 (currently stuck at 3.23.0)
    * lcobucci/jwt 5.5.0 (an aside, this is one of the more frustrating upstream packages to support; this is a dependency of steverhoades/oauth2-openid-connect-server, however since author of lcobucci/jwt has pattern of not supporting breadth of supported php versions, we need to micromanage this one and ensure proper version is selected; since this is a dependency of steverhoades/oauth2-openid-connect-server, would recommend upgrading both at same time; currently stuck at 4.2.1)
    * smarty 5.4.3 (currently stuck at 4.5.5)
    * steverhoades/oauth2-openid-connect-server 3.0.1 (since lcobucci/jwt is a dependency, recommend upgrading both at same time; currently stuck at 2.6.1)
    * zircote/swagger-php 5.0.7 (currently stuck at 3.3.7)
  * Abandoned composer packages:
    * laminas/laminas-json
    * laminas/laminas-loader
    * yubico/u2flib-server
  * Composer issues that need await when openemr no longer support php 8.2:
    * Need to keep phpunit/phpunit at 11.* since 12.* requires php > 8.2
* package.json (https://www.npmjs.com/)
  * OpenEMR breaks with following npm updates:
    * del 8.0.0 (currently stuck at 6.1.1) (package converted to pure ESM)
    * gulp 5.0.0 (currently stuck at 4.0.2)
    * replace-in-file 8.3.0 (currently stuck at 7.2.0) (package converted to pure ESM)
    * bootstrap 5.3.3 (currently stuck at 4.6.2) (Very big project that Tyler is working on)
    * bootswatch 5.3.3 (currently stuck at 4.6.2) (Very big project that Tyler is working on)
    * dwv 0.34.2 (currently stuck at 0.27.1)
  * Abandoned/outdated npm packages:
    * angular (outdated, see here for way forward: https://www.npmjs.com/package/angular)
    * angular-sanitize (outdated, see here for way forward: https://www.npmjs.com/package/angular-sanitize)
    * i18next-xhr-backend (replaced by i18next-http-backend)
    * in the napa section, work on dropping jquery-ui and jquery-ui-themes

output during composer.json install:
```
[...]

PR #11521, opened 2026-04-10
refactor(portal): remove AngularJS from secure_chat and messages

### LINKED ISSUE #6008: angular no longer supported
Not sure what this means (don't know where we are using this package in OpenEMR), but noted this as updating the npm packages (looks like may need to migrate to @angular/core, although have not researched this).
https://docs.angularjs.org/misc/version-support-status

### LINKED ISSUE #8412: PatientPortal Chat Controller Missing
### Describe the bug

@sjpadgett was working on adding or building an extension to the chat feature.

     use OpenEMR\Core\Header;
     use OpenEMR\PatientPortal\Chat\ChatController;
    
     define('C_USER', IS_PORTAL ?: IS_DASHBOARD); 
     define('CHAT_HISTORY', '150');
     define('CHAT_ONLINE_RANGE', '1');
     define('ADMIN_USERNAME_PREFIX', 'adm_');

      // Start application.
       $msgApp = new ChatController();

The controller did not make it into the codebase. 

### To Reproduce
https://komododecks.com/recordings/r26CUlsrK7bv99j9vIIA   <--- in the demo system

See video

### Expected behavior

Duh, chat!

### Client configuration

- Browser:  All
- OpenEMR version: 7.0.3 (4)

rollingventures:
## Summary

AngularJS (v1.8.3) has been end-of-life since January 2022 and is listed in #8177 as an abandoned package requiring removal. It was only used in the two portal messaging files and is flagged by npm audit with multiple high/moderate CVEs (CVE-2024-8372, CVE-2024-21490, and others — see [advisory list](https://www.npmjs.com/package/angular)).

This PR removes AngularJS from the portal messaging surface. `messages.php` is rewritten in vanilla JavaScript; `secure_chat.php` is **deleted outright** (see below). It drops the following dependencies:

- `angular` (1.8.3)
- `angular-sanitize` (1.8.3)
- `angular-summernote` (0.8.1)
- `checklist-model` (1.0.0)

along with their asset registrations in `config/config.yaml`.

## What changed

- **`portal/messaging/messages.php` + new `portal/messaging/js/messages.js`** — Inbox/Sent/All/Archive with threading and compose/reply/forward. The Angular `emrMessageApp` is replaced with a vanilla JS module (state object, `fetch`, event listeners, manual DOM rendering). PHP→JS values are passed via a `window.OE_MESSAGES_CONFIG` block (escaped through `json_encode`/`js_escape`/`xlj`). Summernote (jQuery) and DOMPurify sanitization are preserved. Every interpolation into `innerHTML` goes through an `escapeHtml` helper (attribute-safe) or DOMPurify; message bodies render via `renderMessageBody` (DOMPurify, `FORBID_TAGS: ['a','img']`).
- **`portal/messaging/secure_chat.php` + `portal/messaging/js/secure_chat.js` — deleted.** The secure-chat page is dead and non-functional: its menu entry was removed in #7556 (leaving `viewPortalChats()` as dead code in `user_data_view_model.js`, also removed here), and it instantiates `OpenEMR\PatientPortal\Chat\ChatController`, a class that no longer exists in the codebase, so the page fatals with a 500 on any direct access (#8412). Rather than refactor AngularJS out of a page that cannot run, it is removed entirely.
- **`portal/messaging/handle_note.php` + new `src/Services/PortalMessagingSender.php`** — Staff/dashboard message submissions now derive the sender identity from the authenticated session instead of trusting client-supplied `sender_id`/`sender_name`, closing an impersonation gap. The resolution logic is extracted into `PortalMessagingSender` (unit-tested). Also fixes a staff display-name lookup that always returned null (missing `CONCAT(...) AS display_name` alias).
- **Tests** — `PortalMessagingSenderTest` (PHPUnit, isolated) covers the sender-resolution rules; `portal-messaging-helpers.test.js` (Jest) covers the security-sensitive `messages.js` helpers (`escapeHtml` in text and attribute contexts, `htmlToText`, `renderMessageBody`, `paginate`, `groupChain`).
- **`config/config.yaml`, `package.json`, `package-lock.json`** — Removed the four AngularJS dependencies and their asset definitions.
- **`codecov.yml`** — Excludes the bootstrap/entry-point scripts (`messages.php`, `handle_note.php`) from coverage; the testable logic lives in `PortalMessagingSender`.
- **`.phpstan/baseline/`, `.phpstan/fatal-baseline-caps.php`** — Regenerated to reflect the rewritten `messages.php` and the deleted `secure_chat.php`; `class.notFound` fatal cap lowered 167→166 (the `ChatController` not-found entry is gone).

## What didn't change

- PHP session handling, auth checks, SQL queries, CSRF validation
- Visible UI, layout, styles for the messaging inbox
- Summernote editor (jQuery plugin) and DOMPurify HTML sanitization
- Bootstrap modals

## Verification

Exercised live in a local dev instance (patient portal + staff dashboard):

- [x] Inbox/Sent/All/Archive navigation, thread view, pagination
- [x] Compose, send, and reply across patient portal and staff dashboard
- [x] Stored-XSS payloads in subject/body render inert (DOMPurify + `escapeHtml`)
- [x] `PortalMessagingSenderTest` and `portal-messaging-helpers.test.js` pass
- [x] Pre-commit suite passes: PHPCS, PHPStan (level 10), Rector

## Related

- Fixes part of #8177 (abandoned npm package removal)
- Closes #8412 (missing PatientPortal `ChatController`) — resolved by removing the abandoned secure-chat feature, not by restoring the controller. The page had been unreachable since #7556 and fataled on the missing class; deleting it removes the only consumer.
- Relates to closed #6008 ("angular no longer supported")


portal/messaging/messages.php
```diff
@@ -130,669 +130,887 @@ function getAuthPortalUsers()
...
-                    if (compose.task == 'forward') { // Just overwrite default reply but send to pnotes.
-                        compose.sender_id = $("#selForwardto option:selected").val();
-                        compose.sender_name = $("#selForwardto option:selected").text();
-                        compose.selrecip = compose.recipient_id;
-                    } else {
-                        compose.inputBody = $("#inputBody").summernote('code');
-                    }
-                    return true; // okay to submit
+                // Let the form submit naturally with all values now populated
+                return true;
```
kojiromike: Bug: `submitForm()` populates every hidden input except `#replyid`. The original Angular code kept it in sync with `ng-value='selected.reply_mail_chain'`.

`handle_note.php` reads `$_POST['replyid']` (line 102) and passes it to `sendMail()` as `$reply_noteid`, which ultimately becomes the `reply_mail_chain` column in `onsite_mail`. With this always empty, message threading for replies and forwards could break.

Suggested fix — add before `return true`:
```suggestion
                document.getElementById('replyid').value = state.selected?.reply_mail_chain || '';

                // Let the form submit naturally with all values now populated
                return true;
```

portal/messaging/messages.php
```diff
@@ -130,669 +130,887 @@ function getAuthPortalUsers()
...
+                            <ul class="dropdown-menu dropdown-menu-right">
+                                <li>
+                                    <a class="dropdown-item" href="javascript:;" id="btnReadAll"><?php echo xlt('Mark all as read'); ?></a>
+                                </li>
+                                <li class="dropdown-divider"></li>
+                                <li>
+                                    <a class="dropdown-item" href="" data-mode="add" data-toggle="modal" data-target="#modalCompose"><i class="fa fa-edit"></i> <?php echo xlt('Compose Message'); ?></a>
+                                </li>
+                                <li>
+                                    <a class="dropdown-item" href="javascript:;" id="btnBatchDelete"><i class="fa fa-trash"></i> <?php echo xlt('Send Selected to Archive'); ?></a></li>
```
kojiromike, line 219: Minor: the original had `ng-show='!isTrash'` on this menu item, hiding it when viewing the archive folder. The new code shows it unconditionally. Archiving already-archived messages is likely a no-op on the backend, but it's a behavior change from the original UI.

portal/messaging/messages.php
kojiromike: Gentle reminder — this is still unfixed in the latest commit. `submitForm()` now populates all other hidden inputs but `#replyid` is still never set. Message threading will break for replies and forwards since `handle_note.php` reads `$_POST['replyid']` and passes it as `reply_mail_chain`.

portal/messaging/messages.php
kojiromike: Still showing unconditionally in the latest commit. Low priority but worth a quick fix — hiding the archive option when already in the archive folder matches the original behavior.

kojiromike (commented):
Thanks for the work on removing AngularJS from portal messaging. Two previously flagged items are still open (replied on those threads). One new finding:

**XSS risk in `secure_chat.php` `renderMessages()`**

The message body is rendered via `jsAttr(message.message)` in a template literal that gets inserted as `innerHTML`. `jsAttr` escapes for attribute context (HTML entities for `& < > "`), but it does not sanitize HTML — so `` or other event-handler-based payloads would survive.

The original Angular code used `ng-bind-html` with `$sce` sanitization. The `messages.php` side of this PR correctly uses `renderMessageBody()` with proper sanitization. `secure_chat.php` should use the same approach (DOMPurify or equivalent) for message body rendering.

This is patient portal code handling user-supplied content, so the XSS surface matters.

rollingventures:
Thanks for the detailed review! I've addressed all the feedback:

**Fixed in this commit:**

1. **XSS vulnerability in `secure_chat.php`** - Added `sanitizeHtml()` function that removes script tags and event handlers (like `onerror`, `onclick`) before rendering message bodies. This replaces the unsafe `jsAttr()` usage that didn't properly sanitize HTML for innerHTML context.

2. **Type mismatch with `recip_id`** - All `recip_id` values are now normalized to strings using `String()` before comparisons. Also switched from `indexOf()` to `includes()` for cleaner code.

3. **Missing `#replyid` population** - Added `document.getElementById('replyid').value = state.selected?.mail_chain || '';` in `submitForm()` to ensure message threading works correctly for replies and forwards.

4. **Archive option showing in trash folder** - Added logic to hide the archive menu item when `state.isTrash` is true, matching the original Angular behavior.

The other items from Copilot's review were already addressed in previous commits:
- `replaceShortcodes()` chaining is correct
- `htmlToText()` returns plain text (no double escaping)
- `IS_PORTAL`/`IS_DASHBOARD` use `json_encode()`
- Compose modal populates `#referMsg` correctly

kojiromike:
@aisle-analyzer

rollingventures:
Addressed the Aisle security findings:

### 1. Stored XSS via shortcodes (secure_chat.php)

- Added `safeUrl()` function that validates URLs are `http:` or `https:` only, returning `#` for invalid/dangerous schemes
- Updated `replaceShortcodes()` to use `safeUrl()` for both `[img]` and `[url]` shortcodes
- Enhanced `sanitizeHtml()` to also strip dangerous URL schemes (`javascript:`, `data:`, `vbscript:`) from `href`, `src`, `action`, `formaction`, and `xlink:href` attributes

### 2. Sender impersonation (handle_note.php)

- For staff/dashboard sessions, sender identity (`sender_id`, `sender_name`) is now derived server-side from the authenticated session
- Client-supplied values are ignored for staff sessions
- Patient portal sessions already had proper validation (lines 44-65)

portal/messaging/messages.php
rollingventures: Fixed in fc2775807 — `#replyid` is now populated from `state.selected.reply_mail_chain` before the form submits. Thanks for the catch.

portal/messaging/messages.php
rollingventures: Fixed in 04ceade47 — the archive menu item is now hidden when viewing the trash folder, matching the original `ng-show="!isTrash"` behavior.

portal/messaging/secure_chat.php
```diff
@@ -106,319 +106,510 @@
...
+
+                html += `<div class="direct-chat-msg ${alignClass}">
+                    <div class="direct-chat-info clearfix">
+                        <span class="direct-chat-name ${nameFloat}">${escapeHtml(message.username)}</span>
+                        <span class="direct-chat-timestamp ${timeFloat}">${escapeHtml(message.date)}</span>
+                    </div>
+                    <i class="direct-chat-img fa ${handIcon}" style="cursor: pointer; font-size: 24px" data-sender-id="${escapeHtml(message.sender_id)}" data-is-me="${isMe ? '1' : '0'}" title="${clickTitle}"></i>
+                    <div class="direct-chat-text right">
+                        <div class='px-0' title="${clickTitle}" data-sender-id="${escapeHtml(message.sender_id)}" data-is-me="${isMe ? '1' : '0'}">${sanitizedBody}</div>
+                    </div>
```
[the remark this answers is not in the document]
rollingventures: Addressed across two commits — a4ce6fcd5 added a `safeUrl()` URL-scheme allowlist for `[img]`/`[url]` shortcode expansion (so `[url]javascript:...` resolves to `#`), and 56411180d swaps the hand-rolled sanitizer for `DOMPurify.sanitize()` (loaded as a `Header::setupHeader` asset, matching how `messages.php` already does it). DOMPurify with the html profile blocks `javascript:` URLs and event handlers by default, so the unsafe rendering path you describe is closed off at both the shortcode expansion step and the final sanitize step.

rollingventures:
Quick note on the two failing checks after the rebase:

- **`codecov/patch` (0%)** — the diff is ~91 lines of inline JavaScript in `<script>` heredocs inside the PHP files. PHP coverage tooling doesn't instrument inline JS, and there's no JS test suite for `portal/messaging/*`. The pre-change AngularJS code in these same files was at 0% coverage too, so this isn't a regression — just a structural artifact of where the code lives.
- **`codecov/project` (-9.61pp)** — codecov flagged this with `HEAD has 18 uploads less than BASE` (missing `php8.5`, `inferno`, `phpunit`, `http`, `e2e`/`e2e-tests` flags). Those are CI flag uploads that this PR's run didn't produce; they aren't jobs that broke. Codecov's own summary line above the diff says `All tests successful. No failed tests found.`

All 81 real test/lint/static-analysis jobs are green (PHPUnit isolated x5, integration tests across PHP 8.2-8.5 × MySQL/MariaDB matrix, e2e, api, services, email, PHPStan level 10, PHPCS, Rector, semgrep, composer-require-checker, etc.). Happy to add JS unit tests for the rewritten code in a follow-up if that's wanted, but that would be a larger separate change (extract the inline JS into module files + add Jest tests).

rollingventures:
@kojiromike please run aisle when possible

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,749 @@
...
+            FORBID_TAGS: ['a', 'img']
+        });
+    }
+
+    function htmlToText(html) {
+        const hold = document.createElement('DIV');
+        hold.innerHTML = DOMPurify.sanitize(html, {
+            USE_PROFILES: { html: true },
+            FORBID_TAGS: ['a', 'img']
+        });
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c with an inline nosemgrep suppression. The DOMPurify.sanitize() call runs the HTML profile with FORBID_TAGS=['a','img'], and we only read hold.textContent afterwards — innerHTML is just a parse step.

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,749 @@
...
+                    <span class="col-sm-1 px-1 msg-click" data-idx="${idx}"><span style="${strong}">${escapeHtml(item.message_status)}</span></span>
+                    <span class="col-sm-2 px-1 msg-click" data-idx="${idx}"><span style="${strong}">${escapeHtml(dateStr)}</span></span>
+                    <span class="col-sm-3 px-1 msg-click" data-idx="${idx}">
+                        <a class="btn-link"><span style="${strong}">${escapeHtml(item.sender_name)} to ${escapeHtml(item.recipient_name)}</span></a>
+                    </span>
+                    <span class="col-sm-1 msg-click" data-idx="${idx}">
+                        <a class="btn-link"><span style="${strong}">${escapeHtml(item.title)}</span></a>
+                    </span>
+                    <span class="col-sm-4 px-1 msg-click" data-idx="${idx}"><span style="${strong}">${escapeHtml(bodyPreview)}</span></span>
+                </td>
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c — this match anchors on a template-literal interpolation inside the forEach; the actual innerHTML assignment a few lines below now carries the nosemgrep comment. Every ${} in the loop goes through escapeHtml().

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,749 @@
...
+            </tr>`;
+        });
+        tbody.innerHTML = html;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. Every ${} interpolation in the html string goes through escapeHtml() (textContent-based — only entity-escaped output).

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,749 @@
...
+        container.style.display = state.selected ? '' : 'none';
+        if (!state.selected) return;
+
+        const sel = state.selected;
+        const headerEl = document.getElementById('detailHeader');
+        if (headerEl) {
+            headerEl.innerHTML = `<h5 class="pt-2">
+                <a href="javascript:;" id="btnCloseDetail">${escapeHtml(config.strings.conversationFrom)}</a>
+                <strong>${escapeHtml(sel.sender_name)}</strong> ${escapeHtml(config.strings.regarding)} <strong>${escapeHtml(sel.title)}</strong> ${escapeHtml(config.strings.onPrep)} &lt;${escapeHtml(sel.date)}&gt;
+            </h5>`;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. All ${} interpolations in the h5 template go through escapeHtml().

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,749 @@
...
+                html += `<button class="btn btn-small btn-primary btn-delete-item" data-item-id="${escapeHtml(sel.id)}" title="${escapeHtml(config.strings.archiveTitle)}"><i class="fa fa-trash fa-1x"></i></button>`;
+            }
+            html += `</span>`;
+
+            if (isSelected) {
+                html += `<div class='col jumbotron jumbotron-fluid my-3 p-1 bg-light text-dark rounded border border-info'><span>${renderMessageBody(sel.body)}</span></div>`;
+            }
+            html += `</td></tr>`;
+        });
+        chainTbody.innerHTML = html;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. ${} interpolations use escapeHtml(); the only HTML body comes from renderMessageBody() which runs DOMPurify.sanitize() with the html profile + FORBID_TAGS=['a','img'].

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,749 @@
...
+        if (!pagingEl) return;
+        pagingEl.style.display = state.selected ? 'none' : '';
+
+        const currentPageItems = state.pagedItems[state.currentPage] || [];
+        const startNum = (state.itemsPerPage * state.currentPage) + 1;
+        const endNum = (state.itemsPerPage * state.currentPage) + currentPageItems.length;
+
+        const infoEl = document.getElementById('pagingInfo');
+        if (infoEl) {
+            infoEl.innerHTML = `<strong>${startNum}</strong>~<strong>${endNum}</strong> of <strong>${state.items.length}</strong>`;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. startNum and endNum are integers derived from itemsPerPage * currentPage; state.items.length is a Number — no string interpolation surface.

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,749 @@
...
+        // Populate the "Refer to Message" panel with the sanitized
+        // body of the currently selected message when replying or
+        // forwarding. This replaces the former Angular ng-bind-html.
+        const referMsgEl = document.getElementById('referMsg');
+        const referLabelEl = document.getElementById('referLabel');
+        const referMsgIdEl = document.getElementById('referMsgId');
+        const showRefer = (mode === 'forward' || mode === 'reply') && state.selected && state.selected.mail_chain;
+        if (referMsgEl) {
+            if (showRefer) {
+                referMsgEl.innerHTML = renderMessageBody(state.selected.body);
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. renderMessageBody() runs DOMPurify.sanitize() with the html profile and FORBID_TAGS=['a','img'].

portal/messaging/js/secure_chat.js
```diff
@@ -0,0 +1,531 @@
...
+        const container = document.getElementById('currentRecipients');
+        const users = unique(state.chatusers, 'username');
+        let html = '';
+        users.forEach(user => {
+            const recipId = String(user.recip_id);
+            if (state.pusers.includes(recipId) && recipId !== state.userid) {
+                html += `<label><input type="checkbox" data-recip-id="${escapeHtml(recipId)}" data-section="current" checked> ${escapeHtml(user.username)}</label>`;
+            }
+        });
+        container.innerHTML = html;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. Every ${} interpolation goes through escapeHtml().

portal/messaging/js/secure_chat.js
```diff
@@ -0,0 +1,531 @@
...
+        const users = unique(state.chatusers, 'username');
+        let html = '';
+        users.forEach(user => {
+            const recipId = String(user.recip_id);
+            if (!state.isPortal || (state.isPortal && user.dash)) {
+                const checked = state.pusers.includes(recipId) ? 'checked' : '';
+                html += `<label><input type="checkbox" data-recip-id="${escapeHtml(recipId)}" data-section="available" ${checked}> ${escapeHtml(user.username)}</label>`;
+            }
+        });
+        container.innerHTML = html;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. Every ${} interpolation goes through escapeHtml(); the unquoted `checked` token is the literal 'checked' or ''.

portal/messaging/js/secure_chat.js
```diff
@@ -0,0 +1,531 @@
...
+                    <span class="direct-chat-name ${nameFloat}">${escapeHtml(message.username)}</span>
+                    <span class="direct-chat-timestamp ${timeFloat}">${escapeHtml(message.date)}</span>
+                </div>
+                <i class="direct-chat-img fa ${handIcon}" style="cursor: pointer; font-size: 24px" data-sender-id="${escapeHtml(message.sender_id)}" data-is-me="${isMe ? '1' : '0'}" title="${escapeHtml(clickTitle)}"></i>
+                <div class="direct-chat-text right">
+                    <div class='px-0' title="${escapeHtml(clickTitle)}" data-sender-id="${escapeHtml(message.sender_id)}" data-is-me="${isMe ? '1' : '0'}">${sanitizedBody}</div>
+                </div>
+            </div>`;
+        });
+        container.innerHTML = html;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. ${} interpolations use escapeHtml(); sanitizedBody = DOMPurify.sanitize(message.message) with the html profile (default URL/attribute allowlist).

portal/messaging/js/secure_chat.js
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    function renderOnlines() {
+        const container = document.getElementById('onlineUsers');
+        const users = unique(state.onlines, 'username');
+        let html = '';
+        users.forEach(ol => {
+            html += `<label><input type="checkbox" disabled> ${escapeHtml(ol.username)}</label>`;
+        });
+        container.innerHTML = html;
```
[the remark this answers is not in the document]
rollingventures: Resolved in 9071d087c. The only ${} interpolation goes through escapeHtml().

kojiromike:
@aisle-analyzer

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,774 @@
...
+
+    // --- Helpers ---
+
+    function escapeHtml(text) {
+        if (text === null || text === undefined) return '';
+        const div = document.createElement('div');
+        div.textContent = String(text);
+        return div.innerHTML;
+    }
+
```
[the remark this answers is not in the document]
rollingventures: Fixed in a7ce48da1. `escapeHtml()` in both `messages.js` and `secure_chat.js` now entity-escapes `"` and `'` in addition to `< > &`.

Reproduced the exploit on a live dev stack before the fix; after the fix, sending a message with the exact Aisle payload (`" autofocus onfocus="alert(1)`) as the subject and viewing the thread:

```js
> document.querySelector('button[data-mtitle]').getAttribute('data-mtitle')
'" autofocus onfocus="alert(1)'
> document.querySelector('button[data-mtitle]').hasAttribute('autofocus')
false
> document.querySelector('button[data-mtitle]').hasAttribute('onfocus')
false
> window.__OE_ATTACK_FIRED__
undefined
```

The payload survives as the literal `data-mtitle` value (correct — apps must be able to store quote characters); no attributes injected.

Added 5 regression tests including a DOM-level assertion that parses HTML built with the malicious payload through the exact vulnerable template, so future quote-handling regressions fail at the parser level, not just at the string level.

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,783 @@
...
+        });
+        return hold.textContent || hold.innerText || '';
+    }
+
+    function limitTo(str, limit) {
+        if (!str) return '';
+        return str.length > limit ? str.substring(0, limit) : str;
+    }
+
+    async function postForm(url, params) {
```
kojiromike, line 106: `postForm` never checks `response.ok`. On a 4xx/5xx, `response.text()` returns the error body, `JSON.parse` throws, and the wrapper returns the raw text. Read callers (`getAll`/`getSent`/`getDeleted`) then fall through `Array.isArray(data) ? data : []` and silently render an **empty folder**; write callers (`deleteMessage`, send) treat the error page as success and mutate the UI anyway. AngularJS `$http` rejected on non-2xx, so this is a behavior regression. Guard the status before parsing:

```js
if (!response.ok) {
    throw new Error(`Request failed (${response.status})`);
}
```

portal/messaging/js/secure_chat.js
```diff
@@ -0,0 +1,570 @@
...
+        // Replace [url] shortcodes with safe href URLs
+        msg = msg.replace(/\[url](.*?)\[\/url]/g, (match, url) => {
+            const safe = safeUrl(url);
+            return `<a href='${safe}'>${escapeHtml(url)}</a>`;
+        });
+        msg = msg.replace(/<img /g, "<img class='img-responsive' ");
+        return msg;
+    }
+
+    async function parseJsonOrText(response) {
```
kojiromike: Same `response.ok` gap as `messages.js` `postForm`: `parseJsonOrText` (used by both `postData` and `postForm`) reads `response.text()` regardless of HTTP status, so a non-2xx silently returns the error body as if it were a successful payload. Add the guard here:

```js
if (!response.ok) {
    throw new Error(`Request failed (${response.status})`);
}
```

portal/messaging/handle_note.php
```diff
@@ -75,6 +77,18 @@
     }
     //owner is the user authUser
     $owner = $session->get('authUser');
+
+    // For staff/dashboard sessions, derive sender identity from authenticated
+    // session to prevent impersonation via client-supplied values.
+    $staffSenderId = (string) $session->get('authUser');
```
kojiromike: `(string) $session->get('authUser')` casts `mixed`, which trips PHPStan's `Cannot cast mixed to string` — and the fix here was to bump `.phpstan/baseline/cast.string.php` 1→2 rather than fix the source. Narrow instead of cast (and drop the baseline bump):

```php
$authUser = $session->get('authUser');
$staffSenderId = is_string($authUser) ? $authUser : '';
```

This also makes the later `is_string($resolvedStaffSenderId)` guard meaningful instead of redundant. Then regenerate with `composer phpstan-baseline` so the count drops back.

portal/messaging/js/secure_chat.js
```diff
@@ -0,0 +1,570 @@
...
+    function uncheckAll() {
+        state.pusers = [];
+        getAuthUsers();
+    }
+
+    // --- Init ---
+
+    function init() {
+        listMessages();
+        state.pidMessages = setInterval(listMessages, 3000);
```
kojiromike: `setInterval(listMessages, 3000)` calls `listMessages` with the timer id as its first argument, which `listMessages` reads as `wasMySubmission` (always truthy) — so every polled refresh is treated as your own send and the new-message beep / title flash / browser-notification path is skipped. Incoming messages from other users never notify. The old AngularJS code had the same bug, so it isn't introduced here, but the rewrite is a good moment to fix it:

```js
state.pidMessages = setInterval(() => listMessages(), 3000);
```

portal/messaging/js/messages.js
```diff
@@ -0,0 +1,783 @@
...
+        state.selected = item;
+        renderAll();
+    }
+
+    function closeMessage() {
+        state.selected = null;
+        groupToPages();
+    }
+
+    function readAll() {
```
kojiromike, line 335: `readAll()` ("Mark all as read") only flips `message_status` on the in-memory items and re-renders — it never POSTs to the server, so the change is lost on reload. Pre-existing behavior carried over from the Angular version, flagging since the rewrite touches it. If it should persist, fire the `setread` action for the affected ids.

src/Services/PortalMessagingSender.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Resolve sender identity. If staff identity is available (i.e. the request
+     * came from an authenticated staff session) it takes priority over any
+     * client-supplied sender_id / sender_name. Empty strings are returned in
+     * place of nulls so the result can flow into addPnote() / sendMail() which
+     * declare string parameters.
+     *
+     * @return array{0: string, 1: string} tuple of [sender_id, sender_name]
+     */
+    public static function resolve(
```
kojiromike: `resolve()` is `static`, which diverges from our constructor-DI standard and means `handle_note.php` calls it as a service locator. It's a pure function with no state or dependencies, so this is low-risk — but worth a deliberate call: keep it static as an explicit pure-function exception, or make it an instance method for consistency with `BaseService`. Not blocking.

tests/js/portal-messaging-helpers.test.js
```diff
@@ -0,0 +1,399 @@
...
+loadModule('portal/messaging/js/secure_chat.js');
+
+const messagesHelpers = global.window.__OE_MESSAGES_TEST__;
+const secureChatHelpers = global.window.__OE_SECURE_CHAT_TEST__;
+
+// ---------------------------------------------------------------------------
+// messages.js: escapeHtml — the load-bearing XSS guard for every ${} in
+// template literals fed into innerHTML.
+// ---------------------------------------------------------------------------
+describe('messages.js escapeHtml', () => {
```
kojiromike, line 79: This suite thoroughly covers the security-sensitive pure helpers (escapeHtml attribute context, safeUrl, replaceShortcodes, the DOMPurify wrappers) — good. What's missing is the behavioral logic that's most likely to regress and is just as testable: pagination math (`groupToPages` boundaries, prev/next disabled states, the items-range display), threading (`Chained` grouping by `mail_chain`), and new-message dedup (`lastMessageId` compare). Worth adding cases for those.

portal/messaging/js/messages.js
kojiromike: Fixed in 0179e7be15 — added the `response.ok` guard in `postForm` before parsing the body.

portal/messaging/js/secure_chat.js
kojiromike: Fixed in 0179e7be15 — added the `response.ok` guard in `parseJsonOrText`, which covers both `postData` and `postForm`.

portal/messaging/handle_note.php
kojiromike: Fixed in 0179e7be15 — narrowed with `is_string($authUser)` instead of casting, and regenerated the baseline so the `cast.string` count for this file drops back to 1.

portal/messaging/js/secure_chat.js
kojiromike: Fixed in 0179e7be15 — `setInterval(() => listMessages(), 3000)`, so the timer id is no longer passed as `wasMySubmission`.

portal/messaging/js/messages.js
kojiromike: Fixed in 0179e7be15 — `readAll()` now fires `setread` for each unread item and mirrors the status across the cached folder lists, so it persists on reload.

src/Services/PortalMessagingSender.php
kojiromike: Done in 0179e7be15 — `resolve()` is now an instance method; `handle_note.php` and the isolated test call it on an instance.

tests/js/portal-messaging-helpers.test.js
kojiromike: Added in 0179e7be15 — extracted pure `paginate`/`groupChain` (messages.js) and `lastIdOf` (secure_chat.js) and added Jest cases for pagination boundaries, thread grouping by `mail_chain`, and last-message-id dedup.

kojiromike:
@bradymiller @sjpadgett it seems the "secure_chat" feature of the portal was removed in https://github.com/openemr/openemr/pull/7556. I could not get it to work before or after this change, so I'm removing it. I think if there's interest in resurrecting it it can come after, and shouldn't block removing Angular.

kojiromike (approved):

PR #11534, opened 2026-04-10
chore(ci): bump mariadb to 12.2

stephenwaite:


ci/apache_85_122/docker-compose.yml
```diff
@@ -7,6 +7,6 @@ x-includes:
 
 services:
   mysql:
-    image: mariadb:12.0@sha256:607835cd628b78e2876f6a586d0ec37b296c47683b31ef750002d3d17d3d8f7a
+    image: mariadb:12.2@sha256:b2595ba33da05f355446a34c9cf1ce30807e96ea7cb4e7e0cc283b941b7c5544
```
kojiromike: hang on, I'm still in the middle of fixing dependabot up to handle this...

kojiromike: unless you want to merge this first, and then dependabot can adjust?

kojiromike: @stephenwaite so wait a sec, do we actually want dependabot to bump all our images minor and patch versions, and just leave major alone?

stephenwaite: guess the only way to keep track of EOL is a manual review?

stephenwaite: should dependabot only track the image bumps for services in the production docker?

kojiromike: IMO dependabot should track image bumps everywhere. The only place this is kinda complicated is where we actually want to keep old versions of php and mysql around for testing. There, we have to keep dependabot from updating everything to the same, latest version.

kojiromike: > guess the only way to keep track of EOL is a manual review?

basically, I suppose. I mean, sometimes they'll show up as security issues, but that's not really where we want to see it first. :)

kojiromike (approved):

PR #11550, opened 2026-04-11
feat(infrastructure): Server configs and dockers for front-controller

Firehed:
Following #9943, we want some server configs that are designed to test out and iterate on the front-controller, hopefully to eventually become the "officially sanctioned" way to run stock OpenEMR. This creates such files, along with some surrounding infrastructure to test things out more easily and thoroughly during development.

The docker stuff, in particular, departs with some of the previous approaches. The images and compose configs are NOT production ready (nor intended to be), nor are they intended to solve everyone's development problems. There are no provisions for coverage tools, xdebug, etc. It's more of a "serve from docker, run php tools locally" setup. It uses official PHP images instead of installing it on an OS base. It doesn't auto-run setup, perform upgrades, or anything else of that nature. _**It will not run CCDA services**_. But it does route all traffic through the FC, and is _relatively_ straightforward as deployments go.

- Adds webserver config files referencing the front controller for nginx(php-fpm), apache(php-fpm), and apache(mod_php)
- Adds php and php-fpm config files w/ recommended basic starting point (this needs work, doesn't align with setup recommendations)
- Adds a barebones docker setup for running traffic through the FC
- Amends the FC logging to include the SAPI, since following this is tough testing three deployments in parallel
- Wrapper shell script for opening the browser dealing with the ephemeral ports (this may go)
- Adds a `.dockerignore` (this probably needs to be moved to avoid clobbering other stuff)

Yes, Claude did most of the Apache work, and did some minor tweaks on nginx.


Firehed:
@bradymiller This is the PR I mentioned on the admin call that had some experimental Docker work. I'd have to refresh my memory a bit on where it is today. I think getting it _fully_ working requires adjustments to the CCDA stuff, but I think it's directionally useful

PR #11571, opened 2026-04-12
refactor(ci): modernize auto_prepend.php

kojiromike:
## Summary

- Convert test-type if/elseif chain to a `match(true)` expression
- Convert coverage-driver switch to `match` expression
- Use `str_contains()` instead of `strpos() !== false`
- Use first-class callable syntax for `register_shutdown_function`

## Test plan

- [ ] CI passes — this file is exercised by every CI test run

### DIFF
--- a/ci/auto_prepend.php
+++ b/ci/auto_prepend.php
@@ -32,13 +32,11 @@ const COVERAGE_DIR = '/tmp/openemr-coverage';
 // Inferno tests are identified by INFERNO_TEST environment variable
 // API tests use /apis/ endpoints, E2E tests use other routes
 $requestUri = $_SERVER['REQUEST_URI'] ?? '';
-if (getenv('INFERNO_TEST') === 'true') {
-    $testType = 'inferno';
-} elseif (strpos($requestUri, '/apis/') !== false) {
-    $testType = 'api';
-} else {
-    $testType = 'e2e';
-}
+$testType = match (true) {
+    getenv('INFERNO_TEST') === 'true' => 'inferno',
+    str_contains((string) $requestUri, '/apis/') => 'api',
+    default => 'e2e',
+};
 
 define('TEST_TYPE', $testType);
 define('COVERAGE_SUBDIR', COVERAGE_DIR . '/' . $testType);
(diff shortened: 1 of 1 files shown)


ci/auto_prepend.php
```diff
@@ -62,17 +60,11 @@
...
-            break;
-        default:
-            error_log("Prepend: No coverage driver available (need pcov or xdebug)");
-            break;
-    }
+    match (COVERAGE_DRIVER) {
+        'pcov' => \pcov\start(),
+        'xdebug' => xdebug_start_code_coverage(XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE),
+        default => error_log("Prepend: No coverage driver available (need pcov or xdebug)"),
+    };
```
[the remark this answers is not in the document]
kojiromike: PHP allows void function calls as `match` arms when the result is discarded (statement context). This has been valid since PHP 8.0 — confirmed locally and this pattern was generated by Rector's `ChangeSwitchToMatchRector`.

```php
function void_fn(): void {}
match ('a') {
    'a' => void_fn(),
    default => void_fn(),
};
// No error
```

Keeping as-is.

ci/auto_prepend.php
```diff
@@ -61,18 +59,14 @@
     }
 }
 
+// error_log is intentional — this file runs before the OpenEMR bootstrap,
+// so no PSR-3 logger is available.
```
kojiromike:
```suggestion
```

ci/auto_prepend.php
```diff
@@ -61,18 +59,14 @@
...
-            xdebug_start_code_coverage(XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE);
-            break;
-        default:
-            error_log("Prepend: No coverage driver available (need pcov or xdebug)");
-            break;
-    }
+    match (COVERAGE_DRIVER) {
+        'pcov' => \pcov\start(),
+        'xdebug' => xdebug_start_code_coverage(XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE),
+        default => error_log("Prepend: No coverage driver available (need pcov or xdebug)"),
```
kojiromike:
```suggestion

        // No PSR-3 logger is available before the OpenEMR bootstrap,
        // so error_log is necessary here.
        default => error_log("Prepend: Coverage is enabled, but no coverage driver (pcov or xdebug) is available"),
```

ci/auto_prepend.php
kojiromike: Applied — standalone comment removed, moved inline to the `default` arm.

ci/auto_prepend.php
kojiromike: Applied — comment moved inline and message improved.

PR #11590, opened 2026-04-13
fix(faxsms): stop clobbering pc_apptstatus with channel name during dedup

### LINKED ISSUE #11479: faxsms: dedup overwrites pc_apptstatus with 'EMAIL'/'SMS' literals, destroying real status

kojiromike:
## Summary

- Stop overwriting `pc_apptstatus` with literal `'EMAIL'`/`'SMS'` strings when marking an appointment as already-notified. The dedup now relies solely on the `pc_sendalertemail`/`pc_sendalertsms` columns, which already exist for this purpose.
- Simplify the dedup WHERE clause in `faxsms_getAlertPatientData()` to check only the alert-sent flag, removing the redundant (and harmful) `pc_apptstatus` check.
- Convert `rc_sms_notification_cron_update_entry()` to accept the `NotificationChannel` enum and use a `match` expression instead of string comparisons.

## Context

Split out from #11477. Once the column-mismatch fix in #11477 made dedup actually work, this became the visible problem: every reminder cron run was wiping the front-desk appointment status (Pending, Arrived, No Show, etc.) with a channel name that has no `list_options` entry.

Closes #11479

## Test plan

- [ ] Run the SMS/email notification cron in test mode (`dryrun`) and verify appointments are selected correctly
- [ ] Run in live mode and confirm `pc_apptstatus` is preserved after notification send
- [ ] Verify dedup still works: a second cron run should skip already-notified appointments

### DIFF
--- a/.phpstan/baseline/cast.int.php
+++ b/.phpstan/baseline/cast.int.php
@@ -293,7 +293,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Cannot cast mixed to int\\.$#',
-    'count' => 2,
+    'count' => 6,
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-faxsms/library/rc_sms_notification.php',
 ];
 $ignoreErrors[] = [
--- a/.phpstan/baseline/return.type.php
+++ b/.phpstan/baseline/return.type.php
@@ -692,11 +692,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-faxsms/ModuleManagerListener.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Function rc_sms_notification_cron_update_entry\\(\\) should return int but returns ADORecordSet\\.$#',
-    'count' => 1,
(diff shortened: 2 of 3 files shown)


kojiromike:
Duplicate of #11589.

PR #11603, opened 2026-04-14
fix(tests): Remove error-suppressing catch block

Firehed:
While working on testing the front-controller, I ran into weird behavior where test suites were _clearly_ showing errors yet CI was passing. After much digging and some lost hair, I found that several places in the codebase call `exit` / `exit(0)` — which causes PHPUnit to exit 0 before reaching its result printer, producing false-passes. Bad stuff. This PR removes the pattern from the specific sites that were immediately problematic and unhides the real test failures they were masking.

The now-visible failures are a much wider problem than this PR can tackle, so the ones that surface here are skipped with explanatory `markTestSkipped` messages. Follow-up issues will track each skip and the broader cleanup (AppDispatch constructor running the full request pipeline, every `exit`/`die` in legacy code, the faxsms module not being installed in CI, etc).

[^1]: Over 1000 😭

- Remove catch/`exit` error-suppression in `AppDispatch::getApiService`/`setApiService`
- Change `AccessDeniedHelper::deny()` so it no longer swallows PHPUnit runs
- Change `empty()` to `=== ''` where that was the same pattern
- Set `oe_enable_email` in the globals bag (not just `$GLOBALS`) so the notification cron test can resolve the email service
- Skip four now-visible failing tests with notes for follow-up:
  - `FaxSmsEmailTest` (whole class) — `AppDispatch` constructor runs the full request pipeline; not instantiable from a test
  - `NotificationCronEmailTest` (whole class) — `rc_sms_notification.php` requires the faxsms module schema, which isn't installed in CI
  - `EmailTestServiceTest::directSendDeliversEmail` — send reports success but message never arrives in Mailpit (likely a null-field bug; needs diagnosis)
  - `EmailTestServiceTest::queueTemplatedInsertsAndDelivers` — same symptom on the templated path

Yes — Claude Code assisted with diagnosis (xdebug function tracing to identify the swallowing `exit` sites) and drafting the skip annotations; all changes reviewed before commit.

#### Follow-up issues

Skips introduced here are tracked for follow-up:

- #11622 — Refactor AppDispatch constructor so subclasses can be instantiated in tests (covers `FaxSmsEmailTest` skip)
- #11623 — NotificationCronEmailTest skipped: faxsms module schema not available in CI
- #11624 — EmailTestServiceTest: direct send and templated queue tests report success but message never arrives in Mailpit

### DIFF
--- a/.phpstan/baseline/binaryOp.invalid.php
+++ b/.phpstan/baseline/binaryOp.invalid.php
@@ -21437,11 +21437,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../tests/Tests/Certification/HIT1/US_Core_311/InfernoSinglePatientAPITest.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Binary operation "\\." between mixed and \'/interface/modules…\' results in an error\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../tests/Tests/E2e/FaxSmsEmailTest.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Binary operation "\\-" between mixed and 1 results in an error\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../tests/Tests/Fixtures/BaseFixtureManager.php',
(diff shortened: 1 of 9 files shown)


Firehed:
I also verified by hand that all CI suites are actually completing now.

PR #11615, opened 2026-04-14
fix(ci): Detect PHPUnit exiting without finishing

### LINKED ISSUE #11614: bug(ci): Exit calls in SUT can give false-positive test pass
Re: #8756 - if code calls exit or die in such a way that the process exits "clean" (with a `0` code), this can abort the test suite. Since the PHPUnit process exits with a success code, CI passes when it should fail or warn.

Firehed:
Fixes #11614

- Creates a PHPUnit extension that registers shutdown detection
- Listens for PHPUnit "finished" events
- If shutdown occurs without having received a finished event, it indicates the SUT triggered an early exit. Exit the process with a nonzero code (some vague searching online indicated 70 is a convention for this type of thing, but any positive integer will fix it)
- Tweaked the dev autoloader so we don't need to make further future adjustments.
- Added a safeguard-for-the-safeguard in the test bootstrap script because of course there was a situation where that was needed 

No


kojiromike:
:boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom::boom:

PR #11618, opened 2026-04-14
fix(bootstrap): replace die() with exception for missing session site_id

### LINKED ISSUE #11616: Replace die() with exception in globals.php for missing site_id
## Summary

`interface/globals.php:260` calls `die("Site ID is missing from session data!")` when `$_SESSION['site_id']` is empty in a non-`$ignoreAuth` context. This is problematic for non-web callers (CLI background services, test harnesses) because `die()` terminates the process immediately, bypassing `try`/`catch`/`finally` blocks and shutdown cleanup.

## Problem

When `BackgroundServiceRunner` executes a service whose `require_once` file pulls in `globals.php`, and the session lacks `site_id`, the `die()` call:

1. Kills the process before the runner's `finally` block can release the DB lock (`UPDATE background_services SET running = 0`)
2. Leaves the service in a permanently locked state (`running = 1`) that must be cleared manually
3. Prevents the shutdown handler from running (registered via `register_shutdown_function`, but `die()` inside an included file during `require_once` can short-circuit it depending on context)

## Suggested fix

Replace the `die()` with a thrown exception:

```php
throw new \RuntimeException("Site ID is missing from session data!");
```

This allows callers that wrap `globals.php` inclusion in a `try`/`catch` to handle the failure gracefully, while preserving the same fatal behavior for unprotected web request paths (uncaught exception → 500 error).

## Note

`portal/patient/scripts/app.js:30` checks for the string "Site ID is missing from session data" in AJAX responses. That handler may need updating if the error output format changes, though it's already a fragile pattern (parsing error strings from HTML output).

kojiromike:
## Summary

- Replace `die("Site ID is missing from session data!")` in `globals.php` with `throw new MissingSiteIdException()`
- Introduce `MissingSiteIdException` extending the existing `MissingSiteException`, so callers can catch either the specific session case or the broader "site not identified" family
- This allows callers like `BackgroundServiceRunner` that wrap `globals.php` inclusion in `try`/`catch` to handle the failure gracefully and release DB locks, instead of having the process killed mid-execution

## Notes

`portal/patient/scripts/app.js:30` checks for the string "Site ID is missing from session data" in AJAX responses. That handler relies on parsing the raw `die()` output, which is already fragile. With this change, uncaught exceptions in production won't produce that exact string in the response body (PHP returns a 500 with no body when `display_errors` is off). The JS handler should be updated separately to use a proper error response format.

## Test plan

- [ ] Verify web requests without a session site_id still get a 400 response (the `http_response_code(400)` call is preserved before the throw)
- [ ] Verify CLI callers (e.g. `BackgroundServiceRunner`) can catch `MissingSiteIdException` and clean up gracefully
- [ ] Verify the patient portal session expiry flow still redirects correctly (may need follow-up for `app.js`)

Closes #11616

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -270,7 +270,7 @@ if (empty($siteId) || !empty($_GET['site'])) {
                 require_once("$srcdir/auth.inc.php");
             }
             http_response_code(400);
-            die("Site ID is missing from session data!");
+            throw new \OpenEMR\Common\System\MissingSiteIdException();
         }
 
         $tmp = $_SERVER['HTTP_HOST'];
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -269,7 +269,6 @@ if (empty($siteId) || !empty($_GET['site'])) {
                 $globalsBag->set('srcdir', $srcdir);
                 require_once("$srcdir/auth.inc.php");
             }
-            http_response_code(400);
             throw new \OpenEMR\Common\System\MissingSiteIdException();
         }
 
--- a/src/Common/System/MissingSiteException.php
+++ b/src/Common/System/MissingSiteException.php
@@ -12,7 +12,9 @@
 
 namespace OpenEMR\Common\System;
 
-class MissingSiteException extends \RuntimeException
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+class MissingSiteException extends BadRequestHttpException
 {
     public function __construct(string $message = "Site directory is not configured. OE_SITE_DIR must be set.", int $code = 0, ?\Throwable $previous = null)
     {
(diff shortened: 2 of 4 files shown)


Firehed:
I updated the exception stack slightly: the lower MissingSiteException now extends Symfony HTTPKernel's `BadRequestHttpException` which will set the 400 code correctly; the system error handler knows how to deal with exceptions in that family.

I'm 99% sure this will produce the results you want, so long as nothing is somehow loading globals.php from within a try/catch block and swallowing things.

kojiromike:
@aisle-analyzer

kojiromike:

Backport touches interface/globals.php, which has diverged between master and rel-810. Verify clean apply or rework manually.

PR #11621, opened 2026-04-14
feat(phpstan): forbid unsafe catch blocks and exit/die inside them

Firehed:

Adds two custom PHPStan rules to flag catch-related patterns that silently mask failures, and baselines the existing violations so CI stays green while they are fixed incrementally.

- **`openemr.exitInCatchOrFinally`** (`tests/PHPStan/Rules/ForbiddenExitInCatchFinallyRule.php`) — flags any `exit`/`die` inside a `catch` or `finally` block (ignoring nested function/closure bodies, which form a new scope). These sites swallow the caught exception and, when bare or given a string argument, also terminate the process with status 0 — the false-positive-success shape that motivated the broader exit/die work.
- **`openemr.forbiddenCatchType`** (`tests/PHPStan/Rules/ForbiddenCatchTypeRule.php`) — flags `catch` declarations whose type would capture a forbidden throwable at runtime (inheritance in either direction). Driven by a list pre-populated with `\Error` and `\ErrorException`; extend `ForbiddenCatchTypeRule::FORBIDDEN` to block more types. `catch (\Exception $e)` is flagged because it would catch `\ErrorException`. Catches that end with an unconditional `throw` (re-raise or wrap) are exempt — the failure still propagates to the global exception handler.
- Registers both rules in `.phpstan/extension.neon`.
- Generates split-identifier baselines for the existing violations (`.phpstan/baseline/openemr.exitInCatchOrFinally.php`, `.phpstan/baseline/openemr.forbiddenCatchType.php`) so existing code does not fail CI while it is cleaned up.

Out of scope: fixing the baselined sites themselves, and the broader `ExitException` / unconditional exit-die rule work (tracked separately on other branches).

### DIFF
--- a/.phpstan/baseline/loader.php
+++ b/.phpstan/baseline/loader.php
@@ -113,6 +113,8 @@ return ['includes' => [
     __DIR__ . '/offsetAssign.dimType.php',
     __DIR__ . '/offsetAssign.valueType.php',
     __DIR__ . '/openemr.deprecatedSqlFunction.php',
+    __DIR__ . '/openemr.exitInCatchOrFinally.php',
+    __DIR__ . '/openemr.forbiddenCatchType.php',
     __DIR__ . '/openemr.forbiddenCurlFunction.php',
     __DIR__ . '/openemr.forbiddenErrorLog.php',
     __DIR__ . '/openemr.forbiddenGlobalKeyword.php',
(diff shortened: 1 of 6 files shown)


Firehed:
Will have baseline collisions w/ #11619. I think land that one first and I'll update.

PR #11649, opened 2026-04-15
chore: regenerate FHIR R4 classes with php-fhir v4.0.3

kojiromike:
## Summary

Exploratory PR to evaluate what regenerating the FHIR R4 classes looks like using the current dcarbone/php-fhir generator (v4.0.3) versus the 2019-era generated code that has been manually maintained since.

**This PR is intentionally broken** — it regenerates the R4 type classes but does not update the 149 consumer files (605 import references) that depend on the old namespace structure. The purpose is to visualize the scope of the change and decide whether to proceed.

### What changed in the generated code

- `declare(strict_types=1)` throughout
- Native PHP type hints on all parameters and returns
- PHP 8.1 enums for code-based types (vs string wrapper classes)
- Built-in validation via `_getValidationErrors()`
- New shared library infrastructure (`Encoding/`, `Client/`, `Types/`, `Validation/`)
- Updated FHIR R4 spec (v4.0.1 with errata since Dec 2018)
- 1,176 generated files (vs 895 original)

### Namespace mapping

| Old | New |
|-----|-----|
| `OpenEMR\FHIR\R4\FHIRDomainResource\FHIRPatient` | `OpenEMR\FHIR\Versions\R4\Types\FHIRResource\FHIRDomainResource\FHIRPatient` |
| `OpenEMR\FHIR\R4\FHIRElement\FHIRString` | `OpenEMR\FHIR\Versions\R4\Types\FHIRElement\FHIRString` |
| `OpenEMR\FHIR\R4\FHIRResource\FHIRBundle` | `OpenEMR\FHIR\Versions\R4\Types\FHIRResource\FHIRBundle` |

### Remaining work (if we proceed)

- Update 149 consumer files with new namespace imports
- Adapt constructor patterns (array-based → new API)
- Adapt serialization code
- Regenerate PHPStan baseline
- Reapply security hardening (LIBXML_NONET) if not handled by generator

### Generation script

Included at `tools/fhir-gen/generate.php` — requires downloading R4 XSD schemas from HL7 first (see `.gitignore` for excluded artifacts).


kojiromike:
## Level of Effort: Completing This Migration

### Tier 1: Mechanical namespace renames (~1-2 days, scriptable)

135 unique import paths across 149 files with 605 references need updating. Every old import maps 1:1 to a new path — only 2 are truly gone (see Tier 2). This is a sed/rector job with zero ambiguity.

Example mapping:
- `OpenEMR\FHIR\R4\FHIRDomainResource\FHIRPatient` → `OpenEMR\FHIR\Versions\R4\Types\FHIRResource\FHIRDomainResource\FHIRPatient`
- `OpenEMR\FHIR\R4\FHIRElement\FHIRString` → `OpenEMR\FHIR\Versions\R4\Types\FHIRElement\FHIRString`
- `OpenEMR\FHIR\R4\FHIRResource\FHIRBundle` → `OpenEMR\FHIR\Versions\R4\Types\FHIRResource\FHIRBundle`

### Tier 2: API changes (~2-3 days, requires thought)

- **`get_fhirElementName()`** → **`_getFHIRTypeName()`** — 15 call sites, straightforward rename
- **`PHPFHIRResponseParser`** → **`ResourceParser`** (in `OpenEMR\FHIR\Encoding`) — 2 files need adapted to the new parsing API
- **`PHPFHIRHelper`** — deleted with no direct replacement, 1 test file references it
- **`jsonSerialize()` / `xmlSerialize()`** — ~20 call sites, signatures may differ
- **Constructor patterns** — 596 constructor calls total, but most pass scalar values (`new FHIRString("value")`) which still work with the new named-parameter constructors. Only ~8 use array-based construction (`new FHIRBundleEntry($array)`) that would need rewriting.

### Tier 3: Integration & validation (~2-3 days)

- Regenerate PHPStan baseline for the new files
- Run the FHIR test suite and fix failures
- Verify JSON/XML serialization output matches expected FHIR format
- Check `LIBXML_NONET` security hardening — the old code had this added manually; the new generator may or may not include it
- Composer autoloading updates for new directory structure

### Summary

**~5-8 days of focused work.** The bulk is mechanical (Tier 1). The risk is in Tiers 2-3 where behavior might subtly change — particularly serialization output and the response parser, since those are the integration surface with external systems.

kojiromike:
Surfacing an incidental benefit noticed during an agentic-tooling audit:

`src/FHIR/R4/PHPFHIRParserMap.php` is 1.8 MB, which exceeds the ~1 MB file-size cap most LSPs (and some static analyzers) enforce. In practice the class is silently dropped from the type graph — IDE jump-to-definition fails on it, agent symbol lookups miss it, PHPStan indexing skips it. Users don't know why; they just see flaky tooling on anything that references the parser map.

This PR deletes that file entirely (−61,474 lines) and replaces the runtime with a proper `src/FHIR/Encoding/ResourceParser.php`. Adapting the 149 consumer files is the real work, but the regeneration would restore LSP/analyzer coverage for the parser as a side effect — worth factoring into the decision to push through.

Assisted-by: Claude Code

kojiromike:
@adunsulag what do you think about us trying to implement this? I'm somewhat amazed that regenerating the fhir code is this massive a change, but at least fixing the src/FHIR/R4/PHPFHIRParserMap.php problem is probably worth doing, not to mention the typing improvements and other modernizations. What would break here if we got it wrong?

adunsulag:
If we are passing inferno we should be good.  If we got this wrong we'd see a lot of problems in the GET requests for the FHIR apis.  There are annoying spots where some of the APIs have relied on the array hydration behavior and others have done specific type handling around FHIRString, FHIRId, etc. Those are the problems off the type of my head.  Again, my thoughts are that if it passes inferno, we are probably good to go forward with this.

kojiromike:
Thank goodness @Firehed noticed the problems with inferno before I decided to dive back into this!

Issue #10771, opened 2026-02-13
Database migrations: add utility methods/trait for common operations


Firehed:
This requires some more research on what makes sense to add. A LOT of the utilities in the home-rolled migration tooling are irrelevant when migrations are managed properly through doctrine/migrations (various conditional DDL changes)

Data changes in the `list_options` table seem relatively common, so having a streamlined way to handle that would be nice. There are probably others that get seeded with a lot of data which would similarly benefit. This doesn't all need to be done at once but I want to be mindful of the need and avoid reinventing wheels.

PR #11653, opened 2026-04-15
chore(db): Baseline tooling for doctrine/migrations conversion

Firehed:
In projects using a lot of migration scripts (as OpenEMR inevitably will), there's an annoying behavior where the tooling will inspect the DB's schema (table by table, IIIRC) and get progressively slower as you have more tables and migration. We're talking minutes, not a few seconds.

This adds a trait that works around the issue, and at the same time sets up some default configuration for all tables. It also adds a convenience wrapper for creating PKs, since the default syntax is quite verbose for the common-case.

Adds new migrations tooling trait and tests

Tests were AI


Firehed:
This was progress on #10771.

PR #11659, opened 2026-04-16
fix(messages): resolve index mismatch in deleteItem using ID-based matching

Bakul2006:
Previously, deletion relied on the index from a filtered view (state.items) to access state.allItems, which could result in removing the wrong message. This change ensures the correct item is deleted regardless of filtering or ordering.In place of that I have implemented filtering on the basis of the Message  ID. Now  This will fix up the issue of deleting wrong messages due to the indexing miss-match  in both the files.

- Switched from index-based to ID-based deletion in `deleteItem()`
- Fixed incorrect deletions caused by mismatch between filtered and full message lists

No

### DIFF
--- a/portal/messaging/messages.php
+++ b/portal/messaging/messages.php
@@ -246,13 +246,18 @@ function getAuthPortalUsers()
                     $scope.currentPage = this.n;
                 };
 
-                $scope.deleteItem = function (idx) {
+                $scope.deleteItem = function (item) {
                     if (!confirm($scope.xLate.confirm.one)) return false;
-                    const itemToDelete = $scope.allItems[idx];
-                    const idxInItems = $scope.items.indexOf(itemToDelete);
-                    $scope.deleteMessage(itemToDelete.mail_chain); // Just this user's message
-                    $scope.items.splice(idxInItems, 1);
-                    $scope.search();
+
+                    const itemId = item && item.id ? item.id : item;
+                    const idxInAllItems = $filter('getById')($scope.allItems, itemId);
+                    if (idxInAllItems === null) {
+                        return false;
+                    }
+
+                    const itemToDelete = $scope.allItems[idxInAllItems];
+                    $scope.deleteMessage(itemToDelete.mail_chain || itemToDelete.id);
+                    $scope.selected = null;
                     $scope.init()
                     return false;
                 };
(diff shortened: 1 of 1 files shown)


portal/messaging/messages.php
```diff
@@ -246,13 +246,18 @@ function getAuthPortalUsers()
...
-                    $scope.search();
+
+                    const itemId = item && item.id ? item.id : item;
+                    const idxInAllItems = $filter('getById')($scope.allItems, itemId);
+                    if (idxInAllItems === null) {
+                        return false;
+                    }
+
+                    const itemToDelete = $scope.allItems[idxInAllItems];
+                    $scope.deleteMessage(itemToDelete.mail_chain || itemToDelete.id);
```
kojiromike, line 259: When does `itemToDelete.mail_chain` end up falsy, warranting the `|| itemToDelete.id` fallback? The server-side `updatePortalMailMessageStatus` in `portal/lib/portal_mail.inc.php:293` keys the delete on `WHERE (mail_chain = ? OR id = ?)`, so both paths resolve to the right rows, but the original comment `// Just this user's message` suggested `mail_chain` was chosen deliberately for chain-scoped soft-delete. If `mail_chain` is always populated on a valid message, the fallback is dead code; if it isn't, the semantics drift slightly. Could you add a brief inline comment describing the case the fallback covers?

portal/messaging/messages.php
```diff
@@ -246,13 +246,18 @@ function getAuthPortalUsers()
...
+                    $scope.selected = null;
```
kojiromike, line 260: This looks redundant — the next line calls `$scope.init()`, which calls `isInboxSelected()` → `search()` → `groupToPages()` (line 208), and `groupToPages()` already sets `$scope.selected = null`. Is this an intentional optimistic UI nudge before the async `init()` reload settles, or can it be dropped?

kojiromike (commented):
Thanks for tackling this — the filtered-index vs. `allItems` mismatch is real and ID-based lookup is the right direction.

A few things to address alongside Copilot's points on `init()` behavior and silent-return handling:

**"Both the files"** — the PR description says this fixes "both the files" but only `portal/messaging/messages.php` is modified. Which other file did you intend? (The second `deleteItem(items.indexOf(selected))` at line 678 is inside an HTML comment block, so it's dead code — is there a similar bug in another file like `interface/main/messages/` that was meant to be included?)

**Commit message format.** The commit `Solving the indexing bug of deleting messages` doesn't match Conventional Commits and will fail the CI check. Please reword to match the PR title, e.g. `fix(messages): resolve index mismatch in deleteItem using ID-based matching`.

**AI assistance question.** The `#### Was an AI assistant used? Yes / No` line in the PR description is unanswered. If AI helped write any of this, please add an `Assisted-by` trailer to the commit per the project's CLAUDE.md.

portal/messaging/messages.php
Bakul2006: Yes, this was added to provide an immediate UI update before the async init() flow completes. Happy to remove it if you prefer keeping the logic centralized in init().

portal/messaging/messages.php
Bakul2006: The fallback to `id` was added as a safety measure in case mail_chain is not present on certain messages. e.g. Legacy Data

kojiromike:

One CI failure on this PR — verify whether it's real or a flake before merging/backporting.

Bakul2006:
@kojiromike I checked the failing codecov/patch.
portal/messaging/messages.php
The uncovered lines are from frontend (AngularJS) code in messages.php, 
which isn't currently covered by tests in this repository.So this appears to be expected rather than a regression.

PR #11678, opened 2026-04-16
fix(background-services): recover stuck leases from crashed workers

### LINKED ISSUE #11661: Background services have no recovery path for stuck locks
# Background services have no recovery path for stuck locks

`BackgroundServiceRunner` uses a row-level `running = 1` flag on `background_services` as its lock. The flag is set atomically in a single UPDATE when a worker acquires the service, and cleared in `safeReleaseLock()` when the worker finishes — including via a `register_shutdown_function` handler for abnormal exits.

Shutdown handlers do not fire on SIGKILL, OOM kill, container restart, host crash, or DB connection loss before the release can run. When any of those happen mid-execution, the `running = 1` flag persists indefinitely. Every subsequent tick (cron, AJAX, REST) returns `already_running` and skips the service until a human runs `UPDATE background_services SET running = 0 WHERE name = ?` directly against the database.

The current implementation acknowledges this in `safeReleaseLock()`'s docblock — "the stuck lock must be cleared manually or by a separate lock-recovery mechanism" — but no such mechanism exists.

## Why this matters

For a system intended to run unattended, a single OOM event or container restart can permanently disable a background service. There's no alerting for stuck services and no operator-facing tooling to clear one, so failures tend to be noticed only after someone reports a downstream symptom (missing emails, uncreated UUIDs, unreconciled claims).

## Not a viable alternative: MySQL `GET_LOCK()`

Worth calling out so it doesn't get proposed in review: MySQL's named-lock primitive (`GET_LOCK()` / `RELEASE_LOCK()`) is instance-scoped, not database-scoped. On installs where multiple OpenEMR sites share a MySQL server, two sites' `Email_Service` locks would serialize against each other. Namespacing lock names per site is possible but hits the 64-char limit and adds connection-pool footguns. Stick with the row-level flag and fix crash recovery on top of it.

kojiromike:
## Summary

Fixes #11661 — a single OOM event, SIGKILL, or container restart can permanently disable an OpenEMR background service (Email_Service, UUID_Service, phimail, etc.) because the row-level `running = 1` flag never gets cleared when the worker dies before its shutdown handler runs.

Replace the boolean flag with a timestamp-based lease:

- New column `background_services.lock_expires_at` (nullable `DATETIME`). Values are written via MySQL `NOW() + INTERVAL ? MINUTE` under whatever session timezone applies (OpenEMR syncs the MySQL session timezone to `gbl_time_zone` in `interface/globals.php`), so this is **not** guaranteed to be UTC. Liveness comparisons are self-consistent because every read and write uses `NOW()` in the same session.
- Acquire atomically *either* takes a free lock *or* steals an expired lease in a single UPDATE — no race, no separate sweeper cron needed.
- Release clears the lease (and the legacy `running` flag, which is kept for REST API compatibility but is now derived from the lease on read).
- When an expired lease is stolen, a PSR-3 warning is logged so operators get a signal instead of silent self-healing (crashed workers almost always mean an OOM kill worth investigating).
- New `bin/console background:services unlock --name=X` replaces the documented "run UPDATE directly against the database" workaround for the rare case of a genuinely hung service that shouldn't wait out its lease.

Lease duration is computed per-acquire as `max(60, 2×execute_interval)` capped at 24 hours. This gives short-interval services a reasonable recovery window without letting a misconfigured interval make a stuck lock unrecoverable.

## Why not `GET_LOCK()`?

Called out in the issue, but to summarize: MySQL's named-lock primitive is instance-scoped, not database-scoped. On installs where multiple OpenEMR sites share a MySQL server, two sites' `Email_Service` locks would serialize against each other. Namespacing hits the 64-char limit and adds connection-pool footguns.

## Behavior changes

- Admin report at `interface/reports/background_services.php` now renders "Busy" based on the live lease, so stuck locks from crashed workers (which used to report as "Yes" forever) now correctly show "No".
- The in-memory `running == 1` short-circuit in `BackgroundServiceRunner::run()` is removed — it would have bypassed the expired-lease recovery path. `acquireLock()` is now authoritative.

## Migration

Adds one column via `sql/8_1_0-to-8_1_1_upgrade.sql` using the existing `#IfMissingColumn` block. No data migration needed: existing rows start with `lock_expires_at = NULL`, which the new code treats as "free to acquire". An in-flight worker at upgrade time will have `running = 1` but no lease, so the first new-code acquire could theoretically race with it; this is a one-shot upgrade concern only.

## Test plan

- [x] `composer phpunit-isolated -- --filter BackgroundService` — 32 tests pass
- [x] `prek run --files <changed>` — phpcs, phpcbf, phpstan (level 10), rector, composer-require-checker all pass
- [ ] CI: full isolated + services integration suite, including the new `BackgroundServiceRunnerIntegrationTest` which exercises the real SQL path (acquire on fresh row, live-lease blocks, expired-lease steal, release clears timestamp, not-due path)
- [ ] Manual: verify the admin report renders "Busy: No" for a row with `running = 1, lock_expires_at = NULL` (simulates a stuck lock from the old scheme after upgrade)
- [ ] Manual: `bin/console background:services unlock --name=Email_Service` clears the lease

### DIFF
--- a/interface/reports/background_services.php
+++ b/interface/reports/background_services.php
@@ -122,8 +122,15 @@ if (!AclMain::aclCheckCore('admin', 'super')) {
  <tbody>  <!-- added for better print-ability -->
 <?php
 
-$res = sqlStatement("SELECT *, (`next_run` - INTERVAL `execute_interval` MINUTE) as `last_run_start`" .
-  " FROM `background_services` ORDER BY `sort_order`");
+// `is_busy` reflects the live lease rather than the legacy `running` flag
+// so stuck locks from crashed workers render as not-busy (see GH #11661).
+$res = sqlStatement(
+    "SELECT *,"
+    . " (`next_run` - INTERVAL `execute_interval` MINUTE) AS `last_run_start`,"
+    . " (`lock_expires_at` IS NOT NULL AND `lock_expires_at` > NOW()) AS `is_busy`"
+    . " FROM `background_services`"
+    . " ORDER BY `sort_order`"
+);
 while ($row = sqlFetchArray($res)) {
     ?>
   <tr>
@@ -143,7 +150,7 @@ while ($row = sqlFetchArray($res)) {
           <td align='center'><?php echo xlt('Not Applicable'); ?></td>
         <?php } ?>
 
-          <td align='center'><?php echo ($row['running'] > 0) ? xlt("Yes") : xlt("No"); ?></td>
+          <td align='center'><?php echo ($row['is_busy']) ? xlt("Yes") : xlt("No"); ?></td>
 
         <?php if ($row['running'] > -1) { ?>
           <td align='center'><?php echo text(DateFormatterUtils::oeFormatDateTime($row['last_run_start'], "global", true)); ?></td>
--- a/interface/reports/background_services.php
+++ b/interface/reports/background_services.php
@@ -14,6 +14,7 @@ require_once("../globals.php");
 
 use OpenEMR\Common\Acl\AccessDeniedHelper;
 use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Core\Header;
 use OpenEMR\Services\Utils\DateFormatterUtils;
 
@@ -124,14 +125,14 @@ if (!AclMain::aclCheckCore('admin', 'super')) {
 
 // `is_busy` reflects the live lease rather than the legacy `running` flag
 // so stuck locks from crashed workers render as not-busy (see GH #11661).
-$res = sqlStatement(
-    "SELECT *,"
-    . " (`next_run` - INTERVAL `execute_interval` MINUTE) AS `last_run_start`,"
-    . " (`lock_expires_at` IS NOT NULL AND `lock_expires_at` > NOW()) AS `is_busy`"
-    . " FROM `background_services`"
-    . " ORDER BY `sort_order`"
-);
-while ($row = sqlFetchArray($res)) {
+$rows = QueryUtils::fetchRecordsNoLog(<<<'SQL'
+    SELECT *,
+           (`next_run` - INTERVAL `execute_interval` MINUTE) AS `last_run_start`,
+           (`lock_expires_at` IS NOT NULL AND `lock_expires_at` > NOW()) AS `is_busy`
+      FROM `background_services`
+     ORDER BY `sort_order`
+    SQL, []);
+foreach ($rows as $row) {
     ?>
   <tr>
       <td align='center'><?php echo xlt($row['title']); ?></td>
@@ -172,7 +173,7 @@ while ($row = sqlFetchArray($res)) {
 
   </tr>
     <?php
-} // $row = sqlFetchArray($res) while
+} // foreach $rows
 ?>
 </tbody>
 </table>
--- a/sql/8_1_0-to-8_1_1_upgrade.sql
+++ b/sql/8_1_0-to-8_1_1_upgrade.sql
@@ -120,5 +120,5 @@ ALTER TABLE `form_eye_antseg`
 
 #IfMissingColumn background_services lock_expires_at
 ALTER TABLE `background_services`
-  ADD COLUMN `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration (UTC). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.';
+  ADD COLUMN `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration. Compared with NOW() on acquire, so the stored value uses whatever session timezone is in effect (OpenEMR syncs it to gbl_time_zone). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.';
 #EndIf
--- a/sql/8_1_0-to-8_1_1_upgrade.sql
+++ b/sql/8_1_0-to-8_1_1_upgrade.sql
@@ -117,3 +117,8 @@
 ALTER TABLE `form_eye_antseg`
   MODIFY COLUMN OSCONJ text;
 #EndIf
+
+#IfMissingColumn background_services lock_expires_at
+ALTER TABLE `background_services`
+  ADD COLUMN `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration (UTC). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.';
+#EndIf
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -196,7 +196,7 @@ CREATE TABLE `background_services` (
   `function` varchar(127) NOT NULL COMMENT 'name of background service function',
   `require_once` varchar(255) default NULL COMMENT 'include file (if necessary)',
   `sort_order` int(11) NOT NULL default '100' COMMENT 'lower numbers will be run first',
-  `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration (UTC). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.',
+  `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration. Compared with NOW() on acquire, so the stored value uses whatever session timezone is in effect (OpenEMR syncs it to gbl_time_zone). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.',
   PRIMARY KEY  (`name`)
 ) ENGINE=InnoDB;
 
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -196,6 +196,7 @@ CREATE TABLE `background_services` (
   `function` varchar(127) NOT NULL COMMENT 'name of background service function',
   `require_once` varchar(255) default NULL COMMENT 'include file (if necessary)',
   `sort_order` int(11) NOT NULL default '100' COMMENT 'lower numbers will be run first',
+  `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration (UTC). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.',
   PRIMARY KEY  (`name`)
 ) ENGINE=InnoDB;
 
--- a/src/Common/Command/BackgroundServicesCommand.php
+++ b/src/Common/Command/BackgroundServicesCommand.php
@@ -40,11 +40,11 @@ class BackgroundServicesCommand extends Command implements IGlobalsAware
     {
         $this
             ->setName('background:services')
-            ->setDescription('List, run, or generate crontab entries for background services')
+            ->setDescription('List, run, unlock, or generate crontab entries for background services')
             ->setDefinition(
                 new InputDefinition([
-                    new InputArgument('action', InputArgument::REQUIRED, 'Action to perform: list, run, or crontab'),
-                    new InputOption('name', null, InputOption::VALUE_REQUIRED, 'Service name (required for "run")'),
+                    new InputArgument('action', InputArgument::REQUIRED, 'Action to perform: list, run, unlock, or crontab'),
+                    new InputOption('name', null, InputOption::VALUE_REQUIRED, 'Service name (required for "run" and "unlock")'),
                     new InputOption('force', 'f', InputOption::VALUE_NONE, 'Bypass interval check (for "run")'),
                     new InputOption('php', null, InputOption::VALUE_REQUIRED, 'PHP binary path (for "crontab")', PHP_BINARY),
                 ])
@@ -64,6 +64,7 @@ class BackgroundServicesCommand extends Command implements IGlobalsAware
         return match ($action) {
             'list' => $this->handleList($io),
             'run' => $this->handleRun($input, $io),
+            'unlock' => $this->handleUnlock($input, $io),
             'crontab' => $this->handleCrontab($input, $io),
             default => $this->handleUnknownAction($action, $io),
         };
@@ -170,10 +171,60 @@ class BackgroundServicesCommand extends Command implements IGlobalsAware
         return Command::SUCCESS;
     }
 
+    /**
+     * Clear a service's lease unconditionally.
+     *
+     * Normally not needed — crashed workers' leases expire and are stolen
+     * automatically on the next tick. Use this when a service is genuinely
+     * hung mid-run and an operator has verified it is not making progress
+     * but doesn't want to wait out the lease (see GH #11661).
+     */
+    private function handleUnlock(InputInterface $input, SymfonyStyle $io): int
+    {
+        $name = $input->getOption('name');
+        if (!is_string($name) || $name === '') {
+            $io->error('The --name option is required for the "unlock" action.');
+            return Command::FAILURE;
+        }
+
+        $affected = $this->clearLease($name);
+
+        if ($affected === 0) {
+            $io->error("Service '{$name}' not found.");
+            return Command::FAILURE;
+        }
+
+        $io->success("Lease cleared for service '{$name}'.");
+        return Command::SUCCESS;
+    }
+
     private function handleUnknownAction(string $action, SymfonyStyle $io): int
     {
-        $io->error("Unknown action '{$action}'. Valid actions: list, run, crontab");
+        $io->error("Unknown action '{$action}'. Valid actions: list, run, unlock, crontab");
         return Command::FAILURE;
+    }
+
+    /**
+     * Clear the lease (and legacy running flag) for a service.
+     *
+     * Returns the number of rows affected (0 when the service does not
+     * exist, 1 when the lease was cleared).
+     */
+    protected function clearLease(string $name): int
+    {
+        QueryUtils::sqlStatementThrowException(
+            'UPDATE `background_services` SET `running` = 0, `lock_expires_at` = NULL WHERE `name` = ?',
+            [$name],
+            true,
+        );
+        // affectedRows() on a no-op UPDATE (e.g. already null/0) returns 0,
+        // which we want: it distinguishes a missing service from a live clear.
+        // Use a separate existence check so "already clear" still reports success.
+        $exists = QueryUtils::fetchRecordsNoLog(
+            'SELECT 1 FROM `background_services` WHERE `name` = ? LIMIT 1',
+            [$name],
+        );
+        return $exists === [] ? 0 : 1;
     }
 
     /**
--- a/src/Common/Command/BackgroundServicesCommand.php
+++ b/src/Common/Command/BackgroundServicesCommand.php
@@ -205,10 +205,12 @@ class BackgroundServicesCommand extends Command implements IGlobalsAware
     }
 
     /**
-     * Clear the lease (and legacy running flag) for a service.
+     * Clear the lease (and legacy running flag) for a service. The UPDATE
+     * is executed unconditionally; callers that need to distinguish "already
+     * clear" from "actively cleared" should check lock state beforehand.
      *
-     * Returns the number of rows affected (0 when the service does not
-     * exist, 1 when the lease was cleared).
+     * Returns 1 when the service exists (lease is now clear either way) or
+     * 0 when no service with that name exists.
      */
     protected function clearLease(string $name): int
     {
--- a/src/Common/Command/BackgroundServicesCommand.php
+++ b/src/Common/Command/BackgroundServicesCommand.php
@@ -17,7 +17,7 @@ declare(strict_types=1);
 namespace OpenEMR\Common\Command;
 
 use OpenEMR\Common\Database\QueryUtils;
-use OpenEMR\Common\Database\TableTypes;
+use OpenEMR\Services\Background\BackgroundServiceDefinition;
 use OpenEMR\Services\Background\BackgroundServiceRunner;
 use OpenEMR\Services\IGlobalsAware;
 use OpenEMR\Services\Trait\GlobalInterfaceTrait;
@@ -30,7 +30,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
- * @phpstan-import-type BackgroundServicesRow from TableTypes
+ * @phpstan-import-type BackgroundServicesQueryRow from BackgroundServiceDefinition
  */
 class BackgroundServicesCommand extends Command implements IGlobalsAware
 {
@@ -85,7 +85,10 @@ class BackgroundServicesCommand extends Command implements IGlobalsAware
                 $s['name'],
                 $s['title'],
                 (int) $s['active'] !== 0 ? 'yes' : 'no',
-                (int) $s['running'] === 1 ? 'yes' : 'no',
+                // Prefer the SQL-computed liveness flag over the legacy
+                // `running` column so stuck locks from crashed workers
+                // don't display as "yes" indefinitely.
+                (int) ($s['lease_is_live'] ?? $s['running']) === 1 ? 'yes' : 'no',
                 $this->formatInterval((int) $s['execute_interval']),
                 $s['next_run'],
             ], $services),
@@ -230,25 +233,34 @@ class BackgroundServicesCommand extends Command implements IGlobalsAware
     }
 
     /**
-     * @return list<BackgroundServicesRow>
+     * SQL-computed liveness flag (`lease_is_live`) is selected alongside
+     * the row so display/listing uses the same clock as acquireLock().
+     */
+    private const SELECT_WITH_LEASE_LIVE =
+        'SELECT background_services.*,'
+        . ' (lock_expires_at IS NOT NULL AND lock_expires_at > NOW()) AS lease_is_live'
+        . ' FROM background_services';
+
+    /**
+     * @return list<BackgroundServicesQueryRow>
      */
     protected function fetchServices(): array
     {
-        /** @var list<BackgroundServicesRow> */
+        /** @var list<BackgroundServicesQueryRow> */
         return QueryUtils::fetchRecordsNoLog(
-            'SELECT * FROM background_services ORDER BY sort_order',
+            self::SELECT_WITH_LEASE_LIVE . ' ORDER BY sort_order',
             [],
         );
     }
 
     /**
-     * @return list<BackgroundServicesRow>
+     * @return list<BackgroundServicesQueryRow>
      */
     protected function fetchActiveServices(): array
     {
-        /** @var list<BackgroundServicesRow> */
+        /** @var list<BackgroundServicesQueryRow> */
         return QueryUtils::fetchRecordsNoLog(
-            'SELECT * FROM background_services WHERE active = 1 AND execute_interval > 0 ORDER BY sort_order',
+            self::SELECT_WITH_LEASE_LIVE . ' WHERE active = 1 AND execute_interval > 0 ORDER BY sort_order',
             [],
         );
     }
--- a/src/Common/Command/BackgroundServicesCommand.php
+++ b/src/Common/Command/BackgroundServicesCommand.php
@@ -222,9 +222,11 @@ class BackgroundServicesCommand extends Command implements IGlobalsAware
             [$name],
             true,
         );
-        // affectedRows() on a no-op UPDATE (e.g. already null/0) returns 0,
-        // which we want: it distinguishes a missing service from a live clear.
-        // Use a separate existence check so "already clear" still reports success.
+        // The UPDATE runs unconditionally; its affected-row count would
+        // report 0 for "already clear" and 1 for "actively cleared",
+        // which is a distinction the caller doesn't care about. A
+        // dedicated existence check lets us report "not found" vs
+        // "service exists (lease is clear either way)".
         $exists = QueryUtils::fetchRecordsNoLog(
             'SELECT 1 FROM `background_services` WHERE `name` = ? LIMIT 1',
             [$name],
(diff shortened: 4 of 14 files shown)


src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,133 @@ protected function getServices(?string $serviceName, bool $force): array
...
-        if ($row === false) {
+        // Distinguish why we failed: lease still live, or just not yet due?
+        $currentExpiry = $this->readLeaseExpiry($service['name']);
+        if ($currentExpiry === false) {
             return 'error';
         }
-
-        if (($row['running'] ?? '0') === '1') {
+        if ($currentExpiry !== null) {
             return 'already_running';
```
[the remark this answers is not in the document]
kojiromike: Good catch, this was a real bug. Fixed in c109b0c — the failure-reason branch now checks `lock_expires_at > NOW()` in SQL, so an expired timestamp can't mask `not_due`. Added a regression test (`testAcquireReportsNotDueEvenWhenExpiredLeasePresent`).

sql/database.sql
```diff
@@ -196,6 +196,7 @@ CREATE TABLE `background_services` (
   `function` varchar(127) NOT NULL COMMENT 'name of background service function',
   `require_once` varchar(255) default NULL COMMENT 'include file (if necessary)',
   `sort_order` int(11) NOT NULL default '100' COMMENT 'lower numbers will be run first',
+  `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration (UTC). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.',
```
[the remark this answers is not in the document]
kojiromike: You're right about the timezone. `interface/globals.php:525` sets `SET time_zone = ?` to the PHP-configured timezone (`gbl_time_zone`), so `NOW()` returns session-local time, not UTC. Updated the comment in c109b0c to drop the UTC claim. No correctness issue — all lease comparisons use `NOW()` under the same session, so they're self-consistent whatever timezone applies.

sql/8_1_0-to-8_1_1_upgrade.sql
```diff
@@ -117,3 +117,8 @@
 ALTER TABLE `form_eye_antseg`
   MODIFY COLUMN OSCONJ text;
 #EndIf
+
+#IfMissingColumn background_services lock_expires_at
+ALTER TABLE `background_services`
+  ADD COLUMN `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration (UTC). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.';
```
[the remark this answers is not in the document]
kojiromike: Same fix — comment corrected in c109b0c. The column isn't UTC (OpenEMR syncs the MySQL session time_zone to `gbl_time_zone` in `interface/globals.php`), but since every read and write uses `NOW()` under the same session, the lease comparison remains self-consistent.

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,136 @@ protected function getServices(?string $serviceName, bool $force): array
...
+     * at most MAX_LEASE_MINUTES, otherwise 2x the execution interval so that
+     * a normal run comfortably fits inside its lease.
+     *
+     * @param BackgroundServicesRow $service
+     */
+    private function computeLeaseMinutes(array $service): int
+    {
+        $proposed = max(self::MIN_LEASE_MINUTES, ((int) $service['execute_interval']) * 2);
+        return min($proposed, self::MAX_LEASE_MINUTES);
+    }
```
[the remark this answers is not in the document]
kojiromike: Valid concern, filed as #11683 for follow-up. Scoping it there rather than expanding this PR: a renewal/heartbeat mechanism (or per-service `max_run_minutes`) is a meaningful design change and no current OpenEMR background service runs anywhere near even the 60-minute floor — phimail check, UUID generation, and Email_Service all complete in seconds. The current PR's goal is recovery from crashes; making long-running services safe from concurrency is a separate hardening concern worth its own scope.

src/Services/Background/BackgroundServiceDefinition.php
```diff
@@ -68,6 +74,22 @@ public function toArray(): array
...
+     * we can't interpret as held).
+     */
+    private static function leaseIsLive(?string $lockExpiresAt): bool
+    {
+        if ($lockExpiresAt === null || $lockExpiresAt === '') {
+            return false;
+        }
+        $expiry = strtotime($lockExpiresAt);
+        return $expiry !== false && $expiry > time();
+    }
```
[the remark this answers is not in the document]
kojiromike: Fixed in 9d89260. `BackgroundServiceRegistry::get()` and `list()` now SELECT a `lease_is_live` column computed in SQL as `lock_expires_at > NOW()`, so liveness uses the same clock as `acquireLock()`. `fromDatabaseRow()` prefers that flag when present and only falls back to the PHP-clock comparison for rows constructed by hand (test fixtures). Added a regression test for the SQL-vs-PHP disagreement case.

src/Common/Command/BackgroundServicesCommand.php
```diff
@@ -170,12 +171,62 @@ private function handleCrontab(InputInterface $input, SymfonyStyle $io): int
...
-        $io->error("Unknown action '{$action}'. Valid actions: list, run, crontab");
+        $io->error("Unknown action '{$action}'. Valid actions: list, run, unlock, crontab");
         return Command::FAILURE;
     }
 
+    /**
+     * Clear the lease (and legacy running flag) for a service.
+     *
+     * Returns the number of rows affected (0 when the service does not
+     * exist, 1 when the lease was cleared).
```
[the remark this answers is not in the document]
kojiromike: Docblock corrected in 9d89260 — the UPDATE is unconditional, so the return value distinguishes "service exists" from "service not found", not "cleared" vs "already clear". Kept the existing semantics because operators invoking `unlock` don't care about the distinction (either way the lease is clear afterward); the docblock now matches what the code actually does.

src/Common/Command/BackgroundServicesCommand.php
```diff
@@ -40,11 +40,11 @@ protected function configure(): void
...
             ->setName('background:services')
-            ->setDescription('List, run, or generate crontab entries for background services')
+            ->setDescription('List, run, unlock, or generate crontab entries for background services')
             ->setDefinition(
                 new InputDefinition([
-                    new InputArgument('action', InputArgument::REQUIRED, 'Action to perform: list, run, or crontab'),
-                    new InputOption('name', null, InputOption::VALUE_REQUIRED, 'Service name (required for "run")'),
+                    new InputArgument('action', InputArgument::REQUIRED, 'Action to perform: list, run, unlock, or crontab'),
+                    new InputOption('name', null, InputOption::VALUE_REQUIRED, 'Service name (required for "run" and "unlock")'),
                     new InputOption('force', 'f', InputOption::VALUE_NONE, 'Bypass interval check (for "run")'),
```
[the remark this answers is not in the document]
kojiromike: Fixed in a92dcd1. `fetchServices()` and `fetchActiveServices()` now SELECT the computed `lease_is_live` flag (`lock_expires_at > NOW()`), and `handleList()` uses it in the Running column — same source of truth as the admin report and REST API, so stuck locks from crashed workers render as "no".

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,136 @@ protected function getServices(?string $serviceName, bool $force): array
...
      *
      * @param BackgroundServicesRow $service
      * @return string|null Null on success, reason string on failure
      */
     protected function acquireLock(array $service, bool $force): ?string
     {
-        $sql = 'UPDATE background_services SET running = 1, next_run = NOW() + INTERVAL ?'
-            . ' MINUTE WHERE running < 1 ' . ($force ? '' : 'AND NOW() > next_run ') . 'AND name = ?';
+        $leaseMinutes = $this->computeLeaseMinutes($service);
+        $priorExpiry = $this->readLeaseExpiry($service['name']);
```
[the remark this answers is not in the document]
kojiromike: Good catch, fixed in a92dcd1. Using `$service['lock_expires_at']` directly from the row snapshot instead of re-reading. You're right that the prior-lease signal is inherently best-effort regardless of how it's captured (the atomic UPDATE is what arbitrates), so there's no reason to pay for an extra round-trip on every tick. Removed `readLeaseExpiry()`.

tests/Tests/Services/Background/BackgroundServiceRunnerIntegrationTest.php
```diff
@@ -0,0 +1,190 @@
...
+        $result = $this->runner->callAcquireLock($this->fetchRow(), force: true);
+
+        $this->assertSame('already_running', $result);
+    }
+
+    public function testAcquireStealsExpiredLease(): void
+    {
+        $this->insertService();
+        $this->setLeaseExpiry(date('Y-m-d H:i:s', time() - 600)); // 10 min past
+
```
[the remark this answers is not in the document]
kojiromike: Fixed in 567ecdb. Renamed the helper to `setLeaseExpiryMinutesFromNow(int $minutesFromNow)` and made it emit `NOW() + INTERVAL ? MINUTE` in SQL. The "stolen lease is live" assertion also reads via `lock_expires_at > NOW()` now rather than `strtotime(...) > time()`. One clock source for setup and production code.

tests/Tests/Services/Background/BackgroundServiceRunnerIntegrationTest.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    public function testAcquireReportsNotDueEvenWhenExpiredLeasePresent(): void
+    {
+        // Regression guard: a service with an expired (stale) lease AND
+        // a future next_run should report 'not_due' rather than
+        // 'already_running'. The stale timestamp is not a live lease.
+        $this->insertService(nextRun: date('Y-m-d H:i:s', time() + 600));
+        $this->setLeaseExpiry(date('Y-m-d H:i:s', time() - 600));
+
```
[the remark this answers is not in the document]
kojiromike: Same fix in 567ecdb. `insertService(int $nextRunMinutesFromNow = -100000)` now writes `next_run` as `NOW() + INTERVAL ? MINUTE`, sharing the DB clock with `acquireLock()`'s due-ness check. Default stays comfortably in the past so the common case (always due) still reads naturally.

src/Common/Command/BackgroundServicesCommand.php
```diff
@@ -170,32 +174,93 @@ private function handleCrontab(InputInterface $input, SymfonyStyle $io): int
...
+    protected function clearLease(string $name): int
+    {
+        QueryUtils::sqlStatementThrowException(
+            'UPDATE `background_services` SET `running` = 0, `lock_expires_at` = NULL WHERE `name` = ?',
+            [$name],
+            true,
+        );
+        // affectedRows() on a no-op UPDATE (e.g. already null/0) returns 0,
+        // which we want: it distinguishes a missing service from a live clear.
+        // Use a separate existence check so "already clear" still reports success.
```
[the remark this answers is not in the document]
kojiromike: You're right, the affectedRows reference was a leftover from an earlier draft. Rewrote the inline comment in 567ecdb to describe what the code actually does (unconditional UPDATE, separate existence check, why a dedicated check is used).

interface/reports/background_services.php
```diff
@@ -122,8 +122,15 @@
  <tbody>  <!-- added for better print-ability -->
 <?php
 
-$res = sqlStatement("SELECT *, (`next_run` - INTERVAL `execute_interval` MINUTE) as `last_run_start`" .
-  " FROM `background_services` ORDER BY `sort_order`");
+// `is_busy` reflects the live lease rather than the legacy `running` flag
+// so stuck locks from crashed workers render as not-busy (see GH #11661).
+$res = sqlStatement(
```
kojiromike: use QueryUtils.

interface/reports/background_services.php
```diff
@@ -122,8 +122,15 @@
...
+    "SELECT *,"
```
kojiromike: don't do concatenation here. Use a nowdoc.

src/Common/Command/BackgroundServicesCommand.php
```diff
@@ -170,32 +174,95 @@ private function handleCrontab(InputInterface $input, SymfonyStyle $io): int
...
+        );
+        return $exists === [] ? 0 : 1;
+    }
+
+    /**
+     * SQL-computed liveness flag (`lease_is_live`) is selected alongside
+     * the row so display/listing uses the same clock as acquireLock().
+     */
+    private const SELECT_WITH_LEASE_LIVE =
+        'SELECT background_services.*,'
```
kojiromike: don't concatenate literal strings. Use a nowdoc

src/Services/Background/BackgroundServiceDefinition.php
```diff
@@ -68,6 +94,43 @@ public function toArray(): array
...
+     * @param BackgroundServicesQueryRow $row
+     */
+    private static function resolveRunning(array $row): bool
+    {
+        $liveFlag = $row['lease_is_live'] ?? null;
+        if ($liveFlag !== null) {
+            return (int) $liveFlag === 1;
+        }
+        return self::leaseIsLiveByPhpClock($row['lock_expires_at']);
+    }
```
kojiromike:
```suggestion
    private static function resolveRunning(array $row): bool
    {
        $liveFlag = $row['lease_is_live'] ?? null;
        return ($liveFlag === null)
            ? self::leaseIsLiveByPhpClock($row['lock_expires_at'])
            : (int) $liveFlag === 1;
    }
```

src/Services/Background/BackgroundServiceRegistry.php
```diff
@@ -70,14 +70,24 @@ public function unregister(string $name): void
         );
     }
 
+    /**
+     * Projection that derives lease liveness in SQL so the "running"
+     * signal uses the same clock (and time_zone) as acquireLock(), rather
+     * than PHP's clock via `time()` + `strtotime()`.
+     */
+    private const SELECT_WITH_LEASE_LIVE =
+        'SELECT `background_services`.*,'
```
kojiromike: don't concat literal strings.a

src/Services/Background/BackgroundServiceRegistry.php
```diff
@@ -70,14 +70,24 @@ public function unregister(string $name): void
         );
     }
 
+    /**
+     * Projection that derives lease liveness in SQL so the "running"
+     * signal uses the same clock (and time_zone) as acquireLock(), rather
+     * than PHP's clock via `time()` + `strtotime()`.
+     */
+    private const SELECT_WITH_LEASE_LIVE =
```
kojiromike: is this different from the other one? Can it be reused?

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,120 @@ protected function getServices(?string $serviceName, bool $force): array
...
+        // arbitrates who wins. A dedicated pre-read would not be fully
+        // race-free either, and it costs an extra round-trip per tick.
+        $priorExpiry = $service['lock_expires_at'];
+
+        // Atomic acquire-or-steal. The UPDATE matches only when:
+        //   - no lease is held (lock_expires_at IS NULL), or
+        //   - the existing lease has expired (prior worker crashed)
+        // and, unless --force, the service is due (NOW() > next_run).
+        $sql = 'UPDATE background_services'
+            . ' SET running = 1,'
```
kojiromike: don't concat literal strings

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,120 @@ protected function getServices(?string $serviceName, bool $force): array
...
-        $row = QueryUtils::querySingleRow(
-            'SELECT running FROM background_services WHERE name = ?',
+        // Distinguish why the UPDATE failed: is a LIVE lease held, or
+        // is the lease expired/null and the service simply not yet due?
+        // Checking `lock_expires_at > NOW()` in SQL ensures an expired
+        // lease does not get misreported as 'already_running' when the
+        // true reason is 'not_due'.
+        $liveLease = QueryUtils::querySingleRow(
+            'SELECT 1 AS live FROM background_services'
+            . ' WHERE name = ? AND lock_expires_at > NOW()',
```
kojiromike: don't concat literal strings

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,120 @@ protected function getServices(?string $serviceName, bool $force): array
...
-     * cleared manually or by a separate lock-recovery mechanism.
+     * If the underlying DB update fails, the lease remains set; the next
+     * tick will see it expire naturally and recover — no manual operator
+     * intervention required.
      */
     private function safeReleaseLock(string $serviceName): void
     {
         try {
             $this->releaseLock($serviceName);
         } catch (\Throwable) {
```
kojiromike, line 319: be more precise about the nature of possible acceptable exceptions.

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,120 @@ protected function getServices(?string $serviceName, bool $force): array
...
-
-        if (($row['running'] ?? '0') === '1') {
+        if (is_array($liveLease)) {
             return 'already_running';
         }
-
         return $force ? 'already_running' : 'not_due';
     }
 
     protected function releaseLock(string $serviceName): void
```
kojiromike, line 277: add a docblock that describes the kinds of exceptions this can throw

tests/Tests/Isolated/Common/Command/BackgroundServicesCommandTest.php
```diff
@@ -242,4 +283,14 @@ protected function fetchActiveServices(): array
             fn(array $s) => (int) $s['active'] !== 0 && (int) $s['execute_interval'] > 0,
         ));
     }
+
+    protected function clearLease(string $name): int
```
kojiromike: is this _really_ int, or bool?

interface/reports/background_services.php
kojiromike: Fixed in c21591f. Switched to `QueryUtils::fetchRecordsNoLog()` + `foreach` loop. Two stale `sqlFetchArray` baseline entries dropped as a side effect of the regenerated baseline.

interface/reports/background_services.php
kojiromike: Fixed in c21591f — nowdoc for the multi-line SELECT.

src/Common/Command/BackgroundServicesCommand.php
kojiromike: Fixed in c21591f — the Command's `SELECT_WITH_LEASE_LIVE` was removed entirely and now references `BackgroundServiceRegistry::SELECT_WITH_LEASE_LIVE` (promoted to a public nowdoc const there) so there's no duplication.

src/Services/Background/BackgroundServiceDefinition.php
kojiromike: Superseded in c21591f — `resolveRunning()` and `leaseIsLiveByPhpClock()` are gone. All production callers now go through `BackgroundServiceRegistry::SELECT_WITH_LEASE_LIVE` which supplies `lease_is_live` from SQL (same clock as `acquireLock`), so there's no PHP-clock fallback left to wrap. `fromDatabaseRow()` is now a single-expression lookup on the SQL-computed flag.

src/Services/Background/BackgroundServiceRegistry.php
kojiromike: Fixed in c21591f — nowdoc.

src/Services/Background/BackgroundServiceRegistry.php
kojiromike: Consolidated in c21591f. Promoted `BackgroundServiceRegistry::SELECT_WITH_LEASE_LIVE` to a public const and the Command now references that single source of truth for the projection. The duplicate Command-local const is gone.

src/Services/Background/BackgroundServiceRunner.php
kojiromike: Fixed in c21591f — `acquireLock()`'s UPDATE is now a nowdoc, with the conditional `AND NOW() > next_run` appended as a genuinely-distinct-value concatenation (allowed by the convention).

src/Services/Background/BackgroundServiceRunner.php
kojiromike: Fixed in c21591f — nowdoc for the live-lease diagnostic SELECT.

src/Services/Background/BackgroundServiceRunner.php
kojiromike: Expanded the docblock in c21591f. `safeReleaseLock()` now enumerates what it's absorbing: `SqlQueryException` from transient DB errors (dropped connection, deadlock, lock-wait timeout) and `\Error` subclasses that can surface during a fatal-shutdown unwind. Notes that the `\Throwable` catch is intentional — shutdown handlers can't propagate — and that a lost release is recovery-safe because the lease expires on its own.

src/Services/Background/BackgroundServiceRunner.php
kojiromike: Added in c21591f — `releaseLock()` now has an `@throws \OpenEMR\Common\Database\SqlQueryException` with context on when it triggers (dead connection, missing table/column, deadlock abort) and a pointer to `safeReleaseLock()` for callers that need to survive cleanup failures.

tests/Tests/Isolated/Common/Command/BackgroundServicesCommandTest.php
kojiromike: You're right — it's a boolean ("service exists" vs "service not found"), not a count. Changed `clearLease()` in both the real Command and the stub to return `bool` in c21591f, and collapsed the call site in `handleUnlock` accordingly.

sql/database.sql
```diff
@@ -196,6 +196,7 @@ CREATE TABLE `background_services` (
   `function` varchar(127) NOT NULL COMMENT 'name of background service function',
   `require_once` varchar(255) default NULL COMMENT 'include file (if necessary)',
   `sort_order` int(11) NOT NULL default '100' COMMENT 'lower numbers will be run first',
+  `lock_expires_at` datetime DEFAULT NULL COMMENT 'Lease expiration. Compared with NOW() on acquire, so the stored value uses whatever session timezone is in effect (OpenEMR syncs it to gbl_time_zone). Set on acquire, cleared on release. Expired leases are automatically stolen by the next worker.',
```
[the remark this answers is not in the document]
kojiromike: Good catch — PR description was still claiming UTC. Rewrote it to match the schema comment: the column is a plain `DATETIME` written via MySQL `NOW() + INTERVAL ? MINUTE` under whatever session timezone applies (OpenEMR syncs the session time_zone to `gbl_time_zone` in `interface/globals.php`). Liveness comparisons are self-consistent because reads and writes share the same session, not because the value is UTC.

tests/Tests/Isolated/Common/Command/BackgroundServicesCommandTest.php
```diff
@@ -58,6 +59,8 @@ private static function makeService(
             'function' => 'test_fn',
             'require_once' => null,
             'sort_order' => '100',
+            'lock_expires_at' => $lockExpiresAt,
+            'lease_is_live' => $lockExpiresAt !== null && $running === 1 ? '1' : '0',
         ];
```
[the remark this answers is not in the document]
kojiromike: Good catch — fixed in 16c44b6. Added an explicit `leaseIsLive` parameter to the stub's `makeService`, independent of `running` and `lockExpiresAt`. Also added a regression guard test (`testRunningColumnShowsNoForStuckLegacyFlagWithoutLiveLease`) that constructs exactly the inconsistent state you called out: `running = 1`, expired `lock_expires_at`, `leaseIsLive = false` — and asserts the list renders "no" for it.

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,145 @@ protected function getServices(?string $serviceName, bool $force): array
...
+
+        // Atomic acquire-or-steal. The UPDATE matches only when:
+        //   - no lease is held (lock_expires_at IS NULL), or
+        //   - the existing lease has expired (prior worker crashed)
+        // and, unless --force, the service is due (NOW() > next_run).
+        $sql = <<<'SQL'
+            UPDATE background_services
+               SET running = 1,
+                   lock_expires_at = NOW() + INTERVAL ? MINUTE,
+                   next_run = NOW() + INTERVAL ? MINUTE
```
kojiromike, line 217: Is there any chance these two `NOW() + INTERVAL ? MINUTE` could have a different result?

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,145 @@ protected function getServices(?string $serviceName, bool $force): array
...
+             WHERE name = ?
+               AND (lock_expires_at IS NULL OR lock_expires_at < NOW())
+            SQL;
+        if (!$force) {
+            $sql .= "\n   AND NOW() > next_run";
```
kojiromike: what is the value in introducing whitespace and a newline this way?

src/Services/Background/BackgroundServiceRunner.php
```diff
@@ -145,71 +180,145 @@ protected function getServices(?string $serviceName, bool $force): array
...
+     * The catch is `\Throwable` intentionally — shutdown handlers run in
+     * contexts where even a TypeError must not propagate, and losing the
+     * release is not a correctness problem: the lease expires on its own
+     * and the next tick recovers it.
      */
     private function safeReleaseLock(string $serviceName): void
     {
         try {
             $this->releaseLock($serviceName);
         } catch (\Throwable) {
```
kojiromike, line 319: @Firehed FYI a likely valid catch Throwable here.

src/Common/Command/BackgroundServicesCommand.php
```diff
@@ -170,32 +175,84 @@ private function handleCrontab(InputInterface $input, SymfonyStyle $io): int
...
     /**
-     * @return list<BackgroundServicesRow>
+     * Clear the lease (and legacy running flag) for a service. The UPDATE
+     * is executed unconditionally; callers that need to distinguish "already
+     * clear" from "actively cleared" should check lock state beforehand.
+     *
+     * Returns true when the service exists (its lease is now clear either
+     * way), or false when no service with that name exists.
+     */
+    protected function clearLease(string $name): bool
```
kojiromike, line 217: Cover this with a test.

src/Common/Command/BackgroundServicesCommand.php
```diff
@@ -170,32 +175,84 @@ private function handleCrontab(InputInterface $input, SymfonyStyle $io): int
...
+            'SELECT 1 FROM `background_services` WHERE `name` = ? LIMIT 1',
+            [$name],
+        );
+        return $exists !== [];
+    }
+
+    /**
+     * @return list<BackgroundServicesQueryRow>
      */
     protected function fetchServices(): array
```
kojiromike, line 239: Cover this with a test

src/Common/Command/BackgroundServicesCommand.php
```diff
@@ -170,32 +175,84 @@ private function handleCrontab(InputInterface $input, SymfonyStyle $io): int
...
+            BackgroundServiceRegistry::SELECT_WITH_LEASE_LIVE . ' ORDER BY sort_order',
             [],
         );
     }
 
     /**
-     * @return list<BackgroundServicesRow>
+     * @return list<BackgroundServicesQueryRow>
      */
     protected function fetchActiveServices(): array
```
kojiromike, line 251: Cover this with a test

src/Services/Background/BackgroundServiceRunner.php
kojiromike: The two `NOW()` calls return the same value — MySQL evaluates `NOW()` once per statement, not once per expression. The *intervals* are deliberately different: `lock_expires_at` gets `leaseMinutes` (min 60, max 1440) while `next_run` gets `execute_interval` (e.g. 5). The lease intentionally outlives the scheduling interval so there's room for the service to run before the lease expires.

src/Services/Background/BackgroundServiceRunner.php
kojiromike: None — removed in 2260f02. Changed to `' AND NOW() > next_run'` (single-quoted, no escape sequence). Cosmetic SQL formatting inside a runtime-built string doesn't help readability.

src/Services/Background/BackgroundServiceRunner.php
kojiromike: Acknowledged — leaving this for @Firehed to see. The docblock on `safeReleaseLock` enumerates the expected exception types (`SqlQueryException` for transient DB errors, `\Error` subclasses from shutdown-path teardown) and explains why `\Throwable` is intentional here (shutdown handlers must not propagate).

src/Common/Command/BackgroundServicesCommand.php
kojiromike: Added `BackgroundServicesCommandIntegrationTest` in 2260f02 — covers `clearLease` with three cases: existing service → true, missing service → false, and verifying the UPDATE actually clears `lock_expires_at` + `running`.

src/Common/Command/BackgroundServicesCommand.php
kojiromike: Added `testFetchServicesComputesLeaseIsLive` in 2260f02 — inserts a test service, calls `fetchServices()`, verifies the row includes a falsy `lease_is_live` (no lease held).

src/Common/Command/BackgroundServicesCommand.php
kojiromike: Added `testFetchActiveServicesFiltersCorrectly` in 2260f02 — inserts active, inactive, and manual-mode services; verifies only the active interval-based service appears in the result.

PR #11688, opened 2026-04-16
fix(mailer): hardcode UTF-8 CharSet to avoid null from missing global

### LINKED ISSUE #11624: EmailTestServiceTest: direct send and templated queue tests report success but message never arrives in Mailpit
### Context

In PR #11603 two methods on `tests/Tests/E2e/EmailTestServiceTest.php` were skipped with `markTestSkipped`:

- `directSendDeliversEmail`
- `queueTemplatedInsertsAndDelivers`

Both pass the `assertTrue($results[0]->success)` check (the library reports the send / queue succeeded) but then time out in `waitForEmail(...)` — the message never appears in Mailpit within 30 s.

A near-identical sibling in the same class, `queueInsertsAndDelivers`, passes. The difference:

| Test | Path | Result |
|------|------|--------|
| `directSendDeliversEmail` | `new MyMailer(true)` → `->send()` (direct SMTP) | ✗ fail |
| `queueInsertsAndDelivers` | `emailServiceQueue()` → `emailServiceRun()` (plain queue flush) | ✓ pass |
| `queueTemplatedInsertsAndDelivers` | `emailServiceQueueTemplatedEmail()` → `emailServiceRun()` (templated queue flush) | ✗ fail |

All three tests trigger PHP 8.2 deprecation warnings from `vendor/phpmailer/phpmailer/src/PHPMailer.php`:

- `PHPMailer.php:1618` — `strtolower(null)`
- `PHPMailer.php:2571` — `strtolower(null)`
- `PHPMailer.php:3625` — `strlen(null)`

Suggests a null field is being passed somewhere (subject, from, charset, etc.) during these sends. Unverified hypothesis: the templated path renders an empty subject/body, and `MyMailer::send()` (direct) has a similar null somewhere in its setup that doesn't hit on the plain queue path.

### Diagnostic next step

The fastest way to learn what's actually happening is to dump Mailpit's state when `waitForEmail` times out. Patch `tests/Tests/E2e/Email/EmailTestingTrait.php::waitForEmail()` to log all messages (subjects, recipients, received timestamps) just before returning `null` on timeout. That tells us whether:

- the message arrived with a different / null subject → null-field bug, fix the null source
- no message arrived at all → silent send failure, look at `$mailer->ErrorInfo` / SMTP transcript

### Acceptance

- Root cause identified and fixed (the null-field source, or the silent-send-failure source, or both).
- The three PHPMailer null-deprecations no longer fire from the email code paths these tests cover.
- Both skipped tests re-enabled and passing.

### References

- PR #11603 introduces the skips.
- Skip sites: `tests/Tests/E2e/EmailTestServiceTest.php::directSendDeliversEmail` and `::queueTemplatedInsertsAndDelivers`.
- Email test service: `src/Services/Email/EmailTestService.php`.
- Mailpit wait helper: `tests/Tests/E2e/Email/EmailTestingTrait.php::waitForEmail`.

---

_Issue body authored by Claude Code on behalf of @firehed._

kojiromike:
## Summary

Fixes #11624. Re-enables the two E2E tests that PR #11603 skipped (`directSendDeliversEmail` and `queueTemplatedInsertsAndDelivers`).

## Root cause

`MyMailer::emailMethod()` read `$HTML_CHARSET` via the `global` keyword. That variable is set at the top of `interface/globals.php` — but PHPUnit loads its bootstrap via `include_once` inside `TextUI\Application::loadBootstrapScript()`, so the top-level `$HTML_CHARSET = "UTF-8"` assignment is scoped to that method and never escapes to true global scope. In test context `global $HTML_CHARSET` was null, so `$this->CharSet` ended up null.

A null `CharSet` caused two symptoms:

1. PHPMailer triggered `strlen(null)` / `strtolower(null)` deprecations on PHP 8.2+ (PHPMailer.php lines 1618, 2571, 3625 — the `$this->CharSet` accesses in `preSend()`, `wrapText()`, and `encodeHeader()`).
2. PHPMailer emitted encoded-word subject headers with an empty charset declaration, e.g. `=??Q?OpenEMR_Email_Test_=E2=80=94_Direct_Send?=`. Mailpit cannot decode a header with no declared charset, so the raw encoded form was stored. Q-encoding turns spaces into underscores, so `waitForEmail`'s `stripos($subject, "Direct Send")` failed to match. The `queueInsertsAndDelivers` test passed by accident — its subject substring "Queue" has no space and survived the transformation. "Direct Send" and "Queue Templated" did not.

This also explains the issue's observation that the library reports success but the message "never arrives": the message *does* arrive in Mailpit; the subject just arrives malformed and the test helper's substring match can't find it.

## Fix

Set `CharSet` directly to `PHPMailer::CHARSET_UTF8`. OpenEMR is hardcoded to UTF-8 everywhere (`ini_set('default_charset', 'utf-8')`, `mb_internal_encoding('UTF-8')`, `$HTML_CHARSET = "UTF-8"`), so there is nothing to look up.

Three PHPStan baseline entries are no longer needed after this change and have been regenerated out:

- `openemr.forbiddenGlobalKeyword` for `$HTML_CHARSET` in `postmaster.php` (the `global` is gone)
- `assign.propertyType` for `MyMailer::$CharSet (string) does not accept mixed` (the assigned value is now a `string` constant)
- `deadCode.unreachable` for `EmailTestServiceTest.php` (the `markTestSkipped()` calls that made the test bodies unreachable are gone)

## Follow-up commit

A second commit applies OCE conventions to the touched file: adds `declare(strict_types=1)` to `postmaster.php` and an OpenCoreEMR copyright entry. PHPStan level 10 passes cleanly with `strict_types` enabled, confirming no existing caller relies on type coercion into `MyMailer`'s methods.

## Test plan

- [x] `vendor/bin/phpunit --testsuite email --filter EmailTestServiceTest` — 5/5 pass, including the two previously skipped tests
- [x] Full `email` suite — 10 active tests pass, 11 pre-existing skips in `FaxSmsEmailTest` / `NotificationCronEmailTest` are unrelated
- [x] `composer phpstan` — clean on full codebase after regenerating baseline
- [x] Pre-commit hooks (phpcs, phpcbf, phpstan, rector, composer-require-checker, codespell) all pass

### DIFF
--- a/.phpstan/baseline/assign.propertyType.php
+++ b/.phpstan/baseline/assign.propertyType.php
@@ -367,11 +367,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../library/classes/TreeMenu.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Property MyMailer\\:\\:\\$CharSet \\(string\\) does not accept mixed\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../library/classes/postmaster.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Property MyMailer\\:\\:\\$Password \\(string\\) does not accept string\\|false\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../library/classes/postmaster.php',
--- a/.phpstan/baseline/deadCode.unreachable.php
+++ b/.phpstan/baseline/deadCode.unreachable.php
@@ -288,11 +288,6 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Unreachable statement \\- code above always terminates\\.$#',
-    'count' => 2,
-    'path' => __DIR__ . '/../../tests/Tests/E2e/EmailTestServiceTest.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Unreachable statement \\- code above always terminates\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../tests/Tests/E2e/FaxSmsEmailTest.php',
 ];
--- a/.phpstan/baseline/openemr.forbiddenGlobalKeyword.php
+++ b/.phpstan/baseline/openemr.forbiddenGlobalKeyword.php
@@ -1447,11 +1447,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../library/classes/Prescription.class.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Use of the "global" keyword is forbidden \\(\\$HTML_CHARSET\\)\\. Use dependency injection instead\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../library/classes/postmaster.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Use of the "global" keyword is forbidden \\(\\$code_types\\)\\. Use dependency injection instead\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../library/contraception_billing_scan.inc.php',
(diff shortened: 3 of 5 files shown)


Firehed:
Nice find. Also, ick.

Issue #11677, opened 2026-04-16
Background service CLI returns non-zero exit on no-op outcomes


kojiromike:
`bin/console background:services run --name=<service>` returns exit code `2` (`Command::INVALID`) when the runner reports `not_due`, `already_running`, or `skipped`, and exit code `1` for `not_found` / `error`. Only `executed` returns `0`.

See `src/Common/Command/BackgroundServicesCommand.php:120-125`:

```php
$status = $results[0]['status'] ?? 'not_found';
return match ($status) {
    'executed' => Command::SUCCESS,
    'skipped', 'already_running', 'not_due' => 2,
    default => Command::FAILURE,
};
```

## Why this matters

The three exit-2 statuses are not error conditions — they are normal outcomes of running a service on any schedule denser than, or overlapping with, the service's configured interval. Every generic process supervisor (Kubernetes Job/CronJob, `systemd` with `Restart=on-failure`, `cron` with `MAILTO`, any CI step) treats non-zero exit as failure. Under Kubernetes this produces `CrashLoopBackOff` on every tick where the DB-level interval check skips the run, even though the system is behaving exactly as designed.

Reproduced on `pelican-clinic` with a `*/2 * * * *` CronJob for `Email_Service`: the first tick executed successfully, subsequent ticks exited 2 with a `[WARNING] Service 'Email_Service' is not yet due to run (interval has not elapsed).` message and backed off.

## Proposed direction

Treat these three outcomes as success at the process-exit level. The `status` field and the `[WARNING]` output already convey the distinction for any caller that cares; the exit code should reflect whether the process ran cleanly, not whether it happened to have work to do on this tick.

If a tri-state signal is worth keeping for human/script consumption, it belongs in a `--strict` opt-in flag, not as the default.

## Scope

- `src/Common/Command/BackgroundServicesCommand.php::handleRun`
- `grep` for call sites that branch on exit code 2 specifically (likely none)

Parent: #11662

PR #11699, opened 2026-04-17
test(background-services): add shell-out CLI integration test

### LINKED ISSUE #11686: Add shell-out integration test for bin/console background:services
# Add shell-out integration test for `bin/console background:services`

Currently no test invokes the production cron path the way cron/Kubernetes CronJob actually do. Every `bin/console background:services` unit test uses `Symfony\Component\Console\Tester\CommandTester`, which constructs the command class in-process and bypasses everything between `cron -> /usr/bin/env php -> bin/console -> kernel boot -> command dispatch`. That's precisely the stack where the bugs flagged in this epic hide:

- #11677 / opencoreemr/chart-oce-openemr#114 — exit code on no-op outcomes (caught by unit test, but only because the mapping lives on the command class)
- #11672 — `$_SERVER['PHP_SELF']` missing under CLI (would have been caught by any test that actually shelled out)
- Future: a service function that assumes a web request context, a migration that breaks the Registry upsert, a `bin/console` regression from a Symfony upgrade

## What

Add `tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php` in the existing `services` suite (runs via `build_test services` in `ci/ciLibrary.source`, exercised across the full `ci/apache_*` matrix). The test:

1. Registers a probe service via `BackgroundServiceRegistry` (short interval, a function that flips a sentinel row or writes a temp file).
2. Shells out via `proc_open` to `bin/console background:services run --name=<probe>` (captures stdout, stderr, exit code).
3. Asserts exit 0, expected stdout marker, and that the probe function actually executed (sentinel row present / temp file written).
4. Second case: calls the same binary with no `--name`, asserts exit 0 and that the probe advanced.
5. Third case: calls immediately again (probe will be `not_due`); asserts exit 0 (the #11677 regression shape) and no second execution.
6. Tears down by `unregister()`.

Zero CI-config changes — `services` is already in the test matrix at `.github/workflows/test.yml` line 120, `ciLibrary.source`'s `_exec` runs inside the openemr container so `bin/console` picks up the same DB, config, and PHP environment as production.

## Why

This is the one test shape that proves the epic's deliverables actually work under the execution environment that the chart PR opencoreemr/chart-oce-openemr#114 deploys. Without it, we're fixing bugs in a codebase whose production orchestration path has no regression coverage at all.

## Notes

- `proc_open` is already used elsewhere in the repo (e.g., `tests/Tests/E2e/` for bookend scripts) so the pattern is familiar.
- Needs a small guard for `bin/console`'s path (use `__DIR__ . '/../../../../bin/console'` or similar — verify the services suite's working directory).
- Probe function should be defined in a test-only file that the probe service's `require_once` points at; keep it outside any autoloaded path so it can't leak into production.

kojiromike:
## Summary

Adds an integration test that shells out to `bin/console background:services run` via `Symfony\Component\Process`, proving the full production cron path (cron → php → bin/console → kernel boot → command dispatch) actually runs a registered service. Every other test in this area drives the command class in-process via `CommandTester` and short-circuits that chain.

The test registers a probe service whose function writes to a sentinel file in the temp dir, invokes the real binary, and asserts exit code, stdout marker, and sentinel contents.

Runs in the existing `services` test suite — no CI config changes.

## Status

Draft. Two of the three cases from #11686 are in:

- `testRunByNameExecutesProbeViaShell` — `run --name=<probe>` → exit 0, success marker, sentinel written exactly once.
- `testImmediateRerunIsNotDueAndDoesNotExecuteAgain` — re-run within the interval → exit 2 (current behavior), not-due warning, sentinel unchanged. Inline TODO flips this to exit 0 once #11677 lands.

Deferred:
- Case from the issue that calls `run` with no `--name`. `handleRun` currently errors when `--name` is missing — orthogonal to #11686, will file or bundle later.

Closes #11686 once the remaining case is decided.

## Test plan

- [ ] `docker compose exec openemr /root/devtools services-test` with the new test class
- [ ] Confirm probe sentinel path is stable inside the services container
- [ ] Decide whether to bundle the #11677 fix here or keep them separate

### DIFF
--- a/tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
+++ b/tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
@@ -91,14 +91,10 @@ class BackgroundServicesCliIntegrationTest extends TestCase
             $result['stdout'],
             'Executed run should print the success marker from handleRun().',
         );
-        $this->assertFileExists(
-            $this->sentinelPath,
-            'Probe function must have actually run in the child process.',
-        );
         $this->assertSame(
             1,
             $this->sentinelLineCount(),
-            'Probe function should have executed exactly once.',
+            'Probe function should have executed exactly once in the child process.',
         );
     }
 
@@ -188,13 +184,16 @@ class BackgroundServicesCliIntegrationTest extends TestCase
 
     private function sentinelLineCount(): int
     {
-        if (!is_file($this->sentinelPath)) {
-            return 0;
-        }
+        // Precondition asserted here (rather than defensively handled with a
+        // silent `return 0`) so a missing sentinel surfaces as a clear test
+        // failure — "subprocess didn't run the probe" — instead of an
+        // unhelpful "expected 1, got 0" one level up.
+        $this->assertFileExists(
+            $this->sentinelPath,
+            'Sentinel file must exist before counting probe executions.',
+        );
         $contents = file_get_contents($this->sentinelPath);
-        if ($contents === false || $contents === '') {
-            return 0;
-        }
+        $this->assertNotFalse($contents, 'Failed to read sentinel file.');
         return substr_count($contents, "\n");
     }
 }
--- a/tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
+++ b/tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
@@ -110,16 +110,15 @@ class BackgroundServicesCliIntegrationTest extends TestCase
 
         $second = $this->runConsole(['run', '--name=' . self::PROBE_NAME]);
 
-        // Current behavior: not_due is reported with Command::INVALID (2),
-        // which causes generic supervisors (Kubernetes CronJob, systemd)
-        // to treat a scheduled no-op as a crash. See #11677 — when that
-        // lands this assertion flips to 0. Exit 2 here guards against
-        // regressions in the opposite direction (e.g. not_due falling
-        // through to `default => Command::FAILURE` = 1).
+        // Per #11677 (fixed in #11687), not_due is a no-op, not an error:
+        // it must exit 0 so generic supervisors (Kubernetes CronJob, systemd)
+        // don't treat a scheduled no-op as a crash. This assertion guards
+        // against regressions in either direction — exit 1 (generic failure)
+        // or exit 2 (Command::INVALID, the pre-#11687 behavior).
         $this->assertSame(
-            2,
+            0,
             $second['exit'],
-            "Expected exit 2 for not_due.\nstdout:\n{$second['stdout']}\nstderr:\n{$second['stderr']}",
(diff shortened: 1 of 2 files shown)


tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
```diff
@@ -0,0 +1,200 @@
...
+    private function sentinelLineCount(): int
+    {
+        if (!is_file($this->sentinelPath)) {
+            return 0;
+        }
+        $contents = file_get_contents($this->sentinelPath);
+        if ($contents === false || $contents === '') {
+            return 0;
+        }
+        return substr_count($contents, "\n");
```
kojiromike:
```suggestion
        return substr_count($contents ?: '', "\n");
```

tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
```diff
@@ -0,0 +1,200 @@
...
+    {
+        if (is_file($this->sentinelPath)) {
+            unlink($this->sentinelPath);
+        }
+    }
+
+    private function sentinelLineCount(): int
+    {
+        if (!is_file($this->sentinelPath)) {
+            return 0;
```
kojiromike: This is not covered in testing. Is there any way to use type declarations to guarantee that by the time sentinelLineCount runs, $this->sentinelPath is a file? For example `cliProbeSentinelPath` should guarantee it returns a file.

tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
kojiromike: Took this one step further: dropped both defensive guards (`is_file` and `$contents === false || ''`) and replaced them with explicit `assertFileExists` / `assertNotFalse` calls inside `sentinelLineCount`. That turns the missing/unreadable-file case into a clear test failure instead of a silent `return 0` that would later show up as an unhelpful "expected 1, got 0". See 878d53c221.

tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php
kojiromike: PHP's type system can't express "this path is a file" — and the sentinel legitimately doesn't exist at construction time (it's created by the subprocess). Closest equivalent is an assertion-level precondition, which is what I went with: `sentinelLineCount` now calls `$this->assertFileExists(...)` before reading. That gives us a test-grade failure message ("Sentinel file must exist before counting probe executions") the first time the precondition is violated, rather than the old dead `return 0` that was never covered. See 878d53c221.

Also added `@codeCoverageIgnore` on `markCliProbeSentinel` with a comment explaining the cross-process measurement limitation — same rationale as the data-provider carve-out documented in `CLAUDE.md`.

PR #11700, opened 2026-04-17
fix(orm): fix multiple bugs in ORDataObject::_load_enum

Firehed:

Fixes multiple bugs in the `ORDataObject::_load_enum()` method that could cause undefined variable errors and incorrect caching behavior.

I came across issues that appeared to be suppressed working on #11550, surfaced only in PHP 8.4+ where _something_ was subtly different along the way. I still can't quite figure out what, though I'm reasonably sure there's a tweak to error handling and promotion getting surfaced.

Note: I'm working on a parallel branch to fix some of the type info in ORDataObject, so more phpstan stuff should get fixed there. I'm ignoring the temporary additions.

- **Undefined variable fix**: Return empty array when no matching enum column is found (previously `$enum` was undefined)
- **Cache respects `$blank` parameter**: Cache raw enum values and apply the `$blank` parameter on each retrieval (previously cached results ignored `$blank` on subsequent calls)
- **Condition order fix**: Check `empty($this->_table)` before accessing cache arrays
- **Prevent mutation**: Copy enum values instead of modifying `$col->enums` in place
- **Working cache**: Use a static class property for caching (the previous `OEGlobalsBag::get('static')` approach assigned to a return value, which was a no-op)

---

### DIFF
--- a/.phpstan/baseline/nullCoalesce.offset.php
+++ b/.phpstan/baseline/nullCoalesce.offset.php
@@ -162,6 +162,16 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../library/ajax/payment_ajax.php',
 ];
 $ignoreErrors[] = [
+    'message' => '#^Offset 0 on array\\<string, int\\> on left side of \\?\\? does not exist\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/classes/Document.class.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Offset 0 on array\\<string, int\\> on left side of \\?\\? does not exist\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../library/classes/X12Partner.class.php',
+];
+$ignoreErrors[] = [
     'message' => '#^Offset \'height\' on array\\{0\\: int\\<0, max\\>, 1\\: int\\<0, max\\>, 2\\: int, 3\\: string, mime\\: string, channels\\?\\: int, bits\\?\\: int\\} on left side of \\?\\? does not exist\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../library/classes/thumbnail/Thumbnail.class.php',
--- a/.phpstan/baseline/nullCoalesce.offset.php
+++ b/.phpstan/baseline/nullCoalesce.offset.php
@@ -162,16 +162,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../library/ajax/payment_ajax.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Offset 0 on array\\<string, int\\> on left side of \\?\\? does not exist\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../library/classes/Document.class.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Offset 0 on array\\<string, int\\> on left side of \\?\\? does not exist\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../library/classes/X12Partner.class.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Offset \'height\' on array\\{0\\: int\\<0, max\\>, 1\\: int\\<0, max\\>, 2\\: int, 3\\: string, mime\\: string, channels\\?\\: int, bits\\?\\: int\\} on left side of \\?\\? does not exist\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../library/classes/thumbnail/Thumbnail.class.php',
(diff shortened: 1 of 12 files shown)


src/Common/ORDataObject/ORDataObject.php
```diff
@@ -160,51 +167,53 @@ public function setThrowExceptionOnError($throwException)
...
+    protected function _load_enum(string $field_name, bool $blank = true): array
     {
-        if (
-            !empty(OEGlobalsBag::getInstance()->get('static')['enums'][$this->_table][$field_name])
-            && is_array(OEGlobalsBag::getInstance()->get('static')['enums'][$this->_table][$field_name])
-            && !empty($this->_table)
-        ) {
-            return OEGlobalsBag::getInstance()->get('static')['enums'][$this->_table][$field_name];
-        } else {
+        if (empty($this->_table)) {
```
kojiromike: Per the repo's `CLAUDE.md` ("Legacy Code Is Not the Standard"), new code shouldn't add `empty()` calls — `empty('0')` returning `true` is rarely what you want, and the baseline for `empty.notAllowed` in this file is something the codebase is trying to drive down. Would an explicit check work here?

```suggestion
        if ($this->_table === null || $this->_table === '') {
            return [];
        }
```

src/Common/ORDataObject/ORDataObject.php
```diff
@@ -160,51 +167,53 @@ public function setThrowExceptionOnError($throwException)
...
+            return [];
+        }
+
+        $cacheKey = $this->_table . '.' . $field_name;
```
kojiromike, line 183: The cache key is `table.field_name` only — it doesn't account for `$this->_db`. In a long-running PHP-FPM worker that serves multiple sites with potentially different enum definitions for the same table, this static would return stale entries on cross-site requests. The old `OEGlobalsBag::getInstance()` singleton had the same property when it (in theory) worked, so this isn't a regression introduced here — just worth confirming this matches the deployment model you care about. A `spl_object_id($this->_db)` (or similar) component in the cache key would close it if needed.

.phpstan/baseline/nullCoalesce.offset.php
```diff
@@ -161,6 +161,16 @@
     'count' => 1,
     'path' => __DIR__ . '/../../library/ajax/payment_ajax.php',
 ];
+$ignoreErrors[] = [
```
kojiromike: These two new baseline entries are the downstream effect of tightening `_load_enum`'s return to `array<string, int>`. Both call sites — `Document::__construct` (`$this->type_array[0] ?? ''`) and `X12Partner::__construct` (`$this->processing_format_array[0] ?? null`) — index with integer `0`, but `array_flip` produces string keys, so `[0]` has always returned `null` and the `??` defaults masked it. The PHPStan errors are pointing at real latent bugs, not noise.

The repo's `CLAUDE.md` says "Avoid baselines. Never add new baseline entries — fix the underlying type error." Could these two callers be fixed in this PR (or split into a focused follow-up that lands first), rather than baselined? The callers likely meant something like `array_key_first($this->type_array) ?? ''` to grab the first enum value — though it's worth verifying with whoever knows the original intent, since the `?? ''` may also have been a deliberate "no default" sentinel.

kojiromike (approved):
This looks good to me, but Claude did find a couple of opportunities for improvement…

- The narrower `@return array<string, int>` shape is great, but it surfaced two latent bugs in callers (`Document::__construct` and `X12Partner::__construct` both do `$this->type_array[0] ?? ''`). Their integer offset `[0]` never existed in the returned shape — `array_flip` produces string keys. The defaults (`?? ''`, `?? null`) silently masked it. The PR adds these to the PHPStan baseline; per the repo's `CLAUDE.md` ("Avoid baselines. Never add new baseline entries — fix the underlying type error"), it would be better to fix the call sites in this PR or split them out into a follow-up rather than suppress.
- `empty($this->_table)` — `CLAUDE.md` calls out `empty()` as a legacy antipattern that new code shouldn't pick up. Easy explicit replacement inline.
- The static cache is now correctly populated, but its key omits `$this->_db`. Under PHP-FPM this is fine for a single-database deployment; it's worth a quick sanity check for any multi-database setup where the same `table.field_name` could resolve to different enum sets across requests served by the same worker. (The previous OEGlobalsBag-singleton intent had the same property when it actually worked, so this isn't a regression — just flagging.)

No tests were added — given the bugs are PHP-version-dependent and the `$blank=false` cache-correctness fix is a real behavior change, a small unit/isolated test that exercises both `blank` paths against a stub `MetaColumns` would be a nice-to-have.

Behavior changes look correct and the PHPStan baseline shrinkage on this file is a real win.

src/Common/ORDataObject/ORDataObject.php
Firehed: (Assuming this was a claude comment) FPM workers do not share state across requests, this is not an issue outside of runtimes like Swoole, and OpenEMR has -1000000% chance of working inside of it.

.phpstan/baseline/nullCoalesce.offset.php
Firehed: The PHPStan issues are almost all fixed. I have a follow up branch with additional type-tightening to resolve more stuff like this.

PR #11701, opened 2026-04-17
refactor(ci): Simplify and speed up the integration test build matrix

Firehed:
Our CI test harness is _extremely_ complex, and has a huge amount of setup code. While it was likely required at the time, we've made significant refactoring headway that removes the need for quite a bit of this on many of the tests. 

Through some trial-and-error, I've found that there are five PHPUnit test suites that need a db connection & the tables, but no webserver or anything else. Despite this, those suites were running with all of the heavy initialization code. This change pulls them into their own CI workflow with a much smaller and lighter-weight setup process:

- Run a DB on the action runner directly
- Run the setup commands, pointing to the bare-metal[^bare] database instance
- Run PHPUnit directly on the action runner

[^bare]: It's still Docker, but it's no longer Docker-in-docker (-in-docker?)

For the most part, these tests run so much faster that they finish before the previous ones even start[^speed]. There should be no loss in coverage or accuracy, and the simplification ought to prevent situations we've encountered several times where bugs in the CI procedure suppressed legitimate test failures.

[^speed]: The "collect" step that sits in front the test all workflow takes nearly as long as running the entire new test suite; the "setup" step after it quite a bit longer. The limiting factor on the new path _by far_ is the DB setup process (about 2/3 of the runtime); I expect that to drop from ~45s to probably ~5s once we finish doctrine/migrations conversion.

There are two additional optimizations, since the tests finish in seconds and most of the time is spent on coverage collection:

1) Instead of matrixing-out the test _suites_, all of the suites are run at once in a single `phpunit` execition. This reduces the test coverage collection penalty from once per suite to once for all suites.
1) Test coverage only runs on one select matrix branches. This reduces the runtime of all but one matrix significantly.

Since the performance gain is so large, we're actually able to run the full test matrix on PR instead of just on master. While version-specific regressions are _extremely_ rare, we will know earlier.

1) Takes the prototype `integration tests` workflow and makes it useful:
- Sets up matrix for DB & PHP versions
- Run 5 test suites in PHPUnit

2) Removed the 5 test suites from the even-wider "test all configurations" build matrix, since they're now redundant.

3) Fixes one test setup issue revealed, which evidently came out of running the suites together.

DB suites use currently-supported LTS versions (https://endoflife.date/mariadb + https://endoflife.date/mysql) plus some "latest" variants.

No


.github/workflows/integration-tests.yml
```diff
@@ -16,24 +20,52 @@ permissions:
...
+            image: mysql:9
+            health-cmd: "mysql -uroot -proot -e 'SELECT 1'"
+          - name: MariaDB 10
+            image: mariadb:10
+            health-cmd: "healthcheck.sh --connect --innodb_initialized"
+          - name: MariaDB 11
+            image: mariadb:11
+            health-cmd: "healthcheck.sh --connect --innodb_initialized"
+          - name: MariaDB 12
+            image: mariadb:12
```
bradymiller, line 51: @Firehed ,
for the databases, goal is to run the supported LTS versions and the most recent dev one if there is one (there is not one for mysql right now btw) (also stuck with mysql 5.7 until we get ok from @jesdynf when this has been removed from the aws packages):
mariadb:
10.6 LTS
10.11 LTS
11.4 LTS
11.8 LTS
12.2 RR (rolling release)
mysql:
5.7
8.0 LTS
8.4 LTS
9.7 LTS

Will this then mean each of above are tested with each of below php versions?

Firehed: Should be an easy change to adjust to LTS versions! It wasn't obvious to me how the versions were picked for CI, makes sense now. 

And yes- each DB on each PHP version

bradymiller: wow, runs are super fast. after address above, then looks good to me.
nice stuff!

PR #11707, opened 2026-04-19
refactor(cli): replace allow_cronjobs.php with Symfony console commands

### LINKED ISSUE #11668: Deprecate `library/allow_cronjobs.php` in favor of Symfony console commands
# Deprecate `library/allow_cronjobs.php` in favor of Symfony console commands

`library/allow_cronjobs.php` is a bootstrap shim for standalone PHP scripts run from cron. It:

- Stuffs CLI args (`key=value`) into `$_REQUEST`/`$_GET`/`$_POST`.
- Sets `$ignoreAuth = true` and `$sessionAllowWrite = true`.
- Requires a `site=` argument.

Current callers: `interface/batchcom/batch_phone_notification.php`, `custom/zutil.cli.doc_import.php`.

It's also on the PHPStan `forbiddenRequestGlobals` baseline (three entries), and its arg convention differs from every other CLI surface OpenEMR ships (`bin/console` uses Symfony options; `background_services` cron uses positional args).

## What

1. Migrate the existing callers to Symfony console commands under `src/Common/Command/`, dispatched via `bin/console`.
2. Emit a deprecation notice from `allow_cronjobs.php` in the interim.
3. Remove the file once callers are gone.

## Why

- Eliminates the last `$_REQUEST`-stuffing bootstrap path.
- Unifies CLI tooling on a single entry point (`bin/console`) with consistent argument parsing, help output, and error handling.
- Clears three baseline entries that can never be reduced while the file exists.

kojiromike:
## Summary

Closes #11668.

Replaces `library/allow_cronjobs.php` and its two callers with two Symfony console commands dispatched via `bin/console`:

| Legacy script | Replacement command |
|---|---|
| `interface/batchcom/batch_phone_notification.php` | `php bin/console notifications:phone` |
| `custom/zutil.cli.doc_import.php` | `php bin/console documents:import [--pid --category --owner --limit --in-situ]` |

`library/allow_cronjobs.php` is removed, along with its `$_REQUEST`/`$_GET`/`$_POST`-stuffing bootstrap shim, ad-hoc `key=value` argument parsing, and three `forbiddenRequestGlobals` baseline entries that it held open.

## Behavior notes for operators

Operators who schedule these scripts from cron will need to update their crontabs. Old:

```
/usr/bin/php -f /var/www/html/openemr/interface/batchcom/batch_phone_notification.php site=default
/usr/bin/php -f /var/www/html/openemr/custom/zutil.cli.doc_import.php site=default
```

New:

```
/usr/bin/php /var/www/html/openemr/bin/console --site=default notifications:phone
/usr/bin/php /var/www/html/openemr/bin/console --site=default documents:import
```

## Implementation notes

- Commands live in `src/Common/Command/` and are auto-discovered by `SymfonyCommandRunner`.
- Options use Symfony `InputOption`; no superglobal access.
- Database access uses `QueryUtils` (no direct `sqlStatement`/`sqlQuery`).
- Explicit type narrowing throughout — no casts to silence phpstan.
- PHPStan level 10 passes with **zero new baseline entries** for the new files; the per-identifier baselines are regenerated and net shrink by ~310 entries (all deletions, no additions outside the regen churn).
- `Document`, `MaviqClient`, `RestResponse`, and `addNewDocument` are added to `.composer-require-checker.json`'s `symbol-whitelist`, matching the existing pattern for legacy `library/` symbols such as `C_Document`.

## Test plan

- [ ] `composer phpstan` passes (level 10, full codebase).
- [ ] `composer phpcs` / `composer rector-check` / `composer require-checker` pass.
- [ ] `php bin/console list` shows `notifications:phone` and `documents:import`.
- [ ] Manually invoke `php bin/console --site=<site> notifications:phone` on a site with phone gateway configured.
- [ ] Manually invoke `php bin/console --site=<site> documents:import --pid=<pid> --limit=<n>` against a populated scanner directory.


tests/Tests/Isolated/Common/Command/DocumentImportCommandTest.php
```diff
@@ -0,0 +1,406 @@
...
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Tester\CommandTester;
+
+#[Group('isolated')]
+#[Group('document-import')]
+class DocumentImportCommandTest extends TestCase
+{
+    public static function setUpBeforeClass(): void
```
kojiromike, line 31: This function isn't covered. If it can't be because it happens before coverage kicks in, then it should be properly excluded from coverage.

tests/Tests/Isolated/Common/Command/PhoneNotificationCommandTest.php
```diff
@@ -0,0 +1,414 @@
...
+use PHPUnit\Framework\TestCase;
+use RestResponse;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Tester\CommandTester;
+
+// MaviqClient and RestResponse are legacy globals defined procedurally; PHPUnit
+// resolves class-level `extends` clauses at file-load time, so load the file
+// before the stub class declarations below.
+if (!class_exists(MaviqClient::class, false)) {
```
kojiromike, line 30: This if block isn't covered by tests. If it can't be, it must be properly excluded from coverage analysis

tests/Tests/Isolated/Common/Command/PhoneNotificationCommandTest.php
```diff
@@ -0,0 +1,414 @@
...
+    require_once __DIR__ . '/../../../../../library/maviq_phone_api.php';
+}
+
+#[Group('isolated')]
+#[Group('phone-notifications')]
+class PhoneNotificationCommandTest extends TestCase
+{
+    public static function setUpBeforeClass(): void
```
kojiromike, line 42: This function isn't covered by tests. If it can't be, it should be properly excluded from analysis.

tests/Tests/Isolated/Common/Command/PhoneNotificationCommandTest.php
```diff
@@ -0,0 +1,414 @@
...
+     * @param mixed $path
+     * @param mixed $method
+     * @param mixed $vars
+     */
+    public function sendRequest($path, $method = 'POST', $vars = []): RestResponse
+    {
+        $index = $this->calls;
+        $this->calls++;
+        if (!isset($this->responses[$index])) {
+            throw new \RuntimeException('FakeMaviqClient ran out of canned responses');
```
kojiromike, line 490: This line is not covered by tests. I think this one probably should be.

src/Common/Command/PhoneNotificationCommand.php
```diff
@@ -0,0 +1,335 @@
...
+     */
+    private function intField(array $row, string $key): int
+    {
+        $value = $row[$key] ?? null;
+        if (is_int($value)) {
+            return $value;
+        }
+        if (is_string($value) && is_numeric($value)) {
+            return (int) $value;
+        }
```
kojiromike: Would phpstan complain if you simplified to this?

```suggestion
        if (is_numeric($value)) {
            return (int) $value;
        }
```

src/Common/Command/PhoneNotificationCommand.php
```diff
@@ -0,0 +1,335 @@
...
+            'greeting' => $greeting,
+            'timeRange' => $phoneTimeRange,
+            'type' => 'appointment',
+            'timeZone' => date('P'),
+            'callerId' => $facPhoneMap[$facilityId] ?? '',
+        ];
+
+        $response = $client->sendRequest('appointment', 'POST', $data);
+
+        if ($response instanceof RestResponse && $response->IsError) {
```
kojiromike, line 271: Break this function up a bit.

src/Common/Command/PhoneNotificationCommand.php
```diff
@@ -0,0 +1,335 @@
...
+        $facMsgMap = $facilities['msg_map'];
+        $defaultMessage = $this->defaultMessage($globals);
+
+        $patients = $this->fetchPendingPatients(self::NOTIFICATION_TYPE, $triggerHours);
+        $io->writeln(sprintf('%s: %d', xlt('Total Records Found'), count($patients)));
+
+        $client = $this->createMaviqClient($phoneId, $phoneToken, $phoneUrl);
+        $logDir = $globals->get('phone_reminder_log_dir');
+        $logPath = is_string($logDir) && $logDir !== ''
+            ? $logDir . '/phone_reminder_cronlog_' . date('Ymd') . '.html'
```
kojiromike, line 89: This line isn't covered in tests. It probably should be.

src/Common/Command/DocumentImportCommand.php
```diff
@@ -0,0 +1,347 @@
...
+        }
+
+        return Command::SUCCESS;
+    }
+
+    /**
+     * Pulls in legacy procedural helpers. Extracted so tests can override
+     * with a no-op.
+     */
+    protected function loadHelpers(): void
```
kojiromike, line 149: Everything from loadHelpers to removeFile is not covered by tests. If it can't be covered, it should be annotated as such so coverage analysis accounts for it.

tests/Tests/Isolated/Common/Command/DocumentImportCommandTest.php
kojiromike: Fixed in 42ae53887a — added `@codeCoverageIgnore` to `setUpBeforeClass` (PHPUnit lifecycle hook; bootstrap only).

tests/Tests/Isolated/Common/Command/PhoneNotificationCommandTest.php
kojiromike: Fixed in 42ae53887a — wrapped the file-level `require_once` block in `@codeCoverageIgnoreStart`/`@codeCoverageIgnoreEnd`.

tests/Tests/Isolated/Common/Command/PhoneNotificationCommandTest.php
kojiromike: Fixed in 42ae53887a — added `@codeCoverageIgnore` to `setUpBeforeClass`.

tests/Tests/Isolated/Common/Command/PhoneNotificationCommandTest.php
kojiromike: Fixed in 42ae53887a — added `testFakeMaviqClientThrowsWhenOutOfResponses` which asserts the guard fires when the canned-response queue is exhausted.

src/Common/Command/PhoneNotificationCommand.php
kojiromike: Applied in 42ae53887a — `intField` now reads: `if (is_numeric($value)) { return (int) $value; } return 0;`.

src/Common/Command/PhoneNotificationCommand.php
kojiromike: Fixed in 42ae53887a — extracted `formatApptDate`, `buildRequestData`, `formatErrorLog`, and `formatSuccessLog` helpers. `sendReminder` is now a flat orchestration of those helpers.

src/Common/Command/PhoneNotificationCommand.php
kojiromike: Fixed in 42ae53887a — added `testLogDirProducesPerDayCronLogPath` which sets `phone_reminder_log_dir`, runs a reminder, and asserts the resulting log path matches `{dir}/phone_reminder_cronlog_YYYYMMDD.html`.

src/Common/Command/DocumentImportCommand.php
kojiromike: Fixed in 42ae53887a — added `@codeCoverageIgnore` to every seam in `DocumentImportCommand` (`loadHelpers`, `iterateDirectory`, `resolveCategoryId`, `detectMimeType`, `findExistingDocument`, `persistInSituDocument`, `attachToCategory`, `addDocument`, `removeFile`). The same annotations were also added to `PhoneNotificationCommand`'s seams (`loadHelpers`, `decryptPassword`, `fetchFacilitiesMap`, `fetchPendingPatients`, `createMaviqClient`, `markReminderSent`, `insertLogEntry`, `writeLog`) since the same reasoning applies.

src/Common/Command/PhoneNotificationCommand.php
```diff
@@ -0,0 +1,404 @@
...
+            'error' => false,
+            'log' => $this->formatSuccessLog($firstName, $lastName, $phoneHome, $apptDate, $apptTime),
+        ];
+    }
+
+    private function formatApptDate(string $eventDate): string
+    {
+        $pieces = explode('-', $eventDate);
+        if (count($pieces) !== 3) {
+            return '';
```
kojiromike, line 295: This line is not covered in tests.

src/Common/Command/PhoneNotificationCommand.php
```diff
@@ -0,0 +1,404 @@
...
+        );
+    }
+
+    private function defaultMessage(OEGlobalsBag $globals): string
+    {
+        $messages = $globals->get('phone_appt_message');
+        if (is_array($messages) && isset($messages['Default']) && is_string($messages['Default'])) {
+            return $messages['Default'];
+        }
+        return '';
```
kojiromike, line 375: This line is not covered in tests.

src/Common/Command/PhoneNotificationCommand.php
```diff
@@ -0,0 +1,404 @@
...
+    private function stringField(array $row, string $key): string
+    {
+        $value = $row[$key] ?? null;
+        if (is_string($value)) {
+            return $value;
+        }
+        if (is_int($value) || is_float($value)) {
+            return (string) $value;
+        }
+        return '';
```
kojiromike, line 390: these lines are not covered in tests.

src/Common/Command/PhoneNotificationCommand.php
```diff
@@ -0,0 +1,404 @@
...
+    /**
+     * @param array<string, mixed> $row
+     */
+    private function intField(array $row, string $key): int
+    {
+        $value = $row[$key] ?? null;
+        if (is_numeric($value)) {
+            return (int) $value;
+        }
+        return 0;
```
kojiromike, line 402: this line is not covered in tests.

src/Common/Command/PhoneNotificationCommand.php
kojiromike: Covered in 5b8a0d9855 — `testNonScalarRowFieldsAndMalformedDateFallBackSafely` runs with `pc_eventDate = ''`, so `explode('-', '')` yields one piece and this `return ''` fires.

src/Common/Command/PhoneNotificationCommand.php
kojiromike: Covered in 5b8a0d9855 — the same new test sets `phone_appt_message` to `[]` (array but no `Default` key), so `defaultMessage` falls through to this `return ''`.

src/Common/Command/PhoneNotificationCommand.php
kojiromike: Covered in 5b8a0d9855 — the new test passes `pid = null` on the row, so `stringField` hits the non-string/int/float branch and returns `''`.

src/Common/Command/PhoneNotificationCommand.php
kojiromike: Covered in 5b8a0d9855 — the new test sets `pc_facility = 'not-n'`, which fails `is_numeric()` and hits this `return 0`.

Firehed (approved):
Room to take this further with service extraction, etc, but still a big win. LGTM

Issue #11667, opened 2026-04-16
Deprecate and remove `library/ajax/execute_background_services.php`


kojiromike:
# Deprecate and remove `library/ajax/execute_background_services.php`

Once the main-tab poll and phiMail refresh button have migrated to the REST API, `library/ajax/execute_background_services.php` has no in-repo callers, but it remains a documented cron entry point in the wild (e.g., `library/MedEx/MedEx_background.php:17` still suggests `0 0,4,8,12,16,20 * * * /usr/bin/env php ROOT_DIR/library/ajax/execute_background_services.php`).

## What

1. Emit a deprecation warning (to the PHP error log) when the file is invoked. Mention the two replacement paths:
   - UI: `POST /api/background_service[/:name]/run`
   - CLI: `bin/console background:services run --name=<service>`
2. Keep it functioning for at least one release cycle so operators can update their crontabs.
3. Remove the file in a subsequent release.

## Why

- The file has a confused dual identity (AJAX handler + CLI entry point) with two argument-parsing schemes.
- It bypasses modern request handling, stuffs arguments into `$_GET`/`$_POST`, and has existing entries in the PHPStan `forbiddenRequestGlobals` baseline.
- Its functionality is now fully covered by `BackgroundServiceRunner` via two clean entry points (REST and `bin/console`).

## Depends on

- "Migrate main-tab background services poll to the REST API"
- "Migrate phiMail refresh button to the REST API"
- "Fix MedEx background integration" (updates the docblock example still pointing here)

Issue #11662, opened 2026-04-16
Epic: unify how OpenEMR runs scheduled tasks


kojiromike:
# Epic: unify how OpenEMR runs scheduled tasks

OpenEMR currently has several overlapping mechanisms for running scheduled work:

1. The `background_services` table + `BackgroundServiceRunner`, driven by three different entry points:
   - AJAX poll from `interface/main/tabs/main.php` (runs only while users are logged in).
   - CLI mode of `library/ajax/execute_background_services.php` (running an `ajax/` file from cron).
   - `bin/console background:services run` — the modern path; also generates a ready-to-install crontab.
   - REST API at `POST /api/background_service/:name/run`.
2. The `library/allow_cronjobs.php` shim, used by standalone cron scripts (`interface/batchcom/batch_phone_notification.php`, `custom/zutil.cli.doc_import.php`).
3. Module-specific one-off cron scripts (`modules/sms_email_reminder/cron_*.php`).
4. Raw `INSERT INTO background_services` SQL in module installers that predates `BackgroundServiceRegistry`.

Goal: one execution core (`BackgroundServiceRunner`), two usage shapes:

- **HTTP** (UI and external triggers) via the REST API, using the same `APICSRFTOKEN` / LocalApi pattern the insurance-edit form already uses.
- **CLI** (cron and operator tooling) via `bin/console background:services`.

Everything else becomes redundant and should be deprecated or removed.

## Scope

Child issues:

- [x] #11664 — Add "run all due" entry points (REST + CLI). Subsumes #11677 (CLI exit-code on no-op outcomes).
- [ ] #11665 — Migrate main-tab AJAX poll to the REST endpoint.
- [ ] #11666 — Migrate phiMail refresh button to the REST endpoint.
- [ ] #11667 — Deprecate and remove `library/ajax/execute_background_services.php`.
- [ ] #11668 — Deprecate `library/allow_cronjobs.php` in favor of Symfony console commands.
- [ ] #11670 — Migrate module `background_services` installers to `BackgroundServiceRegistry`.
- [x] #11672 — Fix MedEx background integration (URL hack + stale docblock).
- [x] #11669 — Remove or migrate `modules/sms_email_reminder`.
- [ ] #11671 — Replace `running = -1` sentinel with a real `last_run_at` column.
- [x] #11673 — Decide policy for `BackgroundServiceRegistry::register()`'s handling of `active`.
- [x] #11661 — Background services have no recovery path for stuck locks.
[...]

PR #11727, opened 2026-04-20
refactor(background-services): migrate main-tab poll to REST API

### LINKED ISSUE #11665: Migrate main-tab background services poll to the REST API
# Migrate main-tab background services poll to the REST API

`interface/main/tabs/main.php:252-274` runs a 60-second loop that POSTs to `library/ajax/execute_background_services.php` to tick due background services while a user is logged in.

This should migrate to the REST API, using the same `APICSRFTOKEN` header / LocalApi pattern already in use at `templates/patient/insurance/insurance_edit.html.twig:387-410`.

## Why

- Collapses two execution paths into one (see epic).
- Removes a file whose dual AJAX/CLI identity has caused long-standing confusion.
- Normalizes authentication — the REST stack already handles CSRF, session auth, and error shapes; the legacy endpoint reimplements them inline.

## Depends on

- "Add REST endpoint for running all due background services"

## Caller notes

The current call only inspects `response.status !== 200` and does not parse the body. Parity with the existing behavior is low-risk.

kojiromike:
## Summary

- Migrates the 60-second logged-in poll in `interface/main/tabs/main.php` from the legacy `library/ajax/execute_background_services.php` endpoint to `POST /api/background_service/$run`.
- Uses the LocalApi `APICSRFTOKEN` header pattern already in use by `templates/patient/insurance/insurance_edit.html.twig`.
- Collapses the browser-poll execution path onto the same REST/`BackgroundServiceRunner` stack used by the CLI and existing REST endpoints.

Closes #11665. Part of the #11662 epic. Removal of `execute_background_services.php` is tracked separately in #11667.

## Why

- Normalizes authentication: the REST stack handles CSRF, session auth, and error shapes; the legacy endpoint reimplements them inline.
- No behavior change for the caller (only `response.status !== 200` is inspected; no body parsing is added).
- The REST stack does not touch `SessionTracker`, so no `skip_timeout_reset` equivalent is needed to keep the poll from resetting the session expiration timer.

## Test plan

- [x] Log in; observe DevTools Network tab: `POST /apis/<site>/api/background_service/$run` fires ~10 s after each `goRepeaterServices()` tick (60 s cadence).
- [x] Verify request carries the `APICSRFTOKEN` header and returns 200 with a `results` array.
- [x] Confirm `background_services.next_run` advances for due services.
- [x] With `OPENEMR__NO_BACKGROUND_TASKS=1`, confirm the poll is suppressed.
- [x] Leave a tab idle past the normal session timeout and confirm it still times out (poll should not reset the timer).

## Local verification

All five checklist items were driven manually against a fresh `docker/development-easy` worktree. Evidence captured below.

### URL, auth header, payload (HAR capture)

HAR entry for the first POST after login (`11:14:22 EDT`):

```
POST https://localhost:<port>/apis/default/api/background_service/$run
Headers:
  APICSRFTOKEN: a74987a31a74d6b8bf7cf60eadf317eb6d9688fe
  Cookie: App=OpenEMR; ...
Response: 200 OK (application/json, 160 ms)
Body: {"results":[
  {"name":"phimail","status":"skipped"},
  {"name":"X12_SFTP","status":"skipped"},
  {"name":"UUID_Service","status":"executed"},
  {"name":"Email_Service","status":"executed"}
]}
```

### 60 s cadence (Apache access log)

Consecutive POSTs from the same session, computed from the `access.log` timestamps:

```
15:14:22   (first hit, ~10 s after goRepeaterServices() tick)
15:15:23   (+61 s)
15:16:23   (+60 s)
15:17:24   (+61 s)
```

### `next_run` advancement

Queried `background_services` immediately after the `15:14:22` execution (`now` = `15:16:08`):

| name          | execute_interval | next_run            | delta from execution |
| ------------- | ----------------:| ------------------- | -------------------- |
| Email_Service | 2 min            | 2026-04-21 15:16:22 | +2 min               |
| UUID_Service  | 240 min          | 2026-04-21 19:14:22 | +4 h                 |

Matches the `executed` status for both services in the response payload. `phimail` and `X12_SFTP` remained at their epoch placeholders (they returned `skipped`).

### `OPENEMR__NO_BACKGROUND_TASKS=1` suppresses the `$run` POST

Recreated the `openemr` container with `OPENEMR__NO_BACKGROUND_TASKS=1` in the environment, logged back in, left the tab idle for three full 60 s cadence cycles. Apache access log for that window:

```
$run POSTs since login (15:24:58):
  (none)

dated_reminders_counter POSTs since login:
  15:24:59   (first hit — goRepeaterServices() fires immediately)
  15:25:59   (+60 s cadence tick)
  15:27:00   (+60 s cadence tick)
```

`goRepeaterServices()` continues to run on its 60 s timer (legacy reminders poll still fires), but the `if (!noBackgroundTasks) fetch(...)` gate correctly suppresses only the `$run` call. Reverted the env var afterwards.

### Session timer not reset by the poll (structural)

The `$run` request flows `apis/dispatch.php → ApiApplication → LocalApiAuthorizationController → BackgroundServiceRunner`. None of that path includes `library/auth.inc.php`, and a repo-wide grep of `SessionTracker` across `apis/`, `src/RestControllers/`, `src/Common/Http/`, and `src/Common/Auth/` returns zero hits. `SessionTracker::updateSessionExpiration()` is only called from `library/auth.inc.php:116`, which the REST stack bypasses entirely — so the `$run` poll structurally cannot reset the session expiration timer. This is a stronger guarantee than an empirical timeout test, as it holds for every REST request rather than only the ones observed.

### DIFF
--- a/interface/main/tabs/main.php
+++ b/interface/main/tabs/main.php
@@ -10,10 +10,12 @@
  * @author    Ranganath Pathak <pathak@scrs1.org>
  * @author    Jerry Padgett <sjpadgett@gmail.com>
  * @author    Stephen Nielson <snielson@discoverandchange.com>
+ * @author    Michael A. Smith <michael@opencoreemr.com>
  * @copyright Copyright (c) 2016 Kevin Yeh <kevin.y@integralemr.com>
  * @copyright Copyright (c) 2016-2019 Brady Miller <brady.g.miller@gmail.com>
  * @copyright Copyright (c) 2019 Ranganath Pathak <pathak@scrs1.org>
  * @copyright Copyright (c) 2024 Care Management Solutions, Inc. <stephen.waite@cmsvt.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
@@ -128,6 +130,13 @@ $twig = (new TwigContainer(null, OEGlobalsBag::getInstance()->getKernel()))->get
         var isPortalEnabled = "<?php echo OEGlobalsBag::getInstance()->getBoolean('portal_onsite_two_enable') ?>";
         // Set the csrf_token_js token that is used in the below js/tabs_view_model.js script
         var csrf_token_js = <?php echo js_escape(CsrfUtils::collectCsrfToken($session)); ?>;
+        // Separate CSRF token for calls to the REST/LocalApi stack (sent as the APICSRFTOKEN header).
+        var api_csrf_token_js = <?php echo js_escape(CsrfUtils::collectCsrfToken($session, 'api')); ?>;
+        <?php
+        $sessionSiteId = $session->get('site_id');
+        $sessionSiteIdString = is_string($sessionSiteId) ? $sessionSiteId : '';
+        ?>
+        var site_id_js = <?php echo js_escape($sessionSiteIdString); ?>;
         var userDebug = <?php echo js_escape(OEGlobalsBag::getInstance()->get('user_debug')); ?>;
         var webroot_url = <?php echo js_escape($web_root); ?>;
         var jsLanguageDirection = <?php echo js_escape($session->get('language_direction')); ?> ||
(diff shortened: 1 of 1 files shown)


kojiromike:
Bench-testing this against a fresh `docker/development-easy` install surfaced a regression that isn't caused by this change but *does* block it: `LocalApiAuthorizationController` 500s when the authenticated user has no `users.uuid` yet (fresh installs / pre-UUID upgrades), because `setRequestUser()` ends up calling `Uuid::fromString(null)`. This PR is the first in-repo caller that reaches LocalApi before any OAuth-driven UUID bootstrap has run on a fresh session, so it turns the latent bug into a user-visible 500 on the background-services poll.

Splitting the fix out rather than bundling it here:

- #11735 — `fix(local-api): bootstrap missing user UUID to prevent 500 on fresh installs`. Populates `users.uuid` on demand inside `authorizeRequest()`, mirroring the OAuth password-grant bootstrap in `OpenIDConnect\Repositories\UserRepository::getAccountByPassword()`. Keeps the behavior change reviewable on its own.

Once #11735 merges I'll rebase this PR on top so the background-services migration stays a pure REST-endpoint swap. Happy-path LocalApi call is already verified end-to-end against a container where the fix is applied — response is a clean 200 with the expected `{"results":[...]}` payload.

kojiromike:
Both blockers have landed on master (#11735 and #11736), and I've rebased this PR on top of them. The latent `Uuid::fromString(null)` 500 is fixed, and the UUID bootstrap called from `LocalApiAuthorizationController::setRequestUser()` is now the scoped single-row helper (`UuidRegistry::createMissingUuidForRow`) rather than the whole-table sweep, so the background-services poll stays cheap on the authenticated hot path. Ready for review.

kojiromike:
@aisle-analyzer

PR #11739, opened 2026-04-21
ci: add ScanCode license-scan workflow (PoC)

kojiromike:
## Summary

Adds a GitHub Actions workflow that runs [ScanCode Toolkit](https://github.com/nexB/scancode-toolkit) on files changed in a PR, to detect embedded license text, copyright notices, and package manifests.

This is a **proof of concept**. The intent is to start a conversation about whether OpenEMR wants a per-file license-compliance check in CI. Nothing is wired up as a required check, and the workflow never fails — it only reports findings.

## Motivation

Tools like Snyk check *declared* dependency licenses from `composer.json` / `package.json`. They do not detect cases where source code licensed under a different license (MIT, Apache, BSD, etc.) has been copy-pasted into the repo. ScanCode fingerprints license text inside files themselves, so it can distinguish GPL vs LGPL vs MIT vs Apache findings on a per-file basis. This is the class of problem that came up on #11416 (where code from an MIT-licensed upstream is being imported).

## How it works

- Trigger: `pull_request` against `master` / `rel-*`, plus `workflow_dispatch`.
- Scans only files added/modified in the PR (excluding `vendor`, `node_modules`, `ccdaservice/node_modules`, `Documentation`). Scoping to the diff keeps runtime manageable on a ~1M-LOC repo.
- ScanCode runs with `--license --copyright --package --only-findings`.
- Results are summarized in the GitHub Actions step summary (a table of license expressions + per-file breakdown) and uploaded as a JSON + HTML artifact for 30 days.
- The job is non-fatal — it surfaces findings but does not block merges.

## Not in scope for this PR

- Full-repo scans (would need scheduled runs + caching).
- SARIF upload / Code Scanning integration (ScanCode output would need mapping).
- A policy/allowlist of acceptable license expressions with fail-on-violation.
- SPDX SBOM generation.

Happy to iterate on any of these in follow-ups if there's interest. Also open to scrapping this in favor of [REUSE](https://reuse.software/) (stricter — requires `SPDX-License-Identifier` headers in every file).

## Test plan

- [ ] Workflow runs on this PR and produces a step summary table of license findings for the one changed file (the workflow itself).
- [ ] Artifact `scancode-results` contains `scancode-results.json` and `scancode-results.html`.
- [ ] Confirm the job surfaces the `apache-2.0` / `gpl-3.0` distinctions we'd expect if ScanCode is working.

### DIFF
--- a/.github/workflows/scancode.yml
+++ b/.github/workflows/scancode.yml
@@ -6,6 +6,14 @@ on:
     - master
     - rel-*
   workflow_dispatch:
+    inputs:
+      scan_mode:
+        description: Scan scope
+        type: choice
+        options:
+        - full
+        - diff
+        default: full
 
 permissions:
   contents: read
@@ -17,8 +25,11 @@ concurrency:
 
 jobs:
   scancode:
-    name: ScanCode License Scan (PR diff)
+    name: ScanCode License Scan
     runs-on: ubuntu-latest
+    # Full-repo scans on a ~1M-LOC codebase can take a while; give the job
+    # headroom without hitting the 6-hour ceiling.
+    timeout-minutes: 180
     steps:
     - name: Checkout code
       uses: actions/checkout@v6
@@ -33,15 +44,49 @@ jobs:
     - name: Install ScanCode Toolkit
       run: pipx install scancode-toolkit
 
-    - name: Stage changed files for scan
+    - name: Determine scan mode
+      id: mode
+      env:
+        EVENT: ${{ github.event_name }}
+        DISPATCH_MODE: ${{ github.event.inputs.scan_mode }}
+        HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+      # Resolution rules:
+      #   workflow_dispatch → whatever the user picked (default: full)
+      #   pull_request     → 'full' if the HEAD commit carries trailer
+      #                      `Scancode-Full-Scan: true`, otherwise 'diff'
+      run: |
+        git config --global --add safe.directory "$GITHUB_WORKSPACE"
+        case "$EVENT" in
+          workflow_dispatch)
+            mode="${DISPATCH_MODE:-full}"
+            ;;
+          pull_request)
+            trailer=$(
+              git log -1 --format='%(trailers:key=Scancode-Full-Scan,valueonly)' "$HEAD_SHA" \
+                | tr -d '[:space:]'
+            )
+            if [[ $trailer == 'true' ]]; then
+              mode='full'
+            else
+              mode='diff'
+            fi
+            ;;
+          *)
+            mode='diff'
+            ;;
+        esac
+        printf 'Scan mode: %s\n' "$mode"
+        echo "mode=$mode" >> "$GITHUB_OUTPUT"
+
+    - name: Stage changed files for scan (diff mode)
       id: stage
+      if: steps.mode.outputs.mode == 'diff' && github.event_name == 'pull_request'
       env:
         BASE_SHA: ${{ github.event.pull_request.base.sha }}
         HEAD_SHA: ${{ github.event.pull_request.head.sha }}
-      # Only scan files added or modified in the PR so the check stays fast on
-      # a large repo. Exclude vendored / generated trees.
+      # Only scan files added or modified in the PR so the check stays fast.
+      # Exclude vendored / generated trees.
       run: |
-        git config --global --add safe.directory "$GITHUB_WORKSPACE"
         mkdir -p scancode-input
         mapfile -t changed < <(
           git diff --name-only --diff-filter=AM "$BASE_SHA" "$HEAD_SHA" \
@@ -65,8 +110,8 @@ jobs:
         done
         printf 'Staged %d files for scan.\n' "${#changed[@]}"
 
-    - name: Run ScanCode
-      if: steps.stage.outputs.has_changes == 'true'
+    - name: Run ScanCode (diff)
+      if: steps.mode.outputs.mode == 'diff' && steps.stage.outputs.has_changes == 'true'
       # --license / --copyright: detect license text and copyright notices
       # --package: detect package manifests (composer.json, package.json, etc.)
       # --only-findings: skip files with no license/copyright findings in output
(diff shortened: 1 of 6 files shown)


.github/workflows/scancode.yml
```diff
@@ -0,0 +1,223 @@
...
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+
+    - name: Install ScanCode Toolkit
+      run: pipx install scancode-toolkit
```
kojiromike: should this be version pinned? Better yet, is there an action version we can use here? We want dependabot to maintain the versions of our tools.

.github/workflows/scancode.yml
```diff
@@ -0,0 +1,223 @@
...
+        case "$EVENT" in
+          workflow_dispatch)
+            mode="${DISPATCH_MODE:-full}"
+            ;;
+          pull_request)
+            trailer=$(
+              git log -1 --format='%(trailers:key=Scancode-Full-Scan,valueonly)' "$HEAD_SHA" \
+                | tr -d '[:space:]'
+            )
+            if [[ $trailer == 'true' ]]; then
```
kojiromike:
```suggestion
            if [[ $trailer = 'true' ]]; then
```

.github/workflows/scancode.yml
```diff
@@ -0,0 +1,223 @@
...
+        BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+      # Only scan files added or modified in the PR so the check stays fast.
+      # Exclude vendored / generated trees.
+      run: |
+        mkdir -p scancode-input
+        mapfile -t changed < <(
+          git diff --name-only --diff-filter=AM "$BASE_SHA" "$HEAD_SHA" \
+            -- \
+            ':!vendor' \
```
kojiromike: why exclude what's already in .gitignore?

.github/workflows/scancode.yml
```diff
@@ -0,0 +1,223 @@
...
+        )
+        if (( ${#changed[@]} == 0 )); then
+          echo 'No files to scan.'
+          echo 'has_changes=false' >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+        echo 'has_changes=true' >> "$GITHUB_OUTPUT"
+        for f in "${changed[@]}"; do
+          if [[ -f $f ]]; then
+            mkdir -p "scancode-input/$(dirname "$f")"
```
kojiromike: use a parameter expansion instead of `dirname`

.github/workflows/scancode.yml
```diff
@@ -0,0 +1,223 @@
...
+      # the output without adding useful findings, and skip the on-disk scan
+      # cache dir.
+      run: |
+        scancode \
+          --license \
+          --copyright \
+          --package \
+          --only-findings \
+          --processes "$(nproc)" \
+          --ignore 'vendor/*' \
```
kojiromike: can we use .gitignore or another config file instead of hard-coding config in here?

.github/workflows/scancode.yml
```diff
@@ -0,0 +1,223 @@
...
+          .
+
+    - name: Summarize findings
+      if: |
+        steps.mode.outputs.mode == 'full'
+        || (steps.mode.outputs.mode == 'diff' && steps.stage.outputs.has_changes == 'true')
+      env:
+        SCAN_MODE: ${{ steps.mode.outputs.mode }}
+      run: |
+        python3 - <<'PY' >> "$GITHUB_STEP_SUMMARY"
```
kojiromike: doesn't gha support changing the shell to python so the run value is directly interpreted by python?

kojiromike:
### Previous full-repo run (pre-sharding)

Run: https://github.com/openemr/openemr/actions/runs/24744516376

- **Wall clock:** 38m47s on a single `ubuntu-latest` runner.
- **Throughput:** 8348 files at ~3.60 files/sec with 4 worker processes.
- **Outcome:** ScanCode exited 1, which caused the summary + artifact-upload steps (gated on the previous step's success) to skip — so no findings artifact for that run.

The exit-1 was a **scanner error, not a license finding**. A single file (`interface/modules/zend_modules/public/js/installer/build/static/js/lhc-forms.js.map`) crashed a worker; ScanCode documents this as "scan completed, but some per-file errors." This is orthogonal to license detection, which runs to completion regardless of which licenses it finds.

**Fixes on this push:**

1. Added `*.map`, `*.min.js`, `*.min.css` to `.scancode-ignore` — ScanCode won't open them. Source maps and minified bundles carry no useful license text anyway, and are the class of file most likely to trip the scanner.
2. Wrapped the scancode invocation to treat exit 1 as non-fatal (only exit > 1 fails the job). Belt-and-suspenders for any other pathological files we haven't discovered yet — the summary + artifact steps now run regardless.
3. Fanned the full-repo scan out to an 8-way matrix shard (`hash(path) % 8`). ScanCode's license matcher is CPU-bound, so the speedup should be close to linear — expect roughly 5m wall clock with a modest increase in aggregate runner-minutes.

The empty commit on top of this push carries the `Scancode-Full-Scan: true` trailer to exercise the sharded full-scan path on this PR.

.github/workflows/scancode.yml
kojiromike: Pinned the CLI version via a `SCANCODE_VERSION` env var so all three install steps bump together. Left a comment explaining why I didn't move to `aboutcode-org/scancode-action` — it's still tagged `@beta`, runs ScanCode.io pipelines (different output schema, would require rewriting the summary step), and has no stable release yet. Revisiting for Dependabot once the action stabilizes.

.github/workflows/scancode.yml
kojiromike: Applied.

.github/workflows/scancode.yml
kojiromike: Dropped the `vendor/`, `node_modules/`, and `ccdaservice/node_modules/` excludes — `git diff --name-only` never lists untracked or ignored paths, so those were redundant. Kept `:!Documentation` since it is tracked.

.github/workflows/scancode.yml
kojiromike: Replaced with `d="${f%/*}"` and a `[[ $d = "$f" ]] && d='.'` fallback for top-level files.

.github/workflows/scancode.yml
kojiromike: Extracted to `.scancode-ignore` at the repo root. Full-scan shard job reads it and converts each entry to a `:(exclude)` git pathspec. I kept it as a separate file rather than reusing `.gitignore` because the two lists serve different purposes — `.gitignore` governs what git tracks, `.scancode-ignore` governs what the scanner opens (e.g., we want to skip `Documentation/` for the scan even though it is checked in, and skip `*.map` even though those files are often committed artifacts).

.github/workflows/scancode.yml
kojiromike: Good call — switched the aggregate merge step to `shell: python`. No more heredoc / `python3 - <<EOF` wrapping.

kojiromike:
### Sharded full-repo run — results

Run: https://github.com/openemr/openemr/actions/runs/24746703406

**Wall clock: 6m31s** (vs. 38m47s single-runner baseline — ~6× speedup). The longest shard was `shard 5/8` at ~5m53s; aggregate step ran in 19s. All 8 shards + aggregate completed successfully with no scan errors (the earlier `lhc-forms.js.map` crash is now filtered by `.scancode-ignore`).

---

### Next run: demonstrating an annotated policy failure

The just-pushed commit (`08c2af9c59`) is a temporary \"DO NOT MERGE\" commit that:

1. Extends `tools/ci/scancode-summary.py` with a `SCANCODE_DISALLOWED_LICENSES` env var. When set, the summary step emits GitHub `::error file=…::` annotations for each matching file and exits non-zero.
2. Sets `SCANCODE_DISALLOWED_LICENSES='proprietary-license,commercial-license'` in both Summarize-findings steps.
3. Adds `tools/ci/scancode-demo-fixture.txt` with a clearly proprietary header.

The commit also carries `Scancode-Full-Scan: true` so both the diff path and the sharded full-repo path run — both should detect the fixture, annotate it in the PR diff view, and fail the job. That's what we want to see: ScanCode attributes the license correctly, the policy check catches it, and the failure is linked to the specific file rather than a vague scanner error.

Revert `08c2af9c59` to roll the demo back before merge. The enforcement logic itself is benign when `SCANCODE_DISALLOWED_LICENSES` is unset and could stay as a real feature.

PR #11774, opened 2026-04-22
feat(cache): add PSR-16 cache abstraction with three backends

zmilan:
Split of #11416 

# [2/4] PSR-16 cache abstraction with three backends

## Context

This is **part 2 of 4** in the SSO/OIDC/GCIP split (tracked in the original monolithic branch). The work has been broken into four independently shippable PRs for reviewer clarity:

| No. | Title | Status |
|---|-------|--------|
| 1 | OIDC core (discovery, validator, session refresh, events) | Waiting on No.2 |
| **2** | **PSR-16 cache abstraction with three backends** | **← this PR** |
| 3 | GCIP custom module | Waiting on No.1 |
| 4 | `login_counter_ip_tracker.php` hardening | Awaiting scope clarification |

Infrastructure-only — no current consumer in this PR. Designed for the forthcoming OIDC validator/discovery layer (PR No.1) and available for general reuse across OpenEMR (session caching, rate-limit state, route caching, etc.).

## What this adds

A PSR-16 (`psr/simple-cache`) cache abstraction under `OpenEMR\Common\Cache` with three backend implementations.

### Source (`src/Common/Cache/`)

| File | Purpose |
|------|---------|
| `CacheEntry.php` | Typed envelope for all cached values. Restricts `unserialize()` to this class only (allowed-classes whitelist) to prevent PHP object-injection attacks. |
| `CacheException.php` | Base cache exception. Implements `Psr\SimpleCache\CacheException` marker interface. |
| `CacheInvalidArgumentException.php` | Invalid-key exception. Implements `Psr\SimpleCache\InvalidArgumentException`. |
| `FilesystemCache.php` | PSR-16 backend using serialized files with atomic temp-file + rename writes. Suitable for single-instance deployments. |
| `DatabaseCache.php` | PSR-16 backend using the new `app_cache` table. Suitable for multi-instance deployments sharing a database. Includes `purgeExpired()` for periodic cleanup. |
| `RedisCache.php` | PSR-16 backend using the native `\Redis` class from `ext-redis` (the client OpenEMR already requires — same pattern as `LockingRedisSessionHandler`). Suitable for high-traffic and multi-instance deployments. |

All three concrete backends implement `Psr\SimpleCache\CacheInterface`. All three share:
- Key validation (non-empty, no reserved characters `{}()/\\@:`, length-bounded where the backend requires it)
- `CacheEntry` envelope for safe serialize/unserialize
- Full PSR-16 surface: `get`/`set`/`delete`/`clear`/`has`/`getMultiple`/`setMultiple`/`deleteMultiple`, `ttl` as `int|DateInterval|null`

### Schema

New generic `app_cache` table. Added to:
- `sql/database.sql` (full schema, fresh installs)
- `sql/8_1_0-to-8_1_1_upgrade.sql` (with `#IfNotTable` guard, existing 8.1.0 → 8.1.1 upgrades)
- `sql/patch.sql` (with `#IfNotTable` guard, patch path)

```sql
CREATE TABLE `app_cache` (
    `cache_key` VARCHAR(255) NOT NULL COMMENT 'PSR-16 cache key',
    `cache_value` LONGBLOB NOT NULL COMMENT 'Serialized CacheEntry',
    `expires_at` DATETIME DEFAULT NULL COMMENT 'NULL = no expiration',
    PRIMARY KEY (`cache_key`),
    KEY `idx_expires_at` (`expires_at`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

### Dependencies

- `psr/simple-cache ^3.0` declared as a **direct** dep in `composer.json`. It was already transitively available (brought in by `monolog/monolog` and others) but is now used directly by the new code, so `composer-require-checker` requires the explicit declaration.

### Version bump

- `$v_database` 537 → 538 (both `version.php` and `sql/database.sql` header comment)

### Isolated (`tests/Tests/Isolated/Common/Cache/`, 74 tests)

- `CacheEntryTest` — construction, expiry semantics, safe-unserialize options, immutability
- `FilesystemCacheTest` — round-trip of various types, TTL semantics (int + `DateInterval`), directory validation, reserved-character keys, corrupted-file recovery, atomic writes (no orphan temp files)
- `RedisCacheTest` — same surface against a PHPUnit mock of `\Redis` wired with `willReturnCallback` to simulate an in-memory Redis. No real Redis server required.

### Integration (`tests/Tests/Integration/Common/Cache/`)

- `DatabaseCacheTest` — full surface against a real database, plus `purgeExpired()` behavior.

## Security considerations

- **Object-injection safe**: all backends call `unserialize()` with `CacheEntry::unserializeOptions()` which whitelists only `CacheEntry`. An attacker who writes malicious bytes into cache storage cannot instantiate arbitrary classes.
- **Key validation** rejects characters commonly used in injection attacks (braces, slashes, colons) and empty strings. `DatabaseCache` additionally enforces the 255-char column limit.
- **Atomic writes** in `FilesystemCache` via temp-file + rename — no partial-write races, no orphan temp files.
- **No privilege escalation surface** — cache code never executes cached content; it only returns values to callers.

## Review notes

- The cache classes live in `OpenEMR\Common\Cache` (not the OIDC namespace) because they are genuinely generic PSR-16 implementations with zero OIDC-aware behavior. This makes the abstraction reusable for any future feature that needs caching.
- `RedisCache` uses `\Redis` (the PECL extension) rather than Predis to match OpenEMR's existing Redis usage (`LockingRedisSessionHandler` uses `\Redis`)—no new PHP-library dep.
- The `app_cache` table name is intentionally generic ("application cache") rather than feature-specific, consistent with the framework convention used by Laravel and Symfony.

## Out of scope (explicit)

- **No wiring to any consumer** in this PR. The three backends stand alone. PR No.1 (OIDC core) will be the first consumer.
- **No default-backend selector** / factory. Consumers pick and inject the backend they need.
- **No cache-invalidation tooling** beyond `clear()` and `purgeExpired()` (on DatabaseCache). A cron-style periodic sweep is a follow-up if needed.

## Assisted by Claude Code.

### DIFF
--- a/src/Common/Cache/FilesystemCache.php
+++ b/src/Common/Cache/FilesystemCache.php
@@ -1,0 +1,202 @@
+<?php
+
+/**
+ * PSR-16 simple cache backed by the local filesystem.
+ *
+ * Each cache entry is stored as a serialized file. Suitable for single-instance
+ * deployments and development. For multi-instance or containerized environments,
+ * prefer DatabaseCache or RedisCache.
+ *
+ * @link      https://www.open-emr.org
+ * @author    Milan Zivkovic <zivkovic.milan@gmail.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Common\Cache;
+
+use Psr\SimpleCache\CacheInterface;
+
+final readonly class FilesystemCache implements CacheInterface
+{
+    private string $directory;
+
+    /**
+     * @param string $directory Absolute path to the cache directory.
+     */
+    public function __construct(string $directory)
+    {
+        $realpath = realpath($directory);
+        if ($realpath === false || !is_dir($realpath)) {
+            throw new CacheException("Cache directory does not exist: {$directory}");
+        }
+        if (!is_writable($realpath)) {
+            throw new CacheException("Cache directory is not writable: {$directory}");
+        }
+        $this->directory = $realpath;
+    }
+
+    public function get(string $key, mixed $default = null): mixed
+    {
+        $this->validateKey($key);
+
+        $path = $this->pathForKey($key);
+        if (!is_file($path)) {
+            return $default;
+        }
+
+        $contents = file_get_contents($path);
+        if ($contents === false) {
+            return $default;
+        }
+
+        $entry = @unserialize($contents, CacheEntry::unserializeOptions());
+        if (!$entry instanceof CacheEntry) {
+            @unlink($path);
+            return $default;
+        }
+
+        if ($entry->isExpired(time())) {
+            @unlink($path);
+            return $default;
+        }
+
+        return $entry->value;
+    }
+
+    public function set(string $key, mixed $value, \DateInterval|int|null $ttl = null): bool
+    {
+        $this->validateKey($key);
+
+        $expiresAt = $this->ttlToTimestamp($ttl);
+        $entry = new CacheEntry($value, $expiresAt);
+
+        $path = $this->pathForKey($key);
+        $tempPath = $path . '.' . bin2hex(random_bytes(4)) . '.tmp';
+
+        $written = file_put_contents($tempPath, serialize($entry));
+        if ($written === false) {
+            @unlink($tempPath);
+            return false;
+        }
+
+        return rename($tempPath, $path);
+    }
+
+    public function delete(string $key): bool
+    {
+        $this->validateKey($key);
+
+        $path = $this->pathForKey($key);
+        if (!is_file($path)) {
+            return true;
+        }
+
+        return @unlink($path);
+    }
+
+    public function clear(): bool
+    {
+        $files = glob($this->directory . DIRECTORY_SEPARATOR . '*.cache');
+        if ($files === false) {
+            return false;
+        }
+
+        $success = true;
+        foreach ($files as $file) {
+            if (!@unlink($file)) {
+                $success = false;
+            }
+        }
+
+        return $success;
+    }
+
+    public function has(string $key): bool
+    {
+        return $this->get($key, $this) !== $this;
+    }
+
+    /**
+     * @param iterable<string> $keys
+     * @return iterable<string, mixed>
+     */
+    public function getMultiple(iterable $keys, mixed $default = null): iterable
+    {
+        $results = [];
+        foreach ($keys as $key) {
+            $results[$key] = $this->get($key, $default);
+        }
+
+        return $results;
+    }
+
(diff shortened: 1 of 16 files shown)


src/Common/Cache/FilesystemCache.php
```diff
@@ -0,0 +1,202 @@
...
+use Psr\SimpleCache\CacheInterface;
+
+final readonly class FilesystemCache implements CacheInterface
+{
+    private string $directory;
+
+    /**
+     * @param string $directory Absolute path to the cache directory.
+     */
+    public function __construct(string $directory)
```
Firehed, line 28: If we must maintain this ourselves, it'd be _highly_ preferable to hook through the new Flysystem tooling rather than directly interacting with the FS.

Firehed (changes requested):
Before continuing, make sure this is brought up to date with master - I see a couple things that just changed which are primed for some weirdness.

More broadly - there are [a lot](https://packagist.org/providers/psr/simple-cache-implementation) of PSR-16 cache implementations out there. `symfony/cache` is going to get added imminently as part of ORM work anyway. I would _strongly_ prefer to see us pull in that package rather than re-implement caching from scratch. It [contains implementations](https://github.com/symfony/cache/tree/8.1/Adapter) for everything you've added here and more.

If there's a _very_ good reason for reimplemention I'm open to it, but this feels like the type of thing better off pulled in as library code.

zmilan:
> Before continuing, make sure this is brought up to date with master - I see a couple things that just changed which are primed for some weirdness.
> 
> More broadly - there are [a lot](https://packagist.org/providers/psr/simple-cache-implementation) of PSR-16 cache implementations out there. `symfony/cache` is going to get added imminently as part of ORM work anyway. I would _strongly_ prefer to see us pull in that package rather than re-implement caching from scratch. It [contains implementations](https://github.com/symfony/cache/tree/8.1/Adapter) for everything you've added here and more.
> 
> If there's a _very_ good reason for reimplemention I'm open to it, but this feels like the type of thing better off pulled in as library code.

So, I initially added this as part of OIDC, so that we can cache data... and as far as I know, there was no ready-to-use mechanism for such a purpose... but you are totally right about not writing something ours if there are ready-to-use tools, such as symfony/cache... I think we can close this PR, and I can switch to using symfony/cache instead... thanks... extracting it from the original OIDC PR makes this very clear :)

Firehed:
SGTM! I forgot, the addition of symfony/cache actually landed already, I was just looking at the wrong branch when I posted the comment. So hopefully that will be a straightforward merge :)

bradymiller:
hi @Firehed , where is the cache stored for the symfony/cache implementation that was brought in?

PR #11776, opened 2026-04-22
ci(mago): add advisory-only Mago workflow (PoC)

kojiromike:
## Summary

Adds a **proof-of-concept, advisory-only** GitHub Actions workflow for [Mago](https://mago.carthage.software/) — a Rust-based PHP toolchain that combines linting, static analysis, and formatting in a single tool. The workflow never blocks a PR (every Mago step is wrapped in `continue-on-error`); it exists to surface what Mago would catch so we can decide whether to adopt it, and on what terms (subset of rules, baseline-gated, blocking, etc.).

Take it or leave it.

## What's in the PR

- `.github/workflows/mago.yml` — runs on push/PR to `master`/`rel-*`, uses the official `nhedger/setup-mago@v1` action (version pinned to `1.24.0`), emits GitHub inline annotations, uploads a JSON report, and writes a per-run rule-count summary to the job summary.
- `mago.toml` — scopes linting/analysis to `src/` only, with `vendor/` and `library/` declared as symbol-only `includes` so the analyzer can resolve references without reporting issues in those legacy trees.

## What Mago does vs. what we already run

| Capability | Mago | PHPStan L10 | PHPCS | Rector |
|---|---|---|---|---|
| Type inference / flow analysis | `mago analyze` (311 issue codes) | yes | — | partial |
| Style rules / code smells | `mago lint` (105 rules, 9 categories) | via custom | yes | — |
| Auto-fix safe issues | yes | — | phpcbf | yes |
| Code modernization / PHP version upgrades | **no** | — | — | **yes** |
| Formatting | PER-CS | — | yes | — |
| Baseline support | yes | yes | — | — |
| Architectural rules | `mago guard` | via custom | — | — |
| Security rules | 8 built-in | via strict-rules | custom | — |
| Complexity metrics | halstead / kan-defect / cyclomatic | — | limited | — |

Mago is **not a Rector replacement** — no semantic refactors / PHP-version upgrades. It pairs with our existing stack as a fast feedback layer.

## PoC findings (scoped to `src/Common`, 176 files)

Ran locally on the same tree this PR targets. Timing: lint 0.1 s · analyze 2.2 s. (The docs' "86×/44× faster than Pretty-PHP/PHPStan" claim checked out on this slice.)

### `mago lint` — 2 015 issues

| Count | Rule | Notes |
|------:|------|-------|
| 431 | `no-empty` | matches our documented preference to avoid `empty()` |
| 252 | `no-else-clause` | matches our "no else after return" rule |
| 241 | `identity-comparison` | `==` → `===` |
| 164 | `literal-named-argument` | |
| 152 | `strict-types` | missing `declare(strict_types=1);` |
| 143 | `braced-string-interpolation` | |
| 81 | `no-isset` | |
| 45 | `no-empty-comment` | |
| 45 | `prefer-static-closure` | |
| 44 | `tagged-todo` | |
| 43 | `strict-behavior` | |
| 42 | `no-boolean-flag-parameter` | |
| 41 | `sensitive-parameter` | security |

By level: 825 Warning · 623 Error · 379 Help · 188 Note.

### `mago analyze` — 2 626 issues

| Count | Code | |
|------:|------|--|
| 568 | `mixed-assignment` | PHPStan L10 territory |
| 527 | `mixed-operand` | |
| 329 | `mixed-argument` | |
| 183 | `mixed-array-access` | |
| 151 | `mixed-method-access` | |
| 100 | `mixed-return-statement` | |
| 61 | `deprecated-method` | |
| 52 | `invalid-iterator` | |
| 37 | `possibly-false-argument` | |
| 33 | `deprecated-function` | |
| 29 | `non-existent-class-like` | |
| 26 | `non-existent-method` | |
| 22 | `redundant-null-coalesce` | |
| 21 | `invalid-return-statement` | |
| 16 | `impossible-condition` | dead-code detection |

### A few representative real findings (not just style)

- `src/Common/Command/RegisterApiTestClientCommand.php:86` — condition always `false` (dead code)
- `src/Common/ORDataObject/Person.php:561` — `deactivate` declared `: bool`, actually returns `bool|int`
- `src/Common/ORDataObject/ORDataObject.php:125` — calls undefined `get_id`
- `src/Common/Session/SessionTracker.php:89` — redundant `??` (lhs can never be null)

## Gaps vs. our current toolchain

1. **No Rector replacement.** Mago does not do semantic refactors. Rector would still be needed for modernization passes.
2. **Custom PHPStan rules don't port.** The rules under `tests/PHPStan/Rules/` (forbidden globals, etc.) have no direct equivalent — Mago's rule set isn't user-extensible like PHPStan's.
3. **Encoding quirks.** Three `vendor/symfony/cache` and `vendor/composer` files reported as invalid UTF-8 (lossy conversion applied) — non-fatal, worth filing upstream if we adopt.

## Where it would be a clean win if we adopt

- **Pre-commit speed**: `mago lint --staged` + `mago analyze --staged` finishes in well under a second on changed files.
- **Already aligned**: `no-empty`, `no-else-clause`, `strict-types`, `identity-comparison`, `sensitive-parameter` map directly to rules we already enforce via convention — no custom-rule maintenance.
- **Pairs with, does not replace**: keep PHPStan (custom rules), keep Rector (modernization), add Mago as the fast feedback layer + formatter.

## Suggested follow-ups if we decide to take it

- Generate baselines (`mago lint --generate-baseline …`, same for `analyze`) and flip the workflow from advisory to blocking-for-new-issues-only.
- Add `composer mago-lint` / `composer mago-analyze` scripts.
- Wire into pre-commit.
- File the UTF-8 issues upstream.

## Test plan

- [ ] `Mago (PoC, advisory)` job runs on this PR
- [ ] Job does **not** fail even with thousands of existing findings
- [ ] Step summary shows rule-count tables for lint and analyze
- [ ] JSON reports appear as a downloadable artifact
- [ ] Inline annotations (file/line) appear on the PR "Files changed" tab

Assisted-by: Claude Code

### DIFF
--- a/.github/workflows/mago.yml
+++ b/.github/workflows/mago.yml
@@ -1,13 +1,8 @@
-name: Mago (PoC, advisory)
+name: Mago
 
 # Proof-of-concept workflow for Mago (https://mago.carthage.software/),
 # a Rust-based PHP toolchain that combines linting, static analysis, and
 # formatting in a single tool.
-#
-# This workflow is ADVISORY ONLY. Every step is wrapped in `continue-on-error`
-# so a Mago finding will never block a PR. The intent is to surface what Mago
-# would catch so the team can decide whether to adopt it, and (if so) on what
-# terms (subset of rules, baseline-gated, blocking, etc.).
 
 on:
   push:
@@ -28,7 +23,7 @@ concurrency:
 
 jobs:
   mago:
-    name: Mago PoC
+    name: Mago
     runs-on: ubuntu-24.04
     strategy:
       matrix:
@@ -53,29 +48,28 @@ jobs:
       with:
         version: '1.24.0'
 
+    # Every subsequent step uses `if: always()` so one failing tool does not
+    # short-circuit the rest — the job still fails, but we see all findings
+    # in a single run.
+
     - name: Mago format (check only)
-      id: format
-      continue-on-error: true
+      if: always()
       run: mago format --check
 
-    - name: Mago lint (GitHub annotations)
-      id: lint
-      continue-on-error: true
+    - name: Mago lint
+      if: always()
       run: mago lint --reporting-format=github
 
     - name: Mago lint (JSON for summary)
       if: always()
-      continue-on-error: true
       run: mago lint --reporting-format=json > mago-lint.json
 
-    - name: Mago analyze (GitHub annotations)
-      id: analyze
-      continue-on-error: true
+    - name: Mago analyze
+      if: always()
       run: mago analyze --reporting-format=github
 
     - name: Mago analyze (JSON for summary)
       if: always()
-      continue-on-error: true
       run: mago analyze --reporting-format=json > mago-analyze.json
 
     - name: Summarize findings
--- a/.github/workflows/mago.yml
+++ b/.github/workflows/mago.yml
@@ -1,0 +1,120 @@
+name: Mago (PoC, advisory)
+
+# Proof-of-concept workflow for Mago (https://mago.carthage.software/),
+# a Rust-based PHP toolchain that combines linting, static analysis, and
+# formatting in a single tool.
+#
+# This workflow is ADVISORY ONLY. Every step is wrapped in `continue-on-error`
+# so a Mago finding will never block a PR. The intent is to surface what Mago
+# would catch so the team can decide whether to adopt it, and (if so) on what
+# terms (subset of rules, baseline-gated, blocking, etc.).
+
+on:
+  push:
+    branches:
+    - master
+    - rel-*
+  pull_request:
+    branches:
+    - master
+    - rel-*
+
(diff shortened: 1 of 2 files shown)


kojiromike:
@openemr/admins @Firehed maybe a distraction, but I've been wanting to try this for a while -- it does some significant portion of what phpstan does, but waaaay faster. The main blocker to really useful adoption is that we can't write custom rules. (They do seem amenable to upstreaming rules, but I'm not sure how rapidly we'd be able to take advantage of that.)

kojiromike:
Mago is probably never going to be suitable for OpenEMR as long as it doesn't have any facility for customizing or extending rules. It just goes fast in the wrong direction.

cc @openemr/admins in case this comes up in the future.

Issue #10606, opened 2026-02-06
feat(vitals): add quick-edit from patient dashboard


kojiromike:
## Problem

Entering vitals requires navigating through multiple modules (patient > encounter > vitals form), which is time-consuming for a common workflow.

## Expected Behavior

Add quick-edit for vitals on patient dashboard:
1. Display most recent vitals summary
2. "Add Vitals" button opens modal or inline form
3. Quick entry without full encounter navigation
4. Option to create a quick encounter if none exists for today

## Design Considerations

- Should quick-edit create a new encounter or attach to existing?
- Consider a "Quick Vitals" encounter type
- May want as a configurable feature (enable/disable per practice)

## Source

Sunchu K, et al. "Lessons Learned from OpenEMR Implementation in Graduate Health Informatics Curriculum." AMIA 2024.
https://pmc.ncbi.nlm.nih.gov/articles/PMC12099383/

PR #11777, opened 2026-04-22
feat(vitals): quick vitals form

haileysnodgrass:
https://github.com/openemr/openemr/issues/10606 

This adds a "Quick Vitals Entry" button to the Medical Record Dashboard so providers can input weight, height, pulse, and BP via a quick pop-up, rather than navigating to a separate charting screen.

* **UI:** Added a trigger button to `vitals_fragment.php` and built an AJAX modal workflow in `demographics.php`.
* **Encounter Logic:** Added `quick_vitals_save.php` to smartly link vitals to a same-day encounter (or create a new one) while ignoring historical visits. 
* **Database:** Added `quick_save.php` to securely save the vitals and set the `activity`, `formdir`, and `deleted` flags so the dashboard immediately renders the new data.

Yes (Gemini)

Use the name of the specific tool (e.g. `GitHub Copilot`, `Claude Code`, `ChatGPT`).
If the AI made the commit(s), this trailer was probably added automatically.
-->
N/A

### DIFF
--- a/interface/forms/vitals/quick_edit.php
+++ b/interface/forms/vitals/quick_edit.php
@@ -1,0 +1,49 @@
+<?php
+require_once("../../globals.php");
+
+use OpenEMR\Common\Session\SessionWrapperFactory;
+use OpenEMR\Common\Csrf\CsrfUtils;
+
+$session = SessionWrapperFactory::getInstance()->getActiveSession();
+$pid = $_GET['pid'] ?? null;
+$encounter = $_GET['encounter'] ?? null;
+
+if (!$pid || !$encounter) {
+    die("Missing patient or encounter context.");
+}
+?>
+<div class="p-3">
+    <h4 class="mb-3">Quick Vitals Entry</h4>
+    <form id="quickVitalsForm">
+        <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken(session: $session)); ?>">
+        <input type="hidden" name="pid" value="<?php echo attr($pid); ?>">
+        <input type="hidden" name="encounter" value="<?php echo attr($encounter); ?>">
+
+        <div class="row">
+            <div class="col-6 mb-2">
+                <label>Weight (lbs)</label>
+                <input type="text" name="weight" class="form-control">
+            </div>
+            <div class="col-6 mb-2">
+                <label>Height (in)</label>
+                <input type="text" name="height" class="form-control">
+            </div>
+            <div class="col-6 mb-2">
+                <label>Pulse</label>
+                <input type="text" name="pulse" class="form-control">
+            </div>
+            <div class="col-6 mb-2">
+                <label>BP (Systolic/Diastolic)</label>
+                <div class="input-group">
+                    <input type="text" name="bps" placeholder="Sys" class="form-control">
+                    <input type="text" name="bpd" placeholder="Dia" class="form-control">
+                </div>
+            </div>
+        </div>
+
+        <div class="mt-4 text-right">
+            <button type="button" class="btn btn-secondary btn-cancel-vitals">Cancel</button>
+            <button type="button" id="saveQuickVitals" class="btn btn-primary">Save Vitals</button>
+        </div>
+    </form>
+</div>
--- a/interface/forms/vitals/quick_save.php
+++ b/interface/forms/vitals/quick_save.php
@@ -1,0 +1,51 @@
+<?php
+require_once("../../globals.php");
+
+use OpenEMR\Common\Csrf\CsrfUtils;
+
+header('Content-Type: application/json');
+
+try {
+    CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
+
+    $pid = $_POST['pid'] ?? null;
+    $encounter = $_POST['encounter'] ?? null;
+
+    if (empty($pid) || empty($encounter)) {
+        echo json_encode(["success" => false, "error" => "Missing Context"]);
+        exit;
+    }
+
+    $date = date("Y-m-d H:i:s");
+    
+    // Parse values safely to prevent MySQL strict mode errors (numeric fields need NULL, not empty strings)
+    $weight = !empty($_POST['weight']) ? (float)$_POST['weight'] : null;
+    $height = !empty($_POST['height']) ? (float)$_POST['height'] : null;
+    $pulse  = !empty($_POST['pulse'])  ? (int)$_POST['pulse']   : null;
+    $bps    = !empty($_POST['bps'])    ? (int)$_POST['bps']     : null;
+    $bpd    = !empty($_POST['bpd'])    ? (int)$_POST['bpd']     : null;
+    
+    // 1. Insert into form_vitals. CRITICAL: activity MUST be 1 for the report to display it.
+    $v_sql = "INSERT INTO form_vitals (date, pid, activity, weight, height, pulse, bps, bpd) 
+              VALUES (?, ?, 1, ?, ?, ?, ?, ?)";
+    
+    $v_id = sqlInsert($v_sql, [$date, $pid, $weight, $height, $pulse, $bps, $bpd]);
+
+    // 2. Register the form to the encounter
+    $f_sql = "INSERT INTO forms (date, encounter, form_name, form_id, pid, user, groupname, authorized, deleted, formdir) 
+              VALUES (?, ?, 'Vitals', ?, ?, ?, ?, 1, 0, 'vitals')";
+    
+    sqlStatement($f_sql, [
+        $date, 
+        $encounter, 
+        $v_id, 
+        $pid, 
+        $_SESSION['authUser'] ?? 'admin', 
+        $_SESSION['authProvider'] ?? 'Default'
+    ]);
+
+    echo json_encode(["success" => true]);
+} catch (Throwable $e) {
+    echo json_encode(["success" => false, "error" => $e->getMessage()]);
+}
+exit;
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -626,7 +626,65 @@ $oemr_ui = new OemrUI($arrOeUiSettings);
             placeHtml("track_anything_fragment.php", "track_anything_ps_expand");
             <?php if ($vitals_is_registered && AclMain::aclCheckCore('patients', 'med')) { ?>
             // Initialize the Vitals form if it is registered and user is authorized.
-            placeHtml("vitals_fragment.php", "vitals_ps_expand");
+            placeHtml("vitals_fragment.php", "vitals_ps_expand").then(() => {
+                // Handler for opening the modal
+                $(document).off('click.quickVitals').on('click.quickVitals', '.quick-vitals-btn', function () {
+                    let pid = $(this).data('pid') || top.pid;
+                    let encounter = $(this).data('encounter') || top.encounter;
+
+                    $.post("quick_vitals_save.php", {
+                        pid: pid,
+                        encounter: encounter,
+                        csrf_token_form: <?php echo js_escape(CsrfUtils::collectCsrfToken(session: $session)); ?>
+                    }, function (res) {
+                        if (res.success) {
+                            $.get("/interface/forms/vitals/quick_edit.php", { pid: pid, encounter: res.encounter }, function (html) {
+                                $("#quickVitalsModal, .quick-vitals-backdrop").remove();
+                                $("body").append(`
+                                    <div id="quickVitalsModal" style="display:block; position:fixed; top:50%; left:50%; transform:translate(-50%, -50%); z-index:3000; background:#fff; width:500px; padding:20px; border-radius:8px; box-shadow: 0 0 20px rgba(0,0,0,0.4);">
+                                        ${html}
+                                    </div>
+                                    <div class="modal-backdrop fade show quick-vitals-backdrop" style="z-index:2999; background: rgba(0,0,0,0.5); position:fixed; top:0; left:0; width:100%; height:100%;"></div>
+                                `);
+                            });
+                        } else {
+                            alert("App Error: " + res.error);
+                        }
+                    }, "json").fail(function(xhr) {
+                        console.error("Quick Vitals initialization failed:", xhr.responseText);
+                        alert("An error occurred while loading the vitals form.");
+                    });
+                });
+
+                // 2. SAVE HANDLER
+                $(document).off('click.vitalsSave').on('click.vitalsSave', '#saveQuickVitals', function () {
+                    $.ajax({
+                        url: "/interface/forms/vitals/quick_save.php",
+                        type: "POST",
+                        data: $('#quickVitalsForm').serialize(),
+                        dataType: "json",
+                        success: function (res) {
+                            if (res.success) {
+                                $("#quickVitalsModal, .quick-vitals-backdrop").remove();
+                                // Reload the fragment to show the new data
+                                placeHtml("vitals_fragment.php", "vitals_ps_expand");
+                            } else {
+                                alert("Save Failed: " + res.error);
+                            }
+                        },
+                        error: function (xhr) {
+                            $("#quickVitalsModal, .quick-vitals-backdrop").remove();
+                            console.error("Quick Vitals save failed:", xhr.responseText);
+                            alert("An error occurred while saving vitals. Please check your connection.");
+                        }
+                    });
+                });
+
+                // 3. CANCEL HANDLER
+                $(document).on('click', '.btn-cancel-vitals', function() {
+                    $("#quickVitalsModal, .quick-vitals-backdrop").remove();
+                });
+            });
             <?php } ?>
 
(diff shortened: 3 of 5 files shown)


luisuriarte:
It's very interesting, here in South America, in addition to those values, temperature is also very common.

interface/forms/vitals/quick_save.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
kojiromike, line 1: Two top-level concerns for this file:

**1. Missing ACL check.** This endpoint writes to `form_vitals` and `forms` for any `pid` it's given, but never verifies the caller has access to that patient. The button in `vitals_fragment.php` is shown only when `AclMain::aclCheckCore('patients', 'med')` passes, but that's UI gating — it doesn't stop a direct POST. Add the check at the top of the endpoint:

```suggestion
<?php

declare(strict_types=1);

require_once("../../globals.php");

use OpenEMR\Common\Acl\AclMain;
use OpenEMR\Common\Csrf\CsrfUtils;

header('Content-Type: application/json');

if (!AclMain::aclCheckCore('patients', 'med')) {
    http_response_code(403);
    echo json_encode(['success' => false, 'error' => 'Forbidden']);
    exit;
}
```

**2. `declare(strict_types=1);` is required for new PHP code per the repo's `CLAUDE.md` — included in the suggestion above.**

interface/forms/vitals/quick_save.php
```diff
@@ -0,0 +1,51 @@
...
+require_once("../../globals.php");
+
+use OpenEMR\Common\Csrf\CsrfUtils;
+
+header('Content-Type: application/json');
+
+try {
+    CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
+
+    $pid = $_POST['pid'] ?? null;
```
kojiromike, line 11: Three things going on here and on the next few lines:

- **Raw `$_POST` access.** The repo standards expect `filter_input(INPUT_POST, 'pid', FILTER_VALIDATE_INT)` so the value is type-narrowed and rejected if non-integer.
- **`empty()` is forbidden in new code** (line 14) — `empty('0')` is `true`, which is rarely what you want. Use explicit `=== null || === ''` checks, or rely on `filter_input`'s return value.
- **The `(pid, encounter)` pair isn't validated to belong together.** A caller can submit `pid=A&encounter=<encounter-belonging-to-B>` and attach vitals to the wrong patient's encounter. Re-check that `encounter` belongs to `pid` before inserting (`quick_vitals_save.php` does this — `quick_save.php` should too).

```suggestion
    $pid = filter_input(INPUT_POST, 'pid', FILTER_VALIDATE_INT);
    $encounter = filter_input(INPUT_POST, 'encounter', FILTER_VALIDATE_INT);

    if ($pid === false || $pid === null || $encounter === false || $encounter === null) {
        echo json_encode(['success' => false, 'error' => 'Missing or invalid context']);
        exit;
    }

    // Verify the encounter actually belongs to this patient before writing to it.
    $owner = sqlQuery(
        'SELECT encounter FROM form_encounter WHERE pid = ? AND encounter = ?',
        [$pid, $encounter]
    );
    if (empty($owner['encounter'])) {
        echo json_encode(['success' => false, 'error' => 'Encounter does not belong to patient']);
        exit;
    }
```

interface/forms/vitals/quick_save.php
```diff
@@ -0,0 +1,51 @@
...
+
+    if (empty($pid) || empty($encounter)) {
+        echo json_encode(["success" => false, "error" => "Missing Context"]);
+        exit;
+    }
+
+    $date = date("Y-m-d H:i:s");
+    
+    // Parse values safely to prevent MySQL strict mode errors (numeric fields need NULL, not empty strings)
+    $weight = !empty($_POST['weight']) ? (float)$_POST['weight'] : null;
```
kojiromike, line 22: No range validation on clinical values. `(float)$_POST['weight']` accepts `"abc"` (silently coerced to `0`), negative numbers, and absurdly large values. At minimum, reject negative and obviously-out-of-range numbers — a 0-lb weight or a 999-pulse value is clinical garbage and shouldn't make it into the chart.

Also, `!empty($x) ? (float)$x : null` will treat the legitimate string `"0"` as null. Prefer a parse function that distinguishes "absent" from "zero" — e.g. via `filter_input(..., FILTER_VALIDATE_FLOAT)`.

interface/forms/vitals/quick_save.php
```diff
@@ -0,0 +1,51 @@
...
+    $bpd    = !empty($_POST['bpd'])    ? (int)$_POST['bpd']     : null;
+    
+    // 1. Insert into form_vitals. CRITICAL: activity MUST be 1 for the report to display it.
+    $v_sql = "INSERT INTO form_vitals (date, pid, activity, weight, height, pulse, bps, bpd) 
+              VALUES (?, ?, 1, ?, ?, ?, ?, ?)";
+    
+    $v_id = sqlInsert($v_sql, [$date, $pid, $weight, $height, $pulse, $bps, $bpd]);
+
+    // 2. Register the form to the encounter
+    $f_sql = "INSERT INTO forms (date, encounter, form_name, form_id, pid, user, groupname, authorized, deleted, formdir)
```
kojiromike, line 35: Bypassing the canonical form-registration path. The `addForm()` helper in `library/api.inc.php` is what every other form uses — it sets defaults, looks up `groupname`, and stays consistent with the rest of the codebase (which means downstream code that walks the `forms` table behaves predictably). Hand-rolling the INSERT means future schema changes to `forms` won't be picked up here.

Replace this block with something like:

```php
addForm($encounter, 'Vitals', $v_id, 'vitals', $pid, $userauthorized ?? 1);
```

Also, on line 43–44: `$_SESSION['authUser'] ?? 'admin'` and `?? 'Default'` should not exist. If the session isn't authenticated, the request should already have been rejected (see the ACL note at the top of this file). Silently attributing writes to `admin` makes the audit log lie.

interface/forms/vitals/quick_save.php
```diff
@@ -0,0 +1,51 @@
...
+        $encounter, 
+        $v_id, 
+        $pid, 
+        $_SESSION['authUser'] ?? 'admin', 
+        $_SESSION['authProvider'] ?? 'Default'
+    ]);
+
+    echo json_encode(["success" => true]);
+} catch (Throwable $e) {
+    echo json_encode(["success" => false, "error" => $e->getMessage()]);
```
kojiromike, line 49: Per the repo's `CLAUDE.md`: **"Never expose `$e->getMessage()` in user-facing output."** Exception messages can include SQL fragments, file paths, and class names. Log it instead and return something generic:

```suggestion
} catch (\Throwable $e) {
    (new \OpenEMR\Common\Logging\SystemLogger())->error('Quick vitals save failed', ['exception' => $e]);
    echo json_encode(['success' => false, 'error' => 'Save failed']);
}
```

interface/patient_file/summary/quick_vitals_save.php
```diff
@@ -0,0 +1,55 @@
...
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Session\SessionWrapperFactory;
+
+header('Content-Type: application/json');
+
+try {
+    $session = SessionWrapperFactory::getInstance()->getActiveSession();
+    CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
+
+    $pid = $_POST['pid'] ?? null;
```
kojiromike, line 14: Same three top-level issues as `quick_save.php`:

1. **Missing ACL check** — add `AclMain::aclCheckCore('patients', 'med')` at the top of this endpoint too.
2. **Missing `declare(strict_types=1);`**.
3. **Raw `$_POST` / `empty()`** — use `filter_input(INPUT_POST, 'pid', FILTER_VALIDATE_INT)` and an explicit null/false check instead of `empty($pid)`.

interface/patient_file/summary/quick_vitals_save.php
```diff
@@ -0,0 +1,55 @@
...
+        $row = sqlQuery("SELECT encounter FROM form_encounter WHERE pid = ? AND DATE(date) = CURDATE() ORDER BY date DESC LIMIT 1", [$pid]);
+        if (!empty($row['encounter'])) {
+            $encounter = $row['encounter'];
+        }
+    }
+
+    // 3. STILL empty? That means absolutely no visits today. Create a brand new one!
+    if (empty($encounter)) {
+        $enc_row = sqlQuery("SELECT MAX(encounter) AS max_enc FROM form_encounter");
+        $encounter = ($enc_row['max_enc'] ?? 0) + 1;
```
kojiromike, line 43: **Race condition.** Two concurrent requests will both see the same `MAX(encounter)` and try to insert the same encounter id — at best one fails, at worst they collide silently and vitals attach to the wrong row.

OpenEMR has a `sequences` table for this. The pattern used elsewhere in the codebase (e.g. `interface/forms/newpatient/new.php`) is roughly:

```php
$encounter = sqlInsert("INSERT INTO sequences (id) VALUES (NULL)");
```

That returns a guaranteed-unique encounter id without the read-then-write window. Please use it.

Separately, the `INSERT INTO form_encounter` on line 47 only sets `pid, encounter, date, reason, facility_id, sensitivity` — the existing encounter-creation path also sets `provider_id`, `pc_catid`, `billing_facility`, `class_code`, etc. Some of those have NOT NULL or sensible-default expectations downstream (billing reports, provider attribution). Look at what `new.php` writes and match it.

interface/patient_file/summary/quick_vitals_save.php
```diff
@@ -0,0 +1,55 @@
...
+        
+        $facility = $_SESSION['login_facility'] ?? 1;
+        
+        sqlStatement("INSERT INTO form_encounter (pid, encounter, date, reason, facility_id, sensitivity) VALUES (?, ?, NOW(), 'Quick Vitals', ?, 'normal')", [$pid, $encounter, $facility]);
+    }
+
+    echo json_encode(["success" => true, "encounter" => (int)$encounter]);
+
+} catch (Throwable $e) {
+    echo json_encode(["success" => false, "error" => $e->getMessage()]);
```
kojiromike, line 53: Same as `quick_save.php`: don't expose `$e->getMessage()` to the client. Log the exception and return a generic message — see the inline comment on `quick_save.php` for the suggested shape.

interface/forms/vitals/quick_edit.php
```diff
@@ -0,0 +1,49 @@
+<?php
+require_once("../../globals.php");
+
+use OpenEMR\Common\Session\SessionWrapperFactory;
+use OpenEMR\Common\Csrf\CsrfUtils;
+
+$session = SessionWrapperFactory::getInstance()->getActiveSession();
+$pid = $_GET['pid'] ?? null;
```
kojiromike, line 8: Same standards items as the save endpoints: this file is missing `declare(strict_types=1);`, uses raw `$_GET` (should be `filter_input(INPUT_GET, 'pid', FILTER_VALIDATE_INT)` etc.), and `die("Missing patient or encounter context.")` on line 12 emits plain text into what the JS expects to render as HTML inside the modal — better to return an error fragment or to validate `pid`/`encounter` server-side with a proper error response.

interface/forms/vitals/quick_edit.php
```diff
@@ -0,0 +1,49 @@
...
+<div class="p-3">
+    <h4 class="mb-3">Quick Vitals Entry</h4>
+    <form id="quickVitalsForm">
+        <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken(session: $session)); ?>">
+        <input type="hidden" name="pid" value="<?php echo attr($pid); ?>">
+        <input type="hidden" name="encounter" value="<?php echo attr($encounter); ?>">
+
+        <div class="row">
+            <div class="col-6 mb-2">
+                <label>Weight (lbs)</label>
```
kojiromike, line 24: Two issues for the form labels (applies to lines 24, 28, 32, 36, 45, 46 as well):

1. **Untranslated.** The button on `vitals_fragment.php` uses `xlt('Enter Quick Vitals')` — every label inside this form should be `xlt()`'d too. As is, this UI is English-only.
2. **Hardcoded units.** OpenEMR has a global units setting (lbs/in vs. kg/cm) and the existing vitals form respects it. Quick entry should read the same setting and render the appropriate unit, otherwise users in metric-unit practices will see the wrong label and submit values in the wrong unit.

For the markup itself, also consider `<input type="number" min="0" step="0.01">` instead of `type="text"` so the browser does basic validation.

interface/patient_file/summary/demographics.php
```diff
@@ -626,7 +626,65 @@ function load_location(location) {
...
+
+                    $.post("quick_vitals_save.php", {
+                        pid: pid,
+                        encounter: encounter,
+                        csrf_token_form: <?php echo js_escape(CsrfUtils::collectCsrfToken(session: $session)); ?>
+                    }, function (res) {
+                        if (res.success) {
+                            $.get("/interface/forms/vitals/quick_edit.php", { pid: pid, encounter: res.encounter }, function (html) {
+                                $("#quickVitalsModal, .quick-vitals-backdrop").remove();
+                                $("body").append(`
```
kojiromike, line 643: Two thoughts on the modal construction:

1. **Inline-styled template-literal modal.** The codebase already uses Bootstrap modals (Bootstrap 4.6 per the project tech stack) — using `<div class="modal" tabindex="-1">...</div>` with the standard markup gives you focus trapping, ESC-to-close, and accessibility for free. The hand-rolled `<div style="position:fixed; top:50%; left:50%;...">` reinvents this and skips the a11y behavior.
2. **Hardcoded URL** (`/interface/forms/vitals/quick_edit.php` here, `/interface/forms/vitals/quick_save.php` on line 662). Build these from `webroot_url` so non-root deployments still work.

Also, on the AJAX flow itself: this calls `quick_vitals_save.php` *just to allocate/find an encounter*, and then opens the modal — so you've already created an encounter even if the user clicks Cancel. Consider deferring encounter creation until the actual save.

interface/patient_file/summary/vitals_fragment.php
```diff
@@ -46,3 +45,6 @@
   </span><?php
 } ?>
 </div>
+<button class="btn btn-primary quick-vitals-btn" data-pid="<?php echo attr($pid); ?>" data-encounter="<?php echo attr($session->get('encounter') ?? ''); ?>">
```
kojiromike, line 48: Two things on the new button:

1. **Defense-in-depth ACL gate.** The button is currently shown unconditionally to anyone who can render `vitals_fragment.php`. The caller (`demographics.php`) already gates the inclusion of this fragment behind `AclMain::aclCheckCore('patients', 'med')`, but if this fragment is ever rendered from another path, the button will be visible without that check. Wrapping the button in the same ACL check makes it self-contained.
2. **Missing trailing newline** (line 50 — `\ No newline at end of file`). Same on all four new files. Per PSR-12 / PER-CS, files should end with a single LF.

kojiromike (commented):
Thanks for the contribution — the workflow target (one-click vitals from the dashboard) is a real ergonomic improvement.

I've left inline comments on the specific issues. The biggest blockers, before this can be considered safe to merge, are:

1. **No authorization check** on either save endpoint — any authenticated user can write vitals/encounters for any pid by calling them directly. The fragment's `AclMain::aclCheckCore('patients', 'med')` gate in `demographics.php` only hides the button; it doesn't protect the endpoints.
2. **Encounter-id race** — `MAX(encounter)+1` is not safe under concurrent requests. OpenEMR has a `sequences` table for this; see how `interface/forms/newpatient/new.php` allocates encounter ids.
3. **Exception messages echoed to the client** — `$e->getMessage()` can contain SQL fragments and file paths. The repo's `CLAUDE.md` is explicit about this: log the exception, return a generic message.
4. **Hand-rolled `forms` INSERT** — `library/api.inc.php::addForm()` is the canonical helper. Bypassing it skips defaults and the silent `$_SESSION['authUser'] ?? 'admin'` fallback masks an unauthenticated request as the admin user.

Beyond those, several items from the repo's `CLAUDE.md` aren't followed in the new files: missing `declare(strict_types=1);`, use of `empty()` (which makes `'0'` falsy — surprising for medical values), raw `$_POST` / `$_GET` access instead of `filter_input(...)`, and `sqlQuery` / `sqlStatement` instead of `QueryUtils`. None of those are hard to fix once you've seen one example, but they should be fixed throughout.

A couple of design questions worth thinking about (raised in inline comments): unit handling (the existing vitals form respects a global units setting), translation of the form labels (the button is `xlt`-wrapped, but the inputs aren't), and the modal markup (an inline-styled template literal vs. a real Bootstrap modal). And `@luisuriarte`'s comment about temperature is a good point — the existing vitals form covers it.

Two small process notes:
- The PR is opened from your fork's `master` branch — for future PRs, please make a feature branch (`git checkout -b feat/quick-vitals`) so `master` stays clean to track upstream and follow-up PRs are easier.
- Issue #10606 also asks for a per-practice toggle (configurable enable/disable) and an option to attach to existing vs. new encounter — those aren't covered yet, but they're orthogonal and could land separately.

Happy to look again once the security and correctness items are addressed.

kojiromike (changes requested):

PR #11779, opened 2026-04-23
refactor(filesystem): Move Smarty and mPDF temp directories to system temp

Firehed:
## Summary

- Move Smarty compile directory from `sites/{site}/documents/smarty/` to `sys_get_temp_dir()/openemr-smarty`
- Move mPDF temp directory from `sites/{site}/documents/mpdf/pdf_tmp` to `sys_get_temp_dir()/openemr-mpdf`
- Remove `MPDF_WRITE_DIR` global — temp dir is now resolved at interaction points
- Remove unnecessary directory creation code from `globals.php` (both libraries auto-create, except Smarty_Legacy which now handles it)
- Consolidate Smarty compile directories (GACL/main separation was unnecessary — UIDs are path-based)
- Delete obsolete `documents/smarty/` and `documents/mpdf/` directories

## Rationale

The `documents/` directory is intended for patient documents and persistent data, not cache/temp files. Moving these to system temp:
- Keeps `documents/` clean and focused on its purpose
- Allows system-level temp cleanup policies to apply
- Removes global state dependency for temp paths

## Test plan

- [x] Open the calendar (exercises Smarty_Legacy/pcSmarty)
- [x] Navigate to Administration → Users → ACL Administration (exercises GACL Smarty)
- [ ] Open a document via Files (exercises Controller.class.php Smarty)
- [x] Print a patient report as PDF (exercises mPDF)


Firehed:
@aisle-analyzer

Firehed:
@kojiromike I think the bot feedback is reasonable, but it's also basically the same as before in practice so we're no worse-off. The per-site comment is pointless since multi-site is probably going to go away and in any case next-to-nobody uses it so everything would be under `default` anyway. I'd have preferred to plug a temp directory into flysystem but the libraries in question aren't really compatible with this approach. Thoughts?

Firehed:
@aisle-analyzer

Firehed:
@aisle-analysis

Issue #11792, opened 2026-04-23
Block fatal-category errors from PHPStan baseline


kojiromike:
## Problem

PHPStan's baseline currently grandfathers thousands of errors that represent code which cannot run. These are not style issues or type-inference quirks — they are fatal at load or call time:

- `class.notFound`, `method.notFound`, `staticMethod.notFound`
- `function.notFound`
- `classConstant.notFound`, `constant.notFound`
- `include.fileNotFound`, `includeOnce.fileNotFound`, `requireOnce.fileNotFound`
- `return.missing`, `variable.undefined` (fatal under strict error reporting)

Allowing these into the baseline means a developer can land code that crashes the moment its call path is exercised. The baseline should never accept a new one.

## Current baseline counts

| Identifier | Original | Current | Drained |
|---|---:|---:|---:|
| `variable.undefined` | 3457 | 3064 | -393 |
| `class.notFound` | 255 | 184 | -71 |
| `method.notFound` | 253 | 141 | -112 |
| `constant.notFound` | 125 | 0 | ✅ -125 |
| `return.missing` | 29 | 0 | ✅ -29 |
| `function.notFound` | 13 | 0 | ✅ -13 |
| `include.fileNotFound` | 4 | 0 | ✅ -4 |
| `requireOnce.fileNotFound` | 4 | 0 | ✅ -4 |
| `includeOnce.fileNotFound` | 1 | 0 | ✅ -1 |
| `staticMethod.notFound` | 0 | 0 | — |
| `classConstant.notFound` | 0 | 0 | — |
| **Total** | **4141** | **3389** | **-752** |

Six categories now permanently locked at zero. Phase 2 complete; Phase 3 complete; Phase 4 in progress; Phase 5 in progress.

### Phase 1 — Gate (first PR) ✅

Add a check that fails CI whenever the count of entries in any of the above baseline files exceeds a committed cap. Caps start at today's counts and can only go down. New errors in these categories must be fixed, not baselined.

- Commit a caps file listing per-identifier counts.
- Add an isolated PHPUnit test (or composer script + CI step) that reads the baseline files, counts entries, and asserts each is ≤ its cap.
- Document the workflow: when you regenerate the baseline, either decrement the cap (you fixed some) or the test fails (you added a new one — fix it instead of baselining it).

After Phase 1, the two already-zero categories (`staticMethod.notFound`, `classConstant.notFound`) are permanently locked at zero.

### Phase 2 — Drain the small categories ✅

[...]

PR #11802, opened 2026-04-23
fix(phpstan): drain file-not-found baseline entries

kojiromike:
## Summary

Phase 2 of openemr/openemr#11792 — drive the three file-not-found baseline files to empty and zero their caps.

All four entries across `.phpstan/baseline/{include,includeOnce,requireOnce}.fileNotFound.php` are fixed at the source:

| File | Fix |
| --- | --- |
| `interface/main/calendar/includes/pnHTML.php` | Drop `include 'header.php'` / `include 'footer.php'`; the PostNuke theme files these pointed at were removed long ago. |
| `library/ESign/DbRow/Signable.php` | Delete unused `renderForm()` method; it's never called, and its include path has never been correct. |
| `portal/patient/fwk/libs/verysimple/Phreeze/ExportUtility.php` | Delete unused `OutputAsExcel()` method that required PEAR's PHPExcel (never installed). |
| `portal/patient/fwk/libs/verysimple/Phreeze/PortalController.php` | Delete unused `RenderRSS()` method that required `verysimple/RSS/*` (never installed). |
| `interface/patient_file/summary/demographics_print.php`, `interface/patient_file/printed_fee_sheet.php` | Resolve the optional `custom/*.php` hook paths through `$webserver_root` so PHPStan no longer sees a literal path to a file that doesn't exist in a fresh checkout. Matches the pattern already used in `interface/reports/criteria.tab.php`. |
| `portal/patient/templates/PatientListView.tpl.php` | Fix broken `require_once("../../library/options.inc.php")` to use `$globalsBag->getSrcDir() . '/options.inc.php'`, the same pattern already used earlier in the file. |

Caps in `.phpstan/fatal-baseline-caps.php`:

- `include.fileNotFound.php`: 4 → 0 (file deleted)
- `includeOnce.fileNotFound.php`: 1 → 0 (file deleted)
- `requireOnce.fileNotFound.php`: 4 → 0 (file deleted)
- `class.notFound.php`: 255 → 237 (drive-by: the two deleted Phreeze methods referenced 18 classes that were never autoloadable)

## Test plan

- [x] `composer phpunit-isolated -- --filter FatalBaselineCaps` → 19/19 green
- [x] `composer phpstan-baseline` regenerates clean (no new baseline entries introduced)
- [x] Pre-commit hooks pass (phpstan, phpcs, rector, codespell, composer-require-checker)


.phpstan/baseline/variable.undefined.php
kojiromike, line 1: We should not be increasing these numbers.

interface/patient_file/summary/demographics_print.php
```diff
@@ -18,12 +18,13 @@
...
+$customDemographicsPrint = "$webserver_root/custom/demographics_print.php";
 if (
     !empty(\OpenEMR\Core\OEGlobalsBag::getInstance()->get('gbl_rapid_workflow')) &&
     \OpenEMR\Core\OEGlobalsBag::getInstance()->get('gbl_rapid_workflow') == 'LBFmsivd' &&
-    file_exists('../../../custom/demographics_print.php')
+    file_exists($customDemographicsPrint)
 ) {
-    include('../../../custom/demographics_print.php');
+    include($customDemographicsPrint);
     exit();
```
kojiromike: This should probably be `return` not `exit()`;

portal/patient/templates/PatientListView.tpl.php
```diff
@@ -15,7 +15,7 @@
 
     $globalsBag = OEGlobalsBag::getInstance();
     //require_once ("./../verify_session.php");
-    require_once("../../library/options.inc.php");
+    require_once($globalsBag->getString('srcdir') . "/options.inc.php");
```
kojiromike, line 18: srcdir is {webroot}/library? That doesn't seem right.

```
$ find . -name options.inc.php
./library/options.inc.php
```

.phpstan/baseline/variable.undefined.php
kojiromike: Agreed — this was an unintended regression from the first push. The original diff used `"$webserver_root/..."` (an encapsed string over an untyped global), which bumped both `variable.undefined` and `encapsedStringPart.nonString` counts by 1 in each of the two files.

Fixed in 73247df — switched to `OEGlobalsBag::getInstance()->getProjectDir()` (typed `string` getter, returns the same value as `$webserver_root`). Regenerated baseline now shows those counts back down to their pre-PR values; no net increase anywhere.

interface/patient_file/summary/demographics_print.php
kojiromike: Good call — changed to `return` in 73247df. This file is always the top-level entry point (directly executed by the HTTP router), so `return` cleanly terminates it without the `exit()` sledgehammer.

portal/patient/templates/PatientListView.tpl.php
kojiromike: `srcdir` really is `{webserver_root}/library` — set in `interface/globals.php:323`:

```php
$globalsBag->set('srcdir', "$webserver_root/library");
```

So `$globalsBag->getSrcDir() . '/options.inc.php'` resolves to `{webserver_root}/library/options.inc.php`, which is the same path `find . -name options.inc.php` turned up. It's the same `$srcdir/options.inc.php` idiom already used 3 lines away in `demographics_print.php:30` and in dozens of other spots.

The old path in this file was `"../../library/options.inc.php"` — from `portal/patient/templates/` that resolved to `portal/library/options.inc.php`, which doesn't exist. That's what landed it in the `requireOnce.fileNotFound` baseline in the first place.

PR #11805, opened 2026-04-23
fix(ci): propagate phpunit exit code from inferno tests

Firehed:

Fixes inferno certification tests silently passing when PHPUnit reports errors/failures.

- Capture phpunit exit code explicitly in `run_testsuite()` with `|| exit_code=$?`
- Capture `run_testsuite` exit code in `main()` before any conditional check
- Removes use of `if ! command; then local x=$?` pattern which loses the exit code (the `!` negation makes `$?` become 0)

The `ENABLE_COVERAGE` conditional added in 5b132a761 (Oct 2025) wrapped the phpunit call in an `if` block, which defeats `set -e`. Additionally, `main()` used `if ! run_testsuite` which also swallowed the exit code. The tests have been silently failing for ~6 months.


ci/inferno/run.sh
```diff
@@ -212,9 +214,9 @@ main() {
     fix_redis_permissions
 
     # Run the test suite and capture exit code
-    # shellcheck disable=SC2310
-    if ! run_testsuite; then
-        local exit_code=$?
+    local exit_code=0
+    run_testsuite || exit_code=$?
```
kojiromike, line 237: The three new SC2310 hits (here plus the two `phpunit` calls inside `run_testsuite`) are what's failing ShellCheck CI. SC2310 is pointing out that `set -e` is disabled inside a function called via `||`.

Two ways to clear it:

**Explicit errexit toggle** (no directive, intent self-documenting):

```bash
    local exit_code=0
    set +e
    run_testsuite
    exit_code=$?
    set -e
    if (( exit_code != 0 )); then
```

**`# shellcheck disable=SC2310`** on each site with a short rationale — consistent with the disable that used to live at line 215.

I'd normally suggest just the shellcheck disable, but the gist of this PR seems to be disabling set -e everywhere. Might be the right choice here.

ci/inferno/run.sh
```diff
@@ -115,18 +115,20 @@ run_testsuite() {
...
+                -c "${OPENEMR_DIR}/phpunit.xml" || exit_code=$?
     else
         phpunit --testsuite certification \
                 --log-junit junit-inferno.xml \
-                -c "${OPENEMR_DIR}/phpunit.xml"
+                -c "${OPENEMR_DIR}/phpunit.xml" || exit_code=$?
     fi
 
     echo 'Certification Tests Executed'
+    return "${exit_code}"
```
kojiromike, line 162: With `run_testsuite` always invoked via `|| exit_code=$?`, errexit is fully disabled throughout this function. If `. ci/ciLibrary.source` or `cd -P "${repo_root}"` ever fails, control falls through to `phpunit` and *its* exit code gets returned, masking the real setup failure.

Probably not a real risk, but worth either a line comment clarifying that this function captures the phpunit exit code specifically (not the function's overall health), or explicit `|| return $?` on the preamble commands.

kojiromike:
Once again I find confirmation of my theory that any shell script over 12 lines long should be converted to a real programming language.

Firehed:
Agreed. I'd like to peel this out to bare metal like I'm starting with some of the other test suites and then the need for this goes away.

@adunsulag this is a working starting point to get CI actually failing when the tests fail again; if you want to look into it further, I'd more than welcome your input! I've got a lot of things going on right now so probably not enough bandwidth to get it fully passing again (but I'll spend some claude points on it as I have time)

ci/inferno/run.sh
Firehed: yep, the `set -e` surprise is evidently why this cropped up in the first place. I didn't look super closely at _how_ claude did it yet.

PR #3593, opened 2020-06-03
And then there was only one (portal)

bradymiller:
fixes #3571

Just wanted to see what it would entail to remove the offsite and wordpress portals. Still gotta test this. And still gotta make sure community agrees with this.

PR #11813, opened 2026-04-24
fix(phpstan): drain function-not-found baseline entries

kojiromike:
## Summary

Phase 2 slice of the PHPStan fatal-baseline drain (openemr/openemr#11792). Drives the `function.notFound` cap from 13 to 0 by fixing every entry at the source. No new suppressions, no baseline additions.

### Dead code removed

- `gacl/Cache_Lite/Lite.php`: drop `get_magic_quotes_runtime`/`set_magic_quotes_runtime` calls — both removed in PHP 7, so these have always been fatal on any supported runtime.
- `interface/forms/LBF/new.php`, `interface/orders/single_order_results.php`, `interface/patient_file/report/custom_report.php`: remove the legacy CMS portal integration (`cms_portal_call`, `cms_field_to_lbf`, and their surrounding blocks). The portal backend `interface/cmsportal/portal.inc.php` was deleted in #3593 (2020); the callers were left stranded.
- `interface/main/backup.php`: drop the `gzopen64` → `gzopen` shim (`gzopen64` never exists in modern PHP).
- `interface/main/messages/save.php`: remove the unreachable `updateMessage(...)` branch — nothing defines that function and no caller POSTs `go=Messages`.
- `EncounterccdadispatchTable.php`: empty out the broken `display_layout_rows_group_new` branch to match the adjacent WIP-stub branch (the helper function it called does not exist).
- `library/edihistory/test_edih_sftp_files.php`: delete the entire WIP stub. It is gated behind an env var, references the non-existent `edih_sftp_connect`, and has no callers. Deleting it also cleaned up 61 unrelated baseline entries across 19 files.

### Symbol renames to the modern API

- `library/FeeSheet.class.php`:
  - `updateInvoiceRefNumber()` → `BillingUtilities::updateInvoiceRefNumber()`
  - `acl_check(...)` → `AclMain::aclCheckCore(...)` (the first half of the same expression already used `AclMain::aclCheckCore`)
- `library/classes/NumberToText.class.php`: replace the bare-function `NumberToText(...)` recursion with `(new self(...))->convert()`.

### Guarded dynamic include

- `interface/patient_file/pos_checkout_ippf.php`: `generateCheckoutReceipt()` is supplied by a site-configured custom receipt include (`$gbl_custom_receipt`). Added a `function_exists()` guard with a clear `die()` message so a missing/misconfigured receipt file fails fast instead of producing a PHP fatal.

### Cap updates

`.phpstan/fatal-baseline-caps.php`:

- `function.notFound.php`: 13 → 0 (baseline file deleted)
- `constant.notFound.php`: 125 → 123 (drive-by from CMS portal block removal)
- `variable.undefined.php`: 3457 → 3446 (drive-by)

## Test plan

- [x] `composer phpunit-isolated -- --filter FatalBaselineCaps` — 19/19 passing
- [x] `composer phpstan-baseline` regenerated cleanly (emits "Deleting baseline file … function.notFound.php")
- [x] `composer php-syntax-check` (via pre-commit) on all modified files
- [x] `composer phpcs`, `composer phpstan`, rector, require-checker (via pre-commit)


gacl/Cache_Lite/Lite.php
```diff
@@ -530,14 +530,11 @@ function _read()
...
-            $mqr = get_magic_quotes_runtime();
-            set_magic_quotes_runtime(0);
             if ($this->_readControl) {
                 $hashControl = @fread($fp, 32);
                 $length -= 32;
             }
             $data = @fread($fp, $length);
-            set_magic_quotes_runtime($mqr);
             if ($this->_fileLocking) @flock($fp, LOCK_UN);
             @fclose($fp);
```
kojiromike, line 539: is there a missing try/finally here?

interface/forms/LBF/new.php
```diff
@@ -937,17 +928,14 @@ function warehouse_changed(sel) {
         // small devices. In particular we prefer horizontal arrangement of multiple
         // items in the same row and column.
         echo "<form method='post' class='form-inline' " .
-            "action='$rootdir/forms/LBF/new.php?formname=" . attr_url($formname) . "&id=" . attr_url($formid) . "&portalid=" . attr_url($portalid) . "&formOrigin=" . attr_url($form_origin) . "&isPortal=" . attr_url($patient_portal) . "' " .
+            "action='$rootdir/forms/LBF/new.php?formname=" . attr_url($formname) . "&id=" . attr_url($formid) . "&formOrigin=" . attr_url($form_origin) . "&isPortal=" . attr_url($patient_portal) . "' " .
```
kojiromike, line 931: if we don't already, we should have an issue to clean up manual url construction in favor of a url builder

interface/main/messages/save.php
```diff
@@ -255,11 +255,4 @@
     echo text(json_encode($pidList));
     exit;
 }
-if ($_REQUEST['go'] == "Messages") {
```
kojiromike, line 258: What prompted this removal?

interface/main/backup.php
```diff
@@ -59,13 +59,6 @@
       die('Abort ' . basename(__FILE__) . ' : Missing zlib extensions');
 }
 
-if (!function_exists('gzopen') && function_exists('gzopen64')) {
```
kojiromike, line 62: what prompted this removal?

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -3630,17 +3630,9 @@ public function fetchFormValues($pid, $encounter, $formTables)
...
-                            ob_start();
-                            ?>
-                            <table>
-                                <?php
-                                display_layout_rows_group_new($formDir, '', '', $pid, $value, [$formTables_details[1]], '');
-                                ?>
-                            </table>
-                            <?php
-                            $res[0][$value] = ob_get_clean();
+                            //This section will be used to fetch an LBF group. This has to be completed. We are working on this.
```
kojiromike: What prompted this removal? You either went too far, or you didn't go far enough.

interface/patient_file/report/custom_report.php
```diff
@@ -818,59 +818,35 @@ function getContent()
...
-                    echo $tmp_file;
-                    exit();
+        try {
+            if ($PDF_FAX === 1) {
+                $fax_pdf = $pdf->Output($fn, 'S');
+                $tmp_file = OEGlobalsBag::getInstance()->getString('temporary_files_dir') . '/' . $fn; // is deleted in sendFax...
+                file_put_contents($tmp_file, $fax_pdf);
+                echo $tmp_file;
+                exit();
+            } else {
```
kojiromike: don't do `else` after `exit()`.

interface/patient_file/report/custom_report.php
```diff
@@ -818,59 +818,35 @@ function getContent()
...
-                    file_put_contents($tmp_file, $fax_pdf);
-                    echo $tmp_file;
-                    exit();
+        try {
+            if ($PDF_FAX === 1) {
+                $fax_pdf = $pdf->Output($fn, 'S');
+                $tmp_file = OEGlobalsBag::getInstance()->getString('temporary_files_dir') . '/' . $fn; // is deleted in sendFax...
+                file_put_contents($tmp_file, $fax_pdf);
+                echo $tmp_file;
+                exit();
```
kojiromike: could this be `return` instead?

interface/patient_file/report/custom_report.php
```diff
@@ -818,59 +818,35 @@ function getContent()
...
+            } else {
+                if (!empty($archive_name) && count($staged_docs) > 0) {
```
kojiromike: don't use `empty`

interface/patient_file/report/custom_report.php
```diff
@@ -818,59 +818,35 @@ function getContent()
...
-                'contents' => base64_encode((string) $contents)
-            ]);
-            if ($result['errmsg']) {
-                die(text($result['errmsg']));
             }
-
-            echo "<p class='mt-3'>" . xlt('Report has been sent to the patient.') . "</p>\n";
-            echo "</body></html>\n";
+        } catch (MpdfException $exception) {
+            die(text($exception));
```
kojiromike: `die` is forbidden. What can we do instead?

interface/patient_file/pos_checkout_ippf.php
```diff
@@ -1747,6 +1747,9 @@ function write_old_payment_line($pay_type, $date, $method, $reference, $amount):
             $checkout_times = craGetTimestamps($patient_id, $_GET['enc']);
             $billtime = empty($checkout_times) ? '' : $checkout_times[count($checkout_times) - 1];
         }
+        if (!function_exists('generateCheckoutReceipt')) {
+            die(xlt('The custom receipt include does not define generateCheckoutReceipt().'));
```
kojiromike: `die` is forbidden

kojiromike:
@aisle-analyzer

kojiromike (commented):
oops, this comment was meant for #11801.

gacl/Cache_Lite/Lite.php
kojiromike: This PR only removed two lines inside `_read()` — the `set_magic_quotes_runtime(0)` / `set_magic_quotes_runtime($mqr)` pair that were removed in PHP 7, so they've been fatal on every supported runtime. The surrounding flow (the `@flock` / `@fclose` pattern without try/finally) predates this PR — it's legacy vendor Cache_Lite code that uses `@` error suppression throughout, so nothing in the read path can throw today. Adding try/finally here is a real but separate refactor; happy to file a follow-up if you want, but it's outside this PR's "drain function-not-found" scope.

interface/forms/LBF/new.php
kojiromike: Filed openemr/openemr#11815 for the URL-builder migration. We didn't have an existing tracking issue for it.

interface/main/messages/save.php
kojiromike: The baseline had `Function updateMessage not found` for this file. `updateMessage()` has no definition anywhere in the codebase (searched with `git grep 'function updateMessage'`), and the entry points that target this handler don't POST `go=Messages` — the value dispatched here is `Post`, `Delete`, or the `labels`/`recalls` branches above. The branch is unreachable dead code whose only effect is to throw a fatal on the off chance anyone forged the POST, so I removed it.

interface/main/backup.php
kojiromike: `gzopen64` was a PHP 5 artifact on some 32-bit builds for large-file support — it doesn't exist in any PHP 7+ build. The `if (!function_exists('gzopen') && function_exists('gzopen64'))` guard was always false on every supported runtime, so the inner `function gzopen(...)` shim was dead code; the baseline's `Function gzopen64 not found` entry points at the `return gzopen64(...)` body inside the never-taken branch. Since `gzopen` is a core zlib function (and we already `die()` on missing zlib above), the shim can never run anyway.

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
kojiromike: Fair point. The function called inside this branch was `display_layout_rows_group_new(...)`, which doesn't exist anywhere in the codebase — so this was a WIP stub that happened to be wrapped in plausible-looking scaffolding. The sibling branch immediately above at line 3626 (`if (!$formTables_details[1])`) is already an empty WIP stub with the exact same comment pattern ("This section will be used to fetch… This has to be completed. We are working on this."), so I matched it. The alternative would be to rip out both WIP branches plus their `elseif` dispatcher, but that starts reshaping the CCDA LBF control flow, which is well outside "drain function-not-found." Matching the adjacent precedent felt like the least-invasive fix here.

interface/patient_file/report/custom_report.php
kojiromike: Fixed in e1555ff — flattened the `if ($PDF_FAX === 1) { … exit(); } else { … }` into an early-return-style `if ($PDF_FAX === 1) { … exit(); }` followed by the remaining logic at the outer level.

interface/patient_file/report/custom_report.php
kojiromike: This is top-level script code in an entry point (not a function body), so `return` would `return` from the script itself if it was required; `exit()` is idiomatic here and matches the rest of the file. I kept `exit()` but flattened the `else`-after-`exit` (see the adjacent thread) so the structure reads as early-exit + continuing outer flow.

interface/patient_file/report/custom_report.php
kojiromike: Fixed in e1555ff — swapped `!empty($archive_name) && count($staged_docs) > 0` for `$archive_name !== '' && count($staged_docs) > 0`. `$archive_name` is a string initialized to `''` at line 51, so `!== ''` is the exact intent. Drops one more entry off the `empty.notAllowed` baseline for this file (13 → 12).

interface/patient_file/report/custom_report.php
kojiromike: Fixed in e1555ff — replaced with `throw new \RuntimeException('PDF generation failed', 0, $exception)`, which preserves the original `MpdfException` as the previous exception and lets the global handler log it with the full stack trace. Drops the `openemr.exitInCatchOrFinally` baseline for this file from 2 → 1 (the other is pre-existing at line 811, outside this PR's scope).

interface/patient_file/pos_checkout_ippf.php
kojiromike: Fixed in e1555ff — swapped `die(...)` for `throw new \RuntimeException(...)`. The global exception handler will log and render, and the misconfiguration shows up with a stack trace instead of a silent terminator.

kojiromike:
Thanks for the scan, aisle-research-bot — the finding is correct.

The race is pre-existing: `BillingUtilities::updateInvoiceRefNumber()` has always been a non-atomic read/increment/write, and `interface/patient_file/pos_checkout{,_ippf}.php` already stored its return into `form_encounter.invoice_refno`. This PR only renames a dead global call in `library/FeeSheet.class.php` to the surviving static method — it doesn't change concurrency semantics, add a new call site, or widen the exposure. Fixing it properly needs a transaction/`FOR UPDATE` or an atomic SQL increment plus test coverage for concurrent checkouts, which is meaningfully out of scope for the fatal-baseline drain.

Tracked for separate work in openemr/openemr#11818.

interface/main/messages/save.php
```diff
@@ -255,11 +255,4 @@
...
-    if ($_REQUEST['msg_id']) {
-        $result = updateMessage($_REQUEST['msg_id']);
-        echo json_encode($result);
-        exit;
-    }
-}
 exit;
```
kojiromike: what is the point of `exit` at the end of this file?

interface/patient_file/report/custom_report.php
kojiromike: `exit` and `die` are forbidden functions. You aren't following CLAUDE.md here.

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
kojiromike: Then you didn't go far enough. Clean up this entire WIP branch.

interface/main/messages/save.php
kojiromike: Good catch — removed in 66c6080. Every dispatch branch already calls `exit` or `return` of its own, so the trailing top-level `exit;` was only reachable when no branch matched, in which case falling off the end of the script is identical behavior.

interface/patient_file/report/custom_report.php
kojiromike: Fair — switched `exit()` to `return;` in 66c6080. Since this is top-level script code (inside `if ($PDF_OUTPUT)` in the script body, not a function), `return` exits the include cleanly and skips the remaining branches just like `exit()` did. The two other `die()` calls in this file (lines 811, 818) predate this PR and are out of scope for the fatal-baseline drain.

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
kojiromike: Done in 66c6080 — ripped out both WIP stubs (`!$formTables_details[1]` and `!$formTables_details[3]` branches) and widened the surviving `elseif` to `$formTables_details[0] == 2 && $formTables_details[1] && $formTables_details[3]`. The removed branches were empty foreach-over-foreach loops whose only bodies were "we are working on this" comments, and the helper (`display_layout_rows_group_new`) they were written to call was never defined — so the behavior of both branches was unconditionally no-op, which is equivalent to skipping the `elseif` entirely. Left a comment noting what was removed and why.

PR #11816, opened 2026-04-24
ci(phpstan): add composer phpstan-baseline-reset to wipe and rebuild

kojiromike:
## Summary

Adds a `composer phpstan-baseline-reset` script that deletes every file under `.phpstan/baseline/`, writes a minimal empty `loader.php`, and then runs `composer phpstan-baseline` to regenerate the per-identifier files from scratch.

## Why

The regular `composer phpstan-baseline` reuses the existing baseline as input, which is the wrong starting point when the baseline has drifted, or when the files are in a state PHPStan can't even load — for example, leftover merge conflict markers after resolving an upstream rebase. In those cases the only way to recover is to wipe and rebuild, and this script automates that.

## What's in the change

- `.phpstan/reset-baseline.php` — tiny PHP helper that deletes `.phpstan/baseline/*.php` and writes a valid empty `loader.php`.
- `composer.json` — new `phpstan-baseline-reset` script chains the helper and then `@phpstan-baseline`.
- `.phpstan/README.md` — documents when to prefer `phpstan-baseline-reset` over `phpstan-baseline`.

## Test plan

- [ ] `composer validate --strict` passes
- [ ] `php -l .phpstan/reset-baseline.php` passes
- [ ] On a worktree whose baseline contains leftover merge conflict markers, `composer phpstan-baseline-reset` wipes the directory, writes a valid empty `loader.php`, and then regenerates the full baseline successfully

### DIFF
--- a/.phpstan/README.md
+++ b/.phpstan/README.md
@@ -276,6 +276,14 @@ To regenerate the baseline after fixing violations:
 composer phpstan-baseline
 ```
 
+To wipe and rebuild the baseline from scratch (ignoring whatever is currently on disk):
+
+```bash
+composer phpstan-baseline-reset
+```
+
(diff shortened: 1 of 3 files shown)


kojiromike:
cc @Firehed

PR #11821, opened 2026-04-24
chore(phpstan): drain confident nonObject baselines (21 → 0)

kojiromike:
## Summary

Part of #11792 — drains every confident `*.nonObject` baseline entry (21 across three files) and lowers all five `confidentNonObject` caps to 0.

The `FatalBaselineCapsIsolatedTest` counts nonObject entries whose reported type is definitely non-object (null/false/true/scalar/array) — those are the real runtime `TypeError`s. Excluded are entries reported on `mixed`, `object`, or class unions (where PHPStan just couldn't prove it).

Root-cause fixes, not narrowing:

- **`src/Billing/PaymentGateway.php`** — `submitPaymentCard` / `submitPaymentToken` returned an Omnipay `ResponseInterface` or a string, but the docblock claimed `bool|string`. After `is_string($response)` guards, PHPStan narrowed the remainder to `bool` and flagged every subsequent method call. Corrected return type. Added `@var` at the call sites (`front_payment_cc.php`, `paylib.php`) so gateway-specific methods (`getParsedData`, `getSource`) resolve against the concrete `Omnipay\AuthorizeNetApi\Message\Response` / `Omnipay\Stripe\Message\Response`.
- **`$r->address_zip`** (front_payment_cc + paylib) — property access on an array returned by Stripe `Response::getSource()`. Switched to array access with a null-coalesce guard.
- **`portal/patient/fwk/libs/verysimple/Phreeze/Phreezer.php`** —
  - `GetFieldMaps` / `GetKeyMaps` docblocked `@return Array of FieldMap/KeyMap objects` (PHPStan reads the literal `Array`). Fixed to `array<string, FieldMap>` / `array<string, KeyMap>`.
  - `GetFieldMap` / `GetKeyMap` (singular) claimed to return arrays but actually return the element — corrected to `FieldMap` / `KeyMap`.
  - `$_mapCache` typed `@var array` but assigned a `CacheRam`. Retyped as `ICache` to match the other cache slots.
- **`InstallerController::InstallModuleACL`** — `empty(\$modDir->acl_version)` where `\$modDir` is a filesystem path string. Typo for `\$Module->acl_version`. While fixing, swapped `empty()` for explicit null/empty-string checks.
- **`ResultsCalculator::aggregate_results_by_measure`** — `reset(\$pop_set)` returns `false` on an empty array; the loop immediately dereferenced `->observations`. Added a guard.
- **`FhirOrganizationInsuranceService::parseFhirResource`** — `\$fhirResource::class` in the `!instanceof FHIROrganization` branch (runs when the input is an array). Switched to `get_debug_type()` and widened the param docblock to `FHIROrganization|array`.

## Test plan

- [x] `composer phpstan-baseline` regenerates cleanly
- [x] `composer phpunit-isolated -- --filter FatalBaselineCaps` → 19/19 pass
- [x] Pre-commit hooks green (PHP syntax, PHPCS, PHPStan, Rector, composer-require-checker)
- [ ] CI green

### DIFF
--- a/.phpstan/baseline/assign.propertyType.php
+++ b/.phpstan/baseline/assign.propertyType.php
@@ -537,11 +537,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/Phreeze/Phreezer.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Property Phreezer\\:\\:\\$_mapCache \\(array\\) does not accept CacheRam\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/Phreeze/Phreezer.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Static property VerySimpleStringUtil\\:\\:\\$CONTROL_CODE_CHARS \\(characters\\) does not accept array\\<string, string\\>\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/String/VerySimpleStringUtil.php',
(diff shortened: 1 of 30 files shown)


Firehed:
Nice one. LGTM

PR #11830, opened 2026-04-25
feat(cli): add API enablement flags to install command

Firehed:

Adds CLI flags to `./cli install` for enabling REST/FHIR APIs at install time, replacing manual SQL inserts in CI workflows and/or having to toggle these settings through the UI or manual SQL commands.

- Add 7 new options to `./cli install`:
  - `--enable-rest-api`
  - `--enable-fhir-api`
  - `--enable-portal-api`
  - `--enable-password-grant`
  - `--enable-system-scopes`
  - `--enable-ccda`
  - `--site-address=<url>`
- Uses existing `custom_globals` infrastructure in the installer
- Update `test-frontcontroller.yml` to use the new flags instead of raw SQL


src/Console/Command/InstallCommand.php
```diff
@@ -87,4 +109,29 @@ public function __invoke(
...
+        bool $enablePasswordGrant,
+        bool $enableSystemScopes,
+        bool $enableCcda,
+        string $siteAddress,
+    ): array {
+        $flags = array_filter([
+            GlobalConnectorsEnum::REST_API->value => $enableRestApi ? '1' : null,
+            GlobalConnectorsEnum::REST_FHIR_API->value => $enableFhirApi ? '1' : null,
+            GlobalConnectorsEnum::REST_PORTAL_API->value => $enablePortalApi ? '1' : null,
+            GlobalConnectorsEnum::OAUTH_PASSWORD_GRANT->value => $enablePasswordGrant ? '3' : null,
```
kojiromike: `oauth_password_grant` is a tri-state global — `1` = users only, `2` = patients only, `3` = both (see the checks in `src/Common/Auth/OpenIDConnect/Repositories/UserRepository.php` around lines 91 and 137). A `bool` flag collapses that to off / both, which is fine for the CI use case but loses the other two modes for general installer use.

More broadly, `oauth_password_grant` looks like a good candidate for an int-backed `enum`. At minimum, a brief comment explaining why `3` is the chosen mapping would help.

src/Console/Command/InstallCommand.php
```diff
@@ -87,4 +109,29 @@ public function __invoke(
...
+            GlobalConnectorsEnum::REST_SYSTEM_SCOPES_API->value => $enableSystemScopes ? '1' : null,
+            GlobalConnectorsEnum::SITE_ADDRESS_OAUTH->value => $siteAddress !== '' ? $siteAddress : null,
+            'ccda_alt_service_enable' => $enableCcda ? '3' : null,
```
kojiromike: Every other key in this array routes through `GlobalConnectorsEnum`; `ccda_alt_service_enable` is a bare string literal. Would it make sense to add a `CCDA_ALT_SERVICE_ENABLE` case to the enum so the array stays uniform and the global name is type-checked at compile time? The enum already has a TODO inviting more cases.

Also, `'3'` here is a magic value with no nearby explanation — a one-line comment (or a constant on the enum) clarifying what `3` means for this global would help future readers.

src/Console/Command/InstallCommand.php
Firehed: Filed #11863 to cover this better, going to leave inline comments for now.

src/Console/Command/InstallCommand.php
Firehed: Adding the key, commenting for the enum thing again.

PR #11859, opened 2026-04-27
chore(phpstan): drain Carecoordination module class.notFound + method.notFound

kojiromike:
Drain `class.notFound` and `method.notFound` baseline entries originating in the Carecoordination zend module. Fourth phase-4 module batch following the eRx drain (#11825).

## Source changes

**Controller property native types.** Convert PHPDoc-typed properties on `CarecoordinationController` and `CcdController` to native types (constructor property promotion applied via Rector). PHPStan was resolving docblocks like `@var Documents\Controller\DocumentsController` relative to the `Carecoordination\Controller` namespace, producing `class.notFound` for nonexistent paths.

**`@method` tags for Laminas MVC plugins.** `$this->CommonPlugin()` and `$this->Documents()` resolve through Laminas' plugin manager via `__call`. Add explicit `@method` declarations on the controllers that invoke them, plus `@method \Laminas\Http\Request getRequest()` to narrow the request type so `getPost()`/`getQuery()` resolve.

**Fix `CcdController::uploadAction()` bug.** The method was calling `$this->updateDocumentCategoryUsingCatname()` on itself, which does not exist on the controller. The sibling `CarecoordinationController::uploadAction()` calls it correctly via `$this->documentsController->getDocumentsTable()->updateDocumentCategoryUsingCatname()` — apply the same call here. Dead since at least 2014; PHPStan would have caught it as `method.notFound` if the entry hadn't been baselined.

**Remove dead `CarecoordinationTable::getIssues()`.** Body called `DocumentsTable::getIssues()` which does not exist. Author left an inline `Method not used` TODO; grep confirms no callers.

**Remove dead try/catch in `EncounterccdadispatchController::indexAction()`.** The block wrapped only `echo`/`exit` calls that cannot throw.

**`getCarecoordinationTable()` / `getDocumentsTable()` return types.** Changing the docblock from `@return object` to a real return type collapses several `method.notFound` entries that were `Cannot call method ... on object`.

**DOM narrowing.** `CcdaUserPreferencesTransformer` and `CcdaServiceDocumentRequestorTest` iterate `\DOMNodeList` results and call element-only methods (`insertBefore`, `setAttribute`, `getAttribute`). Add `instanceof \DOMElement` and `instanceof \DOMNode` guards plus a null-check on `lastElementChild` so the type system sees the narrowing.

**Content type narrowing in `EncounterccdadispatchController::indexAction()`.** Initialize `$content = ''` up front and call `strval($content)` before substring operations to avoid Rector flip-flop between `RecastingRemovalRector` (full-codebase) and `NullToStrictStringFuncCallArgRector` (single-file).

## Tests

Adds `CcdaUserPreferencesTransformerTest` covering `transform()` (sort, truncate, default-fallback) and the getter/setter round-trip. Lifts patch coverage on this PR.

## Counts

| Identifier | Before | After | Delta |
|---|---:|---:|---:|
| `class.notFound` | 237 | 229 | -8 |
| `method.notFound` | 184 | 160 | -24 |
| `method.nonObject` | 1995 | 1991 | -4 |
| `argument.type` | 14973 | 14972 | -1 |
| `assign.propertyType` | 388 | 384 | -4 |
| `cast.string` | 807 | 806 | -1 |
| `missingType.parameter` | 9488 | 9487 | -1 |
| `missingType.property` | 6198 | 6195 | -3 |
| `missingType.return` | 5913 | 5912 | -1 |
| `offsetAccess.nonOffsetAccessible` | 10757 | 10756 | -1 |
| `preDec.type` | 9 | 8 | -1 |
| `property.notFound` | 135 | 132 | -3 |
| `return.type` | 2167 | 2164 | -3 |
| `variable.undefined` | 3075 | 3074 | -1 |

Caps in `.phpstan/fatal-baseline-caps.php` updated to match.

Refs #11792.

## Test plan
- [x] `composer phpstan-baseline` regenerates cleanly
- [x] `composer phpunit-isolated -- --filter=FatalBaselineCaps` passes
- [x] `composer phpcs` clean on edited files
- [x] Pre-commit hooks (rector, phpcs, phpstan, codespell, composer-require-checker) all pass

### DIFF
--- a/.phpstan/baseline/catch.neverThrown.php
+++ b/.phpstan/baseline/catch.neverThrown.php
@@ -14,6 +14,11 @@ $ignoreErrors[] = [
 $ignoreErrors[] = [
     'message' => '#^Dead catch \\- Throwable is never thrown in the try block\\.$#',
     'count' => 1,
+    'path' => __DIR__ . '/../../interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Dead catch \\- Throwable is never thrown in the try block\\.$#',
+    'count' => 1,
     'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/Phreeze/PortalController.php',
 ];
 $ignoreErrors[] = [
--- a/.phpstan/baseline/catch.neverThrown.php
+++ b/.phpstan/baseline/catch.neverThrown.php
@@ -14,11 +14,6 @@ $ignoreErrors[] = [
 $ignoreErrors[] = [
     'message' => '#^Dead catch \\- Throwable is never thrown in the try block\\.$#',
     'count' => 1,
-    'path' => __DIR__ . '/../../interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Dead catch \\- Throwable is never thrown in the try block\\.$#',
-    'count' => 1,
     'path' => __DIR__ . '/../../portal/patient/fwk/libs/verysimple/Phreeze/PortalController.php',
 ];
 $ignoreErrors[] = [
--- a/.phpstan/baseline/variable.undefined.php
+++ b/.phpstan/baseline/variable.undefined.php
@@ -10292,11 +10292,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Variable \\$content might not be defined\\.$#',
-    'count' => 8,
-    'path' => __DIR__ . '/../../interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Variable \\$result might not be defined\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php',
--- a/.phpstan/baseline/variable.undefined.php
+++ b/.phpstan/baseline/variable.undefined.php
@@ -10293,7 +10293,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Variable \\$content might not be defined\\.$#',
-    'count' => 6,
+    'count' => 8,
     'path' => __DIR__ . '/../../interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Controller/EncounterccdadispatchController.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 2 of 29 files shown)


.phpstan/baseline/catch.neverThrown.php
```diff
@@ -11,6 +11,11 @@
     'count' => 1,
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-weno/src/Services/DownloadWenoPharmacies.php',
 ];
+$ignoreErrors[] = [
```
kojiromike: This is a weird finding. If Throwable is never thrown, then nothing at all can possibly be thrown, since everything that can be thrown is Throwable. Is this a namespacing problem?

.phpstan/baseline/variable.undefined.php
```diff
@@ -11898,7 +11903,7 @@
 ];
 $ignoreErrors[] = [
     'message' => '#^Variable \\$content might not be defined\\.$#',
-    'count' => 6,
+    'count' => 8,
```
kojiromike: this should not be going up.

tests/Tests/Services/Modules/Carecoordination/Model/CcdaServiceDocumentRequestorTest.php
```diff
@@ -174,6 +176,9 @@ private function replaceRootIdForNodes(DOMXPath $path, DOMXPath $expectedXpath,
 
             $currentNodeId = $path->query(".//hl7:id", $currentNode)->item(0);
             $expectedNodeId = $expectedXpath->query(".//hl7:id", $expectedNode)->item(0);
+            if (!$currentNodeId instanceof \DOMElement || !$expectedNodeId instanceof \DOMElement) {
+                throw new \RuntimeException('Expected DOMElement for hl7:id node');
```
kojiromike: is there a more appropriate exception type to use in tests?

.phpstan/baseline/catch.neverThrown.php
kojiromike: Right — the `try { ob_clean(); ... echo $content; ... exit; } catch (\Throwable)` block had nothing in its body that could throw (echo, header, exit don't throw), so PHPStan correctly flagged the catch as dead. Fixed by removing the dead try/catch wrapper in ca71340562. Also dropped the inner `if (empty($downloadccda))` branch since the only path reaching that point has $downloadccda truthy (the actual download is handled by the earlier `forward()->dispatch(EncountermanagerController::class, 'downloadall')`).

.phpstan/baseline/variable.undefined.php
kojiromike: Agreed. Fixed at the source by initializing `$content = ''` at the top of `indexAction()` so it's always defined regardless of which generation branch runs. Drops 8 entries (6→8 increase reverted plus the original 6) for that file. variable.undefined cap goes 3397 → 3396 in ca71340562.

tests/Tests/Services/Modules/Carecoordination/Model/CcdaServiceDocumentRequestorTest.php
kojiromike: Good call — replaced the `\RuntimeException` throws with `static::assertInstanceOf(\DOMElement::class, ...)` so the failure mode is a PHPUnit assertion (and PHPStan still narrows the type). ca71340562.

PR #11860, opened 2026-04-27
refactor(daysheet): extract DaySheetAggregator from print_daysheet_report_num1

kojiromike:
## Summary

Replaces the per-slot accumulator pattern (`$us0_*..$us19_*`, `$pro0_*..$pro19_*`) in `interface/billing/print_daysheet_report_num1.php` with a typed `OpenEMR\Billing\DaySheet` aggregator. The legacy code silently dropped any rows beyond the 20th unique user or provider; the new aggregator accumulates every distinct key.

Drains `variable.undefined` baseline by 320 entries (3397 → 3077).

Part of #11792.

## What's in the new namespace

- `BillRow` — typed row DTO, parsed at the array boundary
- `SlotTotals` — per-key accumulator (preserves the legacy split-refund policy: insurance/patient amounts split into payment vs refund buckets by sign and code type)
- `DaySheetAggregator` — aggregates rows into per-user, per-provider, and grand totals
- `DaySheetTotals` — result object, filters all-zero slots

Isolated test (`tests/Tests/Isolated/Billing/DaySheet/DaySheetAggregatorTest.php`) covers the legacy semantics including PCP refund exclusion, sign-based payment/refund split, and the 20-slot overflow case.

## What's still in the legacy file

The page footer rendering still depends on legacy variables (`$gtotal_*`, `$totals_only`, `$query_part_day`, `$line_total*`, etc.) that were initialized as side effects of the deleted code. Those need explicit initialization or wiring to `$daySheetTotals->grandTotals` before this is shippable. That's why this is **draft**.

`print_daysheet_report_num2.php` and `print_daysheet_report_num3.php` (sibling reports with similar shape, ~262 baseline entries between them) are out of scope for this PR but the aggregator should be reusable for them.

## Test plan

- [ ] Run end-of-day day sheet report in OpenEMR docker (golden path)
- [ ] Run with `end_of_day_provider_only=1` (provider mode)
- [ ] Run with `end_of_day_totals_only=1` (totals mode)
- [ ] Verify totals match against a dataset with >20 unique users (legacy would drop these)
- [x] `composer phpunit-isolated` passes (DaySheetAggregatorTest)
- [x] `composer phpstan` passes


.phpstan/baseline/cast.string.php
```diff
@@ -3016,6 +3021,11 @@
     'count' => 109,
     'path' => __DIR__ . '/../../src/Billing/Claim.php',
 ];
+$ignoreErrors[] = [
+    'message' => '#^Cannot cast mixed to string\\.$#',
```
kojiromike: narrow, don't cast.

.phpstan/baseline/foreach.nonIterable.php
```diff
@@ -128,7 +128,7 @@
 ];
 $ignoreErrors[] = [
     'message' => '#^Argument of an invalid type mixed supplied for foreach, only iterables are supported\\.$#',
-    'count' => 2,
+    'count' => 4,
```
kojiromike: this count should only go down, not up

.phpstan/baseline/property.nonObject.php
```diff
@@ -156,6 +156,56 @@
...
+    'count' => 2,
+    'path' => __DIR__ . '/../../interface/billing/print_daysheet_report_num1.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Cannot access property \\$providerTotals on mixed\\.$#',
+    'count' => 1,
+    'path' => __DIR__ . '/../../interface/billing/print_daysheet_report_num1.php',
+];
+$ignoreErrors[] = [
+    'message' => '#^Cannot access property \\$userTotals on mixed\\.$#',
```
kojiromike: narrow all these mixed to the right type.

src/Billing/DaySheet/BillRow.php
```diff
@@ -0,0 +1,74 @@
...
+        }
+        return (string) $value;
+    }
+
+    private static function asFloat(mixed $value): float
+    {
+        if (is_int($value) || is_float($value)) {
+            return (float) $value;
+        }
+        if (is_string($value) && is_numeric($value)) {
```
kojiromike:
```suggestion
        if (is_numeric($value)) {
```

kojiromike: Applied. Also collapsed the int/float branch — `is_numeric` covers numeric strings, ints, and floats together.

.phpstan/baseline/cast.string.php
kojiromike: Done. Replaced `(string) $value` in `BillRow::asString` with explicit narrowing (`is_string` returns as-is, `is_int|is_float` casts the safe scalars, everything else returns `''`). Baseline entry is gone.

.phpstan/baseline/foreach.nonIterable.php
kojiromike: Reverted. Root cause was that `$daySheetTotals` was assigned inside `if ($ret = getBillsBetweendayReport(...))` but consumed by the populate blocks at outer scope, so PHPStan inferred mixed there. Initialized `$daySheetTotals = new DaySheetTotals([], [], new SlotTotals(''))` before the conditional — count is back to 2.

.phpstan/baseline/property.nonObject.php
kojiromike: Same root cause as the foreach.nonIterable comment: `$daySheetTotals` was mixed at the populate blocks because it was assigned inside the `if ($ret = ...)` and consumed outside. Initializing it before the conditional removed all 50 entries.

PR #11864, opened 2026-04-27
fix(db): Log all "helpfuldie" sql errors

Firehed:
The older-style `HelpfulDie` does not particularly live up to its name. Fixing it the right way is removal in favor of everything using exceptions, which is an ongoing but massive task. Debugging is overly complex.

This logs the function and full stack trace any time the legacy global-namespace SQL functions catch/suppress/crash any sort of query failure. Yes, `HelpfulDie` makes an attempt to do this, but it's rather limited in how it works and doesn't go through the SystemLogger.

I was hoping this would help with diagnosing some Inferno test issues. It doesn't seem to, but I think it's still a worthwhile addition.

Add an error-level log to all of the `catch` paths inside the older SQL functions, going to the system logger.

No

### DIFF
--- a/library/sql.inc.php
+++ b/library/sql.inc.php
@@ -98,6 +98,7 @@ function sqlStatement($statement, $binds = false)
     try {
         return QueryUtils::sqlStatementThrowException($statement, $binds, noLog: false);
     } catch (SqlQueryException $e) {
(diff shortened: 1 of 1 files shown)


kojiromike (approved):
Did you consider `critical` in lieu of `error`? Other than that this looks good.

PR #11866, opened 2026-04-27
fix(db): Fix SQL upgrade syntax

Firehed:
While working on #11805 I came across a bug when running a SQL upgrade. Invalid syntax! Since the block was conditional, you wouldn't always hit it. The Inferno data did.

Splits the index additions into a syntactically-legal version without changing the results.

Well, technically the new conditionals could change the results, but the intent of what it did before was clear.

No


kojiromike:

The bug is in sql/7_0_3-to-7_0_4_upgrade.sql, an older upgrade-path file. Verify the conditional actually affects rel-810 upgraders. If applied, pair with #11867.

PR #11867, opened 2026-04-27
chore(db): Fix typo

Firehed:
Correction from #11866, which was incorrectly extracted from #11805.


kojiromike:

Typo follow-up to #11866. Apply together.

PR #11868, opened 2026-04-27
fix(billing): cast 835 monetary fields to float for type-strict comparisons

stephenwaite:

The 835 parser in src/Billing/ParseERA.php populates monetary fields (chg, paid, allowed, adj.amount) as strings from explode()-split segment data. Commit 06cbbae9 tightened the error-detection branch in era_callback() to use $svc['paid'] === 0 — but since the parser produces string '0', this strict comparison is always false. As a result, $0-pay denials with codes other than CO-45/CO-59 are silently written off as contractual adjustments instead of being flagged for biller review. T

This fixes the regression by casting monetary fields to `float` at the parser level so the strict comparison works as designed, and updates the comparison itself to compare against `0.0` (float) for type alignment.

- `src/Billing/ParseERA.php`: cast `chg`, `paid`, `allowed`, and `adj.amount` to `float` at every assignment site.
- `interface/billing/sl_eob_process.php`: update the strict comparison from `=== 0` to `=== 0.0`, and update the artificial Claim row's `chg`/`paid` initialization from string `'0'` to float `0.0`.


interface/billing/sl_eob_process.php
```diff
@@ -631,7 +631,7 @@ class: $class,
                     echo getMessageLine($bgcolor, $class, $description . ' ' .
                     sprintf("%.2f", $adj['amount']));
                 } elseif (
-                    $svc['paid'] === 0
+                    $svc['paid'] === 0.0
```
kojiromike, line 637: While we're touching this branch — the condition is hard to read because it mixes a positive check with a negated nested boolean. Could it be extracted into a named flag? Something like:

```suggestion
                } elseif ($svc['paid'] === 0.0 && !$isContractualWriteoff) {
```

…with `$isContractualWriteoff` computed just above the `if/elseif` chain as:

```php
$isContractualWriteoff = $adj['group_code'] === 'CO'
    && in_array($adj['reason_code'], ['45', '59'], true);
```

The name carries the domain meaning (CO-45 / CO-59 are the standard contractual-adjustment CARC codes) so the next reader doesn't have to decode the boolean to know what's being excluded. Happy to leave it for a follow-up if you'd rather keep this PR minimal.

src/Billing/ParseERA.php
```diff
@@ -291,7 +291,7 @@ public static function parseERA($filename, $cb)
                     $out['svc'][$i]['adj'][$j] = [];
                     $out['svc'][$i]['adj'][$j]['group_code'] = $seg[1];
                     $out['svc'][$i]['adj'][$j]['reason_code'] = $seg[$k];
-                    $out['svc'][$i]['adj'][$j]['amount'] = $seg[$k + 1];
```
kojiromike, line 294: Style nit: this is the only `floatval()` call introduced in the diff — the SVC handler below at line 383 uses `(float)`, as does the CAS handler at line 412. Worth using `(float)` here too for consistency.

```suggestion
                    $out['svc'][$i]['adj'][$j]['amount'] = (float)$seg[$k + 1];
```

src/Billing/ParseERA.php
```diff
@@ -409,15 +409,19 @@ public static function parseERA($filename, $cb)
...
+                    $amount = $seg[$k + 1] ?? 0;
+                    if (!is_scalar($amount)) {
+                        $amount = 0;
+                    }
```
kojiromike: This handler now looks quite different from the structurally identical CAS handler at line 294 (claim-level), which just does `(float)$seg[$k + 1]` with no null-coalesce or scalar guard. Since `$seg = explode('*', ...)` a few lines up always produces strings, the `is_scalar` branch is unreachable in practice — and the asymmetry between the two CAS handlers is more confusing than the defensive code is protective.

Would it make sense to match the simpler form here?

```suggestion
                    $out['svc'][$i]['adj'][$j]['amount'] = (float)($seg[$k + 1] ?? 0);
```

The `?? 0` keeps the missing-segment defensiveness (which the claim-level handler doesn't have, but is cheap); dropping the `is_scalar` block trims dead code and lines up the two handlers visually.

kojiromike (approved):
Fix is correct — the regression in `06cbbae9` is real and `=== 0` against a string `'0'` is always false. Casting at the parser boundary is the right shape (parse at the edge, trust the type downstream).

A few small things while we're touching this code:

1. The denial-detection condition reads as a thicket of nested negation. Now that we're in the file, it'd be a good chance to make it self-documenting.
2. `(float)` and `floatval()` are mixed within the same parser. Pick one (`(float)` is the prevailing style in the rest of the diff).
3. The `is_scalar` guard at the new CAS service-level handler is asymmetric with the otherwise-identical CAS claim-level handler a few lines up — and `$seg` always comes from `explode()`, so the values are guaranteed strings. Either both paths need the guard or neither does.

No blocker — the behavioral fix is sound on its own.

src/Billing/ParseERA.php
kojiromike: One more reason to prefer `(float)` here that I missed on the first pass: `floatval()` would also bump an existing PHPStan baseline entry.

`.phpstan/baseline/argument.type.php` already carries:

```
'message' => '#^Parameter \\#1 \\$value of function floatval expects array\\|bool\\|float\\|GMP\\|int\\|resource\\|SimpleXMLElement\\|string\\|null, mixed given\\.$#',
'count' => 3,
'path' => __DIR__ . '/../../src/Billing/ParseERA.php',
```

`$seg` is `mixed` from PHPStan's POV (the `$out`/`$seg` paths are untyped), so this new `floatval($seg[$k + 1])` would push that count from 3 to 4 the next time the baseline is regenerated. The `(float)` cast has no equivalent baseline rule, so swapping avoids growing the baseline.

(Wider note, not a blocker for this PR: most of the ParseERA baseline entries — `binaryOp.invalid`, `assignOp.invalid`, the `argument.type` ones for `count`/`fopen`/`floatval` — all stem from `parseERA(&$out, $cb)` having no type on `$out`. Until `$out` gets a shape, these float casts can't actually shrink the baseline; PHPStan still reads the values back as mixed downstream. A separate effort.)

kojiromike:

One CI failure on this PR — verify whether it's a real failure or a flake before merging/backporting.

stephenwaite:
there's drift occurring from easy-dev phpstan and gh ci

> the easy dev Docker image likely has PHPUnit and its dependencies fully installed in a way that makes CommandTester resolvable to PHPStan, while CI's PHPStan run doesn't resolve it the same way despite having symfony/console.

Issue #10385, opened 2026-01-25
fix: rename misspelled 'declne_to_specfy' value in race list options


kojiromike:
## Summary

The `race` list option contains a misspelled value `declne_to_specfy` that should be `decline_to_specify`.

## Location

- Table: `list_options` (likely `list_id = 'race'`)
- Current value: `declne_to_specfy`
- Correct value: `decline_to_specify`

## Code reference

`src/Services/FHIR/FhirPatientService.php:437` has a workaround with a TODO comment:

```php
if ($race === 'declne_to_specfy') { // TODO: we should rename this misspelled value in the database
```

## Required changes

1. Add SQL migration to rename the value in `list_options`
2. Update any `patient_data` records that reference this value
3. Remove the hardcoded workaround in `FhirPatientService.php`

## Notes

This is a data migration, not just a code change. Need to ensure backwards compatibility for existing patient records.

phantom930:
Hi @kojiromike ,
Can I grab this one for my first contribution?

kojiromike:
@phantom930 sure, go for it.

phantom930:
Hi @kojiromike ,
I created PR #10762  for this issue and could you please review it when you havea  moment?
Thank you.

PR #11876, opened 2026-04-28
fix(db): convert declne_to_specfy in patient_data language and ethnicity

kojiromike:
## Summary

Follow-up to #10762, which renamed the misspelled `declne_to_specfy` race value to `decline_to_specify`. Brady's review on that PR flagged two issues in `sql/8_0_0-to-8_1_0_upgrade.sql` (originally `8_0_0-to-8_0_1_upgrade.sql`) that were never addressed before merge:

- The `patient_data.race` UPDATE was guarded by `#IfColumn patient_data race`, which is always true — the UPDATE ran on every upgrade even when there was nothing to convert. Switch to `#IfRow patient_data race declne_to_specfy` so it's skipped when no rows match.
- The same misspelled value can also appear in `patient_data.language` and `patient_data.ethnicity`. The `list_options` side already covers all three list_ids; this PR mirrors that on the patient row side.

Refs https://github.com/openemr/openemr/issues/10385, follow-up to https://github.com/openemr/openemr/pull/10762.

## Test plan

- [ ] Fresh install of 8.0.0 → upgrade to current master succeeds
- [ ] Tenant with `declne_to_specfy` in `patient_data.race` / `language` / `ethnicity` is migrated to `decline_to_specify`
- [ ] Tenant with no matching rows skips all three `#IfRow` blocks

### DIFF
--- a/sql/8_0_0-to-8_1_0_upgrade.sql
+++ b/sql/8_0_0-to-8_1_0_upgrade.sql
@@ -140,12 +140,21 @@ ALTER TABLE `facility` ADD `organization_type` VARCHAR(50) NOT NULL DEFAULT 'pro
 
 --
 -- Rename the misspelled list_options option_id from 'declne_to_specfy' to 'decline_to_specify',
--- and update any patient_data.race records that reference the old value.
+-- and update any patient_data.race, patient_data.language, and patient_data.ethnicity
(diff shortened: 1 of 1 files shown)


kojiromike:
@bradymiller I guess we'll need to get this in rel-810 too

PR #11888, opened 2026-04-28
fix(csrf): stop rotating CSRF private key on every main_screen.php load

### LINKED ISSUE #11865: main_screen.php rotates CSRF private key on every load, invalidating tokens in long-lived iframes (dated_reminders 60s poll)
## Summary

Loading `interface/main/main_screen.php` unconditionally regenerates the per-session CSRF private key, which silently invalidates CSRF tokens already embedded in pages/iframes rendered before that load. Long-lived widgets that bake their token in at render time and poll on a fixed interval — most visibly `interface/main/dated_reminders/dated_reminders.php` — then start logging `OpenEMR CSRF token authentication error` and returning the auth-error body to the user.

## Reproduction

Reproduced on `master` at `3da4f83cb680cc1c8bd4fd2cbc133ff4e4f26669`, fresh `docker/development-easy` stack, single admin session.

```bash
# 1. Log in (curl, fresh cookie jar) → cookies $JAR
# 2. Capture the CSRF token embedded in the reminders iframe:
curl -ks -b $JAR https://HOST/interface/main/dated_reminders/dated_reminders.php \
  | grep -oE 'csrf_token_form: "[a-f0-9]{40}"'
# → T

# 3. Verify the token works (mimics the periodic poll the JS issues every 60s):
curl -ks -b $JAR -X POST https://HOST/interface/main/dated_reminders/dated_reminders.php \
  -d "drR=0" -d "skip_timeout_reset=1" -d "csrf_token_form=$T"
# → HTTP 200, reminders HTML

# 4. Re-execute main_screen.php (e.g. user opens OpenEMR in a second window,
#    or the top frame reloads). With a valid CSRF this hits the else-branch
#    and rotates the private key:
curl -ks -b $JAR -X POST "https://HOST/interface/main/main_screen.php?auth=login_check" \
  -d "csrf_token_form=$T"
# → HTTP 302

# 5. The polling iframe is still alive and still POSTs $T:
curl -ks -b $JAR -X POST https://HOST/interface/main/dated_reminders/dated_reminders.php \
  -d "drR=0" -d "skip_timeout_reset=1" -d "csrf_token_form=$T"
# → HTTP 403, body "Authentication Error"
# → php-fpm stderr: "OpenEMR CSRF token authentication error"
```

A subsequent GET of `dated_reminders.php` shows a different `csrf_token_form` value, confirming the rotation.

## Mechanism

1. `src/Common/Csrf/CsrfUtils.php:35` — `setupCsrfKey()` writes a new `random_bytes(32)` into `$session['csrf_private_key']` every time it is called.
2. `interface/main/main_screen.php:392` calls `setupCsrfKey($session)` unconditionally — both the new-login `if` branch (line 378) and the migrate-only `else` branch (lines 385-388). The else-branch path only requires a valid CSRF token, so any in-app navigation that re-targets the top frame (refresh, second window, etc.) rotates the key.
3. `interface/main/dated_reminders/dated_reminders.php:139` bakes `CsrfUtils::collectCsrfToken(session: $session)` into the JS at render time, then `updateme(0)` (line 121) polls the same script every 60s (`$updateDelay = 60`, line 27) with `drR=0`.
4. `dated_reminders.php:43` — `isset($_POST['drR'])` is true even for `drR=0`, so every poll runs `CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true)` (line 44). Once the key has rotated, the embedded token no longer matches and the request 403s.

## Impact

- Spurious `OpenEMR CSRF token authentication error` entries in php-fpm stderr — observed in clusters at the 60s polling cadence on a real tenant (~750 errors in one month).
- The user does not see an error: the 403 body lands inside the hidden polling iframe, but the next render of the parent page works fine because the JS re-reads the (new) token. The result is alert noise rather than visible breakage, but it makes real CSRF failures hard to spot.
- Any other long-lived component that captures a CSRF token at render time and continues to use it after the top frame reloads is vulnerable in the same way; `dated_reminders.php` is just the most visible because of its 60s poll.

## Suggested direction (not prescriptive)

The CSRF rotation on every `main_screen.php` load looks like it was meant only for the new-login path. Confining `setupCsrfKey()` to the actual login branch (or to the case where there is no existing private key) would stop invalidating tokens that the same session has already handed out. There may be a security reason for the current behavior — happy to be told otherwise.

## Environment

- OpenEMR `master` @ `3da4f83cb680cc1c8bd4fd2cbc133ff4e4f26669`
- `docker/development-easy/` stack, Apache + php-fpm, MySQL
- PHP 8.x (image default)

kojiromike:
## Summary

Loading `interface/main/main_screen.php` unconditionally regenerated the per-session CSRF private key, silently invalidating CSRF tokens already embedded in pages and iframes rendered before that load. Long-lived widgets that bake their token in at render time and poll on a fixed interval — most visibly `interface/main/dated_reminders/dated_reminders.php` (60s poll) — then logged `OpenEMR CSRF token authentication error` on every poll following any in-app reload of the top frame.

The rotation looks like it was meant only for the new-login path. This PR moves `CsrfUtils::setupCsrfKey($session)` inside the new-login branch so the key is generated once at login and not rotated on subsequent top-frame loads. The `else` branch is reached only after a successful CSRF check, which already requires an existing private key, so no key setup is needed there.

Fixes #11865.

## Test plan

- [ ] Log in, capture the CSRF token rendered into `dated_reminders.php`, reload `main_screen.php`, then re-POST the same token to `dated_reminders.php` — should now return 200 instead of 403.
- [ ] Confirm fresh login still works (the new-login branch still runs `setupCsrfKey`).
- [ ] Confirm `OpenEMR CSRF token authentication error` log entries no longer appear at the 60s polling cadence after navigating between top frames.

### DIFF
--- a/interface/main/main_screen.php
+++ b/interface/main/main_screen.php
@@ -8,9 +8,11 @@
  * @author    Rod Roark <rod@sunsetsystems.com>
  * @author    Brady Miller <brady.g.miller@gmail.com>
  * @author    Ranganath Pathak <pathak@scrs1.org>
+ * @author    Michael A. Smith <michael@opencoreemr.com>
  * @copyright Copyright (c) 2018 Rod Roark <rod@sunsetsystems.com>
  * @copyright Copyright (c) 2018-2019 Brady Miller <brady.g.miller@gmail.com>
  * @copyright Copyright (c) 2019 Ranganath Pathak <pathak@scrs1.org>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
@@ -382,14 +384,17 @@ if (isset($_POST['new_login_session_management'])) {
     } else {
         $_POST["clearPass"] = '';
     }
+    // Set up the csrf private_key
+    //  Note this key always remains private and never leaves server session. It is used to create
+    //  the csrf tokens.
+    //  Generated only on the new-login path. Rotating it on every main_screen.php load
+    //  invalidates CSRF tokens already embedded in long-lived iframes (e.g. dated_reminders,
+    //  which polls every 60s with the token captured at render time).
+    CsrfUtils::setupCsrfKey($session);
 } else {
     CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
     $session->migrate(false);
 }
-// Set up the csrf private_key
-//  Note this key always remains private and never leaves server session. It is used to create
-//  the csrf tokens.
-CsrfUtils::setupCsrfKey($session);
 // Set up the session uuid. This will be used for mapping session setting to database.
 //  At this time only used for lastupdate tracking
 SessionTracker::setupSessionDatabaseTracker();


kojiromike:
@aisle-analyzer

PR #11895, opened 2026-04-28
refactor(billing): drain variable.undefined PHPStan baseline entries

kojiromike:
## Summary

Drain 341 entries from `.phpstan/baseline/variable.undefined.php` by initializing previously-undefined variables across `interface/billing/`. Where the offending variable was a legacy global (`$srcdir`, `$webroot`, `$web_root`, `$webserver_root`) or a `$GLOBALS` lookup, replace it with the typed `OEGlobalsBag` accessor (`getSrcDir()`, `getWebRoot()`, `getString()`, `getBoolean()`).

Cap in `.phpstan/fatal-baseline-caps.php` lowered from 2927 → 2586 to match.

Refs #11792.

## Files touched

`interface/billing/`:
billing_report, billing_tracker, edi_270, edi_271, edih_main, edih_view, edit_payment, era_payments, get_claim_file, indigent_patients_report, new_payment, payment_master.inc, payment_pat_sel.inc, print_billing_report, print_daysheet_report_num1/2/3, search_payments, sl_eob_invoice, sl_eob_search, sl_receipts_report, ub04_dispose, ub04_form, ub04_helpers.

## Test plan

- [x] `composer phpstan-baseline` regenerates with no new entries outside `variable.undefined.php` reductions
- [x] `composer phpstan` passes (pre-commit)
- [x] `composer rector-check` clean
- [x] `composer phpcs` clean
- [ ] CI green


interface/billing/ub04_form.php
```diff
@@ -204,7 +207,7 @@
 var ub04id = new Array();
 payerid = <?php echo js_escape($payerid ?? ''); ?>;
 pid = <?php echo js_escape($pid);?>;
-encounter = <?php echo js_escape($encounter ?? null);?>;
+encounter = <?php echo js_escape($encounter);?>;
 isTemplate = <?php echo js_escape((int) ($isAuthorized === true ? $isAuthorized : false)); ?>;
```
kojiromike: no value to js_escaping an int.

interface/billing/ub04_form.php
```diff
@@ -16,6 +16,9 @@
 
 /* $isAuthorized tells us if the form is for user UI or claim processing and provides another security check */
 global $isAuthorized;
```
kojiromike, line 18:
```suggestion
global $isAuthorized;
$isAuthorized = $isAuthorized === true ? 1 : 0;
```

interface/billing/ub04_form.php
kojiromike:
```suggestion
if ($isAuthorized === 1) {
    $imgurl = "../../../../public/images";  // should use a getKernel() call too!
} else {
    ub04_dispose();
    $pid = $_REQUEST['pid'] ?: '0';
    $encounter = $_REQUEST['enc'] ?: '0';
    $action = $_REQUEST['action'] ?? false ?: false;
    $payerid = $_REQUEST['id'] ?? '0' ?: '0';
    $imgurl = \OpenEMR\Core\OEGlobalsBag::getInstance()->getKernel()->getImagesRelative();
    if ($action === 'payer_defaults') {
        $ub04id = get_payer_defaults($payerid);
    } elseif ($pid && $encounter) {
        $ub04id = json_encode(get_ub04_array($pid, $encounter));
    } else {
        // raise an exception!
    }
}
```

interface/billing/billing_report.php
```diff
@@ -1367,8 +1373,8 @@ function criteriaSelectHasValue(select) {
                                             $rhtml2 .= "<td></td>\n";
                                         }
                                         if (!$iter['id'] && $rowcnt == 1) {
-                                            $rhtml2 .= "<td><input type='checkbox' value='0' name='claims[" . attr($this_encounter_id) . "][bill]' onclick='set_button_states()' id='CheckBoxBilling" . attr($CheckBoxBilling * 1) . "'>&nbsp;</td>\n";
-                                            $CheckBoxBilling++;
+                                            $rhtml2 .= "<td><input type='checkbox' value='0' name='claims[" . attr($this_encounter_id) . "][bill]' onclick='set_button_states()' id='CheckBoxBilling" . attr(($CheckBoxBilling ?? 0) * 1) . "'>&nbsp;</td>\n";
```
kojiromike: if we can guarantee $CheckBoxBilling is numeric, we don't need `attr` or any escaping on it.

interface/billing/edi_270.php
```diff
@@ -21,18 +21,19 @@
...
-require_once("$srcdir/calendar.inc.php");
-require_once("$srcdir/appointments.inc.php");
 
 use OpenEMR\Billing\EDI270;
 use OpenEMR\Common\Csrf\CsrfUtils;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Core\Header;
 use OpenEMR\Core\OEGlobalsBag;
 
+require_once(OEGlobalsBag::getInstance()->getSrcDir() . "/forms.inc.php");
```
kojiromike: repetitive calls to getSrcDir are repetitive

interface/billing/era_payments.php
```diff
@@ -33,6 +30,10 @@
 use OpenEMR\Core\OEGlobalsBag;
 use OpenEMR\OeUI\OemrUI;
 
+require_once(OEGlobalsBag::getInstance()->getSrcDir() . "/patient.inc.php");
+require_once(OEGlobalsBag::getInstance()->get('OE_SITE_DIR') . "/statement.inc.php");
```
kojiromike: is there not a typed call for this one?

interface/billing/era_payments.php
```diff
@@ -431,7 +432,7 @@ function OnloadAction()
         </div>
     </div><!-- End of Container Div-->
     <?php $oemr_ui->oeBelowContainerDiv();?>
-    <script src = '<?php echo $webroot;?>/library/js/oeUI/oeFileUploads.js'></script>
+    <script src = '<?php echo OEGlobalsBag::getInstance()->getString('webroot'); ?>/library/js/oeUI/oeFileUploads.js'></script>
```
kojiromike: is there not a call that returns where teh library or js files are?

interface/billing/new_payment.php
```diff
@@ -32,8 +29,20 @@
...
+require_once OEGlobalsBag::getInstance()->getSrcDir() . "/options.inc.php";
+require_once OEGlobalsBag::getInstance()->getSrcDir() . "/payment.inc.php";
+
 $session = SessionWrapperFactory::getInstance()->getActiveSession();
 
+// payment_master.inc.php (loaded later via require_once) reassigns these.
+// Initialize from request input so PHPStan sees them as `string` / `int` rather
+// than narrowing to literal '' / 0, which would mark downstream comparisons as
+// dead code.
+$PaymentType = (string) (filter_input(INPUT_POST, 'type_name') ?: filter_input(INPUT_GET, 'type_name') ?: '');
```
kojiromike: filter_input can return a typed result if you give it the right filters.

interface/billing/print_daysheet_report_num1.php
```diff
@@ -110,42 +112,28 @@
...
-            // reorder the list starting with array element zero
-                $user_final_list = array_values($user_list);
-                $provider_final_list = array_values($provider_list);
-            // sort array in ascending order
-                sort($user_final_list);
-                sort($provider_final_list);
                 $all4 = array_natsort($ret, 'pid', 'fulname', 'asc');
 
-                if ($_POST['end_of_day_provider_only'] == 1) {
+                if (($_POST['end_of_day_provider_only'] ?? 0) == 1) {
```
kojiromike: use filter_input here and get a typed result

interface/billing/print_daysheet_report_num2.php
```diff
@@ -478,233 +326,48 @@
...
-    $gtotal_patadj += $user_info['patadj'][$i];
-    $gtotal_patpay += $user_info['patpay'][$i];
-
-    ++$i;
+    Printf("<td width=70><span class=text><b>" . xlt("User ") . "</center></b><center>" . text($row['user'])) . "</center>";
+    Printf("<td width=140><span class=text><b><center>" . xlt("Charges") . ' ' . "</center></b><center>" . " %1\$.2f", text((string) $row['fee'])) . "</center>";
+    Printf("<td width=140><span class=text><b><center>" . xlt("Insurance Adj") . '. ' . "</center></b><center>" . "%1\$.2f", text((string) $row['insadj'])) . "</center>";
+    Printf("<td width=140><span class=text><b><center>" . xlt("Insurance Payments") . ' ' . "</center></b><center>" . "%1\$.2f", text((string) $row['inspay'])) . "</center>";
+    Printf("<td width=140><span class=text><b><center>" . xlt("Patient Adj") . '. ' . "</center></b><center>" . "%1\$.2f", text((string) $row['patadj'])) . "</center>";
+    Printf("<td width=140><span class=text><b><center>" . xlt("Patient Payments") . ' ' . "</center></b><center>" . "%1\$.2f", text((string) $row['patpay'])) . "</center>";
```
kojiromike: assign the variables first, using `sprintf` if you need to format something, then output via a single heredoc.

interface/billing/print_daysheet_report_num3.php
```diff
@@ -503,246 +352,61 @@
...
+        || $t['insadj'] != 0
+        || $t['patadj'] != 0
+        || $t['patpay'] != 0,
+));
 
 if ($totals_only != 1) {
     if ($line_total != 0) {
         print "<td class='w-100'><br /><span class='text'><strong><center>" . xlt('Total') . "</strong></center>";
-        Printf("<br /></span></td><td class='w-100'><span class='text'><center>" . " %1\$.2f", text($line_total)) . "</center></span></td>\n<br />";
+        Printf("<br /></span></td><td class='w-100'><span class='text'><center>" . " %1\$.2f", text((string) $line_total)) . "</center></span></td>\n<br />";
```
kojiromike: normalize the capitalization of functions. Narrow, don't cast!

interface/billing/print_daysheet_report_num3.php
```diff
@@ -503,246 +352,61 @@
...
         print "</tr><tr>\n";
     } else {
         print "<td class='w-100'><br /><span class='text'><strong><center>" . xlt('Total') . "</strong></center>";
-        Printf("<br /></span></td><td class='w-100'><span class='text'><center>" . " %1\$.2f", text($line_total_pay)) . "</center></td>\n<br />";
+        Printf("<br /></span></td><td class='w-100'><span class='text'><center>" . " %1\$.2f", text((string) $line_total_pay)) . "</center></td>\n<br />";
         print "</tr><tr>\n";
```
kojiromike: Don't repeat the `print "</tr><tr>\n";` on both halfs of an if/else -- move it out.

interface/billing/print_daysheet_report_num3.php
```diff
@@ -503,246 +352,61 @@
...
-    Printf("<td width='140'><span class='text'><strong><center>" . xlt("Patient Payments") . ' ' . "</center></strong><center>" . "%1\$.2f", text($user_info['patpay'][$i])) . "</center>";
-
-    $gtotal_fee += $user_info['fee'][$i];
-    $gtotal_insadj += $user_info['insadj'][$i];
-    $gtotal_inspay += $user_info['inspay'][$i];
-    $gtotal_patadj += $user_info['patadj'][$i];
-    $gtotal_patpay += $user_info['patpay'][$i];
-
-    ++$i;
+    Printf("<td width='70'><span class='text'><strong><center>" . xlt("User") . ' ' . "</center></strong><center>" . text($row['user'])) . "</center>";
```
kojiromike: same pattern as above.

interface/billing/sl_eob_search.php
```diff
@@ -1151,7 +1149,8 @@ class="fa fa-info-circle oe-superscript" aria-hidden="true"></i>
                                 // yet closed out insurance.
                                 //
                                 if (!$duncount) {
-                                    for ($i = 1; $i <= 3 && SLEOB::arGetPayerID($row['pid'], $row['date'], $i); ++$i) {
+                                    $i = 1;
+                                    for (; $i <= 3 && SLEOB::arGetPayerID($row['pid'], $row['date'], $i); ++$i) {
```
kojiromike: why move the $i = 1 out of the for?

interface/billing/ub04_form.php
kojiromike: Fixed — `$isAuthorized` is now normalized to `int` at the top of the file, so line 211 outputs it directly without `js_escape`.

interface/billing/ub04_form.php
kojiromike: Applied. Normalized `$isAuthorized` to `int` (1 or 0) at the top, then updated the three downstream comparisons to `=== 1`.

interface/billing/ub04_form.php
kojiromike: Applied the structural change — branch on `$isAuthorized === 1`, default `$pid`/`$encounter`/`$ub04id` with `??`. Held off on the `getKernel()->getImagesRelative()` swap and the exception-raising for the no-pid/encounter branch since those are behavior changes outside the variable.undefined drain; happy to file a follow-up if you'd like.

interface/billing/billing_report.php
kojiromike: Fixed. `$CheckBoxBilling` is now type-initialized to `0` before the foreach loop, so PHPStan knows it's `int` and the `attr()` wrapper / quoting are gone at the use sites.

interface/billing/edi_270.php
kojiromike: Hoisted to a single `$srcDir = OEGlobalsBag::getInstance()->getSrcDir();` at the top, then used `$srcDir` for the five `require_once` calls. Same hoist applied to edi_271.php, edih_main.php, edit_payment.php, new_payment.php, search_payments.php, sl_eob_invoice.php, and sl_eob_search.php.

interface/billing/era_payments.php
kojiromike: Yes — switched to `getString('OE_SITE_DIR')`. Same swap applied in sl_eob_search.php.

interface/billing/era_payments.php
kojiromike: Switched to `OEGlobalsBag::getInstance()->getWebRoot()` for the JS path.

interface/billing/new_payment.php
kojiromike: Dropped the casts. `$PaymentType` and `$TypeCode` use `filter_input(...) ?: filter_input(...) ?: ''` (the `?:` chain handles both `false` and `null`); `$CountIndexBelow` uses `FILTER_VALIDATE_INT` for a typed `int` result.

interface/billing/print_daysheet_report_num1.php
kojiromike: Switched both checks to `filter_input(INPUT_POST, '...', FILTER_VALIDATE_INT) === 1`.

interface/billing/print_daysheet_report_num2.php
kojiromike: Refactored. The foreach now `Printf`s each row, then the trailing totals/table block is built with `sprintf` into named variables and emitted via a single heredoc.

interface/billing/print_daysheet_report_num3.php
kojiromike: Fixed `Printf` → `printf` for the name-case violation, and replaced the cast on `$line_total`/`$line_total_pay` with a narrowed selection: `$totalValue = $line_total != 0 ? $line_total : $line_total_pay;`.

interface/billing/print_daysheet_report_num3.php
kojiromike: Hoisted the `</tr><tr>` outside the if/else so both branches just emit their cell.

interface/billing/print_daysheet_report_num3.php
kojiromike: Applied the same narrow-don't-cast + sprintf+heredoc pattern to the foreach loop and the grand-totals block at the bottom.

interface/billing/sl_eob_search.php
kojiromike: Fair point — that was an awkward shape. The original `for (; $i <= 3 && SLEOB::arGetPayerID(...); ++$i) {}` had an empty body with init outside the header, which reads strangely. Rewrote it as a `while` loop with the increment inside the body, which is clearer and still drains the baseline entry:

```php
$i = 1;
while ($i <= 3 && SLEOB::arGetPayerID($row['pid'], $row['date'], $i)) {
    ++$i;
}
```

Issue #665, opened 2017-04-19
Improve pdf output of forms


bradymiller:
Considering this strategy in all the forms to improve the pdf output:
https://github.com/openemr/openemr/pull/660

thoughts?

sjpadgett:
@bradymiller The mPDF composer load works fine except mPDF needs two writable directories tmp and ttfontdata. The default locations are` vendor/mpdf/mpdf/`. If moved then will need to define in bootstrap or code. If code I need to know where they will be. Maybe something like:
`define("_MPDF_TEMP_PATH", __DIR__.'/../mpdf/mpdf/tmp/');
define("_MPDF_TTFONTDATAPATH", __DIR__.'/../mpdf/mpdf/ttfontdata/');`(I think path is right) in the loader.

bradymiller:
To help out with figuring out the best strategy:
Can these paths be modified without touching the library in vendor directory?(ie. place them in interface/globals.php)
If so, then will make sense to place them in the sites/*/documents/mpdf/tmp and sites/*/documents/mpdf/ttfontdata dirs

sjpadgett:
Sure. I prefer to leave globals alone because of the potential troubles with permissions. I'll put in sites/*/documents/mpdf/ and define in scripts unless you prefer globals.

bradymiller:
Just put them in the scripts for now. We can figure out how to centralize it to globals.php (guessing will need to go after the composer autoload happens) later.

sjpadgett:
oh thought you meant paths. Nope, I can mod the globals and be done with it.

PR #11896, opened 2026-04-28
docs: plan for openemr release-automation conductor (openemr-devops#664)

kojiromike:
## Summary

- Adds `docs/release-automation-plan.md` documenting this repo's role as the conductor in the three-PR release-automation flow proposed in openemr/openemr-devops#664.
- Scope here is the long-lived release-prep PR per `rel-*` branch, the annotated tag on merge, and the `repository_dispatch` events emitted to website-openemr and openemr-devops.
- Sibling draft PRs in openemr/openemr-devops and openemr/website-openemr cover the infra and docs slices.
- **Implementation now landed alongside the plan; see "Implementation status" below.**

## Implementation status

Seven commits on top of the original plan-doc commits implement the conductor end-to-end. The PR remains a draft until consumer-side coordination completes (the App is already installed on this repo and the two consumer repos per openemr-devops#664).

### What landed

| # | Commit | Subject |
|---|--------|---------|
| 1 | `264b3a5` | `feat(release): vendor dispatch.schema.json from openemr-devops` |
| 2 | `4f45f40` | `feat(release): vendor TagVerifier from openemr-devops` |
| 3 | `676b208` | `feat(release): add openemr:release-prep console command for conductor` |
| 4 | `591c6bc` | `feat(release): add annotated-tag creation script for conductor merge` |
| 5 | `a113592` | `feat(release): add repository_dispatch script for cross-repo events` |
| 6 | `b594e49` | `ci(release): add release-prep conductor workflow + PR template` |
| 7 | `cbe4460` | `ci(release): add permissions-check workflow for the release App` |

### Layout

- `tools/release/` — sub-package mirroring openemr-devops `tools/release/` with its own composer (php ^8.2, justinrainbow/json-schema, symfony/console, symfony/http-client, symfony/process). Vendored contracts live at `contracts/dispatch.schema.json` and `src/{TagVerifier,TagVerificationResult}.php` — paths the openemr-devops drift checker (`tools/release/bin/check-vendored.php`) expects. Tests under `tests/`.
- `src/Common/Command/ReleasePrepCommand.php` plus `src/Common/Command/ReleasePrep/` (mutators, context, result, interface). The console runner auto-discovers it; the no-arg constructor + optional injection lets unit tests pass fake mutators.
- Tests under `tests/Tests/Isolated/Common/Command/ReleasePrep/` (15 isolated tests, fixture-driven, idempotence-checked).
- Workflows: `.github/workflows/release-prep.yml` (push on `rel-*`, merge handler for the conductor PR) and `.github/workflows/release-permissions-check.yml` (manual `workflow_dispatch`, probes installation membership / contents:write / pull-requests:write / annotated-tag creation / cross-repo dispatch).
- PR template at `.github/PULL_REQUEST_TEMPLATE/release-prep.md` with the irreducibly-manual release-manager checklist.

### Spec ambiguities resolved (documented in plan doc)

- **`_rest_routes.inc.php` → `src/RestControllers/OpenApi/OpenApiDefinitions.php`.** The wiki / earlier draft of this plan said the OpenAPI version constant lives in `_rest_routes.inc.php`, but it actually lives in `OpenApiDefinitions.php` line 17 (`#[OA\Info(title: 'OpenEMR API', version: '8.0.1')]`). Mutator targets the real location; plan updated.
- **`acknowledge_license_cert.html` not in conductor scope.** Wiki-rendered today (the file documents itself as built from the OpenEMR_Acknowledgments wiki page); openemr-devops#664's design moves acknowledgments to the docs PR (`website-openemr`) generated from `git shortlog vPREV..HEAD`. The conductor does not touch this file. Plan doc explains.
- **`contrib/util/installScripts/InstallerAuto.php` not in conductor scope.** The wiki checklist mentions "version refs" here, but a grep shows zero version references in the current code. Skipped.
- **`fsupgrade-N.sh` + upgrade Dockerfile not in conductor scope.** These live in `openemr-devops` under `docker/openemr/<version>/` and are handled by the rotation workflow that landed in openemr-devops#665. Plan doc explains.

### Test status

- **tools/release/**: 16 PHPUnit tests (29 with cross-suite total), 73 assertions. PHPStan level max + strict-rules clean. PHPCS PSR-12 clean. Rector-prepared-sets clean.
- **Application tests**: 15 new isolated tests (43 assertions) for the release-prep command and mutators. PHPStan filtered for changed files clean. PHPCS clean. Composer-require-checker clean.
- **Workflows**: both pass `actionlint`. The `release-prep.yml` was not exercised against a real `rel-*` branch — it fires automatically once merged.

### Maintainer follow-up after merge

1. Run `release-permissions-check` via `workflow_dispatch` to confirm the App's installed permissions match what the conductor needs (each probe fails loudly with the missing scope name).
2. Cut a synthetic `rel-test` (or wait for the next planned `rel-NNN` cut) to exercise the conductor end-to-end. The release-prep PR will appear at `release-prep/<rel-branch>` against the rel branch; merging it creates the annotated tag and fans out `openemr-tag` to the consumer repos.

### Known limitations

- **PHP-source mutators use anchored regex** rather than PHP-Parser. Each mutator targets a single, well-known location and asserts a single match, so silent miss is impossible (the mutator throws on no-match), but a future contributor changing the mutator targets to anything more complex should switch to PHP-Parser per `~/dev/oce/ocemr-docs/main/reference/php-conventions.md`. Conversion is mechanical if requested.
- **Master-branch flow not exercised.** `--scope=master` mutators (version bump + SQL-skeleton scaffold) are tested in isolation but the workflow only invokes `--scope=rel`. The intended master flow (a separate workflow trigger or manual invocation post-tag) is not yet wired up — leave for a follow-up once the rel flow has shipped one cycle.
- **`docker/production/docker-compose.yml` digest preservation.** When `--image-digest` is not supplied, the mutator pins the tag but preserves the existing digest. The workflow currently doesn't pass `--image-digest`; supplying it requires reading the published image's digest after build, which is a separate piece (could chain off the openemr-devops image-build workflow). Left as a follow-up; the produced compose file is still pinned to the version tag, which is the load-bearing change.

## Test plan

- [ ] Review the conductor's mutator coverage against the wiki release checklist for any items I missed.
- [ ] Confirm the dispatch event names + payloads still match the openemr-devops consumer (PR #665 merged).
- [ ] Confirm the workflow's App-token usage is correct for the `RELEASE_APP_ID` / `RELEASE_APP_PRIVATE_KEY` secrets the maintainer installed.
- [ ] After merge: run `release-permissions-check` via `workflow_dispatch`; confirm green.
- [ ] (Optional smoke) Trigger `release-prep.yml` via `workflow_dispatch` against the current `rel-810` branch with `target-version=8.1.0`; confirm a draft PR appears at `release-prep/rel-810`.


.github/workflows/release-permissions-check.yml
```diff
@@ -0,0 +1,151 @@
...
+jobs:
+  probe:
+    name: Probe release-app permissions (${{ matrix.php-version }})
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    strategy:
+      matrix:
+        php-version:
+        - '8.2'
```
kojiromike: Tasks that check the openemr runtime code should use php 8.2. Tasks that are for tooling runtime itself should use the latest production php version.

.github/workflows/release-permissions-check.yml
```diff
@@ -0,0 +1,151 @@
...
+    - name: Install conductor sub-package deps
+      working-directory: tools/release
+      run: composer install --no-interaction --prefer-dist --no-progress
+
+    - name: Probe installation membership
+      env:
+        GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        REPO: ${{ github.repository }}
+      run: |
+        if ! gh api /installation/repositories --jq '.repositories[].full_name' | grep -Fxq "${REPO}"; then
```
kojiromike: I think jq can take over the job from grep and you can avoid the pipeline here.

.github/workflows/release-permissions-check.yml
```diff
@@ -0,0 +1,151 @@
...
+    - name: Probe contents:write (create + delete throwaway branch)
+      id: probe-contents
+      env:
+        GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        REPO: ${{ github.repository }}
+        RUN_ID: ${{ github.run_id }}
+      run: |
+        branch="release-permissions-check/${RUN_ID}"
+        printf 'branch=%s\n' "${branch}" >> "$GITHUB_OUTPUT"
+        sha="$(gh api "repos/${REPO}/git/ref/heads/$(gh api "repos/${REPO}" --jq '.default_branch')" --jq '.object.sha')"
```
kojiromike: Too nested. If the default branch isn't available from repo metadata, create an earlier step to set the default branch in output. Do this in general in favor of multi-line scripts in steps.

.github/workflows/release-permissions-check.yml
```diff
@@ -0,0 +1,151 @@
...
+        if ! gh api -X POST "repos/${REPO}/git/refs" -f "ref=refs/heads/${branch}" -f "sha=${sha}" >/dev/null; then
+          printf 'FAIL: contents:write missing (cannot create branch)\n' >&2
+          exit 1
+        fi
+        printf 'OK: created throwaway branch %s\n' "${branch}"
```
kojiromike: This isn't a good case for printf. echo with interpolation is fine here.

.github/workflows/release-permissions-check.yml
```diff
@@ -0,0 +1,151 @@
...
+        BRANCH: ${{ steps.probe-contents.outputs.branch }}
+      run: |
+        default_branch="$(gh api "repos/${REPO}" --jq '.default_branch')"
+        body='Permissions probe — auto-closed.'
+        pr_number="$(gh api -X POST "repos/${REPO}/pulls" -f "title=release-permissions-check probe" -f "head=${BRANCH}" -f "base=${default_branch}" -f "body=${body}" -F "draft=true" --jq '.number' 2>/dev/null || true)"
+        if [[ -z "${pr_number}" ]]; then
+          printf 'FAIL: pull-requests:write missing (cannot open draft PR)\n' >&2
+          exit 1
+        fi
+        printf 'pr_number=%s\n' "${pr_number}" >> "$GITHUB_OUTPUT"
```
kojiromike: not a good case for printf. Printf is useful when it helps you avoid very complex interpolations, when a string must start with a hyphen, when you need to deal with nested quotes, and when you need formatting. This is none of those things. Echo with interpolation is the better choice here.

.github/workflows/release-permissions-check.yml
```diff
@@ -0,0 +1,151 @@
...
+        if ! gh api -X PATCH "repos/${REPO}/pulls/${pr_number}" -f 'state=closed' >/dev/null; then
+          printf 'FAIL: pull-requests:write missing (cannot close draft PR)\n' >&2
```
kojiromike: no reason to use printf here.

.github/workflows/release-permissions-check.yml
```diff
@@ -0,0 +1,151 @@
...
+        gh api -X DELETE "repos/${REPO}/git/refs/tags/${TAG}" || true
+
+    - name: Cleanup throwaway branch
+      if: always() && steps.probe-contents.outputs.branch != ''
+      env:
+        GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        REPO: ${{ github.repository }}
+        BRANCH: ${{ steps.probe-contents.outputs.branch }}
+      run: |
+        gh api -X DELETE "repos/${REPO}/git/refs/heads/${BRANCH}" || true
```
kojiromike: using || true hides the result of this command no matter what, so if it fails every time it'll appear green. Is there a way to tell github that if this step fails it shouldn't fail the job, but still allow this step to appear as a failure if that's the case?

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,239 @@
...
+  prep:
+    name: Generate release-prep PR (${{ matrix.php-version }})
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    strategy:
+      matrix:
+        php-version:
+        - '8.2'
```
kojiromike: 8.5

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,239 @@
...
+
+    - name: Resolve target version and branch
+      id: resolve
+      env:
+        EVENT_NAME: ${{ github.event_name }}
+        REF: ${{ github.ref }}
+        INPUT_BRANCH: ${{ inputs.branch }}
+        INPUT_VERSION: ${{ inputs.target-version }}
+      run: |
+        if [[ "${EVENT_NAME}" == 'workflow_dispatch' ]]; then
```
kojiromike: Don't use double-equals in bash outside of arithmetic context.

```suggestion
        if [[ "${EVENT_NAME}" = 'workflow_dispatch' ]]; then
```

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,239 @@
...
+          branch="${INPUT_BRANCH}"
+          version="${INPUT_VERSION}"
+        else
+          branch="${REF#refs/heads/}"
+          version="$(php tools/release/bin/branch-to-version.php "${branch}")"
+        fi
+        printf 'branch=%s\n' "${branch}" >> "$GITHUB_OUTPUT"
+        printf 'version=%s\n' "${version}" >> "$GITHUB_OUTPUT"
```
kojiromike:
```suggestion
        case "${EVENT_NAME}" in
            workflow_dispatch)
                echo "branch=${INPUT_BRANCH}"
                echo "version=${INPUT_VERSION}"
                ;;
            *)
                echo "branch=${REF#refs/heads/}"
                printf 'version='
                php tools/release/bin/branch-to-version.php "${branch}"
                ;;
        esac >> "$GITHUB_OUTPUT"
```

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,239 @@
...
+          --skip-globals
+
+    - name: Detect changes
+      id: diff
+      run: |
+        if git diff --quiet; then
+          printf 'changed=false\n' >> "$GITHUB_OUTPUT"
+        else
+          printf 'changed=true\n' >> "$GITHUB_OUTPUT"
+        fi
```
kojiromike:
```suggestion
        if git diff --quiet; then
          echo 'changed=false'
        else
          echo 'changed=true'
        fi >> "$GITHUB_OUTPUT"
```

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,239 @@
...
+        GH_TOKEN: ${{ steps.app-token.outputs.token }}
+      run: |
+        rel_branch='${{ steps.resolve.outputs.branch }}'
+        version='${{ steps.resolve.outputs.version }}'
+        pr_branch="release-prep/${rel_branch}"
+        git add -A
+        git commit -m "release-prep: ${version}"
+        git push --force "https://x-access-token:${GH_TOKEN}@github.com/${{ github.repository }}.git" "HEAD:refs/heads/${pr_branch}"
+
+    - name: Open or update draft PR
```
kojiromike: Is there a reliable and well-maintained third party action for this? Try not to reinvent the wheel. Research what's already done for us.

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,239 @@
...
+      run: composer install --no-interaction --prefer-dist --no-progress
+
+    - name: Resolve version from base branch
+      id: resolve
+      env:
+        BASE_REF: ${{ github.event.pull_request.base.ref }}
+      run: |
+        version="$(php tools/release/bin/branch-to-version.php "${BASE_REF}")"
+        printf 'branch=%s\n' "${BASE_REF}" >> "$GITHUB_OUTPUT"
+        printf 'version=%s\n' "${version}" >> "$GITHUB_OUTPUT"
```
kojiromike:
```suggestion
        {
            echo "branch=$BASE_REF"
            printf 'version='
            php tools/release/bin/branch-to-version.php "${BASE_REF}"
        } >> "$GITHUB_OUTPUT"
```

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,239 @@
...
+          --commit-sha="${MERGE_SHA}" \
+          --conductor-pr-url="${PR_URL}"
+
+    - name: Dispatch openemr-tag to consumers
+      env:
+        RELEASE_APP_TOKEN: ${{ steps.app-token.outputs.token }}
+      run: |
+        version='${{ steps.resolve.outputs.version }}'
+        rel_branch='${{ steps.resolve.outputs.branch }}'
+        tag_name="v$(printf '%s' "${version}" | tr '.' '_')"
```
kojiromike:
```suggestion
        tag_name="v${version//./_}"
```

Avoid unnecessary pipelines and subshells.

src/Common/Command/ReleasePrep/Mutator/GlobalsIncMutator.php
```diff
@@ -0,0 +1,77 @@
...
+        //     'allow_debug_language' => [
+        //         xl('Allow Debugging Language'),
+        //         'bool',                           // data type
+        //         '1',                              // default = true during development...
+        //
+        // Match the entire block so the replacement is anchored on the
+        // 'allow_debug_language' key, then swap '1' → '0' on the third
+        // line of that block.
+        $pattern =
+            '/(\'allow_debug_language\'\s*=>\s*\[\s*'
```
kojiromike: is it really a good idea to use regex here? We have a php parser in the repo for phpstan and rector.

tools/release/phpcs.xml
kojiromike, line 1: isn't the repo root's phpcs going to pick up all this stuff? I'm not super comfortable mixing different coding principles in a nested part of the repo.

kojiromike (commented):
Especially for the mutators: Try to find a better way than regex. There are php parsers, yaml parsers, and even semi-specializer version bumping tools like updatecli out there.

In general, try not to invent the wheel. Pick open source solutions that already exist as long as they're well-regarded and well-maintained.

.github/workflows/release-permissions-check.yml
kojiromike: Done — bumped the permissions-check matrix to PHP 8.5. The probes don't run any openemr application code (just `gh` calls and `tools/release/bin/dispatch.php`), so per the conventions this counts as tooling. Commit `1896276`.

.github/workflows/release-permissions-check.yml
kojiromike: Done — replaced `gh api ... | grep -Fxq` with a `jq select` filter inside the `gh api` call. No second process, no exit-status combo. Commit `1896276`.

.github/workflows/release-permissions-check.yml
kojiromike: Done — switched to `echo`. Commit `1896276`.

.github/workflows/release-permissions-check.yml
kojiromike: Done — switched to `echo`. Commit `1896276`.

.github/workflows/release-permissions-check.yml
kojiromike: Done — switched to `echo`. Commit `1896276`.

.github/workflows/release-permissions-check.yml
kojiromike: Done — both cleanup steps now use `continue-on-error: true` instead of `|| true`. The step shows red on failure but doesn't fail the run, which is exactly the behavior you wanted. Commit `1896276`.

.github/workflows/release-prep.yml
kojiromike: Done — bumped both jobs to PHP 8.5 in commit `33c4140`. Same rationale as the permissions-check workflow: this is conductor tooling (parses application code via AST but never executes it on this PHP version), so it tracks the latest stable rather than the application floor.

.github/workflows/release-prep.yml
kojiromike: Done — switched to `=`. Commit `1896276`.

.github/workflows/release-prep.yml
kojiromike: Done — applied your suggested case-statement pattern with the redirect group. Commit `1896276`.

.github/workflows/release-prep.yml
kojiromike: Resolved by replacing the manual diff/force-push/PR-create logic with `peter-evans/create-pull-request@v7`, which handles diff detection, branch force-push, and PR open-or-update in one step. The `id: diff` step (and the others your suggestion would have rewritten) are gone entirely. Commit `1896276`.

.github/workflows/release-prep.yml
kojiromike: Done — switched to [`peter-evans/create-pull-request@v7`](https://github.com/peter-evans/create-pull-request). 2.7k stars, actively maintained, the de-facto standard for "open or update a long-lived PR with on-disk diff". Removes ~25 lines of bespoke shell. Commit `1896276`.

.github/workflows/release-prep.yml
kojiromike: Done — applied your suggested redirect group. Commit `1896276`.

.github/workflows/release-prep.yml
kojiromike: Done — applied your suggested parameter expansion. Commit `1896276`.

src/Common/Command/ReleasePrep/Mutator/GlobalsIncMutator.php
kojiromike: Done — rewrote `VersionPhpMutator`, `VersionPhpMasterMutator`, `OpenApiVersionMutator`, and `GlobalsIncMutator` on top of nikic/php-parser via a small `AstSourceEditor` helper that locates the target node, then byte-substitutes its source range. This preserves all surrounding formatting (comments, indentation, etc.) while turning a structural mismatch into a hard parse error rather than a silent regex miss.

Added `nikic/php-parser ^5.7` to root `composer.json` `require` (it was already a transitive dep via PHPStan/Rector, so no new vendor footprint). The conductor-only `OpenEMR\Release\` autoload mapping stays in `autoload-dev` so production composer-require-checker still passes.

`DockerComposeProductionMutator` stays on regex deliberately (commented inline): YAML round-tripping through symfony/yaml would drop comments, normalise quote styles, and reorder keys — undesirable for a file maintainers read. The pattern is single-line and asserts a match.

Commit `4f574e8`.

tools/release/phpcs.xml
kojiromike: Done — went bigger than just phpcs after your follow-up clarified that the whole sub-toolchain (phpcs + phpstan + rector + phpunit + composer) was the concern. Deleted the entire `tools/release/{composer.json, composer.lock, phpcs.xml, phpstan.neon, phpunit.xml.dist, rector.php, .gitignore, vendor/}` set. Runtime code stays at `tools/release/{src,bin}` (which is `export-ignore`'d in `.gitattributes` so it never shipped to users either way), but is now picked up by the root composer/phpstan/phpcs/rector/phpunit configs. Tests moved to `tests/Tests/Isolated/Release/` so the existing root `phpunit-isolated.xml` runs them.

Added `justinrainbow/json-schema ^6.8` to root `require-dev` (the only direct dep that wasn't already in root). The `OpenEMR\Release\` PSR-4 mapping went under `autoload-dev` so composer-require-checker doesn't demand the conductor-only deps in production.

Commit `868faf6`.

.github/workflows/release-permissions-check.yml
kojiromike: Done — extracted as a dedicated `id: repo` step that writes `default-branch` to `GITHUB_OUTPUT`. Later steps reference the output instead of nesting `gh api` calls inside other command substitutions. Commit `1896276`.

kojiromike:
Addressed in commits `868faf6` (toolchain consolidation), `1896276` (workflow polish), `4f574e8` (AST mutators), `33c4140` (PHP 8.5 on release-prep). Highlights:

- The four PHP-source mutators (`VersionPhp`, `VersionPhpMaster`, `OpenApiVersion`, `GlobalsInc`) now use `nikic/php-parser` via a small `AstSourceEditor` helper. Each one walks the AST to locate the target node, then byte-substitutes the source range — preserving comments and surrounding formatting while turning a structural mismatch into a hard parse error rather than a silent regex miss.
- `nikic/php-parser ^5.7` added to root `require` (already a transitive dep, so no new vendor footprint).
- `DockerComposeProductionMutator` stays on regex deliberately, with an inline comment: YAML round-tripping through symfony/yaml would drop comments, normalise quote styles, and reorder keys — undesirable for a file maintainers read. The pattern is single-line and asserts a match.

I considered `updatecli` but didn't pull it in — it's a declarative-YAML-manifest tool that's a paradigm shift for a 6-mutator surface, and the AST approach gave us per-mutator unit tests in the same isolated phpunit suite as the rest of the conductor.

Re-request review pending the new push: 4 commits on top of the original stack, all green (52 isolated tests / 140 assertions, PHPStan level 10 + strict-rules clean across 4311 files, PHPCS / Rector / composer-require-checker clean, both workflows pass actionlint).

PR #11906, opened 2026-04-29
fix(sql): Allow sql_upgrade to work on the cli

Firehed:
While sql_upgrade.php has some existing support for running in CLI mode (SAPI detection), the checks that were in place were mostly rendering-related; there was never actually a way to tell it what version to upgrade from, so executing it never did anything.

This is needed for the Inferno test suite - as of today, it relies on a (outdated) schema dump and then runs against the current codebase, which is part of the reason that some tests fail.

Before:
```
Runtime:       PHP 8.5.4 with PCOV 1.0.12
Configuration: /var/www/localhost/htdocs/openemr/phpunit.xml

EEEEFFEEEFF.FF.FFFFFFFFFFFFEFFEFFF...FE.

Time: 01:11.365, Memory: 127.00 MB
...
ERRORS!
Tests: 40, Assertions: 312, Errors: 9, Failures: 25, PHPUnit Warnings: 1, Warnings: 1.
```

After:
```
FFF..FF.......FF..FF.F...E.FE..F....E.                         41 / 41 (100%)
Time: 08:58.933, Memory: 119.00 MB
...
ERRORS!
Tests: 41, Assertions: 846, Errors: 3, Failures: 14, PHPUnit Warnings: 1.
```

This is a hacky approach; we're going to be using doctrine/migrations soon which will obsolete this entirely. Otherwise I'd invest more in separating concerns and tidying this up. It still dumps a bunch of HTML to the console. I don't really care 🤷 

Extracted from #11805

- Look for `--from=X` as a CLI arg and hack it through to the upgrade tool
- Force site=default

Claude wrote the fix, I extracted it from the other branch.


kojiromike:

This is mostly Inferno-driven — most operators don't invoke sql_upgrade from the CLI. Backport only if CLI parity is wanted.

PR #447, opened 2017-01-24
patient portal fix

bradymiller:
This is proposed fix to address the folllowing bug:
https://github.com/openemr/openemr/issues/444

Still need to test this.

This fix will also need to go into rel-500 in order to support the new patient portal in the future.

PR #11907, opened 2026-04-29
fix(faxsms): catch up missed appointment-reminder ticks

kojiromike:
## Summary

`NotificationTaskManager::isWithinCronWindow` enforces both an upper *and* a lower bound on when a reminder is eligible to send: the ideal send time must be within `±cronIntervalHours` of the current tick. That means any gap longer than the cron interval — pod restart, deploy, lead-time change in admin settings, even just a heavily loaded host that misses one tick — silently drops the reminder forever, because the next tick already considers the event "too late".

This PR drops the lower bound. A reminder more than `cronIntervalHours` hours before its ideal send time is still too early, but a reminder past its ideal moment will now send on the next run instead of being skipped. Existing dedup keeps it from sending twice:

- The SQL filter in `faxsms_getAlertPatientData` already excludes events whose `pc_sendalertemail`/`pc_sendalertsms` is `'YES'` (set by `markNotified` after a successful send).
- Recurring events are additionally checked against `notification_log` keyed on `(pc_eid, pc_eventDate, type)`.

To make sure removing the lower bound can't surface a misleading reminder for a past appointment, `AppointmentNotificationRunner::isInCronWindow` now also skips any event whose start time is at or before "now". A stale row that never got a reminder will be quietly left alone once the appointment has begun.

### Files

- `interface/modules/custom_modules/oe-module-faxsms/src/Controller/NotificationTaskManager.php` — drop the lower bound, expand the docblock with the rationale and the dedup mechanisms it relies on.
- `interface/modules/custom_modules/oe-module-faxsms/src/Notification/AppointmentNotificationRunner.php` — add the past-appointment guard.
- `tests/Tests/Isolated/Modules/FaxSMS/Controller/NotificationTaskManagerTest.php` — replace the now-invalid lower-bound test cases with positive cases asserting that late ticks still send.
- `tests/Tests/Isolated/Modules/FaxSMS/Notification/AppointmentNotificationRunnerTest.php` — add `testRunCatchesUpReminderForLateAppointmentStillInFuture` and `testRunSkipsAppointmentThatHasAlreadyStarted`.

## Test plan

- [x] `phpunit-isolated.xml --filter 'NotificationTaskManager|AppointmentNotificationRunner'` passes (45 tests / 89 assertions)
- [x] `phpstan analyse` on the four touched files reports no errors
- [x] PHP syntax check on all four files
- [ ] Reviewer: confirm that for an install whose `Notification_Email_Task` cron interval is 1h and `EMAIL_NOTIFICATION_HOUR` is 20, an appointment that previously missed its 17:15 tick (because the ideal send moment was 17:00 but it arrived 15 min late) sends on the next 18:15 tick instead of being silently dropped

### DIFF
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/NotificationTaskManager.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/NotificationTaskManager.php
@@ -37,7 +37,7 @@ class NotificationTaskManager
     }
 
     /**
-     * Check whether a reminder falls within the cron send window.
+     * Check whether a reminder is eligible to send on this tick.
      *
      * Both parameters use whole-hour granularity. $remainHour is the
      * difference between hours-until-appointment and the configured
@@ -46,15 +46,26 @@ class NotificationTaskManager
      *
      * $cronIntervalHours is the background-service execution interval
      * converted to hours (via getTaskHours()). Values below 1 are clamped
-     * to 1 so the window is never zero-width.
+     * to 1 so the upper bound is never zero-width.
+     *
+     * Only the upper bound is enforced here: a reminder more than
+     * $cronIntervalHours hours before its ideal send time is too early.
+     * No lower bound is applied so a missed tick (pod restart, deploy,
+     * lead-time change) does not silently drop the reminder. Dedup
+     * prevents repeats: the SQL filter excludes events whose
+     * pc_sendalertemail/pc_sendalertsms is already 'YES', and recurring
+     * events are additionally checked against notification_log keyed on
+     * (pc_eid, pc_eventDate, type). The runner is responsible for
+     * skipping appointments that have already started.
      */
     public static function isWithinCronWindow(int $remainHour, int $cronIntervalHours): bool
     {
-        // Enforce a minimum 1-hour window. getTaskHours() returns int, so
-        // sub-hour intervals (e.g. 30 minutes) truncate to 0 — which would
-        // make the window zero-width and silently suppress all sends.
+        // Clamp to a minimum 1-hour window. getTaskHours() returns int, so
+        // sub-hour intervals (e.g. 30 minutes) truncate to 0; without the
+        // clamp, an interval of 0 would only fire on the exact tick where
+        // remainHour == 0 and silently drop everything else.
         $cronIntervalHours = max(1, $cronIntervalHours);
-        return $remainHour >= -$cronIntervalHours && $remainHour <= $cronIntervalHours;
+        return $remainHour <= $cronIntervalHours;
     }
 
     /**
--- a/tests/Tests/Isolated/Modules/FaxSMS/Controller/NotificationTaskManagerTest.php
+++ b/tests/Tests/Isolated/Modules/FaxSMS/Controller/NotificationTaskManagerTest.php
@@ -25,8 +25,36 @@ require_once __DIR__ . '/../../../../../../interface/modules/custom_modules/oe-m
 
 class NotificationTaskManagerTest extends TestCase
 {
-    #[DataProvider('cronWindowProvider')]
-    public function testIsWithinCronWindow(int $remainHour, int $cronInterval, bool $expected): void
+    /**
+     * Upper bound: a reminder more than $cronIntervalHours hours before
+     * its ideal send time is too early. The third column varies row by
+     * row because that is the behavior under test.
+     */
+    #[DataProvider('upperBoundProvider')]
+    public function testIsWithinCronWindowEnforcesUpperBound(int $remainHour, int $cronInterval, bool $expected): void
+    {
+        $this->assertSame($expected, NotificationTaskManager::isWithinCronWindow($remainHour, $cronInterval));
+    }
+
+    /**
+     * No lower bound: a missed tick must still send on the next run.
+     * The runner's per-event guard skips appointments that have already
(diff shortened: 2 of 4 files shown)


tests/Tests/Isolated/Modules/FaxSMS/Controller/NotificationTaskManagerTest.php
```diff
@@ -39,21 +39,25 @@ public function testIsWithinCronWindow(int $remainHour, int $cronInterval, bool
...
-            'within 24h window (negative)'       => [-12, 24, true],
-            'at boundary of 24h window'          => [24, 24, true],
-            'at negative boundary of 24h window' => [-24, 24, true],
-            'outside 24h window'                 => [25, 24, false],
-            'far outside window'                 => [100, 24, false],
-            'old 150h window was a no-op'        => [100, 150, true],
-            'zero interval clamps to 1h'         => [1, 0, true],
-            'zero interval rejects 2h early'     => [2, 0, false],
-            'negative interval clamps to 1h'     => [0, -5, true],
+            // Upper bound is enforced: do not fire too early.
```
kojiromike: Could this be an enum?

kojiromike:
@aisle-analyzer

tests/Tests/Isolated/Modules/FaxSMS/Controller/NotificationTaskManagerTest.php
kojiromike: Considered three enum interpretations:

1. Enum for the `bool` expected value (`Decision::Send`/`Skip`) — pure ceremony; the function returns `bool` and callers gate on it.
2. Enum tagging each row's category (`Category::UpperBound`/`CatchUp`/`Clamp`) — PHPUnit doesn't act on category metadata, so it just duplicates what the comment groupings already say to a human reader.
3. Split into one data provider + test method per behavior bucket (`testIsWithinCronWindowEnforcesUpperBound`, `testIsWithinCronWindowAlwaysAcceptsCatchUpTicks`, `testIsWithinCronWindowClampsSubHourIntervals`) — the test method name carries the category meaning, which is the closest "enum-like" structuring that PHPUnit can actually consume.

Went with #3 in e5bebf6. The win that surfaced doing it: the catch-up rows always expected `true`, so the third column was dead weight. The catch-up provider drops it entirely and the test asserts every row passes — which is the actual invariant under test, not "row N happens to evaluate to row N's hardcoded expected value". Same 14 cases, all green.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/NotificationTaskManager.php
```diff
@@ -46,15 +46,26 @@ public function getTaskHours($type): int
...
-        // sub-hour intervals (e.g. 30 minutes) truncate to 0 — which would
-        // make the window zero-width and silently suppress all sends.
+        // Clamp to a minimum 1-hour window. getTaskHours() returns int, so
+        // sub-hour intervals (e.g. 30 minutes) truncate to 0; without the
+        // clamp, an interval of 0 would only fire on the exact tick where
+        // remainHour == 0 and silently drop everything else.
         $cronIntervalHours = max(1, $cronIntervalHours);
-        return $remainHour >= -$cronIntervalHours && $remainHour <= $cronIntervalHours;
+        return $remainHour <= $cronIntervalHours;
     }
```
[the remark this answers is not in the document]
kojiromike: Thanks aisle-research-bot. Considered the finding; declining the recommendations because the framing overstates the impact and both fixes would re-introduce the bug this PR exists to fix.

**On "unbounded":** the burst is bounded by `appointment_volume × lead_time_window`. For a typical clinic (~50 appts/day, 24h lead time), the worst case after a full-day cron outage is ~50 sends per channel. Sinch REST is ~200ms/req → ~10s; SMTP is ~1–2s/msg → 50–100s. Neither hits a meaningful service rate limit and the *total* SMS/email spend is unchanged because each appointment still fires exactly once (dedup is via `pc_sendalertemail`/`pc_sendalertsms = 'YES'` in `markNotified`, which is the SQL filter the next run reads).

**On atomic claim / concurrency:** valid pre-existing observation, but not new in this PR. The mitigation already in place is `background_services.lock_expires_at` (#11687 / oce-internal #447), which gives the row-level lock that prevents two concurrent runs of the same task task-function. Within a single runner the foreach is sequential, so no intra-process race.

**On the recommendations specifically:**

- Option 1 (restore the lower bound) is the bug this PR fixes. The original `>= -$cronIntervalHours` is exactly what silently dropped reminders whenever a tick was missed.
- Option 2 (cap max lateness at 2h) defeats the overnight / weekend / deploy-window / ArgoCD-sync / lead-time-change cases. Those routinely exceed 2h. The motivating real-world miss for this PR was 1h late (next tick caught it); the next class of misses we expect to handle is 8–14h overnight gaps. A 2h cap silently drops the latter, which is the regression I'm here to prevent.

The genuine residual concern — "is a 12h-late reminder useful to the patient?" — is a UX question, not a security/resource one, and the right place to address it is per-channel (e.g., let admins set a min-useful-lead per template). Out of scope for this PR.

Resolving.

kojiromike:
Per-address dedup, per-tick cap, ordering, and a deferred-work warning are tracked as a follow-up at #11911. Keeping this PR scoped to the catch-up fix.

PR #11912, opened 2026-04-29
feat(auth): add audit logging for failed TOTP, U2F, and OAuth2 MFA attempts

aanand-1706:
<!--
Thanks for sending a pull request!

Your PR title must follow Conventional Commits: type(scope): description
Examples: feat(api): add patient search, fix(calendar): correct date parsing
See CONTRIBUTING.md for details

Please provide context about the nature of your change so that reviewers understand the motivation and we can generate changelogs.

If this resolves an existing issue, link to one or more issues in the short description section, e.g. `Fixes #12345`.

-->

Failed MFA authentication attempts (TOTP, U2F, OAuth2) were not logged, violating ASVS 16.3.1 (CWE-223: Omission of Security-relevant Information). Only successful logins and password failures were recorded in the audit log. 

  - Log failed TOTP authentication attempts in main_screen.php                                                                                                                              
  - Log failed U2F authentication attempts in main_screen.php
  - Log failed OAuth2 MFA attempts in AuthorizationController.php                                                                                                                           
                                                                                                                                                                                            
  Each log entry includes the username, auth group, client IP, and MFA type.   

##### Testing                                                                                                                                                                                   
                                                                                                                                                                                          
  1. Enable TOTP MFA for a user via Admin → MFA Management                                                                                                                                  
  2. Log in and intentionally enter wrong TOTP codes
  3. Check Admin → System → Logs (Event type: All, search "mfa")                                                                                                                          
  4. Confirm failed TOTP attempts appear alongside successful login entries with MFA type metadata   

#### Was an AI assistant used? 
Yes 
<!--
If yes, add an `Assisted-by` trailer to each commit that used AI assistance:

```
git commit --trailer "Assisted-by: Claude Code" -m "fix(calendar): correct date parsing"
```

Use the name of the specific tool (e.g. `GitHub Copilot`, `Claude Code`, `ChatGPT`).
If the AI made the commit(s), this trailer was probably added automatically.
-->

#### Notes                                   
       
This PR introduces mfa as a new event type in the audit log (via `AuthEvent::mfa()`). Any existing dashboards, saved queries, or log filters that rely on specific event values should be updated to include mfa. For example, a query previously filtering on event = 'login' to catch all auth failures will miss MFA-specific failures unless extended to also match event = 'mfa'.                                                                                                                                                                                    
         
### Release notes line (for CHANGELOG.md):                                                                                                                                                    
                                                                                                                                                                                            
#### Security                                                                                                                                                                             
Audit log now records failed MFA authentication attempts (TOTP, U2F, and OAuth2) as `mfa` events, addressing ASVS 16.3.1 / CWE-223 (omission of security-relevant information)


interface/main/main_screen.php
```diff
@@ -234,6 +235,22 @@ function generate_html_end()
                         [$session->get('authUserID')]
                     );
                 } else {
+                    // Log failed TOTP authentication attempt
+                    $ip = collectIpAddresses();
+                    $username = $_POST['authUser'] ?? '';
```
kojiromike: `$_POST['authUser']` is attacker-controlled on this POST — the MFA form is a separate submission from the password form, and the hidden `authUser` field can be tampered with before re-submitting. The user whose MFA is actually being checked is `$userid = $session->get('authUserID')` (set on line 178), which is the value the TOTP secret was just looked up against. Resolving the username from `$userid` would make the audit entry trustworthy:

```suggestion
                    $userRow = privQuery('SELECT username FROM users WHERE id = ?', [$userid]);
                    $username = (is_array($userRow) && isset($userRow['username'])) ? (string) $userRow['username'] : '';
```

Same issue applies to the U2F block below.

interface/main/main_screen.php
```diff
@@ -234,6 +235,22 @@ function generate_html_end()
...
+                    $userInfo = privQuery("SELECT id FROM users WHERE BINARY username = ?", [$username]);
+                    $authGroup = '';
+                    if (!empty($userInfo['id'])) {
```
kojiromike: Two issues on this guard:

1. The repo CLAUDE.md (Coding Standards → Legacy Code Is Not the Standard) specifically calls out `empty()` as a pattern not to introduce in new code — replace with an explicit check.
2. `privQuery()` returns `array|false`, so `$userInfo['id']` will trigger a notice/PHPStan error when the lookup misses. This becomes moot if you switch to resolving the username from `$userid` (no second lookup needed before calling `getAuthGroupForUser`).

interface/main/main_screen.php
```diff
@@ -234,6 +235,22 @@ function generate_html_end()
...
+                        $userService = new \OpenEMR\Services\UserService();
+                        $authGroup = $userService->getAuthGroupForUser($username);
+                    }
+                    EventAuditLogger::getInstance()->newEvent(
+                        'login',
+                        $username,
+                        $authGroup,
```
kojiromike: `UserService::getAuthGroupForUser()` returns `string|bool` (literally `false` when the user is not in a group — see `src/Services/UserService.php:76`). That `false` gets passed straight to `EventAuditLogger::newEvent()` as the `$groupname` parameter, which is declared `string`. At PHPStan level 10 this is a type error; at runtime it coerces to `''`. Either coerce here (`$authGroup = $authGroup === false ? '' : $authGroup;`) or initialize to `''` and only assign on truthy non-false return. Same applies in the U2F and OAuth2 blocks.

interface/main/main_screen.php
```diff
@@ -266,6 +283,22 @@ function generate_html_end()
                 } catch (\u2flib_server\Error $e) {
                     // Authentication failed so we will build the U2F form again.
                     $form_response = '';
+                    // Log failed U2F authentication attempt
```
kojiromike: This block is a near-duplicate of the TOTP failure-logging block above (lines 238–253). Consider extracting a small helper alongside `generate_html_*()` near the top of the file — something like `log_failed_mfa_attempt(int $userid, string $reason): void` — so that the audit shape (and the username-resolution fix) lives in one place and the two paths can't drift.

src/RestControllers/AuthorizationController.php
```diff
@@ -878,6 +879,19 @@ public function userLogin(HttpRestRequest $request): ResponseInterface
             //Check the validity of the authentication token
             if ($request->request->get('user_role') === 'api'  && $mfa->isMfaRequired() && !is_null($mfaToken)) {
                 if (!$mfaToken || !$mfa->check($mfaToken, $request->request->get('mfa_type'))) {
+                    // Log failed MFA authentication attempt
+                    $ip = collectIpAddresses();
+                    $username = $request->request->get('username', '');
+                    $mfaType = $request->request->get('mfa_type', 'unknown');
```
kojiromike: `mfa_type` is request input being interpolated directly into the log message on line 893. A request with `mfa_type=foo%0AINJECTED+ENTRY` would split the audit log line — log injection (CWE-117). Constrain to the known MFA constants before using:

```suggestion
                    $rawMfaType = (string) $request->request->get('mfa_type', '');
                    $mfaType = in_array($rawMfaType, [MfaUtils::TOTP, MfaUtils::U2F], true) ? $rawMfaType : 'unknown';
```

(or whatever closed set `MfaUtils` exposes — an enum would be even better here).

src/RestControllers/AuthorizationController.php
```diff
@@ -878,6 +879,19 @@ public function userLogin(HttpRestRequest $request): ResponseInterface
...
+                    $userService = new UserService();
+                    $authGroup = !empty($username) ? $userService->getAuthGroupForUser($username) : '';
```
kojiromike: Two issues:

1. `empty()` shouldn't appear in new code per the repo coding standards — use an explicit check (`$username === ''`).
2. `UserService::getAuthGroupForUser()` returns `string|bool`; when the user isn't in a group it returns `false`, which then gets passed into `EventAuditLogger::newEvent()` (string parameter). Coerce to `''` here, e.g.:

```suggestion
                    $authGroup = '';
                    if ($username !== '') {
                        $resolvedGroup = $userService->getAuthGroupForUser($username);
                        if (is_string($resolvedGroup)) {
                            $authGroup = $resolvedGroup;
                        }
                    }
```

Also worth considering: `$this->userId` is already populated by the just-completed `verifyLogin()`, so it's a more authoritative source for the user identity than re-reading `username` from the request.

kojiromike (commented):
Audit logging for failed MFA attempts is a worthwhile improvement and the chosen log shape (`'login'`, username, group, 0, "failure: …") matches what `AuthUtils` already does for password failures. A few things to address before merge:

1. **Audit fidelity in the MFA-screen path.** In both blocks in `main_screen.php`, the username being logged comes from `$_POST['authUser']`, which the client can tamper with on the MFA POST. The trustable identifier of the user being challenged is `$session->get('authUserID')` (already in `$userid` at line 178). If you log `$_POST['authUser']`, an attacker can submit an MFA failure that gets recorded against a different account, defeating the point of the audit entry. Resolve the username from `$userid` (or just log with the ID and leave username blank).

2. **Log injection in the OAuth2 path.** `mfa_type` is request input that is then interpolated directly into the audit-log message string. A value like `mfa_type=foo%0AINJECTED+ENTRY` would split the log line. Constrain it to the known MFA constants (`MfaUtils::TOTP`, `MfaUtils::U2F`, etc.) or sanitize for newlines before using.

3. **`empty()` in new code.** The repo CLAUDE.md "Coding Standards" section explicitly calls out `empty()` as a legacy antipattern that should not appear in new code. Use explicit checks instead (`$userInfo === false`, `$username === ''`, `is_array($userInfo) && isset($userInfo['id'])`, etc.).

4. **`getAuthGroupForUser()` returns `string|bool`.** When the user isn't in a group it returns `false`, which then gets passed straight into `EventAuditLogger::newEvent()` as the `$groupname` argument (declared `string`). At PHPStan level 10 this is a type error; at runtime it will end up as `""` via implicit coercion. Coerce to `''` explicitly (or skip the lookup and just log `''`).

5. **Duplication.** The TOTP and U2F failure-logging blocks in `main_screen.php` are nearly identical (and the OAuth2 block is the same shape). Extracting a small helper near the other MFA-support functions in this file would make it harder for the two paths to drift, and would be a natural place to also encode the resolve-username-from-session fix.

Inline comments below.

interface/main/main_screen.php
aanand-1706: Username is no longer taken from `$_POST['authUser']`. It is now resolved server-side from `$userid` (i.e. `$session->get('authUserID')`) via a DB lookup in the extracted helper. In the OAuth2path, `$this->userId` is used with `UserService::getUser()`.

interface/main/main_screen.php
aanand-1706: All uses of `empty()` replaced with explicit checks - `$username !== ''`, `is_array()`, `isset()`.

interface/main/main_screen.php
aanand-1706: In the extracted helper, coerced with `?: ''`. In the OAuth2 path, an explicit `is_string()` guard is used:                                                                                 
```
  $resolvedGroup = $userService->getAuthGroupForUser($username);                                                                                                                            
  if (is_string($resolvedGroup)) {                              
      $authGroup = $resolvedGroup;                                                                                                                                                          
  }
```

interface/main/main_screen.php
aanand-1706: Extracted a `log_failed_mfa_attempt(int $userid, string $reason): void` helper alongside the other MFA support functions at the top of main_screen.php. Both TOTP and U2F failure paths now call it:                                                                                                                                                                                  
```
  log_failed_mfa_attempt($userid, 'TOTP code incorrect');
  log_failed_mfa_attempt($userid, 'U2F authentication error: ' . $e->getMessage());  
```

src/RestControllers/AuthorizationController.php
aanand-1706: `mfa_type` is now validated against `MfaUtils::TOTP` and `MfaUtils::U2F` using a strict `in_array(..., true)` check before being interpolated into the log message. Any other value falls back to `'unknown'`.

src/RestControllers/AuthorizationController.php
aanand-1706: In the OAuth2 path, `$this->userId` is used instead of the request username - it is already populated by the preceding `verifyLogin()` call, so the audit entry is tied to the verified identity rather than client-supplied input.

aanand-1706:
> Audit logging for failed MFA attempts is a worthwhile improvement and the chosen log shape (`'login'`, username, group, 0, "failure: …") matches what `AuthUtils` already does for password failures. A few things to address before merge:
> 
> 1. **Audit fidelity in the MFA-screen path.** In both blocks in `main_screen.php`, the username being logged comes from `$_POST['authUser']`, which the client can tamper with on the MFA POST. The trustable identifier of the user being challenged is `$session->get('authUserID')` (already in `$userid` at line 178). If you log `$_POST['authUser']`, an attacker can submit an MFA failure that gets recorded against a different account, defeating the point of the audit entry. Resolve the username from `$userid` (or just log with the ID and leave username blank).
> 2. **Log injection in the OAuth2 path.** `mfa_type` is request input that is then interpolated directly into the audit-log message string. A value like `mfa_type=foo%0AINJECTED+ENTRY` would split the log line. Constrain it to the known MFA constants (`MfaUtils::TOTP`, `MfaUtils::U2F`, etc.) or sanitize for newlines before using.
> 3. **`empty()` in new code.** The repo CLAUDE.md "Coding Standards" section explicitly calls out `empty()` as a legacy antipattern that should not appear in new code. Use explicit checks instead (`$userInfo === false`, `$username === ''`, `is_array($userInfo) && isset($userInfo['id'])`, etc.).
> 4. **`getAuthGroupForUser()` returns `string|bool`.** When the user isn't in a group it returns `false`, which then gets passed straight into `EventAuditLogger::newEvent()` as the `$groupname` argument (declared `string`). At PHPStan level 10 this is a type error; at runtime it will end up as `""` via implicit coercion. Coerce to `''` explicitly (or skip the lookup and just log `''`).
> 5. **Duplication.** The TOTP and U2F failure-logging blocks in `main_screen.php` are nearly identical (and the OAuth2 block is the same shape). Extracting a small helper near the other MFA-support functions in this file would make it harder for the two paths to drift, and would be a natural place to also encode the resolve-username-from-session fix.
> 
> Inline comments below.

Thank you for the thorough and prompt review - the feedback was very clear and actionable, and helped improve both the correctness and security of the implementation.

kojiromike:
@aisle-analyzer

interface/main/main_screen.php
```diff
@@ -137,6 +138,25 @@ function generate_html_end()
...
+        $userService = new \OpenEMR\Services\UserService();
+        $authGroup = $userService->getAuthGroupForUser($username) ?: '';
+    }
+    EventAuditLogger::getInstance()->newEvent(
+        'login',
+        $username,
+        $authGroup,
+        0,
+        "failure: " . $ip['ip_string'] . ". " . $reason
+    );
```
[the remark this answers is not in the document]
kojiromike: Property of `collectIpAddresses()` (`library/sanitize.inc.php:29`), not introduced by this PR. That helper is the established pattern used by `LogTablesSink`, `AuthUtils`, and other audit call sites. Hardening it requires a trusted-proxy allowlist, restructured audit row fields, and migrating every caller — out of scope here. Tracked in #11915.

**No change needed in this PR.**

src/RestControllers/AuthorizationController.php
```diff
@@ -878,6 +879,27 @@ public function userLogin(HttpRestRequest $request): ResponseInterface
...
+                            $authGroup = $resolvedGroup;
+                        }
+                    }
+                    EventAuditLogger::getInstance()->newEvent(
+                        'login',
+                        $username,
+                        $authGroup,
+                        0,
+                        "failure: " . $ip['ip_string'] . ". OAuth2 MFA ($mfaType) code incorrect"
+                    );
```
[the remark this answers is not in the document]
kojiromike: Same root cause — shared `collectIpAddresses()` helper. Tracked in #11915.

**No change needed in this PR.**

interface/main/main_screen.php
```diff
@@ -266,6 +287,7 @@ function generate_html_end()
                 } catch (\u2flib_server\Error $e) {
                     // Authentication failed so we will build the U2F form again.
                     $form_response = '';
+                    log_failed_mfa_attempt($userid, 'U2F authentication error: ' . $e->getMessage());
                     $errormsg = xl('U2F Key Authentication error') . ": " . $e->getMessage();
```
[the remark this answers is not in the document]
kojiromike: Two parts:

1. **Disclosure via `$e->getMessage()`** — declining. Source is `\u2flib_server\Error` (local U2F library); messages are a fixed set ("Counter too low", "Challenge does not match", etc.) describing why the hardware-key check failed. Persisting that to the admin-only audit row is the point of the entry — collapsing to `U2F_AUTH_FAILED` would lose the only forensically useful payload.

2. **CR/LF injection (CWE-117)** — declining. Source is a hardcoded library, not request input, and `EventAuditLogger::newEvent` writes to a SQL row, not a flat file, so newlines don't split records. If we want defense-in-depth, the right place is `EventAuditLogger::newEvent` itself (one fix for all callers), not this call site.

**No change needed in this PR.**

kojiromike:
The phpstan findings seem legitimate and IMO should be addressed.

aanand-1706:
@aisle-analyzer

aanand-1706:
> The phpstan findings seem legitimate and IMO should be addressed.

@kojiromike  I have made the changes needed for phpstan. Tried running @aisle-analyzer but I'm guessing I don't have access.

kojiromike:
Thanks for catching the audit-log gap — the security motivation is solid, but I'd push back on the shape before merging.

Two concerns with the current approach:

1. **New global function + baseline entry.** `log_failed_mfa_attempt` lands in the global namespace and the only way to make PHPStan accept it is to add an ignore to `.phpstan/baseline/openemr.noGlobalNsFunctions.php`. The baseline is for legacy code we're trying to drain, not somewhere new code should land.
2. **The helper isn't actually shared.** `AuthorizationController::userLogin()` reimplements the same IP/user/group/EAL block inline rather than calling `log_failed_mfa_attempt`, so the abstraction adds a global without removing the duplication it's meant to solve.

Zooming out: a failed MFA attempt is just one flavor of "auth failure event," and that pattern already exists ~25 times in `src/Common/Auth/AuthUtils.php` — every call collects the IP, sets `success = 0`, and formats `"<phase>: <ip>. <reason>"`. Rather than adding an MFA-specific helper, I'd suggest a generic method on `EventAuditLogger`:

```php
public function logAuthFailure(
    AuthEvent $event,           // value object; see below
    ?string $username,          // null when entirely unknown
    string $authGroup,          // '' when not yet resolved
    string $reason,             // non-empty-string; e.g. "TOTP code incorrect"
    ?int $patientId = null,     // portal auth only
): void
```

`AuthEvent` is a small value object rather than an enum, so modules (OIDC, passkeys, SAML, etc.) can define their own event types without patching core:

```php
final readonly class AuthEvent
{
    public function __construct(public string $value)
    {
        if ($value === '') {
            throw new \DomainException('AuthEvent value cannot be empty');
        }
    }

    public static function login(): self    { return new self('login'); }
    public static function mfa(): self      { return new self('mfa'); }
    public static function password(): self { return new self('password'); }
    public static function logout(): self   { return new self('logout'); }
    public static function auth(): self     { return new self('auth'); }
}
```

Notes on the shape:

- Named factories cover core's known event types; modules construct an `AuthEvent` directly with their own string. The wrapper still gives the helper a typed parameter instead of a raw string.
- `mfa` is its own event value, not buried inside `comments` with `event = 'login'`. That makes `WHERE event = 'mfa' AND success = 0` a one-column query — which is what ASVS 16.3.1 reviewers will actually run. Worth calling out in the PR description: `'mfa'` is a new value in `log.event`, so dashboards/saved queries filtering on event values need to know it exists.
- `username` is nullable because some failure paths fire before the username is matched to a row.
- `reason` is `non-empty-string` (PHPStan-enforced) — a failure row with no reason isn't useful for forensics.
- The helper calls `collectIpAddresses()` itself and composes `"<reason>: <ip>"` for the `comments` column, so callers don't redo the formatting and the IP convention lives in one place.
- `patientId` stays optional because patient-portal failure paths in `AuthUtils.php` need to attribute to a patient; staff/MFA paths don't.

That keeps EAL generic (no awareness of TOTP/U2F/OAuth2), removes the IP boilerplate from `AuthUtils.php`, and lets both your new MFA call sites collapse to one line each — no global function, no baseline entry, no duplicated block in `AuthorizationController`.

aanand-1706:
> Thanks for catching the audit-log gap — the security motivation is solid, but I'd push back on the shape before merging.
> 
> Two concerns with the current approach:
> 
> 1. **New global function + baseline entry.** `log_failed_mfa_attempt` lands in the global namespace and the only way to make PHPStan accept it is to add an ignore to `.phpstan/baseline/openemr.noGlobalNsFunctions.php`. The baseline is for legacy code we're trying to drain, not somewhere new code should land.
> 2. **The helper isn't actually shared.** `AuthorizationController::userLogin()` reimplements the same IP/user/group/EAL block inline rather than calling `log_failed_mfa_attempt`, so the abstraction adds a global without removing the duplication it's meant to solve.
> 
> Zooming out: a failed MFA attempt is just one flavor of "auth failure event," and that pattern already exists ~25 times in `src/Common/Auth/AuthUtils.php` — every call collects the IP, sets `success = 0`, and formats `"<phase>: <ip>. <reason>"`. Rather than adding an MFA-specific helper, I'd suggest a generic method on `EventAuditLogger`:
> 
> ```
> public function logAuthFailure(
>     AuthEvent $event,           // value object; see below
>     ?string $username,          // null when entirely unknown
>     string $authGroup,          // '' when not yet resolved
>     string $reason,             // non-empty-string; e.g. "TOTP code incorrect"
>     ?int $patientId = null,     // portal auth only
> ): void
> ```
> 
> `AuthEvent` is a small value object rather than an enum, so modules (OIDC, passkeys, SAML, etc.) can define their own event types without patching core:
> 
> ```
> final readonly class AuthEvent
> {
>     public function __construct(public string $value)
>     {
>         if ($value === '') {
>             throw new \DomainException('AuthEvent value cannot be empty');
>         }
>     }
> 
>     public static function login(): self    { return new self('login'); }
>     public static function mfa(): self      { return new self('mfa'); }
>     public static function password(): self { return new self('password'); }
>     public static function logout(): self   { return new self('logout'); }
>     public static function auth(): self     { return new self('auth'); }
> }
> ```
> 
> Notes on the shape:
> 
> * Named factories cover core's known event types; modules construct an `AuthEvent` directly with their own string. The wrapper still gives the helper a typed parameter instead of a raw string.
> * `mfa` is its own event value, not buried inside `comments` with `event = 'login'`. That makes `WHERE event = 'mfa' AND success = 0` a one-column query — which is what ASVS 16.3.1 reviewers will actually run. Worth calling out in the PR description: `'mfa'` is a new value in `log.event`, so dashboards/saved queries filtering on event values need to know it exists.
> * `username` is nullable because some failure paths fire before the username is matched to a row.
> * `reason` is `non-empty-string` (PHPStan-enforced) — a failure row with no reason isn't useful for forensics.
> * The helper calls `collectIpAddresses()` itself and composes `"<reason>: <ip>"` for the `comments` column, so callers don't redo the formatting and the IP convention lives in one place.
> * `patientId` stays optional because patient-portal failure paths in `AuthUtils.php` need to attribute to a patient; staff/MFA paths don't.
> 
> That keeps EAL generic (no awareness of TOTP/U2F/OAuth2), removes the IP boilerplate from `AuthUtils.php`, and lets both your new MFA call sites collapse to one line each — no global function, no baseline entry, no duplicated block in `AuthorizationController`.

@kojiromike Made the above changes. 

In `AuthorizationController`, `authUser` and `authProvider` are only written to the session in the non-API login branch of `confirmPassword()` - the OAuth2 flow uses `$this->apiAuth = true` which skips `setUserSessionVariables()`, so those session keys are never populated and `$this->userId` (set from the DB after successful password verification) is the only trustworthy source available. So there's a slight difference in implementation between the two MFA call sites.

src/Common/Auth/AuthEvent.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function __construct(public string $value)
+    {
+        if ($value === '') {
+            throw new \DomainException('AuthEvent value cannot be empty');
+        }
+    }
+
+    public static function login(): self    { return new self('login'); }
+    public static function mfa(): self      { return new self('mfa'); }
```
kojiromike: The single-line method bodies + space-padded alignment are unusual in this codebase — I couldn't find another instance of this style in `src/`. PER-CS doesn't strictly forbid it, but for consistency with everything else here I'd expand to standard multi-line form:

```php
public static function login(): self
{
    return new self('login');
}
```

Same applies to `mfa()`, `password()`, `logout()`, and `auth()` below.

src/RestControllers/AuthorizationController.php
```diff
@@ -878,6 +880,20 @@ public function userLogin(HttpRestRequest $request): ResponseInterface
             //Check the validity of the authentication token
             if ($request->request->get('user_role') === 'api'  && $mfa->isMfaRequired() && !is_null($mfaToken)) {
                 if (!$mfaToken || !$mfa->check($mfaToken, $request->request->get('mfa_type'))) {
+                    // Log failed MFA authentication attempt
+                    $rawMfaType = (string) $request->request->get('mfa_type', '');
```
kojiromike: The `(string)` cast on a `mixed` value violates the repo's "narrow, don't cast" rule (CLAUDE.md → Type System / Static Analysis) and silently coerces non-string input — `(string) ['x']` becomes `"Array"` with a warning rather than a type error.

Symfony 7's `InputBag` provides a typed `getString()` that throws `BadRequestException` on non-scalar input — the right behavior at a request boundary, and it mirrors the `OEGlobalsBag::getString()` pattern OpenEMR already uses for typed config access:

```suggestion
                    $rawMfaType = $request->request->getString('mfa_type');
```

kojiromike (commented):
The new design (`EventAuditLogger::logAuthFailure()` + `AuthEvent` value object) cleanly addresses everything from the prior round — session-based username instead of `$_POST`, allowlisted `mfa_type` against log injection, no `empty()`, `is_string()` guard for `getAuthGroupForUser()`'s `string|bool` return, and the duplicated TOTP/U2F blocks collapsed into one helper on EAL. Nice work converging on this shape.

A couple of residual items worth addressing before merge:

- **Tests.** No tests appear in the diff. `AuthEvent` is trivially testable (one constructor with one validation rule, five named factories) and would be worth a small isolated test since it's a brand-new value object with no other coverage paths. `EventAuditLogger::logAuthFailure()` is also exercisable in isolation by asserting against the `newEvent()` row it writes.
- **`AuthUtils.php` migration.** The motivating idea behind moving the helper onto `EventAuditLogger` was that `AuthUtils.php`'s ~25 inline IP-formatting blocks could collapse to one-line `logAuthFailure()` calls. Out of scope for this PR, but a worthwhile follow-up so the new helper actually drains the legacy pattern instead of running alongside it.
- **`mfa` event value.** Good call-out in the PR description. Worth a release-notes line too so operators reading release notes (not PR descriptions) catch it.

Inline comments below for two issues.

src/Common/Auth/AuthEvent.php
aanand-1706: Changed to multi-line form.

src/RestControllers/AuthorizationController.php
aanand-1706: Using `getString()`

aanand-1706:
> The new design (`EventAuditLogger::logAuthFailure()` + `AuthEvent` value object) cleanly addresses everything from the prior round — session-based username instead of `$_POST`, allowlisted `mfa_type` against log injection, no `empty()`, `is_string()` guard for `getAuthGroupForUser()`'s `string|bool` return, and the duplicated TOTP/U2F blocks collapsed into one helper on EAL. Nice work converging on this shape.
> 
> A couple of residual items worth addressing before merge:
> 
> * **Tests.** No tests appear in the diff. `AuthEvent` is trivially testable (one constructor with one validation rule, five named factories) and would be worth a small isolated test since it's a brand-new value object with no other coverage paths. `EventAuditLogger::logAuthFailure()` is also exercisable in isolation by asserting against the `newEvent()` row it writes.
> * **`AuthUtils.php` migration.** The motivating idea behind moving the helper onto `EventAuditLogger` was that `AuthUtils.php`'s ~25 inline IP-formatting blocks could collapse to one-line `logAuthFailure()` calls. Out of scope for this PR, but a worthwhile follow-up so the new helper actually drains the legacy pattern instead of running alongside it.
> * **`mfa` event value.** Good call-out in the PR description. Worth a release-notes line too so operators reading release notes (not PR descriptions) catch it.
> 
> Inline comments below for two issues.

Will refactor the `AuthUtils.php` file in a separate PR.
Made all the other changes.

kojiromike (approved):
Approving pending passing checks.

PR #11938, opened 2026-04-30
fix(test): use INSERT ON DUPLICATE KEY UPDATE for G10 test setup

Firehed:

Fixes G10 certification test failures when the FHIR US Core profile version global doesn't exist in the database.

- Change `setupG10ProfileGlobals()` from UPDATE to INSERT ON DUPLICATE KEY UPDATE so the row is created if it doesn't exist
- Fix argument order in `teardownG10Test()` (value before key name, matching the SQL parameter order)

The `Assisted-by` trailer is included in the commit.

### DIFF
--- a/tests/Tests/Certification/HIT1/G10_Certification/Trait/G10ApiTestTrait.php
+++ b/tests/Tests/Certification/HIT1/G10_Certification/Trait/G10ApiTestTrait.php
@@ -77,19 +77,30 @@ trait G10ApiTestTrait
         self::setupG10ProfileGlobals();
     }
 
-    protected static function setupG10ProfileGlobals() {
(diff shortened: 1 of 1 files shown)


Firehed:
## Investigation findings

After comparing CI logs between master and this branch:

1. **CapabilityStatement tests fail identically on both** - the `us-core-careteam` profile error is unchanged
2. **Inferno Patient tests pass on both** - `4.2.Patient Tests: pass` appears in both logs
3. **This branch adds a new failure** - `testPatient` fails only on branch with sex extension validation errors (SNOMED codes 248153007/248152002 have swapped display names)

The assertion count increased (879 → 903) suggesting more validation is running, but I don't have clear evidence from the logs that the `setupG10ProfileGlobals` change fixes any existing failure. The original hypothesis was that UPDATE silently affected zero rows, but the CapabilityStatement failures are identical before and after.

Recommend closing until we can identify what specific failure this change addresses.

PR #11939, opened 2026-04-30
fix: guard undefined keys and legacy PHP warnings flagged in production logs

### LINKED ISSUE #11935: fix: small undefined-key and legacy-PHP warnings recurring in production logs
## Problem

A handful of small `Undefined array key` and reference-return warnings show up consistently in production logs and have not been guarded. None is fatal on its own, but together they constitute log noise that hides real problems and a few are pre-PHP-7 idioms that PHPStan flags.

### 1. `sql_upgrade.php:380` — undefined `form_old_version` on first GET

```php
$form_old_version = $cliFromVersion ?? $_POST['form_old_version'];
```

When the upgrade page is loaded via GET (or via POST with `form_submit` set but no `form_old_version`), the `?? $_POST['form_old_version']` branch warns.

Production log:

```
PHP Warning:  Undefined array key "form_old_version" in sql_upgrade_temp.php on line 380
```

Trivial fix: `$_POST['form_old_version'] ?? ''` (or whatever the right default is).

### 2. `interface/forms/newpatient/C_EncounterVisitForm.class.php:679` — undefined `onset_date`

```php
if (Utilities::isDateEmpty($encounter['onset_date'])) {
    $encounter['onset_date'] = null;
}
```

Production log:

```
PHP Warning:  Undefined array key "onset_date" in
   interface/forms/newpatient/C_EncounterVisitForm.class.php on line 679
```

When the encounter row predates the `onset_date` column being populated, the array key is missing. The check should use `??` or `array_key_exists()`.

### 3. `interface/main/calendar/modules/PostCalendar/pnuserapi.php:1308` — `Only variable references should be returned by reference`

Production log:

```
PHP Notice:  Only variable references should be returned by reference in
   /var/www/.../interface/main/calendar/modules/PostCalendar/pnuserapi.php on line 1308
```

Legacy PostCalendar PHP4-era idiom: `return &$something_that_is_not_a_variable`. Either drop the `&` or assign to a local variable first.

### 4. Smarty `text` modifier emitted as deprecated

Production log:

```
PHP Deprecated: Using unregistered function "text" in a template is deprecated and
will be removed in a future release. Use Smarty::registerPlugin to explicitly
register a custom modifier.
   vendor/smarty/smarty/libs/sysplugins/smarty_internal_compile_private_modifier.php:113
```

Some Smarty template uses `{$x|text}` (the OpenEMR translation helper) without ever registering it as a Smarty plugin. Will hard-fail on the next Smarty major (the underlying behavior — falling back to the global function — is being removed). Already partially in scope of the Smarty epic #10111, but the `text` modifier is a pre-existing concrete instance worth addressing in isolation.

## Why one ticket

These are independent one-line fixes. Splitting them into four issues would be more PR ceremony than they're worth. Whoever picks this up can knock them off in a single PR or split if it's easier.

kojiromike:
## Summary

Fixes #11935.

Quiets four small recurring warnings/notices from production logs:

- `sql_upgrade.php`: default `$_POST['form_old_version']` to `''` so a GET load (no posted form) does not warn.
- `interface/forms/newpatient/C_EncounterVisitForm.class.php`: guard `$encounter['onset_date']` for legacy rows where the column was never populated.
- `interface/main/calendar/modules/PostCalendar/pnuserapi.php`: assign the event-dispatch result to a local variable before returning, so the by-reference return signature on `postcalendar_userapi_pcGetEvents()` stops triggering "Only variable references should be returned by reference."
- `library/classes/Controller.class.php`: register the `text` modifier as a Smarty plugin so `{$x|text}` no longer falls back to the deprecated unregistered-function path. The implementation narrows to scalar/Stringable values and delegates to the global `text()` helper.

## Test plan

- [ ] Open `sql_upgrade.php` via GET; confirm no `Undefined array key "form_old_version"` warning.
- [ ] View an encounter whose row was created before `onset_date` was tracked; confirm no `Undefined array key "onset_date"` warning.
- [ ] Load the calendar; confirm no `Only variable references should be returned by reference` notice from `pnuserapi.php`.
- [ ] Render a Controller-driven Smarty page that uses `|text` (e.g. document list, insurance company list); confirm output is unchanged and no `Using unregistered function "text"` deprecation appears.

### DIFF
--- a/interface/forms/newpatient/C_EncounterVisitForm.class.php
+++ b/interface/forms/newpatient/C_EncounterVisitForm.class.php
@@ -676,7 +676,7 @@ class C_EncounterVisitForm
         $posOptions = $this->getPosOptionsForTemplate($facilityPosCode);
 // END AI GENERATED CODE
 
-        if (Utilities::isDateEmpty($encounter['onset_date'])) {
+        if (Utilities::isDateEmpty($encounter['onset_date'] ?? null)) {
             $encounter['onset_date'] = null;
         }
 
--- a/interface/main/calendar/modules/PostCalendar/pnuserapi.php
+++ b/interface/main/calendar/modules/PostCalendar/pnuserapi.php
@@ -1305,7 +1305,8 @@ function &postcalendar_userapi_pcGetEvents($args)
     $event->setProviderID($providerID ?? $provider_id ?? null);
 
     $result = OEGlobalsBag::getInstance()->getKernel()->getEventDispatcher()->dispatch($event, CalendarUserGetEventsFilter::EVENT_NAME);
-    return $result->getEventsByDays();
+    $eventsByDays = $result->getEventsByDays();
+    return $eventsByDays;
 }
 
 //===========================
--- a/library/classes/Controller.class.php
+++ b/library/classes/Controller.class.php
@@ -280,17 +280,8 @@ class Controller extends Smarty implements ControllerInterface
 
     private static function escapeForTextNode(mixed $value): string
     {
-        if ($value === null) {
-            return '';
-        }
-        if (is_string($value)) {
-            return htmlspecialchars($value, ENT_NOQUOTES);
-        }
-        if (is_int($value) || is_float($value) || is_bool($value)) {
-            return htmlspecialchars((string) $value, ENT_NOQUOTES);
-        }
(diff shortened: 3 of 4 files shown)


kojiromike:
@aisle-analyzer

PR #11940, opened 2026-04-30
fix(session): clean up callers that re-open read_and_close session

### LINKED ISSUE #11934: fix(session): clean up callers that re-open read-and-close session for writing
## Problem

Production logs show recurring warnings of the form:

```
OpenEMR.WARNING: Session reopened for writing after read_and_close
{"session_name":"OpenEMR","script":"/openemr/library/ajax/dated_reminders_counter.php"}
{"session_name":"OpenEMR","script":"/openemr/interface/main/calendar/index.php"}
{"session_name":"OpenEMR","script":"/openemr/interface/new/new_comprehensive_save.php"}
{"session_name":"OpenEMR","script":"/openemr/interface/patient_file/summary/demographics.php"}
{"session_name":"OpenEMR","script":"/openemr/interface/forms/newpatient/save.php"}
```

These are emitted by `SessionWrapperFactory` (`src/Common/Session/SessionWrapperFactory.php:119`) when a script that opened the session with `read_and_close=true` later writes to `$_SESSION`. The wrapper detects the write and re-opens the session, but logs a warning because the read-and-close optimization has been defeated.

The most chronic offender is `library/ajax/dated_reminders_counter.php`, which fires once per minute per logged-in user.

## Why it matters

- The warning is emitted on a hot path. On a busy server it produces tens of thousands of log lines per day, drowning real warnings.
- Each occurrence means a session lock is acquired and released a second time within the request, which serializes concurrent AJAX requests for the same user — the very thing read-and-close was added to avoid.

## Suggested direction

For each script in the offender list, decide which is correct:

- The script does in fact need to write to the session → open with `read_and_close=false` from the start.
- The script does not need to write to the session → find and remove the assignment that triggers the re-open (often a cached value being eagerly stored).

Implementer's choice per script.

## Notes

The warning instrumentation itself works as designed — this is about cleaning up the callers, not changing the wrapper.

kojiromike:
## Summary

Resolves the recurring `Session reopened for writing after read_and_close` warnings emitted on a hot path in production. Each occurrence acquires and releases the session lock twice in the same request, defeating the read-and-close optimization and serializing concurrent AJAX requests for the same user. Tens of thousands of lines per day on a busy server.

Per-script disposition:

- **`library/ajax/dated_reminders_counter.php`** — removed the `keepAliveTime` session write entirely. The value is never read anywhere in the codebase, so the write was pure dead code (and the most chronic offender, firing once per minute per logged-in user). Verified with `git grep keepAliveTime`.
- **`interface/main/calendar/index.php`** — set `$sessionAllowWrite = true` (writes `pc_facility`, `pc_username`, `viewtype`).
- **`interface/new/new_comprehensive_save.php`** — set `$sessionAllowWrite = true` (writes new patient `pid` via `setpid()`).
- **`interface/patient_file/summary/demographics.php`** — set `$sessionAllowWrite = true` (writes `pid`, `encounter`, `alert_notify_pid`).
- **`interface/forms/newpatient/save.php`** — set `$sessionAllowWrite = true` (writes `encounter` via `setencounter()`).

The four `$sessionAllowWrite = true` additions follow the existing pattern (e.g. `interface/main/main_screen.php:20`, `interface/login/login.php:51`, `interface/usergroup/mfa_totp.php:18`). PHPStan's `ForbidDirectSessionWriteRule` already whitelists this pattern.

Closes #11934

## Test plan

- [ ] Sign in to OpenEMR; confirm the dated-reminders counter still polls and updates the message badge (no regression from removing the dead `keepAliveTime` write).
- [ ] Open the calendar; switch facility / provider / view; confirm selections persist across navigation (writes go to session as before).
- [ ] Add a new patient via the New/Search → Comprehensive form; confirm the new pid becomes the active patient context.
- [ ] Open a patient's Demographics; confirm pid + encounter context is established and CDR alerts behave as before.
- [ ] Create a new encounter via the encounter form; confirm it becomes the active encounter and the encounter list updates.
- [ ] Tail `error_log`; confirm `Session reopened for writing after read_and_close` no longer appears for these five scripts.

### DIFF
--- a/interface/main/calendar/index.php
+++ b/interface/main/calendar/index.php
@@ -16,6 +16,8 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+// Set $sessionAllowWrite to true since pc_facility, pc_username, and viewtype are written to the session below.
+$sessionAllowWrite = true;
 require_once("../../globals.php");
 
 use OpenEMR\Common\Acl\AclMain;
--- a/interface/forms/newpatient/save.php
+++ b/interface/forms/newpatient/save.php
@@ -12,6 +12,8 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+// Set $sessionAllowWrite to true since the encounter is written to the session via setencounter() below.
+$sessionAllowWrite = true;
 require_once(__DIR__ . "/../../globals.php");
 require_once("$srcdir/forms.inc.php");
 require_once("$srcdir/encounter.inc.php");
--- a/interface/new/new_comprehensive_save.php
+++ b/interface/new/new_comprehensive_save.php
@@ -12,6 +12,8 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+// Set $sessionAllowWrite to true since the new patient pid is written to the session via setpid() below.
+$sessionAllowWrite = true;
 require_once("../globals.php");
 
 use OpenEMR\BC\ServiceContainer;
--- a/interface/patient_file/summary/demographics.php
+++ b/interface/patient_file/summary/demographics.php
@@ -27,6 +27,9 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+// Set $sessionAllowWrite to true since this script writes disablePreviousNameAdds to the session
+// on every load, alert_notify_pid whenever CDR is enabled, and pid + encounter when ?set_pid is provided.
+$sessionAllowWrite = true;
 require_once("../../globals.php");
 
 $srcdir = \OpenEMR\Core\OEGlobalsBag::getInstance()->getSrcDir();
--- a/library/ajax/dated_reminders_counter.php
+++ b/library/ajax/dated_reminders_counter.php
@@ -31,8 +31,6 @@ if (SessionTracker::isSessionExpired()) {
     echo json_encode(['timeoutMessage' => 'timeout']);
     exit;
 }
-// keep this below above time out check.
-OpenEMR\Common\Session\SessionUtil::setSession('keepAliveTime', time());
 
 $total_counts = [];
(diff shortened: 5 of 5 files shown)


kojiromike:
@aisle-analyzer

kojiromike:
Thanks aisle-research-bot. All three findings describe pre-existing behavior, not new attack surface introduced here:

- The `setpid()` / `SessionUtil::setSession('encounter', ...)` writes in `demographics.php` and the `pc_username` / `pc_facility` / `viewtype` writes in `calendar/index.php` were happening before this PR. They were just causing the read-and-close wrapper to log `Session reopened for writing after read_and_close` because the session had already been closed for writes.
- `$sessionAllowWrite = true` does not enable a write that did not exist; it only prevents the wrapper from closing the session before the existing write runs (silencing the warning, removing one extra lock acquire/release per request).

So the CSRF and authorization concerns are real but pre-existing, and out of scope for a warning-cleanup PR. Filed as follow-ups so they do not get lost:

- Finding #1 → #11944 (CSRF on `set_pid` / `set_encounterid` patient-context switching)
- Findings #2 and #3 → #11945 (CSRF + authorization on calendar `pc_username`, `pc_facility`, `viewtype`)

interface/main/calendar/index.php
```diff
@@ -16,6 +16,8 @@
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
 
+// Set $sessionAllowWrite to true since pc_facility, pc_username, and viewtype are written to the session below.
+$sessionAllowWrite = true;
```
bradymiller, line 20: hi @kojiromike ,

Wouldn't it still beneficial on long running scripts to not set above.
For example:
session gets read then unlocked (so other concurrent scripts can grab it)
then it grabs the lock again when it sets a variable
and then releases it right away (so other concurrent scripts can grab it)

I think that was the intent in past and why those session write functions do this.

kojiromike: Yeah, you're right. I trusted the bot a bit too much on this one.

PR #11943, opened 2026-04-30
fix(fhir): Resolve incorrect display name for various FHIR APIs

Firehed:
```
18795:2026-04-30T20:36:42.2770380Z Message(type=error): MedicationRequest/96509824-0159-47d1-a735-8cc74aac0807: MedicationRequest.category[0].coding[0].display: Wrong Display Name 'Home/Community' for http://terminology.hl7.org/CodeSystem/medicationrequest-category#community. Valid display is 'Community' (en) (for the language(s) 'en')
```
As well as a number of similar errors hidden inside the Inferno logs.

Master:
```
.FF..FF.......FF..FF.F.....F...F....E.                         41 / 41 (100%)
ERRORS!
Tests: 41, Assertions: 851, Errors: 1, Failures: 11, PHPUnit Warnings: 1.
```

Branch:
```
.FF..FF...........F............F....E.                         41 / 41 (100%)
ERRORS!
Tests: 41, Assertions: 873, Errors: 1, Failures: 6, PHPUnit Warnings: 1.
```
Updated the constant that's rendered to match that indicated by the Inferno test suite.

Yep, Claude did most of it.

### DIFF
--- a/src/Services/FHIR/FhirMedicationRequestService.php
+++ b/src/Services/FHIR/FhirMedicationRequestService.php
@@ -92,6 +92,35 @@ class FhirMedicationRequestService extends FhirServiceBase implements IResourceU
         'discharge' => 'Discharge',
     ];
 
+    /**
+     * Official display values for timing abbreviation codes.
+     * @see http://terminology.hl7.org/CodeSystem/v3-GTSAbbreviation
+     */
+    private const TIMING_DISPLAY = [
+        'BID' => 'BID',
+        'TID' => 'TID',
+        'QID' => 'QID',
+        'AM' => 'AM',
+        'PM' => 'PM',
+        'QD' => 'QD',
+        'QOD' => 'QOD',
+        'Q1H' => 'every hour',
+        'Q2H' => 'every 2 hours',
+        'Q3H' => 'every 3 hours',
+        'Q4H' => 'Q4H',
+        'Q6H' => 'Q6H',
+        'Q8H' => 'every 8 hours',
+        'WK' => 'weekly',
+    ];
+
+    /**
+     * Official display values for route codes (NCI Thesaurus).
+     * @see http://ncicb.nci.nih.gov/xml/owl/EVS/Thesaurus.owl
+     */
+    private const ROUTE_DISPLAY = [
+        'C38288' => 'ORAL',
+    ];
+
     const USCGI_PROFILE_URI = "http://hl7.org/fhir/us/core/StructureDefinition/us-core-medicationrequest";
     /**
      * @deprecated use USCGI_PROFILE_URI
@@ -278,10 +307,12 @@ class FhirMedicationRequestService extends FhirServiceBase implements IResourceU
         }
         // Dose and Rate
         if (!empty($dataRecord['interval_codes'])) {
+            $code = $dataRecord['interval_codes'];
+            $display = self::TIMING_DISPLAY[$code] ?? $dataRecord['interval_notes'];
             $intervalConcept = UtilsService::createCodeableConcept([
-                $dataRecord['interval_codes'] => [
-                    'code' => $dataRecord['interval_codes'],
-                    'description' => $dataRecord['interval_notes'],
+                $code => [
+                    'code' => $code,
+                    'description' => $display,
                     'system' => FhirCodeSystemConstants::HL7_TIMING_ABBREVIATION
                 ]
             ]);
@@ -312,6 +343,15 @@ class FhirMedicationRequestService extends FhirServiceBase implements IResourceU
             if (!empty($dataRecord['route_codes'])) {
                 $codeTypesService = $this->getCodeTypesService();
                 $parsedCodes = $codeTypesService->parseCodesIntoCodeableConcepts($dataRecord['route_codes']);
+                // Override display with official terminology values
+                if (is_array($parsedCodes)) {
+                    foreach ($parsedCodes as $code => &$codeData) {
+                        if (is_string($code) && is_array($codeData) && isset(self::ROUTE_DISPLAY[$code])) {
+                            $codeData['description'] = self::ROUTE_DISPLAY[$code];
+                        }
+                    }
+                    unset($codeData);
+                }
                 $route = UtilsService::createCodeableConcept($parsedCodes);
             } else {
                 $route = new FHIRCodeableConcept();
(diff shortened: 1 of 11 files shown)


Firehed:
## FHIR Display Value Fixes

Added display value mappings to FHIR services to ensure coded elements use official HL7 terminology display values instead of internal titles.

### Fixes Applied

| Error | Fixed In | Change |
|-------|----------|--------|
| MedicationRequest category "Home/Community" | FhirMedicationRequestService | CATEGORY_DISPLAY mapping |
| Observation category "social-history" | FhirObservationTrait | CATEGORY_DISPLAY mapping |
| MedicationRequest timing "Twice daily..." | FhirMedicationRequestService | TIMING_DISPLAY mapping |
| MedicationRequest route "Per Oris" | FhirMedicationRequestService | ROUTE_DISPLAY mapping |
| DocumentReference type "Discharge summary note" | FhirClinicalNotesService | LOINC_DISPLAY mapping |
| DocumentReference type "Consultation note" | FhirClinicalNotesService | LOINC_DISPLAY mapping |
| DiagnosticReport "Urinanalysis..." | FhirDiagnosticReportLaboratoryService | LOINC_DISPLAY mapping |
| Observation "Protein[Mass/Volume]..." | FhirObservationLaboratoryService | LOINC_DISPLAY mapping |
| Observation "Blood pressure systolic..." | FhirObservationVitalsService | Direct fix in COLUMN_MAPPINGS |

### Not Fixed (different issue type)
- `Procedure/...: Unknown code 'SNOMED-CT:36969009'` - missing code in SNOMED, not a display issue

### Notes
- PHPStan baseline will need regeneration since error counts shifted, but no new actual errors were introduced
- These are FHIR-layer overrides; the internal display values (e.g., "Home/Community") are preserved for UI purposes

src/Services/FHIR/FhirMedicationRequestService.php
```diff
@@ -81,6 +81,45 @@ class FhirMedicationRequestService extends FhirServiceBase implements IResourceU
...
+        'QD' => 'QD',
+        'QOD' => 'QOD',
+        'Q1H' => 'every hour',
+        'Q2H' => 'every 2 hours',
+        'Q3H' => 'every 3 hours',
+        'Q4H' => 'Q4H',
+        'Q6H' => 'Q6H',
+        'Q8H' => 'every 8 hours',
+        'WK' => 'weekly',
+    ];
```
kojiromike, line 114: Roughly half of these (`BID`, `TID`, `QID`, `AM`, `PM`, `QD`, `QOD`, `Q4H`, `Q6H`) map a code to a string identical to the code itself. If those are intentional overrides of whatever `$dataRecord['interval_notes']` would have been, please verify each against the published display in `http://terminology.hl7.org/CodeSystem/v3-GTSAbbreviation` — Inferno tends to check display strings against the code system verbatim, and the canonical displays for these are long-form English (e.g. "Two times a day" for BID). If those are not intentional overrides, the entries are dead and can be dropped so the `?? $dataRecord['interval_notes']` fallback handles them.

Also, since the entries below (`Q1H`, `Q2H`, `Q3H`, `Q8H`, `WK`) use lowercase long-form, mixing the two conventions in a single table is going to confuse the next person to maintain it.

src/Services/FHIR/FhirMedicationRequestService.php
```diff
@@ -81,6 +81,45 @@ class FhirMedicationRequestService extends FhirServiceBase implements IResourceU
...
+
+    /**
+     * Official display values for route codes (NCI Thesaurus).
+     * @see http://ncicb.nci.nih.gov/xml/owl/EVS/Thesaurus.owl
+     */
+    private const ROUTE_DISPLAY = [
+        'C38288' => 'ORAL',
+    ];
```
kojiromike, line 122: Single-entry `ROUTE_DISPLAY` here, but the route override block below applies it generically. If only `C38288` needs correction today the table is fine, but worth confirming against NCI Thesaurus that "ORAL" is the published display — the EVS preferred name for C38288 is typically "Oral Route of Administration". If Inferno is happy with "ORAL" then keep it, but the `@see` link doesn't currently document why "ORAL" was chosen over the longer form.

src/Services/FHIR/FhirMedicationRequestService.php
```diff
@@ -302,6 +343,15 @@ public function populateDosageInstruction(FHIRMedicationRequest $medRequestResou
...
                 $parsedCodes = $codeTypesService->parseCodesIntoCodeableConcepts($dataRecord['route_codes']);
+                // Override display with official terminology values
+                if (is_array($parsedCodes)) {
+                    foreach ($parsedCodes as $code => &$codeData) {
+                        if (is_string($code) && is_array($codeData) && isset(self::ROUTE_DISPLAY[$code])) {
+                            $codeData['description'] = self::ROUTE_DISPLAY[$code];
+                        }
+                    }
+                    unset($codeData);
+                }
```
kojiromike, line 354: The `is_array($parsedCodes)` guard implies uncertainty about `parseCodesIntoCodeableConcepts()`'s return type. If it's typed as `array`, the guard is dead and PHPStan would say so; if it's `mixed` and you're working around that, this is a good spot to file/track a follow-up to narrow it at the source rather than guard at every call site. Not blocking the PR.

src/Services/FHIR/FhirMedicationRequestService.php
```diff
@@ -469,24 +519,27 @@ public function populateMedication(FHIRMedicationRequest $medRequestResource, ar
...
+            $code = self::MEDICATION_REQUEST_CATEGORY_COMMUNITY;
             $medRequestResource->addCategory(UtilsService::createCodeableConcept(
                 [
-                    self::MEDICATION_REQUEST_CATEGORY_COMMUNITY => [
-                        'code' => self::MEDICATION_REQUEST_CATEGORY_COMMUNITY,
-                        'description' => xlt(self::MEDICATION_REQUEST_CATEGORY_COMMUNITY_TITLE),
-                        'system' => FhirCodeSystemConstants::HL7_MEDICATION_REQUEST_CATEGORY
+                    $code => [
+                        'code' => $code,
+                        'description' => self::CATEGORY_DISPLAY[$code],
```
kojiromike, line 541: Worth a small comment that this default branch trusts `self::CATEGORY_DISPLAY[self::MEDICATION_REQUEST_CATEGORY_COMMUNITY]` to exist — if someone later removes the `'community'` entry from the table the array access throws. Could be `?? 'Community'` defensively, or add a const-level assertion.

kojiromike (changes requested):
The Inferno failure count dropped from 11 to 6, the baseline shifts in a good direction (-38 total, -21 `array.invalidKey` from the new `is_string()` guards), and the `Blood pressure panel with all children optional` correction in the vitals service matches LOINC 85354-9 exactly. Direction is right.

Two things worth resolving before merge:

1. `TIMING_DISPLAY` looks half-finished. About a third of the entries map an abbreviation to itself (`'BID' => 'BID'`, `'TID' => 'TID'`, `'QID' => 'QID'`, `'AM' => 'AM'`, `'PM' => 'PM'`, `'QD' => 'QD'`, `'QOD' => 'QOD'`, `'Q4H' => 'Q4H'`, `'Q6H' => 'Q6H'`), while the others (`Q1H`, `Q2H`, `Q3H`, `Q8H`, `WK`) use long-form English. The HL7 GTSAbbreviation code system the `@see` link points to publishes a `display` for each of these (e.g. "Two times a day" for BID), and Inferno typically wants the published display verbatim. If the no-op entries exist because the original `interval_notes` already produced the right string for those codes and only the others needed correction, the no-ops are dead lookups — drop them and let the `?? $dataRecord['interval_notes']` fallback handle the rest. If the no-ops are intentional overrides, please confirm against the code system that those exact strings are valid displays, because to a reader they look like placeholders.

2. No unit test pins any of the new display values. The PR's correctness contract is "for code X in system Y, FHIR output's `coding[0].display` is Z" — that's exactly the kind of thing a small `parseOpenEMRRecord`-level assertion catches in 50ms, vs. discovering a regression next time someone runs the full Inferno suite. A handful of assertions across MedicationRequest category/timing/route, DocumentReference type, and DiagnosticReport / Observation laboratory codes would lock the contract you're establishing here.

Two smaller inline points and an architectural observation in the comments.

PR #11949, opened 2026-05-01
fix(login): default login_page_layout when globals row is missing

### LINKED ISSUE #11931: fix(login): login page 500s when login_page_layout global is null
## Problem

The login page returns HTTP 500 when the `login_page_layout` global is missing or null.

Stack trace (from production PHP error log):

```
TypeError: OpenEMR\Events\Core\TemplatePageEvent::setTwigTemplate():
  Argument #1 ($template) must be of type string, null given,
  called in src/Events/Core/TemplatePageEvent.php on line 58
  at src/Events/Core/TemplatePageEvent.php:126
#0 src/Events/Core/TemplatePageEvent.php(58): setTwigTemplate(NULL)
#1 interface/login/login.php(266): TemplatePageEvent->__construct(
     'login/login.php', Array, NULL, Array)
```

## Root cause

`interface/login/login.php:65` reads:

```php
$layout = $globalsBag->get('login_page_layout');
```

…which returns `null` when the row is absent from the `globals` table (e.g. on a freshly-installed site or one whose 7.0.3 → 7.0.4 upgrade did not complete). That `null` is then passed at `login.php:266` as the third argument to `TemplatePageEvent::__construct()`:

```php
$templatePageEvent = new TemplatePageEvent('login/login.php', [], $layout, $viewArgs);
```

The constructor's signature (`src/Events/Core/TemplatePageEvent.php:54`) declares `$twigTemplate = ""` with no parameter type, but immediately forwards it to `setTwigTemplate(string $template)` (line 126), which throws on `null`.

## Impact

Login is unreachable on any site where `globals.login_page_layout` is not set. There is no way for the user to recover without DB intervention.

## Suggested direction

Either:

- Make the call site tolerant of a missing global (provide a default at the read), or
- Make `TemplatePageEvent` accept `?string` for the template, treating `null` as "no override".

Implementer's choice. Both are reasonable; the second also hardens any other callers that pass through the same code path.

kojiromike:
## Summary

Fixes #11931 — the login page returned HTTP 500 when the `login_page_layout` global was absent (freshly seeded databases that haven't run the 7.0.0 → 7.0.1 upgrade). `interface/login/login.php` read it as null and passed it into `TemplatePageEvent`, whose `setTwigTemplate(string)` then `TypeError`'d, locking users out with no way to recover.

## Changes

- **`interface/login/login.php`**: default the layout to the documented `login/layouts/vertical_band.html.twig` (matches the default declared in `library/globals.inc.php:520`). Primary fix — restores login on broken databases.
- **`src/Events/Core/TemplatePageEvent.php`**: add a `: self` return type to `setTwigTemplate()` for fluent use; clears one PHPStan baseline entry.
- **`src/RestControllers/AuthorizationController.php`** + **`src/RestControllers/SMART/SMARTAuthorizationController.php`**: typed the two private `renderTwigPage` / `renderTwigJson` methods as a source-fix for upstream `mixed → string` flow that previously needed PHPStan baseline entries.
- **PHPStan baseline**: net **-80 lines** (no additions; obsolete entries removed by the source fixes above).

## Testing

- New `tests/Tests/Isolated/Events/Core/TemplatePageEventTest.php` (4 cases) covering construction, defaults, round-trip, and fluent return.
- Full isolated suite: 2,776 pass.
- `composer phpstan` clean. `composer phpcs`, `composer rector-check`, `composer require-checker` clean.

## Test plan

- [ ] On a database missing `login_page_layout`: hit the login page → renders with the default layout instead of HTTP 500.
- [ ] On a normally-installed database: login page renders with the configured layout (no behavior change).
- [ ] OAuth2 flow (Authorization, SMART): login screens still render correctly.

### DIFF
--- a/.phpstan/baseline/missingType.parameter.php
+++ b/.phpstan/baseline/missingType.parameter.php
@@ -33187,11 +33187,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../src/Events/Core/TemplatePageEvent.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Method OpenEMR\\\\Events\\\\Core\\\\TemplatePageEvent\\:\\:__construct\\(\\) has parameter \\$twigTemplate with no type specified\\.$#',
-    'count' => 1,
-    'path' => __DIR__ . '/../../src/Events/Core/TemplatePageEvent.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Method OpenEMR\\\\Events\\\\Core\\\\TemplatePageEvent\\:\\:__construct\\(\\) has parameter \\$twigVariables with no type specified\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../src/Events/Core/TemplatePageEvent.php',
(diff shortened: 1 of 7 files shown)


kojiromike:
@aisle-analyzer

PR #11956, opened 2026-05-01
refactor(crypto): migrate call sites to encryptForDatabase/decryptFromDatabase helpers

Firehed:
## Summary

- Migrates all `encryptStandard()`/`decryptStandard()` call sites that use `KeySource::Drive` (the default) to the new helpers
- Improves type safety: `decryptFromDatabase()` throws `CryptoGenException` on failure instead of returning `false`, and returns `string` (not `string|false`)
- Adds plaintext passthrough support in `decryptFromDatabase()` for handling mixed encrypted/unencrypted data

This solves a number of issues:

- The previous code path could silently fail, resulting in broken operations and/or insecure configurations (at least one place would use an empty "key" for something else, though I didn't dig _super_ far in to see how consequential that is). It now throws for a loud failure, which should de-risk corruption or insecure operation
- Type safety is improved, partially through the above
- Nearly all data destined for storage in the DB goes through a single known path, improving visibility of what data needs to be updated during a key rotation. This also simplifies toggles for external key management in Transparent Data Encryption (TDE) environments.

**Files NOT migrated (intentionally excluded):**
- `EventAuditLogger.php` - has its own `enable_auditlog_encryption` gate
- `SMARTLaunchToken.php` - see #12072

There were a handful of other touch-points that didn't quite fit the model which were split into their own PRs. #12050, #12062, #12066, #12069.


Firehed:
@aisle-analyzer

src/Billing/PaymentGateway.php
```diff
@@ -31,8 +31,8 @@ public function __construct($name)
         $this->production = !OEGlobalsBag::getInstance()->getBoolean('gateway_mode_production');
 
         $cryptoGen = ServiceContainer::getCrypto();
-        $this->apiKey = $cryptoGen->decryptStandard(OEGlobalsBag::getInstance()->getString('gateway_api_key'));
-        $this->transactionKey = $cryptoGen->decryptStandard(OEGlobalsBag::getInstance()->getString('gateway_transaction_key'));
+        $this->apiKey = $cryptoGen->decryptFromDatabase(OEGlobalsBag::getInstance()->getString('gateway_api_key'));
+        $this->transactionKey = $cryptoGen->decryptFromDatabase(OEGlobalsBag::getInstance()->getString('gateway_transaction_key'));
```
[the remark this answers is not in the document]
Firehed: This is desirable behavior (failing loudly, let the global handlers deal with it), and the suggestion would not help, just add complexity.

Firehed:
Feedback for Aisle comments:

1) #11958
2) Fixed inline, a8885a3
3) Invalid
4) #11972

interface/patient_file/front_payment.php
```diff
@@ -1584,7 +1584,7 @@ class="form-control"
             ?>
             <script>
                 var ccerr = <?php echo xlj('Invalid Credit Card Number'); ?>
-                var apiLoginID = <?php echo json_encode($cryptoGen->decryptStandard($globalsBag->getString('gateway_api_key'))); ?>;
+                var apiLoginID = <?php echo json_encode($cryptoGen->decryptFromDatabase($globalsBag->getString('gateway_api_key'))); ?>;
```
Firehed, line 1587: Since it's not obviously visible in the PR diff: this is on a branch where the value is a misnomer and it's actually a public key. There's a comment a few lines above clarifying this.

portal/lib/paylib.php
```diff
@@ -52,27 +53,31 @@
...
-    echo 'ok';
+        SessionUtil::setSession('whereto', '#paymentcard');
+
+        SaveAudit($form_pid, $invoice, $ccaudit);
+
+        echo 'ok';
+    } catch (CryptoGenException) {
+        echo 'Decryption failed';
+    }
 }
```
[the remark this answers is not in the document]
Firehed: Existing issue. Fixed in #11958.

.phpstan/baseline/openemr.exitInCatchOrFinally.php
```diff
@@ -66,6 +66,11 @@
     'count' => 1,
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-weno/scripts/weno_log_sync.php',
 ];
+$ignoreErrors[] = [
+    'message' => '#^exit/die inside a catch block swallows the caught exception and aborts the process\\.$#',
```
Firehed, line 70: There are a bunch of these already _outside_ of catch; I don't really want to fight with it right now.

interface/modules/custom_modules/oe-module-weno/src/Services/TransmitProperties.php
```diff
@@ -458,7 +459,11 @@ public static function echoError($errors): void
     public function cipherPayload(): string
     {
         $cipher = "aes-256-cbc"; // AES 256 CBC cipher
-        $enc_key = $this->cryptoGen->decryptStandard(OEGlobalsBag::getInstance()->get('weno_encryption_key'));
+        try {
+            $enc_key = $this->cryptoGen->decryptFromDatabase(OEGlobalsBag::getInstance()->get('weno_encryption_key'));
+        } catch (CryptoGenException) {
+            return "error";
```
Firehed, line 465: yes, this is goofy. follows existing patterns :(

Firehed:
Claude security review (I'm hesitant to invoke Aisle since it posts existing stuff as well right now):

Based on the validation results, I need to filter out all findings with confidence scores below 8:

- **Finding 1** (plaintext passthrough): Confidence 2/10 - FALSE POSITIVE (intentional documented behavior)
- **Finding 2** (edit_globals.php): Confidence 2/10 - FALSE POSITIVE (admin-only page, no attack vector)  
- **Finding 4** (LogProperties deterministic key): Confidence 7/10 - Below threshold (guard partially mitigates, requires key corruption + HTTPS interception)

All identified potential vulnerabilities scored below the 8/10 confidence threshold required for reporting.

---

# Security Review: crypto-database-helpers-migration

## Summary

**No high-confidence security vulnerabilities identified.**

This PR migrates cryptographic operations from `CryptoGen::encryptStandard()`/`decryptStandard()` to new `encryptForDatabase()`/`decryptFromDatabase()` helper methods. The migration introduces:

1. **Plaintext passthrough behavior** - `decryptFromDatabase()` returns unencrypted values unchanged rather than returning `false`. This is intentional, documented behavior for handling legacy unencrypted data during migration.

2. **Exception-based error handling** - Decryption failures now throw `CryptoGenException` instead of returning `false`, making failures more visible.

3. **Explicit key source** - The new methods use `KeySource::Drive` explicitly.

## Findings

No findings met the confidence threshold (≥8/10) required for reporting.

## Notes for Consideration

While not meeting the threshold for security vulnerabilities, the following code quality observations were noted during review:

- Some call sites removed explicit try/catch blocks from the old pattern, relying on exception propagation. This is appropriate behavior in most contexts.
- The `LogProperties.php` module has pre-existing cryptographic weaknesses (hardcoded zero IV) that are not introduced by this PR but could be addressed separately.
- The plaintext passthrough design is explicitly documented in `CryptoInterface.php` and serves the legitimate purpose of transparently handling unencrypted legacy data during migration.

Firehed:
I've been running this locally and doing all sorts of data toggles for various things that is situationally encrypted, and I didn't run into any issues. But I would NOT pull it forward into 8.1.0 unless there's a good external reason for it - I've tested it as best I can with the infra we have and manually, but letting it bake longer would be preferable before it goes into a release.

PR #11973, opened 2026-05-01
feat(crypto): add database encryption opt-out setting

Firehed:

Adds a global setting `database_encryption` that allows installations using database-level encryption (e.g., TDE) to opt out of column-level encryption for data stored via `encryptForDatabase()`.

Motivation: while the current cryptographic operations in OpenEMR are sound, the key management process is _extremely_ fragile, and high-risk. One bad patch can permanently destroy a production installation if backup procedures are insufficient.

Fortunately, modern database-level encryption technologies such as TDE are sufficient[^reqs] for remaining compliant; relying on this for protecting sensitive data drastically reduces the risk of permanent, unrecoverable damage due to a bad file move.

[^reqs]: As best as we've been able to figure out from reading the requirements, at least. A lot of standards defer to other standards, so it's a bit of a moving target.

(plus, most of the data in the database that can be stored encrypted is internal-use-only such as config values, not patient data)

Checking for sufficient DB-level encryption settings will be included as part of an oft-discussed future OCE certification checker module, and an OpenEMR installation that opted-out would get red-flagged if external validation failed.

- Add `database_encryption` boolean setting to `library/globals.inc.php` (defaults to enabled)
- Add `shouldEncryptForDatabase` property to `CryptoGen` that reads from the global setting
- Add `shouldEncryptForDatabase` constructor parameter to `BC\Crypto\Crypto`
- When opted out, `encryptForDatabase()` returns plaintext; `decryptFromDatabase()` passes through unencrypted values
- Fix missing parameter in `Crypto::instance()` static factory
- Add comprehensive tests for opt-out behavior in both crypto implementations

Commits include `Assisted-by: Claude Code` trailers.


Firehed:
https://github.com/openemr/openemr/pull/11956 is intended to land first, but there's not a strict interdependency between them.

src/Common/Crypto/CryptoGen.php
```diff
@@ -148,67 +161,6 @@ public function cryptCheckStandard(?string $value): bool
         }
     }
 
-    /**
```
Firehed, line 151: This was supposed to have been removed in #12096, some sort of merge went sideways.

.phpstan/baseline/openemr.forbiddenInstantiation.php
```diff
@@ -263,7 +263,7 @@
 ];
 $ignoreErrors[] = [
     'message' => '#^Direct instantiation of OpenEMR\\\\Common\\\\Crypto\\\\CryptoGen is discouraged\\. Use OpenEMR\\\\BC\\\\ServiceContainer\\:\\:getCrypto\\(\\) instead\\.$#',
-    'count' => 2,
+    'count' => 8,
```
Firehed, line 266: Testing, not an issue.

PR #11999, opened 2026-05-04
fix(ui): Fix crash from uncaught class in birthday popup

Firehed:
```
openemr-1  | NOTICE: PHP message: OpenEMR.CRITICAL: Uncaught exception! {"exception":"[object] (Error(code: 0): Class \"echoOEGlobalsBag\" not found at /usr/share/nginx/html/openemr/interface/patient_file/birthday_alert/birthday_pop.php:33)
```

I suspect this was the root cause for #11983 (etc) but even without the crash I bet the e2e tests would be too fragile to proceed.

- Fixes the missing space
- Updates baseline to reflect fix

No

### DIFF
--- a/interface/patient_file/birthday_alert/birthday_pop.php
+++ b/interface/patient_file/birthday_alert/birthday_pop.php
@@ -30,7 +30,7 @@ $session = SessionWrapperFactory::getInstance()->getActiveSession();
 </head>
 <body>
     <div style="padding: 15px; text-align: center">
-        <p class="h2"><?php echo xlt('Happy Birthday');?>&ensp;<img src="<?=OEGlobalsBag::getInstance()->getKernel()->getImagesRelative()?>/balloons-154949_960_720.png" height="42" width="42"></p>
+        <p class="h2"><?php echo xlt('Happy Birthday');?>&ensp;<img src="<?php echo OEGlobalsBag::getInstance()->getKernel()->getImagesRelative()?>/balloons-154949_960_720.png" height="42" width="42"></p>
 
         <?php if (OEGlobalsBag::getInstance()->getBoolean('patient_birthday_alert_manual_off')) { ?>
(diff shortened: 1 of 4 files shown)


interface/patient_file/birthday_alert/birthday_pop.php
```diff
@@ -30,7 +30,7 @@
 </head>
 <body>
     <div style="padding: 15px; text-align: center">
-        <p class="h2"><?php echo xlt('Happy Birthday');?>&ensp;<img src="<?php echoOEGlobalsBag::getInstance()->getKernel()->getImagesRelative()?>/balloons-154949_960_720.png" height="42" width="42"></p>
+        <p class="h2"><?php echo xlt('Happy Birthday');?>&ensp;<img src="<?=OEGlobalsBag::getInstance()->getKernel()->getImagesRelative()?>/balloons-154949_960_720.png" height="42" width="42"></p>
```
kojiromike: The repo overwhelmingly prefers `<?php echo ?>` over `<?= ?>` (~14,600 vs. ~20 occurrences), and the existing `<?php echo xlt('Happy Birthday');?>` on this same line is in the long form. Would it make sense to just add the missing space and keep the rest as-is? It's a smaller diff and stays consistent with the surrounding code:

```suggestion
        <p class="h2"><?php echo xlt('Happy Birthday');?>&ensp;<img src="<?php echo OEGlobalsBag::getInstance()->getKernel()->getImagesRelative()?>/balloons-154949_960_720.png" height="42" width="42"></p>
```

Firehed:
~@kojiromike I was going to update that fatal baseline file thing to reflect a class.notFound removal, but the baseline generator script says there are 208 entries (vs 182 in the current cap file) so it wasn't clear how to proceed.~ the isolated tests provided guidance, didn't initially notice.

PR #12000, opened 2026-05-04
feat(crypto): Refactor filesystem encryption

Firehed:

Corollary to #11956

Adds higher-level `encryptForFilesystem()` and `decryptFromFilesystem()` methods to `CryptoInterface` that encapsulate the encryption-enabled check and correct key source selection, simplifying call sites throughout the codebase.

These handle several things that were implemented inline:

- Handling "is encryption enabled" on the encrypt path (`drive_encryption` setting)
- Handling "is encrypted" on the decrypt path (`checkCryptStandard`)
- Passing the appropriate `KeySource`
- Error handling

Importantly, there was a setting for `couchdb_encryption` that behaved similarly to `drive_encryption`. These are now controlled by the same setting. As discussed on the 2026-05-05 admin call, having these separate is likely to lead to confusion and errors. When upgrading, you should have these values in sync to reduce transitional confusion, but there should be no runtime harm if they're not the same value.

> [!IMPORTANT]
> Again, this means the `couchdb_encryption` config will have no effect, and `drive_encryption` handles all document storage.

- Add `encryptForFilesystem()` and `decryptFromFilesystem()` to `CryptoInterface`
- Implement in both `CryptoGen` and `BC\Crypto\Crypto`
- `encryptForFilesystem()` returns plaintext passthrough when drive encryption is disabled
- Uses `KeySource::Database` for filesystem operations (per existing crypto strategy)
- Migrate ~20 call sites from manual `encryptStandard`/`decryptStandard` + conditionals to the new methods
- Add `encryptForDatabase()` and `decryptFromDatabase()` for symmetry (database storage uses `KeySource::Drive`)
- Add comprehensive tests for the new methods
- Updated checks for couchdb setting to use drive; updated text on the setting to clarify the old one is unused and the new one controls both. This will be more forward-compatible as the Document model is migrated to use the Flysystem abstraction.

Commits include `Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>` trailer.

---


.phpstan/baseline/argument.type.php
```diff
@@ -706,13 +706,28 @@
     'count' => 1,
     'path' => __DIR__ . '/../../controllers/C_Document.class.php',
 ];
+$ignoreErrors[] = [
```
Firehed, line 709: Note: trades from below due to adding more type hints. Fixing these properly will occur through adoption of Flysystem in document handling.

library/classes/Document.class.php
```diff
@@ -1191,9 +1187,10 @@ function get_data($force_no_decrypt = false)
     public function decrypt_content($data)
     {
         $cryptoGen = ServiceContainer::getCrypto();
-        $decryptedData = $cryptoGen->decryptStandard($data, keySource: KeySource::Database);
-        if ($decryptedData === false) {
-            throw new RuntimeException("Failed to decrypt the data");
+        try {
+            $decryptedData = $cryptoGen->decryptFromFilesystem($data);
+        } catch (CryptoGenException $e) {
```
Firehed, line 1188: This is the one place I'm doing a catch/rethrow, since I'm not sure if anything downstream is explicitly looking for this runtime exception.

Firehed:
# Security Review Results

**No security vulnerabilities were identified in this PR.**

## Summary

The changes are purely a refactoring of encryption logic with no new attack surfaces introduced:

| Category | Finding |
|----------|---------|
| Input Validation | No changes to input handling |
| Authentication/Authorization | No changes to auth logic |
| Cryptographic Implementation | Methods delegate to existing `encryptStandard()`/`decryptStandard()` using the same AES-256-CBC with HMAC-SHA384 |
| Key Management | Same cross-encryption model preserved (`KeySource::Database` for filesystem ops) |
| Error Handling | Improved - throws `CryptoGenException` instead of returning `false` |
| Backwards Compatibility | Preserved - `cryptCheckStandard()` detects encrypted vs plaintext data |

## Positive Security Notes

1. **Dependency injection** added to `C_Document` for `CryptoInterface` improves testability
2. **Earlier encryption flag setting** in `Document.class.php` ensures consistency before storage operations
3. **Exception-based error handling** makes decryption failures explicit rather than silent

---
🤖 This security review was performed by Claude Code

Firehed (commented):
The coverage warning is quite surprising seeing that this is a net reduction in code and increase in tests. I suspect it's some goofy thing relating to Inferno.

Firehed:
I think this is probably best not merged into 8.1.0 due to amount of time for testing, but it could go either way. You could easily argue this is a proactive security fix since encryption-related errors are less likely to be ignored.

PR #12011, opened 2026-05-05
fix(upgrade): drop redundant categories_seq UPDATE in eye form laser migration

### LINKED ISSUE #11930: fix(sql-upgrade): categories_seq UPDATE collides on PK when table has multiple rows

stephenwaite:
fixes #11930

The categories_seq UPDATE fatals the demo data upgrade when
categories_seq has drifted from its single-row shape (duplicate
PRIMARY KEY → sqlStatementNoLog calls HelpfulDie → upgrade aborts).

remove unnecessary sql to categories_seq

### DIFF
--- a/.phpstan/baseline/openemr.deprecatedSqlFunction.php
+++ b/.phpstan/baseline/openemr.deprecatedSqlFunction.php
@@ -7563,7 +7563,7 @@ $ignoreErrors[] = [
 ];
 $ignoreErrors[] = [
     'message' => '#^Use QueryUtils\\:\\:fetchRecordsNoLog\\(\\) instead of sqlStatementNoLog\\(\\)\\.$#',
-    'count' => 24,
+    'count' => 23,
     'path' => __DIR__ . '/../../src/Services/Utils/SQLUpgradeService.php',
 ];
 $ignoreErrors[] = [
--- a/src/Services/Utils/SQLUpgradeService.php
+++ b/src/Services/Utils/SQLUpgradeService.php
@@ -737,8 +737,6 @@ class SQLUpgradeService implements ISQLUpgradeService
                             $eyeFormCategoryParent['rght'] + 5
                         ]
                     );
-                    // update root node
-                    sqlStatementNoLog("UPDATE `categories` SET `rght` = (SELECT MAX(`rght`) FROM `categories`) + 1 WHERE `id` = 1");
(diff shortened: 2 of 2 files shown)


stephenwaite:
discovered while working in the demo farm, it seems mariadb10.6 exposed this since couldn't replicate it in easy dev

PR #12019, opened 2026-05-05
fix(sql): handle table-qualified columns in escape_sql_column_name()

Firehed:

Fixes regression in `escape_sql_column_name()` when `$long=true` (table-qualified column names like `pnotes.date`), introduced in #11280.

The `$long=true` path had two bugs after backtick-quoting was added:

1. **Whitelist mismatch**: `$columns_options` was built using backtick-quoted table names (e.g., `` `pnotes`.date ``), but input comes unquoted (`pnotes.date`), so validation always failed.

2. **Incorrect quoting**: Output wrapped the entire `table.column` string in backticks (`` `pnotes.date` ``), but MySQL interprets that as a column literally named "pnotes.date". Correct output is `` `pnotes`.`date` ``.

### DIFF
--- a/library/formdata.inc.php
+++ b/library/formdata.inc.php
@@ -138,14 +138,21 @@ function escape_sql_column_name($s, $tables, $long = false, $throwException = fa
     $columns_options = [];
     foreach ($tables_escaped as $table_escaped) {
         $res = sqlStatementNoLog("SHOW COLUMNS FROM " . $table_escaped);
+        // Strip backticks for whitelist comparison; input won't have them
+        $table_for_whitelist = trim($table_escaped, '`');
         while ($row = sqlFetchArray($res)) {
-            $columns_options[] = $long ? $table_escaped . "." . $row['Field'] : $row['Field'];
+            $columns_options[] = $long ? $table_for_whitelist . "." . $row['Field'] : $row['Field'];
         }
     }
 
     // Whitelist against actual columns, then backtick-quote to keep in identifier context
     $dieIfNoMatch = !$throwException;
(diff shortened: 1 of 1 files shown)


Firehed:
Tagging into 8.1.0. The message center has been totally broken for me on master, but not on any currently-tagged release. The regression would end up in 8.1.0.

But if nobody else can repro this on master (default installation docker dev-easy, admin/pass login, try to view message center) maybe it's worth holding off.

Firehed:
# Security Review: `fix/escape-sql-column-name-long-mode`

## Summary

**No high-confidence security vulnerabilities found.**

## Analysis

The changes to `escape_sql_column_name()` in `library/formdata.inc.php` were reviewed for SQL injection and other security concerns.

### Security Controls Verified

1. **Input Sanitization**: Backticks in user input `$s` are rejected before any processing (existing check at line 117-119)

2. **Whitelist Validation**: The `escape_identifier()` function uses `array_search()` for exact-match whitelist validation against actual database column names retrieved via `SHOW COLUMNS`

3. **Safe Output Construction**: The returned value is the matched whitelist entry (from database schema), not the user input, eliminating any injection path

### Changes Reviewed

| Change | Security Impact |
|--------|-----------------|
| Strip backticks from `$table_escaped` for whitelist comparison | **Neutral** - Necessary for comparison since input cannot contain backticks |
| Separate backtick-quoting for `table.column` format | **Positive** - Correct SQL identifier quoting for multi-part identifiers |

### Conclusion

The changes fix a functional bug in the `$long=true` path without weakening security. The existing defense-in-depth controls (backtick rejection, exact-match whitelisting, schema-derived output) remain intact.

PR #8603, opened 2025-07-11
Various Bugs and fax/sms Module Upgrade

sjpadgett:
Fixes #8602 
Fixes #8628

#### Short description of what this resolves:
Fix Demographics view state persisting
Upgrade fax/sms module with new VoIP widget and setting user permissions.

#### Changes proposed in this pull request:
Card name is wrong. 'demographic' versus the correct name 'demographics'

PR #8585, opened 2025-07-10
New web phone service fax/sms module

sjpadgett:
Fixes #8531

#### Changes proposed in this pull request:
New web phone service fax/sms module
several new classes to support voice services including communication event handlers.
move authentication to traits

PR #12020, opened 2026-05-06
refactor(faxsms): remove unimplemented VoiceClient

Firehed:
I came across this while working on cryptography refactors (#11956 primarily). VoiceClient had some crypto tendrils. It seems like it never worked, so I opted to remove it and keep going through all of the touch points.

I made a best-effort approach to follow the LLM claims about this being unreachable or permanently-broken code. The handler logic within the module is rather tricky to trace, but I couldn't find any path to it working as it exists in the project.

🤖 :
## Summary

- Remove `VoiceClient` class and all related voice integration code from the faxsms module
- Remove `ServiceType::VOICE` enum case and voice channel configuration
- Remove voice UI elements from setup pages and user permissions
- Remove `phone_widget.html.twig` template

## Why this is safe

`VoiceClient` was scaffolding for a RingCentral voice integration that was **never implemented**. Git history confirms this:

| Date | Commit | What happened |
|------|--------|---------------|
| Jul 12, 2025 | 3ea9f6ad4 (#8585) | Initial creation with TODO stubs |
| Jul 18, 2025 | 09d2ece84 (#8603) | Added getVoiceCredentials() (credential retrieval only) |
| Later | 5 commits | Infrastructure refactoring (globals, crypto, sessions) |

The core methods have been empty stubs since day one:

```php
function sendFax(): string|bool { return ''; /* TODO */ }
function sendSMS(): mixed { return ''; /* TODO */ }
function sendEmail(): mixed { return ''; /* TODO */ }
function fetchReminderCount(): string|bool { return ''; /* TODO */ }
```

**The class could not function even if instantiated:**

1. Constructor calls `initializeSDK()` but never calls `authenticate()` — the RingCentral SDK would have no access token
2. All API methods are stubs returning empty strings
3. The trait's auth caching (getCachedAuth, cacheAuthData) is never exercised
4. A self-referential `private VoiceClient $client;` property was declared but never used

**No code path ever used VoiceClient for actual voice functionality.** The only entry point (setup_voice.php) just saves credentials to the database — it never makes API calls.

## Test plan

- [ ] PHPStan passes (baseline regenerated)
- [ ] Existing fax/SMS/email functionality unaffected
- [ ] Module setup page loads without voice options
- [ ] User permissions table shows fax/SMS/email columns only


bradymiller (approved):

sjpadgett:
I was asked to resolve or explain why this was removed. This is a sponsored featured.
This is not broken for any version 7.0.4 through 8.0.3. I've tested this up through 8.0.3. 
@bradymiller Why would you approve this removal w/o consulting me? Also fyi all module clients/services are now broken in 8.1.0
The entire justification noted above is bogus and I'll be opening issues to at least fix module to still function and not break on upgrade for the many production practices using this module.
@Firehed Hopefully you won't take this as being harsh as I'm sure your intent included reducing foot print which overall I'm all for! I'm just disappointed no one pinged me for some advise

image

PR #12046, opened 2026-05-06
ci(release): drop single-entry matrix from release-permissions-check

kojiromike:
Manual one-shot probe; the single-entry `php-version` matrix added no value and made the job name harder to read. Inlines `php-version: '8.5'` into the setup-php-composer step.

Follow-up to #11896.


kojiromike:
Closing — keeping the php-version matrix. Only the runner-side matrix in website-openemr#97 is the cleanup intended here.

PR #12050, opened 2026-05-07
fix(faxsms): use filesystem-targeted encryption for cached auth

Firehed:
I came across a cryptointerface touch-point during a refactor that seemed to be using the wrong keysource (and thus method). I pulled it out so it doesn't get swept up in an invalid change.

## Summary
- Switch from `encryptStandard`/`decryptStandard` to `encryptForFilesystem`/`decryptFromFilesystem` for cached RingCentral authentication data
- This results in using the correct key source (database key) for filesystem-stored data per CryptoGen conventions
- Adds explicit error handling for decryption failures with automatic cache cleanup
- Uses `JSON_THROW_ON_ERROR` for stricter JSON validation

### DIFF
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AuthenticateTrait.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AuthenticateTrait.php
@@ -99,14 +99,14 @@ trait AuthenticateTrait
             }
 
             $json = $this->crypto->decryptFromFilesystem($contents);
-            $data = json_decode($json, true, flags: JSON_THROW_ON_ERROR);
+            $data = json_decode((string) $json, true, flags: JSON_THROW_ON_ERROR);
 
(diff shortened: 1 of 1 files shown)


interface/modules/custom_modules/oe-module-faxsms/src/Controller/AuthenticateTrait.php
```diff
@@ -80,23 +81,40 @@ private function authenticateRingCentral(): int|string
...
+
+            $contents = file_get_contents($file);
+            if ($contents === false) {
+                // exists but not readable
+                $error = true;
+                return [];
+            }
+
+            $json = $this->crypto->decryptFromFilesystem($contents);
+            $data = json_decode((string) $json, true, flags: JSON_THROW_ON_ERROR);
```
Firehed, line 102: Absolutely no amount of coercion can get rector to not add this useless cast. It's not context-aware about traits.

PR #12070, opened 2026-05-08
fix(cors): use => instead of , so Access-Control-Allow-Methods is actually sent

harrywinner2:
The associative-array entry on the methods line of `CORSListener::getInitialResponse()` is malformed: a comma between key and value instead of `=>`. PHP parses the result as two positional entries `["Access-Control-Allow-Methods", "GET, HEAD, ..."]`, which Symfony's `HeaderBag` silently drops because they have no string key.

Net effect: every CORS preflight response built by `getInitialResponse()` omits `Access-Control-Allow-Methods`, and any browser doing a non-simple cross-origin request (anything carrying `Authorization`, a `Content-Type` other than the simple subset, etc.) blocks the actual request after what looks server-side like a successful preflight.

One-character fix.

```diff
-            "Access-Control-Allow-Methods", "GET, HEAD, POST, PUT, DELETE, PATCH, TRACE, OPTIONS"
+            "Access-Control-Allow-Methods" => "GET, HEAD, POST, PUT, DELETE, PATCH, TRACE, OPTIONS"
```

### DIFF
--- a/src/RestControllers/Subscriber/CORSListener.php
+++ b/src/RestControllers/Subscriber/CORSListener.php
@@ -66,7 +66,7 @@ class CORSListener implements EventSubscriberInterface
         $response = new Response('', Response::HTTP_OK, [
             'Access-Control-Allow-Credentials' => 'true',
             "Access-Control-Allow-Headers" => "origin, authorization, accept, content-type, content-encoding, x-requested-with",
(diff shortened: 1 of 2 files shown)


CloCkWeRX:
That... seems like a bit of an oversight!

Would a regression test be useful to add?

harrywinner2:
Thats a good call. Let me cook one up and add it to the PR real quick.

CloCkWeRX:
Welp, as an outsider to the project, LGTM!

kojiromike (approved):

Issue #12118, opened 2026-05-11
reactor(audit): Improve immutability and tamper-resistance


Firehed:
Today, the event logging system uses encryption as a means of trying to address certification requirements around immutability and tamper-resistance. The logic is well-intentioned, but somewhat faulty.

[ONC 170.315(d)(10)](https://www.ecfr.gov/current/title-45/subtitle-A/subchapter-D/part-170/subpart-C/section-170.315) states, in effect, that the audit log must be write-once and offer tamper-resistance. Neither of those are particularly solved by encryption[^sign]; the checksumming is certainly the right direction, but since it's a row-level checksum that operates independent of other rows, it wouldn't reveal stuff like deletions.

[^sign]: Encryption does of course have HMAC which indeed is a form of tamper-resistance, but HMACs can be used without encryption which is effectively what the current checksum column is (or aims to be). This still is scoped to row-level tamper resistance, offering nothing in terms of broad immutability. There's _nothing_ stopping an actor that has internal access to just... write code that would re-encrypt a row and still pass HMAC.

Proposal stages:

- Remove encryption from the audit logs (on all columns that use it). The encryption increases risk (loss of access to audit logs) and complexity (key management), but does not actually solve any regulatory needs. I've not found any places where it would contain sensitive data.
  - Deprecate/remove/no-op the `enable_auditlog_encryption` setting 
  - The existing decrypt path can stay to offer a straightforward migration path, but encryption should be removed.
  - Eventually, an upgrade process should perform a decryption (and re-checksumming, if need be) of historic data. Remove the `encrypt` column in the metadata table. Version also probably goes.
- Update table permissions: `log` (and `log_comment_encrypt`) should have permissions set at a MySQL level to allow `INSERT` and `SELECT`, but not `UPDATE`, `DELETE`, `TRUNCATE`, etc. Ideally, the "default" openemr DB user has no access to the table(s) at all; a new user with minimum permissions should exist for dedicated access
- Have some sort of checksum chaining, similar to how git commit hashes work. Tamper detection could examine the chain, re-checksum, and ensure there are no mismatches (this would be very slow in large installs, but I'm not sure that's avoidable)

Realistically, deprioritizing the internal audit logger in favor of ATNA is the best long-term plan, as it's better designed to solve this problem. This also simplifies the integration, since this would mean OpenEMR itself actually has _no_ direct write access to log tables (read maybe, TBD, depends if the log viewer needs to stay in OE at all)

PR #12123, opened 2026-05-11
refactor(audit): Remove redundant encryption from audit log writes

### LINKED ISSUE #12120: refactor(audit): Remove redundant encryption
See #12118 for motivation and details. In short, it's not needed for compliance

Firehed:
Removes support for writing encrypted data to the audit logs. Existing data reads will continue to support decrypting existing data, and have been marked with a reference to this task for added context.

This is the first step towards #12118, though it doesn't complete it. Fixes #12120.

tl;dr: the Audit Log, per ONC (see linked issue), needs to be append-only and tamper-resistant+tamper-evident, but encryption is not part of the requirements. Encryption (which defaulted to off) does not enhance immutability on top of the existing checksumming system. Its presence is really just a liability, since it risks locking you out of critical audit data if there's an issue with key management.

> 170.315(d)(10) 
> (i) By default, be set to record actions related to electronic health information in accordance with the standard specified in [§ 170.210(e)(1)](https://www.ecfr.gov/current/title-45/section-170.210#p-170.210(e)(1)).
> (ii) If technology permits auditing to be disabled, the ability to do so must be restricted to a limited set of users.
> (iii) Actions recorded related to electronic health information must not be capable of being changed, overwritten, or deleted by the technology.
> (iv) Technology must be able to detect whether the audit log has been altered.

This change makes no regressions on certification requirements. There are follow-up issues linked from the parent to enhance the processes described in subsections iii and iv, independent of this change.

The encryption is also _uniquely_ problematic in the application setup, bootstrapping, and runtime: there's a patched-around circular dependency between the database and the audit logger. DB queries go through audit logging, audit logging could use encryption, and the encryption keys need to be obtained through a database connection (even when using the keys on the filesystem since they're encrypted with the db keys).

Through a combination of global state manipulation and careful routing through the "no log" query utilities, we've managed to work around it, but it's _extremely_ fragile and deeply incompatible with trying to move forward into a DI world. The fact that dropping audit log encryption fixes this is a huge bonus (I believe this means DBAL/ORM should be _relatively_ straightforward to progress on now, rather than requiring layers of indirection and complexity), but not directly a motivating factor.

- Remove `enable_auditlog_encryption` config value
- Remove reading it from EAL
- Unthread both the preference and the cryptography tools from EAL, the Audit Event DTO, and the sinks
- Remove now-unreachable test paths

Testing:
- CI
- Verifying setting no longer in the admin>config menu
- Verifying audit log screen still works, including when display data that was previously encrypted
- Verifying audit log tamper report still works, including data ranges including encrypted data

Review only.

### DIFF
--- a/interface/logview/logview.php
+++ b/interface/logview/logview.php
@@ -321,6 +321,7 @@ if (!empty($_GET)) {
                                                 $patterns = ['/^success/', '/^failure/', '/ encounter/'];
                                                 $replace = [xl('success'), xl('failure'), sprintf(' %s', xl('encounter'))];
 
+                                                // Note: new data no longer written encrypted. Kept for compatibility. See #12118.
                                                 $commentEncrStatus = !empty($iter['encrypt']) ? $iter['encrypt'] : "No";
                                                 $encryptVersion = !empty($iter['version']) ? $iter['version'] : 0;
 
--- a/interface/logview/logview.php
+++ b/interface/logview/logview.php
@@ -321,7 +321,7 @@ if (!empty($_GET)) {
                                                 $patterns = ['/^success/', '/^failure/', '/ encounter/'];
                                                 $replace = [xl('success'), xl('failure'), sprintf(' %s', xl('encounter'))];
 
-                                                // Note: new data no longer written encrypted. Kept for compatibility. See #12118.
+                                                // Note: new data no longer written encrypted. Kept for compatibility. See #12118+12120.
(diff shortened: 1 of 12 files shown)


bradymiller:
hi @Firehed ,
Agree not needed for ONC. Looked through historical ONC stuff and unable to find this requirement in their either. I dropped the isntruction to turn this on in https://www.open-emr.org/wiki/index.php/OpenEMR_8.1.0_ONC_Ambulatory_EHR_Certification_Requirements#Requirements

PR #12133, opened 2026-05-13
fix(physical-exam): preserve persisted diagnosis ordering on save

### LINKED ISSUE #12029: fix(physical_exam): edit_diagnoses overwrites custom ordering values with loop counter on Save
## Bug

`interface/forms/physical_exam/edit_diagnoses.php` renders the **Order** input from the loop counter `$i`, not from the persisted `ordering` column. If a physical-exam line has customized order values saved (e.g. `10, 20, 30`), opening the editor and clicking **Save** silently rewrites them to `1, 2, 3`.

## Reproduction

1. Open a Physical Exam encounter form for a patient.
2. Add diagnoses to a line and save with non-1-based ordering values (e.g. `10`, `20`, `30`).
3. Open the same line's "-- Edit --" diagnosis dialog.
4. Without changing anything, click **Save**.
5. Inspect `form_physical_exam_diagnoses` for that `line_id` — the `ordering` values are now `1, 2, 3`.

## Code

`interface/forms/physical_exam/edit_diagnoses.php`, current state (post #12018):

```php
<?php for ($i = 1; $drow = sqlFetchArray($dres); ++$i) { ?>
 <tr>
  <td><input type='text' size='3' maxlength='5' name='form_ordering[<?php echo attr($i); ?>]' value='<?php echo attr($i); ?>' /></td>
  <td><input type='text' size='20' maxlength='250' name='form_diagnosis[<?php echo attr($i); ?>]' value='<?php echo attr($drow['diagnosis']); ?>' style='width:100%' /></td>
 </tr>
<?php } ?>
```

The `value='<?php echo attr($i); ?>'` for the order input should be `value='<?php echo attr($drow['ordering']); ?>'`. The save loop in the same file at lines 53-64 already uses `$_POST['form_ordering']` correctly — the bug is purely in how the existing values are populated into the form.

## Fix

```diff
-  <td><input type='text' size='3' maxlength='5' name='form_ordering[<?php echo attr($i); ?>]' value='<?php echo attr($i); ?>' /></td>
+  <td><input type='text' size='3' maxlength='5' name='form_ordering[<?php echo attr($i); ?>]' value='<?php echo attr($drow['ordering']); ?>' /></td>
```

The five blank rows below the existing-row loop correctly default to the loop counter (continuing the numbering) — that part is fine. Only the existing-row branch is wrong.

## Provenance

`git log -L 99,103:interface/forms/physical_exam/edit_diagnoses.php` shows the bug has been present since 2006 (commit `17d4ccccb2`, "added physical_exam encounter form"). Surfaced by Copilot review on PR #12018 (the security fix that activated the previously-disabled ACL gate on this file — now that the gate runs, more users will reach the editor and the data-loss bug becomes more visible).

## Out of scope for #12018

#12018 is scoped to the security fix (ACL dead-code) and a baseline-noise revert. This ordering bug is a separate data-integrity issue and gets its own PR.

## Verification

- Manual: save customized orderings, open the editor, save without changes, confirm orderings persist.
- Add an isolated test if the form has any test coverage to bind to.

## Pre-flight reading

- `interface/forms/physical_exam/edit_diagnoses.php` — full file context.
- `interface/forms/physical_exam/new.php:151` — how the dialog is invoked.
- The `form_physical_exam_diagnoses` schema in `sql/`.

### LINKED ISSUE #12030: refactor(physical_exam): unify edit_diagnoses.php two for-loops into one foreach
## Why this ticket exists

Follow-up from Copilot review on #12018. `interface/forms/physical_exam/edit_diagnoses.php` renders existing diagnoses and 5 trailing blank rows using two `for` loops that share the `$i` counter across loop boundaries:

```php
<?php for ($i = 1; $drow = sqlFetchArray($dres); ++$i) { ?>
 <tr>
  <td><input type='text' size='3' maxlength='5' name='form_ordering[<?php echo attr($i); ?>]' value='<?php echo attr($i); ?>' /></td>
  <td><input type='text' size='20' maxlength='250' name='form_diagnosis[<?php echo attr($i); ?>]' value='<?php echo attr($drow['diagnosis']); ?>' style='width:100%' /></td>
 </tr>
<?php } ?>

<?php for ($j = 0; $j < 5; ++$j, ++$i) { ?>
 <tr>
  <td><input type='text' size='3' name='form_ordering[<?php echo attr($i); ?>]' value='<?php echo $i?>' /></td>
  <td><input type='text' size='20' name='form_diagnosis[<?php echo attr($i); ?>]' style='width:100%' /></td>
 </tr>
<?php } ?>
```

The cross-loop `$i` sharing is fragile (works at runtime because PHP's `for` init runs unconditionally, but not legible at a glance). The two loops also drift on markup details:

- Existing rows: `size='3' maxlength='5'` and `size='20' maxlength='250'`.
- Blank rows: `size='3'` (no maxlength) and `size='20'` (no maxlength).
- Existing rows escape the order value: `value='<?php echo attr($i); ?>'`.
- Blank rows skip the escape: `value='<?php echo $i?>'`.

PHPStan flags `$i might not be defined` 4 times in the second loop (verified — the openemr context analyzes the file at top level with PHP/HTML interleave, where variable tracking is coarser than inside a function body). Plus secondary entries in `echo.nonString.php` and `preInc.type.php` for the `mixed` $i reads.

## Proposed refactor

Build one list of (existing-or-blank) rows, render with one `foreach`. No `$i`, no second loop, no PHPStan complaint, consistent markup:

```php
$existing = [];
foreach (sqlStatement(
    "SELECT ordering, diagnosis FROM form_physical_exam_diagnoses WHERE line_id = ? ORDER BY ordering, diagnosis",
    [$line_id]
) as $drow) {
    $existing[] = $drow;
}
$rows = $existing;
$rows = array_pad($rows, count($existing) + 5, ['ordering' => null, 'diagnosis' => '']);
?>
...
<?php foreach ($rows as $idx => $drow):
    $i = $idx + 1;
    $ordering = $drow['ordering'] ?? $i;
    $diagnosis = $drow['diagnosis'] ?? ''; ?>
 <tr>
  <td><input type='text' size='3' maxlength='5' name='form_ordering[<?php echo attr($i); ?>]' value='<?php echo attr($ordering); ?>' /></td>
  <td><input type='text' size='20' maxlength='250' name='form_diagnosis[<?php echo attr($i); ?>]' value='<?php echo attr($diagnosis); ?>' style='width:100%' /></td>
 </tr>
<?php endforeach; ?>
```

Note this also fixes #12029 (the existing-row Order input value should come from `$drow['ordering']`, not from the loop counter — opening + saving without changes silently rewrites custom orderings to 1..N). The two refactors are naturally one PR — landing them separately would mean the foreach refactor either re-introduces the bug or has to be sequenced after #12029.

## What this drains

- `variable.undefined`: 4 occurrences / 1 entry. Cap moves 559 → 558.
- `echo.nonString`: 1 entry.
- `preInc.type`: 1 entry.
- `argument.type`: net 0 (the existing `attr(int<1, max>)` and `attr(mixed)` complaints both go away because `$i` and `$drow['ordering']` are both narrowed at the source — `$i` from `$idx + 1`, `$drow['ordering']` already in `attr()`-friendly shape).

## Why a separate PR from #12018

#12018 is scoped to the security fix (ACL dead-code) and the baseline-revert that followed Copilot review. The refactor + #12029 fix is a data-integrity bug fix that deserves its own diff and reviewer attention, and shouldn't ride on a security PR's review thread.

## Verification

- Open the editor for a line with no existing diagnoses → 5 blank rows render as `form_ordering[1]` through `form_ordering[5]`.
- Open the editor for a line with N existing diagnoses → N rows + 5 blank rows render with continuous numbering.
- Open the editor for a line with custom orderings (e.g. 10, 20, 30) → existing rows show 10/20/30 in the Order input (this is the #12029 fix).
- Save without changes → orderings persist.
- `composer phpstan-baseline` regenerates with all four expected baselines down (variable.undefined, echo.nonString, preInc.type — and net 0 on argument.type). Cap lowered to match.
- `composer phpunit-isolated -- --filter FatalBaselineCapsIsolatedTest` passes.

## Pre-flight reading

- `interface/forms/physical_exam/edit_diagnoses.php` — current shape.
- `interface/forms/physical_exam/new.php:151` — how the dialog is invoked.
- #12018 — security fix that landed the ACL gate; commit messages explain the PHPStan false positive.
- #12029 — the ordering-overwrite bug this refactor naturally fixes.

pradhankukiran:

Fixes #12029.
Fixes #12030.

The Physical Exam diagnosis editor rendered existing row order values from the row loop counter instead of the persisted `ordering` column. Opening the editor and saving without changes could rewrite custom orderings like 10, 20, 30 to 1, 2, 3.

- Build one existing-plus-blank diagnosis row list before rendering.
- Render existing and blank rows through one `foreach` instead of two shared-counter loops.
- Preserve persisted ordering values for existing rows while keeping 5 blank rows numbered after them.
- Move the diagnosis save loop into a small helper so the regression test can exercise the same save path as the page.
- Keep row markup consistent, including maxlength attributes and escaped values.
- Add a regression test that renders the editor, feeds the rendered values through the save helper, and verifies persisted ordering survives the save round trip.
- Remove the drained PHPStan baseline entries and lower the variable.undefined cap.

Validation:
- `php -l interface/forms/physical_exam/edit_diagnoses.php`
- `php -l interface/forms/physical_exam/diagnosis_helpers.php`
- `php -l tests/Tests/Services/Forms/PhysicalExam/EditDiagnosesTest.php`
- `composer phpcs -- --standard=phpcs.xml.dist interface/forms/physical_exam/edit_diagnoses.php interface/forms/physical_exam/diagnosis_helpers.php tests/Tests/Services/Forms/PhysicalExam/EditDiagnosesTest.php`
- `vendor/bin/phpstan analyze --memory-limit=4G --configuration=phpstan.neon.dist interface/forms/physical_exam/edit_diagnoses.php interface/forms/physical_exam/diagnosis_helpers.php tests/Tests/Services/Forms/PhysicalExam/EditDiagnosesTest.php`
- `vendor/bin/phpunit -c phpunit.xml --testsuite services --filter EditDiagnosesTest`
- `composer phpunit-isolated -- --filter FatalBaselineCapsIsolatedTest`

Codex assisted with issue review, implementation, and validation.


interface/forms/physical_exam/edit_diagnoses.php
```diff
@@ -99,17 +109,16 @@
...
-  <td><input type='text' size='3' maxlength='5' name='form_ordering[<?php echo attr($i); ?>]' value='<?php echo attr($i); ?>' /></td>
-  <td><input type='text' size='20' maxlength='250' name='form_diagnosis[<?php echo attr($i); ?>]' value='<?php echo attr($drow['diagnosis']); ?>' style='width:100%' /></td>
- </tr>
-<?php } ?>
-
-<?php for ($j = 0; $j < 5; ++$j, ++$i) { ?>
+<?php foreach ($diagnosisRows as $idx => $drow) {
+    $i = $idx + 1;
+    $orderingValue = $drow['ordering'] ?? $i;
+    $ordering = is_scalar($orderingValue) ? (string) $orderingValue : (string) $i;
```
kojiromike: What case is the `is_scalar($orderingValue)` guard defending against? `sqlFetchArray` on `SELECT ordering, diagnosis` returns string|null per column (the `??` above already handles null), and the schema for `form_physical_exam_diagnoses.ordering` is a small integer column, so `$orderingValue` is `string|int` by the time we reach this line. If this is here purely to silence PHPStan rather than to guard a real runtime case, a `(string) ($drow['ordering'] ?? $i)` would say the same thing more directly. Same question on line 117 for `diagnosis`.

kojiromike (changes requested):
The ordering fix is straightforward and correct: existing rows now read `ordering` from the persisted row instead of being overwritten with the loop counter, and the loop unification preserves render output for the existing-row case while keeping the five trailing blanks numbered sequentially.

A couple of items worth resolving before merge:

- The PR title is `refactor(physical-exam): ...`, but this PR closes #12030, which is a behavior bug (re-saving rewrites custom ordering). Conventional Commits would expect `fix(` here — a refactor by definition doesn't change behavior, and a release note generated from this title would not mention the ordering bug at all. Suggest retitling to something like `fix(physical-exam): preserve persisted diagnosis ordering on save` and noting the rendering unification in the body.
- No tests are added or updated. The pre-existing tree has no coverage of `edit_diagnoses.php` at all (rendering or save handler), so this PR isn't making things worse, but the regression that #12030 reports — open editor, save without changes, ordering values get rewritten — is exactly the kind of round-trip that a small integration test against the save handler + a re-render would catch and lock in. Worth at least acknowledging in the PR description whether you considered adding coverage, given the regression is silent and would only surface after the next save.

One inline question on the new render block.

interface/forms/physical_exam/edit_diagnoses.php
pradhankukiran: Addressed in d4ecc272f59791170cddb03b969d7f18d164b53c. I replaced the render-time `is_scalar()` guards with direct casts and added a narrow PHPStan type assertion where the `sqlFetchArray()` rows are collected, so the runtime path stays direct while static analysis still knows the selected column types.

pradhankukiran:
Thanks for the review. Addressed in d4ecc272f59791170cddb03b969d7f18d164b53c: retitled the PR to `fix(physical-exam): preserve persisted diagnosis ordering on save`, simplified the render casts per the inline comment, and added a services-suite regression test that seeds physical exam diagnoses and verifies the editor renders persisted ordering values. I also updated the PR description with the new coverage/validation notes.

.phpstan/fatal-baseline-caps.php
```diff
@@ -49,7 +49,7 @@
         'return.missing.php' => 0,
         'staticMethod.notFound.php' => 0,
         'trait.notFound.php' => 0,
-        'variable.undefined.php' => 547,
+        'variable.undefined.php' => 546,
```
kojiromike: This cap is causing the `PHP 8.6 - Isolated Tests` job to fail: the test asserts `assertSame($cap, $count)`, and on PHP 8.6 the actual entry count for `variable.undefined.php` is now `545`, not `546`. The diff removed one `$ignoreErrors[]` entry (the physical_exam/edit_diagnoses.php one with `count => 4`), and the test counts entries by `preg_match_all('/\$ignoreErrors\[\] = \[/')` — so the entry count drops by exactly one regardless of how many occurrences that entry covered.

Options:

1. Drop the cap to `545` to match what PHP 8.6 reports. If older PHP versions still report `546`, this PR would need to wait until the corresponding baseline regeneration on those versions also drops the entry.
2. Re-run `composer phpstan-baseline` on PHP 8.6 specifically and use whatever value all required matrix versions agree on.

Whichever path, this needs to land before the PR can merge — `assertSame` will not accept `count < cap`.

interface/forms/physical_exam/edit_diagnoses.php
```diff
@@ -99,17 +111,14 @@
   <td width='95%'><?php echo xlt('Diagnosis'); ?></td>
  </tr>
 
-<?php for ($i = 1; $drow = sqlFetchArray($dres); ++$i) { ?>
- <tr>
```
kojiromike, line 103: Minor: since `$drow['ordering']` is typed as `int|string|null` and is then cast to `(string)`, the `?? $i` fallback fires only when the column is `null`. The persisted schema stores `ordering` as `int NOT NULL DEFAULT 0`, so for any row read from `form_physical_exam_diagnoses` the value should be non-null and the fallback is dead. Worth either dropping the `?? $i` (it can't trigger for existing rows, and blank rows are synthesized with `null` precisely so this fallback fires) or adding a brief comment that the fallback only applies to the synthetic blank rows from `array_pad`. As-is, a reader has to trace `$diagnosisRows` back up to see why a non-null DB column needs a null guard.

tests/Tests/Services/Forms/PhysicalExam/EditDiagnosesTest.php
```diff
@@ -0,0 +1,98 @@
...
+            $output = (string) ob_get_clean();
+        } finally {
+            while (ob_get_level() > $bufferLevel) {
+                ob_end_clean();
+            }
+        }
+
+        $this->assertStringContainsString("name='form_ordering[1]' value='10'", $output);
+        $this->assertStringContainsString("name='form_ordering[2]' value='20'", $output);
+        $this->assertStringContainsString("name='form_ordering[3]' value='30'", $output);
```
kojiromike: This is good render-side coverage, but #12029 is reported as a save-time data-loss bug ("open editor, save without changes, ordering rewritten to 1..N"). The save loop lives at lines 47-65 of `edit_diagnoses.php` and reads `$_POST['form_ordering']` / `$_POST['form_diagnosis']`. A test that POSTs the form values back (the same values the editor just rendered) and re-reads `form_physical_exam_diagnoses` would close the loop the issue actually describes. As written, this test would also pass if the save handler silently dropped the `ordering` column — the render is correct, but the round-trip is what the bug was about.

kojiromike (changes requested):

interface/forms/physical_exam/edit_diagnoses.php
pradhankukiran: Done. Added a brief comment above the fallback clarifying that it is for the synthetic blank rows created by `array_pad`, not for persisted diagnosis rows.

.phpstan/fatal-baseline-caps.php
pradhankukiran: Done. I rebased the PR branch onto current upstream/master so the local baseline matches the merge result, kept variable.undefined.php at 545, and reran `composer phpunit-isolated -- --filter FatalBaselineCapsIsolatedTest` successfully (`OK (20 tests, 20 assertions)`).

tests/Tests/Services/Forms/PhysicalExam/EditDiagnosesTest.php
pradhankukiran: Done. The regression test now renders the editor, extracts the rendered `form_diagnosis` and `form_ordering` values, sends those through the same save helper used by the POST handler, and re-reads the table to verify the persisted 10/20/30 ordering survives the save round trip. The focused services test passes locally (`OK (1 test, 6 assertions)`).

src/Forms/PhysicalExam/DiagnosisHelper.php
```diff
@@ -0,0 +1,45 @@
+<?php
```
kojiromike, line 1: This new file is missing `declare(strict_types=1)`. The repo's CLAUDE.md requires it on every new PHP file ("Every new PHP file starts with `declare(strict_types=1)`"), and the PR's own test file already has it. It matters here specifically: `save()` casts both `$ordering` and `$diagnosis` with `(string)` and writes `$ordering` into an `int` column. Without strict types, the surrounding loose coercion is exactly the silent-data-corruption class the rule guards against.

```suggestion
<?php

declare(strict_types=1);

```

kojiromike (changes requested):

src/Forms/PhysicalExam/DiagnosisHelper.php
pradhankukiran: Fixed in 70e04895f. I added declare(strict_types=1) to the new helper file, placed after the file docblock to satisfy PHPCS.

src/Forms/PhysicalExam/DiagnosisHelper.php
```diff
@@ -0,0 +1,47 @@
...
+            }
+
+            $query = "INSERT INTO form_physical_exam_diagnoses (
+                line_id, ordering, diagnosis
+                ) VALUES (
+                ?, ?, ?
+                )";
+            QueryUtils::sqlStatementThrowException($query, [$lineId, (string) $ordering, (string) $diagnosis]);
+        }
+    }
```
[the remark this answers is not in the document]
pradhankukiran: Fixed in 2e5592da5. Diagnosis saves now run inside QueryUtils::inTransaction(), so the delete and inserts commit or roll back together.

interface/forms/physical_exam/diagnosis_helpers.php
```diff
@@ -0,0 +1,47 @@
...
+
+        foreach ($formDiagnoses as $i => $diagnosis) {
+            if (!is_scalar($diagnosis) || (string) $diagnosis === '') {
+                continue;
+            }
+
+            $ordering = $formOrderings[$i] ?? '';
+            if (!is_scalar($ordering)) {
+                $ordering = '';
+            }
```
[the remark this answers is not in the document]
pradhankukiran: Fixed in fc8f02667. Ordering values are now normalized to integers before insert; valid integer values are preserved and malformed/missing values fall back to the row key, then 0 only if needed. Added a regression test for malformed ordering input.

src/Forms/PhysicalExam/DiagnosisHelper.php
```diff
@@ -0,0 +1,64 @@
...
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    OpenEMR Contributors
+ * @copyright Copyright (c) 2026 OpenEMR Contributors
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Forms\PhysicalExam;
```
kojiromike, line 15: This class declares `OpenEMR\Forms\PhysicalExam`, but the repo's Composer PSR-4 mapping only covers `"OpenEMR\\": "src"` — so the class is not autoloadable, and every consumer has to know to `require_once` this file first (the page and the test both do). A future caller that just writes `use OpenEMR\Forms\PhysicalExam\DiagnosisHelper;` and calls it will get a fatal "class not found" even though the namespace looks autoloadable. The repo's CLAUDE.md asks for new classes to go in `/src/` for exactly this reason.

Suggest moving this to `src/Forms/PhysicalExam/DiagnosisHelper.php` (same namespace, filename matching the class). Then the `require_once` lines in both `edit_diagnoses.php` and the test can be dropped and the autoloader does the rest. (I'm aware `interface/forms/newpatient/C_EncounterVisitForm.class.php` does the same thing today — that's a legacy exception rather than a pattern to extend.)

interface/forms/physical_exam/edit_diagnoses.php
```diff
@@ -49,22 +51,13 @@
...
-            ) VALUES (
-            ?, ?, ?
-            )";
-            sqlStatement($query, [$line_id, $ordering, $diagnosis]);
-        }
-    }
+    $form_diagnoses = $_POST['form_diagnosis'] ?? [];
+    $form_orderings = $_POST['form_ordering'] ?? [];
+    DiagnosisHelper::save(
+        is_scalar($line_id) ? (string) $line_id : '',
```
kojiromike: When `$line_id` is not a scalar (e.g. a crafted `lineid[]=x` request), this silently proceeds with `''` — so the save runs `DELETE ... WHERE line_id = ''` and can insert rows keyed on an empty `line_id`, quietly accepting a malformed request. Would it make sense to fail fast instead — bail out with an error when `lineid` is missing or non-scalar, before the save — rather than coercing to a default? The `is_array(...) ? ... : []` narrowing on the two POST arrays is fine (an empty array is a sensible degenerate case), but an empty-string line id is never a valid target row.

kojiromike (changes requested):
Thanks for working through all the earlier rounds — the fix, the transactional save, the ordering normalization, and the render/save round-trip test all look solid at this point. I tried to poke holes in `normalizeOrdering()` (negative values, whitespace-padded numeric strings, the 0-collapse concern from the earlier review) and in the render refactor (blank-row numbering, the `"0"`-diagnosis truthiness change — which this PR actually improves, since a literal `"0"` diagnosis is now kept instead of silently dropped), and couldn't find a behavioral regression.

Two remaining items, both on the new code.

Note on the failing `codecov/project` check: the -2.20% project drop comes from carried-forward flags that didn't upload on this run (several flags report `?`), not from this PR — patch coverage is 94%. Nothing for you to action there.

src/Forms/PhysicalExam/DiagnosisHelper.php
pradhankukiran: Fixed in 01085e23b. DiagnosisHelper now lives at src/Forms/PhysicalExam/DiagnosisHelper.php under the autoloaded OpenEMR\\ namespace, and the manual require_once calls were removed from the page and test.

interface/forms/physical_exam/edit_diagnoses.php
pradhankukiran: Fixed in 3b18e6013. The editor now normalizes lineid once up front and returns HTTP 400 for missing, non-scalar, or blank values before any save can run. Added test coverage for malformed line id handling.

src/Forms/PhysicalExam/DiagnosisHelper.php
```diff
@@ -0,0 +1,78 @@
...
+        if (!is_scalar($lineId)) {
+            return null;
+        }
+
+        $lineId = (string) $lineId;
+        if (trim($lineId) === '') {
+            return null;
+        }
+
+        return $lineId;
```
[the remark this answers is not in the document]
pradhankukiran: Fixed in 83d2a3b06. normalizeLineId() now accepts only strings and integers, trims the returned ID, and rejects booleans/floats/blank values before any save path can use them. Added coverage for the trimmed and rejected cases.

kojiromike (approved):
All feedback from the previous round is addressed at HEAD (0ecbc6a):

- **Helper is now autoloadable.** `DiagnosisHelper` moved to `src/Forms/PhysicalExam/DiagnosisHelper.php` under the PSR-4 `OpenEMR\` root, and both manual `require_once` calls (page and test) are gone. No stale references to the old `interface/forms/physical_exam/diagnosis_helpers.php` path remain.
- **Malformed `lineid` now fails fast.** `normalizeLineId()` accepts only `int|string`, trims, rejects blank, and the page returns HTTP 400 before the ACL check or any save path — no more silent `DELETE ... WHERE line_id = ''`. The tightening round (83d2a3b) also covers CodeRabbit's follow-up: booleans and floats are rejected, and the returned ID is trimmed so render and save target the same key. `testNormalizeLineIdRejectsMalformedValues` locks all of this in.
- **Baseline sync is correct.** The two `attr_url`/`text` "mixed given" entries for `edit_diagnoses.php` drop out because `$line_id` is now a guaranteed `string` — a genuine drain, not a suppression. The isolated baseline-caps tests pass across the PHP matrix.

The only failing check is `codecov/project`, which is carryforward-flag noise (several flags uploaded `?` this run) — patch coverage is ~94% and all tests pass.

Thanks for the thorough follow-through across all the rounds — this is ready.

PR #12140, opened 2026-05-14
fix(globals): Restore windows webserver_root normalization.

### LINKED ISSUE #12134: bug(globals): Windows XAMPP asset paths break because prior $webserver_root slash normalization was removed from globals.php
### Describe the bug
On Windows/XAMPP, $server_document_root is normalized from backslashes to forward slashes, but $webserver_root is no longer normalized the same way. This causes the $web_root calculation to compare paths with different slash styles and generate invalid web asset paths.

For example, instead of loading:
https://somedomain.com/public/themes/style_dark.css?v=82
OpenEMR generates:
https://xampp/htdocs/openemr/public/themes/style_dark.css?v=82
which fails because xampp is interpreted as a hostname.

This was previously handled by normalizing $webserver_root on Windows, and the fix should simply restore that missing code in globals.php.

The PR will add back the prior Windows-specific normalization:

if (IS_WINDOWS) {
    $webserver_root = str_replace("\\", "/", $webserver_root);
}
This is a simple restoration of previously existing logic and resolves broken asset URL generation under Windows/XAMPP virtual-host setups.

### To Reproduce
Build and run on Windows

1. Add a '....'

### Client configuration

- Browser:
- OpenEMR version: dev-master
- Operating system: windows 11 xampp

sjpadgett:
Fixes #12134 

add back previously removed IF_WINDOWS conditional to normalize path slashes in globals.php.

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -44,6 +44,9 @@ if (!defined('IS_WINDOWS')) {
 // If not working, can set manually below.
 // Auto collect the full absolute directory path for openemr.
 $webserver_root = dirname(__FILE__, 2);
+if (IS_WINDOWS) {
+    $webserver_root = str_replace("\\", "/", $webserver_root);
+}
 
 /**
  * Allow a `.env` file to be read in and applied as $_SERVER variables.


sjpadgett:
Best I can tell this issue is new and has not been released in any recent 8.0.0 patches

stephenwaite (approved):
thanks @sjpadgett !

sjpadgett:
fyi, I have no plans to try and fix the CodeCov report because it has no meaning to this PR. May have meaning otherwise but just letting you know if waiting to merge because of error. A pain patching this every time I pull master.
Also, while I'm here I thought I'd give a heads up that there're at least two long time sponsors of openemr that use and paid for the Voice Client in the faxsms module.

kojiromike:
We need to backport this to rel-810

PR #12144, opened 2026-05-15
fix(portal): correct deleteItem target and readMessage issue

### LINKED ISSUE #11644: bug(portal): deleteItem index mismatch and readMessage folder switch in messaging
## `deleteItem()` index mismatch between `state.items` and `state.allItems`

In `portal/messaging/messages.php`, `deleteItem(idx)` receives an index from `state.items.indexOf(state.selected)` but uses it to look up `state.allItems[idx]`. Since `state.items` can be `inboxItems`, `sentItems`, etc. (filtered subsets with different ordering), the index doesn't correspond to the same message in `state.allItems`. This can delete the wrong message.

## `readMessage()` always switches to "All" folder

`readMessage()` unconditionally sets `state.isAll = true` and `state.items = state.allItems`, regardless of which folder the user was viewing. Clicking a message in Inbox or Sent switches the user's context to the All folder.

## Context

Both bugs are pre-existing — they were carried over from the original AngularJS implementation into the vanilla JS rewrite in #11521. They are not regressions introduced by that PR.

techaround:
**Short description of what this changes or resolves:**
Fixes [#11644](https://github.com/openemr/openemr/issues/11644)

**Changes proposed in this pull request:**
deleteItem operated on the wrong array.
The handler received idx = items.indexOf(selected) (an index into $scope.items — which can be inboxItems, sentItems, etc.) but dereferenced it as $scope.allItems[idx]. Each folder array is populated by its own XHR, so the per-folder object references are distinct; $scope.items.indexOf(itemToDelete) then returned -1, and splice(-1, 1) silently removed the last row in the list. In the worst case the wrong mail_chain was POSTed to handle_note.php?task=delete. Now passes the item directly (deleteItem(selected)) and removes the deleted thread from every local folder array.

readMessage unconditionally switched the user to the All folder.
Every click on a message in Inbox / Sent / Archive ran $scope.isAll = true; $scope.items = $scope.allItems, silently flipping the user out of their chosen folder. Removed the folder switch; readMessage now only marks the message read and sets $scope.selected. (Note: this bug was masking bug 1 in the common click-then-delete flow — by the time the user clicked delete, items === allItems and the indexing accident produced the correct result. Once the folder switch is removed, the deleteItem fix becomes load-bearing.)

Post-delete UI updates without a page refresh.

**Was an AI assistant used? No**


portal/messaging/messages.php
```diff
@@ -246,14 +246,26 @@ function getAuthPortalUsers()
...
-                    const itemToDelete = $scope.allItems[idx];
-                    const idxInItems = $scope.items.indexOf(itemToDelete);
-                    $scope.deleteMessage(itemToDelete.mail_chain); // Just this user's message
-                    $scope.items.splice(idxInItems, 1);
-                    $scope.search();
-                    $scope.init()
+                    const removeFromList = function (list) {
+                        if (!Array.isArray(list)) return;
+                        for (let i = list.length - 1; i >= 0; i--) {
+                            if (list[i] && list[i].mail_chain === item.mail_chain) {
```
kojiromike: Two follow-ups on this comparison:

**1. Strict equality risks a silent no-op on type drift.** `mail_chain` values come from `handle_note.php` responses; PHP/MySQL round-trips frequently surface the same column as a string from one endpoint and an integer from another (mysqli's default vs. a hand-cast). If any of the four endpoints that populate these arrays ever returns `mail_chain` as a different scalar type, `===` returns false and the loop quietly removes nothing. The old index-based code couldn't hit this. Worth normalizing on one side, e.g.:

```suggestion
                            if (list[i] && String(list[i].mail_chain) === String(item.mail_chain)) {
```

**2. Sweeping by `mail_chain` archives the whole thread.** Every row sharing `mail_chain` is spliced from all four arrays. The old code removed a single row by index. If `allItems` (or any folder) ever renders per-message rather than per-thread, this silently changes delete semantics from "this row" to "the entire chain." Worth pinning the assumption with a brief comment and/or confirming server-side semantics for `task=delete` with `noteid=<mail_chain>` (whole-thread archive vs. user's side only).

portal/messaging/messages.php
```diff
@@ -246,14 +246,26 @@ function getAuthPortalUsers()
...
+                                list.splice(i, 1);
+                            }
+                        }
+                    };
+                    $scope.deleteMessage(item.mail_chain).then(function () {
+                        removeFromList($scope.items);
+                        removeFromList($scope.inboxItems);
+                        removeFromList($scope.sentItems);
+                        removeFromList($scope.allItems);
+                        $scope.selected = null;
```
kojiromike: Small race window introduced here: if the user clicks delete on message A and then opens message B (which sets `$scope.selected = item`) before A's POST returns, A's `.then` clobbers `$scope.selected` back to `null` and B's reading pane disappears.

A one-line guard is enough — only null `selected` when it's still the deleted message:

```suggestion
                        if ($scope.selected && $scope.selected.mail_chain === item.mail_chain) {
                            $scope.selected = null;
                        }
```

portal/messaging/messages.php
```diff
@@ -274,12 +286,11 @@ function getAuthPortalUsers()
                 };
 
                 $scope.deleteMessage = function (id) {
-                    $http.post('handle_note.php', $.param({'task': 'delete', 'noteid': id, 'csrf_token_form': $scope.csrf})).then(function successCallback(response) {
+                    return $http.post('handle_note.php', $.param({'task': 'delete', 'noteid': id, 'csrf_token_form': $scope.csrf})).then(function successCallback(response) {
                         return true;
                     }, function errorCallback(response) {
                         alert(response.data);
```
kojiromike, line 294: The `errorCallback` handles the failed POST with `alert()` but doesn't return a rejected promise. The outer promise (returned by `deleteMessage`) therefore resolves successfully even when the server rejected, so the `.then(...)` chain in `deleteItem` (lines 260–268) runs and removes the thread from `items`/`inboxItems`/`sentItems`/`allItems` regardless. UI shows the message archived; server still has it.

Returning the rejection (or restructuring as `.then(success).catch(error)`) lets the failure propagate so `deleteItem` only updates local state when the delete actually succeeded:

```suggestion
                        alert(response.data);
                        return $q.reject(response);
```

(Requires `$q` in the controller's DI list if it isn't already injected. Alternatively, drop the inline `errorCallback` entirely and add a `.catch(function (response) { alert(response.data); })` after the `.then` in `deleteItem`.)

portal/messaging/messages.php
```diff
@@ -634,19 +652,19 @@ function getAuthPortalUsers()
...
                                         <span class="col-sm-1" style="max-width: 5px;"><input type="checkbox" checklist-model="item.deleted" value={{item.deleted}}></span>
 
-                                        <span class="col-sm-1 px-1"  ng-click="readMessage($index)" ><span ng-class="{strong: !item.read}">{{item.message_status}}</span></span>
-                                        <span class="col-sm-2 px-1"  ng-click="readMessage($index)" ><span ng-class="{strong: !item.read}">{{item.date | date:'yyyy-MM-dd hh:mm'}}</span></span>
-                                        <span class="col-sm-3 px-1"  ng-click="readMessage($index)" >
-                                            <a ng-click="readMessage($index)" class="btn-link">
+                                        <span class="col-sm-1 px-1"  ng-click="readMessage(item)" ><span ng-class="{strong: !item.read}">{{item.message_status}}</span></span>
+                                        <span class="col-sm-2 px-1"  ng-click="readMessage(item)" ><span ng-class="{strong: !item.read}">{{item.date | date:'yyyy-MM-dd hh:mm'}}</span></span>
+                                        <span class="col-sm-3 px-1">
```
kojiromike: Minor UX regression in the listing template: this `<span class="col-sm-3 px-1">` (and the `<span class="col-sm-1">` a few lines down at new line 662) had `ng-click="readMessage($index)"` on the wrapper in master — the whole column was clickable. After this PR, only the inner `<a ng-click="readMessage(item)">` fires. Clicking the surrounding column area (e.g., padding around the link, especially on touch) no longer marks-read / selects the message.

If removing the outer `ng-click` was intentional (e.g., to avoid the bubbling Copilot flagged), restoring the wrapper handler and adding `$event.stopPropagation()` on the inner `<a>` would preserve the larger click target while killing the duplicate fire. If unintentional, just put the wrapper `ng-click` back.

kojiromike (commented):
Thanks for picking this up — both bugs are real and the structural shape of the fix (passing the object instead of an index, fanning state changes across the per-folder arrays) is the right direction.

A few inline notes. The main one is the rejection-swallowing in `deleteMessage`; the other two are smaller correctness/UX nits introduced by the rework.

Confirmed during review:
- The new client-side per-thread removal is consistent with the server: `updatePortalMailMessageStatus()` already soft-deletes every row matching `mail_chain OR id` for the owner, so removing every local row with that `mail_chain` matches what the server has done. The old client only spliced one row, which was inconsistent with server behavior — this PR closes that gap.
- Reverse-iteration `splice` in `removeFromList` is correct (won't skip elements).
- Marking read across all folder arrays is necessary — each folder is its own XHR result with distinct object refs.
- Nested-`ng-click` bubbling in listing rows (raised by Copilot) is unchanged by this PR — pre-existing, separate concern.

kojiromike (approved):
All three of my prior comments addressed. Re-reviewed at HEAD — no remaining concerns.

techaround:
@kojiromike do you want me to add test coverage for this change? Looks like it would add a lot of code to do properly.

kojiromike:
I do not think you can reasonably add tests for the portal today. Worse, I don't think adding tests is reasonable without first replacing Angular with something modern and supported. So our only options today are manual testing and writing e2e tests.

PR #11347, opened 2026-03-30
fix(security): replace addslashes() with js_escape() in JS contexts

haoyu-haoyu:
## Summary

Replace all `addslashes()` calls used for JavaScript string escaping with `js_escape()` across 22 files (~42 call sites).

`addslashes()` only escapes quotes, backslashes, and NUL bytes — it is **not sufficient for JavaScript escaping**. Characters like `</script>`, line terminators (U+2028, U+2029), and other special sequences are not handled. `js_escape()` uses `json_encode()` which correctly handles all JavaScript-sensitive characters per RFC 8259.

Since `js_escape()` returns a JSON-encoded value that includes its own quotes, all manually added surrounding quotes were removed to prevent double-quoting.

### Categories addressed

| Category | Count | Pattern |
|----------|-------|---------|
| JS `alert()`/`confirm()` calls | ~30 | `alert('" . addslashes($x) . "')` → `alert(" . js_escape($x) . ")` |
| `attr(addslashes())` in onclick handlers | ~12 | `attr(addslashes($x))` → `attr(js_escape($x))` |

### Categories NOT addressed (separate PRs per #11259)

- SQL/log string contexts (`gacl/admin/`, `payment.inc.php`, `eye_mag/save.php`)
- CSV contexts (`BillingExport.csv.php`)
- Smarty compiler internals and test helpers (leave alone per issue)

### Files changed (22)

`ccr/createCCR.php`, `custom/import_xml.php`, `interface/billing/sl_eob_search.php`, `interface/drugs/add_edit_drug.php`, `interface/drugs/add_edit_lot.php`, `interface/drugs/destroy_lot.php`, `interface/drugs/drug_inventory.php`, `interface/fax/fax_dispatch.php`, `interface/fax/faxq.php`, `interface/forms/CAMOS/new.php`, `interface/forms/LBF/new.php`, `interface/forms/eye_mag/SpectacleRx.php`, `interface/forms/eye_mag/a_issue.php`, `interface/main/calendar/add_edit_event.php`, `interface/main/finder/multi_patients_finder.php`, `interface/main/messages/messages.php`, `interface/patient_file/pos_checkout_ippf.php`, `interface/reports/destroyed_drugs_report.php`, `library/custom_template/add_context.php`, `library/custom_template/add_template.php`, `library/custom_template/delete_category.php`, `library/dated_reminder_functions.php`

## Test plan

- [ ] Verify `js_escape()` output includes quotes (e.g., `js_escape("hello")` → `"hello"`)
[...]

PR #12145, opened 2026-05-15
chore: clean up js_escape() patterns

### LINKED ISSUE #11613: chore: clean up js_escape() patterns after #11347
After merging #11347 (addslashes → js_escape), several patterns remain that can be cleaned up:

1. **`js_escape(xl(...))` → `xlj(...)`** — `xlj()` already exists in `library/htmlspecialchars.inc.php` and does exactly `js_escape(xl($key))`. All `js_escape(xl('...'))` call sites should use `xlj()` instead.

2. **Long inline `<?php echo js_escape(...) ?>` lines** — Many of the converted lines are deeply embedded in long HTML strings. These should be extracted to variables for readability.

3. **`js_escape('literal-string')`** — `json_encode` on a plain ASCII literal just adds quotes. These calls can be replaced with the quoted string directly (e.g., `js_escape('foo')` → `"foo"`).

Kasirocswell:
**Summary**
This PR addresses issue #11613 by cleaning up `js_escape()` patterns that remained after #11347 was merged.

**Changes Made**

1. Replace `js_escape(xl(...))` with `xlj(...)`
Replaced 15 instances across 6 files where `js_escape(xl('...'))` was called without variable concatenation. The `xlj()` function already exists in [library/htmlspecialchars.inc.php] and does exactly `js_escape(xl($key))`.

2. Extract long inline `js_escape(...)` to variables
Extracted 7 deeply embedded `js_escape()` calls to `$msg` variables for improved readability.

3. Replace `js_escape('literal-string')` with quoted strings
Replaced 16 instances where `js_escape()` was called on plain ASCII literals. Since `json_encode` on a plain literal just adds quotes, these can be replaced directly.

**Note**
The remaining `js_escape(xl(...) . $variable)` patterns in `sphere/` files concatenate translated strings with runtime variables and cannot use `xlj()` since it only accepts a single translation key. These were correctly left as `js_escape()` but extracted to variables per item 2.

**Testing**
- All affected pages should display alerts/confirms correctly
- No behavioral changes expected
- CI checks should pass

Fixes #11613

cc @kojiromike @msummers42

### DIFF
--- a/interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
+++ b/interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
@@ -102,7 +102,7 @@ if ($zip->open($storeLocation) === true) {
         EventAuditLogger::getInstance()->newEvent("pharmacy_log", $session->get('authUser'), $session->get('authProvider'), 0, ($isError['messageText']));
         $wenoLog->insertWenoLog("Pharmacy Directory", "Failed");
         // no need to continue so send error to UI alert and die.
-        die('"Pharmacy download failed."');
+        die(js_escape('Pharmacy download failed.'));
     }
     // process the csv file
     // Number of rows imported or false if error
@@ -145,7 +145,7 @@ if ($zip->open($storeLocation) === true) {
     error_log('Pharmacy download zip open failed.');
     $wenoLog->insertWenoLog("Pharmacy Directory", "Pharmacy download zip open failed.");
     // no need to continue so send error to UI alert and die.
-    die('"Pharmacy download zip open failed."');
+    die(js_escape('Pharmacy download zip open failed.'));
 }
 
 function download_zipfile($fileUrl, $zipped_file): void
--- a/interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
+++ b/interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
@@ -102,7 +102,7 @@ if ($zip->open($storeLocation) === true) {
         EventAuditLogger::getInstance()->newEvent("pharmacy_log", $session->get('authUser'), $session->get('authProvider'), 0, ($isError['messageText']));
         $wenoLog->insertWenoLog("Pharmacy Directory", "Failed");
         // no need to continue so send error to UI alert and die.
-        die(js_escape('Pharmacy download failed.'));
+        die('"Pharmacy download failed."');
     }
     // process the csv file
     // Number of rows imported or false if error
@@ -145,7 +145,7 @@ if ($zip->open($storeLocation) === true) {
     error_log('Pharmacy download zip open failed.');
     $wenoLog->insertWenoLog("Pharmacy Directory", "Pharmacy download zip open failed.");
     // no need to continue so send error to UI alert and die.
-    die(js_escape('Pharmacy download zip open failed.'));
+    die('"Pharmacy download zip open failed."');
 }
 
 function download_zipfile($fileUrl, $zipped_file): void
--- a/library/custom_template/add_context.php
+++ b/library/custom_template/add_context.php
@@ -98,7 +98,7 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                     document.designation_managment.submit();
                 }
                 else{
-                    alert(<?php echo js_escape(xl("Context name can't be empty"));?>);
+                    alert(<?php echo xlj("Context name can't be empty");?>);
                 }
             }
             function deleteme(id){
@@ -107,10 +107,10 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                 CheckContextLive(id);
                 stat = document.getElementById('stat').value;
                 if(stat==1){
-                    msg = <?php echo js_escape(xl('This context contains categories, which will be deleted. Do you still want to continue?'));?>;
+                    msg = <?php echo xlj('This context contains categories, which will be deleted. Do you still want to continue?');?>;
                 }
                 else{
-                    msg = <?php echo js_escape(xl('Do you want to delete this?'));?>;
+                    msg = <?php echo xlj('Do you want to delete this?');?>;
                 }
                 if(confirm(msg)){
                 document.getElementById('action').value='delete';
@@ -133,7 +133,7 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                 document.designation_managment.submit();
                 }
                 else{
-                   alert(<?php echo js_escape(xl("Context name can't be empty"));?>);
+                   alert(<?php echo xlj("Context name can't be empty");?>);
                 }
             }
             function CheckContextLive(id){
--- a/library/custom_template/add_context.php
+++ b/library/custom_template/add_context.php
@@ -98,7 +98,7 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                     document.designation_managment.submit();
                 }
                 else{
-                    alert("<?php echo xl("Context name can't be empty"); ?>");
+                    alert(<?php echo xlj("Context name can't be empty"); ?>);
                 }
             }
             function deleteme(id){
@@ -107,10 +107,10 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                 CheckContextLive(id);
                 stat = document.getElementById('stat').value;
                 if(stat==1){
-                    msg = "<?php echo xl('This context contains categories, which will be deleted. Do you still want to continue?'); ?>";
+                    msg = <?php echo xlj('This context contains categories, which will be deleted. Do you still want to continue?'); ?>;
                 }
                 else{
-                    msg = "<?php echo xl('Do you want to delete this?'); ?>";
+                    msg = <?php echo xlj('Do you want to delete this?'); ?>;
                 }
                 if(confirm(msg)){
                 document.getElementById('action').value='delete';
@@ -133,7 +133,7 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                 document.designation_managment.submit();
                 }
                 else{
-                   alert("<?php echo xl("Context name can't be empty"); ?>");
+                   alert(<?php echo xlj("Context name can't be empty"); ?>);
                 }
             }
             function CheckContextLive(id){
--- a/library/custom_template/add_context.php
+++ b/library/custom_template/add_context.php
@@ -98,7 +98,7 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                     document.designation_managment.submit();
                 }
                 else{
-                    alert(<?php echo xlj("Context name can't be empty");?>);
+                    alert("<?php echo xl("Context name can't be empty"); ?>");
                 }
             }
             function deleteme(id){
@@ -107,10 +107,10 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                 CheckContextLive(id);
                 stat = document.getElementById('stat').value;
                 if(stat==1){
-                    msg = <?php echo xlj('This context contains categories, which will be deleted. Do you still want to continue?');?>;
+                    msg = "<?php echo xl('This context contains categories, which will be deleted. Do you still want to continue?'); ?>";
                 }
                 else{
-                    msg = <?php echo xlj('Do you want to delete this?');?>;
+                    msg = "<?php echo xl('Do you want to delete this?'); ?>";
                 }
                 if(confirm(msg)){
                 document.getElementById('action').value='delete';
@@ -133,7 +133,7 @@ if (trim($_POST['contextname'] ?? '') != '' && $_POST['action'] == 'add') {
                 document.designation_managment.submit();
                 }
                 else{
-                   alert(<?php echo xlj("Context name can't be empty");?>);
+                   alert("<?php echo xl("Context name can't be empty"); ?>");
                 }
             }
             function CheckContextLive(id){
--- a/library/custom_template/add_template.php
+++ b/library/custom_template/add_template.php
@@ -65,12 +65,12 @@ $list_id = $_REQUEST['list_id'];
                 async: false,
                 success: function(thedata){
                         if(thedata=="Fail"){
-                            alert(document.getElementById('template_name').value + " " + "<?php echo xl('already exists'); ?>");
+                            alert(document.getElementById('template_name').value + " " + <?php echo xlj('already exists'); ?>);
                             return false;
                         }
                         else{
                             mainform.getElementById('templateDD').innerHTML = thedata;
-                            alert("<?php echo xl('Successfully added category'); ?>" + " " + document.getElementById('template_name').value);
+                            alert(<?php echo xlj('Successfully added category'); ?> + " " + document.getElementById('template_name').value);
                             //window.opener.opener.location.reload();
                             dlgclose();
                         }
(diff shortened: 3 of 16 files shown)


interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
```diff
@@ -102,7 +102,7 @@
         EventAuditLogger::getInstance()->newEvent("pharmacy_log", $session->get('authUser'), $session->get('authProvider'), 0, ($isError['messageText']));
         $wenoLog->insertWenoLog("Pharmacy Directory", "Failed");
         // no need to continue so send error to UI alert and die.
-        die(js_escape('Pharmacy download failed.'));
+        die('"Pharmacy download failed."');
```
kojiromike: No need to fix this here, as it's out of scope, but `die()` is already forbidden in our standards, and these in particular seem likely to result in a `200 OK` when an error happens. If there isn't a followup ticket already, please create one.

kojiromike: Also, it's not clear to me why the double-quoting or js_escape is even necessary in this context. Is this going into a js context?

library/custom_template/add_context.php
```diff
@@ -98,7 +98,7 @@ function checkSubmit(){
                     document.designation_managment.submit();
                 }
                 else{
-                    alert(<?php echo js_escape(xl("Context name can't be empty"));?>);
+                    alert(<?php echo xlj("Context name can't be empty");?>);
```
kojiromike: js escaping isn't necessary for a literal string like this. You just need the nested quotes.

library/custom_template/add_context.php
```diff
@@ -107,10 +107,10 @@ function deleteme(id){
                 CheckContextLive(id);
                 stat = document.getElementById('stat').value;
                 if(stat==1){
-                    msg = <?php echo js_escape(xl('This context contains categories, which will be deleted. Do you still want to continue?'));?>;
+                    msg = <?php echo xlj('This context contains categories, which will be deleted. Do you still want to continue?');?>;
```
kojiromike: same as above, s/xlj/xl and add quotes.

library/custom_template/add_context.php
```diff
@@ -133,7 +133,7 @@ function checkUpdate(){
                 document.designation_managment.submit();
                 }
                 else{
-                   alert(<?php echo js_escape(xl("Context name can't be empty"));?>);
+                   alert(<?php echo xlj("Context name can't be empty");?>);
```
kojiromike: same as above

library/custom_template/add_template.php
```diff
@@ -65,12 +65,12 @@ function add_template(){
                 async: false,
                 success: function(thedata){
                         if(thedata=="Fail"){
-                            alert(document.getElementById('template_name').value + " " + <?php echo js_escape(xl('already exists'));?>);
+                            alert(document.getElementById('template_name').value + " " + <?php echo xlj('already exists');?>);
```
kojiromike: xl

library/custom_template/add_template.php
```diff
@@ -65,12 +65,12 @@ function add_template(){
...
                             return false;
                         }
                         else{
                             mainform.getElementById('templateDD').innerHTML = thedata;
-                            alert(<?php echo js_escape(xl('Successfully added category'));?> + " " + document.getElementById('template_name').value);
+                            alert(<?php echo xlj('Successfully added category');?> + " " + document.getElementById('template_name').value);
```
kojiromike: xl

kojiromike (changes requested):
You caught some cases where a literal string didn't need js_escape, but just quotes. But there are many others.

interface/modules/custom_modules/oe-module-weno/scripts/file_download.php
Kasirocswell: it isn't a JS context. The caller (templates/download_log_viewer.php) fetches this via $.ajax and assigns the response with .textContent, so js_escape was never needed. My '"..."' change actually leaves the literal quotes visible in the error message.

I'll revert this hunk and fold the proper fix (die('...') → plain text, plus the die() / 200-OK issue) into the follow-up ticket.

library/custom_template/delete_category.php
```diff
@@ -74,7 +74,7 @@ function delete_category(id){
                 },
                 success: function(thedata){
                             if(thedata){
-                                alert(<?php echo js_escape('There are currently other users of the category you are trying to delete. Please contact them and ask them to delete it. Categories may not be deleted while in use. This Categories are currently used by \n');?> + thedata);
+                                alert("<?php echo xl('There are currently other users of the category you are trying to delete. Please contact them and ask them to delete it. Categories may not be deleted while in use. This Categories are currently used by'); ?>\n" + thedata);
```
kojiromike: Same translation-injection issue here, plus the `\n` belongs outside the JSON-encoded translation. `xlj()` emits its own surrounding quotes, so concatenate the newline as a separate JS string.

```suggestion
                                alert(<?php echo xlj('There are currently other users of the category you are trying to delete. Please contact them and ask them to delete it. Categories may not be deleted while in use. This Categories are currently used by'); ?> + "\n" + thedata);
```

library/custom_template/add_context.php
```diff
@@ -98,7 +98,7 @@ function checkSubmit(){
                     document.designation_managment.submit();
                 }
                 else{
-                    alert(<?php echo js_escape(xl("Context name can't be empty"));?>);
+                    alert("<?php echo xl("Context name can't be empty"); ?>");
```
kojiromike: Revert to `xlj()` — apologies for the earlier misdirection. `xl()` is translator-controlled, so wrapping its output in literal JS quotes would let a translation containing `"`, `\`, or a newline break out of the alert.

```suggestion
                    alert(<?php echo xlj("Context name can't be empty"); ?>);
```

library/custom_template/add_context.php
```diff
@@ -107,10 +107,10 @@ function deleteme(id){
                 CheckContextLive(id);
                 stat = document.getElementById('stat').value;
                 if(stat==1){
-                    msg = <?php echo js_escape(xl('This context contains categories, which will be deleted. Do you still want to continue?'));?>;
+                    msg = "<?php echo xl('This context contains categories, which will be deleted. Do you still want to continue?'); ?>";
```
kojiromike: Same — please revert to `xlj()`.

```suggestion
                    msg = <?php echo xlj('This context contains categories, which will be deleted. Do you still want to continue?'); ?>;
```

library/custom_template/add_context.php
```diff
@@ -107,10 +107,10 @@ function deleteme(id){
...
                 }
                 else{
-                    msg = <?php echo js_escape(xl('Do you want to delete this?'));?>;
+                    msg = "<?php echo xl('Do you want to delete this?'); ?>";
```
kojiromike: Same — please revert to `xlj()`.

```suggestion
                    msg = <?php echo xlj('Do you want to delete this?'); ?>;
```

library/custom_template/add_context.php
```diff
@@ -133,7 +133,7 @@ function checkUpdate(){
                 document.designation_managment.submit();
                 }
                 else{
-                   alert(<?php echo js_escape(xl("Context name can't be empty"));?>);
+                   alert("<?php echo xl("Context name can't be empty"); ?>");
```
kojiromike: Same — please revert to `xlj()`.

```suggestion
                   alert(<?php echo xlj("Context name can't be empty"); ?>);
```

library/custom_template/add_template.php
```diff
@@ -65,12 +65,12 @@ function add_template(){
                 async: false,
                 success: function(thedata){
                         if(thedata=="Fail"){
-                            alert(document.getElementById('template_name').value + " " + <?php echo js_escape(xl('already exists'));?>);
+                            alert(document.getElementById('template_name').value + " " + "<?php echo xl('already exists'); ?>");
```
kojiromike: Revert to `xlj()` — sorry, my earlier `s/xlj/xl/` advice was wrong for `xl()` output. A translation containing `"`, `\`, or a newline would break out of the alert string; `xlj()` JSON-encodes the result so that can't happen.

```suggestion
                            alert(document.getElementById('template_name').value + " " + <?php echo xlj('already exists'); ?>);
```

library/custom_template/add_template.php
```diff
@@ -65,12 +65,12 @@ function add_template(){
...
                             return false;
                         }
                         else{
                             mainform.getElementById('templateDD').innerHTML = thedata;
-                            alert(<?php echo js_escape(xl('Successfully added category'));?> + " " + document.getElementById('template_name').value);
+                            alert("<?php echo xl('Successfully added category'); ?>" + " " + document.getElementById('template_name').value);
```
kojiromike: Same — please revert to `xlj()`.

```suggestion
                            alert(<?php echo xlj('Successfully added category'); ?> + " " + document.getElementById('template_name').value);
```

library/custom_template/add_template.php
```diff
@@ -81,11 +81,11 @@ function add_template(){
                 });
                 }
                 else{
-                    alert(<?php echo js_escape(xl('You should select at least one context'));?>);
+                    alert("<?php echo xl('You should select at least one context'); ?>");
```
kojiromike: Same — please revert to `xlj()`.

```suggestion
                    alert(<?php echo xlj('You should select at least one context'); ?>);
```

library/custom_template/add_template.php
```diff
@@ -81,11 +81,11 @@ function add_template(){
...
                 }
             }
             else{
-                alert(<?php echo js_escape(xl('Category name is empty'));?>);
+                alert("<?php echo xl('Category name is empty'); ?>");
```
kojiromike: Same — please revert to `xlj()`.

```suggestion
                alert(<?php echo xlj('Category name is empty'); ?>);
```

library/custom_template/delete_category.php
```diff
@@ -53,7 +53,7 @@ function delete_full_category(id){
                      source: "delete_full_category"
                 },
                 success: function(thedata){
-                            alert(<?php echo js_escape(xl('Deleted Successfully.'));?>);
+                            alert("<?php echo xl('Deleted Successfully.'); ?>");
```
kojiromike: Revert to `xlj()` — `xl()` is translator-controlled, so the literal-quotes pattern is unsafe here even though it was fine for the truly literal cases elsewhere in the PR.

```suggestion
                            alert(<?php echo xlj('Deleted Successfully.'); ?>);
```

library/custom_template/delete_category.php
```diff
@@ -63,7 +63,7 @@ function delete_full_category(id){
         }
         function delete_category(id){
             top.restoreSession();
-            if(confirm(<?php echo js_escape(xl('Do you want to delete?'));?>)){
+            if(confirm("<?php echo xl('Do you want to delete?'); ?>")){
```
kojiromike: Same — please revert to `xlj()`.

```suggestion
            if(confirm(<?php echo xlj('Do you want to delete?'); ?>)){
```

library/custom_template/delete_category.php
kojiromike: Same translation-injection concern, plus the `\n` belongs outside the JSON-encoded translation. `xlj()` emits its own surrounding quotes, so concatenate the newline as a separate JS string.

```suggestion
                                alert(<?php echo xlj('There are currently other users of the category you are trying to delete. Please contact them and ask them to delete it. Categories may not be deleted while in use. This Categories are currently used by'); ?> + "\n" + thedata);
```

kojiromike (changes requested):
🤦 I am sorry, I realized I led you down the wrong road. While `echo js_escape('foo');` trivially converts to `echo '"foo"';`, once translation is involved it isn't safe anymore, on the (admittedly slim) chance that there's an attack vector through the translation database. So we need to use `xlj`, not just `xl` in those cases.

kojiromike (approved):
LGTM. All prior review points are addressed in the current diff.

kojiromike:
While you're here — two more sites I'd like to fold into this PR rather than spinning up a follow-up. Both are the same `js_escape(xl/xlt/xla(...))` family, just with the `xlt`/`xla` variants:

**1. `interface/super/edit_list.php:310`**

```php
echo "<script>let error=" . js_escape(xlt("The new list") . " [" . $_POST['newlistname'] . "] " . xlt("already exists! Please try again.")) . ";</script>";
```

The result feeds `alertMsg(error)`, which renders via `${jsText(message)}` — so `xlt()` HTML-escapes once, `jsText()` HTML-escapes again, and the user sees `&amp;` instead of `&` for any non-ASCII in the translation. Switch to per-translation `xlj()` with JS-side concat:

```suggestion
        echo "<script>let error=" . xlj("The new list") . " + ' [' + " . js_escape($_POST['newlistname']) . " + '] ' + " . xlj("already exists! Please try again.") . ";</script>";
```

**2. `interface/forms/procedure_order/common.php:1153-1154`**

```php
let prmt = <?php echo js_escape(xla("How many sets of specimen labels to create?") .
    "\n" . xla("Each test in order gets a label.")); ?>;
let count = prompt(prmt, '1');
```

`xla()` is attribute-escaping, but the sink is `prompt()` — a plain-text browser dialog — so any `"` or `&` in the translation renders as `&quot;` / `&amp;`. Same fix shape:

```suggestion
            let prmt = <?php echo xlj("How many sets of specimen labels to create?"); ?> + "\n" + <?php echo xlj("Each test in order gets a label."); ?>;
```

I checked the other `js_escape(xlt(...))` sites in the codebase (faxsms controllers, immunization_report) — those genuinely target HTML rendering, so `xlt()` is correct there. Just these two.

Kasirocswell:
@kojiromike - quick update.

I've folded in the two additional sites you flagged:

interface/super/edit_list.php:310 - js_escape(xlt(...) . $_POST['newlistname'] . xlt(...)) was double-HTML-escaping (xlt → jsText) and concatenating raw $_POST inside js_escape(). Now uses per-translation xlj() with JS-side concat, and wraps the user input in its own js_escape($_POST['newlistname']).
interface/forms/procedure_order/common.php:1153-1154 - js_escape(xla(...) . "\n" . xla(...)) was attribute-escaping for a plain-text prompt() sink. Now splits into two xlj() calls concatenated with a "\n" JS literal.

CI was green on every functional check (PHPStan level 10, Rector, PHPCS, integration matrix across PHP 8.2–8.5 × MySQL/MariaDB, isolated tests, JS/CSS lint, Semgrep, spell-check). The only red mark was codecov/patch reporting 0.00% of diff hit (target 50.00%).

That's structurally expected here: every line in the diff is inline <script> inside legacy interface/, library/custom_template/, portal/sign/, or sphere/ PHP pages, none of which have any existing PHPUnit coverage. Should I attempt to add tests for these sites, or is it reasonable to merge as-is given that this is a pure XSS-hardening cleanup with no behavioral change?

kojiromike:
I don't think adding tests is feasible within the scope of this pr.

kojiromike:
…however, phpstan is reporting a failure.

Kasirocswell:
@kojiromike 
$_POST['newlistname'] was mixed, so js_escape() (which is typed string by design - library/htmlspecialchars.inc.php explicitly forbids widening it) tripped PHPStan. Fixed by reading the value into a local and narrowing with is_string() per CLAUDE.md's "narrow, don't cast" guidance, keeping the file's $_POST baseline count unchanged.

kojiromike (approved):
Re-approving after the follow-up commits. Both added sites (edit_list.php, procedure_order/common.php) match the suggested shapes, and the PHPStan narrowing on `$_POST['newlistname']` is correct.

Kasirocswell:
@kojiromike
All required checks are passing now; the only red is codecov/patch at 0.00%, which you already agreed is out of scope here since every changed line is inline <script> in legacy interface/, library/custom_template/, portal/sign/, and sphere/ pages with no existing PHPUnit coverage. Would you mind merging when you have a moment?

PR #12146, opened 2026-05-15
refactor(ui): replace vendored SearchHighlight.js with vanilla TreeWalker helper

### LINKED ISSUE #10110: Replace vendored SearchHighlight.js with modern alternative
## Summary

`library/js/SearchHighlight.js` is a vendored jQuery plugin from 2007 (Renato Formato). It could be replaced with a modern solution.

## Current State

- Used in ~3 files
- Depends on jQuery
- Provides search term highlighting functionality

## Proposed Solution

Options:
1. Use native browser `CSS.highlights` API (modern browsers)
2. Use a maintained library like `mark.js`
3. Write a small vanilla JS replacement (the functionality is simple)

## Benefits

- Remove legacy jQuery dependency
- Better maintained code
- Smaller bundle size

progradedteam-dev:

Fixes #10110. Replaces the unmaintained 212-line vendored `library/js/SearchHighlight.js` jQuery plugin (and the parallel HTML-regex highlighter in `custom_report.js` / `portal_custom_report.php`) with a small vanilla DOM-walking module that uses `TreeWalker` + `Range`. No jQuery dependency, no new npm dependency.

The previous code ran regex over `innerHTML` and then a second regex to repair broken HTML tags — fragile against any text containing markup characters. The new module walks text nodes individually, finds matches, and wraps each in `<mark class="hilite">` using `splitText()` + `insertBefore()`. `TreeWalker`, `NodeFilter`, `splitText`, and `insertBefore` are supported in every browser within OpenEMR's declared `browserslist`.

Per the design discussion in #10110, this is the zero-dependency approach kojiromike asked us to try before falling back to a third-party library.

- Add `library/js/searchHighlight.js` exposing `window.OpenEMRSearchHighlight` with `search()`, `highlight()`, `unhighlight()`, `buildRegex()`, and `escapeRegExp()`
- Update `config/config.yaml` asset registry to point the `search-highlight` asset at the new filename
- Rewrite `mark_hilight()`, `remove_mark_all()`, and `get_word_count()` in `interface/patient_file/report/custom_report.js` and `portal/report/portal_custom_report.php` to delegate to the new module
- Drop the now-redundant `doSearch()`, `reset_highlight()`, and `remove_mark()` helpers
- Delete `library/js/SearchHighlight.js`

**Behavior notes:**
- The new highlighter walks text nodes individually, so matches that span HTML element boundaries (e.g., `lorem <em>ipsum</em>`) are not joined. The previous HTML-regex approach attempted to bridge tags but was fragile; `acrossElements` support can be added in a follow-up if needed.
- Multi-word queries highlight each token separately rather than the full phrase. Phrase navigation in `next`/`prev` still works because `get_word_count()` returns the per-form mark count.
- Output shape is unchanged — `<mark class="hilite">` with sequential `result_N` ids — so existing CSS and the `next`/`prev`/`hilite2` navigation logic are unmodified.

**Tested:**
- ESLint clean on changed files
- Functional unit tests via in-page `evaluate_script`: single-word matching, multi-word, `unhighlight()`, `SKIP_TAGS` exclusion for `<script>` and `<textarea>`, and case sensitivity
- Manual: `portal/report/portal_custom_report.php` loads the new helper with no console errors
- Manual: `interface/patient_file/report/custom_report.php` loads the new helper via the central `Header::setupHeader(['search-highlight'])` asset registry

#### Was an AI assistant used? No

cc @kojiromike @msummers42


library/js/searchHighlight.js
```diff
@@ -0,0 +1,250 @@
...
+     * @returns {RegExp|null}
+     */
+    function buildRegex(query, opts) {
+        const options = opts || {};
+        const trimmed = String(query == null ? '' : query).trim();
+        if (!trimmed) {
+            return null;
+        }
+
+        const tokens = trimmed.split(/\s+/).map(escapeRegExp);
```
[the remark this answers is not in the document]
progradedteam-dev: Fixed, buildRegex now splits on [\s,]+ to match the legacy plugin's [\s,+.]+ delimiter behavior.

library/js/searchHighlight.js
```diff
@@ -0,0 +1,250 @@
...
+ * @copyright Copyright (c) 2026 OpenEMR Contributors
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+/* This module is pure vanilla JavaScript — no jQuery dependency. */
+
+(function (window) {
+    'use strict';
+
+    // Tags whose text content should never be highlighted.
+    const SKIP_TAGS = new Set(['SCRIPT', 'STYLE', 'TEXTAREA', 'NOSCRIPT']);
```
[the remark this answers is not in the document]
progradedteam-dev: Worth declining, the browserslist entry is stale. jQuery 3.x already dropped IE 8, Angular 1.8 never supported it, and const/let/arrow functions appear throughout dozens of existing library/js/ files. The old SearchHighlight.js had the same incompatibility since it depended on jQuery 3.x. This is not a regression.

interface/patient_file/report/custom_report.js
```diff
@@ -236,30 +185,23 @@ function clear_last_visit(){
...
+    // this form's container. The new highlighter wraps each token of the
+    // phrase as a separate <mark>, so phrase navigation steps over the
+    // whole group of token-marks.
+    var container = document.getElementById('search_div_' + form_id + '_' + form_dir);
+    if (!container) {
+      return 0;
     }
-  }else{
-    return 1;
+    return container.querySelectorAll('.hilite').length;
```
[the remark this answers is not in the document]
progradedteam-dev: Good catch. Fixed in the follow-up commit, `get_word_count` now derives the count directly from the query tokens (`trimmed.split(/[\s,]+/).filter(...)`) instead of querying live `.hilite` elements. The DOM-based approach was broken because `next()`/`prev()` promote all `.hilite` wrappers to `.hilite2` before `get_word_count` is called on the next click, so the selector always returned 0.

portal/report/portal_custom_report.php
```diff
@@ -389,30 +339,23 @@ function clear_last_visit(){
...
+      // this form's container. The new highlighter wraps each token of the
+      // phrase as a separate <mark>, so phrase navigation steps over the
+      // whole group of token-marks.
+      var container = document.getElementById('search_div_' + form_id + '_' + form_dir);
+      if (!container) {
+        return 0;
       }
-    }else{
-      return 1;
+      return container.querySelectorAll('.hilite').length;
```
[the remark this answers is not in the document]
progradedteam-dev: Same fix applied here, `get_word_count` in the portal report now also counts query tokens from the string rather than querying `.hilite` DOM elements. Same root cause as the clinician report: the class swap to `hilite2` happened before this function ran on subsequent clicks.

library/js/searchHighlight.js
kojiromike, line 23: Would you be open to adding a Jest test for this module in `tests/js/`? The repo has Jest set up (`jest.config.js`, see `tests/js/pdf-utils.test.js` for the pattern). This file is pure logic with well-defined edge cases (token splitting on `[\s,]+`, `SKIP_TAGS` exclusion, zero-length match infinite-loop guard, multi-token wrap order). A small jsdom-backed test would lock those in — manual in-browser verification is hard to regress against.

progradedteam-dev: sure, no problem

progradedteam-dev: @kojiromike Done, added `tests/js/searchHighlight.test.js` (29 tests, jsdom-backed). Covers token splitting on whitespace and commas, SKIP_TAGS exclusion, zero-length match guard, multi-token wrap order, case sensitivity, and the unhighlight/normalize round-trip.

library/js/searchHighlight.js
```diff
@@ -0,0 +1,250 @@
...
+     * @returns {RegExp|null}
+     */
+    function buildRegex(query, opts) {
+        const options = opts || {};
+        const trimmed = String(query == null ? '' : query).trim();
+        if (!trimmed) {
+            return null;
+        }
+
+        const tokens = trimmed.split(/[\s,]+/).map(escapeRegExp).filter(Boolean);
```
kojiromike, line 55: Separator-only queries (e.g. `,,,`) leave `tokens` empty and produce `new RegExp('', 'gi')`, which the highlighter then runs against every text node for no inserts. Bail out:

```suggestion
        const tokens = trimmed.split(/[\s,]+/).map(escapeRegExp).filter(Boolean);
        if (tokens.length === 0) {
            return null;
        }
```

library/js/searchHighlight.js
```diff
@@ -0,0 +1,250 @@
...
+ * search. Uses TreeWalker + Range; no jQuery dependency.
+ *
+ * Highlights wrap each match in `<mark class="hilite">...</mark>` (or any
+ * configured tag/class), matching the existing CSS selectors.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    OpenEMR Contributors
+ * @copyright Copyright (c) 2026 OpenEMR Contributors
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
```
kojiromike, line 15: Please post explicit values for author and copyright.

library/js/searchHighlight.js
progradedteam-dev: Good catch. Fixed, added a `tokens.length === 0` bail-out after the filter so separator-only inputs like `,,,` return `null` rather than producing `new RegExp('', 'gi')`. Also added a test case covering this edge case (30 tests total now).

library/js/searchHighlight.js
progradedteam-dev: Done, updated with explicit author and copyright.

package.json
```diff
@@ -59,6 +59,7 @@
     "gulp-sourcemaps": "3.0.0",
     "gulp-watch": "5.0.1",
     "jest": "29.7.0",
+    "jest-environment-jsdom": "^30.4.1",
```
kojiromike: `jest-environment-jsdom@^30` is paired with `jest@29.7.0` here, which is a major-version skew — `jest-environment-jsdom` is supposed to track Jest's major version. The lockfile churn (+967) is largely driven by this: nested copies of `@jest/environment@30.4.1`, `@jest/types@30.4.1`, `jest-mock@30.4.1`, `pretty-format@30.4.1`, etc. are pulled in as duplicates of the Jest-29 copies already in `node_modules` from the rest of the toolchain. Tests may pass today, but the environment is loading a different `@jest/environment` than the runner.

Pinning to the matching line also keeps the style consistent with the rest of `devDependencies` (no other entry uses `^`).

```suggestion
    "jest-environment-jsdom": "29.7.0",
```

kojiromike (commented):
Iteration since the prior rounds looks clean — the tokens.length === 0 bail-out and the query-token-derived w_count both landed. One pre-merge concern on the new devDependency.

library/js/searchHighlight.js
```diff
@@ -0,0 +1,253 @@
...
+ * mark_hilight() regex approach previously used in the patient custom-report
+ * search. Uses TreeWalker + Range; no jQuery dependency.
+ *
+ * Highlights wrap each match in `<mark class="hilite">...</mark>` (or any
+ * configured tag/class), matching the existing CSS selectors.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Andrew Alanis <progradedteam@gmail.com>
+ * @copyright Copyright (c) 2026 Andrew Alanis
```
kojiromike: I believe OpenCoreEMR is paying for this work. If I have that right, please change the copyright to…

```suggestion
 * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
```

progradedteam-dev: Yep, you're right.

progradedteam-dev: Updated, copyright now attributes OpenCoreEMR Inc.

package.json
progradedteam-dev: Good catch on the major-version skew. Pinned to `29.7.0` and regenerated the lockfile; the duplicate @jest/environment, @jest/types, jest-mock, and pretty-format trees are gone.

kojiromike (approved):
Iteration converged — splitText cursor-walk, token-derived w_count, separator-only bail-out, jest-environment-jsdom pin, and OpenCoreEMR copyright all landed cleanly. 30 jsdom-backed tests lock the module surface. Merge once CI completes.

progradedteam-dev:
@kojiromike looks like `All Checks Passed` is only being blocked by `codecov/patch` and `codecov/project`, and both seem like noise:

* `codecov/patch` shows 0% on 20 lines, but those lines are inline JS inside `portal/report/portal_custom_report.php`. JS coverage tools cannot trace JS inside a PHP template, and PHP coverage does not execute the JS. The same logic is already covered by the 30 Jest tests for `library/js/searchHighlight.js`.
* `codecov/project` shows -2.17%, which appeared after the rebase. The flag-level numbers point to incomplete Codecov uploads from this run versus the new base. The Codecov bot summary still shows ✅ “Project coverage is 22.86%.”

Would you be open to merging through these or overriding the Codecov gate? I’m happy to extract the portal inline JS into a separate `.js` file in a follow-up PR if you’d rather close the coverage gap structurally.

Issue #10602, opened 2026-02-06
feat(vitals): add input validation for numeric fields


kojiromike:
## Problem

Users can enter invalid data in vitals form fields:
- Text in numeric fields
- Excessively large numbers (unrealistic values)
- Negative values where inappropriate (e.g., blood pressure)

This affects clinical decision support and data quality.

## Expected Behavior

1. Numeric fields should only accept numeric input (client-side and server-side)
2. Fields should have reasonable min/max constraints based on clinical norms
3. Values outside typical ranges should show a warning (not block entry, as outliers can be valid)

## Source

Sunchu K, et al. "Lessons Learned from OpenEMR Implementation in Graduate Health Informatics Curriculum." AMIA 2024.
https://pmc.ncbi.nlm.nih.gov/articles/PMC12099383/

PR #12163, opened 2026-05-18
feat(vitals): add input validation for numeric fields

lilit1121:
#10602 

Adds comprehensive client-side and server-side validation for the vitals form to prevent invalid data entry (non-numeric, negative, out-of-range values) and warn on clinically unusual values.          
                                                                                                                                                                                                           
  - New VitalsFieldRanges class as single source of truth for hard bounds and clinical warning ranges across all 15 numeric vitals fields (including metric equivalents)
  - Extended JS validation from 6 hardcoded fields to all numeric inputs via data-* attributes                                                                                                           
  - Added validateVitalField() with real-time validation on change events — errors block submission, warnings allow it
  - Added inputmode="decimal" and data-min/max/warning-min/warning-max attributes to textbox and conversion Twig templates
  - Added FormVitals::validate() method for server-side range checking as a safety net
  - Added .warning CSS styles (amber border) and inline warning messages
  - Preserved support for weight # format (e.g. 5#4 for lbs/oz)

  Was an AI assistant used? Yes

### DIFF
--- a/interface/forms/vitals/C_FormVitals.class.php
+++ b/interface/forms/vitals/C_FormVitals.class.php
@@ -22,6 +22,7 @@ use OpenEMR\Common\Forms\FormVitalDetails;
 use OpenEMR\Common\Forms\FormVitals;
 use OpenEMR\Common\Forms\ReasonStatusCodes;
 use OpenEMR\Common\Forms\VitalsFieldRanges;
+use OpenEMR\Common\Session\SessionUtil;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Common\Uuid\UuidRegistry;
@@ -31,10 +32,7 @@ use OpenEMR\Services\VitalsService;
 
 class C_FormVitals
 {
-    /**
-     * @var FormVitals
-     */
-    public $vitals;
+    public FormVitals $vitals;
 
     public $template_dir;
     public $form_id;
@@ -354,7 +352,7 @@ class C_FormVitals
 
         $session = SessionWrapperFactory::getInstance()->getActiveSession();
         $validationErrors = $session->get('vitals_validation_errors', []);
-        $session->remove('vitals_validation_errors');
+        SessionUtil::unsetSession(['vitals_validation_errors']);
 
         $data = [
             'vitals' => $vitals
@@ -463,8 +461,7 @@ class C_FormVitals
 
         $validationResult = $this->vitals->validate();
         if ($validationResult['errors'] !== []) {
-            $session = SessionWrapperFactory::getInstance()->getActiveSession();
-            $session->set('vitals_validation_errors', $validationResult['errors']);
+            SessionUtil::setSession('vitals_validation_errors', $validationResult['errors']);
             return;
         }
 
--- a/interface/forms/vitals/C_FormVitals.class.php
+++ b/interface/forms/vitals/C_FormVitals.class.php
@@ -172,15 +172,15 @@ class C_FormVitals
             ]
             ,[
                 'type' => 'textbox'
-                , xl('Pulse') => 'title'
+                ,'title' => xl('Pulse')
                 // eventually we could just grab the raw values...
                 ,'vitalsValue' => "get_pulse"
                 ,'precision' => 0
                 ,'input' => 'pulse'
                 ,'unit' => 'per min'
-                , xl('per min') => 'unitLabel'
+                ,'unitLabel' => xl('per min')
                 ,'codes' => 'LOINC:8867-4'
-                ,'valiuguufdation' => VitalsFieldRanges::getRangeForField('pulse')
+                ,'validation' => VitalsFieldRanges::getRangeForField('pulse')
             ]
             ,[
                 'type' => 'textbox'
@@ -190,6 +190,7 @@ class C_FormVitals
                 ,'precision' => 0
                 ,'input' => 'respiration'
                 ,'unit' => 'per min'
+                ,'unitLabel' => xl('per min')
                 ,'codes' => 'LOINC:9279-1'
                 ,'validation' => VitalsFieldRanges::getRangeForField('respiration')
             ]
@@ -352,8 +353,12 @@ class C_FormVitals
         }
 
         $session = SessionWrapperFactory::getInstance()->getActiveSession();
+        $validationErrors = $session->get('vitals_validation_errors', []);
+        $session->remove('vitals_validation_errors');
+
         $data = [
             'vitals' => $vitals
+            ,'validationErrors' => $validationErrors
             ,'vitalFields' => $vitalFields
             ,'FORM_ACTION' => OEGlobalsBag::getInstance()->getWebRoot()
             ,'DONT_SAVE_LINK' => OEGlobalsBag::getInstance()->get('form_exit_url')
--- a/interface/forms/vitals/C_FormVitals.class.php
+++ b/interface/forms/vitals/C_FormVitals.class.php
@@ -21,6 +21,7 @@ use OpenEMR\Common\Forms\BmiCategory;
 use OpenEMR\Common\Forms\FormVitalDetails;
 use OpenEMR\Common\Forms\FormVitals;
 use OpenEMR\Common\Forms\ReasonStatusCodes;
+use OpenEMR\Common\Forms\VitalsFieldRanges;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Common\Twig\TwigContainer;
 use OpenEMR\Common\Uuid\UuidRegistry;
@@ -130,6 +131,7 @@ class C_FormVitals
                 ,'precision' => 2
                 ,'vitalsValueUSAHelpTitle' => xl("Decimal pounds or pounds and ounces separated by #(e.g. 5#4)")
                 ,'codes' => 'LOINC:29463-7'
+                ,'validation' => VitalsFieldRanges::getRangeForField('weight')
             ]
             ,[
                 'type' => 'textbox_conversion'
@@ -144,6 +146,7 @@ class C_FormVitals
                 ,'unitMetricLabel' => xl('cm')
                 ,'precision' => 2
                 ,'codes' => 'LOINC:8302-2'
+                ,'validation' => VitalsFieldRanges::getRangeForField('height')
             ]
             ,[
                 'type' => 'textbox'
@@ -154,6 +157,7 @@ class C_FormVitals
                 ,'unit' => 'mmHg'
                 ,'unitLabel' => xl('mmHg')
                 ,'codes' => 'LOINC:8480-6'
+                ,'validation' => VitalsFieldRanges::getRangeForField('bps')
             ]
             ,[
                 'type' => 'textbox'
@@ -164,17 +168,19 @@ class C_FormVitals
                 ,'unit' => 'mmHg'
                 ,'unitLabel' => xl('mmHg')
                 ,'codes' => 'LOINC:8462-4'
+                ,'validation' => VitalsFieldRanges::getRangeForField('bpd')
             ]
             ,[
                 'type' => 'textbox'
-                ,'title' => xl('Pulse')
+                , xl('Pulse') => 'title'
                 // eventually we could just grab the raw values...
                 ,'vitalsValue' => "get_pulse"
                 ,'precision' => 0
                 ,'input' => 'pulse'
                 ,'unit' => 'per min'
-                ,'unitLabel' => xl('per min')
+                , xl('per min') => 'unitLabel'
                 ,'codes' => 'LOINC:8867-4'
+                ,'valiuguufdation' => VitalsFieldRanges::getRangeForField('pulse')
             ]
             ,[
                 'type' => 'textbox'
@@ -184,8 +190,8 @@ class C_FormVitals
                 ,'precision' => 0
                 ,'input' => 'respiration'
                 ,'unit' => 'per min'
-                ,'unitLabel' => xl('per min')
                 ,'codes' => 'LOINC:9279-1'
+                ,'validation' => VitalsFieldRanges::getRangeForField('respiration')
             ]
             ,[
                 'type' => 'textbox_conversion'
@@ -200,6 +206,7 @@ class C_FormVitals
                 ,'unitMetricLabel' => xl('C')
                 ,'precision' => 2
                 ,'codes' => 'LOINC:8310-5'
+                ,'validation' => VitalsFieldRanges::getRangeForField('temperature')
             ]
             ,[
                 'type' => 'template'
@@ -215,6 +222,7 @@ class C_FormVitals
                 ,'unit' => '%'
                 ,'unitLabel' => '%'
                 ,'codes' => 'LOINC:59408-5'
+                ,'validation' => VitalsFieldRanges::getRangeForField('oxygen_saturation')
             ]
             ,[
                 'type' => 'textbox'
@@ -226,6 +234,7 @@ class C_FormVitals
                 ,'unit' => 'l/min'
                 ,'unitLabel' => xl('l/min')
                 ,'codes' => 'LOINC:3151-8'
+                ,'validation' => VitalsFieldRanges::getRangeForField('oxygen_flow_rate')
             ]
             ,[
                 'type' => 'textbox'
@@ -237,6 +246,7 @@ class C_FormVitals
                 ,'unit' => '%'
                 ,'unitLabel' => '%'
                 ,'codes' => 'LOINC:3150-0'
+                ,'validation' => VitalsFieldRanges::getRangeForField('inhaled_oxygen_concentration')
             ]
             ,[
                 'type' => 'textbox_conversion'
@@ -253,6 +263,7 @@ class C_FormVitals
                 // hide_circumferences
                 ,'hide' => OEGlobalsBag::getInstance()->get('gbl_vitals_options') > 0
                 ,'codes' => "LOINC:9843-4"
+                ,'validation' => VitalsFieldRanges::getRangeForField('head_circ')
             ]
             ,[
                 'type' => 'textbox_conversion'
@@ -269,6 +280,7 @@ class C_FormVitals
                 // hide_circumferences
                 ,'hide' => OEGlobalsBag::getInstance()->get('gbl_vitals_options') > 0
                 ,'codes' => "LOINC:9843-4"
+                ,'validation' => VitalsFieldRanges::getRangeForField('waist_circ')
             ]
             ,[
                 'type' => 'template'
@@ -287,6 +299,7 @@ class C_FormVitals
                 ,'unit' => '%'
                 ,'unitLabel' => '%'
                 ,'codes' => 'LOINC:77606-2'
+                ,'validation' => VitalsFieldRanges::getRangeForField('ped_weight_height')
                 ,'hide' => !$show_pediatric_fields
             ]
             ,[
@@ -298,6 +311,7 @@ class C_FormVitals
                 ,'unit' => '%'
                 ,'unitLabel' => '%'
                 ,'codes' => 'LOINC:59576-9'
+                ,'validation' => VitalsFieldRanges::getRangeForField('ped_bmi')
                 ,'hide' => !$show_pediatric_fields
             ]
             ,[
@@ -309,6 +323,7 @@ class C_FormVitals
                 ,'unit' => '%'
                 ,'unitLabel' => '%'
                 ,'codes' => 'LOINC:8289-1'
+                ,'validation' => VitalsFieldRanges::getRangeForField('ped_head_circ')
                 ,'hide' => !$show_pediatric_fields
             ]
             ,[
@@ -440,6 +455,13 @@ class C_FormVitals
         $this->vitals = new FormVitals();
         $this->vitals->populate_array($vitalsArray);
         $this->populate_object($this->vitals);
+
+        $validationResult = $this->vitals->validate();
+        if ($validationResult['errors'] !== []) {
+            $session = SessionWrapperFactory::getInstance()->getActiveSession();
+            $session->set('vitals_validation_errors', $validationResult['errors']);
+            return;
+        }
 
         $vitalsService->saveVitalsForm($this->vitals);
         return;
--- a/interface/forms/vitals/save.php
+++ b/interface/forms/vitals/save.php
@@ -31,7 +31,7 @@ $c->default_action_process();
 $validationErrors = $session->get('vitals_validation_errors');
 if ($validationErrors !== null && $validationErrors !== []) {
     $formUrl = OEGlobalsBag::getInstance()->getWebRoot()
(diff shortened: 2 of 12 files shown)


lilit1121:
@msummers42 @kojiromike sending you this PR for review.
Thanks!

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -164,17 +168,19 @@ public function default_action()
                 ,'unit' => 'mmHg'
                 ,'unitLabel' => xl('mmHg')
                 ,'codes' => 'LOINC:8462-4'
+                ,'validation' => VitalsFieldRanges::getRangeForField('bpd')
             ]
             ,[
                 'type' => 'textbox'
-                ,'title' => xl('Pulse')
+                , xl('Pulse') => 'title'
```
kojiromike: Key and value are swapped. As written, the array key becomes the translated string for "Pulse" and the value becomes the literal string `'title'`. The actual `title` key disappears from this field config.

```suggestion
                ,'title' => xl('Pulse')
```

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -164,17 +168,19 @@ public function default_action()
...
                 // eventually we could just grab the raw values...
                 ,'vitalsValue' => "get_pulse"
                 ,'precision' => 0
                 ,'input' => 'pulse'
                 ,'unit' => 'per min'
-                ,'unitLabel' => xl('per min')
+                , xl('per min') => 'unitLabel'
```
kojiromike: Same key/value swap as the `title` line above — Pulse loses its `unitLabel`.

```suggestion
                ,'unitLabel' => xl('per min')
```

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -164,17 +168,19 @@ public function default_action()
...
                 ,'codes' => 'LOINC:8867-4'
+                ,'valiuguufdation' => VitalsFieldRanges::getRangeForField('pulse')
```
kojiromike: Typo: `'valiuguufdation'` should be `'validation'`. As written, Pulse never gets a `validation` entry, so the textbox template's `field.validation is defined` check is false and the input renders with no `data-min/max/warning-*` attributes — the new JS validator silently skips pulse.

```suggestion
                ,'validation' => VitalsFieldRanges::getRangeForField('pulse')
```

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -184,8 +190,8 @@ public function default_action()
                 ,'precision' => 0
                 ,'input' => 'respiration'
                 ,'unit' => 'per min'
-                ,'unitLabel' => xl('per min')
                 ,'codes' => 'LOINC:9279-1'
+                ,'validation' => VitalsFieldRanges::getRangeForField('respiration')
```
kojiromike, line 193: The previous line `,'unitLabel' => xl('per min')` was removed in this hunk and not restored. The respiration field will render without its unit label. Please add it back:

```suggestion
                ,'unitLabel' => xl('per min')
                ,'codes' => 'LOINC:9279-1'
                ,'validation' => VitalsFieldRanges::getRangeForField('respiration')
```

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -441,6 +456,13 @@ public function default_action_process()
         $this->vitals->populate_array($vitalsArray);
         $this->populate_object($this->vitals);
 
+        $validationResult = $this->vitals->validate();
+        if ($validationResult['errors'] !== []) {
+            $session = SessionWrapperFactory::getInstance()->getActiveSession();
+            $session->set('vitals_validation_errors', $validationResult['errors']);
+            return;
```
kojiromike, line 465: Returning here means the form silently fails to save: nothing in this PR reads `vitals_validation_errors` from the session, and there's no redirect/flash so the user sees the form close (or render with default state) as if save succeeded. The JS validator presumably catches most cases first, but the PR description calls server-side validation a "safety net" — for it to be one, the failure has to surface to the user (re-render the form with errors, redirect with a flash, etc.). Could you wire the error path to something the user actually sees?

src/Common/Forms/FormVitals.php
```diff
@@ -467,6 +467,69 @@ public function has_reason_for_column($column)
...
+            'head_circ' => 'head_circ',
+            'waist_circ' => 'waist_circ',
+            'ped_weight_height' => 'ped_weight_height',
+            'ped_bmi' => 'ped_bmi',
+            'ped_head_circ' => 'ped_head_circ',
+        ];
+
+        foreach ($fieldPropertyMap as $fieldName => $property) {
+            $value = $this->$property;
+            if ($value === null || $value === '' || $value === 0 || $value === '0') {
```
kojiromike: Treating `0`/`'0'` as "not entered" is risky for vitals. A blood pressure of 0, pulse of 0, or respiration of 0 is a clinically critical value that should be flagged (and likely rejected) by validation, not silently skipped. Suggest narrowing this to `null` and `''` only:

```suggestion
            if ($value === null || $value === '') {
```

src/Common/Forms/FormVitals.php
```diff
@@ -467,6 +467,69 @@ public function has_reason_for_column($column)
...
+                continue;
+            }
+
+            if ($numericValue < 0) {
+                $errors[] = "Field '{$fieldName}' cannot be negative.";
+                continue;
+            }
+
+            if ($numericValue < $range['min'] || $numericValue > $range['max']) {
+                $errors[] = "Field '{$fieldName}' is outside acceptable range ({$range['min']}–{$range['max']}).";
```
kojiromike: Error strings here (and on lines 505, 516, 526) are untranslated English with machine field names (`oxygen_saturation`, etc.) interpolated. If these are ever surfaced to the user (see the related comment on `C_FormVitals::default_action_process` about the silent-fail path), they need `xl()` wrapping and human-readable labels.

src/Common/Forms/VitalsFieldRanges.php
```diff
@@ -0,0 +1,85 @@
...
+            ],
+            'respiration' => [
+                'min' => 0, 'max' => 150, 'warningMin' => 2, 'warningMax' => 100,
+            ],
+            'temperature' => [
+                'min' => 0, 'max' => 120, 'warningMin' => 80, 'warningMax' => 115,
+                'metricMin' => 0, 'metricMax' => 48.9, 'metricWarningMin' => 26.7, 'metricWarningMax' => 46.1,
+            ],
+            'oxygen_saturation' => [
+                'min' => 0, 'max' => 100, 'warningMin' => 0, 'warningMax' => 100,
```
kojiromike: `warningMin`/`warningMax` equal the hard `min`/`max` here, so the warning band can never fire — any in-range value passes silently, any out-of-range value is a hard error. The same fully-inert pattern appears for `inhaled_oxygen_concentration`, `ped_weight_height`, `ped_bmi`, and `ped_head_circ`. A few other fields (`oxygen_flow_rate`, `head_circ`, `waist_circ`) have an inert *lower* warning bound (`warningMin == min == 0`) while the upper band still fires; less broken but probably also not intended. The class docblock promises "clinically-informed validation ranges" — these look like placeholder values. Could you either fill in real warning bands (the AMIA paper referenced in #10602 is a starting point, plus standard pediatric growth chart bounds) or drop the warning machinery for these fields?

kojiromike (changes requested):
Thanks for taking this on — centralizing the ranges and generalizing the JS validator from 6 hard-coded fields to all numeric inputs is the right direction. A few things to address before this can ship, though; the Pulse and Respiration field entries in `C_FormVitals::default_action()` were damaged in this diff in a way that suggests the form wasn't opened after the change.

**Blocking issues**

1. **Pulse entry is broken** (3 separate problems in 4 adjacent lines):
   - `, xl('Pulse') => 'title'` — key and value swapped; the translated string becomes the array key and the literal string `'title'` becomes its value. The `title` key is now missing.
   - `, xl('per min') => 'unitLabel'` — same swap for `unitLabel`.
   - `'valiuguufdation'` — typo; pulse never receives a `validation` entry, so the data attributes (and therefore the new JS validator) silently skip pulse.

2. **Respiration loses its `unitLabel`** — the old `,'unitLabel' => xl('per min')` line was deleted and not replaced.

3. **Server-side validation silently fails the save.** In `default_action_process()` the new branch sets `vitals_validation_errors` in the session and `return`s before calling `saveVitalsForm()`. There's no redirect, no flash, and nothing in this PR reads that session key — from the user's perspective the form just closes and the data is gone. The JS validator presumably catches most cases, but the comment in the PR description calls this a "safety net," which means it has to surface the error somehow.

**Other findings**

4. **`$value === 0 || $value === '0'` skip in `FormVitals::validate()`** treats zero as "not entered." For BP, pulse, respiration, and oxygen saturation, zero is a clinically critical value that should be flagged, not silently dropped. Use `null`/empty-string only.

5. **Server-side error messages are not translated** (no `xl()`), and they interpolate field machine-names like `'oxygen_saturation'`. Even if #3 is fixed to surface them, they aren't user-facing in their current form.

6. **Several warning bands are inert.** In `VitalsFieldRanges`, `oxygen_saturation`, `inhaled_oxygen_concentration`, `ped_weight_height`, `ped_bmi`, and `ped_head_circ` all have `warningMin`/`warningMax` equal to the hard `min`/`max`, so the warning band can never fire — any in-range value passes silently, any out-of-range value is a hard error. A few other fields (`oxygen_flow_rate`, `head_circ`, `waist_circ`) have an inert *lower* warning bound (`warningMin == min == 0`) while the upper band still fires; those are less broken but probably still not what was intended. These all look like placeholder values rather than the "clinically-informed" ranges the class docblock promises.

7. **No tests.** `FormVitals::validate()` and `VitalsFieldRanges` are both straightforward to cover with isolated tests (no DB or session required). At minimum: each field's hard-bound rejection, warning-band hit, valid value, and the non-numeric branch.

The blocking issues (#1–#3) all involve obvious-on-first-load breakage of the Pulse field and silent data loss on save. Please open the vitals form in a running OpenEMR instance, enter pulse + respiration values (both valid and out-of-range), and confirm the form renders correctly and either saves or shows an error before pushing again.

interface/forms/vitals/C_FormVitals.class.php
lilit1121: Fixed!

interface/forms/vitals/C_FormVitals.class.php
lilit1121: Fixed!

interface/forms/vitals/C_FormVitals.class.php
lilit1121: Fixed!

interface/forms/vitals/C_FormVitals.class.php
lilit1121: Fixed!

interface/forms/vitals/C_FormVitals.class.php
lilit1121: Done!

src/Common/Forms/FormVitals.php
lilit1121: Done!

src/Common/Forms/FormVitals.php
lilit1121: Done!

src/Common/Forms/VitalsFieldRanges.php
lilit1121: Done!

interface/forms/vitals/save.php
```diff
@@ -26,5 +26,13 @@
 CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
 
 $c = new C_FormVitals();
-echo $c->default_action_process();
-@formJump();
+$c->default_action_process();
+
+$validationErrors = $session->get('vitals_validation_errors');
+if (!empty($validationErrors)) {
```
kojiromike: New code shouldn't use `empty()` — the repo CLAUDE.md calls it out specifically (the `empty("0")` truthiness trap, etc.). Since this value comes from `$session->get('vitals_validation_errors')` with no default, it's either `null` or a `list<string>`. An explicit check makes the intent obvious and stays inside the documented standard:

```suggestion
if ($validationErrors !== null && $validationErrors !== []) {
```

kojiromike (changes requested):
Thanks for turning this around quickly. I re-checked each prior thread against `c0fa79da` and the eight findings from the previous round all look resolved:

- Pulse `title`, `unitLabel`, and `validation` keys are correct (`C_FormVitals.class.php:175–183`).
- Respiration's `unitLabel` is restored (`C_FormVitals.class.php:193`).
- Server-side validation errors now surface — stored in session by `default_action_process()`, redirected back to the form by `save.php`, displayed by the alert block in `vitals.html.twig`.
- `FormVitals::validate()` no longer collapses `0` into "not entered" — only `null`/`''` are skipped.
- Error strings are wrapped with `xl()` and use the same human labels as the field titles.
- Warning bands are now mostly distinct from hard bounds. The remaining `warningMax == max` cases (`oxygen_saturation`, `inhaled_oxygen_concentration`) are at 100%, the physiological ceiling, so the "inert upper band" there is correct rather than a placeholder. Isolated tests for the field map and `validate()` are also in.

One follow-up before this is ready to ship: **`empty()` in new code** — see the inline on `save.php:32`. The repo CLAUDE.md forbids `empty()` in new code; please switch to an explicit check.

I'll hold off on resolving the prior threads until that's addressed so it's easy to scan the full history.

src/Common/Forms/FormVitals.php
```diff
@@ -467,6 +467,89 @@ public function has_reason_for_column($column)
...
+            'ped_bmi' => xl('Pediatric BMI Percentile'),
+            'ped_head_circ' => xl('Pediatric Head Circumference Percentile'),
+        ];
+
+        foreach ($fieldPropertyMap as $fieldName => $property) {
+            $value = $this->$property;
+            if ($value === null || $value === '') {
+                continue;
+            }
+
```
kojiromike, line 485: Wrapping each fragment in `xl()` does technically address the prior "untranslated" point, but the resulting strings still can't be properly localized: gettext entries are per-fragment, so translators see `Field`, `is outside acceptable range`, etc. in isolation and can't reorder the sentence for languages where word order differs (German verb-final, Japanese SOV, Arabic RTL, etc.). The `'{$label}'` is also welded into the middle of the English sentence rather than being a placeholder.

Usual fix is a single `xl()` template per message with `sprintf`-style slots, e.g.:

```suggestion
                $errors[] = sprintf(xl('Field "%1$s" is outside acceptable range (%2$s–%3$s).'), $label, $range['min'], $range['max']);
```

Same shape applies to the `has a non-numeric value`, `cannot be negative`, and `is outside typical clinical range` strings on the surrounding lines.

kojiromike (changes requested):
Re-checked `c0fa79d` / `108251d` — all nine prior blockers are resolved:

- Pulse `title`, `unitLabel`, `validation` keys correct; Respiration `unitLabel` restored.
- Server-side validation errors now surface end-to-end (session → `save.php` redirect → alert in `vitals.html.twig`).
- `validate()` no longer collapses `0` into "not entered".
- Warning bands distinct from hard bounds where they meaningfully can be; the remaining `warningMax == max == 100` cases (`oxygen_saturation`, `inhaled_oxygen_concentration`) are at the physiological ceiling, so the upper band being inert there is correct rather than a placeholder.
- Isolated `validate()` tests cover valid/non-numeric/negative/above-max/null/empty/zero per field.
- `save.php` uses the explicit `!== null && !== []` check rather than `empty()`.

One i18n point left (inline). Otherwise this is ready.

src/Common/Forms/FormVitals.php
lilit1121: Done!

kojiromike (approved):

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -441,6 +461,13 @@ public function default_action_process()
         $this->vitals->populate_array($vitalsArray);
         $this->populate_object($this->vitals);
 
+        $validationResult = $this->vitals->validate();
```
kojiromike, line 462: PHPStan reports `Cannot call method validate() on mixed` here, and `Cannot access offset 'errors' on mixed` on the next line. The cause isn't this call — it's that `$this->vitals` is declared as `public $vitals;` (untyped) near the top of the class. PHPStan reads the property type from the declaration, not from the local `$this->vitals = new FormVitals()` above, because the property could be reassigned between the `new` and the read (`populate_object()` even takes it by reference). Typing the property fixes both errors.

Change the property declaration (line ~37) to:

```php
public FormVitals $vitals;
```

Non-nullable — it's only ever assigned at line 460 and every read happens later in the same method, so there's no real null state to model.

Then swap the session write here to `SessionUtil::setSession()` to address the separate phpstan error on line 467 (`$session->set()` is forbidden because it silently fails on `read_and_close` sessions):

```suggestion
        $validationResult = $this->vitals->validate();
        if ($validationResult['errors'] !== []) {
            SessionUtil::setSession('vitals_validation_errors', $validationResult['errors']);
            return;
        }
```

Don't forget the `use OpenEMR\Common\Session\SessionUtil;` at the top of the file (if not already present).

interface/forms/vitals/C_FormVitals.class.php
```diff
@@ -337,8 +353,12 @@ public function default_action()
         }
 
         $session = SessionWrapperFactory::getInstance()->getActiveSession();
+        $validationErrors = $session->get('vitals_validation_errors', []);
+        $session->remove('vitals_validation_errors');
```
kojiromike: PHPStan: `Direct $session->remove() is forbidden — it silently fails on read_and_close sessions. Use SessionUtil::unsetSession() instead.`

```suggestion
        $validationErrors = SessionUtil::getSession('vitals_validation_errors', []);
        SessionUtil::unsetSession(['vitals_validation_errors']);
```

(Switching the read on the line above to `SessionUtil::getSession()` for consistency even though phpstan didn't flag it — same `read_and_close` concern applies.)

interface/forms/vitals/save.php
```diff
@@ -26,5 +26,13 @@
...
 
 $c = new C_FormVitals();
-echo $c->default_action_process();
-@formJump();
+$c->default_action_process();
+
+$validationErrors = $session->get('vitals_validation_errors');
+if ($validationErrors !== null && $validationErrors !== []) {
+    $formUrl = OEGlobalsBag::getInstance()->getWebRoot()
+        . '/interface/forms/vitals/new.php?id=' . urlencode($_POST['id'] ?? '');
```
kojiromike: PHPStan reports two errors on this line: `Direct access to $_POST is forbidden` and `Parameter #1 $string of function urlencode expects string, mixed given.`

The form posts `id` from a hidden input bound to `vitals.get_id()` in the twig template — it's the `form_vitals` primary key, an integer. So `urlencode` is doing nothing useful here even before the phpstan error; validate as an int and concatenate directly:

```suggestion
        . '/interface/forms/vitals/new.php?id=' . (int) filter_input(INPUT_POST, 'id', FILTER_VALIDATE_INT);
```

`filter_input(..., FILTER_VALIDATE_INT)` returns `int|false|null`; the `(int)` cast normalizes a missing/invalid id to `0`, which `new.php` already treats as "fresh form" (see `setFormId(0)`).

src/Common/Forms/FormVitals.php
```diff
@@ -467,6 +467,89 @@ public function has_reason_for_column($column)
...
+            'oxygen_flow_rate' => xl('Oxygen Flow Rate'),
+            'inhaled_oxygen_concentration' => xl('Inhaled Oxygen Concentration'),
+            'head_circ' => xl('Head Circumference'),
+            'waist_circ' => xl('Waist Circumference'),
+            'ped_weight_height' => xl('Pediatric Weight Height Percentile'),
+            'ped_bmi' => xl('Pediatric BMI Percentile'),
+            'ped_head_circ' => xl('Pediatric Head Circumference Percentile'),
+        ];
+
+        foreach ($fieldPropertyMap as $fieldName => $property) {
```
kojiromike: Zooming out from the phpstan error one line down — there's a larger structural issue worth a look. `validate()` keeps **three** parallel data structures that all have to stay in sync:

1. `$fieldPropertyMap` (this method, 15 entries) — every entry maps a field name to itself, so it's degenerate; it could be `array_keys($fieldLabels)`.
2. `$fieldLabels` (this method, 15 entries) — display names for the same 15 fields.
3. `VitalsFieldRanges::getRanges()` (the other file, same 15 fields) — bounds for those fields.

If someone adds a new vital to `VitalsFieldRanges` and forgets to add it here, the `$range === null` branch silently skips validation — no error, no warning. That's the kind of drift the type system can't catch and a reviewer will eventually miss.

The cheap fix is to make `VitalsFieldRanges` the single source of truth — add `label` to each entry there, drop both local maps, and iterate. The validate body collapses to:

```php
public function validate(): array
{
    $errors = [];
    $warnings = [];

    foreach (VitalsFieldRanges::getRanges() as $fieldName => $field) {
        $value = $this->$fieldName;
        if ($value === null || $value === '') {
            continue;
        }

        $label = $field['label'];

        if (!is_numeric($value)) {
            $errors[] = sprintf(xl('Field "%1$s" has a non-numeric value.'), $label);
            continue;
        }

        $numericValue = (float) $value;

        if ($numericValue < 0) {
            $errors[] = sprintf(xl('Field "%1$s" cannot be negative.'), $label);
            continue;
        }

        if ($numericValue < $field['min'] || $numericValue > $field['max']) {
            $errors[] = sprintf(xl('Field "%1$s" is outside acceptable range (%2$s–%3$s).'), $label, $field['min'], $field['max']);
            continue;
        }

        if ($numericValue < $field['warningMin'] || $numericValue > $field['warningMax']) {
            $warnings[] = sprintf(xl('Field "%1$s" is outside typical clinical range (%2$s–%3$s).'), $label, $field['warningMin'], $field['warningMax']);
        }
    }

    return ['errors' => $errors, 'warnings' => $warnings];
}
```

This also makes the phpstan complaint on line 522 (`?? $fieldName` is dead) moot — `$field['label']` is always present by the array shape contract.

kojiromike (changes requested):
PHPStan (8.5) failed on `70f3917` with five errors against new code. All real, all in PR code, all fixable. Inline comments cover the four fixes; one of them (on `validate()`) also raises a structural simplification that goes beyond the phpstan error.

src/Common/Forms/FormVitals.php
lilit1121: Done!

interface/forms/vitals/save.php
lilit1121: Done!

interface/forms/vitals/C_FormVitals.class.php
lilit1121: The replacement for remove() has been done, but getSession() function can not be applied as it does not exists in SessionUtil class.

interface/forms/vitals/C_FormVitals.class.php
lilit1121: Done!

kojiromike (approved):

PR #12170, opened 2026-05-20
fix: bypass max_input_vars truncation in edit_layout and edit_payment

### LINKED ISSUE #11494: Batch Payments "Modify Payments" silently fails for patients with many encounters (max_input_vars truncation)
## Problem

The Batch Payments "Modify Payments" form (`interface/billing/edit_payment.php`) renders one set of inputs per encounter row for the selected patient and POSTs them all in a single request. For patients with large numbers of charges/encounters, the request exceeds PHP's `max_input_vars` limit. PHP silently truncates the POST body, the `payment_id` and other required fields are dropped, and the user sees a blank white screen with no payments allocated.

This has been reported by users in the community: https://community.open-emr.org/t/white-screen-on-modify-payments-in-batch-payments-fix-for-large-encounter-lists/26685

## Current state

OpenEMR currently treats this as a configuration prerequisite:

- `Documentation/INSTALL` recommends `max_input_vars` ≥ 3000.
- `setup.php` flags `max_input_vars < 3000` as an error on the health check.
- All shipped Docker `php.ini` files set `max_input_vars = 3000`.

3000 is enough headroom for most installs but is still a finite ceiling. Sites with unusually large patients hit it, and the failure mode (silent POST truncation → blank screen) gives the user no actionable information.

## Why this should be a code fix

Relying on `max_input_vars` as the only line of defense has two problems:

1. The ceiling is arbitrary — there is no patient-size limit at which we can guarantee the form will work.
2. The failure is silent. Even when it fires, nothing surfaces a useful error to the user or to the logs.

## Suggested direction (not prescriptive)

Bound the request size so it cannot exceed `max_input_vars` regardless of how many encounters a patient has. Possible approaches:

- Paginate/batch the encounter rows in the Modify Payments UI.
- Submit allocations incrementally (e.g., per encounter or per page) instead of one giant form post.
- At minimum, detect a truncated POST and return a clear error instead of a blank page.

The implementer should pick the approach that best fits the existing Batch Payments flow.

## Acceptance criteria

- A patient with an arbitrarily large number of open encounters can have payments modified through the Batch Payments → Modify Payments flow without hitting `max_input_vars`.
- If the request is somehow still truncated, the user sees a clear error rather than a blank page.

### LINKED ISSUE #11933: fix(layout-editor): edit_layout.php exceeds max_input_vars and silently corrupts the save
## Problem

The Layout Editor (`interface/super/edit_layout.php`) renders one set of inputs per layout row and POSTs them all in a single form. Layouts with many fields (or many conditions per field) exceed PHP's `max_input_vars` limit, the POST body is silently truncated, and the resulting save corrupts data.

Observed in production logs:

```
PHP Warning:  PHP Request Startup: Input variables exceeded 3000.
  To increase the limit change max_input_vars in php.ini. in Unknown on line 0
  /var/www/.../interface/super/edit_layout.php

PHP Warning:  unserialize(): Error at offset 0 of 3 bytes in
  /var/www/.../interface/super/edit_layout.php on line 945

PHP Warning:  Undefined array key "validation" in
  /var/www/.../interface/super/edit_layout.php on line 472   (× many)
PHP Warning:  Undefined array key "action" in
  /var/www/.../interface/super/edit_layout.php on line 427   (× many)
```

The truncated POST drops `validation` / `action` keys for the trailing rows, and the `conditions` field for rows near the cutoff arrives as a malformed 3-byte fragment that fails `unserialize()`.

This is the same root pattern as #11494 (Batch Payments → Modify Payments), but for a different page. The layout-editor form is more visible because every site admin who edits a long layout (LBF demographics, history, problem list with many fields, etc.) can hit it.

## Why this should be a code fix

- 3000 is an arbitrary ceiling. Some shipped layouts already approach it; sites that customize layouts can exceed it.
- The failure mode is silent corruption (truncated `conditions`, missing keys), not a clean error.

## Suggested direction

Pick whichever fits the existing edit_layout flow:

- Submit row edits incrementally (per row, or chunked) rather than one giant form.
- Detect a truncated POST and refuse to save with a clear error.
- Both.

## Related

- #11494 — same root cause on Batch Payments. Whatever pattern lands there should probably extend to this page.

### LINKED ISSUE #12165: CSRF check missing on interface/billing/edit_payment.php mutating POST branches
## Summary

`interface/billing/edit_payment.php` has no CSRF token check on any of its mutating POST branches, and the in-page form does not render a `csrf_token_form` hidden input. Every other comparable mutating page in the codebase (e.g. `interface/super/edit_layout.php`) calls `CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true)` at the top of each save branch and emits the token in the form. This file is an exception to that pattern.

## Affected mutating branches

- `DeletePaymentDistribution` — deletes an `ar_activity` row (logged as deleted, but the integrity impact is real).
- `ModifyPayments` / `FinishPayments` — re-allocates payment distributions across encounters and closes/finalizes an `ar_session`.

A user with `acct/bill` or `acct/eob` write access (the ACL gate at the top of the file) can reach any of these without a CSRF token currently being validated.

## Why this is being filed as a public issue, not a GHSA

OpenEMR sets `cookie_samesite = Strict` on the core session (`src/Common/Session/SessionConfigurationBuilder.php`). With Strict, a third-party origin cannot cause the browser to attach the session cookie to a cross-origin POST, so the classic cross-site CSRF vector against this page is already blocked at the browser layer regardless of what the server does.

Residual exploitability after `SameSite=Strict`:

- A same-origin XSS could host a request that carries the cookie. But if reflected/stored XSS exists elsewhere in OpenEMR, the attacker already controls the session and CSRF on this page is the least of the user's problems.
- Pre-SameSite browsers (effectively IE11 and very old Safari). Negligible population for a clinical app.
- Cookie-scope mistakes at the deployment level (cookie issued for a parent domain shared with an untrusted subdomain). Not a defect in this file.

A rough CVSS read: `AV:N/AC:H/PR:N/UI:R/S:U/C:N/I:L/A:N` — Low. The fix is defense-in-depth and consistency with the rest of the codebase, not an actively exploitable hole. Filing privately would gate a 4-line, mechanically obvious patch behind the coordinated-disclosure timeline without a corresponding reduction in real-world risk.

## Proposed fix

Three changes in `interface/billing/edit_payment.php`:

1. Import `OpenEMR\Common\Csrf\CsrfUtils`.
2. Add `CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);` at the top of each mutating branch — `DeletePaymentDistribution`, and `ModifyPayments` / `FinishPayments`.
3. Add `<input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken()); ?>" />` to the in-page form.

The only form posting to `edit_payment.php` is the one in the file itself; all other callers (`search_payments.php`, `front_payment.php`, `new_payment.php`) link via GET hyperlinks or a `header("Location: ...")` redirect. No external POST callers to keep compatible with.

## Suggested verification

- Confirm Save and Delete from the payment editor still work in the browser with a valid session.
- Confirm a POST to either mutating branch with a stripped or stale `csrf_token_form` is rejected (`dieOnFail: true` ends the request).

progradedteam-dev:

Two legacy forms — `interface/super/edit_layout.php` (Layout editor save) and `interface/billing/edit_payment.php` (Batch Payments Modify/Finish Payments) — serialise one input subarray per row. Large layouts and patients with many encounters push the input count past PHP's `max_input_vars` ceiling, at which point PHP silently truncates `$_POST` and the trailing rows fall on the floor. The Layout editor saves with missing fields; the Batch Payments user sees a blank white screen.

Per the design-sketch discussion on #11933, this uses server-side raw-body parsing rather than a client-side JSON-blob transport (a11y/o11y concerns).

Closes #11933
Closes #11494
Closes #12165

**New abstraction-layer pair under `OpenEMR\Common\Http`:**

- `RawRequestBodyReader` — wraps `php://input` behind a typed string-returning API; throws `RawPostParserException` on read failure.
- `RawPostParser` — re-parses the raw `application/x-www-form-urlencoded` body and writes the result back into `$_POST` so legacy downstream code (`formData()`, `trimPost()`, `DistributionInsert()`) keeps working without per-call-site changes. Multipart bodies are rejected up front because PHP consumes them before `php://input` is populated.
- `ABSTRACTION_CLASSES` allowlist in `tests/PHPStan/Rules/ForbiddenRequestGlobalsRule.php` is extended to recognise `RawPostParser` as the single sanctioned point of `$_POST` mutation.

**Implementation note worth flagging:** `parse_str()` itself respects `max_input_vars` in PHP 7+, so the naïve single-call parse sketched in the design discussion on #11933 would still truncate (would silently drop variables and emit a runtime warning). `RawPostParser` splits the body into chunks below the configured ceiling, parses each with `parse_str()`, and merges with `array_replace_recursive`. Chunk size is computed at parse time so a runtime ini change is honoured. Documented caveat: `array_replace_recursive` does not append-merge bracket-bare auto-indexed arrays (`foo[]=1&foo[]=2`); both target endpoints use explicit indices (`fld[1][id]`, `Payment3`), so this is not a concern in practice.

**CSRF gap closed (#12165):** `edit_payment.php` had no CSRF check on its `DeletePaymentDistribution`, `ModifyPayments`, or `FinishPayments` handlers. A hidden `csrf_token_form` is now rendered in the shared form, and `CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true)` gates each handler. Bare/invalid tokens get HTTP 400.

**UX polish per review:** Kojiromike's review note that the original sketch's `die()` on a parse-failed save was poor UX is addressed — `edit_layout.php` now sets a `$saveError` variable, the save loop short-circuits when it is set, and the editor body renders a Bootstrap `alert-danger` banner explaining the failure. The user's browser-side edits are not discarded.

**Files changed:**

| File | Purpose |
|---|---|
| `src/Common/Http/RawPostParser.php` (new) | Chunked `parse_str` boundary helper |
| `src/Common/Http/RawRequestBodyReader.php` (new) | Typed `php://input` reader |
| `src/Common/Http/RawPostParserException.php` (new) | Typed exception |
| `tests/Tests/Isolated/Common/Http/RawPostParserIsolatedTest.php` (new) | 9 cases incl. 5000-vars body recovery + chunk-boundary correctness |
| `tests/Tests/Isolated/Common/Http/RawRequestBodyReaderIsolatedTest.php` (new) | 3 cases — read / cache / unreadable-stream |
| `interface/billing/edit_payment.php` | CSRF check + token field + `applyToGlobals()` call |
| `interface/super/edit_layout.php` | `applyToGlobals()` call + in-editor error banner |
| `tests/PHPStan/Rules/ForbiddenRequestGlobalsRule.php` | Add `RawPostParser` to allowlist |
| `.phpstan/baseline/openemr.forbiddenRequestGlobals.php` | edit_layout `$_POST` count 64→63 |
| `.phpstan/baseline/offsetAccess.nonOffsetAccessible.php` | Removed an `int<1, max>` entry made obsolete by improved typing |

**Verification:**

| Check | Result |
|---|---|
| Isolated PHPUnit (`RawPostParser` + `RawRequestBodyReader`) | 12 tests / 27 assertions |
| Full-codebase PHPStan level 10 | clean |
| Rector | clean |
| phpcs / phpcbf | clean |
| Conventional Commits validator | clean |

End-to-end browser test against the Demographics layout: **127 layout rows × 3352 form inputs** with `max_input_vars=3000`. Without the fix, trailing rows are truncated. With the fix, a save of the last row (`related_persons`) persists to `layout_options` as expected. CSRF check verified for all three mutating modes — bare or wrong `csrf_token_form` returns HTTP 400.

**Suggested follow-up (not in this PR):** Per kojiromike's review note about an admin-side healthcheck — when an editor renders, it could compute its field count and compare against `ini_get('max_input_vars')`, surfacing a banner before the user clicks Save if their environment would have truncated under the pre-fix code. Worth a sibling issue.

#### Was an AI assistant used? No

cc @kojiromike @msummers42

### DIFF
--- a/interface/billing/edit_payment.php
+++ b/interface/billing/edit_payment.php
@@ -54,6 +54,9 @@ $screen = 'edit_payment';
 
 $recorder = new Recorder();
 
+/** @var string|null $saveError User-visible error when the body could not be re-parsed. */
+$saveError = null;
+
 // Deletion of payment distribution code
 
 if (isset($_POST["mode"])) {
@@ -89,23 +92,35 @@ if (isset($_POST["mode"])) {
 //Modify Payment Code.
 //===============================================================================
 
-if (isset($_POST["mode"])) {
-    if ($_POST["mode"] == "ModifyPayments" || $_POST["mode"] == "FinishPayments") {
-        CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
+// CSRF + raw-body re-parse for the two mutating modes. Done in a top
+// guard block so a parser failure can fail closed: $saveError is set, the
+// existing write block below short-circuits on it, and the page renders
+// the banner instead of silently committing a partial post.
+$paymentMode = filter_input(INPUT_POST, 'mode');
+if (in_array($paymentMode, ['ModifyPayments', 'FinishPayments'], true)) {
+    CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
+    try {
         // ModifyPayments posts one Payment$CountRow, AdjAmount$CountRow,
-        // etc. set per encounter row. For patients with many encounters this
-        // exceeds PHP's max_input_vars, silently truncating $_POST and
-        // producing a blank page. Re-parse the raw body so all rows survive,
-        // then repopulate $_POST — the per-row code below reads via
-        // formData(), trimPost(), and DistributionInsert(), all of which
-        // pull from the superglobal directly.
-        try {
-            RawPostParser::fromGlobals()->applyToGlobals();
-        } catch (RawPostParserException $e) {
-            ServiceContainer::getLogger()->warning('edit_payment.php: raw POST parse failed', ['exception' => $e]);
-            // Fall through with PHP's truncated $_POST; behaviour matches
-            // the pre-fix state rather than crashing the request.
-        }
+        // etc. set per encounter row. For patients with many encounters
+        // this exceeds PHP's max_input_vars and silently truncates $_POST.
+        // applyToGlobals re-parses the raw body and also rebuilds $_REQUEST
+        // so loop-bound reads (CountIndexAbove/Below) see the full payload.
+        RawPostParser::fromGlobals()->applyToGlobals();
+    } catch (RawPostParserException $e) {
+        $saveError = xl('Save did not complete: the form data could not be re-parsed. Please contact your administrator. No payments were modified.');
+        ServiceContainer::getLogger()->error('raw POST parse failed', [
+            'component' => 'edit_payment',
+            'mode' => $paymentMode,
+            'payment_id' => filter_input(INPUT_POST, 'payment_id', FILTER_VALIDATE_INT),
+            'content_length' => filter_input(INPUT_SERVER, 'CONTENT_LENGTH', FILTER_VALIDATE_INT),
+            'max_input_vars' => filter_var(ini_get('max_input_vars'), FILTER_VALIDATE_INT),
+            'exception' => $e,
+        ]);
+    }
+}
+
+if ($saveError === null && isset($_POST["mode"])) {
+    if ($_POST["mode"] == "ModifyPayments" || $_POST["mode"] == "FinishPayments") {
         $payment_id = $_REQUEST['payment_id'];
         //ar_session Code
         //===============================================================================
@@ -604,6 +619,11 @@ $ResultSearchSub = sqlStatement(
 </head>
 <body class="body_top" onload="OnloadAction()">
     <div class="container-fluid">
+        <?php if ($saveError !== null) { ?>
+            <div class="alert alert-danger m-2" role="alert" id="payment-save-error">
+                <?php echo text($saveError); ?>
+            </div>
+        <?php } ?>
         <?php
         if (($_REQUEST['ParentPage'] ?? '') == 'new_payment') {
             ?>
--- a/interface/billing/edit_payment.php
+++ b/interface/billing/edit_payment.php
@@ -22,8 +22,12 @@
 
 require_once("../globals.php");
 require_once("../../custom/code_types.inc.php");
+use OpenEMR\BC\ServiceContainer;
 use OpenEMR\Common\Acl\AccessDeniedHelper;
 use OpenEMR\Common\Acl\AclMain;
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Http\RawPostParser;
+use OpenEMR\Common\Http\RawPostParserException;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Core\Header;
 use OpenEMR\Core\OEGlobalsBag;
@@ -54,6 +58,7 @@ $recorder = new Recorder();
 
 if (isset($_POST["mode"])) {
     if ($_POST["mode"] == "DeletePaymentDistribution") {
+        CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
         $DeletePaymentDistributionId = (isset($_POST['DeletePaymentDistributionId']) ? trim((string) $_POST['DeletePaymentDistributionId']) : '');
         $DeletePaymentDistributionIdArray = explode('_', $DeletePaymentDistributionId);
         $payment_id = $DeletePaymentDistributionIdArray[0];
@@ -86,6 +91,21 @@ if (isset($_POST["mode"])) {
 
 if (isset($_POST["mode"])) {
     if ($_POST["mode"] == "ModifyPayments" || $_POST["mode"] == "FinishPayments") {
+        CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
+        // ModifyPayments posts one Payment$CountRow, AdjAmount$CountRow,
+        // etc. set per encounter row. For patients with many encounters this
+        // exceeds PHP's max_input_vars, silently truncating $_POST and
+        // producing a blank page. Re-parse the raw body so all rows survive,
+        // then repopulate $_POST — the per-row code below reads via
+        // formData(), trimPost(), and DistributionInsert(), all of which
+        // pull from the superglobal directly.
+        try {
+            RawPostParser::fromGlobals()->applyToGlobals();
+        } catch (RawPostParserException $e) {
+            ServiceContainer::getLogger()->warning('edit_payment.php: raw POST parse failed', ['exception' => $e]);
+            // Fall through with PHP's truncated $_POST; behaviour matches
+            // the pre-fix state rather than crashing the request.
+        }
         $payment_id = $_REQUEST['payment_id'];
         //ar_session Code
         //===============================================================================
@@ -619,6 +639,7 @@ $ResultSearchSub = sqlStatement(
         $onclick = empty($payment_id) ? "top.restoreSession();return SavePayment();" : "return false;";
         ?>
         <form class="form" name='new_payment' method='post' action="edit_payment.php" onsubmit='<?php echo $onclick; ?>'>
+            <input type="hidden" name="csrf_token_form" value="<?php echo attr(CsrfUtils::collectCsrfToken(session: $session)); ?>" />
             <?php
             if (!empty($payment_id)) { ?>
             <fieldset>
--- a/interface/super/edit_layout.php
+++ b/interface/super/edit_layout.php
@@ -452,12 +452,21 @@ if (!empty($_POST['formaction']) && ($_POST['formaction'] == "save") && $layout_
         $parsedPost = RawPostParser::fromGlobals()->applyToGlobals();
     } catch (RawPostParserException $e) {
         $saveError = xl('Save did not complete: the form data could not be re-parsed. Please contact your administrator. Your edits in the browser have not been discarded.');
-        ServiceContainer::getLogger()->warning('edit_layout.php: raw POST parse failed', ['exception' => $e]);
+        ServiceContainer::getLogger()->warning('raw POST parse failed', [
+            'component' => 'edit_layout',
+            'layout_id' => $layout_id,
+            'content_length' => filter_input(INPUT_SERVER, 'CONTENT_LENGTH', FILTER_VALIDATE_INT),
+            'max_input_vars' => filter_var(ini_get('max_input_vars'), FILTER_VALIDATE_INT),
+            'exception' => $e,
+        ]);
     }
     $fldPayload = $parsedPost['fld'] ?? null;
     if ($saveError === null && (!is_array($fldPayload) || $fldPayload === [])) {
         $saveError = xl('Save did not complete: no field rows were submitted. Please contact your administrator.');
-        ServiceContainer::getLogger()->warning('edit_layout.php: save received no fld[] payload');
+        ServiceContainer::getLogger()->warning('save received no fld[] payload', [
+            'component' => 'edit_layout',
+            'layout_id' => $layout_id,
+        ]);
     }
     // If we are saving, then save.
     $fld = is_array($fldPayload) ? $fldPayload : [];
--- a/interface/super/edit_layout.php
+++ b/interface/super/edit_layout.php
@@ -21,9 +21,12 @@ require_once(\OpenEMR\Core\OEGlobalsBag::getInstance()->getSrcDir() . "/layout.i
 /** @var list<int> $typesUsingList */
 /** @var array<string,string> $sources */
 
+use OpenEMR\BC\ServiceContainer;
 use OpenEMR\Common\Acl\AccessDeniedHelper;
 use OpenEMR\Common\Acl\AclMain;
 use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Http\RawPostParser;
+use OpenEMR\Common\Http\RawPostParserException;
 use OpenEMR\Common\Logging\EventAuditLogger;
 use OpenEMR\Common\Session\SessionWrapperFactory;
 use OpenEMR\Core\Header;
@@ -435,11 +438,30 @@ $lbfonly = str_starts_with(attr($layout_id), 'LBF') ? "" : "style='display:none;
 
 // Handle the Form actions
 
+$saveError = null;
 if (!empty($_POST['formaction']) && ($_POST['formaction'] == "save") && $layout_id) {
     CsrfUtils::checkCsrfInput(INPUT_POST, dieOnFail: true);
+    // The save form serialises one fld[N][...] subarray per layout row. For
+    // large layouts this exceeds PHP's max_input_vars, silently truncating
+    // $_POST['fld'] and dropping the tail rows from the save. Re-parse the
+    // raw body so every row survives. On parser failure we keep the
+    // truncated $_POST (current pre-fix behaviour) and surface an error
+    // banner in the editor instead of die()ing on the user.
+    $parsedPost = [];
+    try {
+        $parsedPost = RawPostParser::fromGlobals()->applyToGlobals();
+    } catch (RawPostParserException $e) {
+        $saveError = xl('Save did not complete: the form data could not be re-parsed. Please contact your administrator. Your edits in the browser have not been discarded.');
+        ServiceContainer::getLogger()->warning('edit_layout.php: raw POST parse failed', ['exception' => $e]);
+    }
+    $fldPayload = $parsedPost['fld'] ?? null;
+    if ($saveError === null && (!is_array($fldPayload) || $fldPayload === [])) {
+        $saveError = xl('Save did not complete: no field rows were submitted. Please contact your administrator.');
+        ServiceContainer::getLogger()->warning('edit_layout.php: save received no fld[] payload');
+    }
     // If we are saving, then save.
-    $fld = $_POST['fld'];
-    for ($lino = 1; isset($fld[$lino]['id']); ++$lino) {
+    $fld = is_array($fldPayload) ? $fldPayload : [];
+    for ($lino = 1; $saveError === null && isset($fld[$lino]['id']); ++$lino) {
         $iter = $fld[$lino];
         $field_id = trim((string) $iter['id']);
         $field_id_original = trim((string) $iter['originalid']);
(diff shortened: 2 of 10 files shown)


progradedteam-dev:
@kojiromike could you review this PR?

The failing `codecov/project` check is an upload-artifact: codecov's diff reads `-1.95%` because the head has fewer per-flag uploads than the base (note the `?` markers on the `http`, `inferno`, `phpunit`, and `php8.5` flags; the apache/e2e/api/email/mariadb/services flags all only moved by `-0.01%`). The codecov bot's own summary actually shows ✅ "Project coverage is 22.91%" on the absolute number — only the delta check trips.

`codecov/patch` is at 83.04% with 29 lines uncovered:
- `src/Common/Http/RawPostParser.php` (75%) — the uncovered lines are `applyToGlobals()` and `fromGlobals()`, which mutate `$_POST` / read `$_SERVER`. I deliberately kept the isolated tests free of superglobal mutation; happy to add tests that swap them in `setUp`/`tearDown` if you'd prefer.
- `interface/super/edit_layout.php` (12%) and `interface/billing/edit_payment.php` (0%) — legacy PHP scripts not coverable by unit tests without a full HTTP stack. These lines were exercised manually: the Demographics layout (127 rows / 3352 inputs vs `max_input_vars=3000`) round-trips a save of the trailing `related_persons` row to `layout_options`, and curl POSTs to `edit_payment.php` without/with-wrong `csrf_token_form` return HTTP 400 on all three mutating modes. Browser/curl checks aren't tracked by codecov.

Would you like the `applyToGlobals`/`fromGlobals` tests added, or are you OK merging through the codecov noise?

interface/billing/edit_payment.php
```diff
@@ -86,6 +91,21 @@
...
+        // ModifyPayments posts one Payment$CountRow, AdjAmount$CountRow,
+        // etc. set per encounter row. For patients with many encounters this
+        // exceeds PHP's max_input_vars, silently truncating $_POST and
+        // producing a blank page. Re-parse the raw body so all rows survive,
+        // then repopulate $_POST — the per-row code below reads via
+        // formData(), trimPost(), and DistributionInsert(), all of which
+        // pull from the superglobal directly.
+        try {
+            RawPostParser::fromGlobals()->applyToGlobals();
+        } catch (RawPostParserException $e) {
```
kojiromike: Two concerns on this catch block — one bug, one design call:

**1. The fix is incomplete: `$_REQUEST` isn't repopulated, but the handler reads `$_REQUEST` for security-relevant fields downstream.**

`applyToGlobals()` only writes `$_POST`. But this branch reads `$_REQUEST` at multiple points later:

- L144 `$CountIndexBelow = $_REQUEST['CountIndexBelow'];` — **loop bound** for the per-encounter distribution write
- L143 `$CountIndexAbove = $_REQUEST['CountIndexAbove'];`
- L145 `$hidden_patient_code = $_REQUEST['hidden_patient_code'];`
- L327 `if ($_REQUEST['global_amount'] == 'yes')` and L332 `$_REQUEST['global_reset']` — money disposition fields
- L348 `$payment_id = !empty($payment_id) ? (int) $payment_id : (int) $_REQUEST['payment_id'];`

`$_REQUEST` is built once at request start from PHP's truncated `$_POST` (per `request_order`); it is never re-derived. In the exact scenario this PR exists to fix — a form past `max_input_vars` — `$_REQUEST['CountIndexBelow']` is still the truncated value while the per-row `$_POST["Payment$N"]` data is now fully restored. Net effect: the loop iterates over a truncated row range while the row data is complete → silent partial financial save, looks successful to the user.

Either rebuild `$_REQUEST` after `applyToGlobals()` (something like `$_REQUEST = array_replace($_GET, $_POST, $_COOKIE)` respecting `request_order`), or migrate these reads to `$_POST` in the same diff. A regression test posting > `max_input_vars` pairs and asserting every `CountRow` row was written would pin the invariant.

**2. Silent fall-through on a money endpoint commits a partial mutation.**

The in-line comment says "behaviour matches the pre-fix state rather than crashing" — but pre-fix state *was* the bug (#12165's blank screen). On a financial-mutation endpoint, the alternative (banner + early return, like `edit_layout.php` does) is one extra branch and keeps the failure mode loud. Consider mirroring `edit_layout.php`'s `$saveError` pattern here so a `RawPostParserException` surfaces a visible error rather than a successful-looking-but-truncated post. (CWE-755, OWASP ASVS V5.1.4.)

src/Common/Http/RawPostParser.php
```diff
@@ -0,0 +1,173 @@
...
+
+        $chunkSize = $this->chunkSize();
+        $pairs = explode('&', $raw);
+        $parsed = [];
+        foreach (array_chunk($pairs, $chunkSize) as $chunk) {
+            $chunkParsed = [];
+            parse_str(implode('&', $chunk), $chunkParsed);
+            $parsed = array_replace_recursive($parsed, $chunkParsed);
+        }
+        return $this->cache = $parsed;
```
kojiromike, line 178: Two adjacent failure modes here that the current code doesn't guard against — both produce silently *worse* behavior than the truncation bug being fixed:

**(a) `post_max_size` truncation is invisible.** PHP with `enable_post_data_reading=On` (default) reads the body up to `post_max_size` and discards the rest, then exposes the truncated portion via `php://input`. `parse_str` parses truncated bytes cleanly (the last pair just gets dropped on the floor), no exception fires, the parser silently produces a partial-but-plausible `$_POST`. Pre-PR: oversized POSTs hit a hard ceiling and the user noticed. Post-PR: oversized POSTs to `edit_payment` would silently write a subset of rows.

A cheap check before the final `return`:

```suggestion
        $expected = (int) ($_SERVER['CONTENT_LENGTH'] ?? strlen($raw));
        if (strlen($raw) !== $expected) {
            throw new RawPostParserException(
                sprintf('Raw body truncated: read %d bytes, Content-Length %d', strlen($raw), $expected),
            );
        }
        return $this->cache = $parsed;
```

(`$_SERVER['CONTENT_LENGTH']` read via the same `filter_input(INPUT_SERVER, ...)` pattern you use for `CONTENT_TYPE` if you want to keep the boundary discipline.)

**(b) Empty parsed result silently clobbers `$_POST`.** At the `$raw === ''` branch above, `parse()` returns `[]` and `applyToGlobals()` then assigns `$_POST = []`. If anything in a future request lifecycle has already consumed `php://input` before this branch runs (test bootstrap, middleware, a profiler), the user's submission gets wiped — strictly worse than the truncation it's fixing. Worth a defensive guard in `applyToGlobals()`: refuse to overwrite a non-empty `$_POST` with an empty parsed array, or throw.

src/Common/Http/RawPostParser.php
```diff
@@ -0,0 +1,173 @@
...
+ * @author    Andrew Alanis <progradedteam@gmail.com>
+ * @copyright Copyright (c) 2026 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Common\Http;
+
+class RawPostParser
```
kojiromike: Two notes on the class itself:

**1. The `array_replace_recursive` limitation is documented but the PHPStan allowlist makes it everyone's problem.** Today's two call sites use explicit indices (`fld[1][id]`, `Payment3`) so the `foo[]=1&foo[]=2` divergence doesn't bite. But the `ABSTRACTION_CLASSES` entry now sanctions `RawPostParser` as the org-wide `$_POST` mutator. The next contributor using it on a form with `<input type="checkbox" name="permissions[]">` will ship a regression invisible to any unit test that exercises a body small enough for PHP's native `parse_str` to handle. Two options that would prevent that:

- Detect `[]=`-style auto-indexed keys in `$pairs` and `throw new RawPostParserException('append-merge unsupported; use explicit indices');` — fails loud at the first reuse.
- Or actually reproduce `parse_str`'s append-merge for integer-keyed sub-arrays so the shape matches the native parser.

The first is much cheaper and keeps the abstraction's scope honest.

**2. Mark the class `final`.** `RawPostParser`, `RawRequestBodyReader`, and `RawPostParserException` are all bare `class` declarations. These are boundary classes with no extension contract; `final` is the right default per the immutability/OO guidance in `php-conventions.md`. (`$contentType` and `$reader` are already `readonly` — good — but the class itself should also be `final`.)

src/Common/Http/RawPostParser.php
```diff
@@ -0,0 +1,173 @@
...
+     */
+    public function parse(): array
+    {
+        if ($this->cache !== null) {
+            return $this->cache;
+        }
+
+        if (stripos($this->contentType, 'multipart/form-data') !== false) {
+            throw new RawPostParserException(
+                'RawPostParser cannot read multipart/form-data bodies; '
```
kojiromike: Multi-line `.`-concatenated string literal. Per `php-conventions.md` §"No String Concatenation for Line Wrapping", concatenating literal fragments to fit a column width is a smell — easy to drop a trailing space, hard to diff. This message fits on one line:

```suggestion
            throw new RawPostParserException(
                'RawPostParser cannot read multipart/form-data bodies; PHP consumes them before php://input is available.',
            );
```

Or use a nowdoc if you prefer the multi-line shape.

interface/super/edit_layout.php
```diff
@@ -435,11 +438,30 @@ function encodeModifier($jsonArray)
...
+    // $_POST['fld'] and dropping the tail rows from the save. Re-parse the
+    // raw body so every row survives. On parser failure we keep the
+    // truncated $_POST (current pre-fix behaviour) and surface an error
+    // banner in the editor instead of die()ing on the user.
+    $parsedPost = [];
+    try {
+        $parsedPost = RawPostParser::fromGlobals()->applyToGlobals();
+    } catch (RawPostParserException $e) {
+        $saveError = xl('Save did not complete: the form data could not be re-parsed. Please contact your administrator. Your edits in the browser have not been discarded.');
+        ServiceContainer::getLogger()->warning('edit_layout.php: raw POST parse failed', ['exception' => $e]);
```
kojiromike: Per `coding-conventions.md` §"PSR-3 Logging Context", variable/identifying data goes in the context array; the message stays a static identifier. Filename in the message string breaks log-aggregator grouping (every aggregator groups by message template). Also missing useful operational context here.

```suggestion
        ServiceContainer::getLogger()->warning('raw POST parse failed', [
            'component' => 'edit_layout',
            'layout_id' => $layout_id,
            'content_length' => filter_input(INPUT_SERVER, 'CONTENT_LENGTH', FILTER_VALIDATE_INT),
            'max_input_vars' => filter_var(ini_get('max_input_vars'), FILTER_VALIDATE_INT),
            'exception' => $e,
        ]);
```

Same treatment for the "save received no fld[] payload" log a few lines down.

interface/billing/edit_payment.php
```diff
@@ -86,6 +91,21 @@
...
+            ServiceContainer::getLogger()->warning('edit_payment.php: raw POST parse failed', ['exception' => $e]);
```
kojiromike: Same PSR-3 message/context split as the edit_layout.php sibling log call — static message, identifying data in context. Two additional notes for this one:

- This is a financial-mutation code path that, per the catch comment above, *proceeds to write* a truncated payment. PSR-3 §1.1 reserves `warning` for "exceptional occurrences that are not errors"; a known-truncated billing write is closer to `error`. (Less load-bearing if you switch this path to fail-closed per the other comment.)
- Useful context keys for this endpoint: `payment_id`, `mode`, `content_length`, `max_input_vars`.

```suggestion
            ServiceContainer::getLogger()->error('raw POST parse failed', [
                'component' => 'edit_payment',
                'mode' => filter_input(INPUT_POST, 'mode'),
                'payment_id' => filter_input(INPUT_POST, 'payment_id', FILTER_VALIDATE_INT),
                'content_length' => filter_input(INPUT_SERVER, 'CONTENT_LENGTH', FILTER_VALIDATE_INT),
                'max_input_vars' => filter_var(ini_get('max_input_vars'), FILTER_VALIDATE_INT),
                'exception' => $e,
            ]);
```

src/Common/Http/RawRequestBodyReader.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @throws RawPostParserException if the stream cannot be read.
+     */
+    public function read(): string
+    {
+        if ($this->cache !== null) {
+            return $this->cache;
+        }
+
+        $raw = @file_get_contents($this->streamIdentifier);
```
kojiromike: The `@` suppression swallows the underlying warning, so the thrown exception's message contains only the stream path — no reason. For `php://input` in normal SAPI use this is fine, but the constructor also accepts arbitrary stream identifiers (tempfile / `data://` in tests, `open_basedir`-restricted paths on hardened php.ini installs), and the current shape gives a debugger nothing to work with when those fail.

```suggestion
        $raw = file_get_contents($this->streamIdentifier);
        if ($raw === false) {
            $reason = error_get_last()['message'] ?? 'unknown error';
            throw new RawPostParserException(
                sprintf('Unable to read raw request body from %s: %s', $this->streamIdentifier, $reason),
            );
        }
```

The PHP manual on the [error control operator](https://www.php.net/manual/en/language.operators.errorcontrol.php) explicitly discourages `@` for exactly this case — wanting to know what went wrong.

kojiromike (changes requested):
Thanks! the abstraction is well-tested, the chunked `parse_str` strategy is correct, and adding CSRF coverage on `edit_payment.php`'s three mutating modes is overdue. A few items below are blockers in my read; the rest are smaller standards/hardening notes.

kojiromike (approved):
Round of fixes addresses every concern from the prior review:

- `$_REQUEST` rebuild in `applyToGlobals()` correctly honours `request_order` (with `variables_order` / `'GP'` fallback) and uses `array_replace` left-to-right so later sources win, matching PHP's own builder.
- Fail-closed branch in `edit_payment.php` mirrors `edit_layout.php`'s `$saveError` pattern; severity bumped to `error` for the financial path.
- `post_max_size` truncation guard (`strlen($raw) < $contentLength`) and the bracket-bare key rejection both have isolated tests that fail on revert.
- `applyToGlobals()` anti-clobber guard correctly returns existing `$_POST` unchanged when an earlier layer has already consumed `php://input`.
- All three classes now `final`; `@` suppression dropped with `error_get_last()` reason threaded into the exception; both log call sites switched to PSR-3 message/context split.

Tests grew from 9 → 14 on `RawPostParser`, covering the Content-Length truncation, matching-length success, bracket-bare rejection, `$_REQUEST` rebuild, and anti-clobber paths.

Looks good to me.

PR #12218, opened 2026-05-21
fix(requisition): correct responsible-party rendering and refactor lab helpers

### LINKED ISSUE #8291: Update the requisition display and ...
As one of my former project, this needed some renovation to stop polluting the error log with undefined messages. A feature is added. The feature is to be able to select from multiple lab orders in a single encounter.

juggernautsei:
Fixes #8291 — corrects lab requisition responsible-party rendering and refactors requisition/lab helper data handling for clearer structure, safer typing, print/UI behavior, and static-analysis compliance.

- Refactor lab requisition form (`interface/forms/requisition/new.php`) layout/data flow and print-ready UI
- Improve lab helpers in `library/lab.inc.php` (QueryUtils, normalization, pure presentation helpers, safer IDs)
- Extract pure helpers for coercion, city/name formatting, AOE aggregation, barcode resolution, relationship/billing display
- Update document viewer (`templates/documents/general_view.html`) for DICOM pop-out metadata and valid plain-text embed markup
- Add/expand isolated coverage in `tests/Tests/Isolated/library/LabIncTest.php`
- Address PR review feedback, merge latest `master`, and sync PHPStan baseline with CI regeneration

Yes

AI assistance used on follow-up commits (Warp). Commit trailers include `Assisted-by` / `Co-Authored-By` where applicable.

#### Testing
- `LabIncTest`: 39 tests green locally
- PHPStan clean on branch-changed PHP paths (`lab.inc.php`, `requisition/new.php`, `LabIncTest.php`)
- Applied CI-regenerated PHPStan baseline artifact to clear baseline-stability failure
- CI matrix + Codecov/patch on latest head remain the authority after checks re-run

#### Conversation
https://app.warp.dev/conversation/caff8512-28d3-4c4d-81f1-4e34fd94cb42


interface/forms/requisition/new.php
```diff
@@ -4,7 +4,7 @@
  *  Lab Requisition Form
  *
  * @package   OpenEMR
- * @link      https://www.open-emr.org
+ * @link      http://www.open-emr.org
```
kojiromike: Don't downgrade `https://` to `http://`. Please keep the `https` URL (same applies to `library/lab.inc.php` line 7).

interface/forms/requisition/new.php
```diff
@@ -13,22 +13,14 @@
  */
 
 require_once(__DIR__ . "/../../globals.php");
-
-use OpenEMR\Common\Session\SessionWrapperFactory;
-use OpenEMR\Core\Header;
-use OpenEMR\Core\OEGlobalsBag;
-
-// Hoist legacy `globals.php` locals so PHPStan can see them (#11792 Phase 5).
-$srcdir = OEGlobalsBag::getInstance()->getSrcDir();
```
kojiromike, line 23: This hoist was added for #11792 Phase 5 so PHPStan can resolve `$srcdir` in the `require_once` lines below. Removing the hoist *and* the `OEGlobalsBag` import while keeping `require_once("$srcdir/...")` makes `$srcdir` undefined — PHPStan fails with 6 errors here (`Variable $srcdir might not be defined`). Please restore the hoist (and the `use OpenEMR\Core\OEGlobalsBag;` import), or inline the path:

```suggestion
require_once(OEGlobalsBag::getInstance()->getSrcDir() . "/api.inc.php");
require_once(OEGlobalsBag::getInstance()->getSrcDir() . "/patient.inc.php");
require_once(OEGlobalsBag::getInstance()->getSrcDir() . "/options.inc.php");
require_once(OEGlobalsBag::getInstance()->getSrcDir() . "/lab.inc.php");
```

(and re-add `use OpenEMR\Core\OEGlobalsBag;` near the top).

interface/forms/requisition/new.php
```diff
@@ -37,29 +29,56 @@
 
 global $pid ;
 
-$encounter = $session->get('encounter');
+$encounter = $_SESSION['encounter'];
```
kojiromike: Going from `$session->get('encounter')` back to `$_SESSION['encounter']` is a regression — the session-wrapper abstraction was introduced specifically so this code stops reading the superglobal directly. Please restore the wrapper:

```suggestion
$session = SessionWrapperFactory::getInstance()->getActiveSession();
$encounter = $session->get('encounter');
```

(and re-add `use OpenEMR\Common\Session\SessionWrapperFactory;`).

interface/forms/requisition/new.php
```diff
@@ -244,16 +263,11 @@
...
-                        if (!empty($ins[0]['subscriber_relationship']) && $ins[0]['subscriber_relationship'] == 'child') {
-                            echo xlt("Parent");}
-
-                        ?><br />
-
+                       <div class="pFill">
+                        <?php echo !empty($responsibleParty['name'])        ? text($responsibleParty['name'])        : '/'; ?><br />
+                        <?php echo !empty($responsibleParty['address'])     ? text($responsibleParty['address'])     : '/'; ?><br />
+                        <?php echo !empty($responsibleParty['city_st_zip']) ? text($responsibleParty['city_st_zip']) : '/'; ?><br />
+                        <?php echo !empty($responsibleParty['relationship']) ? text($responsibleParty['relationship']) : '/'; ?><br />
```
kojiromike: `$responsibleParty['relationship']` is populated with `xlt('Client Billing')`, `xlt('Self')`, or `getListItemTitle('sub_relation', ...)` — all of those already return HTML-escaped strings. Wrapping them in `text()` here double-escapes (an apostrophe in a list-item title would render as `&amp;#039;`).

The cleanest fix is to store the raw values in `$responsibleParty` and apply the escaping/translation at the template:

- store `'Client Billing'` (string literal) instead of `xlt('Client Billing')`, then render with `<?php echo xlt($responsibleParty['relationship']); ?>`
- store the raw `subscriber_relationship` code, then render with `<?php echo text(getListItemTitle('sub_relation', $responsibleParty['relationship'])); ?>`

Applies to the assignments at lines 62, 70, and 79.

interface/forms/requisition/new.php
```diff
@@ -306,33 +320,31 @@
...
                <tr style="height:125px">
                    <td style="vertical-align:top; width:400px;">
                        <div class="notes">
-                         <span clas="fs-4"><strong><?php echo xlt('Test Ordered') ?>:</strong></span><br />
-                            <?php echo text($order[2]) . " " . text($order[3]); ?><br />
-                            <?php echo text($order[17]) . " " . text($order[16]); ?><br />
-                            <?php echo text($order[28]) . " " . text($order[29]); ?><br />
+                         <span class="fs-4"><strong><?php echo xlt('Test Ordered') ?>:</strong></span><br />
+                            <?php echo text($order[2]) . ' ' . text($order[3]); ?><br />
+                            <?php echo isset($order[16]) ? text($order[16]) . ' ' . text($order[17] ?? '') : ''; ?><br />
```
kojiromike: The index offsets shifted: this was `text($order[17]) . ' ' . text($order[16])` before and is now `text($order[16]) . ' ' . text($order[17] ?? '')`. With 14 elements per code row (per the new docblock on `getProceduresInfo`), `$order[16]` is the *second* row's `procedure_code` and `$order[17]` is its `procedure_name` — different fields than the old indexing read. Same on line 347 for `$order[30]/[31]` vs the old `$order[28]/[29]`, and on lines 358-359 for `$order[18]/[32]` vs old `$order[18]/[30]`.

Double-check against a real multi-code procedure order that the output rows show what you intended. The deeper problem is that the flat numeric array is brittle — consider making `getProceduresInfo` return an array of associative rows and iterating them here.

library/lab.inc.php
```diff
@@ -4,166 +4,212 @@
...
-
-
-    return $rez;
+/**
+ * Returns the Quest lab provider configuration row.
+ */
+function getLabconfig(): array|false
+{
+    return QueryUtils::querySingleRow(
+        "SELECT recv_app_id, recv_fac_id FROM procedure_providers WHERE name = 'Quest'",
```
kojiromike: Hardcoding `WHERE name = 'Quest'` will silently return no rows for every deployment whose lab provider has a different `name` value in `procedure_providers`. The previous code's TODO was to extend the function to accept a provider — that's the change needed, not pinning to Quest. Suggest:

```suggestion
function getLabconfig(int $ppid): array|false
{
    return QueryUtils::querySingleRow(
        "SELECT recv_app_id, recv_fac_id FROM procedure_providers WHERE ppid = ?",
        [$ppid]
    );
}
```

and update the call in `interface/forms/requisition/new.php` to pass the order's `$lab`.

library/lab.inc.php
```diff
@@ -4,166 +4,212 @@
...
     }
+    return QueryUtils::querySingleRow(
+        "SELECT title, fname, lname, street, city, state, zip, organization, phone
+         FROM users WHERE id = ?",
+        [$parts[1]]
+    );
+}
 
-    return false;
+function formatPhone($phone): string
```
kojiromike: `formatPhone()` is dead code (no call site anywhere in this diff) and duplicates the existing `OpenEMR\Common\ValueObjects\PhoneNumber` value object, which wraps libphonenumber and already provides `formatLocal()`, `formatGlobal()`, `toNational()`, `toE164()`, etc. (plus `TypedPhoneNumber` / `PhoneNumberService`). A raw 7/10-digit regex is strictly worse — it drops extensions, ignores locale, and mangles `+country` numbers. Remove it; if a formatter is needed when this is wired up, use `PhoneNumber::formatLocal()` / `formatGlobal()` rather than a new helper.

kojiromike (changes requested):
Several blockers here — please address before this is ready to merge. The PR is touching two files (`interface/forms/requisition/new.php`, `library/lab.inc.php`) and CI is red on PHPStan, Rector, Styling, PR Title, and codecov/patch.

**Blockers**

1. **PHPStan regression: `$srcdir` is undefined.** The PR removes both the `use OpenEMR\Core\OEGlobalsBag;` import and the `$srcdir = OEGlobalsBag::getInstance()->getSrcDir();` hoist (the comment said it was added for #11792 Phase 5), but the four `require_once("$srcdir/...")` lines below still depend on `$srcdir`. PHPStan correctly fails with six `Variable $srcdir might not be defined` / `Part $srcdir (mixed) of encapsed string cannot be cast to string` errors. The hoist must be restored, or replaced with another mechanism (e.g. `OEGlobalsBag::getInstance()->getSrcDir()` inlined into each `require_once`).
2. **Session abstraction regression.** Replacing `$session->get('encounter')` with `$_SESSION['encounter']` reintroduces direct superglobal access in code that had already been migrated to `SessionWrapperFactory`. Restore the abstraction.
3. **`getLabconfig()` hardcodes `WHERE name = 'Quest'`.** The previous implementation returned the first procedure_providers row (with a TODO to fix when there's more than one provider). Hardcoding `'Quest'` is not the fix — it silently breaks every deployment whose lab provider is not literally named "Quest". This should take the provider id/name as a parameter (`getLabconfig(int $ppid)`), or be selected against `procedure_providers.ppid = $lab` derived from the order being printed.
4. **Double-escaping on responsibleParty render.** In new.php around lines 267-270, `$responsibleParty['relationship']` is set to `xlt('Client Billing')` / `xlt('Self')` / `getListItemTitle('sub_relation', ...)`, all of which already return HTML-escaped strings. Wrapping them again in `text(...)` double-escapes — apostrophes and quotes will display as `&amp;#039;` etc. Either store the un-escaped string in the array and let the template call `text()`/`xlt()`, or remove the `text()` wrap on already-escaped values. Same shape applies if any of `name`/`address`/`city_st_zip` ever gets fed an `xlt()`'d default.

**Other findings**

5. **`@link` URL downgraded from `https://` to `http://`** in both files. Don't downgrade — keep `https`.
6. **PR title fails Conventional Commits.** `refactor lab requisition form and data structures` has no `type(scope):` prefix, which is why the workflow is failing. Suggest `fix(requisition): correct responsible-party rendering` since this closes a UI bug, not pure refactor.
7. **`formatPhone()` duplicates the existing `PhoneNumber` value object and is never called.** It's added to `lab.inc.php` (line 551) but has no call site anywhere in the diff — pure dead code. It also reimplements, more crudely, `OpenEMR\Common\ValueObjects\PhoneNumber` (which wraps libphonenumber and already offers `formatLocal()`, `formatGlobal()`, `toNational()`, `toE164()`, etc.; see also `TypedPhoneNumber` and `PhoneNumberService`). A naïve 7/10-digit regex is strictly worse — no extension handling, no i18n, breaks on `+country` codes. Drop it; if a formatter is ever wired in here, use `PhoneNumber::formatLocal()`/`formatGlobal()`.
8. **Order array index drift.** The render block now reads `$order[16]`, `$order[17]`, `$order[30]`, `$order[31]`, `$order[32]` where the old code read `$order[17]`, `$order[16]`, `$order[28]`, `$order[29]`, `$order[30]`. With 14 elements per code row, the new indices land on different fields than the old ones (e.g. row-2 `procedure_code` vs row-2 `procedure_name`). Please verify against a real multi-code order which field you actually want displayed on each line — the silent index shuffle is the kind of thing the docblock at the top of `getProceduresInfo` papers over rather than catches. Consider returning an array of associative rows rather than a flat numeric blob; the indexing scheme is the root of the bug.
9. **No tests added.** Codecov reports 0% patch coverage across 126 changed lines. The newly-added helpers (`getProcedureBillingType`, `formatPhone`, `getLabconfig`) are all small, pure-ish functions that are straightforward to unit-test in `tests/Tests/Isolated/`. At minimum, the billing-type → responsible-party mapping logic in `new.php` should be extracted into a testable function and covered, since that's the behavioral fix this PR is trying to ship.

juggernautsei:
@kojiromike Thanks for the thorough review. I've addressed all 9 findings in commit 52b79adb7:

**Blockers**

1. **`$srcdir` undefined** — Restored `OEGlobalsBag` import and `$srcdir` hoist so PHPStan can resolve the `require_once` paths.
2. **Session abstraction regression** — Restored `SessionWrapperFactory` usage instead of raw `$_SESSION['encounter']`.
3. **`getLabconfig()` hardcoded `'Quest'`** — Changed signature to `getLabconfig(int $ppid)` and query by `ppid`. Call site now passes `(int) $lab` from the procedure order.
4. **Double-escaping on responsible party** — Extracted a `buildResponsibleParty()` pure function that stores raw values. The template now applies `xlt()` for literal labels (C/P billing) and `text(getListItemTitle(...))` for list codes (T billing) via a `relationship_is_list` flag.

**Other findings**

5. **`@link` URLs** — Restored `https://` in both files.
6. **PR title** — Updated to Conventional Commits format.
7. **`formatPhone()` dead code** — Removed.
8. **Order array index drift** — Refactored `getProceduresInfo()` to return an array of associative rows instead of a flat numeric blob. Template now iterates rows with named keys — no more brittle index math.
9. **Tests** — Added `tests/Tests/Isolated/LabIncTest.php` with 9 tests (28 assertions) covering all `buildResponsibleParty()` branches.

Ready for re-review when you have a chance.

kojiromike:
@coderabbitai review

juggernautsei:
## Ready for re-review

Addressed the requested changes from review https://github.com/openemr/openemr/pull/12218#pullrequestreview-4453330293 and follow-up CI/PHPStan findings.

### Review blockers
1. Restored `OEGlobalsBag` `$srcdir` hoist and `SessionWrapperFactory` (no direct `$_SESSION`)
2. `getLabconfig(int $ppid)` — no longer hardcodes Quest
3. Responsible-party relationship no longer double-escaped (raw values + render-time `xlt()`/`text()`)
4. `@link` kept as `https://`
5. Removed dead `formatPhone()`
6. `getProceduresInfo()` returns associative rows; template iterates them
7. Isolated tests for `buildResponsibleParty()`

### Additional fixes in latest commit
- PHPStan type narrowing for `pid` / `encounter` / `oid` / `provider_id` / `lab_id`
- Restored AOE Q&A via `getProcedureOrderAnswers()`
- `getLabProviders()` normalizes `false` → `null`
- Print-ready official requisition UI redesign
- Dropped fixed PHPStan baseline ignores for this form

@kojiromike please take a final look when you have a chance.

interface/forms/requisition/new.php
```diff
@@ -35,349 +35,835 @@
...
+    $patientDob = !empty($patientData['DOB']) ? oeFormatShortDate($patientData['DOB']) : '';
+    $collectionDate = !empty($firstRow['date_collected'])
+        ? oeFormatDateTime($firstRow['date_collected'])
+        : '';
+    $orderDate = !empty($firstRow['date_ordered'])
+        ? oeFormatDateTime($firstRow['date_ordered'])
+        : '';
+
+    if (!empty($responsibleParty['relationship'])) {
+        $relationshipDisplay = ($responsibleParty['relationship_is_list'] ?? false)
```
kojiromike: `xlt()` here receives a runtime value, so `'Client Billing'` and `'Self'` never reach `contrib/util/language_translations/collectConstants.pl` — it only collects literal `xl*('...')` arguments and explicitly skips `$`-prefixed ones. Those two labels will render untranslated in every locale. The `$billingLabel` `match` five lines below already does this right with literal `xl()` calls; mirroring that shape for the non-list branch would fix it.

tests/Tests/Isolated/LabIncTest.php
```diff
@@ -0,0 +1,186 @@
...
+
+        $this->assertSame('Bob Jones', $result['name']);
+        $this->assertSame('', $result['relationship']);
+        $this->assertTrue($result['relationship_is_list']);
+    }
+
+    /**
+     * Test getProcedureOrderAnswers is defined for requisition AOE rendering.
+     */
+    public function testGetProcedureOrderAnswersFunctionExists(): void
```
kojiromike, line 197: `function_exists()` assertions pass as long as the file parses, so this test cannot fail for any reason `php -l` wouldn't already catch. Meanwhile the four DB-touching helpers this PR adds or rewrites — `getProcedureBillingType()`, `getLabconfig()`, `getProcedureOrderAnswers()`, and `getBarId()`'s orphaned-requisition cleanup branch — have no behavioral coverage at all (Codecov puts `lab.inc.php` at 33% patch). Could these get real tests in a DB-backed suite, or the smoke test be dropped and the gap called out explicitly?

tests/Tests/Isolated/LabIncTest.php
```diff
@@ -0,0 +1,186 @@
...
+ * Unit tests for library/lab.inc.php helper functions.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Tests\Isolated;
```
kojiromike, line 13: Tests for `library/` files live under `tests/Tests/Isolated/library/` with namespace `OpenEMR\Tests\Isolated\library` — see `GlobalFunctionsTest.php`, which also carries `#[Small]`. Could this move to `tests/Tests/Isolated/library/LabIncTest.php` to match?

kojiromike (changes requested):
Several blockers here — please address before this is ready to merge. The PR is touching two files (`interface/forms/requisition/new.php`, `library/lab.inc.php`) and CI is red on PHPStan, Rector, Styling, PR Title, and codecov/patch.

**Blockers**

1. **PHPStan regression: `$srcdir` is undefined.** The PR removes both the `use OpenEMR\Core\OEGlobalsBag;` import and the `$srcdir = OEGlobalsBag::getInstance()->getSrcDir();` hoist (the comment said it was added for #11792 Phase 5), but the four `require_once("$srcdir/...")` lines below still depend on `$srcdir`. PHPStan correctly fails with six `Variable $srcdir might not be defined` / `Part $srcdir (mixed) of encapsed string cannot be cast to string` errors. The hoist must be restored, or replaced with another mechanism (e.g. `OEGlobalsBag::getInstance()->getSrcDir()` inlined into each `require_once`).
2. **Session abstraction regression.** Replacing `$session->get('encounter')` with `$_SESSION['encounter']` reintroduces direct superglobal access in code that had already been migrated to `SessionWrapperFactory`. Restore the abstraction.
3. **`getLabconfig()` hardcodes `WHERE name = 'Quest'`.** The previous implementation returned the first procedure_providers row (with a TODO to fix when there's more than one provider). Hardcoding `'Quest'` is not the fix — it silently breaks every deployment whose lab provider is not literally named "Quest". This should take the provider id/name as a parameter (`getLabconfig(int $ppid)`), or be selected against `procedure_providers.ppid = $lab` derived from the order being printed.
4. **Double-escaping on responsibleParty render.** In new.php around lines 267-270, `$responsibleParty['relationship']` is set to `xlt('Client Billing')` / `xlt('Self')` / `getListItemTitle('sub_relation', ...)`, all of which already return HTML-escaped strings. Wrapping them again in `text(...)` double-escapes — apostrophes and quotes will display as `&amp;#039;` etc. Either store the un-escaped string in the array and let the template call `text()`/`xlt()`, or remove the `text()` wrap on already-escaped values. Same shape applies if any of `name`/`address`/`city_st_zip` ever gets fed an `xlt()`'d default.

**Other findings**

5. **`@link` URL downgraded from `https://` to `http://`** in both files. Don't downgrade — keep `https`.
6. **PR title fails Conventional Commits.** `refactor lab requisition form and data structures` has no `type(scope):` prefix, which is why the workflow is failing. Suggest `fix(requisition): correct responsible-party rendering` since this closes a UI bug, not pure refactor.
7. **`formatPhone()` duplicates the existing `PhoneNumber` value object and is never called.** It's added to `lab.inc.php` (line 551) but has no call site anywhere in the diff — pure dead code. It also reimplements, more crudely, `OpenEMR\Common\ValueObjects\PhoneNumber` (which wraps libphonenumber and already offers `formatLocal()`, `formatGlobal()`, `toNational()`, `toE164()`, etc.; see also `TypedPhoneNumber` and `PhoneNumberService`). A naïve 7/10-digit regex is strictly worse — no extension handling, no i18n, breaks on `+country` codes. Drop it; if a formatter is ever wired in here, use `PhoneNumber::formatLocal()`/`formatGlobal()`.
8. **Order array index drift.** The render block now reads `$order[16]`, `$order[17]`, `$order[30]`, `$order[31]`, `$order[32]` where the old code read `$order[17]`, `$order[16]`, `$order[28]`, `$order[29]`, `$order[30]`. With 14 elements per code row, the new indices land on different fields than the old ones (e.g. row-2 `procedure_code` vs row-2 `procedure_name`). Please verify against a real multi-code order which field you actually want displayed on each line — the silent index shuffle is the kind of thing the docblock at the top of `getProceduresInfo` papers over rather than catches. Consider returning an array of associative rows rather than a flat numeric blob; the indexing scheme is the root of the bug.
9. **No tests added.** Codecov reports 0% patch coverage across 126 changed lines. The newly-added helpers (`getProcedureBillingType`, `formatPhone`, `getLabconfig`) are all small, pure-ish functions that are straightforward to unit-test in `tests/Tests/Isolated/`. At minimum, the billing-type → responsible-party mapping logic in `new.php` should be extracted into a testable function and covered, since that's the behavioral fix this PR is trying to ship.

juggernautsei:
@kojiromike The remaining error has nothing to do with my pull request. 
Job: 95885305350 (Isolated Tests)
Root cause: PHPUnit failures: 15 Twig template rendering tests compared rendered output against stored fixtures and found mismatches (many tests like portal/partial/_nav_icon, portal/login/autologin, calendar views, patient card templates). The test runner failed due to these assertion failures (Rendered output does not match fixture). The log also shows some non-fatal warnings (disk full in FaxDocumentService during media store operations) but the CI job failed because of the PHPUnit assertion failures.
You admin need to fix this

juggernautsei:
## Re-review requested

Addressed the **codecov/patch** gap (was ~34% vs 50% target) by extracting pure requisition/lab presentation helpers into `library/lab.inc.php` and expanding isolated coverage in `tests/Tests/Isolated/LabIncTest.php`.

### What changed
- Coercion helpers: `lab_coerce_non_negative_int`, `lab_coerce_optional_int`, `lab_coerce_provider_id_string`
- Formatting helpers: `lab_format_city_state_zip`, `lab_format_person_name`
- Display strategy helpers: `lab_has_responsible_party`, `lab_relationship_display`, `lab_billing_type_label_key`, `lab_resolve_existing_barcode`
- Aggregation helpers: `lab_collect_aoe_answers`, `lab_normalize_insurance_rows`
- Requisition form now calls these helpers instead of inline logic
- Isolated tests expanded (39 tests / 111 assertions green locally)
- PHPStan clean on changed paths

Please re-review when convenient. Codecov will recompute after CI uploads coverage for this commit.

juggernautsei:
@kojiromike re-review requested when you have a chance.

Codecov/patch was failing at ~34% (target 50%). Latest push extracts pure requisition helpers into `library/lab.inc.php` and expands `LabIncTest` (39 tests green locally; PHPStan clean on changed paths). Formal `RequestReviews` from this account is blocked by permissions, so pinging here instead.

tests/Tests/Isolated/LabIncTest.php
juggernautsei: Moved in the latest commits to `tests/Tests/Isolated/library/LabIncTest.php` with namespace `OpenEMR\Tests\Isolated\library` and `#[Small]`, matching `GlobalFunctionsTest.php`.

tests/Tests/Isolated/LabIncTest.php
juggernautsei: Agreed the `function_exists()` smoke test was weak — removed it.

Coverage focus remains on pure helpers in isolated tests (`buildResponsibleParty`, coercion/formatting, AOE aggregation, barcode resolution, facility-id validation). The DB-touching helpers (`getProcedureBillingType`, `getLabconfig`, `getProcedureOrderAnswers`, `getBarId`) still need a DB-backed suite for behavioral coverage; calling that out explicitly rather than faking it with existence checks.

templates/documents/general_view.html
```diff
@@ -15,10 +15,61 @@
...
+        alert({xlj t='Unable to open document viewer. Please try Download instead.'});
+        return false;
+    }
+
+    if (popsrc.indexOf('as_file=') === -1 && popsrc.indexOf('dicom_frame.php') === -1) {
+        popsrc += (popsrc.endsWith('&') || popsrc.endsWith('?')) ? 'as_file=false' : (popsrc.indexOf('?') === -1 ? '?as_file=false' : '&as_file=false');
+    }
+
+    if ((docMime === 'application/dicom' || docMime === 'application/dicom+zip') && popsrc.indexOf('dicom_frame.php') === -1) {
+        popsrc = root + '/library/dicom_frame.php?web_path=' + encodeURIComponent(popsrc);
```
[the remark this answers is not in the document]
juggernautsei: Addressed: pop-out and inline DICOM iframe now pass `patient_id` and `document_id` (button data attributes + `dicom_frame.php` query params).

templates/documents/general_view.html
```diff
@@ -233,8 +296,11 @@ <h4>{xlt t='Contents'}</h4>
 	</tr>
 	<tr id="DocContents" class="h-100">
 		<td>
-      {if $file->get_mimetype() eq "image/tiff" or $file->get_mimetype() eq "text/plain"}
-			<embed  style="height:84vh; border: 0px" type="{$file->get_mimetype()|attr}" src="{$web_path|attr}as_file=false"></embed>
+      {if $file->get_mimetype() eq "image/tiff"}
+            {* Firefox cannot render TIFF in iframe/embed; use img like core 8.2. *}
+            <img style="max-width:100%; height:auto; max-height:84vh; border:0" alt="{$file->get_name()|attr}" src="{$web_path|attr}as_file=false" />
+            {elseif $file->get_mimetype() eq "text/plain"}
+			<embed  style="height:84vh; border: 0px" type="text/plain" src="{$web_path|attr}as_file=false"></embed>
```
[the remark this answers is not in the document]
juggernautsei: Removed the invalid `</embed>` end tag for the text/plain viewer.

juggernautsei:
## PR status update + feedback addressed

### Current status
- Merged latest `upstream/master` (branch was `CONFLICTING` / `DIRTY`) and resolved the PHPStan baseline conflict.
- Review decision was still **CHANGES_REQUESTED** from @kojiromike on an earlier commit; original blockers were already fixed on branch and remaining open items are addressed below.
- Local verification after merge: `LabIncTest` 39 tests green; PHPStan clean on changed PHP paths.

### Remaining open review items just fixed
1. **Test location** — moved to `tests/Tests/Isolated/library/LabIncTest.php` + `#[Small]`
2. **Weak `function_exists` smoke** — removed; kept pure-helper coverage; called out DB-helper gap explicitly
3. **Document viewer** — pass `patient_id`/`document_id` into `dicom_frame.php`; remove invalid `</embed>`
4. **`getFacilityInfo`** — require exactly `XX_<positive-digit-id>`
5. **City/state/zip punctuation** — use shared formatter so empty locality doesn’t render bare commas

@kojiromike ready for another pass when you have time. Codecov/patch will recompute after CI coverage upload for the new head.

juggernautsei:
Pushed a follow-up to clear the CI **PHPStan baseline stability** failure:

- Analysis itself was not introducing new errors
- Synced `.phpstan/baseline/openemr.noGlobalNsFunctions.php` from the CI-regenerated `phpstan-baseline-php8.5` artifact

Branch paths remain PHPStan-clean locally. Waiting on CI re-run for the new head.

juggernautsei:
@kojiromike — re-review requested on the latest head.

### Why ready for another pass
- Original blockers from your earlier review are addressed on branch (`OEGlobalsBag` paths, session wrapper, `getLabconfig($ppid)`, no double-escaping, https links, associative order rows, tests, etc.)
- Remaining open thread items addressed:
  - `LabIncTest` moved to `tests/Tests/Isolated/library/` with `#[Small]`
  - weak `function_exists()` smoke removed
  - DICOM pop-out/iframe now passes `patient_id` / `document_id`
  - invalid `</embed>` removed
  - `getFacilityInfo()` tightened; city/state/zip formatting cleaned up
- Merged latest `master` (merge conflicts cleared; PR is **MERGEABLE**)
- PHPStan baseline stability synced from CI artifact on latest push
- Branch-changed paths PHPStan-clean locally; isolated lab tests green

Latest head: current PR tip on `refactor/8291-lab-requisition-form`.

Formal `RequestReviews` from this fork account is permission-limited, so pinging you here. Thanks for another look when you have time.

PR #12209, opened 2026-05-21
fix(faxsms): FaxSMS AppDispatch typed session property fatal before initialization

sjpadgett:
The FaxSMS module can throw a fatal error when loading the fax service from `messageUI.php`.
Fixes #12208 
Error:

```text
Typed property OpenEMR\Modules\FaxSMS\Controller\AppDispatch::$_session must not be accessed before initialization

Example stack path:

messageUI.php
AppDispatch::getApiService('fax')
AppDispatch::getServiceInstance('fax')
RCFaxClient->__construct()
RCFaxClient->getCredentials()
AppDispatch->getSetup()
AppDispatch->getSession('authUserID')

The issue is caused by service clients being instantiated through the static service factory path before AppDispatch::__construct() has initialized the typed $_session property.

Because $_session is a typed property, PHP throws a fatal error when getSession() accesses it before initialization.

Expected behavior:

FaxSMS service clients should be able to safely access session-backed setup/credentials regardless of whether they are reached through the normal controller constructor path or the static service factory path.

Proposed fix:

Add a small runtime initializer in AppDispatch that safely initializes request globals, the active session wrapper, crypto service, and default auth message before they are used. Then call it from both __construct() and getSession().

This keeps the fix contained in AppDispatch and avoids changing individual service client constructors.

fix(faxsms): Initialize AppDispatch runtime context before session access

This fixes a fatal error in the FaxSMS module where `AppDispatch::$_session` could be accessed before initialization.

The failure occurs when a service client is created through the static service factory path:

AppDispatch::getApiService()
AppDispatch::getServiceInstance()
new RCFaxClient()
RCFaxClient->getCredentials()
AppDispatch->getSetup()
AppDispatch->getSession()

In that path, AppDispatch::__construct() has not necessarily completed before getSession() is called. Since $_session is a typed property, PHP throws:

Typed property OpenEMR\Modules\FaxSMS\Controller\AppDispatch::$_session must not be accessed before initialization

This PR adds a small initializeRuntimeContext() helper to AppDispatch and calls it from both the constructor and getSession().

The helper initializes:

request/query/post/server/cookie references
active Symfony session wrapper
crypto service
default auth error message

This keeps the fix local to AppDispatch and makes the class safe for both normal controller dispatch and static service factory usage.

No service client behavior is changed.

The relevant source shows $_session is typed/readonly and getSession() directly dereferences it, while the static factory can instantiate service clients before the constructor path initializes the property.


#### Was an AI assistant used? Yes entirely ChatGPT

PR #12224, opened 2026-05-21
fix(faxsms): lazy-init AppDispatch session to survive static factory path

### LINKED ISSUE #12208: bug(faxsms): FaxSMS AppDispatch typed session property fatal before initialization
### Describe the bug
The FaxSMS module can throw a fatal error when loading the fax service from `messageUI.php`.
Only broke in 8.1.0 and master dev.
Error:
```text
Typed property OpenEMR\Modules\FaxSMS\Controller\AppDispatch::$_session must not be accessed before initialization

Example stack path:
messageUI.php
AppDispatch::getApiService('fax')
AppDispatch::getServiceInstance('fax')
RCFaxClient->__construct()
RCFaxClient->getCredentials()
AppDispatch->getSetup()
AppDispatch->getSession('authUserID')

The issue is caused by service clients being instantiated through the static service factory path before AppDispatch::__construct() has initialized the typed $_session property.

Because $_session is a typed property, PHP throws a fatal error when getSession() accesses it before initialization.

Expected behavior:

FaxSMS service clients should be able to safely access session-backed setup/credentials regardless of whether they are reached through the normal controller constructor path or the static service factory path.

Proposed fix:

Add a small runtime initializer in AppDispatch that safely initializes request globals, the active session wrapper, crypto service, and default auth message before they are used. Then call it from both __construct() and getSession().

This keeps the fix contained in AppDispatch and avoids changing individual service client constructors.

### To Reproduce
<!-- if Applicable add screenshots -->

Steps to reproduce the behavior:
Install and enable any module client fax/sms/email
relog in
From Services menu access a service.
### Client configuration

- Browser:
- OpenEMR version: 8.1.0
- Operating system: windows

kojiromike:
## Summary

- Fixes #12208: `Typed property OpenEMR\Modules\FaxSMS\Controller\AppDispatch::$_session must not be accessed before initialization` when reaching a fax/SMS service client through the static factory path (`AppDispatch::getApiService('fax')` → `new RCFaxClient()` → `getCredentials()` → `getSetup()` → `getSession()`).
- Drops `readonly` from `$_session`, makes it nullable, and routes reads through a private `session()` accessor that lazy-initializes via `SessionWrapperFactory`. The normal `__construct()` path still primes the property; the static factory path now resolves it on first access.

Picks up the intent of #12209 (Jerry Padgett's earlier attempt) without the Reflection-based workaround.

## Why this shape

Every service client subclass (`RCFaxClient`, `EtherFaxActions`, `TwilioSMSClient`, `EmailClient`, `ClickatellSMSClient`, `SignalWireClient`) runs `$this->getCredentials()` inside its own constructor *before* calling `parent::__construct()`. They can't call `parent::__construct()` earlier — the parent constructor runs `verifyAcl()` → `dispatchActions()` → `render()` (which `exit`s). So the only contained fix is to make `$_session` resolvable on first read instead of requiring constructor priming.

## Regression test

Added under `tests/Tests/Isolated/Modules/FaxSMS/AppDispatchLazyInitTest.php`. Reproduces the issue's call shape (an anonymous subclass that touches `getSession()` before `parent::__construct()`). Without the fix, the test errors with the exact fatal from the issue. With the fix it passes.

`custom_modules/` PSR-4 prefixes are registered dynamically by the module manager at runtime, so the isolated test suite has no way to autoload `OpenEMR\Modules\FaxSMS\...`. The test bootstraps the module's PSR-4 prefix itself in `setUpBeforeClass()` via the live Composer `ClassLoader` — a one-test workaround that keeps the regression covered until a more general module-autoload bootstrap is wired into the isolated suite.

## Test plan

- [x] `composer phpunit-isolated -- --filter AppDispatchLazyInitTest` passes on the fix
- [x] Same test errors with the exact `Typed property ... must not be accessed before initialization` fatal when the fix is reverted
- [x] PHPStan, PHPCS, Rector, Composer Require Checker (pre-commit) all green
- [x] Manual smoke: enable a FaxSMS service in a clean install, log out/in, hit the service from the Services menu — should no longer fatal

### DIFF
--- a/.phpstan/fatal-baseline-caps.php
+++ b/.phpstan/fatal-baseline-caps.php
@@ -35,7 +35,7 @@ declare(strict_types=1);
 
 return [
     'all' => [
-        'class.notFound.php' => 181,
+        'class.notFound.php' => 180,
         'classConstant.notFound.php' => 0,
         'constant.notFound.php' => 0,
         'function.notFound.php' => 0,
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/AppDispatch.php
@@ -42,7 +42,7 @@ abstract class AppDispatch
     protected $_currentAction;
     protected $credentials;
     private $_request, $_response, $_query, $_post, $_server, $_cookies;
-    private readonly SessionInterface $_session;
+    private ?SessionInterface $_session = null;
     protected $authUser;
 
     /**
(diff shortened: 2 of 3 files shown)


sjpadgett (approved):
Test okay for me, thanks.

PR #12266, opened 2026-05-26
feat(storage): fail fast when CLI runs as non-web user before file writes

bradymiller:
## Summary

Files written by a PHP process running as a different UID than the web server end up with restrictive ownership that apache cannot read later. The result is opaque downstream errors like `PHP Fatal: Smarty error: unable to write to compile_dir` or `Key path … is not readable` — far from the actual mistake ("re-run PHP as the apache user").

This PR adds `WebUserGuard` to surface the mismatch *at the write call site*, with an error message that names both UIDs and the exact command to fix it.

## How it works

The guard compares the current process's effective UID against the owner of `{siteDir}/documents` — a directory OpenEMR guarantees is writable by the web server at runtime (it's where patient documents land). Whoever owns that directory IS the runtime web user, by definition.

Example error message:

```
Web-user mismatch: 'cache directory at /tmp/openemr-smarty' would be done by UID 0,
but '/var/www/localhost/htdocs/openemr/sites/default/documents' (a directory the
web server writes to) is owned by UID 1000. Files this process creates would not
be readable by the web server, which would cause opaque fatal errors later
(unable to write to compile_dir, key path not readable, etc). Re-run PHP as UID
1000 — e.g. via `su -s /bin/sh $(id -un 1000) -c '<your command>'`.
```

## Wired into

- **`CacheDirectory::for()`** — covers `/tmp/openemr-smarty` (Smarty compile) and `/tmp/openemr-mpdf` (mPDF temp), plus any future `CacheDirectory` scope automatically (no per-scope maintenance).
- **`OAuth2KeyConfig::createOrRecreateKeys()`** — covers the OAuth2 private/public key generation under `sites/{site}/documents/certificates`.

## Background — what failures this catches

Reproduced today on master in a dev container running as root: hitting any Smarty-rendered page (Calendar, Practice Settings, New Documents) returns 500 with the compile-dir fatal in the PHP error log. Similarly, hitting any FHIR API endpoint returns 500 with "Key path ... is not readable" when the OAuth keys were generated by a root-running CLI invocation. The same shape applies to production environments where admins run `sudo php sql_upgrade.php` or where cron jobs are misconfigured to run as root.

## Cross-OS safety

- Skipped entirely on non-POSIX systems (Windows uses ACLs, not UID semantics; the failure mode doesn't naturally exist there).
- UID resolution prefers `posix_geteuid()` and falls back to running `id -u` via Symfony Process when the `posix` extension isn't loaded (common in stripped-down PHP CLI builds).
- `posix_geteuid` / `posix_getpwnam` are added to `.composer-require-checker.json`'s symbol whitelist — both are guarded with `function_exists` checks, so `ext-posix` is **not** declared as a hard composer dependency.

## Test plan

- [x] `composer phpstan` — clean
- [x] Isolated suite — 3033/3033 pass (3 new `WebUserGuardTest` cases + existing `CacheDirectoryTest` 21/21 still pass)
- [x] Manual verification end-to-end: as root from CLI, `(new CacheDirectory())->for('openemr-smarty')` now throws with the exact message above; as apache (`su apache -s /bin/sh -c '…'`), it creates the dir cleanly with apache ownership.
- [ ] CI green
- [ ] No regressions in api/services suites (will verify on PR)

## Notes for review

- The fix is intentionally *fail-loud*. There's a softer "auto-chown on root-creation" variant we discussed but decided against — it'd silently paper over a real configuration mistake, both in dev (where surfacing the issue is wanted) and in production (where it'd let admins continue running PHP as root without realizing the implication for future cron-vs-apache interleavings).
- Coverage is currently only the two known-impacted writers (cache dirs + OAuth keys). Other lazy-file-creation paths in OpenEMR may have the same issue and would benefit from the same guard. Out of scope for this PR but the helper is designed to drop into any analogous call site (`WebUserGuard::assertSafe('what you're about to write')`).

### DIFF
--- a/.composer-require-checker.json
+++ b/.composer-require-checker.json
@@ -80,6 +80,8 @@
     "lookup_code_descriptions",
     "note",
     "opcache_invalidate",
+    "posix_geteuid",
+    "posix_getpwnam",
     "privQuery",
     "privStatement",
     "resolve_rules_sql",
--- a/src/Common/Auth/OAuth2KeyConfig.php
+++ b/src/Common/Auth/OAuth2KeyConfig.php
@@ -197,9 +197,18 @@ class OAuth2KeyConfig
         // Fail fast if this process is not running as the web user.
         // Root-owned key files would brick OAuth at runtime: the web
         // server would log "Key path ... does not exist or is not
-        // readable" and return 500. See WebUserGuard for the full
-        // reasoning.
-        WebUserGuard::assertSafe('OAuth2 key generation at ' . $this->privateKey);
+        // readable" and return 500. The reference path is derived from
+        // the explicit private-key location (which already encodes the
+        // configured site dir) rather than $GLOBALS['OE_SITE_DIR'], so
+        // multisite callers that instantiate OAuth2KeyConfig with a
+        // non-default $siteDir validate against the correct site's
+        // documents directory. See WebUserGuard for the full reasoning.
+        WebUserGuard::assertSafe(
+            'OAuth2 key generation at ' . $this->privateKey,
+            // {siteDir}/documents/certificates/oaprivate.key
+            //   -> dirname x2 -> {siteDir}/documents
+            dirname($this->privateKey, 2),
+        );
 
         // Collect info from $this->oauth2KeyMissing to determine what is missing for logging purposes and then reset it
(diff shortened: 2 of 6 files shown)


bradymiller:
closing since going with different strategy: https://github.com/openemr/openemr/pull/12267

PR #743, opened 2017-05-15
Start removing register_globals

robertdown:
Let's remove register_globals!

PR #12267, opened 2026-05-26
feat(cli): refuse to run OpenEMR CLI as root

bradymiller:
## Summary

OpenEMR has no CLI script that legitimately needs root. Running PHP as root from the command line creates files (cache dirs, generated keys, Smarty compile output) owned by UID 0 with restrictive permissions; the web server, running as a non-root user, then cannot read them, and the failure surfaces far from the cause — opaque `PHP Fatal: ... unable to write to compile_dir` deep in template rendering, or `Key path … is not readable` in OAuth, etc.

Sites that run their web server as root have a configuration problem of their own; accommodating them here only hides it.

This PR adds a small `RootCliGuard::assertNotRoot()` and wires it into every CLI entry point that loads or instantiates OpenEMR runtime code. The check itself is intentionally narrow:

- **CLI only** — web requests short-circuit via `PHP_SAPI !== 'cli'`.
- **POSIX only** — Windows uses ACLs; this failure mode doesn't naturally exist there.
- **UID 0 only** — the broader "wrong non-root user" case was the domain of the (now-closed) WebUserGuard #12266; this gate refuses root, full stop.

The error message points at the actionable fix:

```
OpenEMR CLI scripts must not be run as root (UID 0). Doing so creates
files the web server cannot read, leading to opaque fatal errors later
(unable to write to compile_dir, key path not readable, etc). Re-run as
the web user (typically 'apache' or 'www-data') — e.g. `su -s /bin/sh
apache -c '[your command]'`.
```

## Gate sites

Seven entry points, each chosen to be the minimal hook that covers a distinct failure mode:

| Location | Covers |
|---|---|
| `interface/globals.php` (after autoload, behind `PHP_SAPI === 'cli'`) | Everything that bootstraps OpenEMR runtime — `sql_upgrade.php`, `acl_upgrade.php`, `sql_patch.php`, contrib utilities, etc. |
| `bin/console` | Symfony console commands, including the `--skip-globals` minimal-bootstrap path |
| `bin/command-runner` | Dispatched commands like `CreateAPIDocumentation` that write swagger yaml |
| `bin/generate-phpstan-types.php` | PHPStan type-alias generator |
| `bin/get-v-database.php` | DB version extractor (read-only today, gated for policy uniformity) |
| `bin/refresh-reserved-word-supplement.php` | Monthly cron in `.github/workflows/refresh-reserved-word-supplement.yml` |
| `library/classes/Installer.class.php::quick_install()` | Both CLI install entry points (`InstallerAuto.php` in this repo and `auto_configure.php` in openemr-devops) plus the web installer `setup.php` (no-ops via the SAPI check) |

Anything that goes through one of these eventually trips the same `RootCliGuard::assertNotRoot()`.

## PHPUnit carveout

CI runs PHPUnit as root inside the test container, and the non-isolated suite loads `interface/globals.php` for DB-backed tests. The globals.php and Installer gates are wrapped in `if (!defined('PHPUNIT_COMPOSER_INSTALL'))` so PHPUnit's bootstrap and `InstallerTest` exercise the rest of the class without the gate firing. The constant is defined only by PHPUnit's autoloader, never in production. The `RootCliGuardTest` itself calls `assertNotRoot()` directly (bypassing the carveout) so the throw-as-root path stays meaningful.

## Test infrastructure

- `tests/Tests/Services/Background/BackgroundServicesCliIntegrationTest.php` spawns `php bin/console background:services` as a subprocess. The PHPUNIT carveout doesn't reach the subprocess, so the integration test now resolves the web-server user (`apache` on the apache image, `www-data` on nginx + php-fpm) and `su`'s the subprocess into it. UID resolution is done via Symfony Process (`id -u`) — the dev container doesn't ship `ext-posix`.
- `tests/Tests/Isolated/Common/Command/RootCliGuardTest.php` — 3 isolated tests covering throw-as-root, non-root pass, and subprocess-dropped-to-apache.
- `tests/Tests/Isolated/Common/Command/RootCliGuardTest.php` and `BackgroundServicesCliIntegrationTest` both use a Symfony-Process-only UID/user resolution path; the `posix_geteuid()` branch was removed since alpine PHP CLI doesn't ship the extension (the symbol was also dropped from `.composer-require-checker.json`).
- `ci/ciLibrary.source::install_configure()` wraps the `InstallerAuto.php` call in `su -p -s /bin/sh <web user> -c …` because that step runs on both the apache and the nginx (`dev-php-fpm`) matrices and dev-php-fpm doesn't carry the openemr-devops devtools library. A `web_user_in_openemr` helper probes `apache` then `www-data` so the same wrap works across the full matrix.

## Companion PRs (both merged)

- **openemr/openemr-devops#743** (merged) — drops privileges to apache across `devtoolsLibrary.source`, all `fsupgrade-*.sh` upgrade scripts, and the `openemr.sh` entrypoint, using `su-exec` for argv-passthrough. Affects flex, 8.1.1, and binary images.
- **openemr/demo_farm_openemr#109** (merged) — same privilege-drop pattern applied to `demo_build.sh` (resolves apache vs www-data from the existing `$alpineOs` flag).

With #743 landed and the flex image SHA in `docker/development-easy/docker-compose.yml` already bumped via dependabot, this PR's earlier `upgradeOpenEMR()` override in `post_install_upgrade()` is no longer needed — `devtoolsLibrary.source` now drops privileges in `upgradeOpenEMR` directly via `run_php_as_apache` (su-exec under the hood). The override was removed in the latest commit on this branch.

## Audit

- No `composer.json` script transitively loads `interface/globals.php`, so running `composer install` as root in build/CI containers is unaffected.
- All discovered OpenEMR-shipped CLI scripts route through one of the gate sites above. The Installer-class gate also covers openemr-devops's `auto_configure.php`, which has its own entry script but reaches `Installer::quick_install()` the same way `InstallerAuto.php` does.

## Test plan

- [x] PHPStan clean at level 10 (full project, no new baseline entries).
- [x] Isolated suite green on PHP 8.2 / 8.3 / 8.4 / 8.5 / 8.6 (includes RootCliGuardTest + InstallerTest under the PHPUNIT carveout).
- [x] Full matrix green: api / services / e2e / email across apache + nginx + Redis Sentinel mTLS + Upgrade-from-5.0.0 (4 PHP × 4 DB versions × 4 stacks).
- [x] Integration test (`BackgroundServicesCliIntegrationTest`) green via `su -p -s /bin/sh apache -c …` subprocess on apache stacks and `… www-data …` on nginx stacks.
- [x] Manual repro inside the running container:
  - `php bin/console list` as root → throws with full message.
  - `su -s /bin/sh apache -c 'php bin/console list'` → runs successfully.
  - `php sql_upgrade.php --from=…` as root → throws (globals.php anchor).
  - `php contrib/util/installScripts/InstallerAuto.php` as root → throws (Installer-class anchor).
  - Web login page returns 200 with a clean error log (web SAPI unaffected).

## Related

- **openemr/openemr-devops#743** (merged) — devops companion: privilege drops in flex, 8.1.1, binary devtools + openemr.sh + fsupgrade chain via su-exec.
- **openemr/demo_farm_openemr#109** (merged) — demo farm `demo_build.sh` privilege drops.
- **WebUserGuard #12266** (closed) — was a defense-in-depth at write sites; closed in favor of the simpler upstream-only refusal in this PR.

### DIFF
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -218,7 +218,22 @@ post_install_upgrade() {
     # rather than the `VAR=val cmd` form — demoData also references
     # $MYSQL_HOST / $MYSQL_ROOT_PASS directly, and the inline form would
     # scope them only to the prepareVariables call.
-    _exec sh -c 'export MYSQL_HOST=mysql MYSQL_ROOT_PASS=root && . /root/devtoolsLibrary.source && prepareVariables && demoData'
+    #
+    # upgradeOpenEMR() is overridden locally so sql_upgrade.php runs as
+    # apache rather than root. RootCliGuard refuses root for OpenEMR CLI
+    # scripts (the failure it catches: a root-owned write here would
+    # brick apache later). The full fix lives in openemr-devops/devtools
+    # — once devtoolsLibrary.source itself drops privileges, this
+    # override can be removed.
+    _exec sh -c '
+        export MYSQL_HOST=mysql MYSQL_ROOT_PASS=root
+        . /root/devtoolsLibrary.source
+        prepareVariables
+        upgradeOpenEMR() {
+            su -p -s /bin/sh apache -c "php -f /var/www/localhost/htdocs/openemr/sql_upgrade.php -- --from=\"$1\""
+        }
+        demoData
+    '
 
     # demoData drops all tables and reseeds from the 5.0.0 dump (which
     # has its own globals values), so the CI-specific globals that
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -238,23 +238,7 @@ post_install_upgrade() {
     # rather than the `VAR=val cmd` form — demoData also references
     # $MYSQL_HOST / $MYSQL_ROOT_PASS directly, and the inline form would
     # scope them only to the prepareVariables call.
-    #
-    # upgradeOpenEMR() is overridden locally so sql_upgrade.php runs as
-    # apache rather than root. RootCliGuard refuses root for OpenEMR CLI
-    # scripts (the failure it catches: a root-owned write here would
-    # brick apache later). The full fix lives in openemr-devops/devtools
-    # — once devtoolsLibrary.source itself drops privileges, this
-    # override can be removed.
-    # shellcheck disable=SC2016  # $1 is intentionally unexpanded by the outer shell — it refers to the overridden upgradeOpenEMR()'s positional arg, evaluated inside the inner shell that sources devtoolsLibrary.
-    _exec sh -c '
-        export MYSQL_HOST=mysql MYSQL_ROOT_PASS=root
-        . /root/devtoolsLibrary.source
-        prepareVariables
-        upgradeOpenEMR() {
-            su -p -s /bin/sh apache -c "php -f /var/www/localhost/htdocs/openemr/sql_upgrade.php -- --from=\"$1\""
-        }
-        demoData
-    '
+    _exec sh -c 'export MYSQL_HOST=mysql MYSQL_ROOT_PASS=root && . /root/devtoolsLibrary.source && prepareVariables && demoData'
 
     # demoData drops all tables and reseeds from the 5.0.0 dump (which
     # has its own globals values), so the CI-specific globals that
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -182,7 +182,12 @@ install_configure() {
     # we need this to occur on the docker image of OpenEMR, on local systems this may be the same but in the ci engine its inside the docker container
     # until we get the devops repo setup we'll just grab the auto_configure.php script and use that
     _exec sh -c 'curl -v https://raw.githubusercontent.com/openemr/openemr-devops/refs/heads/master/docker/openemr/flex/auto_configure.php > /root/auto_configure.php'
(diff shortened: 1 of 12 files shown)


ci/ciLibrary.source
```diff
@@ -218,7 +238,23 @@ post_install_upgrade() {
...
+    # shellcheck disable=SC2016  # $1 is intentionally unexpanded by the outer shell — it refers to the overridden upgradeOpenEMR()'s positional arg, evaluated inside the inner shell that sources devtoolsLibrary.
+    _exec sh -c '
+        export MYSQL_HOST=mysql MYSQL_ROOT_PASS=root
+        . /root/devtoolsLibrary.source
+        prepareVariables
+        upgradeOpenEMR() {
+            su -p -s /bin/sh apache -c "php -f /var/www/localhost/htdocs/openemr/sql_upgrade.php -- --from=\"$1\""
+        }
+        demoData
+    '
```
bradymiller: after land https://github.com/openemr/openemr-devops/pull/743 and rebuild flex, then will remove this section for ci testing before bringing in this PR

bradymiller:
note will need to rebase this PR to master and fixup the new bin script after the following PR lands: https://github.com/openemr/openemr/pull/12335

PR #12273, opened 2026-05-27
fix(globals): stop forcing E_ALL error reporting and add reporting options

### LINKED ISSUE #12265: bug(user debug): globals.php forces E_ALL and overrides php.ini error reporting
### Describe the bug

`interface/globals.php` currently calls:

```php
error_reporting(E_ALL);

Because globals.php is loaded by most scripts, this forces OpenEMR to ignore the server/admin configured php.ini error reporting policy on nearly every request.

For example, if php.ini is configured to suppress notices/deprecations:

error_reporting = E_ALL & ~E_NOTICE & ~E_DEPRECATED & ~E_STRICT
display_errors = Off
log_errors = On

OpenEMR resets the runtime mask back to E_ALL.

Impact

This can cause notices, warnings, deprecations, user warnings, and user deprecations to be logged or displayed even when the server administrator intentionally disabled those levels in php.ini.

This is especially noisy in production and Windows/XAMPP environments where third-party libraries or filesystem permission checks may emit notices/warnings that the admin has already chosen not to report.

Previous behavior

For years this code narrowed reporting instead of widening it:
Intent with this pattern is to act on existing reporting and mask out noise.

if ($globalsBag->getInt('user_debug', 0) > 1) {
    error_reporting(error_reporting() & ~E_WARNING & ~E_NOTICE & ~E_USER_WARNING & ~E_USER_DEPRECATED);
    ini_set('display_errors', 1);
}

The original intent was to show application errors only when user debugging was enabled, not to force all PHP reporting levels globally
### Expected behavior
Do not call error_reporting(E_ALL) from globals.php.

Default behavior should respect php.ini / server runtime policy. A dedicated user_php_debug global should control explicit PHP display/debug modes:

0: use server defaults
2: display PHP errors only
3: display PHP errors and warnings only
4: display current runtime/php.ini reporting level without changing the reporting mask

This avoids overriding administrator logging policy while still allowing explicit debug modes.



### Client configuration

- Browser:
- OpenEMR version: 8.1.0
- Operating system: windows

sjpadgett:
Fixes: #12265
### Summary

This fixes a regression where `interface/globals.php` forces:
error_reporting(E_ALL);
on nearly every OpenEMR request.

Since globals.php is loaded by most scripts, this overrides the server/admin configured php.ini error reporting policy and can cause notices, warnings, deprecations, user warnings, and user deprecations to be logged or displayed unexpectedly.

Background

The previous long-standing behavior did not widen reporting to E_ALL. It narrowed noisy user-level warnings/deprecations and, when user debugging was enabled, displayed application errors only.

The current error_reporting(E_ALL) pattern changes that behavior by globally enabling all reporting levels regardless of php.ini.

Changes
Removes the unconditional error_reporting(E_ALL) override.
Adds/uses user_php_debug as the dedicated PHP runtime debug option.
Keeps default mode aligned with php.ini / server runtime settings.
Keeps user_debug separate for setup/JavaScript console reporting.
Expected behavior
Default mode does not alter error_reporting() or display_errors.
user_php_debug = 0 Use Server Defaults (Display off).
user_php_debug = 2 Display Errors Only.
user_php_debug = 3 Display Errors and Warnings Only
user_php_debug = 4 Display Current Runtime Reporting

Why

OpenEMR should not silently override administrator error reporting policy from globals.php. Debug display should be explicit and controlled by a PHP-specific global.

#### Was an AI assistant used? Yes `Assisted-by` Claude

<!--
If yes, add an `Assisted-by` trailer to each commit that used AI assistance:

### DIFF
--- a/.phpstan/baseline/offsetAccess.nonOffsetAccessible.php
+++ b/.phpstan/baseline/offsetAccess.nonOffsetAccessible.php
@@ -14212,16 +14212,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-faxsms/setup.php',
 ];
 $ignoreErrors[] = [
-    'message' => '#^Cannot access offset \'appKey\' on mixed\\.$#',
-    'count' => 4,
-    'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-faxsms/setup.php',
-];
-$ignoreErrors[] = [
-    'message' => '#^Cannot access offset \'appSecret\' on mixed\\.$#',
-    'count' => 2,
-    'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-faxsms/setup.php',
-];
-$ignoreErrors[] = [
     'message' => '#^Cannot access offset \'fax_number\' on mixed\\.$#',
     'count' => 1,
     'path' => __DIR__ . '/../../interface/modules/custom_modules/oe-module-faxsms/setup.php',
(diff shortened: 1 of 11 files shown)


sjpadgett:
Still some strangeness to address! CCDA, appt email notifications iffy, notification logs not loading in email panel, pathing issues and so on. Mostly these are one/two line fixes so I'm just including here because should be no surprise, I'm lazy

kojiromike (approved):

sjpadgett:
@kojiromike I wasn't finished as noted with my comment on notifications.

Issue #6945, opened 2023-10-20
Implement ONC 2015 Certification (b)(10) Electronic Health Information Export 


adunsulag:
We need to implement the following ONC 2015 Certification (b)(10) Electronic Health Information Export.
https://www.healthit.gov/test-method/electronic-health-information-export

The requirements for this are as follows:

 (b)(10)(i)(A) - Enable a user to timely create an export file(s) with all
 of a single patient’s electronic health information stored at the time
 of certification by the product, of which the Health IT Module is a part.
 (b)(10)(i)(B) -  A user must be able to execute this capability at any time
           the user chooses and without subsequent developer assistance
           to operate.
(b)(10)(i)(C) - Limit the ability of users who can create export file(s) in at
         least one of these two ways: (1) To a specific set of
         identified users (2) As a system administrative function
         [Feature should implement option 2].
 (b)(10)(i)(D) - The export files(s) created must be electronic and in a
         computable format.
 (b)(10)(i)(E) - The publicly accessible hyperlink of the export’s format must
         be included with the exported file(s).
 (b)(10)(ii)   - Create an export of all the electronic health information
         that can be stored at the time of certification by product of
         which the Health IT Module is a part.

 For purposes of Electronic Health Information the information exported by
 this feature should follow the ONC definition given which is:
  EHI means “electronic protected health information” (ePHI) as defined
  in 45 CFR 160.103 to the extent that it would be included in a
  designated record set as defined in 45 CFR 164.501, regardless of whether
  the group of records are used or maintained by or for a covered entity.
  But EHI does not include psychotherapy notes as defined in 45 CFR
  164.501 or information compiled in reasonable anticipation of, or for
  use in, a civil, criminal, or administrative action or proceeding.

 The regulation text for 45 CFR 160.103 defines "health information" as the
 following:
  Health information means any information, including genetic information,
  whether oral or recorded in any form or medium, that: (1) Is created
  or received by a health care provider, health plan, public health
  authority, employer, life insurer, school or university, or health care
  clearinghouse; and (2) Relates to the past, present, or future physical
  or mental health or condition of an individual; the provision of health
  care to an individual; or the past, present, or future payment for the
  provision of health care to an individual.

 The regulation text for 45 CFR 160.103 defines "protected health information"
 as the following:
  Protected health information means individually identifiable health
  information: 
  (1) Except as provided in paragraph (2) of this definition,
  that is:
      (i) Transmitted by electronic media;
      (ii) Maintained in electronic media; or
[...]

PR #12317, opened 2026-05-29
build: exclude generated EHI schema docs from release archives

kojiromike:
## Summary

- `Documentation/EHI_Export` is ~140M of SchemaSpy-generated HTML, diagrams, and tool jars — generated schema documentation, not application code.
- The EHI exporter reads only two descriptors from it at runtime (`docs/openemr.openemr.xml`, `schemaspy/schemas/openemr.meta.xml`); the rest is human-facing docs that belong on the website, not in the install package.
- Mark the generated output `export-ignore` so release archives and GitHub source downloads drop it, keeping the two runtime files via `-export-ignore` carve-outs. The bundled `Documentation/` tree drops from ~150M to ~10.5M.

The single-`*` globs match immediate children only, leaving the parent directories walkable so the carve-outs survive — a `/**` glob would prune the ancestor directories before `git archive` could re-include the two files.

## Verification

- `git archive` of `Documentation/EHI_Export` now yields exactly the two runtime XMLs.
- `EhiExporter` uses `xmlConfigPath` only to read those two files (constructor + `EhiExporter.php:358-359`); it does not copy the HTML tree into the patient export, so the ONC-certified EHI export feature is unaffected.

## Note

This targets `master`. The packaged release archives are built from the `rel-*` branches, so a backport (e.g. to `rel-810`) is needed for it to shrink a given release's package.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -35,3 +35,16 @@ semgrep.yaml                   export-ignore
 CLAUDE.md                      export-ignore
 DOCKER_README.md               export-ignore
 README-Isolated-Testing.md     export-ignore
+
+# Exclude the very large generated EHI schema documentation (~140M of
+# SchemaSpy HTML, diagrams, jars) from release archives; it is published to
+# the website instead of shipped in the install package. The EHI exporter
+# only reads two descriptors at runtime, so keep just those via single-*
+# globs (which match immediate children only, leaving the parent dirs
+# walkable) plus -export-ignore carve-outs for the two files.
+Documentation/EHI_Export/docs/* export-ignore
+Documentation/EHI_Export/docs/openemr.openemr.xml -export-ignore
+Documentation/EHI_Export/schemaspy/* export-ignore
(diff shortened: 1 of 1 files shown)


adunsulag:
Note that we intentionally made the decision to include these documentation assets in case people needed to customize the ehi export as the URL for the schema / data catalog points to the source system the export came from and NOT the OpenEMR repo.  The original implementation of that was here: https://github.com/openemr/openemr/issues/6945.  In hindsight we probably don't want to support people's custom database changes / extensions but we will need to change the EHI exporter to point to a public website.

I'd suggest changing up the addDocumentationReadme function in the EhiExporter.php to point to the public URL otherwise excluding the EHI docs will force the releases to be out of compliance with ONC.

https://github.com/openemr/openemr/blob/af55a287d35386ab01e5c0ba6880566d6acd6fb0/interface/modules/custom_modules/oe-module-ehi-exporter/src/Services/EhiExporter.php#L744

PR #12335, opened 2026-05-30
feat(phpstan): SqlReservedWordRule + automated MySQL/MariaDB drift detection

bradymiller:
## Summary

Adds a custom PHPStan rule that flags bare (unqualified) reserved-word column references in SQL string literals, plus a scheduled workflow that keeps the rule's reserved-word knowledge in sync with the latest MySQL and MariaDB engines without manual maintenance.

Same class of bug as #12329 (\`ORDER BY rank\`) and #12340 (\`AND system = ?\`) — passes every MariaDB CI leg and MySQL 5.7, fails MySQL 8/8.4/9.3 the moment a test path exercises the code. Both prior fixes for this bug class were one-off discoveries; this PR establishes ongoing automated coverage.

## What this PR ships

| | |
|---|---|
| **Rule** | \`tests/PHPStan/Rules/Sql/SqlReservedWordRule.php\` + 3 collaborator classes |
| **Automation** | \`bin/refresh-reserved-word-supplement.php\` + \`.github/workflows/refresh-reserved-word-supplement.yml\` |
| **Snapshots** | \`tests/PHPStan/Rules/Sql/snapshots/{mysql,mariadb}.tsv\` (versioned audit trail) |
| **Dev dep** | \`phpmyadmin/sql-parser\` ^6.0 |
| **Tests** | 24 hermetic tests via PHPStan \`RuleTestCase\` + PHPUnit \`TestCase\` |

First tested custom rule in \`tests/PHPStan/Rules/\` — pattern reusable for the other 18.

## Why bare vs qualified matters

Empirical test against MySQL 8.4 (verified in throwaway Docker containers):

| Position | Bare `interval` | `p.interval` (qualified) |
|---|---|---|
| `SELECT` | **ERROR 1064** | works |
| `WHERE = ?` | **ERROR 1064** | works |
| `ORDER BY` | **ERROR 1064** | works |
| `UPDATE SET` | **ERROR 1064** | works |

MySQL accepts reserved words as identifiers after a `.` without quoting. The rule skips every qualified form accordingly. An earlier iteration flagged 18 sites that turned out to all be table-qualified; those edits were reverted.

## Architecture

Four collaborators:

- **`ReservedWordRegistry`** — union of `phpmyadmin/sql-parser`'s MySQL + MariaDB Context classes (auto-resolved to the highest-numbered one per engine) plus a `RESERVED_WORD_SUPPLEMENT` constant that covers the gap. **The supplement is refreshed weekly by the workflow below**, so the only manual touch point is reviewing the bot PR when the supplement drifts.
- **`SchemaColumnRegistry`** — walks `sql/database.sql` via the parser and collects every defined identifier. JSON cache with mtime invalidation; atomic write with symlink refusal.
- **`SqlSinkResolver`** — closed list of SQL entry points across `FuncCall` (procedural `sqlStatement`/etc.), `StaticCall` (`QueryUtils::*`), and `MethodCall` (`DatabaseQueryTrait` instance methods).
- **`SqlReservedWordRule`** — walks the parser's parsed `Statement` tree. Position-aware (no false positives on `INTERVAL` in `DATE_ADD`). Skips qualified refs. INSERT column lists use a focused token-stream walk on top of the parsed tree.

The `(reserved ∩ schema-identifier)` AND-gate keeps the rule silent on unrelated SQL.

## Automation

`.github/workflows/refresh-reserved-word-supplement.yml` runs weekly (Mondays 06:00 UTC, master + upstream-repo only):

1. Spins up `mysql:latest` and `mariadb:latest` as service containers
2. Reads MySQL's reserved set from `INFORMATION_SCHEMA.KEYWORDS.RESERVED`
3. Derives MariaDB's reserved set by probing each keyword with `SELECT <word>` (1064 = reserved, 1054 = identifier)
4. Computes the gap vs `phpmyadmin/sql-parser`'s shipped sets
5. Rewrites `RESERVED_WORD_SUPPLEMENT` + commits TSV snapshots
6. Opens a PR if anything changed

The TSV snapshots include engine versions in headers, so each refresh PR's diff is self-attributing — reviewer can see immediately whether new reserveds came from an engine bump or a phpmyadmin update.

The script is dry-run by default; `--write` required to mutate files. CLI-only (PHP_SAPI guard + the existing `bin/.htaccess`). Safe-write via `realpath` validation + symlink refusal + atomic temp-rename.

## Coverage

| Position | Covered |
|---|---|
| SELECT expression list | ✓ |
| FROM table refs | ✓ |
| WHERE / HAVING conditions | ✓ |
| ORDER BY / GROUP BY expressions | ✓ |
| JOIN ON conditions + joined table refs | ✓ |
| UPDATE SET column targets | ✓ |
| DELETE WHERE | ✓ |
| INSERT INTO foo (col, ...) column lists | ✓ |

Known gap: `INSERT INTO foo PARTITION (...) (cols)` — the walker treats the first parens after the table as the column list. No OpenEMR call sites use this.

## Tests

24 hermetic tests (no DB, runs under `composer phpunit-isolated`):

- 21 rule tests covering positive cases, qualified-skip, INSERT column lists (incl. `ON DUPLICATE KEY UPDATE` and SELECT-form exit cases), trait sinks, INSERT scope, etc.
- 3 cache-path tests for `SchemaColumnRegistry`

## Verification

- [x] 24/24 tests pass: `composer phpunit-isolated --filter 'SqlReservedWord|SchemaColumnRegistry'`
- [x] Full PHPStan run on the codebase: zero findings
- [x] Pre-commit hook suite passes on every commit
- [x] Empirical MySQL/MariaDB version matrix verified in throwaway Docker containers
- [x] Refresh script tested end-to-end against live `mysql:latest` + `mariadb:latest`; supplement matches what the cron would compute
- [ ] CI matrix turns green across all MySQL legs

## Related PRs (already merged)

- **#12340** — standalone `system` backtick fix in `ContactTelecomService`. Discovered by an earlier iteration of this rule; split out for backportability to rel-* maintenance branches.
- **#12341** — CI `All Checks Passed` aggregate now ignores `codecov/patch` (was already ignoring `codecov/project`). Unblocks the aggregate gate on both PRs.

## Commits on this branch (squash-merges to one)

20 commits documenting the design evolution — rule rewrite to parser-driven, server-derived supplement, INSERT and trait coverage expansion, security-bot followups, snapshot relocation, review-comment followups. Net diff: rule infrastructure + dev dep + `v_database` 539 (cosmetic schema cleanup of two historical tables).


tests/PHPStan/Rules/Sql/SchemaColumnRegistry.php
```diff
@@ -0,0 +1,125 @@
...
+
+        $identifiers = $this->parse($schemaPath);
+
+        if ($cachePath !== null) {
+            $dir = dirname($cachePath);
+            if (!is_dir($dir)) {
+                @mkdir($dir, 0755, true);
+            }
+            $serialized = "<?php\n\nreturn " . var_export($identifiers, true) . ";\n";
+            @file_put_contents($cachePath, $serialized);
```
kojiromike: `@file_put_contents()` here isn't atomic, and PHPStan runs its analysis across parallel worker processes — each builds its own container and runs `load()`. On a clean CI checkout `tmp-phpstan/sql-schema-identifiers.php` doesn't exist, so every worker misses the cache and races to write the same file at once. A worker that `@include`s a half-written file (line 82) hits a **parse error**, which `@` cannot suppress — that's a hard fatal that aborts the whole PHPStan run, i.e. an intermittent CI failure.

The usual fix is write-to-temp-then-rename (atomic on the same filesystem):

```suggestion
            $serialized = "<?php\n\nreturn " . var_export($identifiers, true) . ";\n";
            $tmp = $cachePath . '.' . getmypid() . '.tmp';
            if (@file_put_contents($tmp, $serialized) !== false) {
                @rename($tmp, $cachePath);
            }
```

Separately: the entire cache path (write, mtime invalidation, stale-cache reload) is untested — the test suite passes `''` to disable caching, so this is the most intricate code in the class with zero coverage. Worth a small test that constructs the registry twice against a temp cache file and asserts the second read uses the cache.

bradymiller:
@aisle-analyzer

bradymiller:
@aisle-analyzer

bradymiller:
ok, I think this is ready for prime time. Basically should avoid these annoying sql reserved words from getting into the codebase anymore (and will also deal with codebase when a new reserved word gets created by mysql or mariadb). And should not require any effort on our part except for the rare accepting of a PR when a new reserved word is found (that is not covered by the phpmyadmin stuff) which is checked weekly via a github action.

PR #12336, opened 2026-05-31
fix(reminders): Numerous bug fixes

### LINKED ISSUE #12307: bug(reminders) Appointment reminder runner fails when no reminders are due and no notifications for recurring events
### Describe the bug
- The Fax/SMS appointment reminder runner can throw a fatal error when the reminder scan completes without processing any appointments inside the active cron notification window.
The failure occurs because AppointmentNotificationRunner::run() only initializes $result inside the appointment processing loop. When there are no pending appointments, or when pending appointments are skipped because they are outside the cron window, $result remains undefined but is still used after the loop for logging and as the method return value.
- Recurring events has never worked. Worth a look!
- pc_appstatus not updated on reminder success.
- Error Insurance company search. [29-May-2026 22:31:04 America/New_York] OpenEMR.CRITICAL: Uncaught exception! {"exception":"[object] (TypeError(code: 0): OpenEMR\\Services\\AddressService::getOneByForeignId(): Argument #1 ($foreignId) must be of type int, string given, called in C:\\xampp\\htdocs\\openemr\\interface\\practice\\ins_search.php on line 255 at C:/xampp/htdocs/openemr/src/Services/AddressService.php:140)
[stacktrace]
#0 C:/xampp/htdocs/openemr/interface/practice/ins_search.php(255): OpenEMR/Services/AddressService->getOneByForeignId('')
#1 {main}
"} []
- Missing getPatientDetails() endpoint for module contact dialog.
- Search/Add save error TypeError: window[opts.callBack.call] is not a function
URL: https://opensourcedemr.com/library/dialog.js?v=82
Line: 759 Column: 51
Error object: "{}"
- Notification logs are not available in Email or SMS service page logs;
- Uncaught Reference Error presents in Flow Board: translations is not defined https://opensourcedemr.com/interface/main/messages/js/reminder_appts.js?v=82:420
-  bug double Send() in mailEmail() and a static call on the instance method getSetup()

Error
PHP Warning: Undefined variable $result in
interface/modules/custom_modules/oe-module-faxsms/src/Notification/AppointmentNotificationRunner.php on line 119
OpenEMR.CRITICAL: Uncaught exception!
Error: Call to a member function toArray() on null
Stack trace points to:
AppointmentNotificationRunner->run()
library/rc_sms_notification.php

Cause
AppointmentNotificationRunner::run() initializes the reminder counters before the loop:
$scanned = count($appointments);
$inWindow = 0;
$sent = 0;
$skippedInvalid = 0;
$failed = 0;

However, the ReminderRunResult object is created only inside the foreach ($appointments as $row) loop after a reminder row is processed. If the loop does not execute, or every appointment is skipped by isInCronWindow(), $result is never created.

The method then unconditionally calls:

...$result->toArray(),

which causes a fatal error because $result is undefined/null.

### To Reproduce

Enable email client and setup credentials
Run Services->Notifications-> Test Email Reminders

### Client configuration

- Browser:
- OpenEMR version: v8.1.0 and dev master
- Operating system:

sjpadgett:
## Summary
Fixes #12307

This PR stabilizes several Fax/SMS appointment reminder, recall, dialog, logging, and insurance search workflows. The changes focus on preventing fatal/runtime errors, improving reminder status handling, restoring missing module support endpoints, and making the user-facing reminder/recall pages more reliable.

## Changes

* Fixed appointment reminder runner handling when no appointments are processed during the active cron notification window.

  * The runner now safely handles empty or skipped reminder batches instead of referencing an uninitialized result value after the scan completes.

* Improved reminder runner result handling and logging.

  * Reminder scans now return and log a consistent result even when no appointment notification is sent.

* Corrected appointment reminder status updates after successful reminder processing.

  * Appointment reminder success now updates the appointment status field as expected.

* Improved recurring appointment reminder handling.

  * Updated reminder processing so recurring events are evaluated more reliably by the reminder scan.

* Fixed insurance company search error caused by invalid address lookup input.

  * Insurance search now guards/casts the foreign ID before calling the address service, preventing type errors when the value is empty or not numeric.

* Added the missing patient details endpoint needed by the module contact dialog.

  * Module contact dialogs can now retrieve patient contact details through the expected endpoint.

* Fixed search/add dialog callback handling.

  * Dialog save callbacks now verify the requested JavaScript callback exists before invoking it, preventing runtime errors when an invalid or missing callback is provided.

* Restored notification log visibility for Email and SMS service pages.

  * Email and SMS service log pages now display their related notification logs as expected.

* Fixed reminder/recall JavaScript translation initialization.

  * The Flow Board reminder page now defines its translation object using the standard JavaScript translation helper, preventing `translations is not defined` runtime errors.

## Why this matters

These fixes address several edge cases where reminder and recall workflows could fail silently or throw fatal errors. They also improve visibility into notification activity, restore missing dialog support, and make the reminder UI safer when filtering or showing empty result sets.

Overall, this PR makes the Fax/SMS reminder workflow more reliable across cron scans, appointment status updates, recurring events, service logs, contact dialogs, and recall filtering.


sjpadgett:
I'm finished.

sjpadgett:
I have trouble running the isolated test on windows. The tests appear to have bugs anyway I'll leave to you guys..
Fixing baseline now. To separate concerns I'll open new PR for the module abstract class and services cleanup. I've been moving a little slow and may take me a couple days to finish so if that's a problem with release schedule let me know.

sjpadgett:
I give up! My local run is
```
sjpad@sjpadgett MINGW32 ~/Documents/openemr (reminders)
$ XDEBUG_MODE=off vendor/bin/phpstan analyze   --memory-limit=8G   -c .phpstan/phpstan.ci.neon   --error-format=table
 4299/4299 [============================] 100%
 [OK] No errors
```
Without line numbers and not able to recreate I can't fix phpstan and this shouldn't be this hard to get a PR in.
If you decide not to merge I recommend fixing the bugs that breaks reminders minimum.

sjpadgett:
I need this so I don't have to deal with conflicts in my next PR for module!

bradymiller:
hi @sjpadgett, agree looks like something amiss. asked claude code to look at this (whom is much smarter than me :) ) and here is the analysis:

**(added by me after the post: sorry for Claude's tone here; i wrongly assumed it would be sycophantic :) )**

---

Took a look at the failing checks — there are three real failures (the two `codecov/*` entries are statistical and on the All-Checks-Passed ignore list). All three look tractable, and one of them is a subtle trap that I think explains why the failures look weirder than the code change itself.

## 1. PHP 8.3 Isolated Tests — eager DB call from constructor (15 errors)

The new constructor body at `src/Notification/AppointmentNotificationRunner.php:71` runs:

```php
$this->backgroundRunning = $this->isBackgroundTask();
```

`isBackgroundTask()` (added in this PR, line 426) instantiates `NotificationTaskManager` and calls `getServiceStatus()`, which this PR also changed to use `QueryUtils::querySingleRow()`. In isolated tests there's no DB, so `QueryUtils::Execute()` is called on `null` and the construction blows up. That cascades through all 15 existing tests that build a runner.

**Suggested fix** — move the call out of the constructor so it's not invoked at object-graph-build time. Cheapest is to make it lazy:

```php
// drop the property assignment from the constructor; then in run():
if (!$this->isBackgroundTask()) {
    // ... per-send logging block
}
```

Or inject the boolean as a constructor parameter so tests can pass `false`. Either way the existing test file (`AppointmentNotificationRunnerTest.php`) shouldn't need changes — the eager DB call at construction time is the bug; the tests are catching it correctly.

## 2. PHPStan (8.5) — baseline files had line endings flipped LF → CRLF (14 errors)

This is the one I think is causing the most confusion, because all 14 reported errors are in files this PR doesn't touch (`pnadmin.php`, `PatientReporter.php`, etc.) — which makes it look like CI is being unreasonable.

Here's what actually happened. Compare these baseline diff lines:

```diff
-    'message' => '#^Binary operation "\\.\\=" between "select\\\\n\\\\t\\\\t\\\\t `patient…" and mixed results in an error\\.$#',
+    'message' => '#^Binary operation "\\.\\=" between "select\\\\r\\\\n\\\\t\\\\t\\\\t …" and mixed results in an error\\.$#',
```

Something in the editing pass flipped `\\n` to `\\r\\n` (CRLF) inside the regex patterns in the baseline files. The PHP source files those baselines suppress errors for still have LF line endings, so PHPStan emits errors with `\n` patterns — but the baseline regexes now look for `\r\n` and no longer match. The suppressions break and the errors surface as "new" failures, even though no one actually changed the underlying source.

It looks like 11 + 1 + 1 + 1 of the 14 are exactly this LF/CRLF flip in entries for files outside the PR's scope.

**Suggested fix** — regenerate the baselines rather than editing them by hand:

```bash
composer phpstan-baseline          # in-place regenerate
# or, for a clean slate:
composer phpstan-baseline-reset
```

(Container equivalents: `openemr-cmd phpstan-generate` / `openemr-cmd phpstan-generate-reset`.)

This rewrites all the baseline files with the correct LF patterns so the suppressions match again. The count reductions in this PR (e.g. `EtherFaxActions.php` going `20 → 17`) look like legit fixes to real errors and will survive the regenerate.

Worth noting: CLAUDE.md says the project policy is to avoid adding baseline entries at all — the long-term direction is to fix the underlying mixed-type concatenations in source. But the regenerate above is the quickest path to green CI.

## 3. Rector PHP Analysis (8.5) — auto-fixable, 1 file

Single suggestion in `src/Controller/AppDispatch.php:686` — Rector's `SimplifyIfElseToTernaryRector` wants to collapse a 5-line `if/else` block into a ternary:

```php
$status = $mail->send() ? xlt("Email successfully sent.") : xlt("Error: Email failed") . ' ' . text($mail->ErrorInfo);
```

Just:

```bash
composer rector-fix       # or: openemr-cmd rector-process
```

Accept the change, commit, push.

## TL;DR

1. Move `isBackgroundTask()` out of the constructor — it triggers a DB call before any test setup can intervene.
2. Run `composer phpstan-baseline-reset` to fix the baseline LF/CRLF flip. Most of the PHPStan errors are in files you never touched; they got broken by hand-editing the baselines.
3. Run `composer rector-fix` for the one ternary suggestion.

None of these are CI being hostile — #1 is a real bug the tests caught, #2 is a Windows/CRLF accident that's a one-command fix, #3 is a sixty-second auto-fix.

sjpadgett:
Thanks @bradymiller I've noticed several source files from master having widows endings. Someone is committing commits w/o git line ending checks. My IDE catches and auto fixes these and this is a new catch for me as this is first I've noted.
The test I will refactor when I refactor for separation of concerns and testability in my next PR.
I try to fix with your recommendation on this one. thanks.

sjpadgett:
arrrg! While my git client will fix line ending looks like when I regen'ed baseline from bash line ending flag isn't honored so my boo boo!

sjpadgett:
Well something has changed dealing with line endings in my repo where I can't commit line ending fixed files. I'm playing wack a mole as my local won't generate a diff so I can commit. You guy will need to help me out.

I gave up on the isolated test as claude or chatgpt couldn't get it easily.
I do understand the reasoning for the failure and points out a bigger issue overall. If background:services run is dispatched from inside the app rather than as a clean separate process, then admin vs background isn't separable by entry point at al; an app-initiated background tick and an interactive use can be live in the same runtime. Which means using a service-state flag to disambiguate them is leaning on exactly the boundary that's fuzziest. So the answer isn't detect the run context more in line with test pass but that the runner shouldn't be detecting context at all. Should be passed in from background task call.

sjpadgett:
Yep tired of this

PR #12398, opened 2026-06-04
refactor: install oe-module-dorn as a Composer dependency

claimrevolution:
## What

Extracts the bundled **DORN** (Diagnostic Ordering Result Network) lab-integration module from the tree and installs it as the standalone Composer package [`claimrevolution/oe-module-dorn`](https://packagist.org/packages/claimrevolution/oe-module-dorn), mirroring the ClaimRev Connect extraction in #11265.

The `openemr/oe-module-installer-plugin` relocates the package into `interface/modules/custom_modules/oe-module-dorn` at `composer install` time, so the module continues to be discovered and loaded exactly as before — no change to how OpenEMR finds it.

## Why

- Lets the module ship and version on its own cadence (Packagist) instead of being tied to the OpenEMR release cycle.
- Resolves a latent fatal on the **8.0.x** patch line: the bundled copy calls `OEGlobalsBag::getKernel()` directly, which exists on flex/master but **not** on 8.0.x, so the module 500s every page when active there. The packaged version resolves the kernel through a version-tolerant helper.

## Changes

- Remove `interface/modules/custom_modules/oe-module-dorn/` (the bundled copy).
- Add `claimrevolution/oe-module-dorn: ^1.0` to `composer.json` (and `composer.lock`).
- Add a `.gitignore` entry for the composer-installed module path, matching the existing `oe-module-claimrev-connect` block.

## Testing

`composer require` + an actual authenticated page load with the module registered and active on:

- `openemr/openemr:flex` (master)
- `openemr/openemr:8.0.0.3` (8.0.x patch line) — confirms the `getKernel` 500 is gone
- `openemr/openemr:7.0.4` (the release where DORN was first bundled)

All three: login `200`, module pages render, no fatals.


kojiromike:
@claimrevolution you also need to add ignores to the other files that do things like phpstan and phpcs. Same as #11265

PR #12484, opened 2026-06-13
feat(faxsms): encrypt unassigned inbound faxes at rest

bradymiller:
## Summary

When the SignalWire fax provider receives an inbound fax it cannot auto-match to a patient, the module currently stages the raw PDF under `sites/<site>/documents/received_faxes/unassigned/`. The patient-matched branch in the same `storeFaxDocument` method already routes through the `Document` class `createDocument` call, which honors the `drive_encryption` global; this PR extends the same encryption-at-rest behavior to the unassigned branch.

- The unassigned write in `FaxDocumentService::storeFaxDocument` wraps the bytes via `encryptForFilesystem` before writing the file.
- The read paths — `FaxDocumentService::assignFaxToPatient` and the SignalWire controller's `viewFaxPdf` / `download` endpoints — decrypt via `decryptFromFilesystem`.
- With `drive_encryption` off, `encryptForFilesystem` is a no-op, so behavior is unchanged.
- Legacy plaintext files left on disk from before this change are detected by `cryptCheckStandard` and returned unchanged, so no data migration is required.
- MIME detection in `assignFaxToPatient` now inspects the decrypted buffer rather than the file path (which would have read the encrypted blob after the write change).
- Stale entries cleaned up in `.phpstan/baseline/` for the two touched files (`$mediaPath` narrowed via `is_string()` at source, eliminating several `mixed` warnings; readfile entries removed since the streaming now goes through the crypto helper).

EtherFax and RingCentral inbound paths are unaffected — they don't go through this unassigned-on-disk staging.

## Test plan

- [ ] With `drive_encryption` on: send an unassigned inbound fax via SignalWire webhook, confirm the on-disk file under `sites/<site>/documents/received_faxes/unassigned/` is ciphertext (not a recognizable PDF header)
- [ ] With `drive_encryption` on: view the unassigned fax via the queue UI's "View" and "Download" buttons, confirm the PDF renders / downloads correctly
- [ ] Assign that unassigned fax to a patient, confirm the resulting patient document opens correctly through the standard Documents UI
- [ ] With `drive_encryption` off: confirm the same flows work and on-disk files remain plaintext
- [ ] Legacy compatibility: drop a known plaintext file into `sites/<site>/documents/received_faxes/unassigned/`, point an existing queue row at it, and confirm view/download/assign all still work

### DIFF
--- a/interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
+++ b/interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
@@ -15,6 +15,7 @@ namespace OpenEMR\Modules\FaxSMS\Controller;
 
 use Document;
 use OpenEMR\BC\ServiceContainer;
+use OpenEMR\Common\Crypto\CryptoGenException;
 use OpenEMR\Common\Crypto\CryptoInterface;
 use OpenEMR\Common\Database\QueryUtils;
 use OpenEMR\Common\Session\SessionWrapperFactory;
@@ -23,6 +24,8 @@ use OpenEMR\Core\OEGlobalsBag;
 use OpenEMR\Modules\FaxSMS\Exception\FaxDocumentException;
 use OpenEMR\Modules\FaxSMS\Exception\FaxNotFoundException;
 use OpenEMR\Services\PhoneNumberService;
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
 
 class FaxDocumentService
 {
@@ -176,9 +179,13 @@ class FaxDocumentService
             // legacy plaintext files left on disk before this change.
             $raw = file_get_contents($mediaPath);
             if ($raw === false) {
-                throw new FaxDocumentException("Failed to read fax media file: {$mediaPath}");
+                throw new FaxDocumentException("Failed to read fax media file");
             }
-            $mediaContent = $this->crypto->decryptFromFilesystem($raw);
+            try {
+                $mediaContent = $this->crypto->decryptFromFilesystem($raw);
+            } catch (CryptoGenException $e) {
+                throw new FaxDocumentException("Failed to decrypt fax media file", 0, $e);
+            }
             $details = json_decode($fax['details_json'] ?? '{}', true);
             $fromNumber = $details['from'] ?? $fax['calling_number'] ?? 'Unknown';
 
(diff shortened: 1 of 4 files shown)


bradymiller:
hi @sjpadgett , here's the PR that previously discussed.

bradymiller:
all copilot concerns were addressed

sjpadgett:
Okay Brady, give me a minute to pull and i'll test.

bradymiller:
thanks @sjpadgett , also have another PR cooking (just couple line changes) regarding filename issue that will likely pop up in an hour or so (when AI assistant figures out some phpstan stuff). I'll  ping you tomorrow on it.

sjpadgett:
@bradymiller I'm thinking perhaps we should go a head and encrypt/decrypt all the fax vendors i.e. etherfax/RC. Seems to keep document interface consistent and establish the pattern for future new vendor services.

sjpadgett:
@bradymiller I'm working out some bugs and general implementation issues. 
[13-Jun-2026 07:42:53 America/New_York] SignalWireClient.assignFax(): ERROR - Fax media file not found: 
[13-Jun-2026 07:43:17 America/New_York] SignalWireClient.getCredentials(): DEBUG - faxNumber after E.164 formatting: +18556921099
[13-Jun-2026 07:43:17 America/New_York] SignalWire initialization error: Class "SignalWire\Rest\Client" not found

I have an ongoing account assigned to me/openemr by the powers to be for testing and maintaining etherfax. I also just created a yrial account for SignalWire. I'll send you the credentials where at min the etherfax account you should archive for future openemr testing.

sjpadgett:
Hi @bradymiller 
I'm running into several structural issues where signalwire isn't close to production in current state. A big hurdle is the SDK. The signalwire SDK package puts us in dependency hell e.g. depends on Twilio SDK <= v6.33.0 and lesser Guzzle version etc. So it'd be best to just manage the API with a Guzzle/HTTP client I had already a part of my restructure for a singular client. Plus I hope to get rid of Twilio and RC SDK and on it goes!
I originally had a separate fax server when Twilio supported fax that would have covered these concerns.
Anyway I'm working on the sdk client replacement now so can test but I'll take another look at this to see if I see anything glaring if you want to merge and then leave with me to follow up.

interface/modules/custom_modules/oe-module-faxsms/src/Controller/FaxDocumentService.php
```diff
@@ -113,9 +120,12 @@ public function storeFaxDocument(
                     'patient_id' => $patientId
                 ];
             } else {
-                // Store in unassigned directory
+                // Store in unassigned directory, honoring the drive_encryption
+                // global the same way Document::createDocument does for the
+                // patient-matched branch above. encryptForFilesystem is a
+                // no-op when drive_encryption is off.
                 $mediaPath = $this->receivedFaxesPath . '/unassigned/' . $filename;
```
sjpadgett, line 127: I originally didn't see a need to use the file system at all and still don't. Back when I was reviewing and helping Sherwin get the new service in before ONC, I remember you had serious concerns. At the time though we knew SignalWire wasn't going to be production ready so the security concern was in reality moot. I had intended to follow up and never got around to it.
The temp file need will go away is my plan but if you feel it is safer merging this to cover us, I'm going to proceed by removing this need. My restructure will be staged with this first up in case I run out of time before new release.

sjpadgett:
@bradymiller 
I've decided to implement two modes of handling fax for signalwire:
- Simple polling setup — best default for new OpenEMR users, no OpenEMR webhook.
- Webhook setup — optional “instant inbound fax” mode, but requires a public OpenEMR endpoint. Harder account setup and limits all our folks that run local. This is also legacy so, stuck.

So this PR will be needed after all and thanks for doing.

bradymiller:
hi @sjpadgett ,
Happy to bring this in. Let me know if you want me to bring this in now. Then I may throw up a concept only PR that applies encryption to files that are pending fax out.

sjpadgett:
@bradymiller Yep, merge and I'll go a head and test though I feel, even in enterprise, I can handle without webhooks via polling. But since I'm 3/4 way there to support both maybe I'll keep if you guys are happy with it's security.

Issue #12469, opened 2026-06-11
Decouple code-to-list-option mappings from the SQL upgrade process


Firehed:
## Problem

`CodeTypeEventsSubscriber` hooks into `SQLUpgradeEvent::EVENT_UPGRADE_POST` to populate `list_options.codes` with SNOMED/CPT4 codes. This creates several problems:

1. **Tied to legacy upgrader** — Moving to Doctrine migrations means these events won't fire (or will require an additional weird integration path)
2. **Runs repeatedly** — Executes on every upgrade, with fragile "shouldUpdate" checks to skip redundant work
3. **Overwrites data** — Uses `SET codes = ...`, risking trampling user customizations or invalidating previous imports
4. **Two sync paths** — Also listens to `CodeTypeInstalledEvent`, duplicating logic

## Goals

- No hook into the SQL upgrade process
- Code type import is a one-time operation; list mappings update automatically
- User data is not at risk of being overwritten
- Downstream consumers (`addCoding()`, FHIR services, etc.) work unchanged

## Implementation Ideas

1) Flip the relationship: remove the code from list_options, add list option references either to the codes table or some new intermediate join table
2) Perform mapping purely in PHP: it appears that fairly few list_option values can even be affected by this upgrader. Cut the path out entirely

@adunsulag maybe you have additional thoughts here? I'm not deeply familiar with how this coding process works or is used downstream, but if I'm following the subscriber's implementation correctly, option 2 is _more or less_ what's already happening!

PR #12503, opened 2026-06-15
refactor(modules): Move CodeTypes module alongside surrounding core code

Firehed:
One sticking point of progressing on Doctrine Migrations is an event hook used exclusively (at least in bundled code) by the code types zend module. It listens for migrations completing and, if certain conditions are met, updates some list_options code types. See #12469.

The simplest path right now is to move the existing tool _from a module into core_ so it sits alongside other code type logic and works on the autoload path. This module does nothing else so it's a pretty small change overall. Long term it should probably all move out but that's a different discussion.

Note: as implemented, this means the code hook always runs post-migration rather than only if the module was enabled, as before. Its internal checks will still guard against goofy behavior; installations that haven't imported any additional codes shouldn't see any change. We _could_ keep this config-gated and basically migrate the "is module enabled" to "should this run", but my understanding of the codes is that if you're using them, you'd basically always want this enabled.

- Move CodeTypesEventSubscriber hook class into core, and remove the surrounding module boilerplate.
- Adds some missing type info, cleaning up a number of PHPStan issues
- Separates the upgrade process itself from the event receiver, which will ease plugging this into Doctrine Migration's _own_ event hooks (this is a little goofy since it's a different interface name requiring the same method with a functionally different signature so it'll take a little coercing)

Yes

## Summary by CodeRabbit

* **Refactor**
  * Migrated the Code Types event handling from a module-based registration to a services-based approach, including updated event subscriber behavior during request initialization.
  * Improved typing and method signatures for the Code Types event subscriber to increase reliability and clarity.
* **Chores**
  * Refreshed PHPStan baseline ignores to match the updated code organization and types.


kojiromike:
@coderabbitai review

src/Services/CodeTypes/CodeTypeEventsSubscriber.php
```diff
@@ -87,27 +90,37 @@ class CodeTypeEventsSubscriber implements EventSubscriberInterface
...
+        };
+
+        $this->updateActivatedMappings($logger);
+
+    }
+
+    /**
+     * @param InternalLogger $logger
+     */
+    public function updateActivatedMappings($logger): void
```
kojiromike, line 116: Now that `updateActivatedMappings()` is split out as a public method and the subscriber runs unconditionally post-migration (rather than gated on the module being enabled), consider adding a test that exercises this path.

kojiromike (commented):
comment inline

adunsulag (approved):
Other than @kojiromike's comment, I see no problems with this PR.

PR #12510, opened 2026-06-16
refactor(security): Remove unused client SSL certificate auth feature

Firehed:
## Summary

- Remove the SSL Certificate Administration page and related certificate generation code
- Remove global settings for client SSL certificate auth
- Remove menu entry for Certificates under Admin -> System
- Stop capturing $_SERVER['SSL_CLIENT_S_DN_CN'] in audit logs (always write empty string)
- Clean up PHPStan baseline entries for deleted files

This has **absolutely no effect** on server certificates. When someone is talking about SSL/TLS certificates, they're talking about server certs, not client certs. Any production instance of OE absolutely needs to be served over https; this change has nothing to do with that.

If you're one of the few people that even knows what client certificates are, you probably already know how to put an appropriate proxy in front of OE. Nobody uses them in practice because the browser UX is effectively unusable.

For all intents and purposes, passkeys/WebAuthn have solved this problem in a much better way. Adding them to OE is a good idea, but well out of scope here.

## Background

The client certificate authentication feature was added in February 2010 citing ATNA compliance. However, it was never actually integrated with OpenEMR's authentication system:

- The ssl_certificates_admin.php page could generate certificates
- The audit logger captured the certificate CN if Apache set SSL_CLIENT_S_DN_CN
- **But no code in the login flow ever checked or validated client certificates** (in theory Apache did, but it also issued the key material in the first place, so the whole key management model was pretty dicey; it also required plaintext CA keys to issue them)

The feature also broke almost immediately: in April 2010 (less than 2 months after it was added), globals moved to a database-backed system, but ssl_certificates_admin.php still wrote to the old interface/globals.php file. The Save Certificate Settings button appeared to work but had no effect.

Also, even if it did work, this data was _not used by ATNA logging at all_. Not now, nor historically. IDK if this was a bug, a mis-statement, or something else, but it never did what it claimed to do.

The `crt_user` column in the log table is left in place for backward compatibility with existing audit data, but is now always written as empty string.

The same removed UI pages would also generate self-signed SSL certs and provide instructions for connecting them into Apache. That is absolutely not something that OpenEMR itself ever should have been doing even then, and certainly not now. There are very few legitimate use-cases of self-signed certs, and anyone that has one of those use-cases would know how to do the cert generation. Everyone else should be using LetsEncrypt (or equivalent) or a managed service provider to get properly-signed certs with up-to-date security settings.

## Summary by CodeRabbit

* **Removals**
  * Removed SSL certificate administration functionality (generate/manage/download).
  * Removed client SSL certificate authentication support and related configuration.
  * Removed **Admin → System → Certificates** menu item.
* **Audit Logging**
  * Updated audit logging behavior to stop recording the SSL client certificate common name value.
* **Tests / Quality**
  * Updated E2E, unit, and isolated logging expectations to match the new audit behavior.
  * Adjusted static analysis baselines to reflect configuration and diagnostic changes.


bradymiller:
hi @Firehed , Remember in distant past adding devtools/openemr-cmd to test this thing. Rec checking out code related to this at following devtool/openemr-cmd functions to see how it worked (if removing this in codebase, then would also wipe these calls from devtools/openemr-cmd):
```
  scc, setup-client-cert             Turn on client based cert with designated package
  lcc, list-client-certs             To list the certificate packages
  pcc, put-client-cert               To copy certificate package into the docker
```

Firehed:
Will check that out and most likely send a corresponding pull to devops. Thanks!

PR #12519, opened 2026-06-17
fix(calendar): restore appointments hidden after 8.0→8.1 upgrade

### LINKED ISSUE #12518: Calendar appointments disappear after 8.1.0 upgrade: pc_endDate '0000-00-00'→NULL conversion silently no-ops due to migration statement ordering
### Describe the bug

The 8.0.0 → 8.1.0 upgrade silently fails to convert pre-existing `openemr_postcalendar_events.pc_endDate` values of `'0000-00-00'` to `NULL`. As a result, **every open-ended (non-recurring) appointment created before the upgrade disappears from the calendar** after upgrading, even though the rows still exist and appear in reports.

The migration completes successfully with no error, so the data loss is invisible until a user notices appointments missing from historical/future calendar dates.

### Root cause

In `sql/8_0_0-to-8_1_0_upgrade.sql`, the statements run in this order:

```sql
SET @currentSQLMode = (SELECT @@sql_mode);
SET sql_mode = '';
DELETE FROM `openemr_postcalendar_events` WHERE `pc_eventDate` = '0000-00-00';
UPDATE `openemr_postcalendar_events` SET `pc_endDate` = NULL WHERE `pc_endDate` = '0000-00-00';   -- (A)
SET sql_mode = @currentSQLMode;

ALTER TABLE `openemr_postcalendar_events` MODIFY `pc_eventDate` date NOT NULL;

#IfNotColumnTypeDefault openemr_postcalendar_events pc_endDate date NULL
ALTER TABLE `openemr_postcalendar_events` MODIFY `pc_endDate` date DEFAULT NULL;                   -- (B)
#EndIf
```

The UPDATE at **(A)** runs *before* the ALTER at **(B)**. In the 8.0.0 baseline schema the column is:

```sql
`pc_endDate` date NOT NULL default '0000-00-00',
```

So at the moment (A) executes, `pc_endDate` is still `NOT NULL`. Because (A) runs under `sql_mode = ''`, assigning `NULL` to a `NOT NULL DATE` column does **not** raise an error — MySQL/MariaDB silently coerces the `NULL` back to the implicit date default `'0000-00-00'`. The UPDATE reports rows matched/changed and "succeeds," but the values on disk are unchanged.

The ALTER at (B) then makes the column nullable — but too late; the data was already coerced one step earlier. Net result: the column ends up correctly `DEFAULT NULL`, the version is stamped 538, the job exits 0, yet every legacy row still holds `'0000-00-00'`.

The `sql_mode = ''` line is itself necessary (it lets the zero-date literals through under `NO_ZERO_DATE`), but combined with the ordering it converts what should be a hard error into a silent no-op.

### Why this hides the appointments

The calendar event query in `interface/main/calendar/modules/PostCalendar/pnuserapi.php` was updated for 8.1.0 to treat an open-ended event as `pc_endDate IS NULL`:

```sql
-- 8.1.0 (pcQueryEvents)
AND ((a.pc_endDate >= '$start' AND a.pc_eventDate <= '$end')
  OR (a.pc_endDate IS NULL AND a.pc_eventDate >= '$start' AND a.pc_eventDate <= '$end'))
```

(8.0.0 matched `a.pc_endDate = '0000-00-00'` here instead.)

A leftover row with `pc_endDate = '0000-00-00'` satisfies **neither** branch: `'0000-00-00' >= '$start'` is false, and it is not `NULL`. So the appointment is silently filtered out of every day/week/month view. Recurring events (which carry a real end date) and any appointment created after the upgrade (`NULL`) are unaffected — which is why the loss looks selective.

### To Reproduce

1. On an 8.0.0 instance, create a normal non-recurring appointment (stored with `pc_endDate = '0000-00-00'`).
2. Upgrade to 8.1.0 (`sql_upgrade.php`).
3. Open the calendar on that appointment's date.

**Result:** the appointment is missing from the calendar, though it still exists in the DB and shows in the Appointments report.

Confirming query (use `pc_endDate + 0 = 0`; a literal `'0000-00-00'` in a strict client session throws `ERROR 1525`):

```sql
SELECT COUNT(*) FROM openemr_postcalendar_events WHERE pc_endDate + 0 = 0;
-- > 0 after upgrade == affected
```

### Expected behavior

After the upgrade, all pre-existing open-ended appointments have `pc_endDate = NULL` and remain visible on the calendar.

### Suggested fix

Reorder so the column is made nullable **before** the data is converted (and, ideally, don't rely on `sql_mode = ''` for the value change):

```sql
-- 1. Make the column nullable FIRST
#IfNotColumnTypeDefault openemr_postcalendar_events pc_endDate date NULL
ALTER TABLE `openemr_postcalendar_events` MODIFY `pc_endDate` date DEFAULT NULL;
#EndIf

-- 2. THEN convert the sentinel; now the NULL persists instead of coercing back
SET @currentSQLMode = (SELECT @@sql_mode);
SET sql_mode = '';
DELETE FROM `openemr_postcalendar_events` WHERE `pc_eventDate` = '0000-00-00';
UPDATE `openemr_postcalendar_events` SET `pc_endDate` = NULL WHERE `pc_endDate` = '0000-00-00';
SET sql_mode = @currentSQLMode;

ALTER TABLE `openemr_postcalendar_events` MODIFY `pc_eventDate` date NOT NULL;
```

### Remediation for already-upgraded instances

Because these instances are already at `v_database = 538` (column already nullable), re-running just the data conversion now persists correctly and restores visibility immediately (no schema change, no restart):

```sql
SET SESSION sql_mode = '';
UPDATE openemr_postcalendar_events SET pc_endDate = NULL WHERE pc_endDate = '0000-00-00';
```

### Additional context

The upgrade file (current `master`) still has the original ordering. This is likely related to #11737 (leftover `'0000-00-00'` values surfacing as `-0001` years via unguarded `new DateTime()`), another symptom of the same incomplete cleanup from #11179. The migration's clean exit and the `sql_mode = ''` coercion are what keep this from being noticed during the upgrade itself.

### OpenEMR Version

8.1.0 (`v_database` 538), upgraded from 8.0.0 (`v_database` 531).

kojiromike:
## Summary

The 8.0.0→8.1.0 migration silently failed to convert `pc_endDate = '0000-00-00'` to `NULL` for pre-existing appointments, causing every open-ended appointment created before the upgrade to disappear from all calendar views after upgrading to 8.1.0.

**Root cause (already filed as #12518):** The migration ran the `UPDATE pc_endDate = NULL` *before* the `ALTER TABLE ... MODIFY pc_endDate ... DEFAULT NULL`. While the column was still `NOT NULL`, MySQL/MariaDB under `sql_mode = ''` silently coerces `NULL` to the implicit default `'0000-00-00'`. The UPDATE matched rows and reported success, but no values changed on disk. The subsequent ALTER made the column nullable, but by then the data was already coerced.

**Why appointments disappear:** The 8.1.0 calendar query was updated to match open-ended events with `pc_endDate IS NULL`. A leftover `'0000-00-00'` value satisfies neither branch (`'0000-00-00' >= $start` is false, and it is not `NULL`), so the row is filtered out of every day/week/month view.

## Fix

Adds a remediation step to `8_1_0-to-8_1_1_upgrade.sql` that:

1. Ensures `pc_endDate` is nullable (guards against any edge case; already applied on affected instances by the prior migration's ALTER).
2. Converts any remaining `'0000-00-00'` sentinels to `NULL`, restoring calendar visibility for affected appointments.

The UPDATE is idempotent — instances without stale rows are unaffected.

Closes #12518

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed calendar issues where open-ended appointments could fail to display correctly when their end date contained placeholder values.
* **Chores**
  * Updated the database schema version to keep it in sync with the application’s expected version.

### DIFF
--- a/sql/8_1_0-to-8_1_1_upgrade.sql
+++ b/sql/8_1_0-to-8_1_1_upgrade.sql
@@ -113,6 +113,7 @@
 --    desc: Add encounter to the form_misc_billing_options table
 --    arguments: none
 
+-- v_database 540: Fix calendar appointments (pc_endDate zero-date sentinels).
 --
 -- Fix calendar appointments disappearing after 8.0.0 → 8.1.0 upgrade.
 -- The 8.0.0→8.1.0 migration ran UPDATE pc_endDate=NULL before the column was
--- a/sql/8_1_0-to-8_1_1_upgrade.sql
+++ b/sql/8_1_0-to-8_1_1_upgrade.sql
@@ -113,4 +113,24 @@
 --    desc: Add encounter to the form_misc_billing_options table
 --    arguments: none
 
--- v_database 539: cosmetic schema edit; no migration. Marker only — remove on any future real upgrade.
+--
+-- Fix calendar appointments disappearing after 8.0.0 → 8.1.0 upgrade.
+-- The 8.0.0→8.1.0 migration ran UPDATE pc_endDate=NULL before the column was
+-- made nullable, so MySQL/MariaDB silently coerced each NULL back to '0000-00-00'.
+-- The calendar query in 8.1.0 matches open-ended events on pc_endDate IS NULL,
+-- so any row still holding '0000-00-00' is invisible in all calendar views.
+-- See: https://github.com/openemr/openemr/issues/12518
+--
+
+-- Ensure pc_endDate is nullable before converting sentinels (idempotent on
+-- instances where the column was already corrected by the earlier ALTER).
+#IfNotColumnTypeDefault openemr_postcalendar_events pc_endDate date NULL
+ALTER TABLE `openemr_postcalendar_events` MODIFY `pc_endDate` date DEFAULT NULL;
+#EndIf
+
+-- Convert any remaining zero-date sentinels to NULL so open-ended appointments
+-- reappear in calendar views.
+SET @currentSQLMode = (SELECT @@sql_mode);
+SET sql_mode = '';
+UPDATE `openemr_postcalendar_events` SET `pc_endDate` = NULL WHERE `pc_endDate` = '0000-00-00';
+SET sql_mode = @currentSQLMode;
--- a/sql/database.sql
+++ b/sql/database.sql
@@ -3,7 +3,7 @@
 --
 -- Keep v_database in sync with $v_database in version.php.
 -- CI will fail if they don't match.
--- v_database: 539
+-- v_database: 540
 --
 
 --
--- a/version.php
+++ b/version.php
@@ -31,7 +31,7 @@ $v_realpatch = '0';
 //
 // Keep in sync with the v_database comment in sql/database.sql.
 // CI will fail if they don't match.
-$v_database = 539;
+$v_database = 540;
 
 // Access control version identifier, this is to be incremented whenever there
 // is a access control change in the course of development.  It is used


sql/8_1_0-to-8_1_1_upgrade.sql
```diff
@@ -113,4 +113,24 @@
...
+-- made nullable, so MySQL/MariaDB silently coerced each NULL back to '0000-00-00'.
+-- The calendar query in 8.1.0 matches open-ended events on pc_endDate IS NULL,
+-- so any row still holding '0000-00-00' is invisible in all calendar views.
+-- See: https://github.com/openemr/openemr/issues/12518
+--
+
+-- Ensure pc_endDate is nullable before converting sentinels (idempotent on
+-- instances where the column was already corrected by the earlier ALTER).
+#IfNotColumnTypeDefault openemr_postcalendar_events pc_endDate date NULL
+ALTER TABLE `openemr_postcalendar_events` MODIFY `pc_endDate` date DEFAULT NULL;
```
Firehed, line 129: AIUI this is strictly redundant assuming 800->810 ran, but it's harmless as a "make sure" task.

Firehed (approved):

PR #12400, opened 2026-06-04
fix(reminders): Numerous bug fixes

sjpadgett:
Fixes #12307

This PR stabilizes several Fax/SMS appointment reminder, recall, dialog, logging, and insurance search workflows. The changes focus on preventing fatal/runtime errors, improving reminder status handling, restoring missing module support endpoints, and making the user-facing reminder/recall pages more reliable.

## Changes

* Fixed appointment reminder runner handling when no appointments are processed during the active cron notification window.

  * The runner now safely handles empty or skipped reminder batches instead of referencing an uninitialized result value after the scan completes.

* Improved reminder runner result handling and logging.

  * Reminder scans now return and log a consistent result even when no appointment notification is sent.

* Corrected appointment reminder status updates after successful reminder processing.

  * Appointment reminder success now updates the appointment status field as expected.

* Improved recurring appointment reminder handling.

  * Updated reminder processing so recurring events are evaluated more reliably by the reminder scan.

* Fixed insurance company search error caused by invalid address lookup input.

  * Insurance search now guards/casts the foreign ID before calling the address service, preventing type errors when the value is empty or not numeric.

* Added the missing patient details endpoint needed by the module contact dialog.

  * Module contact dialogs can now retrieve patient contact details through the expected endpoint.

* Fixed search/add dialog callback handling.

  * Dialog save callbacks now verify the requested JavaScript callback exists before invoking it, preventing runtime errors when an invalid or missing callback is provided.

* Restored notification log visibility for Email and SMS service pages.

  * Email and SMS service log pages now display their related notification logs as expected.

* Fixed reminder/recall JavaScript translation initialization.

  * The Flow Board reminder page now defines its translation object using the standard JavaScript translation helper, preventing `translations is not defined` runtime errors.

## Why this matters

These fixes address several edge cases where reminder and recall workflows could fail silently or throw fatal errors. They also improve visibility into notification activity, restore missing dialog support, and make the reminder UI safer when filtering or showing empty result sets.

Overall, this PR makes the Fax/SMS reminder workflow more reliable across cron scans, appointment status updates, recurring events, service logs, contact dialogs, and recall filtering.

## Summary by CodeRabbit

* **New Features**
  * Added a patient-details endpoint for the messaging module.
  * Improved insurance company form saving using service-driven insert/update mapping.
  * Refined fax/sms notification reporting: removed legacy notification-log endpoints and introduced call-log placeholder output.
* **Bug Fixes**
[...]

PR #12523, opened 2026-06-17
chore: phpstan fixes and other unrelated fixes

### LINKED ISSUE #12307: bug(reminders) Appointment reminder runner fails when no reminders are due and no notifications for recurring events

bradymiller:
Follow-up to #12400. The merged PR landed with a few items that didn't go in cleanly and a few that left rough edges; this PR fixes them and removes some content that was added by accident.

Fixes #12307.

**Bug fixes**

- **#12307 — insurance search TypeError.** #12400 claimed to fix this but the merged `ins_search.php` ended up with both the old buggy code and the proposed replacement coexisting (lines 228–235). The last assignment, `$ins_co_phone = (new PhoneNumberService())->getOneByForeignId($ins_id) ?? null`, still calls the service unconditionally — when `?ins=` is missing or non-numeric, `$ins_id` is `null` and the call re-throws the same TypeError #12307 reports. Replace those eight lines with the clean four-line guarded version using `filter_input(... FILTER_VALIDATE_INT)`.
- **`$ins_co` null-safe form rendering.** When creating a new insurance company (no `?ins=`), `$ins_co` was `null`, and six form fields plus two dropdown comparisons read `$ins_co['key'] ?? '' / ?? null` — PHP 8 emits a "trying to access offset on null" warning before the null-coalesce fires. Default to `[]` (via `(array)` so `getOneById`'s untyped return also collapses) so every offset read is safe.

**Tighter contracts**

- **`AppDispatch::apiFetchPatientDetails()`** — narrowed from `bool|string` to `string`, with a try/catch around `json_encode(..., JSON_THROW_ON_ERROR)`. The "always returns a JSON object string" contract is kept; on the rare UTF-8 encoding failure the endpoint logs via `ServiceContainer::getLogger()` and returns `'{}'` rather than a 500.
- **`AppDispatch::mailEmail()` and `getNotificationLog()`** — both catches were assigning `$e->getMessage()` into the user-facing response, which can leak SMTP credentials or stack trace context. Replace with a PSR-3 logger call and a generic `xlt()` message.
- **`AppDispatch::validEmail()`** kept as the original single-line `return ValidationUtils::isValidEmail($email);` — #12400 rewrote it to a verbose `if/return` block; mine reverts that.
- **`AppDispatch::setModuleType()`, `mailEmail()` PascalCase calls, and `formatPhone()`** — picked up the rest of the upstream-modernization #12400 was midway through but didn't land cleanly: `SessionUtil::setSession()` for the module-type write so it persists past the read-and-close, PHPMailer's camelCase API instead of the deprecated PascalCase aliases (clears six baseline entries), and `PhoneNumber::tryParse()?->toE164()` in place of `PhoneNumberService::toE164()` so this entry point uses the same value object as the rest of the project.
- **`InsuranceCompanyService::saveFromForm()`** — was returning `array|string` where the string was an error message, the design that CodeRabbit flagged and that defeats static analysis on every caller. Change to "returns array on success, throws RuntimeException on failure" so the array shape can flow through and the caller doesn't have to branch on `is_array()`. Caller in `ins_search.php` drops its `is_array/throw` block since the exception now propagates.
- **`InsuranceCompanyService::insert()` and `getInsuranceDisplayName()` and `getDisplayNameForInsuranceRecord()`** got proper `int|string` and `string` return types so the downstream `js_escape()` / `text()` / `attr()` calls don't see `mixed`. `insert()` adds a runtime `is_int||is_string` guard on `$freshId` so an unexpected `false` from `QueryUtils::generateId()` fails fast rather than silently writing a bad row.
- **`setup_services.php`** — the status-line guard `($currentStatus[$selectedService]) !== false` was issuing an undefined-index notice when `$selectedService` was the empty default; add `?? false`.
- **`ins_search.php`** — the CQM SOP dropdown was missing the selected-attribute logic the other two dropdowns had, so the saved value wasn't pre-selected when editing.

**phpstan**

- Removes a duplicate copy of several `Forbidden*Rule.php` files at `.phpstan/` (canonical location is `tests/PHPStan/Rules/`), two `NoCoversAnnotation*` files that duplicate the existing `ForbiddenCoversRule`, a diverged `SoftRequiredArgumentsRule.php` copy, the `MIGRATION_GUIDE.md` draft, and an unreferenced 16,738-line `phpstan-remaining-baseline.neon` that no `.neon` include path picked up. ~17,000 lines total.
- **Zero new baseline entries.** The five baseline additions that crept in during early commits were all resolved at source (return-type declarations, removing dead `is_string()` fallbacks, narrowing `$ins_co` to `array`, removing a tautological `assertArrayHasKey`, and a `(string)` cast on `getDisplayNameForInsuranceRecord`'s return). Baseline regenerations during the PR's history are pure entry **removals** plus a couple of count adjustments — no new suppressions slip in.

Stacks cleanly on master. The PR keeps upstream's formatting choices on files that weren't otherwise touched — no cosmetic-revert noise in the diff.


bradymiller:
some weirdness here. tried to fix by reverting prior commit and fix phpstan and then build out the fixes on top of that. but some strangeness going on. will see if can fix it up better.

PR #12526, opened 2026-06-18
refactor(oe-module-faxsms): Fix SignalWire fax support in oe-module-faxsms

### LINKED ISSUE #12730: refactor(module): Fix SignalWire as a fax vendor to oe-module-faxsms (legacy SDK is unusable)
The faxsms module supports RingCentral and etherFAX for fax. We want SignalWire as a third option — it's a low-cost, programmable-fax provider that fits the module's existing send/receive/poll model well.
The blocker is the SDK. The module's SignalWire client is written against the legacy signalwire/signalwire Packagist package, using its Twilio-fork idioms:
phpuse SignalWire\Rest\Client;
new Client($projectId, $apiToken, ['signalwireSpaceUrl' => $spaceUrl]);
$client->fax->v1->faxes->create([...]);
$client->fax->v1->faxes->read([...], 100);
$client->fax->v1->faxes->getContext($jobId)->fetch();
Two problems with depending on that package:

The legacy package is effectively abandoned (copyright 2018–2021) and drags in a full Relay/websocket half — hundreds of generated files across 20+ API namespaces, almost none of which the module touches. That's exactly the transitive dependency weight we try to avoid.
The maintained replacement is a different API. github.com/signalwire/signalwire-php now ships as signalwire/sdk (v1.x) with a ground-up REST client (SignalWire\REST\RestClient, named-arg constructor, returns plain arrays) and hard dependencies on workerman/workerman + phrity/websocket. Adopting it means rewriting the client against a different surface and pulling in a websocket stack — the opposite of what we want.

So Composer can't give us a clean, lightweight path either way.
What the module actually uses is tiny — four operations, each mapping directly onto SignalWire's Compatibility (LaML) Faxes REST API over HTTP Basic auth:

create → POST /api/laml/2010-04-01/Accounts/<project>/Faxes
read → GET .../Faxes with date filters
getContext($sid)->fetch() → GET .../Faxes/<sid>

Secondary items surfaced while wiring this up (can be folded in or split out):

The setup form marks vendor-specific inputs (smsnumber, phone) required in markup, but hides them by vendor. When a non-matching vendor is selected the hidden-but-required field is unfocusable and HTML5 validation throws "An invalid form control … is not focusable" instead of submitting.
Outbound fax media URL handling needs SSRF protection (the provider's pull model requires a fetchable URL).
No automated coverage exists for the SignalWire path.

Proposal: replace the package dependency with a self-contained shim that reproduces only the LaML Fax surface the module consumes, backed by the Guzzle client OpenEMR already bundles (so it's mockable in the isolated test suite and adds zero new dependencies).

sjpadgett:
Fixes #12730

## Summary

This PR adds SignalWire as a fax vendor in `oe-module-faxsms`. It implements inbound (polling) and outbound (send) fax against SignalWire's Compatibility / LaML Fax API, using a small vendored REST shim rather than the upstream SignalWire SDK, and follows a stateless "provider is the system of record" model so that protected health information (PHI) is only written to OpenEMR when a user explicitly files a fax to a chart.

## Background / motivation

SignalWire exposes a Twilio-compatible (LaML) fax surface, which lets us list, fetch, send, and delete faxes through ordinary REST calls. The current first-party SignalWire PHP SDK is a ground-up rewrite that pulls in a WebSocket/event-loop stack (workerman et al.) we don't want in the OpenEMR tree. Rather than vendor that weight, this PR adds a thin shim that speaks the small slice of the Compatibility REST API the module actually uses and rides on the Guzzle client already present in core — zero new runtime dependencies.

The design intentionally mirrors how the existing RingCentral integration behaves: the provider holds the faxes and their status, and OpenEMR treats its list as the inbox. Faxes are only downloaded and stored locally when a user keeps one (files it to a patient chart); otherwise they remain on SignalWire. For cloud-hosted practices in particular (where storage is a recurring cost), this avoids accumulating fax PHI on disk that nobody asked to keep.

## What's included

**Vendored SignalWire REST shim.** A self-contained client (`Client.php`) under the module's `RestClient/SignalWire/Rest` namespace, reproducing only the surface the controller uses: `fax->v1->faxes->create()`, `->read()` (with auto-pagination), `->getContext($sid)->fetch()`, and `->getContext($sid)->delete()`. All HTTP goes through Guzzle, so the transport is injectable and mockable in tests. Maps the API's snake_case fields to the camelCase shape the controller expects.

**Stateless inbound polling.** `getPending()` renders the inbox directly from the live SignalWire fax list on each check — no reads from or writes to `oe_faxsms_queue`. Received faxes get action icons; in-progress (`receiving`) faxes show a status badge; failed faxes are hidden. Rows are populated from the SignalWire fax JSON (`status`, `from`, `to`, `num_pages`, `date_created`).

**Document actions via the shared contract.** View, download, and delete all route through the same `getDocument()` → `viewFax` handler every other vendor uses, rather than introducing vendor-specific functions. `viewFax()` returns base64 for the in-modal viewer, stages an encrypted temp file for the download flow, or deletes the fax from SignalWire; `disposeDocument()`/`sendFile()` mirror the existing vendor download handoff.

**File-to-chart and dismiss.** Filing a fax to a patient downloads it from SignalWire, stores it as the patient's document, and then deletes the upstream copy so the provider stops holding it. Dismiss deletes the fax from SignalWire so it drops off the polled inbox. Because handled faxes are removed upstream, the live list is always the set of unhandled faxes — no local "handled" bookkeeping is required.

**Outbound send with a hardened media handout.** `sendFax()` sends via the shim's `create()`. SignalWire's send API fetches the document from a URL we provide, so the outgoing PDF must be reachable — but it is PHI and must not sit at a public static path. Instead it is staged **outside the web root** and served by a new login-less endpoint, `faxMedia.php`, gated solely by a short-lived, encrypted, tamper-proof token (carrying the file name, site, and an expiry). The endpoint validates the token and expiry, enforces a strict random-hex filename (no traversal), confirms the file is a real PDF, streams a clean PDF body (errors/whitespace/gzip suppressed), and relies on a TTL sweep for cleanup. The send path also verifies the source is a PDF before handing SignalWire a URL, so a non-PDF fails clearly on our side instead of as a cryptic provider error.

**Live reminder count.** The unread badge counts inbound received faxes live from SignalWire instead of querying the queue, consistent with the stateless model.

## Design decisions and rationale

- **Shim over SDK** — keeps OpenEMR's dependency surface minimal and avoids an event-loop stack we don't need; reuses core Guzzle.
- **Provider as system of record (no queue table)** — eliminates DB/provider status drift, avoids persisting unwanted PHI, and lowers storage cost. `oe_faxsms_queue` is left to etherFAX, whose model requires it.
- **Deferred storage** — download/store happens only on a user action (file-to-chart); the upstream copy is deleted once kept, so SignalWire isn't left holding it either.
- **Consistency with existing vendors** — document actions go through the shared `getDocument()`/`viewFax` contract rather than bespoke handlers, so the UI and JS are unchanged.

## Security / PHI considerations

- Outbound media is never exposed as a static public file. The handout endpoint is login-less by necessity (the provider has no session) but is gated by an encrypted, expiring, single-purpose token; tampering fails authenticated decryption, expired tokens are rejected, and only strict random-hex filenames resolve.
- The PHI exposure window is bounded by the token TTL and the staging directory lives outside the web root.
- Inbound faxes are not written to disk until a user files them to a chart.

## Database / configuration

- No schema changes. (SignalWire does not use `oe_faxsms_queue`.)
- Credentials (Project ID, Space URL, API Token) are entered in the module setup.
- For inbound polling, the SignalWire number must be configured for Compatibility fax (cXML `<Receive>`); faxes only appear in the LaML list when received that way.
- Outbound send requires the OpenEMR host to be reachable by SignalWire's network (a publicly resolvable address).

## Testing

- Isolated PHPUnit tests for the REST shim (create/read/pagination/fetch/error mapping) using a mocked Guzzle handler, and for the download-and-store service.
- A standalone CLI smoke-test script that exercises list/fetch read-only against a live space and diffs the live JSON against the shim's field assumptions.
- Manual round-trip: receive, view, download, file-to-chart, dismiss, and send.

## Known limitations / follow-ups

- **Non-PDF source documents.** SignalWire sends PDF; charted faxes are frequently TIFF. The send path currently rejects non-PDF media with a clear error rather than converting. TIFF→PDF conversion on send (as the RingCentral path does) is a sensible follow-up if practices need to fax charted TIFFs.
- **Webhook.** The polling path is the focus of this PR; the inbound webhook receiver is secondary and can be tracked separately.
- **Module restructure.** A broader effort to generalize the vendor surface — so adding a new provider is a drop-in rather than a clone — is planned as its own change and is out of scope here.

## Notes for reviewers

- The shim lives entirely within the module and adds no Composer dependencies.
- `faxMedia.php` is intentionally login-less; please review the token gating with that in mind.
- Action routing (`viewFax`, `disposeDocument`, `assignFax`) follows the same dispatch convention as the existing vendors.


sjpadgett:
@bradymiller Still some to do on this but wanted to give you opportunity to see where it's going concerning file management.

sjpadgett:
I may be trying too be too clever with faxMedia.php. 
@github-advanced-security[bot] called me out. You can bet my Claude(Goober) will get a strong talking to! lol

sjpadgett:
Would someone please explain what procedure you follow for phpstan commits in PR's.

kojiromike:
> Would someone please explain what procedure you follow for phpstan commits in PR's.

Sure — the short version is: PHPStan runs at level 10 in CI (`composer phpstan`, config `phpstan.neon.dist`), and the check fails when your changes introduce errors that aren't already in the baseline. So the procedure for a PR is to fix the new errors introduced there.

Workflow:

1. Run it locally before pushing: `composer phpstan`. If you've got the docker dev pre-commit hooks installed it runs automatically on commit — see [CONTRIBUTING.md](https://github.com/openemr/openemr/blob/master/CONTRIBUTING.md) (pre-commit setup + the `composer phpstan` command).
2. Fix at the source. Most of what trips people up is the custom rules (no `$GLOBALS` access, no `error_log()`, no `curl_*`, no `call_user_func`, no `empty()`, etc.), and every one of them has a documented before/after in [`.phpstan/README.md`](https://github.com/openemr/openemr/blob/master/.phpstan/README.md). Worth a skim — it'll explain most failures you'll see.
3. Don't hand-edit the baseline, and don't add your new code's errors to it. If you ever do need to regenerate it (rare for a feature PR), it's `composer phpstan-baseline` — never edit the files under `.phpstan/baseline/` by hand.
4. If you hit something you believe is a genuine false positive, say so in the PR rather than baselining it quietly, so a reviewer can confirm it's actually a false positive and not a real bug being masked.

---

@bradymiller I guess there are some tools in the dockers and openemr-cmd for phpstan. I haven't used those, but feel free to comment if there's a different way you do it.

sjpadgett:
I'm on windows git bash. I could use docker but I want to get this to work for other dev's using windows. So my issue isn't getting to work on docker/wsl but using git bash on windows so I can update contrib md.
Thank you though because you've verified the workflow for me.

sjpadgett:
I like the coderabbitai/codepilot because it will make PRs so much easier for those community contributors. I got in trouble using in my last PR sneaking in some weirdness Brady caught. I'm pretty sure was on me for not pulling back to local and testing but I want to try again on this PR.

sjpadgett:
@coderabbitai review

sjpadgett:
Nice work on easy dev improvements! Works great on Docker for Windows. Also just noticed Recent Patients tab on Dashboard. How long has that been there? Kudos and really useful. I may include in Context Manager with the other improvements for that feature.
coderabbitai is proving very useful though one needs to pay careful attention to the suggestions up/down stream ramifications. Depends on scope. The irony is that we have one AI tool correcting another AI tool. It's fun watching Claude try to make excuses! lol.

sjpadgett:
Will stay in draft because adding some more new stuff e.g. Cover Page and looking to refactor RC SDK but it is much more involved than SW and Twilio. I'm also looking to refactor oeHTTP client over to Guzzle os module is consistent throughout.
I've been looking to pull that client I did years ago cause it was never used/needed for the original intent. It's used in a lot more places than I thought so something to discuss.

sjpadgett:
@bradymiller @kojiromike I put in the oeHTTP client years ago. I have a couple places I use it in module. I want to refactor if we aren't going to keep. I know we dropped curl so are we going to allow a mix of clients e.g. Guzzle and/or oeHTTP client(built around Guzzle) or get rid/replace oeHTTP?

sjpadgett:
I still have a pretty extensive security push coming tonight. Testing now.

interface/modules/custom_modules/oe-module-faxsms/messageUI.php
```diff
@@ -877,385 +159,74 @@ function makeRingoutCall(callTo, callFrom = '', id = '') {
...
-                                                <th><?php echo xlt("Pages") ?></th>
-                                                <th><?php echo xlt("From") ?></th>
-                                                <th><?php echo xlt("To") ?></th>
-                                                <th><?php echo xlt("Status") ?></th>
-                                                <th><?php echo xlt("Actions") ?></th>
+                                                <?php foreach ($faxsmsCols as $faxsmsCol) {
+                                                    if (is_array($faxsmsCol) && isset($faxsmsCol['raw'])) {
+                                                        echo is_string($faxsmsCol['raw']) ? $faxsmsCol['raw'] : ''; // nosemgrep: echoed-request -- controlled pre-built markup from MessageUiProfile
+                                                    } else {
+                                                        echo '<th>' . (is_string($faxsmsCol) ? $faxsmsCol : '') . '</th>'; // nosemgrep: echoed-request -- xlt()-escaped column label from MessageUiProfile
```
sjpadgett: Per comment these echo are already escaped by builder in MessageUIProfile these suggestions cause a double escape. See comment in that class.

interface/modules/custom_modules/oe-module-faxsms/src/View/MessageUiProfile.php
```diff
@@ -0,0 +1,265 @@
...
+    /** Text-escaped translated label. */
+    private static function t(string $key): string
+    {
+        return xlt($key);
+    }
+
+    /** Attribute-escaped translated label. */
+    private static function a(string $key): string
+    {
+        return xla($key);
```
sjpadgett, line 263: Here is escaping for UI view.

sjpadgett:
@bradymiller

sjpadgett:
Is anybody going to address this PR. It's ready for merge before it gets stale.

bradymiller (approved):

bradymiller:
nice stuff @sjpadgett ! bringing in after it passes testing. will port it to rel-820.

bradymiller:
just brought this back to rel-820

sjpadgett:
thanks Brady!

PR #12538, opened 2026-06-19
fix(insurance): clear phone_numbers before re-insert in legacy persist + cleanup migration

bradymiller:
## Summary

Third follow-up to PR #10326 (the *PhoneNumberService consolidation* refactor from 2026-01-27). #12529 fixed the `TypeError` on non-NANP phones; #12534 added client-side validation. This PR fixes the third bug from the same source — `InsuranceCompany::persist()` and `Pharmacy::persist()` accumulating duplicate `phone_numbers` rows on every save — and ships a one-time cleanup of rows already in production.

## Symptom

On *Practice Settings → Insurance Companies* (and *Procedures → Pharmacies*), edit any company, click Save without changing anything. The list view now shows the same company name **four times**. Save again: nine. Again: sixteen.

The `insurance_companies` table isn't actually getting duplicate rows — it's still one. What's happening:

- `InsuranceCompany::persist()` and `Pharmacy::persist()` call `PhoneNumberService::insert()` once per entry in `$this->phone_numbers`. The accompanying comment says *"PhoneNumberService handles upsert logic"* — but the service's `insert()` is a plain INSERT. Each save appends a fresh WORK row + FAX row to `phone_numbers` for the same foreign id.
- `InsuranceCompanyService::search()` (the list query) LEFT JOINs `phone_numbers` twice on `i.id` — once filtered to `type=WORK`, once to `type=FAX`. With N work rows × M fax rows for one company, the join produces N×M list rows.

Both `set_number()` methods in the legacy classes already enforce *"replace existing phone of same type, or add new"* at the object layer, so multi-row-per-`(foreign_id, type)` is always a persist-bug artifact, never intentional state.

## Fix

Three pieces:

**1. `library/classes/InsuranceCompany.class.php` and `library/classes/Pharmacy.class.php` — `persist()`:** clear `phone_numbers` for the foreign id before the insert loop so the table mirrors `$this->phone_numbers` after every save.

```php
QueryUtils::sqlStatementThrowException(
    "DELETE FROM phone_numbers WHERE foreign_id = ?",
    [$this->id]
);
foreach ($this->phone_numbers as $phone) {
    // ... existing null-digits guard from #12529, then insert ...
}
```

**2. `sql/8_1_0-to-8_1_1_upgrade.sql` — one-time cleanup of duplicates already in production:**

```sql
DELETE p1 FROM phone_numbers p1
INNER JOIN phone_numbers p2
    ON p1.foreign_id = p2.foreign_id
    AND p1.type = p2.type
    AND p1.id < p2.id;
```

Self-join DELETE keeps the highest `id` per `(foreign_id, type)` and drops older duplicates. *Highest id* is right because if the user ever changed the phone, the latest save carries the corrected value. Idempotent — no-op on instances that never accumulated dupes, so re-running the upgrade is safe.

**3. Regression tests** — `InsuranceCompanyServiceTest::testLegacyPersistDoesNotDuplicatePhoneRowsOnResave` and the matching `PharmacyPersistTest`. Each covers: create, resave-unchanged, resave-with-changed-phone — asserts exactly one row per type each time, and that the row that survives carries the most recent number.

## Verification

- Direct PHP repro through `InsuranceCompany::persist()` and `Pharmacy::persist()` shows the new clear-then-insert leaves exactly one row per `(foreign_id, type)` across all three of the test scenarios.
- Selenium repro through the actual *Insurance Company* edit form: create → 3 resaves → list count stays at 1 each time (versus 1 → 4 → 9 → 16 without the fix).
- Manual sanity check of the cleanup SQL against fake duplicates in `phone_numbers` — 5 rows (3 WORK + 2 FAX) collapse to 2 (one row per type, both holding the highest-id record's values).
- Pre-commit chain green (phpstan, phpcbf, phpcs, rector, codespell, conventional-commits, composer-require-checker).

## Scope notes

- Affects only the legacy `library/classes/{InsuranceCompany,Pharmacy}.class.php::persist()` paths. The modern `InsuranceCompanyService::update()` already calls `phoneNumberService->update()` (true UPDATE WHERE foreign_id AND type), so the modern API path doesn't have this bug and isn't touched here.
- The cleanup query DELETEs across all `foreign_id` values, not just insurance / pharmacy ones — `phone_numbers` is shared, but the single-row-per-`(foreign_id, type)` invariant should hold across every writer anyway. If a non-shared writer somewhere depends on multiple rows per type for the same foreign id, this would be incorrect — I couldn't find such a writer in the codebase.

## Test plan

- [ ] Edit an existing insurance company, click Save, return to list — name should appear once, not multiple times.
- [ ] Repeat for a pharmacy.
- [ ] On a database with existing duplicates (e.g., a long-running instance), run the upgrade — list view stops multiplying, no data lost (highest-id row per type kept).
- [ ] On a clean database, run the upgrade — no-op (idempotent).

### DIFF
--- a/library/classes/InsuranceCompany.class.php
+++ b/library/classes/InsuranceCompany.class.php
@@ -337,33 +337,41 @@ class InsuranceCompany extends ORDataObject
 
     public function persist()
     {
-        parent::persist();
-        $this->address->persist($this->id);
-        $phoneService = new PhoneNumberService();
-        // Reset phone_numbers rows for this company before inserting so the
-        // table mirrors $this->phone_numbers after each save. Without the
-        // clear, persist() accumulates a fresh row on every save (since
-        // PhoneNumberService::insert is a plain INSERT, contrary to the
-        // earlier "handles upsert logic" comment) and the list-view LEFT
-        // JOINs on phone_numbers multiply each company across
-        // (work_count * fax_count) result rows.
-        QueryUtils::sqlStatementThrowException(
-            "DELETE FROM phone_numbers WHERE foreign_id = ?",
-            [$this->id]
-        );
-        foreach ($this->phone_numbers as $phone) {
-            $nationalDigits = $phone->phoneNumber->getNationalDigits();
-            if ($nationalDigits === null) {
-                // The legacy phone_numbers table stores 10-digit NANP parts
-                // (area_code / prefix / number). Skip numbers we can't
-                // represent in that schema rather than throwing a TypeError
-                // from PhoneNumberService::getPhoneParts().
-                continue;
+        // Wrap the whole logical save (parent row, address row, phone_numbers
+        // delete + re-insert) in a single transaction. The delete-then-insert
+        // below is what stops the duplicate-phone-row accumulation that PR
+        // #10326 introduced; without the transaction, a connection drop
+        // between the DELETE and the final INSERT would leave the record
+        // with partial or no phone data instead of just the buggy duplicates.
+        QueryUtils::inTransaction(function (): void {
+            parent::persist();
+            $this->address->persist($this->id);
+            $phoneService = new PhoneNumberService();
+            // Reset phone_numbers rows for this company so the table mirrors
+            // $this->phone_numbers after the loop. PhoneNumberService::insert
+            // is a plain INSERT (contrary to the earlier "handles upsert
+            // logic" comment), so without the clear persist() accumulates a
+            // fresh WORK + FAX row on every save and the list-view LEFT JOINs
+            // on phone_numbers multiply each company across
+            // (work_count * fax_count) result rows.
+            QueryUtils::sqlStatementThrowException(
+                "DELETE FROM phone_numbers WHERE foreign_id = ?",
+                [$this->id]
+            );
+            foreach ($this->phone_numbers as $phone) {
+                $nationalDigits = $phone->phoneNumber->getNationalDigits();
+                if ($nationalDigits === null) {
+                    // The legacy phone_numbers table stores 10-digit NANP parts
+                    // (area_code / prefix / number). Skip numbers we can't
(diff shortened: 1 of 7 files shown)


bradymiller:
@coderabbitai thanks for the catch on transactional safety — addressed in d273d647. Both `InsuranceCompany::persist()` and `Pharmacy::persist()` now wrap the whole logical save (parent row + address row + phone_numbers delete-then-insert) in `QueryUtils::inTransaction()`, matching the existing in-repo pattern at e.g. `src/Billing/BillingUtilities.php:1677`.

The atomicity gap was real and you're right that this PR introduced it — pre-fix the persist path was additive (every save INSERTed more rows on top), so a mid-loop failure left the existing rows intact. After the delete-then-insert switch a mid-loop failure could leave the record with the prior rows already gone and only a partial set of new rows in place. The wrap closes that window.

No new rollback-path test — agree with the "Heavy lift" tag on injecting a mid-persist failure cleanly. The existing regression tests still exercise the happy path through the transaction, and `inTransaction()` itself has framework-level coverage.

bradymiller:
@coderabbitai good call on the test-file fragility — addressed in 1180e14f. Both regression tests now build a type-keyed lookup from the SELECT result and assert against `\$byType[PhoneType::WORK->value]` / `\$byType[PhoneType::FAX->value]` explicitly, rather than relying on `ORDER BY type` putting WORK at index 0. Also tightened the regression by asserting on the FAX row's `area_code` / `prefix` / `number` values (the original only checked WORK, which would let a drifted FAX row slip through).

Minor implementation note: PHPStan level 10 here rejects `(int) \$row['type']` since `\$row['type']` is `mixed` — used `assertIsInt(\$rowType)` to narrow for both the test assertion and PHPStan, matching the project's "narrow, don't cast" convention.

PR #12539, opened 2026-06-19
feat(dx): Add REPL tool for dev envs

Firehed:

During the course of expanding modernization (ORM/DBAL, DI, etc), it's helpful to be able to quickly validate integrations - particularly around the database. Instead of a slew of test scripts with repeated bootstrapping, this adds a REPL tool where the application is already bootstrapped, and exposes the Doctrine ORM EntityManagerInterface as `$em`. Other utilities/variables may get added in later based on their utility value.

- Adds [PsySH](https://psysh.org) to the CLI, available at `./cli shell` (alias: `./cli repl`)
- Adds ext-posix to dev dependencies as a formality, though it's installed by default and must be explicitly disabled

#### Testing

Due to the nature of the package, it's not really possible to add any sort of *meaningful* unit tests as it's based around interactive inputs. Testing was manual:

- Ran `./cli repl` and `./cli shell`, saw it ran and behave as expected
- Ran `composer install --no-dev` to strip dev dependencies, saw it correctly detect the missing class and fail gracefully without disrupting other CLI commands

The test case included is mostly a smoke test.

No

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -137,6 +137,7 @@
         "zircote/swagger-php": "^6.0"
     },
     "require-dev": {
+        "ext-posix": "*",
         "ergebnis/composer-normalize": "^2.48",
         "iamcal/sql-parser": "^0.7.0",
         "justinrainbow/json-schema": "^6.8",
--- a/composer.json
+++ b/composer.json
@@ -150,6 +150,7 @@
         "phpstan/phpstan-strict-rules": "^2.0",
         "phpunit/php-code-coverage": "^11.0",
         "phpunit/phpunit": "^11.0",
+        "psy/psysh": "^0.12.23",
         "ramsey/conventional-commits": "^1.5",
         "rector/rector": "^2.1",
         "roave/security-advisories": "dev-latest",
--- a/src/Console/Command/ShellCommand.php
+++ b/src/Console/Command/ShellCommand.php
@@ -1,0 +1,45 @@
+<?php
+
+/**
+ * Wrap PsySH in a console command with some pre-configured variables.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Eric Stern <eric@opencoreemr.com>
+ * @copyright Copyright (c) 2025 OpenCoreEMR Inc <https://opencoreemr.com/>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ * @see       https://physh.org
+ */
+
+declare(strict_types=1);
+
+namespace OpenEMR\Console\Command;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Psy\Shell;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+
+#[AsCommand(
+    name: 'shell',
+    description: 'Run a REPL in the bootstrapped application',
+    aliases: ['repl'],
+)]
+class ShellCommand extends Command
+{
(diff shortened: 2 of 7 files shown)


src/Console/Command/ShellCommand.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        if (!class_exists(Shell::class)) {
+            // PhySH is installed as a dev dependency, so this condition is
+            // unreachable from test environments.
+            // @codeCoverageIgnoreStart
+            $output->writeln('Shell is only available in dev installations.');
+            return 1;
+            // @codeCoverageIgnoreEnd
+        }
+        $shell = new Shell();
```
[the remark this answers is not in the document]
Firehed: Done this way intentionally: besides not having an interface to reference, trying to DI this through autowiring will crash the CLI in non-dev environments.

tests/Tests/Isolated/Console/Command/ShellCommandTest.php
```diff
@@ -0,0 +1,41 @@
...
+use Symfony\Component\Console\Output\NullOutput;
+
+#[Group('isolated')]
+class ShellCommandTest extends TestCase
+{
+    public function testInvokeStartsShellAndExits(): void
+    {
+        if (posix_isatty(STDIN)) {
+            self::markTestSkipped('PsySH reads from STDIN; skipping in interactive mode. Run tests with `< /dev/null` at the end to test non-interactively.');
+        }
```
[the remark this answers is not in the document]
Firehed: Per PHP documentation, POSIX is enabled by default. However, I've added it as a dev dependency just to be safe.

bradymiller:
hi @Firehed ,
Does the `ext-posix` mean it needs to be installed in the alpine dev dockers? And will this fail in native windows dev environments?

composer.json
```diff
@@ -137,6 +137,7 @@
         "zircote/swagger-php": "^6.0"
     },
     "require-dev": {
+        "ext-posix": "*",
```
sjpadgett, line 140: @bradymiller @Firehed 
psysh and the shell command are a nice addition — no objection to keeping them. One snag from the Windows side: the `"ext-posix": "*"` in `require-dev` breaks `composer install` on Windows, because Windows PHP ships no posix extension, and a composer platform requirement is enforced at install time (and recorded in the lock's `platform-dev`), so every Windows contributor hits a hard failure.

Digging in, the only thing in the tree that actually touches posix is the `posix_isatty()` call in `ShellCommandTest`. So rather than require the extension at install, we can let that test skip where posix isn't available — the standard cross-platform pattern — with `#[RequiresPhpExtension('posix')]`. PHPUnit evaluates it before the method runs, so it stays green on CI (posix present) and skips cleanly on Windows. That lets us drop the `ext-posix` line from `require-dev` and refresh the lock, and psysh is untouched (it only *suggests* posix anyway).

Happy to push that — test guard + the one-line removal + lock refresh — if it's easier on your end. Just say the word.

bradymiller: hi @sjpadgett , Agree with your plan to drop above line.

bradymiller: fyi, this broke easy dev. fixing it by putting support for posix in the flex docker (https://github.com/openemr/openemr/pull/12548). still agree to remove above line though since should just live as a suggestion (then won't hard fail windows environments or environments that simply don't want to use it).

Firehed: Sounds fine to me! Sorry about the disruption, the PHP docs said it was installed by default and did not mention windows.

Issue #12438, opened 2026-06-09
fix(e2e): patient dashboard fails to render after successful create on PHP 8.2 cell


kojiromike:
## Summary

On the `PHP 8.2 - apache - mariadb 11.8.8 / e2e` matrix cell specifically, an Apache worker process **segfaults (signal 11)** while serving the patient Medical Record Dashboard (`demographics.php`) on the patient **re-open** path. The connection drops mid-response with no HTTP status, the tab iframe is left stuck on "Loading…", and the test's 30s `waitFor` for the `Medical Record Dashboard - <name>` heading times out. Every downstream `testPatientOpen` then cascades to ERROR.

This is the dominant remaining failure on `master`'s `Test All Configurations` now that the Docker Hub registry timeouts have been resolved (#12425, #12431). It has failed this one cell on most recent runs while other e2e cells pass or fail only sporadically — so it is intermittent, not deterministic, but heavily concentrated in this cell.

It is the patient **re-open** path that crashes, not create. `testPatientAdd` **passes** (~20.4s — the create-path load `demographics.php?set_pid=1&is_new=1` renders fine and the heading appears). The timeout is in `testPatientOpen` at `PatientOpenTrait.php:83`, on the open path (`set_pid=1`, no `is_new`). The patient is created successfully (the save `POST` returns HTTP 200 and the row exists); re-opening that patient via the finder is what segfaults.

## Root cause: Apache worker segfault (signal 11) serving `demographics.php`

The dashboard does not "fail to render" client-side — the Apache worker process crashes mid-response, the connection drops with no HTTP status, and the tab iframe is left on "Loading…" until the `waitFor` times out. The crash is logged by Apache core, not PHP — which is why an earlier pass grepping for PHP `CRITICAL`/`Uncaught`/`Fatal`/HTTP 500 missed it:

```
[Tue Jun 09 13:49:52 2026] [core:notice] [pid 1:tid 1] AH00051: child pid 153 exit signal Segmentation fault (11), possible coredump in /var/www
```

In run [27210506632](https://github.com/openemr/openemr/actions/runs/27210506632) (job 80338571311) there are **23 distinct segfaults**, recurring every ~35s from 13:49:52 to teardown, in 1:1 lockstep with the browser's `demographics.php?set_pid=1` reload attempts:

| reload (`demographics.php?set_pid=1`) | segfault |
|---|---|
| 13:50:06 | 13:50:07 (pid 127) |
[...]

PR #12546, opened 2026-06-19
refactor(codes): Migrate code remapping to DBAL/ORM

Firehed:
Closes #12503 (alternative approach), and resolve the last hard-technical gap in switching from the home-rolled migration system to doctrine/migrations.

Updates the CPT4/SNOMED-to-`list_options` remapping hook so that it works with the new migrations tooling (alongside the previous approach).

- Adds minimum-viable ORM models for `list_options`, `codes`, and `code_types` tables
- Refactor the remapper to use ORM models, replacing a large number of bare queries
- Updates the migration events to run the handler after Doctrine events
- Backports the enw tool to run in the old event bindings for the existing migrations tools

Claude


Firehed:
disregard the last two debugging commits - I think #12438 struck in a very roundabout way

PR #12566, opened 2026-06-20
chore(composer): ext-posix in `require-dev` breaks `composer install` on Windows

sjpadgett:
Windows PHP ships no posix extension, and a composer platform requirement is enforced at install time (and recorded in the lock's `platform-dev`), so every Windows contributor hits a hard failure. The only thing in the tree that actually touches posix is the `posix_isatty()` call in `ShellCommandTest` . We can let that test skip where posix isn't available — the standard cross-platform pattern — with `#[RequiresPhpExtension('posix')]`. PHPUnit evaluates it before the method runs

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -137,7 +137,6 @@
         "zircote/swagger-php": "^6.0"
     },
     "require-dev": {
-        "ext-posix": "*",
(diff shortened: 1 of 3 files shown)


sjpadgett:
Would someone merge this for me?

PR #12569, opened 2026-06-20
fix(core): prevent {headerTemplate} from rendering the page header twice

### LINKED ISSUE #12567: bug: smarty setupHeader function is duplicating the header in smarty template files
### Describe the bug

On the patient documents upload screen if you attempt to click the Open Patient Template dropdown list the button does not expand the template list for the user.  Digging into the event listeners the bootstrap select event is being added twice due to the bootstrap file being included twice which causes the list to open and immediately close.  To the end user it appears as if nothing is working.

Tracking this down it looks like in commit 4c5c27f8adb51cf8a887ccfff8b49706c313e999 (from January 2026) a change was made to fix a phpstan issue and cause the function to always return the output instead of being an else condition.  However, in smarty if you return a string in addition to echoing the content (default behavior of the function) it ALSO echos the string out to the screen.  This results in the header javascript functions being echoed out twice causing the bootstrap dropdown handler to be added twice to the button.  

Areas of code this impacts:
- pain map form (clickmap area of code)
- prior auth form
- review of systems form
- patient documents form (general_list.html)
- insurance companies list (Practice Settings -> Insurance Companies)
- practice settings
- prescription lookup

### To Reproduce

1. Start easy dev docker development with default database (no patients installed)
2. Add a patient
3. Click on documents
4. Click on a document categories folder
5. Click on the Open Patient Template dropdown list
6. See that no dropdown list opens
7. Inspect the html and see the head script & css are duplicated

Expected the dropdown list to show.

### Client configuration

- Browser: Chrome and Firefox
- OpenEMR version: 8.0.0+
- Operating system: Ubuntu 24.0.4 LTS

dac-ai-worker:
Fixes #12567

## What

The `{headerTemplate}` Smarty plugin (`library/smarty/plugins/function.headerTemplate.php`) did:

```php
return Header::setupHeader($assets);
```

`Header::setupHeader()` has `$echoOutput = true` by default, so it **echoes** the header markup *and* returns it. Because a Smarty function plugin's return value is printed at the tag location, a single `{headerTemplate}` tag emitted the entire `<head>` asset block (jQuery, Bootstrap, etc.) **twice**.

## Why it matters

The duplicate `jquery.min.js` + `bootstrap.bundle.min.js` load registers Bootstrap's dropdown `data-api` click handler on `document` twice. A single toggle click then runs Bootstrap's `toggle` twice — open, then immediately close — so the menu flickers and appears to do nothing.

This is visible on the patient **Documents** upload screen: the "Open Patient Template" dropdown opens and instantly closes. It affects every Smarty page that uses `{headerTemplate}` (documents, practice settings, insurance companies, prescription lookup, review of systems, prior auth, clickmap), which all double-load their assets.

This regressed when the `return` in `setupHeader()` was made unconditional (so it now both echoes and returns); direct PHP callers ignore the return value, so only the Smarty path doubled.

## Fix

Pass `$echoOutput = false` so the plugin only **returns** the markup and Smarty prints a single copy. One line + explanatory comment. The sibling `{assetsTemplate}` plugin already calls `setupAssets()` in return-only form and is unaffected.

## Testing

- Rendered `{headerTemplate assets='datetime-picker|select2'}` through OpenEMR's real Smarty stack: `bootstrap.bundle.min.js`, `jquery.min.js`, `select2.full.min.js` each appear **once** (were twice before).
- Authenticated HTTP fetch of the Documents and prescription-lookup pages: single header; the documents dropdown opens and stays open.
- `composer phpcs` / `php -l` clean; core PHPUnit unit suite green.

### DIFF
--- a/library/smarty/plugins/function.headerTemplate.php
+++ b/library/smarty/plugins/function.headerTemplate.php
@@ -33,5 +33,11 @@ function smarty_function_headerTemplate($params, &$smarty)
         $assets = explode('|', (string) $params['assets']);
     }
 
-    return Header::setupHeader($assets);
+    // Pass $echoOutput = false so setupHeader() only returns the markup. As a Smarty
+    // function plugin, the value returned here is printed at the {headerTemplate} tag.
+    // With the default $echoOutput = true, setupHeader() would also echo the markup,
+    // emitting the entire header (jQuery, Bootstrap, etc.) twice from a single tag --
+    // double-binding Bootstrap's data-api handlers (e.g. dropdowns open then instantly
+    // close). See OpenEMR\Core\Header::setupHeader().
+    return Header::setupHeader($assets, false);
 }
--- a/tests/Tests/Unit/Core/HeaderTest.php
+++ b/tests/Tests/Unit/Core/HeaderTest.php
@@ -1,0 +1,142 @@
+<?php
+
+/**
+ * Tests for the {headerTemplate} Smarty plugin / OpenEMR\Core\Header.
+ *
+ * Regression guard for the bug where a single {headerTemplate} tag emitted the
+ * page header (jQuery, Bootstrap, etc.) twice: Header::setupHeader() defaults to
+ * $echoOutput = true (echo + return), and a Smarty function plugin's return value
+ * is printed at the tag, so the plugin must call setupHeader() in return-only mode.
+ *
+ * @package   OpenEMR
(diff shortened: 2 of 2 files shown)


bradymiller:
Just brought this in and also ported to rel-820 branch (for soon to be released 8.2.0). Thanks for the fix @dac-ai-worker !

PR #12581, opened 2026-06-21
refactor(calendar): remove dead register_globals emulation in pnAPI.php

kiffa-australis256:
## What

`interface/main/calendar/includes/pnAPI.php` (lines **40–58**) still carries the
old PostNuke `register_globals` shim:

```php
// interface/main/calendar/includes/pnAPI.php:40-58
if (ini_get('register_globals') != 1) {
    $supers = ['_REQUEST','_ENV','_SERVER','_POST','_GET',
               '_COOKIE','_SESSION','_FILES','_GLOBALS'];
    foreach ($supers as $__s) {
        if ((isset(${$__s}) == true) && (is_array(${$__s}) == true)) {
            extract(${$__s}, EXTR_OVERWRITE);   // line 53
        }
    }
    unset($supers);
}
```

## Why

`register_globals` was removed back in PHP 5.4, so on 8.2 the `!= 1` check
(line 40) is always true and this block runs on every calendar request — the
`extract(..., EXTR_OVERWRITE)` at **line 53** pulls
`$_GET`/`$_POST`/`$_REQUEST`/`$_COOKIE` into local scope. That lets a request set
arbitrary local variables by picking parameter names, which feels like a footgun
worth removing before something downstream relies on an uninitialized variable.

I didn't find a concrete exploit (it runs after `globals.php`, and `index.php`
re-assigns the vars it uses via `pnVarCleanFromInput`), so this is cleanup rather
than a security fix — but since there's nothing left to emulate, it seems safe to
just delete it.

## Change

```diff
--- a/interface/main/calendar/includes/pnAPI.php   (lines 40-58)
-if (ini_get('register_globals') != 1) {
-    $supers = ['_REQUEST','_ENV','_SERVER','_POST','_GET',
-               '_COOKIE','_SESSION','_FILES','_GLOBALS'];
-    foreach ($supers as $__s) {
-        if ((isset(${$__s}) == true) && (is_array(${$__s}) == true)) {
-            extract(${$__s}, EXTR_OVERWRITE);
-        }
-    }
-    unset($supers);
-}
```

## Notes for reviewers

Worth a quick check that no PostCalendar function reached from the calendar
relies on a variable this used to inject, but the modern `pnVarCleanFromInput`
usage suggests it's safe to drop. No behavior change intended, hence `refactor`.

Was an AI assistant used? Yes — Claude Code

### DIFF
--- a/interface/main/calendar/includes/pnAPI.php
+++ b/interface/main/calendar/includes/pnAPI.php
@@ -234,12 +234,16 @@ function pnVarCleanFromInput()
     $resarray = [];
     foreach (func_get_args() as $var) {
     // Get var
-        global ${$var};
-        if (empty($var)) {
+        if (empty($var) || !is_string($var)) {
             return;
         }
 
-        $ourvar = ${$var};
+        // Historically this read a global of the same name that a
+        // register_globals emulation (removed) populated from the request
+        // superglobals. Read the request directly instead.
+        $ourvar = filter_input(INPUT_GET, $var)
+            ?? filter_input(INPUT_POST, $var)
+            ?? filter_input(INPUT_COOKIE, $var);
         if (!isset($ourvar)) {
             array_push($resarray, null);
             continue;
--- a/interface/main/calendar/includes/pnAPI.php
+++ b/interface/main/calendar/includes/pnAPI.php
@@ -33,31 +33,6 @@ use OpenEMR\BC\Database;
  *
  */
 
-/*        Allows Postnuke to work with register_globals set to off
- *        Patch for php 4.2.x or greater
- */
-
-if (ini_get('register_globals') != 1) {
-    $supers = ['_REQUEST',
-                            '_ENV',
-                            '_SERVER',
-                            '_POST',
-                            '_GET',
-                            '_COOKIE',
-                            '_SESSION',
-                            '_FILES',
-                            '_GLOBALS' ];
-
-    foreach ($supers as $__s) {
-        if ((isset(${$__s}) == true) && (is_array(${$__s}) == true)) {
-            extract(${$__s}, EXTR_OVERWRITE);
-        }
-    }
-
-    unset($supers);
-}
-
-
 /*
  * State of modules
  */
--- a/interface/main/calendar/includes/pnAPI.php
+++ b/interface/main/calendar/includes/pnAPI.php
@@ -234,7 +234,7 @@ function pnVarCleanFromInput()
     $resarray = [];
     foreach (func_get_args() as $var) {
     // Get var
-        if (empty($var) || !is_string($var)) {
+        if ($var === '' || !is_string($var)) {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


Firehed (changes requested):
Hi @kiffa-australis256  - thanks for the contribution! While this is absolutely a desirable change from a security and variable scoping standpoint, it does appear to be causing issues in CI, indicating that some code is relying on this behavior. The CI logs may provide some useful information as to the source, as might existing baselined PHPStan errors.

If you're able to rectify the problem(s) we ought to be able to get this merged (you'll also want to update the phpstan baseline to account for the removal, but best do that when done).

kiffa-australis256:
Thanks! Tracked it down: `pnVarCleanFromInput()` (the calendar's input reader — used for the `module`/`func`/`type` routing in `index.php` and throughout PostCalendar) read request values via `global ${$var}`, which only worked because the register_globals emulation populated those globals. So removing the block left it returning `null`.

Pushed a follow-up that reads the request directly instead:

```php
     foreach (func_get_args() as $var) {
     // Get var
-        global ${$var};
-        if (empty($var)) {
+        if (empty($var) || !is_string($var)) {
             return;
         }

-        $ourvar = ${$var};
+        // Historically this read a global of the same name that a
+        // register_globals emulation (removed) populated from the request
+        // superglobals. Read the request directly instead.
+        $ourvar = filter_input(INPUT_GET, $var)
+            ?? filter_input(INPUT_POST, $var)
+            ?? filter_input(INPUT_COOKIE, $var);
         if (!isset($ourvar)) {
```

`filter_input()` keeps PHPStan's forbidden-superglobal rule happy (no new baseline needed), and I dropped the 3 now-unmatched baseline entries for the removed code (`extract()` arg, `$_GLOBALS`, dynamic `global`).

Locally `composer phpstan` is clean and `composer phpcs` passes on the file. Session-backed values (`pc_username`/`pc_facility`) are read via `$session->get()` elsewhere, so this only affects genuine request params.

interface/main/calendar/includes/pnAPI.php
```diff
@@ -259,12 +234,16 @@ function pnVarCleanFromInput()
...
-        if (empty($var)) {
+        if (empty($var) || !is_string($var)) {
             return;
         }
 
-        $ourvar = ${$var};
+        // Historically this read a global of the same name that a
+        // register_globals emulation (removed) populated from the request
+        // superglobals. Read the request directly instead.
+        $ourvar = filter_input(INPUT_GET, $var)
```
kojiromike, line 244: We have moved from using `filter_input` to using a symfony request object to handle these variables. If you are able to do that now, great. If not, I wouldn't block this PR on that basis.

interface/main/calendar/includes/pnAPI.php
```diff
@@ -259,12 +234,16 @@ function pnVarCleanFromInput()
     $resarray = [];
     foreach (func_get_args() as $var) {
     // Get var
-        global ${$var};
-        if (empty($var)) {
+        if (empty($var) || !is_string($var)) {
```
kojiromike:
```suggestion
        if ($var === '' || !is_string($var)) {
```

I rather think we don't want to treat `"0"` as empty, but we can, if we must.

kiffa-australis256: Applied the $var === '' change — thanks. Migrating to a Symfony Request object should be done in a separate PR, as the scope of this one would grow beyond the register_globals removal.

PR #12602, opened 2026-06-22
fix(ci): improve coredump generation for segfault debugging

Firehed:
## Summary

Investigating why core dumps aren't being generated when Apache workers segfault in CI (issue #12438/#12439). The CI configures coredumps correctly (ulimits, core_pattern, suid_dumpable) but no cores appear when crashes occur.

Changes:
- Add verification steps to `enable_coredumps()` to confirm settings are effective inside the container
- Add test crash that runs before real tests to verify cores are being written
- Add `_coredump_diagnostics()` to log Docker security options, AppArmor profile, capabilities, and ulimits when cores aren't produced
- Disable AppArmor and seccomp in `compose-shared-apache.yml` (Docker's default security profiles can block core file creation even when ulimits allow it)

## Test plan

- [ ] CI runs and the "Enable coredumps" step shows diagnostic output
- [ ] Test crash produces a core file (SUCCESS messages in output)
- [ ] If a real segfault occurs, a core dump is captured and backtraced

### DIFF
--- a/ci/ciLibrary.source
+++ b/ci/ciLibrary.source
@@ -664,13 +664,170 @@ readonly COREDUMP_DIR
 #      the web user, which clears the dumpable flag; mode 2 restores it
 #      (core owned by root, readable by root).
 # core_pattern is a kernel global shared with the container.
+#
+# After configuration, runs a test crash to verify cores are actually being
+# written. If the test crash doesn't produce a core, logs diagnostics to help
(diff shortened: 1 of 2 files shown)


Firehed:
I'd been pulling the changes from this branch into another that was experiencing the stack traces, to no avail. In any case, this was never meant to land.

PR #12689, opened 2026-06-29
refactor(session): Adopt DTO for settings & tidy cookie path

Firehed:
Motivating factors:

- disambiguates `web_root` from session cookie processing logic. I was originally misled to believe that it was in reference to location on disk, when in fact it's actually the cookie's path (used for security scoping)
- continuing to untangle the audit logger setup from other application bootstrapping (also `web_root`)

- Create and adopt DTO with well-known types for session config
- Thread it through the touch points
- Remove some redundant/inappropriate `empty()` checks in the cookie path handling

Some mechanical stuff mostly with tests

### DIFF
--- a/src/Common/Session/SessionConfigurationBuilder.php
+++ b/src/Common/Session/SessionConfigurationBuilder.php
@@ -1,131 +1,60 @@
 <?php
 
-/*
- * SessionConfigurationBuilder.php
- * @package openemr
+/**
+ * @package   OpenEMR
  * @link      https://www.open-emr.org
  * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
  */
+
+declare(strict_types=1);
 
 namespace OpenEMR\Common\Session;
 
 class SessionConfigurationBuilder
 {
-    /** @var array<string, mixed> */
-    private array $config = [];
-
-    public function __construct()
+    public static function forCore(string $webRoot = '', bool $readOnly = true): SessionConfiguration
     {
-        // Set default values that are common across all session types
-        $this->config = [
-            'gc_maxlifetime' => SessionUtil::DEFAULT_GC_MAXLIFETIME,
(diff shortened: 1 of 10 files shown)


src/Common/Session/SessionConfigurationBuilder.php
```diff
@@ -1,131 +1,60 @@
...
-            'use_only_cookies' => true,
-            'cookie_samesite' => 'Strict',
-            'cookie_secure' => false,
-            'cookie_httponly' => true
-        ];
-
-        // Add PHP version-specific settings
-        if (version_compare(phpversion(), '8.4.0', '<')) {
-            $this->config['sid_bits_per_character'] = 6;
-            $this->config['sid_length'] = 48;
```
bradymiller, line 33: hi @Firehed ,
Are we purposefully dropping the above setting of sid_bits_per_character and sid_length ?

Firehed: Oversight on my part, good catch. I'll revisit that.

Firehed: I looked at the PHP history here, specifically https://wiki.php.net/rfc/deprecations_php_8_4#sessionsid_length_and_sessionsid_bits_per_character. Given the reasoning behind the deprecation and the security explanation within the RFC (the system defaults maintain 128 bits of entropy), I don't think it makes sense to reintroduce the logic. The notes about the WAFs in particular seem especially relevant for OpenEMR, leading to the change likely causing more problems than it solves.

bradymiller: hi @Firehed , Sounds good. Did this in distant past when seemed like a good idea to go ultra secure. I do recall the special characters breaking some openemr session logic that had to address at the time. Fare thee well, sid_bits_per_character, you will be missed :)

PR #30, opened 2014-10-27
pull test

mac8905:
pull test

PR #12749, opened 2026-07-02
feat(fqhc): card-per-row layout for report tables on phones

Simonparkershames:
## Summary
- Follow-up to #30: FQHC report/eligibility-worklist tables stopped forcing page-level scroll on phones, but individual tables still scrolled sideways. Below 640px, `.fqhc-table` now renders each row as a bordered card (first cell as title, remaining cells as label/value pairs from the visually-hidden header via `data-label`).
- Adds the `data-label` attributes needed for this across all 10 tables in `report.html.twig` and `eligibility-worklist.html.twig`.

## Certification impact
Markup/CSS-only change confined to the `oe-module-fqhc` custom module (report and eligibility-worklist Twig templates + `fqhc.css`). No certified core code touched.

## Test plan
- [x] Verified against the real rendering pipeline (actual `fqhc-card` custom element + `fqhc.css` + design tokens) by installing/enabling the module locally and injecting sample rows into `report.php` — row titles, label/value pairs, detail rows, and totals rows all render correctly, no desktop regression.
- [ ] Recommend a manual DevTools narrow-viewport check before merge (see note below).

Note: verification used a forced-CSS technique rather than an actual resized viewport (no window-resize control was available in the tooling used), so a quick manual check in browser DevTools responsive mode is worth doing if convenient.


Simonparkershames:
Opened against the wrong repository by mistake (tooling default picked the upstream parent instead of my fork). Closing immediately — no action needed.

PR #12751, opened 2026-07-02
refactor: eliminate session data from OEGlobalsBag

Firehed:
## Summary

Migrates session-specific values away from `OEGlobalsBag` to dedicated session utilities, addressing most items in #12737.

**Migrated values:**
- `pid` → `PatientSessionUtil::getPid()` 
- `encounter` → `EncounterSessionUtil::getEncounter()` / `getOrInitialize()`
- `userauthorized` → `$session->get('userauthorized')`
- `authUserID` → `$session->get('authUserID')`
- `_SESSION['language_direction']` → `$session->get('language_direction')`

These keys have been added into newly-added deprecation warnings, which should hard-fail in development and log in prod. I've left them in the bag until we know there's no access.

**Key changes:**
- Added `EncounterSessionUtil::getOrInitialize()` to consolidate the "get encounter or create with today's date" pattern
- Replaced `global $pid` / `global $encounter` declarations with session utility calls across forms, controllers, services, and library code
- Updated PHPStan baseline to remove entries fixed by these changes

**Not addressed (per issue scope):**
- `bootstrap_pid` in portal code
- `ignoreAuth` (not accessed via OEGlobalsBag, just a plain global variable)

## Test plan
- [ ] Verify patient selection and encounter handling in forms (vitals, SOAP, ROS, prior auth, clickmap)
- [ ] Verify eye_mag form operations  
- [ ] Verify patient document handling
- [ ] Verify prescription interactions
- [ ] Run existing test suites

Closes #12737

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #12764, opened 2026-07-03
chore(phpstan): Fix weird variant type in legacy phreeze path

Firehed:
This came up while working on cleaning up globals. Phreeze used `variant` instead of `mixed` which maybe was a convention to someone at one point. Split out of #12751 since it was already scope-creeping.

- Change `variant` to `mixed` in a bunch of signatures
- Fix a couple other chase-the-error issues
- Update baseline (almost all removal, a couple trades)

Yes


Firehed:
Overrode the coverage check since this is extreme-legacy code.

PR #12791, opened 2026-07-06
ci(docker): add image-to-image upgrade path test workflow

bradymiller:
## Summary

Adds `.github/workflows/docker-test-upgrade.yml` — a new workflow that validates the docker image upgrade path from a prior openemr release to a newer one, using the exact production compose stack end users deploy.

## Pattern

1. Write in-workflow compose override that replaces the sha-pinned `image:` from `docker/production/docker-compose.yml` with a plain tag ref (initially `openemr/openemr:latest`, currently 8.0.0.3 per release-targets.yml). Also relaxes the healthcheck to hit `/` instead of `/meta/health/readyz` for compatibility with older images that predate readyz.
2. `docker compose up --wait` — pulls Docker Hub image, boots stack, waits for healthy.
3. Sanity check web responds.
4. `docker compose down` — preserves named volumes (`databasevolume`, `sitevolume`, `logvolume01`).
5. Sed the compose override to point at `openemr/openemr:next` (rel-820's daily-rebuilt candidate for 8.2.0).
6. `docker compose up --wait` — pulls new image, boots stack. The production entrypoint (`docker/release/openemr.sh`) detects the existing installation and runs the auto-upgrade path (`fsupgrade-N.sh` + `sql_upgrade.php`).
7. Sanity check web still responds after upgrade.

If any upgrade step fails, `docker compose up --wait` times out and the workflow fails with the container logs attached.

## Triggers

- **`push` on `rel-*`** with paths filter matching release-adjacent files (`docker-test-upgrade.yml`, `docker/production/docker-compose.yml`, `version.php`). Fires on release-prep merges to give pre-tag confidence.
- **Daily schedule** at 12:00 UTC — catches Docker Hub tag drift independent of this repo (e.g. Alpine base image regression breaking `latest`).
- **`workflow_dispatch`** — manual tag-pair testing, defaults to `latest` → `next`, but supports arbitrary combos (e.g. `7.0.4` → `8.0.0.3` for a two-release skip, or `8.1.0-2026-06-01` → `next` for a specific dated variant).

Not wired to master pushes yet — target on master would be `dev` (currently master's tag), which represents a much larger jump than most real-world upgrades. Deferring master coverage until the rel-* case proves stable.

## Complements existing tests

| Workflow | Tests |
|---|---|
| `docker-test-release.yml` (existing) | "Does the current branch's Dockerfile build and boot from scratch?" |
| `docker-test-upgrade.yml` (this) | "Does an existing prior-version installation upgrade cleanly to what this branch will ship?" |

Neither is redundant with the other. Together they cover the fresh-install and upgrade-in-place cases.

## Exploratory PR

This is a first cut to see what happens. Failure modes I'd expect if any:

- Prior image's healthcheck endpoint choice — if `latest` (currently 8.0.0.3) doesn't respond to `/` fast enough with the relaxed healthcheck, `docker compose up --wait` will time out. The 3-minute start_period should be plenty but worth watching the first run.
- Auto-upgrade path — this is the whole point. If `fsupgrade-<N>.sh` or `sql_upgrade.php` have any hidden bugs on real 8.0.0 → 8.2.0 data, they surface here.
- Volume persistence across `docker compose down` — verified with an explicit check step; should work but the whole design depends on it.

## Test plan

- [ ] Trigger the workflow via `workflow_dispatch` on master with defaults (`latest` → `next`) — smoke-test that the pattern works end-to-end
- [ ] Trigger with `latest` → `latest` (upgrade from 8.0.0.3 to 8.0.0.3) to prove no-op upgrade works
- [ ] Trigger with `7.0.4` → `latest` (two-release skip) to exercise a real cross-version upgrade
- [ ] Push a version.php bump on a rel-820 test branch and confirm the push trigger fires
- [ ] Wait one day and confirm the scheduled run succeeds

### DIFF
--- a/.github/workflows/docker-test-upgrade.yml
+++ b/.github/workflows/docker-test-upgrade.yml
@@ -1,0 +1,229 @@
+# Docker image-to-image upgrade path validation.
+#
+# Boots the production compose stack from a prior openemr Docker Hub image
+# (default: openemr/openemr:latest -- the current shipped-stable, per
+# .github/release-targets.yml this points at rel-800's v8_0_0_3), lets it
+# install fresh, stops the stack while preserving the named volumes
+# (databasevolume, sitevolume, logvolume01), then boots the same stack
+# against a newer image (default: openemr/openemr:next -- the daily-rebuilt
+# release-in-flight, per release-targets.yml this points at rel-820 today,
+# reflecting what will ship as 8.2.0). The second boot triggers the
+# production entrypoint's auto-upgrade path (fsupgrade-<N>.sh +
+# sql_upgrade.php).
+#
+# Complements docker-test-release.yml: that validates "does the current
+# branch's Dockerfile build and boot?" -- this validates "does an existing
+# installation cleanly upgrade to what this branch will ship?"
+#
+# Uses docker/production/docker-compose.yml as the base compose file --
+# that's what end users deploy, so we test the exact stack shape they'll
+# use. An in-workflow override file replaces the sha-pinned image ref
+# with a plain tag ref so we can swap between FROM_TAG and TO_TAG without
+# editing the compose file itself. The healthcheck is also relaxed to
+# hit `/` instead of `/meta/health/readyz`, since older openemr images
+# may not have the readyz endpoint (added later).
+#
+# Triggers:
+#
+#   1. push on rel-* branches when release-adjacent files change --
+#      catches the moment a release-prep merge pushes a version.php bump
+#      onto rel-*, giving pre-tag confidence that the shipped image will
+#      cleanly upgrade an existing 8.0.0.x installation.
+#
+#   2. schedule daily at 12:00 UTC -- catches drift where "next" or
+#      "latest" tags on Docker Hub degrade independent of this repo
+#      (e.g. Alpine base image regression).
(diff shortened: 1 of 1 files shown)


bradymiller:
Parked in favor of the acceptance-testing migration plan. The compose-override / volume-persistence / image-swap scaffolding here is 40% of what the eventual `acceptance-docker.yml` needs — carrying it forward as reference material rather than shipping the smoke-only version now.

Under the acceptance-testing plan, upgrade validation lives in `tests/Acceptance/` as PHPUnit test classes that run against the shipped artifact from outside, exercising the actual data-integrity + app-functional properties (seed data pre-upgrade → verify post-upgrade). Smoke-only 'does it respond with 2xx' as this PR does is a strict subset of that.

Sequence:
1. Land acceptance-testing planning doc (in progress)
2. Design + land `tests/Acceptance/` harness + `acceptance-docker.yml` (this workflow, evolved) + `acceptance-package.yml`

Branch (`docker-test-upgrade-workflow`) kept alive as reference material for the workflow scaffolding — compose override yaml pattern, `COMPOSE_PROJECT_NAME` pinning, volume-persistence check, image-swap sed step.

PR #12807, opened 2026-07-06
refactor(session): Route pid access through PatientSessionUtil

### LINKED ISSUE #12737: Refactor: eliminate session data from OEGlobalsBag
During analysis, a number of session-specific values were found to be accessed through OEGlobalsBag, rather than the session tooling. It should be migrated to the session-specific tooling, and removed from anything globals-adjacent (the bag, fallbacks, etc).

Likely items: `pid`, `encounter`, `bootstrap_pid`, `userauthorized`, `authUserID`, `ignoreAuth`, `_SESSION`. These were AI discovered and may not be comprehensive or entirely accurate.

See parent issue for more context.

Firehed:
## Summary
- Mark `pid` as deprecated in `OEGlobalsBag` with deprecation warning directing to `PatientSessionUtil::getPid()`
- Convert all `global $pid` and `$session->get('pid')` usages to use the accessor
- Remove insecure pid manipulation from `TwilioSMSClient::index()` that was setting pid from request parameters

Closes #12737 (pid portion only)

## Test plan
- [x] Verify deprecation warning is emitted when accessing `$globalsBag->get('pid')`
- [x] Test patient context flows work correctly with the accessor
- [x] Run PHPStan to verify no type errors


interface/forms/eye_mag/php/eye_mag_functions.php
```diff
@@ -2086,14 +2087,14 @@ function build_PMSFH($pid)
...
+    $pid = PatientSessionUtil::getPid();
     global $PMSFH_titles;
     $count = [];
     $total_PMSFH = 0;
     $header1 = '';
     $display_PMSFH = [];
     if (!($PMFSH ?? '')) {
-        $PMSFH = build_PMSFH($pid);
+        $PMSFH = build_PMSFH((string) $pid);
     }
```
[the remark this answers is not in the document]
Firehed: Well, valid, but existing issue unrelated to this change AFAICT.

Firehed:
I don't know of a practical way to test _every_ flow that references `pid`, but I went through some of the core stuff and didn't run into any obvious issues. The bulk of this is a fairly mechanical change - any of several sources for pid (explicit `global`, implied scope-leaked `$pid`, `OEGlobalsBag`, and sessions),moving to the utility.

interface/patient_file/summary/demographics.php
```diff
@@ -70,9 +71,7 @@
 
 $session = SessionWrapperFactory::getInstance()->getActiveSession();
 
-if (!isset($pid)) {
-    $pid = $session->get('pid') ?? $_GET['pid'] ?? null;
-}
+$pid = PatientSessionUtil::getPid() ?: ($_GET['pid'] ?? null);
```
kojiromike, line 74: This rewrite resurrects a previously-dead fallback. Before this change, `$pid` was always set by `globals.php` (line 748 assigns `PatientSessionUtil::getPid()`), so the `!isset($pid)` guard meant the `$_GET['pid']` branch never ran. Now it's reachable — and the only values that can reach it are non-numeric strings: `globals.php:742` already copies a numeric `$_GET['pid']` into the session before this line runs, so `getPid()` returns it. A request like `demographics.php?pid=abc` now puts the raw string `'abc'` into `$pid`, which then flows into `findByPid()`, `allergy_conflict()`, and the SQL below as a string where every other path guarantees an int.

Could this just drop the fallback?

```suggestion
$pid = PatientSessionUtil::getPid();
```

controllers/C_Document.class.php
```diff
@@ -646,7 +647,7 @@ public function retrieve_action(?string $patient_id, $document_id, $as_file = tr
         // For patient_picture context, non-portal users may only request the session's active patient.
         if ($context === 'patient_picture') {
             if (!($session->has('patient_portal_onsite_two') && $session->has('pid'))) {
-                $allowed_pid = OEGlobalsBag::getInstance()->get('pid') ?? 0;
+                $allowed_pid = PatientSessionUtil::getPid() ?? 0;
```
kojiromike, line 650: `getPid()` is declared `: int` (non-nullable), so the `?? 0` is dead code — and it visually implies a nullability that no longer exists.

```suggestion
                $allowed_pid = PatientSessionUtil::getPid();
```

interface/modules/zend_modules/module/Immunization/src/Immunization/Form/ImmunizationForm.php
```diff
@@ -14,12 +14,14 @@
...
 use Application\Listener\Listener;
 use Laminas\Form\Form;
+use OpenEMR\Common\Session\PatientSessionUtil;
 
 class ImmunizationForm extends Form
 {
     public function __construct()
     {
-        global $pid, $encounter;
+        $pid = PatientSessionUtil::getPid();
```
kojiromike, line 23: `$pid` is never referenced anywhere in this file — the old `global $pid;` was a dead import, and this conversion turns it into a dead local plus a pointless session read. `global $encounter;` on the next line is equally unused (single occurrence in the file). Both lines can go, which would also drop the ImmunizationForm entry from the `openemr.forbiddenGlobalKeyword` baseline.

```suggestion
```

kojiromike (approved):
Nice cleanup — routing everything through one typed accessor is a real improvement, and the `TwilioSMSClient::index()` removal kills a genuinely sketchy pattern. Three small findings below, none raised in earlier reviews.

One broader note on coverage: `PatientSessionUtil::getPid()` has no direct unit tests, and this PR makes its normalization contract (missing/non-numeric → `0`) load-bearing everywhere — e.g. the `!== 0` guards in the `soap_*` files and the `?:` in demographics.php both depend on it. A small isolated test covering missing / numeric-string / non-numeric session values would pin that contract down cheaply. (The codecov/patch failure is mostly the legacy call sites, which is expected — but this piece is new-contract territory and easy to test.)

The Docker build failures are unrelated infra (upstream php-fpm binary download returning HTTP 500).

Issue #12343, opened 2026-06-01
Expand tests/Tests/Api/ coverage: 142 of 172 REST route+method combinations are untested


kojiromike:
## Summary

The `tests/Tests/Api/` suite exercises only a small fraction of the REST API surface. Of the **172** route+method combinations declared across the three route tables, **30 (17%)** are touched by a test and **142 are not**. Most resources have no API-level test at all — including write paths (POST/PUT/DELETE) where a regression silently corrupts or drops data, and where ONC certification depends on correct behavior.

This issue inventories every existing route and its current API-test status so the gaps can be closed incrementally, one resource (or one method) at a time.

## Why this matters

- **Regression safety.** The service layer has unit/CRUD tests, but the HTTP layer (routing, OAuth scope enforcement, status codes, `Location` headers, `OperationOutcome`/error bodies) is almost entirely unverified. A broken route, a wrong scope check, or a 500 in place of a 400 would not be caught.
- **Certification.** FHIR R4 / US Core and the patient-access APIs are certification-relevant; their request/response conformance should be regression-tested.
- **Onboarding.** Each missing entry is a small, self-contained, well-patterned unit of work — good first-issue material.

## How coverage was determined

- **Routes**: extracted from `apis/routes/_rest_routes_fhir_r4_us_core_3_1_0.inc.php`, `_rest_routes_standard.inc.php`, and `_rest_routes_portal.inc.php`.
- **Coverage**: a route counts as covered if any test under `tests/Tests/Api/` exercises that exact method+endpoint, either via an HTTP call through `ApiTestClient` or via an in-process REST-controller method call.

## Scope notes

- The list reflects the current default branch. Write endpoints for additional FHIR resources under active development are tracked with their own PR and are intentionally excluded here.
- A few endpoints exercised by tests are not in the route-table inventory and so are not listed: FHIR `POST /fhir/Patient/_search`, the bulk-export operation routes (`GET /fhir/Group/:uuid/$export`, `$bulkdata-status`), and the `/meta/health/*` liveness/readiness endpoints. These already have some coverage.
- Known issue surfaced while mapping coverage: the one test touching `POST /api/insurance_company` asserts a **500** caused by a missing validate method — i.e. it documents a broken endpoint rather than verifying correct behavior. Worth a separate fix.

## Pattern to follow

[...]

PR #12808, opened 2026-07-06
test(api): add tests for FHIR Condition endpoints

baselaka:
Generated-By: Cursor
Assisted-By: Claude
Assisted-By: Codex

Adds API-level test coverage for the two FHIR Condition read routes (`GET /fhir/Condition` and `GET /fhir/Condition/:uuid`), closing two entries from the coverage inventory in #12343. Tests drive the real HTTP path through `ApiTestClient` (real OAuth token, real request/response), not in-process controller calls. Test-only — no changes to production code under `src/`.

Refs #12343

**`tests/Tests/Api/ConditionFhirApiTest.php`** (new) — eight tests:

- Search returns a collection bundle
- Search by `_id` returns the seeded condition (exactly one entry)
- Read-one returns a Condition resource with the US Core elements present on this path (clinicalStatus, category, code, subject)
- Search with no match returns an empty bundle (HTTP 200, not 404)
- Read-one with a malformed uuid returns 400
- Read-one with a well-formed but non-existent uuid returns 404
- Unauthenticated search returns 401
- Unauthenticated read-one returns 401

**`tests/Tests/Fixtures/FixtureManager.php`**

- `installConditionFixtures()` — seeds a patient + a problem-list condition and returns `{ patientUuid, conditionUuid }`.
- `removeConditionFixtures()` — cleans up the seeded rows.

Condition has no POST route, so data is seeded through the fixture layer rather than by POSTing a resource (the pattern `PatientFhirApiTest` uses).

**Notes on current OpenEMR behavior** — these tests assert current behavior, not strict FHIR conventions:

- Search bundles use `type: "collection"` (`FhirResourcesService::createBundle`), not FHIR's usual `"searchset"`.
- Errors from `FhirGenericRestController` / `RestControllerHelper::handleFhirProcessingResult` return `validationErrors` (400) or `[]` (404), not `OperationOutcome`.
- Problem-list Conditions currently surface the diagnosis as `code.text` only; no structured `code.coding` is emitted even when `lists.diagnosis` holds an ICD-10 code. The read-one test asserts `code.text`. This coding gap is raised separately for maintainer input (see #12343) and could be covered by a follow-up assertion.

Pinning real behavior is intentional for regression coverage.

**Testing**

- Full `Api` suite; all 8 `ConditionFhirApiTest` cases pass against unmodified production code.
- Run twice consecutively — fixture teardown is idempotent (no leftover seeded rows between runs).
- 401 responses confirmed to be OAuth resource-server denials.
- Both changed files pass PHPStan level 10 (targeted analysis).

### DIFF
--- a/tests/Tests/Api/ConditionFhirApiTest.php
+++ b/tests/Tests/Api/ConditionFhirApiTest.php
@@ -1,5 +1,17 @@
 <?php
 
+/**
+ * FHIR Condition API tests.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @author    Vasilii Tereshchenko <vasilii.tereshchenko@gmail.com>
+ * @copyright Copyright (c) 2026 Vasilii Tereshchenko <vasilii.tereshchenko@gmail.com>
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+declare(strict_types=1);
+
 namespace OpenEMR\Tests\Api;
 
 use OpenEMR\Tests\Fixtures\FixtureManager;
--- a/tests/Tests/Api/ConditionFhirApiTest.php
+++ b/tests/Tests/Api/ConditionFhirApiTest.php
@@ -1,0 +1,294 @@
+<?php
+
+namespace OpenEMR\Tests\Api;
+
+use OpenEMR\Tests\Fixtures\FixtureManager;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Tests for the FHIR Condition endpoints:
+ *   GET /fhir/Condition          (search -> returns a Bundle)
+ *   GET /fhir/Condition/:uuid     (read one -> returns a single Condition)
+ *
+ * Modeled on PatientFhirApiTest, but with one big difference: there is NO
+ * POST /fhir/Condition route, so we cannot create test data by POSTing a
+ * Condition the way the Patient test POSTs a Patient. Instead we seed a
+ * patient + a medical-problem row directly through FixtureManager in setUp,
+ * and remove it in tearDown via ConditionFixtureManager::removeFixtures().
+ *
+ * OpenEMR vs FHIR conventions (tests pin current server behavior, not the spec):
+ * - Search bundles use type "collection" (FhirResourcesService), not FHIR's "searchset".
+ * - getOne errors return validationErrors (400) or an empty JSON array (404), not OperationOutcome.
+ *   A follow-up issue could align these with FHIR if desired.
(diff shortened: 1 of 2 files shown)


kojiromike:
Thanks for the contribution! Looks pretty good at a glance. Please address coderabbit's review before I trigger CI.

baselaka:
Thanks @kojiromike  I've addressed CodeRabbit's comment now. Ready for CI whenever you're able to trigger it.

tests/Tests/Api/ConditionFhirApiTest.php
```diff
@@ -0,0 +1,306 @@
...
+            "/apis/default/fhir/Condition",
+            ['_id' => $randomUuid]
+        );
+        $this->assertEquals(Response::HTTP_OK, $result->getStatusCode());
+
+        $bundle = json_decode((string) $result->getBody()->getContents(), true);
+        $this->assertIsArray($bundle);
+        $this->assertEquals("Bundle", $bundle['resourceType'] ?? null);
+        $this->assertEquals("collection", $bundle['type'] ?? null);
+        $this->assertEquals(0, $bundle['total'] ?? null, "No-match search should have total 0");
```
kojiromike: `assertEquals(0, $bundle['total'] ?? null)` passes even when `total` is absent from the response: the null-coalesce yields `null`, and PHPUnit's loose comparison treats `0 == null` as equal. So this assertion can't actually catch a bundle that drops the `total` field. `assertSame()` closes that hole:

```suggestion
        $this->assertSame(0, $bundle['total'] ?? null, "No-match search should have total 0");
```

kojiromike (changes requested):
Thanks — this is careful, well-scoped test coverage, and the class docblock explaining where OpenEMR diverges from FHIR conventions (collection vs searchset, validationErrors vs OperationOutcome) is genuinely useful for the next person. I verified the pinned behaviors against the production code (`FhirResourcesService::createBundle`, `RestControllerHelper::handleFhirProcessingResult`, and the `code.text` title fallback in `FhirConditionTrait::populateCode`) and the assertions match.

One inline comment on a loose assertion that can silently pass, below.

Also: CodeRabbit's earlier nitpick about the inline `/** @var array<string, mixed> $resource */` annotations at lines 146 and 183 is still outstanding — its proposed fix (loosen `assertConditionCodeText()`'s `@param` to `array<array-key, mixed>` and drop the two `@var` lines) is the right shape, since the helper already re-narrows `$resource['code']` with `assertIsArray()` before touching it. CLAUDE.md's static-analysis section asks new code to avoid inline `@var` casts, so it'd be good to fold that in while you're in the file.

tests/Tests/Api/ConditionFhirApiTest.php
baselaka: Good catch — the ?? null was masking a dropped total. Switched to assertSame(), and removed the inline @var annotations by loosening the helper's @param to array<array-key, mixed>.

kojiromike (approved):

PR #12820, opened 2026-07-07
feat(portal): Add view only mode to portal appointments

### LINKED ISSUE #12819: feat:(portal): Add view only mode to portal appointments
This has been a longstanding users request in portal for those admin not wanting patients involed in their scheduling.
 I have time so lets make users happy.

sjpadgett:
Fixes #12819

add feature enable to globals
refactor templated to exclude new and edit buttons replace signerAlertMsg with asynAlertMsg from utils removed alert remove from reset logout timer. Why was it there!

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -3169,6 +3169,13 @@ $GLOBALS_METADATA = [
             xl('Allow Patient to make and view appointments online.')
         ],
 
+        'view_only_portal_appointments' => [
+            xl('Allow Online Appointments View Only.'),
+            'bool',                           // data type
+            '0',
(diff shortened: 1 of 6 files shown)


sjpadgett:
@bradymiller I checked and there is not any "drift" in v8.2.0 except in global.inc-php:
```
    'allow_debug_language' => [
        xl('Allow Debugging Language'),
        'bool',                           // data type
        '0',                              // default = true during development and false for production releases
        '1',                              // default = true during development and false for production releases
        xl('This will allow selection of the debugging (\'dummy\') language.')
    ],
```
Which can let slide as no diff to release however still shouldn't be IMO as it forces the Admin to turn off a debug default! New user will not know what it is unless doing translations maybe!
I'd like to give users at least one requested feature in this release.

PR #12826, opened 2026-07-08
fix(sql-upgrade): server-status polling now terminates reliably

bradymiller:
## Summary

Fixes a race in the `sql_upgrade.php` browser-visible server-status polling loop. On fast/minimal-data upgrades, the JS polling continues indefinitely after the upgrade actually completes — Apache workers idle, MySQL idle, upgrade successful, but the "Server Status" area keeps showing new timestamps as the browser hammers the poll endpoint on a recursive loop. Users interpret this as "upgrade stuck," when in reality the DB has already reached the target version.

## Reproduction

Fresh 8.0.0 install (tarball extracted, mounted into a flex container) upgraded to 8.2.0-dev on Docker loopback network. The UUID population phase finished with only 27 UUIDs backfilled in under a second (small install → few UUIDs). The pre-existing `sleep(2)` delivery margin wasn't enough for the `doPoll=0` chunk to reach the browser through the pipeline's buffering, so polling ran for 10+ minutes at ~1,900 queries/sec against MySQL until the browser tab was manually paused.

## Root cause

`sql_upgrade.php` uses a recursive JS polling loop that hits `library/ajax/sql_server_status.php` continuously to display active DB queries. The termination signal is an inline `<script>doPoll = 0;</script>` chunk streamed via `flush_echo()` at the end of each phase, wrapped by a `sleep(2)` for delivery margin (a pre-existing partial mitigation — the code comment literally says "fixes odd bug, where if process goes to fast, then the polling does not stop").

The `sleep(2)` protects against JS scheduling ordering. It does **not** protect against chunk-delivery buffering. If PHP output buffering, Apache mod_deflate, kernel TCP send buffer (Nagle), or the browser fetch stream absorbs the `doPoll=0` chunk, the browser never sees the stop signal even after the server-side sleep completes.

### 1. Bump server-side delivery margin

`sleep(2)` → `sleep(5)` in both places it appears (after each SQL upgrade file and after the UUID population phase). Cheap increase in the happy-path delivery margin.

### 2. Client-side fallback termination

Add a counter + wall-clock idle guard to the recursive `serverStatus()` JS function:

- Track consecutive empty poll responses (`emptyPollCount`) and the timestamp of the last non-empty poll (`lastNonEmptyPollTime`)
- On each non-empty response: reset both counters
- On each empty response while `doPoll === 1`: increment the counter
- When BOTH thresholds trip (100 consecutive empties AND 15 seconds since last non-empty): treat as proof the phase completed and the primary `doPoll=0` signal was lost, auto-terminate the loop
- Reset both counters at each phase boundary so a slow-starting phase doesn't inherit stale idle state

**Thresholds are conservative** because failure modes are asymmetric:
- **Premature stop** (bad): user sees "done" while upgrade still runs → could close the browser mid-upgrade
- **Late stop** (fine): "done" message shows 15 seconds after actual completion → cosmetic annoyance only

Two-signal AND gate (count + wall-clock) protects against future phases that might have longer PHP-side compute steps between DB batches.

## Interaction between the two layers

On the happy path (~99% of pipelines), the `sleep(5)` delivery margin gets `doPoll=0` to the browser in time, polling terminates promptly (typically <100ms after actual work completes), and the fallback counter never reaches its threshold. On the sad path (chunk truly swallowed), the fallback kicks in after ~15 seconds. Neither mechanism changes the happy-path behavior.

## Test plan

- [x] PHP syntax check on the modified file
- [ ] Reproduce the bug on a fresh 8.0.0 → 8.2.0-dev upgrade with the fix applied — expect polling to terminate cleanly after UUID phase (either via the `sleep(5)` or the fallback)
- [ ] Verify progress display during a phase is unchanged (server status area still shows active queries scrolling while work is happening)
- [ ] Verify the "Server Status" pause toggle still works normally

### DIFF
--- a/sql_upgrade.php
+++ b/sql_upgrade.php
@@ -52,6 +52,12 @@ if ($response !== true) {
 @ob_end_clean();
 // Disable PHP timeout.  This will not work in safe mode.
 @ini_set('max_execution_time', '0');
+// Decouple the upgrade continuation from browser connection state --
+// prevents a client disconnect (tab close, network glitch, or a false-
+// positive fallback trip in the polling loop that scares the user into
+// closing the tab) from aborting the sql_upgrade script mid-flight,
+// which could leave the schema in an inconsistent partial-upgrade state.
+ignore_user_abort(true);
 if (ob_get_level() === 0) {
     ob_start();
 }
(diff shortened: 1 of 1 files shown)


stephenwaite:
freedom-roadtrip

PR #12840, opened 2026-07-08
fix(sql-upgrade): version-check backup + smart dropdown default

bradymiller:
## Summary

Third and final round on the sql_upgrade polling termination problem. Replaces the previous rounds' client-side counter mechanism (which was polluted by poll-endpoint bootstrap queries) with a direct check against actual completion state via the DB's \`version.v_database\` column.

Also adds two mitigations for the "user re-runs sql_upgrade on already-completed install" edge case (dropdown default preselect + initial-version gate on the version check).

## Root cause of the residual issue after #12826 / #12831

The previous rounds landed:
- \`sleep(5)\` bump on the streaming \`doPoll=0\` chunk (primary termination)
- Client-side empty-poll counter + wall-clock idle gate (fallback)
- \`ignore_user_abort(true)\` (safety)
- \`\$skipAuditLog = true\` on the poll endpoint (DB load reduction)

End-to-end reproduction after both PRs merged revealed the fallback counter never trips because the poll endpoint's own \`interface/globals.php\` bootstrap runs a stable-but-large set of read queries per request (globals table load, translation lookups, schema introspection, modules table checks, etc.). Those queries appear in subsequent polls' \`INFORMATION_SCHEMA.PROCESSLIST\` snapshots and reset \`emptyPollCount\` to zero.

Pattern-filtering the specific noise queries out of the poll's SQL is whack-a-mole — every round of testing revealed more patterns; each round of \`globals.php\` additions in the future would add more.

## Fix — two independent termination mechanisms

Instead of fighting the noise, add a mechanism that doesn't depend on the "no DB activity" signal at all.

### Primary — sleep(5) + doPoll=0 chunk (preserved)

Unchanged from earlier rounds. Streaming HTML chunk delivers the stop signal on happy path.

### Backup — 15s version-check via new endpoint (new)

- New endpoint: \`library/ajax/sql_upgrade_version_check.php\` — returns the DB's \`v_database\` column as a plain integer
- JS \`setInterval\` on 15s cadence hits the endpoint, terminates polling when the returned value has *increased* since page load AND caught up to the target from \`version.php\`
- Immune to output buffering, poll endpoint self-noise, and every other failure mode that affected the counter design
- \`sql_upgrade.php\`'s very last DB write is \`\$versionService->update()\`, so a match unambiguously means the upgrade completed

Either mechanism wins → polling display terminates cleanly.

## Edge-case handling for re-runs

Two mitigations for the "user re-runs sql_upgrade on an install already at the target v_database" case:

**1. Dropdown default now preselects the current DB version.** Reduces the mistake pattern at the source — user starts at the right version rather than defaulting to the most recent one regardless of DB state.

**2. Initial-version gate on the version check.** JS captures \`INITIAL_DB_VERSION\` at page render. Version-check only fires COMPLETE when the returned DB version has *increased* since page load. Prevents the display from terminating prematurely when the DB was already at target.

## Reverts from earlier rounds

Removes:
- \`emptyPollCount\`, \`lastNonEmptyPollTime\`, \`EMPTY_POLL_THRESHOLD\`, \`IDLE_TIME_THRESHOLD_MS\`
- The counter reset in the \`if (start === 1)\` block
- The softened "Status polling stopped..." warning wording

Kept:
- \`sleep(5)\`
- \`ignore_user_abort(true)\` (critical safety)
- \`\$skipAuditLog = true\` on the poll endpoint (90 percent DB load reduction — still valuable)

## Test plan

- [x] End-to-end reproduction: fresh 8.0.0 tarball install → flex container → upgrade to 8.3.0-dev candidate → Server Status displays activity → terminates cleanly via primary path or version-check backup, "End watching server processes" + "Database and Access Control upgrade finished" messages render
- [x] Dropdown correctly preselects "8.0.0" on the freshly-installed rig
- [x] Re-run on already-upgraded install (edge case): dropdown preselects the current version (guides user); if user still forces the old version selection, version-check gate blocks premature COMPLETE (falls back to original polling behavior — cosmetic issue only, no data harm, ignore_user_abort protects tab close)
- [ ] rel-820 backport for 8.2.0 (companion PR)

### DIFF
--- a/library/ajax/sql_upgrade_version_check.php
+++ b/library/ajax/sql_upgrade_version_check.php
@@ -1,0 +1,58 @@
+<?php
+
+/**
+ * sql_upgrade_version_check.php
+ *
+ * Lightweight backup completion signal for sql_upgrade.php's status polling.
+ *
+ * Returns the current DB v_database as a plain integer. sql_upgrade.php's
+ * JS compares that against its captured-at-page-load INITIAL_DB_VERSION
+ * and TARGET_DB_VERSION constants; completion is inferred when the
+ * returned value has *increased* since page load AND caught up to the
+ * target (see the version-check block in sql_upgrade.php).
+ *
+ * sql_upgrade.php's last DB write is $versionService->update(), so the
+ * v_database column atomically reflects "upgrade complete." This survives
+ * every failure mode that can affect the main serverStatus() polling
+ * loop's termination signal (streaming chunk delivery through output
+ * buffering, poll endpoint self-noise from globals.php bootstrap queries
+ * polluting the "no DB activity" signal, etc.).
+ *
+ * Called on a JS setInterval() from sql_upgrade.php's page, independent
+ * of the main polling loop. Whichever mechanism notices completion first
+ * ends the polling display.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
+ */
+
+$ignoreAuth = true;
+// Match sql_server_status.php's bootstrap posture -- keys table may be in
+// a partial state during upgrade, connection pooling would confuse the
+// per-request lifecycle, and per-poll audit log INSERTs are pure noise.
+// These flags gate behavior inside globals.php itself so they must be
+// set as $GLOBALS entries before globals.php loads and the autoloader is
+// available (matches sql_server_status.php pattern).
+// @phpstan-ignore openemr.forbiddenGlobalsAccess (Required before globals.php loads the autoloader)
+$GLOBALS['ongoing_sql_upgrade'] = true;
+// @phpstan-ignore openemr.forbiddenGlobalsAccess (Required before globals.php loads the autoloader)
+$GLOBALS['connection_pooling_off'] = true;
+$skipAuditLog = true;
+require_once(__DIR__ . '/../../interface/globals.php');
+
+use OpenEMR\Common\Csrf\CsrfUtils;
+use OpenEMR\Common\Database\QueryUtils;
+use OpenEMR\Common\Session\SessionWrapperFactory;
+
+$session = SessionWrapperFactory::getInstance()->getActiveSession();
+// Same guard as sql_server_status.php -- only sql_upgrade.php's page
+// should be able to invoke this endpoint (CSRF token minted there).
+CsrfUtils::checkCsrfInput(INPUT_POST, subject: 'sqlupgrade', dieOnFail: true);
+
+header('Cache-Control: no-cache');
+header('Content-Type: text/plain; charset=utf-8');
+
+$row = QueryUtils::querySingleRow("SELECT v_database FROM version LIMIT 1") ?: [];
+/** @var array{v_database?: int|string} $row */
+echo (int)($row['v_database'] ?? 0);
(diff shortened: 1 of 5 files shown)


library/ajax/sql_upgrade_version_check.php
```diff
@@ -0,0 +1,58 @@
...
+ * buffering, poll endpoint self-noise from globals.php bootstrap queries
+ * polluting the "no DB activity" signal, etc.).
+ *
+ * Called on a JS setInterval() from sql_upgrade.php's page, independent
+ * of the main polling loop. Whichever mechanism notices completion first
+ * ends the polling display.
+ *
+ * @package   OpenEMR
+ * @link      https://www.open-emr.org
+ * @license   https://github.com/openemr/openemr/blob/master/LICENSE GNU General Public License 3
```
bradymiller, line 29: author name/copyright

library/ajax/sql_upgrade_version_check.php
```diff
@@ -0,0 +1,58 @@
...
+// Same guard as sql_server_status.php -- only sql_upgrade.php's page
+// should be able to invoke this endpoint (CSRF token minted there).
+CsrfUtils::checkCsrfInput(INPUT_POST, subject: 'sqlupgrade', dieOnFail: true);
+
+header('Cache-Control: no-cache');
+header('Content-Type: text/plain; charset=utf-8');
+
+$row = QueryUtils::querySingleRow("SELECT v_database FROM version LIMIT 1") ?: [];
+/** @var array{v_database?: int|string} $row */
+echo (int)($row['v_database'] ?? 0);
```
bradymiller: lets not use v_database. best to use major/minor/patch for matching.

sql_upgrade.php
```diff
@@ -81,6 +82,22 @@
...
 
+// Capture the DB version at the start of this request. Two version-
+// dependent features below use this:
+//   1. The prior-version dropdown preselects the row that matches the
+//      current DB state, guiding users to the right starting point and
+//      reducing the "re-run on already-complete install" mistake.
+//   2. The client-side version-check polling loop only fires COMPLETE
+//      when v_database has *increased* since page load AND caught up to
+//      the target -- avoids firing COMPLETE prematurely if someone runs
+//      the upgrade on an install already at the target v_database.
```
bradymiller: as above, lets not use v_database

bradymiller:
## Review-response commit landed (6e5f19f0b4)

Addresses all feedback from Brady + CodeRabbit:

**Brady:**
- ✓ Added `@author` + `@copyright` block on the new endpoint (matches `sql_server_status.php` pattern)
- ✓ Switched from `v_database` (int) to `v_major.v_minor.v_patch` (string) for the completion comparison. More semantic — `v_database` is a schema counter that could increment on tag changes too; human-readable version is what we mean when saying "did the install reach the target release?"

**CodeRabbit:**
- ✓ Added `declare(strict_types=1)` to the new endpoint (placed after the file-level docblock per phpcs)
- ✓ Removed inline `@var` casts on both `QueryUtils::querySingleRow` result sites — replaced with `is_array()` + `is_string()` narrowing
- ✓ Removed the two baseline count bumps (`cast.int` for `sql_upgrade.php` and the new `v_database might not be defined` entry) — no longer needed since the underlying type issues are fixed via proper narrowing (and `v_database` isn't used at all anymore after switching to major/minor/patch)
- ✓ Uses `isset() && is_string()` narrowing for the require-source variables (`$v_major`, `$v_minor`, `$v_patch`)

Endpoint API changed — returns `v_major.v_minor.v_patch` string instead of a `v_database` integer. JS compares strings; captures `INITIAL_VERSION` + `TARGET_VERSION` at page load via `json_encode()` for safe string embedding. Same three-scenario handling as before.

rel-820 backport #12841 updated with matching patch-id `471f7b2f8aef543b9fc1d1a2203918e11f41976f`.

PR #12861, opened 2026-07-09
fix(calendar): stop "Undefined array key 0" on appointment search

kojiromike:

The appointment search page (calendar → Search) renders a PHP warning directly into the iframe HTML, repeated once per provider:

```
Warning: Undefined array key 0 in .../interface/main/calendar/modules/PostCalendar/pnuser.php
```

In `postcalendar_user_search()`, the provider `<select>` builder read `$sessionPcUsername[0]` unconditionally inside the provider loop. The `pc_username` session value may be:

- a single username (string),
- a list of usernames (from the multi-select `pc_username[]` input), or
- absent (defaulted to `[]`).

So indexing `[0]` emitted `Undefined array key 0` on every loop iteration whenever it was empty. This normalizes to the first username once, above the loop, and guards the comparison. It also fixes a latent bug where a plain-string `pc_username` matched only its first character (string offset access), rather than the whole username.

The now-matched PHPStan baseline entry for this file (`Cannot access offset 0 on mixed`) is removed; the baseline was regenerated with `composer phpstan-baseline` and only that one entry changed.

### DIFF
--- a/.phpstan/baseline/offsetAccess.nonOffsetAccessible.php
+++ b/.phpstan/baseline/offsetAccess.nonOffsetAccessible.php
@@ -11217,11 +11217,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../../interface/main/calendar/modules/PostCalendar/pnuser.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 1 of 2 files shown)


stephenwaite:
freedom-roadtrip

PR #12879, opened 2026-07-11
fix(documents): Run fax dispatch through standard document handling

Firehed:
This path wasn't using the standard Document class for this scan/fax/pdf tooling, which can result in some system settings not being respected with respect to document handling.

Refactor fix_dispatch handling to use Document class instead of raw sql. Update phpstan baseline

Yes


Firehed:
the e2e thing seems to be a regression on master, rather than the usual test flakes?

kojiromike (approved):

PR #12919, opened 2026-07-12
fix(smart): handle optional JWKS metadata during app registration.

### LINKED ISSUE #12917: bug(smart) Fix SMART app registration when using JWKS metadata
### Problem

SMART application registration can fail when JWKS metadata is submitted through either the web registration form or the dynamic client registration endpoint.

Two related failures occur depending on how the `jwks` value is submitted.

When the web registration form leaves inline JWKS blank, it may still submit:

```json
"jwks": ""
```

The registration endpoint then attempts to read `jwks` as an array and returns:

```text
Unexpected value for parameter "jwks": expecting "array", got "string".
```

When a dynamic registration client submits JWKS as a decoded JSON object, Symfony may reject the non-scalar value and return:

```text
Input value "jwks" contains a non-scalar value.
```

### Cause

The registration controller uses Symfony `InputBag` accessors that are specific to one input shape:

* `all('jwks')` expects an array and rejects a JSON string.
* `get('jwks')` expects a scalar and rejects a decoded JSON object.

Valid dynamic client registration requests may provide JWKS metadata as either:

* a JSON object
* a JSON-encoded string

The web registration form also submits empty optional JWKS values instead of omitting them.

### Expected behavior

* Blank `jwks_uri` and `jwks` fields should be omitted from the registration request.
* An application may provide a JWKS URI without also providing inline JWKS.
* Inline JWKS should be accepted as either a JSON object or a JSON-encoded string.
* Invalid JWKS metadata should return an appropriate `invalid_client_metadata` response.
* A valid JWKS object should be normalized and stored as JSON.

### Client configuration

- Browser:
- OpenEMR version: 803 dev
- Operating system: all

sjpadgett:
Fixes: #12917

Fix SMART application registration when an application supplies a JWKS URI without an inline JSON Web Key Set, and improve handling of JWKS metadata submitted through dynamic client registration.

The web-based SMART registration form previously submitted empty strings for optional `jwks_uri` and `jwks` fields. The registration controller then attempted to retrieve `jwks` using Symfony `InputBag::all('jwks')`, which expects an array and throws when the submitted value is a string.

Dynamic registration clients may also submit JWKS metadata as either:

* a JSON object
* a JSON-encoded string fix(smart): handle optional JWKS metadata during app registration.

Symfony’s individual `InputBag` accessors do not safely support both representations:

* `all('jwks')` rejects scalar JSON strings
* `get('jwks')` rejects non-scalar arrays

This change reads the complete request payload and extracts the `jwks` value directly, allowing both valid representations to be normalized and validated.

### SMART registration form

* Do not include `jwks_uri` when the field is blank.
* Do not include `jwks` when the field is blank.
* Parse inline JWKS text as JSON before submission.
* Reject invalid inline JSON before sending the registration request.
* Clarify that either a JWKS URI or an inline JWKS value may be supplied for system scopes.

### Dynamic client registration endpoint

* Read `jwks` from the complete request payload rather than using a shape-specific `InputBag` accessor.
* Accept JWKS metadata supplied as either a decoded JSON object or JSON string.
* Decode string values using `JSON_THROW_ON_ERROR`.
* Validate that the resulting value is an array/object.
* Store the normalized JWKS as JSON.
* Return `invalid_client_metadata` for invalid JWKS values.

### Documentation

* Clarify that EHR-launched SMART applications are launched from the **SMART Enabled Apps** card on the patient Summary or Demographics screen.
* Explain that the launcher is not a main-menu or global-toolbar item.
* Document the requirements for the SMART card to appear.
* Add troubleshooting guidance for disabled clients, missing `launch` scope, invalid launch URIs, disabled FHIR support, and collapsed dashboard cards.

## Problem reproduced

Register a confidential SMART client while providing a valid JWKS URI and leaving the inline JWKS field blank.

Previously, registration could fail with:

```text
Unexpected value for parameter "jwks": expecting "array", got "string".
```

A dynamic registration client submitting a JWKS object could also fail with:

```text
Input value "jwks" contains a non-scalar value.
```

## Expected behavior

* Blank optional JWKS fields are omitted.
* A client may register with a JWKS URI without also supplying inline JWKS.
* Inline JWKS may be supplied as a JSON object or JSON-encoded string.
* Invalid JWKS metadata returns an appropriate client-metadata error.
* Approved EHR-launch applications appear in the patient **SMART Enabled Apps** card.

## Testing

* Confirmed PHP syntax for `interface/smart/register-app.php`.
* Confirmed PHP syntax for `src/RestControllers/AuthorizationController.php`.
* Registered a SMART application using only a JWKS URI.
* Registered a client with inline JWKS metadata.
* Confirmed blank JWKS fields are omitted from web registration.
* Confirmed malformed inline JWKS is rejected.
* Confirmed the registered EHR-launch client appears under **Patient Summary → SMART Enabled Apps** after approval and enabling.

## Notes

This change does not alter the application-approval process. Provider and system applications may still require administrator approval based on the configured OAuth2 approval policy.

Assisted-by: ChatGPT


src/RestControllers/AuthorizationController.php
```diff
@@ -337,11 +337,19 @@ public function clientRegistration(HttpRestRequest $request): ResponseInterface
...
                         }
-                        $params[$key] = json_encode($jwks);
+                        if (!is_array($jwks)) {
+                            throw new OAuthServerException(
+                                'jwks must be a JSON object',
+                                0,
+                                'invalid_client_metadata'
+                            );
+                        }
+                        $params[$key] = json_encode($jwks, JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
sjpadgett: Umm Okay! I wish you AI folks would get on the same page!:)

sjpadgett:
Coderabbitai has some good security issues I may address because I think I wrote the controller but don't want to get side tracked at the moment so maybe in a day or two.

sjpadgett:
@bradymiller This is done and ready.

sjpadgett:
lol I guess I now can't even use my "I'm too lazy" excuse now that we have all the tools do it right!

sjpadgett:
@bradymiller This can be merged.

bradymiller:
hi @adunsulag , jwks seems solid; just wanted to get your blessing on this PR.

src/RestControllers/AuthorizationController.php
```diff
@@ -563,7 +582,7 @@ public function oauthAuthorizationFlow(HttpRestRequest $httpRequest): ResponseIn
         $request = $this->convertHttpRestRequestToServerRequest($httpRequest);
         $session = $this->session;
         if (!empty($httpRequest->getQueryParam('nonce'))) {
-            $session->set('nonce', $httpRequest->getQueryParam('nonce'));
+            SessionUtil::setSession('nonce', $httpRequest->getQueryParam('nonce'));
```
bradymiller: hi @sjpadgett , is this a bug where $session->set is not working?

sjpadgett: good catch Brady. I am trusting ai too much. Aggravating! 
This was introduced during PHPStan cleanup. `SessionUtil::setSession()` looked like an equivalent way to persist the nonce, but the later OIDC flow reads it specifically from `$this->session`, so the replacement was not behaviorally equivalent.

The existing tests passed because they did not cover the complete authorization-to-ID-token nonce flow. I’ve restored the original session-object write and will add or verify coverage for that path.

sjpadgett:
Testing well against JWK system but testing now for PKCE against several different scopes and roles.
Afterward will add better coverage in tests.

sjpadgett:
@bradymiller I still need to finish standard API testing the authorizationController. FHIR with public PKCE, Confidential and JWK working. Tomorrow if okay.

adunsulag (approved):

PR #12924, opened 2026-07-12
refactor(ci): extract byte-identical manifest parsing into a shared script + BATS coverage

bradymiller:
## Summary

Consolidates the three-way duplication of manifest-parsing logic that produced #12916 (first workflow yq bug) and #12920 (second workflow yq bug). Both were the same class of bug: the workflow's inline yq expression drifted from the sync script's `read_manifest_entries` + `filter_by_branch` (in `lib/glob-expand.sh`) because they were maintained independently.

Now there's **one source of truth**: the shared lib. Both scripts and the workflow use it. Test coverage for the workflow's manifest-parsing extends naturally.

## What changed

**New script** `.github/scripts/list-manifest-paths.sh <manifest> <target-branch>` — reads the manifest via `read_manifest_entries`, filters via `filter_by_branch`, prints one path per line. Under 70 lines including comments. No new logic; just wires existing lib helpers into a callable driver.

**`sync-byte-identical.yml`** — replaces the inline yq in the "Compute add-paths input for peter-evans" step with a bash invocation of the new script. Materializes master's copy of both the script + manifest into RUNNER_TEMP (Apply step already materializes `lib/glob-expand.sh` alongside sync-byte-identical.sh; layout works because list-manifest-paths.sh source-loads by relative path).

**`test-byte-identical-scripts.yml`** — adds the new BATS suite directory to the CI run.

**New BATS suite** (10 cases in `tests/bats/ci-scripts/list-manifest-paths/`):

- Usage: missing manifest arg → exit 2
- Usage: missing target-branch arg → exit 2
- Precondition: missing manifest file → exit 2 with clear message
- String-only manifest: all paths printed
- Object-form entry excluded from target filtered out
- Object-form entry included for non-excluded target still printed
- Mixed string + object entries both shapes coexist correctly
- Empty manifest → exit 0, no output
- All entries excluded for target → exit 0, no output
- Target not in any exclude list → all entries printed

## Results

- **53/53 BATS pass locally** (10 new list-manifest-paths + 27 sync + 16 validate).
- Shellcheck clean on all three scripts.

## Why not in the byte-identical set

Like `sync-byte-identical.sh` and `sync-byte-identical.yml`, the new script is master-only-firing (only invoked from sync-byte-identical.yml, which is master-only). RUNNER_TEMP materialization always uses master's copy fresh at workflow-run time, so rel-branch drift doesn't affect execution.

## Cost analysis

−12 lines from the workflow, +67 lines new script + tests. Net add. Every future manifest-schema change now touches ONE place (the lib) that has BATS coverage, instead of three places where two were untested. Given we hit that class of bug twice in the last day, the ratio is favorable.

## Follow-ups deferred

- Similar treatment for the RUNNER_TEMP-extraction pattern in the Apply step could reduce coupling further. Not urgent — that step just materializes known files, doesn't parse anything.
- Workflow-level test coverage via actionlint or nektos/act. Big lift, low return until we see a workflow-shape bug that isn't already covered by scripts.

## Test plan

- [ ] `test-byte-identical-scripts.yml` runs all three suites (53 tests total)
- [ ] `validate-byte-identical.yml` canary still green
- [ ] After merge: sync workflow re-fires cleanly, produces sync PRs against three rel branches (rel-820 = small; rel-800/rel-704 = manifest-only-ish, no orphan release-mechanism paths)

### DIFF
--- a/.github/scripts/list-manifest-paths.sh
+++ b/.github/scripts/list-manifest-paths.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+#
+# Print the byte-identical manifest paths applicable to a given rel
+# branch, one per line. Object-form entries whose `exclude-branches:`
+# list contains the target branch are filtered out.
+#
+# Usage:
+#   list-manifest-paths.sh <manifest-file> <target-branch>
+#
+# Example:
+#   list-manifest-paths.sh .github/byte-identical.yml rel-820
+#
+# Rationale: sync-byte-identical.yml's "Compute add-paths input for
+# peter-evans" step needs the same filter the sync script itself uses
+# (`read_manifest_entries` + `filter_by_branch` in lib/glob-expand.sh).
+# Before this script existed the workflow had its own inline yq
+# expression that got the filter wrong twice (#12916 then #12920).
+# Sharing the logic through the lib eliminates that class of bug.
+#
+# This script is master-only-firing (only invoked from
+# sync-byte-identical.yml, which is master-only). Not in the byte-
+# identical set. If a rel-branch copy of this script drifts from
+# master's, no consumer cares -- master's version is materialized
+# fresh into RUNNER_TEMP at workflow run time.
+#
+# Exit codes
+#   0   normal completion (with or without paths printed)
+#   2   precondition error: missing args, missing yq, missing manifest
+
+set -euo pipefail
+
+if [[ $# -lt 2 ]]; then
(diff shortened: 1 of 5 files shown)


bradymiller:
Real bug — process substitution with `mapfile` swallows the script's exit code. Fix pushed as 5e0e0b4abd: write script stdout to `$RUNNER_TEMP/paths.txt` first, then mapfile from the file, so GitHub Actions' default `bash -eo pipefail` propagates a nonzero exit from `list-manifest-paths.sh`.

PR #12925, opened 2026-07-12
feat(release): move ChangelogGenerator + friends from devops, wire ChangelogMutator

bradymiller:
PR 4 of the "Changelog surface early-migration slice" (see `docs/release-mechanism-migration-from-devops.md`).

## What moves

Four release-mechanism classes come across from `openemr-devops/tools/release/` into openemr's `tools/release/`, plus two bin wrappers that invoke them:

- `tools/release/src/GitHubApi.php` — gh CLI wrapper (dependency of the below)
- `tools/release/src/ChangelogGenerator.php` — commit-range → PR list → conventional-commits categorization → markdown
- `tools/release/src/CompatibilityDeriver.php` — reads `ci/` on a rel branch, computes minimum tested PHP/DB versions
- `tools/release/src/CompatibilityNotesRenderer.php` — renders + injects the "Minimum supported versions" markdown section
- `tools/release/bin/changelog.php` — CLI wrapper for ChangelogGenerator
- `tools/release/bin/derive-compatibility.php` — CLI wrapper for CompatibilityDeriver + CompatibilityNotesRenderer

Autoload paths adjusted to `dirname(__DIR__, 3)` to resolve to openemr's repo root (matching existing bin scripts like `dispatch.php`). Package/license headers switched to openemr. Devops keeps its copies for now — `PreflightChecker` still uses `GitHubApi` there — until a future consolidation PR.

## What ports

The four private-static filter methods from website-openemr's `tools/release-docs/src/ReleaseNotesGenerator.php`:

- `isNoise`, `isDockerBump`, `isNoOpVersionBump`, `scopeOf`
- Constants: `RELEASE_BOT`, `DEPENDABOT`, `DEPENDABOT_DOCKER_GROUPS`, `MACHINERY_SCOPES`

Applied via a `filterNoise()` pre-step in `generate()` so bot re-pins, docker-image bumps, release-machinery PRs, backports, and `[TEST]` tickets get dropped BEFORE categorization + rendering. Composer/npm dependabot bumps pass through as before — those are real user-facing package changes.

`GitHubApi::prsForCommits()` extended to also return the PR author login (needed by the filter to distinguish dependabot from human PRs).

## Compare-link fix

New optional `compareLinkOverride` parameter on `ChangelogGenerator::generate()`. When set, the rendered `## [<v>](compare-url)` heading uses `<override>` for the head ref while the git-range API call still uses `\$head` (which must exist as a ref). Lets the mutator render `vPREV...vNEW` (aspirational immutable-tag URL) at release-prep time when the target tag has not yet been created, enumerating commits from the rel branch tip in the meantime. On the finalize-partner PR post-tag, both refs equal `vNEW` and the URL becomes valid.

## Injected clock

`ChangelogGenerator` now takes an optional `Psr\Clock\ClockInterface` (defaults to `Lcobucci\Clock\SystemClock` with the system timezone, matching the pattern in `HolidayService`). Preserves the documented rerun-idempotence guarantee across a day boundary; tests inject `FrozenClock` for deterministic date output.

## Markdown safety

PR titles, area labels, and advisory summaries are contributor-controlled and were being interpolated directly into the CHANGELOG.md — a `[text](url)` in a merged PR title could plant a spoofed link into the release notes. Two defenses added to `ChangelogGenerator`:

- **`escapeMarkdown()`** — escapes `[` and `]` in contributor-controlled strings before Markdown interpolation. Neutralizes `[text](url)` link-injection while preserving backticks (code identifiers), asterisks, and other Markdown that survives in real PR titles.
- **`sanitizeGitHubUrl()`** — only emits URLs that point at `https://github.com/openemr/`; off-origin URLs (mirror-referencing advisories, rewritten PR URLs) become a neutral placeholder pointing at the repo root.

`ChangelogMutator::applySection()` uses `preg_replace_callback()` (not `preg_replace()`) when replacing an existing section — `$body` from live PR titles / GHSA summaries can legitimately contain literal `$1` / `\1` sequences that `preg_replace` would interpret as backreferences and corrupt the rewritten section.

## New mutator

`tools/release/src/Mutator/ChangelogMutator.php` under namespace `OpenEMR\\Release\\Mutator` (autoload-dev, alongside its `ChangelogGenerator` + `GitHubApi` dependency chain, preserving the original dev/prod split for `OpenEMR\\Release\\`). It:

- Resolves prev tag via \`git tag --list 'v*_*_*' --sort=-v:refname\` filtered to versions strictly less than target
- Resolves range head to target tag when it exists, rel branch otherwise
- Prepends the section after the \`# CHANGELOG.md\` header, or replaces the existing \`## [<target>]\` section wholesale if one is present (idempotent for reruns within a release cycle)

## Wiring

`ReleasePrepCommand`'s default mutator lists wire in the mutator on BOTH scopes via a `class_exists()`-guarded FQCN reference. Under a production `composer install --no-dev` (which never invokes `openemr:release-prep` anyway) the class isn't autoloadable and the guard silently skips it. The mutator FQCN is added to `.composer-require-checker.json`'s symbol-whitelist for the same "documented cross-boundary" reason a handful of existing entries use it (e.g. `OpenEMR\\Common\\Utils\\Error`).

Wired at:

- **`buildDefaultRelMutators`** — runs during release-prep PR (rel-branch scope, prepends CHANGELOG entry using rel-branch tip for the range and vNEW for the URL)
- **`buildDefaultMasterMutators`** — runs during release-finalize partner PR (master scope, post-tag: both range head + URL resolve to vNEW → the persisted CHANGELOG entry becomes the canonical immutable-tag diff shape)

## Tests

At `tests/Tests/Isolated/Release/`:

- **ChangelogGeneratorTest** — 20 cases covering each `isNoise` branch (release-bot / `[TEST]` / backport / machinery-scope / chore-release / dependabot no-op / dependabot docker-path / dependabot docker-group), composer bumps kept, section ordering, developer-changes bucket, area sub-heading, `compareLinkOverride` behaviour, `FrozenClock`-driven date output, and Markdown-injection defenses (bracket escaping in PR titles + area labels + advisory summaries; PR URL sanitization; advisory URL sanitization).
- **ChangelogMutatorTest** — 8 cases covering prepend, replace (in-place), idempotence, range-head selection, prev-tag resolution, MutatorResult shape.

Isolated suite: 3744 tests, all pass. PHPStan clean. Composer-require-checker clean (via the whitelist entry described above).

## Live smoke test

Ran the mutator against real \`v8_2_0...rel-820\` via live gh api (fake target `8.2.99` so prev-tag resolved to `v8_2_0` and rangeHead fell back to `rel-820`, tag not created). Output:

```
## [8.2.99](https://github.com/openemr/openemr/compare/v8_2_0...v8_2_99) - 2026-07-13

### Changed

  - sync PR + delete orphan docker-byte-identical files (rel-820) ([#12901](https://github.com/openemr/openemr/pull/12901))
  - sync stale byte-identical manifest refs on rel-820 (post-#12896) ([#12905](https://github.com/openemr/openemr/pull/12905))
```

Aspirational \`v8_2_0...v8_2_99\` URL renders as expected, prepend inserts cleanly above the existing 8.2.0 section, 8.2.0 section preserved intact. Existing CHANGELOG.md restored after run.

## Coordinated

Depends on **openemr-devops#857** (retire `changelog-pr.php`). Landing either first without the other would either produce no CHANGELOG entry (devops merges first) or double-write CHANGELOG.md (openemr merges first). Coordinate landing same day; ideally this PR first so the mutator is in place before devops stops opening the two post-tag PRs.

## Deferred to follow-ups

- Fixture regression that regenerates 8.2.0's committed CHANGELOG.md entry from frozen inputs (would need ~100 real PR objects captured to fixtures; skipped for scope). The behaviour is exercised by the unit tests above; the shape is regression-guarded by the section-ordering / URL-format / idempotence assertions.
- Compatibility injection (CompatibilityDeriver + Renderer) into the mutator flow. Kept out of scope; existing 8.2.0 CHANGELOG entries include a Minimum-supported-versions section that a follow-up CompatibilityMutator can add. Not blocking the changelog surface migration.
- `CompatibilityNotesRenderer::inject()` idempotence — CR flagged that a second `inject()` call on an already-injected notes file duplicates the "### Minimum supported versions" section. Legit finding but pre-existing behavior from devops's copy and not called by anything this PR wires today. Will be addressed alongside the follow-up CompatibilityMutator.
- Retire `src/Common/Command/CreateReleaseChangelogCommand.php` (Stephen Nielson's older milestone-driven Guzzle-based helper, 374 lines, redundant with what the new ChangelogGenerator + ChangelogMutator flow produces). Slated for after PR 6 (website retire) so we've fully cut over first.

## Test plan

- [ ] CI green
- [ ] After merge: land openemr-devops#857 within minutes
- [ ] Verify next release cycle populates CHANGELOG.md via the mutator (release-prep PR includes CHANGELOG.md in its file list)


bradymiller:
CodeRabbit round-1 addressed as 73590d7b83:

- **#5 ChangelogMutator preg_replace backref injection (Major)** — real bug, fixed with `preg_replace_callback` + static closure returning replacement verbatim.
- **#3 date('Y-m-d') breaks rerun-idempotence (Major)** — injected `Psr\Clock\ClockInterface`, defaults to `Lcobucci\Clock\SystemClock` (same pattern as `HolidayService`). New isolated test uses `FrozenClock` to pin the date.
- **#2 changelog.php return-value check (Minor)** — `mkdir` + `file_put_contents` failures now return exit 1 with a stderr message.
- **#1 ReleasePrepCommand comment (Minor)** — comment rewritten to describe what the code actually does (FQCN via `::class`, autoload guarded by `class_exists`, symbol-whitelisted in `.composer-require-checker.json`).

Deferring **#4 CompatibilityNotesRenderer.inject() idempotence** to a follow-up. Legit finding but pre-existing behavior from devops's copy and not called by anything this PR wires today (`ChangelogMutator` doesn't thread `CompatibilityDeriver` through in this scope). Will be addressed alongside the follow-up `CompatibilityMutator`.

Also re-ran the live-gh-api smoke test — same output, still generates a valid 8.2.99 section against `v8_2_0...rel-820` with the aspirational `v8_2_0...v8_2_99` compare URL.

bradymiller:
Reverted `6dd4097659` as `4666440275`. Back to `::class` + `.composer-require-checker.json` whitelist entry.

Rationale: the rector-skip approach (skip `StringClassNameToClassConstantRector` for the whole file) had the wrong blast radius — rector's skip granularity is file-level, so any future string-class-name literal added to `ReleasePrepCommand.php` would silently escape conversion too. That's overreach for the actual scope of the exception (one line). Whitelisting the single symbol in composer-require-checker keeps rector's rule active codebase-wide and confines the exception to exactly the class name that actually crosses the dev/prod boundary.

bradymiller:
CR round-2 (Markdown injection via contributor-controlled fields) addressed as 24a5931b5e:

- **`escapeMarkdown()`** — escapes `[` → `\[` and `]` → `\]` in PR titles, area labels, and advisory summaries. Neutralizes `[text](url)` link-injection while preserving backticks (code identifiers) + other Markdown that survives in real PR titles today.
- **`sanitizeGitHubUrl()`** — only emits URLs that point at `https://github.com/openemr/`; off-origin URLs (mirror-referencing advisories, rewritten PR URLs) become a neutral placeholder pointing at the repo root.
- Applied to `formatAdvisories()` (severity + summary + ghsa_id + url) and `formatByCategory()` (area label + PR title + PR url). Extracted per-PR line to `formatPrLine()` so the unlabeled + area-grouped paths stay consistent.

4 new isolated tests cover each branch. 29/29 filter/generator cases pass.

Inherited from prio devops code — not a new issue introduced by this PR — but fixed at the port site rather than devops-side because devops's copy is being retired in the follow-up PRs anyway.

bradymiller:
CR round-3 addressed as `abcc9bf217`.

Fair catch — my assertion `assertStringContainsString('https://github.com/openemr/openemr', $out)` was tautologically satisfied by the release heading's compare link, so it passed regardless of whether the PR-line sanitizer regressed. Tightened to `([#1](https://github.com/openemr/openemr)` which matches only the PR bullet-line shape and would fail if the sanitizer dropped/altered/blanked the URL.

PR #12928, opened 2026-07-13
feat(release): CompatibilityMutator + fix inject() idempotence (PR 8)

bradymiller:
PR 8 of the "Changelog surface early-migration slice" (see `docs/release-mechanism-migration-from-devops.md`).

**Sequencing note (resequenced 2026-07-13):** originally slotted as a non-blocking follow-up, but promoted to land **before PR 5** to close the compat-gap window. Without this, PR 5's section-extract would pull an entry missing the "### Minimum supported versions" block for 8.2.1+, since ChangelogMutator alone doesn't produce that section.

## New mutator

`tools/release/src/Mutator/CompatibilityMutator.php` under namespace `OpenEMR\\Release\\Mutator` — same autoload-dev placement as `ChangelogMutator`, alongside its `CompatibilityDeriver` + `CompatibilityNotesRenderer` deps.

**Wired AFTER ChangelogMutator** on both rel + master scopes via the same `class_exists()`-guarded FQCN pattern. Ordering matters — CompatibilityMutator injects into the `## [X.Y.Z]` heading that ChangelogMutator writes.

`appendOptionalReleaseMutators()` refactored to loop over an array of FQCNs (was one hardcoded entry).

## CI-directory sourcing

On both scopes the mutator materializes rel branch's `ci/*/docker-compose.yml` files via `git ls-tree` + `git show` per file (not `git archive` — `.gitattributes` marks `ci/` as `export-ignore`, correctly, since ci configs shouldn't ship in release tarballs, but that means `git archive` silently produces an empty tarball).

Master scope needs this materialization because master's `ci/` can diverge from rel-820's during the release window (master accepts newer CI additions the release wasn't tested against). Rel scope uses the same code path for uniformity — one code path, tested once.

## `inject()` idempotence fix

CR round-2 finding on PR 4 (openemr/openemr#12925): `CompatibilityNotesRenderer::inject()` was non-idempotent — a second call on already-injected notes duplicated the section by blindly splicing after the first `## ` heading. This PR is the first live consumer, so we own the behavior.

Fix: strips any existing `### Minimum supported versions` block (bounded by heading + next `##`-prefixed heading of any depth or EOF) before injecting. Rerun with same section → no diff. Rerun with different minimums → stale block replaced instead of duplicated.

## Latent bug in ChangelogMutator (fixed here)

Noticed while planning PR 8: `release-prep.yml`'s **rel-scope** invocation did NOT pass `--rel-branch`, so `$context->relBranch` was null on rel scope. ChangelogMutator's `resolveRangeHead()` throws when target tag doesn't exist AND relBranch is null — exactly the pre-tag release-prep-PR-open case. **Would have blown up on the first real 8.2.1 release-prep run.**

Fix: pass `--rel-branch` on rel scope too (workflow already computes `steps.resolve.outputs.branch`). CompatibilityMutator needs the same input for its git-tree reads + matrix URL.

## Whitelist

`.composer-require-checker.json` gets a second whitelist entry (`OpenEMR\\Release\\Mutator\\CompatibilityMutator`) alongside `ChangelogMutator`, same "documented cross-boundary" reason.

## Tests

At `tests/Tests/Isolated/Release/`:

- **`CompatibilityMutatorTest`** — 6 cases exercising real `CompatibilityDeriver` + `Renderer` against a scratch git repo with seeded `ci/` dirs + `rel-820` branch: relBranch required, injection ordering (compat before Fixed), matrixUrl uses rel-branch, rerun idempotence, master-scope reads rel branch's `ci/` (not master's `ci/` after intentional drift), MutatorResult shape.
- **`CompatibilityNotesRendererTest`** — 5 cases covering render shape, injection after version heading, prepend when no heading, idempotence (rerun no diff), and stale-section replacement.

Isolated suite: 3756 tests, all pass. PHPStan clean. Composer-require-checker clean.

## Live smoke test

Ran the mutator against real rel-820's `ci/` via git ls-tree + git show. Output:

```
## [8.2.99](https://github.com/openemr/openemr/compare/v8_2_0...v8_2_99) - 2026-07-13

### Minimum supported versions

- **PHP** 8.2+
- **MariaDB** 10.6+
- **MySQL** 5.7+

See the [tested CI matrix](https://github.com/openemr/openemr/tree/rel-820/ci) for all tested version combinations.
```

Exactly the shape 8.2.0's committed CHANGELOG entry has today.

## Coordinated

**Coordinated rel-820 companion PR** (opening after this one merges) to add the `OpenEMR\\Release\\Mutator\\CompatibilityMutator` whitelist entry to rel-820's `.composer-require-checker.json`. Same shape as #12927 was for ChangelogMutator's whitelist, since `.composer-require-checker.json` isn't in the byte-identical set.

## Test plan

- [ ] CI green
- [ ] After merge: open rel-820 companion PR for whitelist
- [ ] After both land: sync workflow re-fires cleanly against rel-820 (byte-identical picks up the new mutator + wiring + renderer fix)
- [ ] Next real release cycle (probably 8.2.1) is the acceptance test end-to-end

### DIFF
--- a/.codespell-ignore-words.txt
+++ b/.codespell-ignore-words.txt
@@ -112,6 +112,7 @@ cancelled
 # These require separate PRs with proper testing (see SPELLCHECK_BATCHES.md)
 cant
 collapsable
+deriver
 detatch
 durationm
 ect
--- a/.composer-require-checker.json
+++ b/.composer-require-checker.json
@@ -25,6 +25,7 @@
         "OpenEMR\\Common\\Http\\UploadedFile",
         "OpenEMR\\Gacl\\Hashed_Cache_Lite",
         "OpenEMR\\Release\\Mutator\\ChangelogMutator",
+        "OpenEMR\\Release\\Mutator\\CompatibilityMutator",
         "Pdo\\Mysql",
         "PHPUnit\\Framework\\Attributes\\Test",
         "Psy\\Shell",
--- a/.github/workflows/release-prep.yml
+++ b/.github/workflows/release-prep.yml
@@ -247,10 +247,12 @@ jobs:
       if: steps.resolve.outputs.should-run != 'false'
       env:
         VERSION: ${{ steps.resolve.outputs.version }}
+        REL_BRANCH: ${{ steps.resolve.outputs.branch }}
       run: |
         php bin/console openemr:release-prep \
           --skip-globals \
           --target-version="${VERSION}" \
+          --rel-branch="${REL_BRANCH}" \
           --scope=rel
 
     # Open/update the long-lived release-prep PR via peter-evans/create-pull-request,
--- a/src/Common/Command/ReleasePrepCommand.php
+++ b/src/Common/Command/ReleasePrepCommand.php
@@ -217,24 +217,32 @@ final class ReleasePrepCommand extends Command
      * composer-require-checker does not flag the cross-boundary
      * dependency.
      *
+     * Ordering matters: CompatibilityMutator must run AFTER
+     * ChangelogMutator because it injects into the `## [X.Y.Z]` heading
+     * ChangelogMutator writes.
+     *
      * @param list<MutatorInterface> $mutators
      */
     private function appendOptionalReleaseMutators(array &$mutators): void
     {
-        // FQCN via ::class: composer-require-checker sees the reference
-        // (which is why the class name is whitelisted in
+        // FQCNs via ::class: composer-require-checker sees the reference
+        // (which is why the class names are whitelisted in
         // .composer-require-checker.json), but autoload is not triggered
         // by the reference itself — only by the guarded new below. Under
-        // a production `composer install --no-dev` the class is not
+        // a production `composer install --no-dev` the classes are not
         // autoloadable, class_exists() returns false, and the mutator is
         // silently skipped. Kept as ::class (rather than a plain string
         // literal) so an IDE can refactor the FQCN alongside the file if
         // it ever moves.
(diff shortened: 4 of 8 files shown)


bradymiller:
CR round-1 addressed as \`46684d8818\`. Both findings were real:

**Critical — inject() strip regex not scoped to target section**: My prior fix scanned the whole notes buffer for the existing block. On a real multi-release CHANGELOG where an OLDER `## [X.Y.Z]` had its own compat block, the strip would delete the older release's block. Test fixtures only had one `## ` heading so this wasn't caught.

Fix: scope the strip to the FIRST `## ` section's contents (heading → next `## ` or EOF). Older release sections keep their compat blocks intact. New test in `CompatibilityNotesRendererTest` uses a two-release fixture to assert this + a companion end-to-end test in `CompatibilityMutatorTest`.

**Major — bare `rel-820` name not resolvable in master-checkout**: Correct — `master-checkout` has `refs/remotes/origin/rel-820` (via `fetch-depth: 0`) but not `refs/heads/rel-820`. My prior code used the bare name in `git ls-tree` + `git show`, which would miss on master scope.

Fix: new `resolveRelBranchRef()` tries the bare name (rel scope), falls back to `origin/<rel>` (master scope), throws with a fetch hint otherwise. Both git operations use the resolved ref. New test simulates the exact master-checkout shape (bare clone + `--single-branch`, explicit fetch refspec) and asserts the mutator succeeds; companion test asserts the throw message when neither ref resolves.

15/15 tests pass. phpstan clean.

Issue #12439, opened 2026-06-09
ci: capture openemr PHP log and phpunit junit in the e2e failure-diagnostics artifact


kojiromike:
## Summary

When an e2e/Selenium matrix cell fails, the "Capture selenium diagnostics on failure" step (`.github/workflows/test.yml:663-705`) uploads the selenium log, `docker compose ps`, `dmesg`, and docker stats — but **not** the openemr container's PHP error log, nor the phpunit junit / diagnostic stdout. As a result, diagnosing a real failure requires extracting and watching the recorded session video frame-by-frame, which is slow and error-prone.

This is motivated directly by the 8.2 dashboard-render investigation (#12438): the productive evidence ultimately came from the video, because neither the openemr PHP log nor the phpunit junit was reachable from the downloadable artifact. (The `E2e container logs` step does `dc logs openemr`, but only echoes to the job log, which is truncated to the summary tail — so the #11858 per-step console dumps that go to phpunit stdout aren't recoverable that way either.)

## Proposed change

Extend the failure-diagnostics step (`test.yml:663-705`) to also collect, into the uploaded artifact:

- the openemr container's PHP error log (e.g. `dc logs openemr` redirected to a file under the diagnostics dir, and/or the in-container Apache/PHP error log), and
- the phpunit junit XML and diagnostic stdout from the e2e run.

## Why

So the next e2e failure can be root-caused from the downloadable artifact alone, without needing to extract video frames. This makes the flaky-e2e work (#12428 and friends) substantially faster to triage.

Issue #12423, opened 2026-06-08
CI: master "Test All Configurations" red since 2026-06-06 due to Docker Hub registry pull timeouts and e2e flakes


kojiromike:
## Summary

The `Test All Configurations` workflow has been failing on `master` since 2026-06-06 (last green run: 2026-06-06 10:46 UTC). Investigation of the recent failed runs shows the failures are **not** product/test regressions but **CI infrastructure flakiness** — predominantly Docker Hub registry pull timeouts when bringing up the compose stack, plus intermittent e2e/Selenium flakes.

Because the matrix has no resilience around image pulls, a single transient registry timeout fails the whole run, and the same failures surface as red "Upgrade from 5.0.0 setup" / e2e checks on essentially every open PR.

## Evidence

Recent failed `master` runs of `Test All Configurations`:

- [Run 27118824678](https://github.com/openemr/openemr/actions/runs/27118824678) (2026-06-08) — `PHP 8.4 - nginx - mariadb 11.8.8 / e2e`:
  ```
  mysql Error Get "https://registry-1.docker.io/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
  Error response from daemon: Get "https://registry-1.docker.io/v2/": ...
  ##[error]Process completed with exit code 1.
  ```
- [Run 27074532822](https://github.com/openemr/openemr/actions/runs/27074532822) (2026-06-06) — `PHP 8.2 - nginx - mariadb 11.8.8 / services`:
  ```
  openemr Error Get "https://registry-1.docker.io/v2/": context deadline exceeded
  Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded
  ```
  (plus a `PHP 8.5 - apache - mysql 9.3.0 / e2e` failure in the same run)
- [Run 27105227451](https://github.com/openemr/openemr/actions/runs/27105227451) (2026-06-07) — `PHP 8.5 - apache - mariadb 11.8.8 - Redis Sentinel / e2e`: selenium container came up healthy, then a test-level exit 1 (e2e flake, no infra error).

## Impact

- `master` shows red, masking any genuine regressions that might land.
- Open PRs inherit red checks from the same transient failures, making it hard to tell real failures from flakes and blocking auto-merge.

## Suggested directions

- Add retry/backoff around `docker compose up --pull` / image pulls (e.g. retry the compose bring-up step on registry errors).
[...]

PR #12975, opened 2026-07-14
[DEBUG - do not merge] ci: capture apache coredump for #12438-class segfault hitting upgrade-e2e

bradymiller:
## What this is

**Not for merge.** Diagnostic PR to capture a coredump backtrace for the intermittent apache-worker segfault (`AH00052 ... Segmentation fault (11)`) that has been dominating the `PHP 8.2 - apache - mariadb 11.8.8 - Upgrade from 5.0.0 / e2e` cell since 2026-07-13.

## Backstory

The failure was initially chased at the Twig layer — every failing run's login page returned a 264-byte body containing `Call to a member function setSourceContext() on array in login/partials/html/login_details.html.twig at line 23`. Iterations b9357df → 9f8fe99 → 2ced9f0 → 3fcc0e4 → f7ea791 → 84bfb11 (see this branch's history) established that this Twig throw was a **downstream artifact** of an apache-worker segfault mid-render, not the root cause. Cleaner evidence in 84bfb11's `openemr.error.log`:

```
[Wed Jul 15 06:24:45 2026] [core:notice] [pid 1:tid 1] AH00052: child pid 111 exit signal Segmentation fault (11)
[Wed Jul 15 06:24:46 2026] [core:notice] [pid 1:tid 1] AH00052: child pid 123 exit signal Segmentation fault (11)
... (dozens more, one per test request)
```

This is the same class as #12438. Coredump capture from #12445 has been in place since 2026-06-10 but no core has actually been captured yet — including on this run.

## The fix in this PR

`enable_coredumps` in `ci/ciLibrary.source` sets `fs.suid_dumpable=2` and `kernel.core_pattern`. Both are kernel settings checked at the moment a process changes UID. But apache is already running at that point — its prefork workers dropped root → `apache` earlier during `docker compose up`, so their `dumpable` flag is stuck at 0 and their cores are dropped by the kernel. That's why `coredumps.none.txt` keeps reporting "No coredumps found."

`httpd -k graceful` (SIGUSR1 to the parent apache process) forks a fresh worker pool that inherits `dumpable=1` under the new sysctls. Cores land as root/mode 600 — fine, `capture_coredumps` runs `gdb` as root.

## Debug hooks retained

- `interface/login/login.php` — logs full Twig exception chain via PSR-3 (in case the segfault manifests as a Twig error again)
- `tests/Tests/E2e/Login/LoginTrait.php` — dumps browser body to STDERR on login form filter failure
- `ci/compose-shared-apache.yml` — `OPENEMR__ENVIRONMENT=dev` (no runtime effect, since `variables_order=GPCS` in the container leaves `$_ENV` empty for this key, but kept in case that changes)

All three are harmless when the failure is a native crash below PHP.

## What to check on next CI run

If this cell fails again, the `crash-diagnostics-apache_82_118_upgrade` artifact should contain `core.httpd.<pid>.<ts>.backtrace.txt` — that's the `thread apply all bt` + `info sharedlibrary` naming the faulting `.so`. Once we know the extension, root-cause work moves to either the flex-3.22 image build or the offending extension version.

Refs #12423, #12438, #12439, #12445.


bradymiller:
Superseded by #13001 which landed the actionable fix (coredump-capture that actually captures cores, apparent side-effect of resolving the flake). Closing this debug PR — its investigation is captured in #13001's body.

PR #12980, opened 2026-07-14
fix(menu): use xlt() instead of text() for patient menu labels

GustavoOtazu:
## Summary
- Replace `text()` with `xlt()` in `PatientMenuRole.php` for three menu label outputs (lines 200, 206, 214)
- `text()` only escapes output; `xlt()` escapes **and** translates — these labels were not being translated in non-English locales

## Test plan
- [ ] Switch OpenEMR to a non-English locale (e.g., Spanish)
- [ ] Open a patient's demographics page
- [ ] Verify that the patient menu bar labels are now translated

### DIFF
--- a/src/Menu/PatientMenuRole.php
+++ b/src/Menu/PatientMenuRole.php
@@ -197,13 +197,13 @@ class PatientMenuRole extends MenuRole
             if (!empty($value->children)) {
                 // create dropdown if there are children (bootstrap3 horizontal nav bar with dropdown)
                 $class = $value->class ?? '';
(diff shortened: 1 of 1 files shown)


kojiromike:
In order to identify and automatically catalog the text to translate, the `xl*()` class of functions requires its arguments to be literal strings, not variables. If those labels are to be translated, they need to be translated at definition time, not here.

adunsulag (changes requested):
Need's changes with the issue identified by coderabbitai.

PR #12986, opened 2026-07-15
fix(fax): resolve all 157 phpstan findings in fax_dispatch.php

kojiromike:
Modernizes `interface/fax/fax_dispatch.php` so it passes PHPStan level 10 with zero baseline entries, and drains its 157 entries from the baseline.

## What
- Replace `$_GET`/`$_POST` access with Symfony Request typed getters
- Replace all 10 `exec()` calls with Symfony Process using array arguments (the `mogrify *.tif` shell glob now enumerates files via `FilesystemIterator`)
- Replace deprecated `sqlStatement`/`sqlQuery`/`sqlFetchArray`/`sqlInsert`/`addForm` with `QueryUtils` and `FormService`
- Replace false-returning file APIs (`fopen`/`fread`/`tempnam`) with Symfony Filesystem
- Convert the global-namespace `getKittens()`/`mergeTiffs()` functions to an iterative traversal and a closure; drop the `global` keyword
- Lower the `variable.undefined` fatal-baseline cap 505 → 500

## Why
Beyond static-analysis hygiene, this fixes a real latent bug: `$erow` was referenced but never defined, so the "New scanned encounter note for visit on …" patient note always had an empty visit date (and is an undefined-variable error on PHP 8). The date is now queried from `form_encounter` for the selected visit.

Behavior notes: process failures now report combined stdout+stderr instead of only the last stdout line; a missing `sites/<site>/faxcover.txt` now throws instead of silently faxing a blank cover page.

### DIFF
--- a/interface/fax/fax_dispatch.php
+++ b/interface/fax/fax_dispatch.php
@@ -24,6 +24,7 @@ use OpenEMR\Core\OEGlobalsBag;
 use OpenEMR\Services\FormService;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Process\Process;
 
 $globalsBag = OEGlobalsBag::getInstance();
@@ -39,6 +40,19 @@ $session = SessionWrapperFactory::getInstance()->getActiveSession();
 $request = Request::createFromGlobals();
 $filesystem = new Filesystem();
 $userauthorized = $globalsBag->getInt('userauthorized');
+
+// Stop with a real HTTP error status instead of die(): callers and proxies
+// see the failure, and the nonzero exit fails loudly outside a web SAPI.
+// Response::send() skips the status line if headers already went out.
+$abort = static function (int $statusCode, string $message): never {
+    (new Response($message, $statusCode))->send();
+    exit(1);
+};
+
+$site_id = $session->get('site_id');
+if (!is_string($site_id) || $site_id === '') {
+    $abort(Response::HTTP_UNAUTHORIZED, "Site ID is missing from the session.");
+}
 
 $fileParam = $request->query->getString('file');
 $scanParam = $request->query->getString('scan');
@@ -64,7 +78,7 @@ if ($fileParam !== '') {
 
     $filepath = $globalsBag->getString('scanner_output_directory') . '/' . $filename;
 } else {
-    die("No filename was given.");
+    $abort(Response::HTTP_BAD_REQUEST, "No filename was given.");
 }
 
 $dotPos = strrpos($filename, '.');
@@ -89,7 +103,7 @@ $processError = (static fn(Process $process): string => (string) $process->getEx
 // This merges the tiff files for the selected pages into one tiff file.
 // $images holds the form_images checkboxes to the right of the images.
 //
-$mergeTiffs = static function (array $images) use ($faxcache, $filesystem, $runProcess, $processError): string {
+$mergeTiffs = static function (array $images) use ($faxcache, $filesystem, $runProcess, $processError, $abort): string {
     $tiffNames = [];
     foreach ($images as $name) {
         if (is_string($name)) {
@@ -98,7 +112,7 @@ $mergeTiffs = static function (array $images) use ($faxcache, $filesystem, $runP
     }
 
     if (count($tiffNames) === 0) {
-        die(xlt("Internal error - no pages were selected!"));
+        $abort(Response::HTTP_BAD_REQUEST, xlt("Internal error - no pages were selected!"));
     }
 
     // Remove any merge output from a previous run so a failed tiffcp cannot
@@ -133,7 +147,7 @@ if ($request->request->getString('form_save') !== '') {
     if ($request->request->getBoolean('form_cb_copy')) {
         $patient_id = $request->request->getInt('form_pid');
         if ($patient_id === 0) {
-            die(xlt('Internal error - patient ID was not provided!'));
+            $abort(Response::HTTP_BAD_REQUEST, xlt('Internal error - patient ID was not provided!'));
         }
 
         // If copying to patient documents...
@@ -281,7 +295,7 @@ if ($request->request->getString('form_save') !== '') {
                 // we already have a jpeg for each page in faxcache.
                 $process = $runProcess(['convert', '-resize', '800', '-density', '96', $tmp_name, '-append', $imagepath]);
                 if (!$process->isSuccessful()) {
-                    die("convert returned " . text($processError($process)));
+                    $abort(Response::HTTP_INTERNAL_SERVER_ERROR, "convert returned " . text($processError($process)));
                 }
             }
 
@@ -382,7 +396,7 @@ if ($request->request->getString('form_save') !== '') {
         if ($action_taken) {
             $dh = opendir($faxcache);
             if (! $dh) {
-                die("Cannot read " . text($faxcache));
+                $abort(Response::HTTP_INTERNAL_SERVER_ERROR, "Cannot read " . text($faxcache));
             }
 
             $form_cb_delete = '2';
@@ -411,7 +425,7 @@ if ($request->request->getString('form_save') !== '') {
         if (is_dir($faxcache)) {
             $dh = opendir($faxcache);
             if (! $dh) {
-                die("Cannot read " . text($faxcache));
+                $abort(Response::HTTP_INTERNAL_SERVER_ERROR, "Cannot read " . text($faxcache));
             }
 
             while (($tmp = readdir($dh)) !== false) {
@@ -464,9 +478,9 @@ $using_scanned_notes = is_numeric($scannedNotesCount) && (int) $scannedNotesCoun
 if (! is_dir($faxcache)) {
     // Remove the partial cache on any failure so a later request rebuilds it
     // instead of silently serving an incomplete set of pages.
-    $failCache = static function (string $message) use ($faxcache, $filesystem): never {
+    $failCache = static function (string $message) use ($faxcache, $filesystem, $abort): never {
         $filesystem->remove($faxcache);
-        die($message);
+        $abort(Response::HTTP_INTERNAL_SERVER_ERROR, $message);
     };
 
     $filesystem->mkdir($faxcache);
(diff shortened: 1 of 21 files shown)


interface/fax/fax_dispatch.php
```diff
@@ -843,6 +922,10 @@ function allCheckboxes(issel) {
 ksort($jpgarray);
 $page = 0;
 $site_id = $session->get('site_id');
+if (!is_string($site_id)) {
+    die("Site ID is missing from the session.");
```
Firehed: ❗ proper session guard/exit nonzero

interface/fax/fax_dispatch.php
```diff
@@ -7,153 +7,168 @@
...
-    $childcount = 0;
-    while ($crow = sqlFetchArray($cres)) {
-        ++$childcount;
-        getKittens($crow['id'], ($catstring ? "$catstring / " : "") .
-        ($catid ? $crow['name'] : ''), $categories);
-    }
+// Run an external command with Symfony Process so arguments are escaped
+// safely, returning the finished Process for exit-code and output checks.
+/** @param list<string> $command */
+$runProcess = static function (array $command, ?string $cwd = null): Process {
```
Firehed, line 92: I'm sad, but less sad than I was.

interface/fax/fax_dispatch.php
```diff
@@ -7,153 +7,168 @@
...
 use OpenEMR\Core\OEGlobalsBag;
+use OpenEMR\Services\FormService;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Process\Process;
+
+$globalsBag = OEGlobalsBag::getInstance();
+$srcdir = $globalsBag->getSrcDir();
+
+require_once "$srcdir/patient.inc.php";
```
Firehed, line 34: I realize these are still needed, but at least some of them I think can potentially get moved to the autoload path to remove this stuff. Out of scope for this PR.

Firehed (changes requested):
should fix the die() per previous discussions, otherwise looks ok

interface/fax/fax_dispatch.php
```diff
@@ -7,153 +7,168 @@
...
-    if ($tmp2) {
-        $msg .= "tiffcp returned $tmp2: $tmp0 ";
+    $process = $runProcess(array_merge(['tiffcp'], $tiffNames, ['temp.tif']), $faxcache);
+    if (!$process->isSuccessful()) {
+        return 'tiffcp returned ' . $processError($process) . ' ';
     }
 
-    return $msg;
-}
+    return '';
```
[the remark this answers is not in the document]
kojiromike: Fixed in 0c832d7. `$mergeTiffs` now removes any pre-existing `temp.tif` before running `tiffcp`, so a failed merge can never leave stale pages behind. The document flow only runs `tiff2pdf` when the merge succeeded, and the forward flow only runs `sendfax` when both the cover page (enscript) and the merge succeeded.

interface/fax/fax_dispatch.php
```diff
@@ -416,53 +436,105 @@ function mergeTiffs()
...
+        }
+    }
+
+    sort($pageTiffs);
+    if (count($pageTiffs) > 0) {
+        $process = $runProcess(array_merge(['mogrify', '-resize', '750x970', '-format', 'jpg'], $pageTiffs), $faxcache);
+        if (!$process->isSuccessful()) {
+            die("mogrify returned " . text($processError($process)) . "; ext is '" . text($ext) . "'; filepath is '" . text($filepath) . "'");
+        }
     }
```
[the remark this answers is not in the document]
kojiromike: Fixed in 0c832d7. All failure paths in the cache build (`convert`, both `tiffsplit` invocations, `mogrify`) now remove the partially built `$faxcache` before terminating, so the next request rebuilds it from scratch.

kojiromike:
Addressed the CodeRabbit findings in 0c832d7:

- **Stale merge output / dependent commands** (inline, critical): `temp.tif` is now removed before every merge, and `tiff2pdf`/`sendfax` are gated on merge and cover-page success.
- **Partial faxcache on build failure** (inline, major): every failure path in the cache build now removes the partial `$faxcache` before terminating.
- **Archive/delete result unchecked** (outside diff): the delete-all branch now uses the throwing `Filesystem::rename()`/`remove()`, so the cache erase can no longer run when the source fax was not actually archived or deleted.
- **Scanned-note flow persists records before generating the image** (outside diff, heavy lift): pre-existing behavior and a design change (validation ordering + transactionality), so split into a follow-up: #12988.
- **Inject `FormService` instead of `new`** (nitpick): agreed in principle, but this is a legacy procedural entry script with no constructor to inject into — doing this properly means converting the script to a controller, which is beyond this PR's PHPStan-cleanup scope and tracked by the broader backend-modernization effort.

interface/fax/fax_dispatch.php
kojiromike: Fixed in 4c43f19. The site_id guard is hoisted to the top of the script (before any output) and now returns a 401 via a Symfony Response instead of die(). All other error paths go through the same `$abort` helper: 400 for bad input, 500 for command/filesystem failures, and it exits nonzero. The only remaining `die()`s on this page's path are inside `check_file_dir_name()` in library/sanitize.inc.php — happy to clean that up separately since it has many callers.

interface/fax/fax_dispatch.php
kojiromike: Agreed — shelling out to tiffcp/enscript/sendfax at request time is inherently sad; Process with array args at least removes the injection surface and the lost-exit-code problem. A proper fax service abstraction is the real fix, but that's a rewrite of this whole page rather than a PHPStan pass.

interface/fax/fax_dispatch.php
kojiromike: Agreed, out of scope here — several of these (pnotes, forms, gprelations) could become autoloadable service calls. This PR already swaps `addForm()` for `FormService::addForm()`; the remaining requires are for functions this page still calls directly (addPnote, fixDate, generate_form_field, check_file_dir_name).

interface/fax/fax_dispatch.php
kojiromike: Follow-up in df9b23e: swapped the exiting helper for `$sendError` + top-level `return`. Same 4xx/5xx response, but nothing kills the process anymore, so the script stays includable/testable. The faxcache build now returns an error string to the top level instead of terminating from inside a closure.

kojiromike:
Extracted the error-ending pattern into a reusable `OpenEMR\Common\Http\RequestTerminator` (03402da). It sends a Symfony Response, then terminates via an injectable closure: the default exits the process with `(int) (status >= 400)`, and tests inject a non-exiting closure so request-ending paths remain executable under PHPUnit (isolated unit tests included). fax_dispatch.php now uses it; other legacy scripts can adopt it in follow-up PRs as an alternative to die().

src/Common/Http/RequestTerminator.php
```diff
@@ -0,0 +1,68 @@
...
+ * is an error (>= 400), zero otherwise. Tests inject their own closure
+ * (throw or record) so request-ending code paths remain executable under
+ * PHPUnit instead of killing the test runner.
+ */
+final readonly class RequestTerminator
+{
+    private Closure $terminator;
+
+    /**
+     * @param (Closure(int): void)|null $terminator receives the intended
```
Firehed:
```suggestion
     * @param (Closure(int): never)|null $terminator receives the intended
```

src/Common/Http/RequestTerminator.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Send the response, then terminate the request.
+     *
+     * With the default terminator this never returns. Top-level script
+     * callers should still follow it with `return;` so control flow also
+     * ends when a test injects a non-exiting terminator.
+     */
+    public function respond(Response $response): void
```
Firehed:
```suggestion
    public function respond(Response $response): never
```

src/Common/Http/RequestTerminator.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $response->send();
+        ($this->terminator)((int) ($response->getStatusCode() >= 400));
+    }
+
+    /**
+     * Convenience wrapper for the common case: an error body with status.
+     */
+    public function error(int $statusCode, string $message): void
```
Firehed:
```suggestion
    public function error(int $statusCode, string $message): never
```

Firehed (commented):
I think you can tighten up this approach by forcing the terminating closure to `return never`, but otherwise this seems like a good improvement to the legacy codepaths. Might be worth discouraging its use in _new_ code in favor of proper response building/handling, though.

src/Common/Http/RequestTerminator.php
kojiromike: Applied in f6fe6bff5f — the closure is now typed `Closure(int): never` and both `respond()` and `error()` declare native `: never`. The test double throws instead of recording, and the now-dead `return;` statements after `error()` calls in fax_dispatch.php are removed.

src/Common/Http/RequestTerminator.php
kojiromike: Applied in f6fe6bff5f — the closure is now typed `Closure(int): never` and both `respond()` and `error()` declare native `: never`. The test double throws instead of recording, and the now-dead `return;` statements after `error()` calls in fax_dispatch.php are removed.

src/Common/Http/RequestTerminator.php
kojiromike: Applied in f6fe6bff5f — the closure is now typed `Closure(int): never` and both `respond()` and `error()` declare native `: never`. The test double throws instead of recording, and the now-dead `return;` statements after `error()` calls in fax_dispatch.php are removed.

kojiromike:
Rebased onto master (the only conflict was the `variable.undefined` fatal-baseline cap: master moved 505→504 while this branch removed 5 fax entries, so the merged cap is 499 — verified against the merged baseline and the caps test passes).

@Firehed's suggestions are applied in f6fe6bff5f: the terminator closure is typed `Closure(int): never` and `respond()`/`error()` declare native `: never`. Consequences carried through:
- The test double now throws (a `never` closure can't return), and the tests assert the exit code via the exception.
- The `return;` statements after `$terminator->error()` calls in fax_dispatch.php were removed — with `never` they're dead code and PHPStan sees the control flow directly.
- Per the suggestion to discourage use in new code, the class docblock now says explicitly: this exists to retrofit legacy procedural scripts; new endpoints should build and return a Response through a controller.

The CodeRabbit nitpick (unused `$exitCode` in a test closure) is moot — that closure was removed in the `never` restructuring.

Firehed (approved):

PR #12987, opened 2026-07-15
refactor(networking): Set up HTTP client in ServiceContainer and DI

Firehed:
There's a mix of networking paths for external calls in OpenEMR:

- direct cURL calls (~12)
- `oeHttp` family (rarely used)
- directly using Guzzle (~20)
- file_get_contents (hard to skim what's networking vs disk; both are problems)
- several other network-request libraries are installed that I didn't check touch-points

This inconsistency leads to poor enforcement of good defaults: timeouts, certificate validation, error detection and handling, and more. It also makes adding middleware to the request lifecycle _extremely_ ad-hoc (example use-cases: operational telemetry, audit logging, dealing with proxies). Having spot-checked a few implementations, it looks like _most_ of the non-Guzzle locations are at risk of runtime errors at minimum. Effectively all of them cannot be unit-tested.

There's already a migration guide in place touching on this, but it predates some recent refactoring. Plus, as written, most points are still not easily testable as it encouraged directly instantiating the client. This aims to address all of those by making the preferred tools easy to use and updating documentation to encourage it. "Pit of success" and all that.

> [!NOTE]
> This adds PSR-18 as well as Guzzle's ClientInterface intentionally. The former being framework-agnostic is generally preferred, but Guzzle adds _a lot_ of ergonomics that make it worth exposing directly (basically most of what oeHttp aims to do, but more complete and it's not something to maintain). Since it's already installed and the de-facto library for basically _every_ modern PHP project, I'm not overly worried about the soft coupling.

- Adds PSR-18's ClientInterface and Guzzle's ClientInterface accessors to ServiceContainer and the DI container
- Updates oeHttp's internals to use it, and cleans up some of its types for correctness
- Update documentation to push towards either of the interfaces, as needed, and discourage oeHttp
- Adds `ErrorResponseException::throwIfError($response)` to handle the common-case under the PSR-18 requirements

PHPStan mostly a net win; I've improved type safety and in doing so revealed some other at-risk areas. I fixed some unambiguous ones and left one in place since this is "detect existing error" not "adding a new error".

Mostly bouncing ideas, some boilerplate for the error promotion tool


src/BC/ServiceContainer.php
```diff
@@ -133,6 +139,41 @@ public static function getCrypto(): Crypto\CryptoInterface
...
+            GuzzleClientInterface::class,
+            static fn() => new Client([
+                // See config/services.php for details
+                RequestOptions::ALLOW_REDIRECTS => true,
+                RequestOptions::CONNECT_TIMEOUT => 5,
+                RequestOptions::HTTP_ERRORS => false,
+                RequestOptions::TIMEOUT => 15,
+            ]),
+        );
+    }
```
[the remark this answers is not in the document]
Firehed: This cannot be fixed as-is, see comment.

src/Common/Http/oeHttpRequest.php
```diff
@@ -143,22 +149,22 @@ public function setParams($params)
         ]));
     }
 
-    public function usingBaseUri($baseUri)
+    public function usingBaseUri(string $baseUri)
     {
-        $baseUri = str_ends_with((string) $baseUri, '/') ? $baseUri : $baseUri . '/';
```
kojiromike, line 148: This leftover cast is what's failing the Rector CI check (`RecastingRemovalRector`) now that `$baseUri` is natively `string`:

```suggestion
        $baseUri = str_ends_with($baseUri, '/') ? $baseUri : $baseUri . '/';
```

src/Common/Http/oeHttpRequest.php
```diff
@@ -26,7 +27,9 @@ class oeHttpRequest extends oeHttp
     private string $bodyFormat;
     private array $options;
 
-    public function __construct($client)
+    private static ClientInterface $client;
```
kojiromike: Does this need to be `static`? Every construction reassigns it, so with two live `oeHttpRequest` instances the last-constructed one's client wins for both (`send()` reads `self::$client`). Today it's harmless because every caller passes the same shared `ServiceContainer::getGuzzle()` instance, but an instance property would remove the last-writer-wins global state entirely — which feels in the spirit of this PR, given it deletes `oeHttp::$client` for the same reason.

```suggestion
    private ClientInterface $client;
```

(with `$this->client = $client;` in the constructor and `$this->client->request(...)` in `send()`)

src/Common/Http/oeOAuth.php
```diff
@@ -12,8 +12,11 @@
 
 namespace OpenEMR\Common\Http;
 
-use GuzzleHttp\Client;
-use GuzzleHttp\HandlerStack;
+use GuzzleHttp\{
+    Client,
+    ClientInterface,
```
kojiromike: `ClientInterface` isn't referenced anywhere in this class (`oeHttp`/`oeHttpRequest` import it themselves) — leftover?

```suggestion
```

src/BC/ServiceContainer.php
```diff
@@ -133,6 +139,41 @@ public static function getCrypto(): Crypto\CryptoInterface
...
+     * 1) Async request handling options
+     * 2) Request-specific options (an array keyed by RequestOptions constants)
+     *
+     * While we generally prefer binding to vendor-agnostic interfaces,
+     * pragmatism wins out here - there's often a _lot_ of boilerplate in
+     * constructing all of the PSR messages.
+     */
+    public static function getGuzzle(): GuzzleClientInterface & ClientInterface
+    {
+        /** @phpstan-ignore return.type (can't represent the native union with the generics in an array key) */
```
kojiromike: Could this narrow with `instanceof` instead of suppressing? `override()` only validates against `GuzzleClientInterface`, so a registered double that doesn't also implement PSR-18 would currently surface as a runtime `TypeError` at the `return` (native return types are enforced). An explicit check keeps PHPStan clean without the ignore and turns that edge case into a clear failure:

```php
$client = self::resolveOrCreate(
    GuzzleClientInterface::class,
    static fn() => new Client([/* ... */]),
);
if (!$client instanceof ClientInterface) {
    throw new \LogicException('HTTP client override must also implement PSR-18 ClientInterface');
}
return $client;
```

config/services.php
```diff
@@ -89,6 +94,21 @@
...
+    // ClientInterface for its convenience APIs. At present, these resolve the
+    // same thing, but the PSR-18 could (if for some reason we wanted to)
+    // provide a different implementation in the future.
+    ClientInterface::class => Client::class,
+    Client::class => fn () => new Client([
+        // PSR-18 makes no rule about following redirects either way
+        RequestOptions::ALLOW_REDIRECTS => true,
+        // Establish _some_ baseline timeouts
+        RequestOptions::CONNECT_TIMEOUT => 5,
+        RequestOptions::TIMEOUT => 15,
```
kojiromike, line 107: Question on the 15s ceiling as it applies to the pre-existing `oeHttp` call sites, which previously ran with no timeout at all: `ImportPharmacies` hits the NPI registry (`npiregistry.cms.hhs.gov` — the same endpoint the code comments already describe as operationally quirky), and the faxsms module downloads fax media through `FaxDocumentService::fetchMediaBytes()`. If either regularly runs long, they'd now fail mid-operation.

Baseline timeouts are the right default — just wondering whether those two call sites deserve a per-request `RequestOptions::TIMEOUT` override in this PR, or whether "they should fail fast and we'll tune if it bites" is the intended stance.

src/Common/Http/ErrorResponseException.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+/**
+ * Caller-side exception for error HTTP responses (>= 400).
+ *
+ * PSR-18 clients must not throw on 4xx/5xx, so this is opt-in for callers that
+ * prefer exceptions over inspecting the status. It implements the base
+ * ClientExceptionInterface rather than RequestExceptionInterface, which would
+ * require the originating request this only has the response.
+ *
```
kojiromike, line 10:
```suggestion
 * require the originating request; this only has the response.
```

kojiromike (approved):
Nice cleanup — consolidating on a container-provided client with sane defaults is exactly the right direction, and `ErrorResponseException` + its test coverage look solid.

A few items inline. Two are quick mechanical fixes (the leftover cast that's failing the Rector CI check, and an unused import); the rest are questions/suggestions on the new `ServiceContainer` accessors and the timeout defaults.

One test-coverage note that doesn't fit an inline comment: `tests/Tests/Isolated/BC/ServiceContainerTest.php` has a smoke test for every existing accessor, but the two new ones (`getGuzzle()`, `getHttpClient()`) don't have entries. Cheap to add following the existing pattern — and asserting `getHttpClient()` returns the same instance as `getGuzzle()` would pin the intended shared-instance behavior. (`ErrorResponseException` itself is well covered by the new isolated test.)

src/Common/Http/oeHttpRequest.php
Firehed: I was avoiding changing that, but yeah, that's probably the better choice given how it works in practice.

src/BC/ServiceContainer.php
Firehed: oh, actually that probably would work! will try.

config/services.php
Firehed: > "they should fail fast and we'll tune if it bites" is the intended stance.

Yes, that was my plan.

Firehed: Also, the "quirks" in ImportPharmacies _probably_ don't apply anymore - they had a goofy cert some six years ago based on git history; I'd really hope it's been fixed since then. Haven't tested though.

